PR #2124, opened 2023-02-01
Adjust build min test to be more frugal

zhangyiatmicrosoft:
## Launch Checklist

Previous check was +-5% of 750KB, this means each PR was allowed to change size up to 37KB minified!!!

Based on our data of millions of users, 1k size increasement usually means +1ms of load time, and each change needs to be justified.

Should not use percentage because the net byte change will get bigger and bigger.

Changes:
- Allow up to 1K increasement per PR. Or it needs manual attention. Practically, 1K minified usually means ~4k of debug scripts, and that is already a lot of code for one PR.
- Allow up to 4K reduction. This is just to catch bugs that may have dropped things unintentionally.

### DIFF
--- a/test/build/min.test.ts
+++ b/test/build/min.test.ts
@@ -34,11 +34,20 @@ describe('test min build', () => {
     });
 
     test('bundle size stays the same', async () => {
-        const bytes = (await fs.promises.stat('dist/maplibre-gl.js')).size;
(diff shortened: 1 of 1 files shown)


HarelM (approved):
Let's see when and how it fails.

PR #2126, opened 2023-02-02
Using HTMLImageElement for raster images

pramilk:
## Issues
1. Perf:  Currently all raster images are fetched using fetch instead of HTMLImageElement. It's an expensive operation with following code flow fetch(url) --> response.arrayBuffer() --> create blob from arrayBuffer --> createImageBitmap(blob) --> imageBitmap.
2. As per previous comments in code, it's using fetch due to an issue described here: https://github.com/mapbox/mapbox-gl-js/issues/1470
3. HTMLImageElement can't be used in all cases as it does not provide cache information as required when  options.refreshExpiredTiles is true (default is true), 
4. There is no option to override refreshExpiredTiles at a tile source level. In our case (Bing Maps), only traffic tiles requires refresh.

## Fix
1. Expose a refreshExpiredTiles property for vector/raster/raster-dom source, which will override the default maps options.refreshExpiredTiles, if its set to true/false.
2.  Use HTMLImageElement element to download raster/raster-dom tiles when refreshExpiredTiles is false

## Alternative design option
- Instead of having a boolen refreshExpiredTiles on tile sources, we can have a property like expiryDuration, so that these tiles can also be downloaded using HTMLImageElement.
  - Issue with this approach is decoupling of cache data between tiles and style endpoints. This might cause issues in some cases.

## Perf: 
1. https://jsbench.me/flldmrgkeb/1
2. On [Bing maps](https://www.bing.com/maps) A/B testing using HTMLImageElement instead of fetch for raster tiles, showed a substantial improvement. PLT - Page load time.
image

## Pending:
I have not added any tests in this draft PR.

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
Three's a PR that changes the implantation of image queue, it's in advanced stages.
Also there's a PR to move to promises.
All are conflicting, so this will have to wait for the others I believe.
Thanks for taking that time to contribute!

pramilk:
> There's a PR that changes the implantation of image queue, it's in advanced stages. Also there's a PR to move to promises. All are conflicting, so this will have to wait for the others I believe. Thanks for taking that time to contribute!

I see imageRequest changes were already commited (merged with it). Promises PR are still pending, but looking at its changes, there is no conflicts with this PR.

Can you please take a look at this PR. If it looks good then I can add tests.

HarelM:
It looks like there are two features in this PR which is making it hard to review. Can you split the two (html image element and "force" refresh)

pramilk:
Added design proposal for the changes in the sources JSON.  Will hold on to this PR, till the specs are updated.

https://github.com/maplibre/maplibre-gl-style-spec/pull/59

pramilk:
Simplified this PR. Will handle specification changes to allow tile refresh per tile source as per [this design proposal](https://github.com/maplibre/maplibre-gl-style-spec/issues/61) separately. Also added unit tests. Please take a look.

src/util/image_request.test.ts
```diff
@@ -31,72 +15,6 @@ describe('ImageRequest', () => {
         server.restore();
     });
 
-    test('getArrayBuffer, 404', done => {
```
pramilk, line 34: These deleted tests already exist in [ajax.test.ts](https://github.com/maplibre/maplibre-gl-js/blob/99ad3d7d0881cfa4b29207d9789c572bbc84fd0f/src/util/ajax.test.ts)

pramilk:
@HarelM, Can you please take a relook at this PR.

HarelM:
Thanks for pinging, I'll look into it in the next few days...

src/util/image_request.ts
```diff
@@ -139,35 +143,64 @@ namespace ImageRequest {
...
-                if (!isThrottled()) {
-                    processQueue();
+    const onImageResponse = (
+        itemInQueue: ImageRequestQueueItem,
+        callback:GetImageCallback,
+        err?: Error | null,
+        data?: HTMLImageElement | ArrayBuffer | null,
+        cacheControl?: string | null,
+        expires?: string | null): void => {
+        if (err) {
```
HarelM, line 178: consider simplifying all this else-if and use return statements.

src/util/image_request.ts
```diff
@@ -139,35 +143,64 @@ namespace ImageRequest {
...
+            callback(err);
+        } else if (data instanceof HTMLImageElement) {
+            callback(null, data);
```
HarelM, line 181: is this case vaild only in the new code that was added or in other cases? I seem to recall that getArrayBuffer can use an image in some cases?

src/util/image_request.ts
```diff
@@ -214,6 +247,37 @@ namespace ImageRequest {
 
         return imageRequestQueue.length;
     };
+
+    const makeImageRequest = (requestParameters: RequestParameters, callback: GetImageCallback): Cancelable  => {
```
HarelM, line 251: Consider renaming this method to be more like `getArrayBuffer` - maybe `getImageUsingHtmlImage`, IDK, this name is a bit too generic and confuses with the doImageRequest which was added/renamed...

HarelM:
I'm not sure what was the coverage before this change, but after the change it is:
```
image_request.ts                           |   95.37 |     87.8 |   90.47 |   95.28 | 185,237,257,275-277 
ajax.ts                                    |    54.7 |    34.92 |      48 |   53.84 | 91,100-170,190,222-228,233,236-237,269,272-277
```
So it would be better to make sure we cover all `image_request.ts` file as part of this change and that ajax tests removal don't reduce coverage there.

src/util/image_request.ts
pramilk: Yes this is for only New code. Here is the flow
New Code:
GetImage(..) --> doImageRequest(..) --> onImageResponse()

Old Code:

GetImage(..) --> doImageRequest(..) --> onImageResponse() --> arrayBufferToCanvasImageSource() --> 
if  createImageBitmap supported {arrayBufferToImageBitmap() --> createImageBitmap() }
else { arrayBufferToImage()}

arrayBufferToImage is the function which contains the code you are referring about.

src/util/image_request.ts
pramilk: These nested loops are coming from old code. Likely will have to add promise to simplify it, which is out of scope of this PR.

src/util/image_request.ts
pramilk: Good suggestion. Updating.

pramilk:
> Thanks for taking the time to open this PR! I'm not sure what was the coverage before this change, but after the change it is:
> 
> ```
> image_request.ts                           |   95.37 |     87.8 |   90.47 |   95.28 | 185,237,257,275-277 
> ajax.ts                                    |    54.7 |    34.92 |      48 |   53.84 | 91,100-170,190,222-228,233,236-237,269,272-277
> ```
> 
> So it would be better to make sure we cover all `image_request.ts` file as part of this change and that ajax tests removal don't reduce coverage there.

There is a 1-2% drop in test coverage. Will update tests for missing coverage.

src/util/image_request.ts
HarelM: I would consider splitting this method instead of using it twice.
If I understand correctly - when using `getImageUsingHtmlImage` I always get a `HtmlImageElement` back (or an error, but let's leave the error a side for a second) so the rest of this method is irrelevant, right?
Same goes for the `getArrayBuffer` - I always get an array and not HtmlImageElement - so this method is confusing, I think.

pramilk:
Looks like it will be a issue in "REGISTERED_PROTOCOLS" scenario. Will look into it more.

HarelM:
Make sure to add the relevant tests so that future changes will be covered by these new tests.

pramilk:
This PR is now ready for review. Have added some more code coverage and addressed your previous concerns. Fix for addProtocol to handle images is merged (https://github.com/maplibre/maplibre-gl-js/commit/538fe6845242b9ce1b4de5644381f07b5d4779ce).

src/util/ajax.ts
```diff
@@ -91,6 +91,8 @@ export const getReferrer = isWorker() ?
     () => (self as any).worker && (self as any).worker.referrer :
     () => (window.location.protocol === 'blob:' ? window.parent : window).location.href;
 
+export const getProtocolAction = url => config.REGISTERED_PROTOCOLS[url.substring(0, url.indexOf('://'))];
```
HarelM, line 94: Why is this exported?

pramilk:
@HarelM, Please take a look at this PR.

HarelM:
There are comments I left which are still unanswered and a conflict. Please let me know when these are addressed/answered. Thanks! :-)

pramilk:
I went back to check, if there is any unanswered questions and can't find any. If you could point to any unanswered question, I will be happy to answer it. Conflit was in changeLog. Update it.

HarelM:
https://github.com/maplibre/maplibre-gl-js/pull/2126#discussion_r1168330256
https://github.com/maplibre/maplibre-gl-js/pull/2126#discussion_r1163328621

pramilk:
> [#2126 (comment)](https://github.com/maplibre/maplibre-gl-js/pull/2126#discussion_r1168330256) 
> [#2126 (comment)](https://github.com/maplibre/maplibre-gl-js/pull/2126#discussion_r1163328621)
I have responsed to both of them. Please let me know, if there is any additional details required.

HarelM:
I don't see where (It might be somewhere here in the main thread, but I'll have a hard time finding the right one), can you respond in the relevant thread?
image

pramilk:
> I don't see where (It might be somewhere here in the main thread, but I'll have a hard time finding the right one), can you respond in the relevant thread? image

I did replied in the thread it self and I can see it there. Not sure why its not showing up for you.

image

pramilk (commented):
Updated.

src/util/image_request.ts
HarelM: I think it's the other way around, seems to me like this method violates the SOLID principle of single responsibility, as only half of the method is relevant to the calling callback, which opens up the possiblity of adding bugs when trying to change this method.
Having said that, it's not that bad that it can't pass the review. So please let me know if you want to keep it this way or split.

HarelM (approved):

src/util/image_request.ts
pramilk: If you have a strong feeling against current approach, I can update it. Otherwise I will like to keep it as is.  Please let me know.

HarelM: I think you did a great job, this is more in the area of nit-picking.

PR #2131, opened 2023-02-03
Improve performance by processing style layers in parallel

pramilk:
## Change

Style processing (createStyleLayer method) is one of the most expensive operations performed during initial load and also during style change.  This takes anywhere from 100-400ms, depending on machine speed, number of layers, expressions used in style etc.

- Currently: Styles are validated (If validation is enabled) --> Styles are processed on main thread (via createStyleLayer function) --> broadcasted to all worker threads --> Styles are processed on each worker thread (via createStyleLayer function). 

- New change: Styles are validated (If validation is enabled) --> broadcasted to all worker threads --> Styles are processed in parallel (via createStyleLayer function) on both main and worker thread.

Note: In our observation doing a prewarm or an external worker as per this PR (https://github.com/maplibre/maplibre-gl-js/pull/2130), gives much better results as worker threads are ready by the time it gets the broadcast message to truly allow parallel processing.

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


src/style/style.ts
```diff
@@ -319,8 +319,11 @@ class Style extends Evented {
 
         const layers = deref(this.stylesheet.layers);
 
-        this._order = layers.map((layer) => layer.id);
+        // Broadcast layers to workers first, so that expensive style processing (createStyleLayer)
+        // can happen in parallel on both main and worker threads.
+        this.dispatcher.broadcast('setLayers', layers.filter(layer => layer.type !== 'custom'));
```
HarelM, line 324: This logic is not exactly the same - there's a custom layer filtering here. Can you elaborate on the reasoning behind this?

pramilk: Previous code at line 332 (old code) was calling 'this._serializeLayers(this._order)' and if you look its implementation at line 447, its removing the custom layers.

HarelM: But then it means it's not sending the serialized layers, this is strange to me...

pramilk: Here is a bit more detailed explanation:
Previous (Old Code): 
1.1. JSON Style object is validated (If validation is enabled) - Line 301
1.2. Each Layer in Style Json is converted to a layer object (via createStyleLayer function) - Line 327
1.3. Layer object is serialized back to Layer Json - Line 330 (Done in [style_layer.ts](https://github.com/maplibre/maplibre-gl-js/blob/51054671229c68d798fa06a64968aa35688f6a0f/src/style/style_layer.ts) line 217)
1.4. All of these searialized Layer Json is broadcasted to Workers (Except Custom Layers) - Line 332
1.5. On Worker - Layer Json is again created back to Layer object using the same createStyleLayer  function

New code
2.1. Json Style object is validated (If validation is enabled) - Line 301
2.2. All of Layer Json is broadcasted to Workers (Except Custom Layers) - Line 324
2.3. Each Layer in Style Json is converted to a layer object (via createStyleLayer function) - Line 330
2.4. On Worker - Layer Json is again created back to Layer object using the same createStyleLayer function

Looking into layer.serialize() method (Done in [style_layer.ts](https://github.com/maplibre/maplibre-gl-js/blob/51054671229c68d798fa06a64968aa35688f6a0f/src/style/style_layer.ts) line 217)
1. New json created removes any empty paint/layout json.  - Not a issue
2. this.Visibility will be undefined as its only set after setLayoutProperty is called. In our case Serialization is happening before setLayoutProperty can be called on a layer.


I compared JSON generated (Json sent to worker) before and after the change for all bing maps styles and default Map libre style. All of them were exactly same. Bing maps styles did showed difference for some of layer.paint JSON, as some of the layers had empty paint Json ( `paint: {}`) which is not a issue.

Its actually safer to have same Json on both Worker and Main thread. Old code do pause a potential of error, if somehow layer.serialize() method creates a different Json and worker will be potentially out of sync.

HarelM: Oh, I see what confused me. Never mind...

HarelM:
Can you please add a test to test that somehow?

pramilk:
> Can you please add a test to test that somehow?

Added some tests. Also while testing realized that "custom" filter is not required as a style Json does not support it. Added some more tests for it.

HarelM (approved):

HarelM:
THANKS!

PR #2152, opened 2023-02-07
Add method to enable/disable cooperative gestures

### LINKED ISSUE #1488: Disable cooperative gestures in fullscreen
The `cooperativeGestures` feature (https://github.com/maplibre/maplibre-gl-js/issues/234) is great! It is super helpful when integrating a map in a website.
It can be even more useful when it automatically disables itself after the user put the map in full-screen mode (via clicking on a `maplibregl.FullscreenControl()` button).

Because once the map is in full-screen mode the user won't accidentally move the map since it is the only thing visible on the screen. It only makes the map harder to use at this point.

So my suggestion is to have the map in full-screen mode just work as if the cooperative gestures were never enabled in the first place. When the full-screen is closed, cooperative gestures should again function normally.

### LINKED ISSUE #2057: Add a method to enable / disable cooperativeGestures
## Motivation

I have a button on my map to toggle a full screen mode (position:fixed; width+height:100%).

My use case is that I need to have cooperativeGestures enabled when my map is inlined somewhere in my page, but I want to disable it when the full screen mode is enabled.

There is currently no clean way to enable or disable cooperativeGestures programmatically.

This is the dirty way I have to do when I toggle cooperativeGestures :
```
if (fullscreen) {
	// This part is not optimal, but it's OK
	map._cooperativeGestures = false;

	// This part is really uggly...
	map.handlers._handlersById.touchPan._minTouches = 1;
} else {
	map._cooperativeGestures = true;
	map.handlers._handlersById.touchPan._minTouches = 2;
}
```

I can live with that, but I won't sleep well :smiling_face_with_tear: 

## Design Alternatives

We could have something like that :
```
map.enableCooperativeGestures();
map.disableCooperativeGestures();
```

Or (same way than interactions) :
```
map.cooperativeGestures.enable();
map.cooperativeGestures.disable();
```

If it's too complicated, please at least improve the part in **TouchPanHandler**. Instead of computing **TouchePanHandler._minTouches** in the constructor, compute it on the fly by checking if cooperativeGestures are still enabled or not.

kevinschaul:
**Add method to enable/disable cooperativeGestures** (fixes #2057)

**Disable cooperative gestures in fullscreen, using these new methods** (fixes #1488)

**Add a debug page for cooperative gestures** because the feature cannot
be tested with a fullscreen map, and the only other test page is
a fullscreen map.

**Change how meta key is detected for cooperative gestures**

UIEvents like 'wheel' include properties for whether some keys are
currently pressed, including ctrl and meta. [0] This should be less
prone to error, specifically when the user presses or depresses one of
these keys while the browser document is not active.

[0] https://w3c.github.io/uievents/#dom-mouseevent-ctrlkey

## Launch Checklist

 - [x] Link to related issues.


src/ui/control/fullscreen_control.ts
```diff
@@ -5,6 +5,7 @@ import {warnOnce} from '../../util/util';
 import {Event, Evented} from '../../util/evented';
 import type Map from '../map';
 import type {IControl} from './control';
+import type {GestureOptions} from '../map';
```
HarelM, line 8: Can this be unified with the above import from the map file?

src/ui/map.ts
```diff
@@ -301,7 +301,7 @@ class Map extends Camera {
     _interactive: boolean;
     _cooperativeGestures: boolean | GestureOptions;
     _cooperativeGesturesScreen: HTMLElement;
-    _metaPress: boolean;
+    _metaKey: string;
```
HarelM, line 304: The type of this field is not `string` but rather `keyof MouseEvent` or something similar.
Also I'm not sure this change brings a lot of value. I personally don't like to "poke" inside object properties using a string that is defined no where near where it is used. It creates a problem when you try to refactor or "find usages" in the IDE.

src/ui/map.ts
```diff
@@ -2599,35 +2626,37 @@ class Map extends Camera {
...
             desktopMessage = typeof this._cooperativeGestures !== 'boolean' && this._cooperativeGestures.macHelpText ? this._cooperativeGestures.macHelpText : 'Use ⌘ + scroll to zoom the map';
-            modifierKeyName = 'Meta';
+            this._metaKey = 'metaKey';
         }
         const mobileMessage = typeof this._cooperativeGestures !== 'boolean' && this._cooperativeGestures.mobileHelpText ? this._cooperativeGestures.mobileHelpText : 'Use two fingers to move the map';
         this._cooperativeGesturesScreen.innerHTML = `
             <div class="maplibregl-desktop-message">${desktopMessage}</div>
             <div class="maplibregl-mobile-message">${mobileMessage}</div>
         `;
-        document.addEventListener('keydown', (event) => {
```
HarelM, line 2617: But I see how this reduces the need for these event handlers... Not sure what's the best solution then...

src/ui/map.ts
```diff
@@ -2599,35 +2626,37 @@ class Map extends Camera {
         this._container.addEventListener('scroll', this._onMapScroll, false);
     }
 
+    _cooperativeGesturesOnWheel(e) {
```
HarelM, line 2629: What type is `e`?

src/ui/map.ts
```diff
@@ -2599,35 +2626,37 @@ class Map extends Camera {
...
+        this._canvasContainer.addEventListener('wheel', this._cooperativeGesturesOnWheel, false);
         // Remove the traditional pan classes
         this._canvasContainer.classList.remove('maplibregl-touch-drag-pan');
     }
 
+    _destroyCooperativeGestures() {
+        DOM.remove(this._cooperativeGesturesScreen);
+        this._canvasContainer.removeEventListener('wheel', this._cooperativeGesturesOnWheel, false);
+        // Add the traditional pan classes back
+        this._canvasContainer.classList.add('maplibregl-touch-drag-pan');
```
HarelM, line 2657: Is it possible that this was not defined in the first place and now we are adding it "by mistake"? meaning in the setup we remove this class, but we don't know that it was there to begin with, now we are adding this class which might not exist in the first place...

test/debug-pages/cooperative-gestures.html
```diff
@@ -0,0 +1,157 @@
+<!DOCTYPE html>
```
HarelM, line 1: No need to add debug pages, feel free to add a documentation page to the docs repo after this gets merged.

HarelM:
Amazing work! Thanks!!
I've added some comments for you to look at.
Also the tests are failing.
Ping me if you need me to run the tests on the CI.

src/ui/map.ts
kevinschaul: @ewagstaff do you remember why cooperative gestures needs to remove the touch-drag-pan class? https://github.com/maplibre/maplibre-gl-js/commit/557cbcbf99db4a3d03765baa3356252c9dc2917b#diff-827a9d3af39afc00e2d3fc7d6cf4d8b164a9f7f1d8d8ecd0fdde83ba888a6c83R2461

ewagstaff: Testing on my branch, if I comment out the class removal, mobile users get stuck on the map while scrolling down the page with one finger. There might have been a better way to handle that problem!

src/ui/map.ts
```diff
@@ -2599,35 +2625,39 @@ class Map extends Camera {
         this._container.addEventListener('scroll', this._onMapScroll, false);
     }
 
+    _getMetaKey() {
```
HarelM, line 2628: This still needs a return value - `keyof MouseWheelEvent` or something similar.
Also I'm not sure if it's not better to "chace" it like you did previously, as this will be called every hover, mouse movement, click, scroll etc I believe...

src/ui/map.ts
kevinschaul: Thanks, that's helpful! Will keep digging into it.

kevinschaul: I've adjusted this to apply the relevant css rule to a new class (`.maplibregl-cooperative-gestures`). I think that makes it a bit more clear, and makes removing the class easy in the destroy function. What do you think about this approach @HarelM?

src/ui/map.ts
kevinschaul: Thanks, I just added the return value. I think this will only be called every wheel event when cooperative gestures is enabled. I don't see where else it would be called.

I agree this is a bit odd though. I need some way to keep track of the right meta key, and share that information with the scroll_zoom handler. I'm not sure how to do that besides storing it on Map somehow -- either as a property or a function that returns the property.

HarelM: If this is needed only in the handler, I would recommend storing it there when you construct the handler, maybe...

kevinschaul:
https://github.com/WPMedia/gfx-maplibre-basemaps/tree/maperture

src/ui/map.ts
kevinschaul: It's needed in both map and the scroll_zoom handler. But I think I've found a clearer way to define and initialize it.

src/ui/map.ts
HarelM: LGTM

HarelM:
Overall looks good. Can you explain why the extra pan-x pan-y are needed now when they were not needed before?

src/css/maplibre-gl.css
```diff
@@ -53,6 +53,11 @@
     touch-action: none;
 }
 
+.maplibregl-canvas-container.maplibregl-touch-drag-pan.maplibregl-cooperative-gestures,
+.maplibregl-canvas-container.maplibregl-touch-drag-pan.maplibregl-cooperative-gestures .maplibregl-canvas {
+    touch-action: pan-x pan-y;
```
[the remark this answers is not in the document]
kevinschaul: Previously the cooperative gestures code removed the class `.maplibregl-touch-drag-pan` so that `touch-action: pan-x pan-y;` would be applied: https://github.com/maplibre/maplibre-gl-js/blob/7366b5c29cc664f694ab90869845398dabb5dcd5/src/css/maplibre-gl.css#L43

I thought adding this class specific to cooperative gestures was a bit more clear. This way `_destroyCooperativeGestures()` can remove this class without needing to know whether to add the `.maplibre-touch-drag-pan` class back in.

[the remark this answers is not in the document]
HarelM: Cool, thanks for the info!

HarelM (approved):

kevinschaul:
Thanks for the help with the review!

Issue #2095, opened 2023-01-24
map.fitBounds throws outside of bounds error when terrain, center and zoom set and bounds to fit on other side of AM


llambanna:
**maplibre-gl-js version**: 3.0.0-pre.3

**browser**:  chrome Version 109.0.5414.87 (Official Build) (x86_64)
 
### Steps to Trigger Behavior

 1. include `center` and `zoom` when instantiate map
 2. include terrain 
 3. use `fitBounds` to move to area the other side of AM from current location

### Link to Demonstration

click on the Fit to NZ button and error throws

https://jsbin.com/qevebesote/edit?html,output

### Expected Behavior

it zooms to NZ

### Actual Behavior

it throws an error:

> maplibre-gl-dev.js:29359 Uncaught Error: x=-7, y=320, z=9 outside of bounds. 0<=x<512, 0<=y<512 0<=z<=25 
    at new CanonicalTileID (maplibre-gl-dev.js:29359:19)
    at new OverscaledTileID (maplibre-gl-dev.js:29407:26)
    at Transform.getElevation (maplibre-gl-dev.js:45892:24)
    at Map._prepareElevation (maplibre-gl-dev.js:49563:48)
    at Map.flyTo (maplibre-gl-dev.js:49781:18)
    at Map._fitInternal (maplibre-gl-dev.js:49338:18)
    at Map.fitBounds (maplibre-gl-dev.js:49295:21)
    at HTMLButtonElement.<anonymous> (runner:49:13)

HarelM:
Thanks for taking the time to report this! Seems like an issue with getting the elevation, maybe missing a wrap of some sort...

PR #2155, opened 2023-02-09
fix fitBounds crossing AM with 3d - add wrap for lng in transform getElevation 

llambanna:
wrap `lnglat` point so `fitBounds` across anti-meridian no longer throws error

## Launch Checklist

 - [x] Link to related issues. https://github.com/maplibre/maplibre-gl-js/issues/2095

### DIFF
--- a/src/geo/transform.test.ts
+++ b/src/geo/transform.test.ts
@@ -459,4 +459,13 @@ describe('transform', () => {
         expect(transform.getBounds().getNorthWest().toArray()).toStrictEqual(transform.pointLocation(new Point(0, top)).toArray());
     });
 
+    test('getElevation with lng less than -180 wraps correctly', () => {
+        const WRAPPED_OVERSCALETILEID_DOES_NOT_THROW = 4;
+        const terrain = {
+            getElevation: () => WRAPPED_OVERSCALETILEID_DOES_NOT_THROW
+        } as any as Terrain;
+        const transform = new Transform(0, 22, 0, 85, true);
+        expect(transform.getElevation(new LngLat(-183, 40), terrain)).toBe(WRAPPED_OVERSCALETILEID_DOES_NOT_THROW);
+    });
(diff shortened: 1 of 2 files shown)


HarelM:
Thanks for taking the time to fix this!
Can you add a small unit tests around this bug?

llambanna:
@HarelM i tried but i have a mac M1 and i can't get the tests to run, i get this error:
>  dlopen(/Users/admin/code/maplibre-gl-js/node_modules/canvas/build/Release/canvas.node, 0x0001): symbol not found in flat namespace '_cairo_fill'

despite following the arm64 approach here: https://github.com/koordinates/maplibre-gl-js/blob/main/CONTRIBUTING.md#osx

HarelM:
You can use the CI to run the tests in theory, which is less than optimal I guess.
But since this is a bug fix, it has to have a test to make sure this bug won't "return".

llambanna:
@HarelM yep, that would be a painful way to write a test, but it would work, eventually
i'll see what i can do

HarelM:
Thanks! Also if there's a way to make it work on M1 as well it would be super, but you'll need to investigate and test it as I don't have one unfortunately (and I'm also not very keen on buying one due to all this incompatibility mess).

HarelM (approved):

HarelM:
Forgot to ask, can you please add an entry to the changelog file?

llambanna:
@HarelM yep, i'll add something to changelog

if i get it working on an M1 i'll update the docs with what i did

HarelM:
Thanks!! It might not be just the docs, but you already know that 😀

llambanna:
> Thanks!! It might not be just the docs, but you already know that 😀

@HarelM turns out it was just the docs 😌 

have a PR for steps: https://github.com/maplibre/maplibre-gl-js/pull/2196

PR #2156, opened 2023-02-09
Break circular dep: index.ts ... web_worker.ts -> src/index.ts

zhangyiatmicrosoft:
## Launch Checklist

Long circular depedency:
<code>src/index.ts -> src/ui/map.ts -> src/style/style.ts -> src/util/global_worker_pool.ts -> src/util/worker_pool.ts -> src/util/web_worker.ts -> src/index.ts</code>

It looks very reasonable to move workerURL to config object instead of piggyback on main maplibre.

 - [x] Link to related issues.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -142,7 +142,13 @@ const exported = {
         config.MAX_PARALLEL_IMAGE_REQUESTS = numRequests;
     },
 
-    workerUrl: '',
+    get workerUrl(): string {
+        return config.WORKER_URL;
+    },
+
+    set workerUrl(value: string) {
+        config.WORKER_URL = value;
+    },
 
     /**
      * Sets a custom load tile function that will be called when using a source that starts with a custom url schema.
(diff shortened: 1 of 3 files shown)


HarelM:
There's a need to update the documentation I believe and also a note in the changelog that this is a breaking change.
I would recommend opening an issue in the docs repo after this is merged so that we don't forget to update the documentation there (about the workers, csp etc).
Overall, this look good. THANKS!

HarelM:
A side note, this will also cause conflicts with #2130

zhangyiatmicrosoft:
Regarding doc changes, actually no need, I think.
Everyone can still use maplibre.workerUrl = "foo", it is just internally wraps and sends to config object. No one (outside or inside) needs to and should care.
I have been trying to keep it exactly as before and reduce entries in Changelog to only absolutely necessary.

HarelM:
My mistake, I complete read the code wrong. Thanks for pointing this out! :-)

HarelM (approved):

rotu:
I like this change :-) Thank you!

PR #2013, opened 2022-12-30
Use resizeObserver instead of window resize events

rotu:
Previously, if a map was created before the page was rendered, it would have the wrong size. Or if the map was resized by anything other than resizing the window, it would also have the wrong size.

Now, the map will have the correct size no matter why the container element is resized.

PR #2157, opened 2023-02-09
Use resizeObserver instead of window resize events, fix and clean tests

HarelM:
This adds the relevant unit test.
Cleans the other tests that has duplicate code in every before function.
Clear some test that did not need some methods.

@rotu Feel free to review :-)

Replaces #2013

### DIFF
--- a/src/source/vector_tile_worker_source.test.ts
+++ b/src/source/vector_tile_worker_source.test.ts
@@ -13,9 +13,6 @@ import {setPerformance} from '../util/test/util';
 describe('vector tile worker source', () => {
     const actor = {send: () => {}} as any as Actor;
     let server: FakeServer;
-    let originalGetEntriesByName;
-    let originalMeasure;
-    let originalMark;
 
     beforeEach(() => {
         global.fetch = null;
--- a/src/source/vector_tile_worker_source.test.ts
+++ b/src/source/vector_tile_worker_source.test.ts
@@ -21,18 +21,12 @@ describe('vector tile worker source', () => {
         global.fetch = null;
         server = fakeServer.create();
         setPerformance();
-        originalGetEntriesByName = window.performance.getEntriesByName;
-        originalMeasure = window.performance.measure;
-        originalMark = window.performance.mark;
 
     });
 
     afterEach(() => {
         server.restore();
         jest.clearAllMocks();
-        window.performance.getEntriesByName = originalGetEntriesByName;
-        window.performance.measure = originalMeasure;
-        window.performance.mark = originalMark;
     });
     test('VectorTileWorkerSource#abortTile aborts pending request', () => {
         const source = new VectorTileWorkerSource(actor, new StyleLayerIndex(), []);
--- a/src/ui/handler/mouse_handler_interface.test.ts
+++ b/src/ui/handler/mouse_handler_interface.test.ts
@@ -1,13 +1,6 @@
 import Point from '@mapbox/point-geometry';
 
-import {setMatchMedia, setPerformance, setWebGlContext} from '../../util/test/util';
 import {generateMousePanHandler, generateMousePitchHandler, generateMouseRotationHandler} from './mouse';
-
-beforeEach(() => {
-    setPerformance();
-    setWebGlContext();
-    setMatchMedia();
-});
 
 describe('mouse handler tests', () => {
     test('MouseRotateHandler', () => {
--- a/src/ui/handler/one_finger_touch_drag_handler_interface.test.ts
+++ b/src/ui/handler/one_finger_touch_drag_handler_interface.test.ts
@@ -1,13 +1,6 @@
 import Point from '@mapbox/point-geometry';
 
-import {setMatchMedia, setPerformance, setWebGlContext} from '../../util/test/util';
 import {generateOneFingerTouchPitchHandler, generateOneFingerTouchRotationHandler} from './one_finger_touch_drag';
-
-beforeEach(() => {
-    setPerformance();
-    setWebGlContext();
-    setMatchMedia();
-});
 
 const testTouch = {identifier: 0} as Touch;
 
--- a/src/ui/map.test.ts
+++ b/src/ui/map.test.ts
@@ -1,5 +1,5 @@
 import Map, {MapOptions} from './map';
-import {createMap, setMatchMedia, setPerformance, setWebGlContext, setErrorWebGlContext} from '../util/test/util';
+import {createMap, setErrorWebGlContext, beforeMapTest} from '../util/test/util';
 import LngLat from '../geo/lng_lat';
 import Tile from '../source/tile';
 import {OverscaledTileID} from '../source/tile_id';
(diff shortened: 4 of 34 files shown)


test/integration/render/mock_browser_for_node.ts
```diff
@@ -38,6 +38,12 @@ global.Blob = window.Blob;
 global.URL = window.URL;
 global.fetch = window.fetch;
 global.document = window.document;
+global.ResizeObserver = function (callback) {
```
rotu, line 41: This doesn't emulate a resize observer. In particular, in the real thing, there's no guarantee that callback is ever called (and especially not synchronously!).

If the idea is (as I think it is) not to mock anything but to prevent the program from crashing in unrelated tests, then it's worth a comment to that effect.

HarelM:
This is strange... Why would only a small part of the render test fail? Why are they affected from the resize mechanism change?

test/integration/render/mock_browser_for_node.ts
HarelM: Well... Not calling the callback causes some tests to fail. It might need a timeout in theory and maybe a different implantation, but generally speaking, it needs something.

src/ui/map.ts
```diff
@@ -476,8 +476,12 @@ class Map extends Camera {
 
         if (typeof window !== 'undefined') {
```
rotu, line 478: This might not be the PR to address it, but this line bugs me. If `window` is `undefined` then we should be failing very loudly, not silently ignoring it!

rotu: It would also allow _resizeObserver to be definitely assigned, so later `this._resizeObserver?.disconnect()` could get rid of the `?`

test/integration/render/mock_browser_for_node.ts
HarelM: Looks like you were right...

src/ui/map.ts
HarelM: I don't have hard feelings for this `if` in general. From my point of view, it can be removed completely.
In what scenario `window` is not defined?

src/ui/handler/mouse_handler_interface.test.ts
```diff
@@ -1,14 +1,7 @@
 import Point from '@mapbox/point-geometry';
 
-import {setMatchMedia, setPerformance, setWebGlContext} from '../../util/test/util';
 import {generateMousePanHandler, generateMousePitchHandler, generateMouseRotationHandler} from './mouse';
 
-beforeEach(() => {
```
wipfli, line 6: Did we add them here with no reason?

src/ui/handler/one_finger_touch_drag_handler_interface.test.ts
```diff
@@ -1,14 +1,7 @@
 import Point from '@mapbox/point-geometry';
 
-import {setMatchMedia, setPerformance, setWebGlContext} from '../../util/test/util';
 import {generateOneFingerTouchPitchHandler, generateOneFingerTouchRotationHandler} from './one_finger_touch_drag';
 
-beforeEach(() => {
```
wipfli, line 6: Did we add them here unnecessarily?

src/ui/map.test.ts
```diff
@@ -722,41 +720,48 @@ describe('Map', () => {
...
+                observe: spy
+            }));
 
             createMap();
+
+            expect(spy).toHaveBeenCalled();
         });
 
         test('do not resize if trackResize is false', () => {
+            let observerCb = null;
```
wipfli, line 735: Can you add a type to observeCb? Rename to observeCallback

wipfli: Or observerCallback...

HarelM: I have taught you well, young padawan.

src/ui/handler/one_finger_touch_drag_handler_interface.test.ts
HarelM: Yes.

src/ui/handler/mouse_handler_interface.test.ts
HarelM: I believe they were there, but are not needed.

src/ui/map.test.ts
HarelM: Done.

src/source/vector_tile_worker_source.test.ts
```diff
@@ -13,26 +13,17 @@ import {setPerformance} from '../util/test/util';
...
     let server: FakeServer;
-    let originalGetEntriesByName;
-    let originalMeasure;
-    let originalMark;
 
     beforeEach(() => {
         global.fetch = null;
         server = fakeServer.create();
         setPerformance();
-        originalGetEntriesByName = window.performance.getEntriesByName;
```
wipfli, line 24: Do you remeber why we had the mark stuff?

test/integration/render/mock_browser_for_node.ts
```diff
@@ -38,6 +38,12 @@ global.Blob = window.Blob;
 global.URL = window.URL;
 global.fetch = window.fetch;
 global.document = window.document;
+global.ResizeObserver = function () {
+    return {
+        observe: () => {},
+        disconnect: () => {}
+    };
+} as any;
```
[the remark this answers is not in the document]
HarelM: Sure.

src/source/vector_tile_worker_source.test.ts
HarelM: I think I added a comment in a code review that this is needed in order to revert back to original state. but other tests don't do it, so there's no real value in keeping it. either we do it in all test or not. and since the tests are passing and not interfering to one another, there's no reason to keep it.

src/ui/map.ts
```diff
@@ -476,8 +476,12 @@ class Map extends Camera {
...
             addEventListener('online', this._onWindowOnline, false);
-            addEventListener('resize', this._onWindowResize, false);
-            addEventListener('orientationchange', this._onWindowResize, false);
+            this._resizeObserver = new ResizeObserver((entries) => {
```
wipfli, line 479: What are the entries?

wipfli (changes requested):
Looks good to me in general, thanks! I think since the behavior for resizing containers has changes, we should treat this as a breaking change with associated changelog remark. Please also add a test where you only resize the map container but not the window.

src/ui/map.ts
HarelM: This is the ResizerObserver API entries, I did not check what's in them.

HarelM:
The unit tests are covering the resize functionality without knowing if the window or the container were resized.
What did you have in mind in terms of testing?

wipfli:
Ah OK I thought the test simulate resizing of the window and then  check that our code gets called. But if the tests already resizing from container, we do not need more tests.

wipfli (approved):
Looking forward to use this new behavior. Thanks!

HarelM:
I added 2 browser tests, just in case...

PR #1809, opened 2022-11-04
Extract sprite image on demand to reduce memory usage and improve performance

pramilk:
Here is the current workflow in terms of how sprite image is processed
1. Sprite (Json and PNG) is downloaded, 
2. Sprite image is drawn on 2d canvas
3. Based on sprite.Json all individual images are extracted and loaded in memory for later consumption (E.g. there are 418 images in https://api.maptiler.com/maps/streets/sprite.png)
4. When images are requested via imageManager.getImage or imageManager.getImages, image(s) are returned.

Issues:
1. All images are extracted upfront, leading to more processing and higher memory consumption. 
2. Very few images may be required for any given map view, but still we are extracting all images.

Fix:
1. Sprite (Json and PNG) is downloaded, (Same as production)
2. Sprite image is drawn on 2d canvas, (Same as production)
3. Based on sprite.Json, only individual images metadata is stored along with the canvas context (no imageData at this time).
4. When the image is requested, extract the imageData from canvas context and return image.

We are seeing 10-12% reduction in memory footprint with some small performance gains in our A/B testing on Bing Maps.

## Launch Checklist

    
 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.
 - [ ] Link to related issues.
 - [ ] Include before/after visuals or gifs if this PR includes visual changes.
 - [x] Write tests for all new functionality.
 - [ ] Document any changes to public APIs.
 - [ ] Post benchmark scores.
 - [x] Manually test the debug page.
 - [x] Add an entry to `CHANGELOG.md` under the `## main` section.

PR #2165, opened 2023-02-11
[BUGFIX] Fix ImageManager null image on first image populated in styleimagemissing

### LINKED ISSUE #2159: Sprite raster data missing from imageManager
### Summary

OSM Americana relies on access to sprite raster data in map.style.imageManager.images in order to render highway shields in a "styleimagemissing" hook. At runtime, javascript code retrieves the raster data associated with a shield graphic, draws a route number on the graphic, and stores the new graphic back in the sprite sheet in memory. However, between 2.4.0 and the current 3.0.0 mainline code, sprite raster data is no longer accessible in imageManager.

If not corrected, OSM Americana would be unable to render highway shields after 2.4.0.

### Details

**maplibre-gl-js version**:

3.0.0-pre.3 (actually, current main branch as of the date of this bug report)

**browser**:

Google Chrome, Ubuntu

### Steps to Trigger Behavior

1. Any time after the map is loaded, run the code below.

```javascript
  for (const [key, value] of Object.entries(map.style.imageManager.images)) {
    console.log(`${key}, ${value.data}`);
  }
```
2. The expected output is a series of lines in the console that show:

```
<name of sprite>, [object Object]
```
3. However, in the current mainline, the output is:
```
<name of sprite>, null
```

### Expected Behavior

Clients are able to access sprite raster data in `map.style.imageManager.images`.

### Actual Behavior

Raster data is null in `map.style.imageManager.images`.

### LINKED ISSUE #2164: [BUG] ImageManager fails to populate callback when styleimagemissing populates an image
While troubleshooting the fix for #2159, I discovered a bug in ImageManager._notify(), which is responsible for retrieving images from the image manager.

ImageManager uses the following algorithm:
1. Attempt to retrieve an image and save the result
2. If the image is not present, fire a '[styleimagemissing](https://maplibre.org/maplibre-gl-js-docs/example/add-image-missing-generated/)' event and log a warning.
3. Send the saved image to the caller as a callback

However, it is possible that the handler for `styleimagemissing` has invoked map.addImage() to add the missing image.

In this case, the algorithm does two incorrect things:

1. A warning is incorrectly written to the console indicating that an image could not be loaded, even though it was loaded by the `styleimagemissing` handler
2. A `null` is returned to the caller instead of the newly-loaded image.

This logic is below:

https://github.com/maplibre/maplibre-gl-js/blob/98f17bfccca46ac7b30cf896973c8ad7ee92b566/src/render/image_manager.ts#L199-L224

### Proposed fix

Proposed is that the algorithm change to:
1. Attempt to retrieve an image and save the result
2. If the image is not present, fire a '[styleimagemissing](https://maplibre.org/maplibre-gl-js-docs/example/add-image-missing-generated/)' event
3. If `styleimagemissing` has generated the requested image, save the result, otherwise log a warning
4. Send the saved image to the caller as a callback

### Impact if not fixed

1. For implementations that invoke addImage() in the `styleimagemissing` handler, the first instance of a newly-requested image will return a null instead of the requested image, while the second and subsequent instances will return normally. The impact on the map will depend on how the client handles null images. In testing I've observed garbled buffer data drawn to the map where I expect an image to appear.
2. Incorrect warnings are written to the console.

ZeLonewolf:
Fixes #2164 
Fixes #2159

This PR fixes an issue where images populated during a `styleimagemissing` event were not returned upstream. This means that only the second and subsequent calls to retrieve an image of the same name were working as expected. I discovered this while encountering image corruption while attempting to implement the suggested fix in #2159.  Before and after screen shots are shown below:

Before
image

After
image

## Launch Checklist

 - [x] Link to related issues.


HarelM:
Can you please add a test to simulate this flow?
Also I've seen the style missing event change in location in the PR to optimize performance, what was the solution there to the style image missing?

ZeLonewolf:
> Can you please add a test to simulate this flow?

Added this in 7d960c9. It turns out there was already a test case, however, it did not have a check to verify that the the image actually got returned in the callback.  I've added that check.

HarelM:
I'm not sure I'm fully align with this solution as it can be that styleimagemissing will addImage asynchronously, what would be the expected behavior in this case?

ZeLonewolf:
> Also I've seen the style missing event change in location in the PR to optimize performance, what was the solution there to the style image missing?

I'm not quite sure that I understand the question. I changed the location of the `stylemissingimage` in the `_notify()` function because the logic was wrong, not because of a performance concern. The function needed to fire the event, but then _retrieve the newly-created graphic_ to return in the callback.  The old logic returned null. This location change allows the for `stylemissingimage` to generate a new graphic _before_ the `if (image) {` block.

ZeLonewolf:
> I'm not sure I'm fully align with this solution as it can be that styleimagemissing will addImage asynchronously, what would be the expected behavior in this case?

I'm not familiar with this capability. Based on a reading of the `getImages()` function, this functionality is not changed. The current behavior in that situation is to return a null value in the callback, and the behavior in this PR would be the same.

https://github.com/maplibre/maplibre-gl-js/blob/98f17bfccca46ac7b30cf896973c8ad7ee92b566/src/render/image_manager.ts#L179-L197

HarelM:
What I meant to say is that the following PR changed the behavior:
https://github.com/maplibre/maplibre-gl-js/pull/1809
It can be seen that the event is sent before trying to use the image, apparently for a reason, but this is not covered in test, and thus passed my review (as it seems there's no good reason to raise the event at that particular point).
Please make sure the test you changed is failing before your fix, and passes afterwards.
Generally speaking, assuming styleimagemissing is synchronous is weak, but it was assumed before, so I don't mind keeping this assumption.

ZeLonewolf:
I've improved the test case and created a failing version in #2167 as a prover. I should have started with a failing test case in the commit history in this PR but here we are.

The CI is reporting network issues and I request that the workflow be re-run.

ZeLonewolf:
> What I meant to say is that the following PR changed the behavior: #1809 It can be seen that the event is sent before trying to use the image, apparently for a reason, but this is not covered in test, and thus passed my review (as it seems there's no good reason to raise the event at that particular point).

Thanks for the pointer to #1809.  It introduced the bug that this PR intends to fix and it certainly explains why I was seeing corruption in the current main but not in 2.4.0!

The logic of getImage() ensures that no extra deep copies are made if an image isn't present:

https://github.com/maplibre/maplibre-gl-js/blob/9e8fa84a12d91be75b0a4a5adbe44b2309d46492/src/render/image_manager.ts#L89

Therefore, the second call to getImage() should not have a performance concern, since the first call would have short-circuited.

And, now that we've got a proper test case implemented, this will ensure that we won't regress.

ZeLonewolf:
> Generally speaking, assuming styleimagemissing is synchronous is weak, but it was assumed before, so I don't mind keeping this assumption.

I agree with this assessment. I will say that, as a style author, I don't know how this could be done asynchronously, and I would be interested in your thoughts as to how we could do that.

In our [performance benchmarks](https://zelonewolf.github.io/openstreetmap-americana/shieldtest.html), I routinely see shield rendering times on the order of 200μs to 1ms, though this will vary based on client browser/hardware performance of course.

With that I believe I'm caught up on all review notes.

pramilk:
Sorry @ZeLonewolf for causing this issue. I did not considered this scenario. Thanks for taking time to fix it. Your changes looks good to me.

HarelM:
Not sure I underrated why the test is falling... Can you take a look into it?

ZeLonewolf:
> Not sure I underrated why the test is falling... Can you take a look into it?

The windows CI runner is sometimes reporting timeout errors that I don't understand, and it's happening broadly across the project for some reason.  The current run here is passing.

HarelM (approved):

PR #2167, opened 2023-02-11
Failing unit test prover for #2164

ZeLonewolf:
This PR is proof of a failing unit test for #2164 and can be safely discarded once merged.


HarelM:
There's no need to prove a test is falling :-) simply to say you test it and made sure it fails before the fix. 😀
Can this be closed?

ZeLonewolf:
Yay, it fails!

PR #2168, opened 2023-02-11
Add Map.getImage()

### LINKED ISSUE #2162: Supported method for retrieving sprite images
## Motivation

Currently, there is no documented method for retrieving a sprite image from a mapLibre map. This has caused style authors such as myself to hunt around in the codebase looking for a way to access sprite data, such as I did in #2159 where I accessed internal (but nonetheless exposed) variables. Retrieving sprite images is needed for runtime icon manipulation, such as we do in OSM Americana. There is presently a method `map.style.getImage(id)` in the code, however this is not publicly documented.

## Design Alternatives

This functionality is currently accessible via `map.style.getImage(id)`, however, this option is undocumented.

## Design

The recommended implementation is to implement `Map.getImage()` to delegate back to map.style.getImage(id) as is used in other parts of map.ts.

### Implementation

Additional work would be needed to implement this across the maplibre-gl variants.  However, the naming convention of `getImage()` follows the same naming conventions elsewhere in this file.

ZeLonewolf:
Fixes #2162 

This PR adds `Map.getImage()`, which completes the CRUD methods for image access, along with `Map.addImage()`, `Map.updateImage()`, and `Map.removeImage()`.

## Launch Checklist

 - [x] Link to related issues.

### DIFF
--- a/src/ui/map.test.ts
+++ b/src/ui/map.test.ts
@@ -2205,108 +2205,110 @@ describe('Map', () => {
         });
     });
 
-    test('map getImage matches addImage', done => {
+    test('map getImage matches addImage, uintArray', done => {
         const map = createMap();
+        const id = 'add-get-uint';
+        const inputImage = {width: 2, height: 1, data: new Uint8Array(8)};
 
-        //Create graphics of various types, each with a different size
-        const idImageUint = 'add-get-uint';
-        const inputImageUint = {width: 2, height: 1, data: new Uint8Array(8)};
+        map.addImage(id, inputImage);
+        expect(map.hasImage(id)).toBeTruthy();
 
-        const idImageUintClamped = 'add-get-uint-clamped';
-        const inputImageUintClamped = {width: 1, height: 2, data: new Uint8ClampedArray(8)};
+        const gotImage = map.getImage(id);
+        expect(gotImage.data.width).toEqual(inputImage.width);
+        expect(gotImage.data.height).toEqual(inputImage.height);
+        expect(gotImage.sdf).toBe(false);
 
-        const idImageData = 'add-get-image-data';
-        const inputImageData = new ImageData(1, 3);
+        done();
+    });
 
-        const idStyleImage = 'add-get-style-image';
-        const inputStyleImage: StyleImageInterface = {
+    test('map getImage matches addImage, uintClampedArray', done => {
+        const map = createMap();
+        const id = 'add-get-uint-clamped';
+        const inputImage = {width: 1, height: 2, data: new Uint8ClampedArray(8)};
+
+        map.addImage(id, inputImage);
+        expect(map.hasImage(id)).toBeTruthy();
+
+        const gotImage = map.getImage(id);
+        expect(gotImage.data.width).toEqual(inputImage.width);
+        expect(gotImage.data.height).toEqual(inputImage.height);
+        expect(gotImage.sdf).toBe(false);
+
+        done();
+    });
+
+    test('map getImage matches addImage, ImageData', done => {
+        const map = createMap();
+        const id = 'add-get-image-data';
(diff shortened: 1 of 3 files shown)


ZeLonewolf:
Going to add some more test cases

src/ui/map.test.ts
```diff
@@ -2204,6 +2206,112 @@ describe('Map', () => {
...
+        expect(gotStyleImageClamped.data.width).toEqual(inputStyleImageClamped.width);
+        expect(gotStyleImageClamped.data.height).toEqual(inputStyleImageClamped.height);
+
+        expect(gotStyleImageSDF.data.width).toEqual(inputStyleImageSDF.width);
+        expect(gotStyleImageSDF.data.height).toEqual(inputStyleImageSDF.height);
+
+        //Verify retrieved graphic has same image contents
+        for (let i = 0; i < gotImageUint.data.data.length; i++) {
+            expect(gotImageUint.data.data[i]).toEqual(inputImageUint.data[i]);
+        }
```
1ec5, line 2285: Unlike `toBe`, [`toEqual`](https://jestjs.io/docs/expect#toequalvalue) tests deep equality, so a `for` loop shouldn’t be necessary.

ZeLonewolf: The reason I did this is that Uint8ClampedArray is internally repackaged into Uint8Array in some cases and so toEqual would fail because they're different types.  The `for` loop makes this check agnostic to the underlying data storage array.

ZeLonewolf:
This is good to go.  I tried and failed to make the unit tests more compact, but at least it's testing all of the non-HTML image types.

src/ui/map.test.ts
```diff
@@ -2204,6 +2205,112 @@ describe('Map', () => {
         });
     });
 
+    test('map getImage matches addImage', done => {
```
HarelM, line 2208: Can you please split this to different tests, each type of image in a different test?

HarelM: I would also recommend adding these tests to the `stye.test.ts` file, and only add a single test here, as the logic in map.ts is only wrapping the functionality inside and I prefer to test it where the logic is, hopefully.
i.e. there should be a proportion between the number of tests and the logic inside the method, unless you say that these are integration tests and not unit tests so my argument is not accurate.

ZeLonewolf: First pass, tests separated in b6ee6e6e3

ZeLonewolf: I looked through `style.test.ts`, which seems to be unit testing the `Style` class, if I understand it correctly. This PR is specifically adding `getImage()` as the corollary to `addImage()` and so the unit tests I've added here as specifically verifying that the various image types that I'm adding are also retrieved correctly. I don't understand how that fits into the `Style` class.

I explored adding additional checks into the sprite image extraction test, however, it appears that the Style checks are all using a stub map, so it (a) doesn't work and (b) wouldn't be testing the new code here anyways.

HarelM: I see, so those are more of integration tests. No problem.

src/ui/map.test.ts
```diff
@@ -2204,6 +2205,114 @@ describe('Map', () => {
         });
     });
 
+    test('map getImage matches addImage, uintArray', done => {
```
HarelM, line 2208: This doesn't look like an async test, why is the "done" needed here?

HarelM (approved):

src/ui/map.test.ts
ZeLonewolf: Looks like it isn't needed.  I'm still learning how the unit tests work so just a bit of growing pains there as they say.

I removed the done hooks in c6e7e7bab to be a no-argument callback, hopefully that's the right way to do a synchronous test.

Issue #2149, opened 2023-02-06
Split style-spec to a different repo


HarelM:
## Motivation
The style spec is a whole subject that deserves its own repo.
This new repo should facilitate for documentation and the style validation code.
It should also include the typescript definitions in that package (in npm).
This way, this repo can consume this package no matter how it is implemented internally.
As a starting point I would advise to simply move parts of this repo as is (without any code changes) and make sure everything works as expected, once this is done we can discuss the next step for this new repo, either creating a similar package for the native repo to use, refresh technology, change to json-scheme or whatever.
But the first step is to split this apart and define a proper interface.

## Design Alternatives

Keep everything together in a single repo.

## Design

Describe above - split to a different package, define an interface/API and use it

## Acceptance criteria for the bounty:
1. Style spec definition, validation, documentation are in a new repo, including CI of tests and publishing of package and docs.
2. The code is removed from this repo, and this repo only depends on packages produced by the other repo

Link to bounty:
https://github.com/maplibre/maplibre/issues/191

HarelM:
Cc: @birkskyum @wipfli @louwers @ovivoda

birkskyum:
These four issues contain quite a lot of gotchas and pitfalls. I'd be able to plan and resolve them all, but it would likely take me a full week, which I unfortunately can't carve out for a bounty this size ~25$/h. Let's see if someone faster wants to give it a go.

* (blocking for issues below) https://github.com/maplibre/maplibre-gl-js-docs/issues/384 
* (blocking for issues below) https://github.com/maplibre/maplibre-gl-js-docs/issues/389 
* https://github.com/maplibre/maplibre-gl-js/issues/2150
* https://github.com/maplibre/maplibre-gl-js/issues/2149

HarelM:
Thanks for the summary of this. The current state is worse than I thought. I'll see what can be done.

krishnaglodha:
Hi @HarelM , please let me know if the task is available, I'd like to contribute to it

HarelM:
Please coordinate with @birkskyum as he has started looking into it and started working on it.
This issue is dependant on different issues and it's probably the last one to be finished.

birkskyum:
@krishnaglodha , ping me in slack and i can give you a status on where we are

birkskyum:
Style spec is split out from maplibre-gl-js repo already, so i consider this issue complete. It's unfair to turn this into an umbrella ticket after the fact just to withhold the bounty.

[...]

PR #2198, opened 2023-02-20
Use @maplibre/maplibre-gl-style-spec npm package

### LINKED ISSUE #2194: Use @maplibre/maplibre-gl-style-spec in MapLibre GL JS
Make maplibre-gl-js use the style-spec through the published style-spec npm package instead of references files in the nested subdirectory.

- Point all references to style-spec towards the [npm package ](https://www.npmjs.com/package/@maplibre/maplibre-gl-style-spec)
- It might be necessary to expose additional private apis from if there are classes that can be reached form the repo but not the style-spec package

This will make it easier to migrate the style spec later on.

birkskyum:
Closes #2194

This PR removes style-spec from the repo in favor of references to the [npm package](https://www.npmjs.com/package/@maplibre/maplibre-gl-style-spec)


build/generate-style-code.ts
```diff
@@ -2,7 +2,7 @@
 
 import * as fs from 'fs';
 
-import spec from '../src/style-spec/reference/v8.json' assert {type: 'json'};
+import spec from '@maplibre/maplibre-gl-style-spec/tsc/src/reference/v8.json' assert {type: 'json'};
```
HarelM, line 5: I would feel a lot better if we do not know about the internal folder structure of this library...
I.e. export everything that is needed in the top level API.

build/rollup_plugin_minify_style_spec.ts
```diff
@@ -8,7 +8,7 @@ export default function minifyStyleSpec(): Plugin {
     return {
         name: 'minify-style-spec',
         transform: (source, id) => {
-            if (!/style\-spec[\\/]reference[\\/]v[0-9]+\.json$/.test(id)) {
+            if (!/reference[\\/]v[0-9]+\.json$/.test(id)) {
```
HarelM, line 11: Is this relevant?

src/data/bucket/fill_bucket.test.ts
```diff
@@ -7,7 +7,7 @@ import Point from '@mapbox/point-geometry';
 import segment from '../segment';
 import FillBucket from './fill_bucket';
 import FillStyleLayer from '../../style/style_layer/fill_style_layer';
-import {LayerSpecification} from '../../style-spec/types.g';
+import {LayerSpecification} from '@maplibre/maplibre-gl-style-spec/dist';
```
HarelM, line 10: We shouldn't be using the `dist` folder I believe, this similar to the other comment I made.

src/style/properties.ts
```diff
@@ -100,6 +100,7 @@ export class PropertyValue<T, R> {
     constructor(property: Property<T, R>, value: PropertyValueSpecification<T> | void) {
         this.property = property;
         this.value = value;
+        // @ts-ignore
```
HarelM, line 103: Why was this added?

build/rollup_plugin_minify_style_spec.ts
birkskyum: Yes, it takes about 100 kb off the production build.

src/util/util.ts
```diff
@@ -254,8 +254,30 @@ export function filterObject(input: any, iterator: Function, context?: any): any
...
 }
 
-import deepEqual from '../style-spec/util/deep_equal';
-export {deepEqual};
+/**
+ * Deeply compares two object literals.
+ * @param a
+ * @param b
+ */
+export function deepEqual(a?: unknown | null, b?: unknown | null): boolean {
```
HarelM, line 262: Can't we use a library for this? I would feel safer that the library is fully tested instead of reimplementing this...

src/style/properties.ts
birkskyum: It returned this

```ts
(alias) normalizePropertyExpression<T>(value: PropertyValueSpecification<T>, specification: StylePropertySpecification): StylePropertyExpression
import normalizePropertyExpression

Excessive stack depth comparing types '["array", ExpressionSpecification | ExpressionInputType, unknown]' and '["array", ExpressionSpecification | ExpressionInputType, unknown]'.ts(2321)
```

test/build/min.test.ts
```diff
@@ -45,7 +45,7 @@ describe('test min build', () => {
         const decreaseQuota = 4096;
 
         // feel free to update this value after you've checked that it has changed on purpose :-)
-        const expectedBytes = 754120;
+        const expectedBytes = 756547;
```
HarelM, line 48: This seems like a big change in size, can you explain why? I would not expect a size change from this "simple" movement of code.

HarelM:
@birkskyum THANKS!!
See my comments.
Overall, very nice work! I don't see major code changes which is a good sign.

test/build/min.test.ts
birkskyum: I think it's caused by the source maps on some of the style-spec files.

src/util/util.ts
birkskyum: We even have 2 deepEqual functions. Another one is used for testing. I tried merging them before and it was quite a headache, because they have to be the right amount of deep/shallow for our code to pass

test/build/min.test.ts
HarelM: Can you look into it and better understand what cause this?

build/generate-style-code.ts
birkskyum: I think it's a major restructuring to pull everything into the index bundle, because of name collisions etc., which would make this splitting PR quite confusing. Like you say "I don't see major code changes which is a good sign."

test/build/min.test.ts
birkskyum: Yes, it could also simply be the tsc compiled files that aren't minified.

build/generate-style-code.ts
HarelM: Knowing the library internal folder structure is almost as bad as having it as a submodule, IMHO.
Expanding the styke-spec.ts to export more things shouldn't be a huge effort to do, and make the library more explicit about what is being exported.
We should also rename the `stype-spec.ts` to `index.ts` in the style-spec repo to indicate the entry point of the library. 
The changes in this repo should still be minimal.
If we have collision, we can use `import {something as somethingelse}`, but I don't think it will be a problem, knowing the code etc...

HarelM: Please note that some are already being exported, `v8, Color` etc...

src/style/properties.ts
HarelM: This isn't great... Please see if there's a different way to solve this besides ignoring, maybe configuring the depth comparison to be a bit more deeper...? IDK...

build/rollup_plugin_minify_style_spec.ts
birkskyum: apparently it is

src/util/util.ts
birkskyum: I'll open an issue when the dust settles from this split.

src/style/properties.ts
birkskyum: It's resolve now, because we reworked style-spec

test/build/min.test.ts
birkskyum: After reworking style-spec to export more of the API, the bundle size now increased even further to +100kb...

HarelM: This is strange/alarming...

build/rollup_plugin_minify_style_spec.ts
birkskyum: I see what you mean now. This is the main reason the bundle size suddenly grew. When the reference is imported as 

`import {latest} from '@maplbire/maplibre-gl-style-spec'` there is no /reference/v8.json path to match against, so the minifcation doesn't happen.

HarelM: I'm happy to know that sometimes I get things right 😀

test/build/min.test.ts
birkskyum: Alright, after using the minifyed style-spec package, the build size is at 779kb, so 25 kb larger. One place I know we add some kb is in the style-code where we import directly from '@maplibre/maplibre-gl-style-spec/tsc/src/expression/types';, and anything we import there, which might also be exported from the bundle, will be included twice in the production build

birkskyum: In the style-spec, we i.e. import and export the following:

```ts
import {FormattedType, NullType, Type, toString} from './expression/types';
import ResolvedImage from './expression/types/resolved_image';
import Formatted, {FormattedSection} from './expression/types/formatted';
```

These are imported explicitly in maplibre-gl-js.

In generate-style-code, we have this

```ts
            `import {
    ${overridables.reduce((imports, prop) => { imports.push(runtimeType(prop)); return imports; }, []).join(',\n    ')}
} from '@maplibre/maplibre-gl-style-spec/tsc/src/expression/types';
```

I worry there is an overlap there

HarelM: I'll be surprised if rollup doesn't handle this, but maybe you can play around with these imports to see what happens, can't you?

HarelM: The minified version is a bit more cryptic, i would consider looking at the dev file of both branches and run a diff tool against them to see if you can spot the cause of this size diff. I did that in the migration to typescript to better understand what changes were made that I didn't expect...

HarelM: You could also try the following:
Bring back the files or use another branch out of main, defined in tsconfig the style-spec folder as a reference package, change all the imports to use that type of import i.e. absolute package type import instead of relative, make sure you add the relevant things that needs to be exported and only then, if this doesn't cause the size test to fall try and use the package from npm.

birkskyum: I found an issue. So, the style-spec package use the import 'url', which is a part of node. When I then try to build maplibre-gl-js prod build, it says the following:

Screenshot 2023-02-23 at 08 11 01

I tried installing an npm package called 'url', which added 23 kb. So the issue is that this node module can't be found when we use style-spec though the npm package.

HarelM: I think this is a mistake, you don't need an extra package called `url` to parse a url in javascript as far as I know.
I would try and remove that import and see what happens.

birkskyum: Exactly, the error I just posted is the result when it's removed.

HarelM: I mean remove the line `import url from 'url'` in style-spec. use `new URL(...)` instead, which I think should work in style-spec.
And then try using the newly created package in maplibre-gl-js.
Is this what you tried and the above is the error message you get? because I would find it stange that we do not use a package and rollup says that it can't be found...

birkskyum: Without the 'url' module, the bundle is now 768 kb, so adding 14 kb

HarelM: This probably means we are missing something still... :-)

HarelM: Consider creating a local npm package and import it from a local folder instead of all the round trip though npm...

HarelM: yes, I think you can run `npm i ../../somefolder/someotherfolder`

birkskyum: omg, i can't believe i only learn how easy this is now. thank you

birkskyum: it actually causes a lot of rollup errors, so I guess I'm stuck with the roundtrips for now.

HarelM: You probably need to npm publish locally the style-spec (create a tgz or something) and then npm install in here.

birkskyum: Alright, I remove the format/migrate/validate from being exported by style-spec, because they are only used in the style-spec binaries which I think are okay to call down into the files. At least this removes this unused code from maplibre-gl-js build.

birkskyum: We're down to 756 kb

HarelM: Is there a place in the code where we import * from style spec? Removing publicly exported stuff that are not in use should be handled by rollup I believe...

birkskyum: I agree, but I don't think there is any wildcard import

birkskyum: I've been comparing the dev build quite a lot, but things are shuffled around, and I don't know where to find these 2kb. At least everything else checks out now.

HarelM: Yeah, but it still strange for me that rollup picks stuff that are not in use.
I would still consider doing another attempt at using paths in typescript config to use an "absolute path" to the style-spec folder:
https://www.typescriptlang.org/tsconfig#paths
It will still require most of the changes you made, but allow you to see if and when these 2kb are added...

HarelM: i.e. keep the style-spec folder in this repo, change the config and update the imports, fix the factory and some other small stuff and see if the size changes.
If not, and it's the same code as in the other repo, you'd know that it has to do with the npm package and how rollup is using it, i think...

birkskyum: I've made a branch where everything uses the tsconfig path. The result was a 766kb production build, which is even larger than the current.

HarelM: Bummer... Ok, I guess we'll need to live with this extra 2kb. It might be worth opening an issue to remind us that this change caused a size increase and might be worth investigating how to improve the bundle size in the future...

HarelM (approved):

HarelM:
I've updated the initial comment as this doesn't fully solve #2149 - when the docs are generated from the new repo the this can be considered as completed.
Having said that, this does resolve #2194.

birkskyum:
Okay, I thought that was covered by https://github.com/maplibre/maplibre-gl-js-docs/issues/400. I'm make a stable 18.0.1 release of the style spec before merging this, so GL JS doesn't depend on a prerelease

HarelM:
@birkskyum I've just installed the latest version using npm (i.e. pre-5).
It seems that due to this change the typescript types have moved from maplibre-gl package to the style-spec package.
What do you think we should do? 
Add instructions in the readme or try and export the style-spec types in the maplibre.d.ts file (i.e. re-export)?

HarelM:
`StyleSpecification` for example, `LayerSpecification`.
See here:
https://www.unpkg.com/browse/maplibre-gl@3.0.0-pre.5/dist/maplibre-gl.d.ts
Line 6.

birkskyum:
Interesting - my plate is a bit full for the time being though, so I don't have an informed opinion.

PR #2213, opened 2023-02-20
power preference option to webgl context

marucjmar:
Adds at **powerPreference** as a map option. The option is used when creating the WebGL context. 

## Launch Checklist


HarelM:
Can you please elaborate a bit on the use case for this new parameter?
Since this is the third parameter I would consider moving to something more generic to solve this problem, wouldn't you?

pramilk:
As per https://caniuse.com/?search=powerPreference, its currently only supported in Firefox on MacOS.

PR #2216, opened 2023-02-21
Fix format expression typing

aSalama-aya:
## Launch Checklist

Fix `format` expression typing to make `text-color` optional as it is used in [examples](https://maplibre.org/maplibre-gl-js-docs/example/display-and-style-rich-text-labels/)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### 🐞 Bug fixes
 
+- Fix `format` expression type to make `text-color` optional.
 - Fix issue unloading sprite sheet when using `setStyle(style, {diff:true})` ([#2146](https://github.com/maplibre/maplibre-gl-js/pull/2146))
 - Fix wrap coords in `getTerrain` when `fitBounds` accross the AM ([#2155](https://github.com/maplibre/maplibre-gl-js/pull/2155))
 - _...Add new stuff here..._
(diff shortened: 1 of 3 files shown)


HarelM:
Thanks for taking the time to fix this!!
We are currently looking into moving the style spec out to a different repo so this conflicts with this effort.
Once the migration is complete we can merge this to the other repo, but I would wait a bit to see that everything is as expected, post migration...

aSalama-aya:
@HarelM Thanks for your comment!, any timeline for when moving the specs to a separate repo could happen?
Currently I'm getting ts errors trying to use format expression without providing `text-color`, so I hope this could be solved soon to avoid tons of ts-ignore :)

HarelM:
It's underway, see #2198
I hope we will be able to finish this within a couple of weeks.

AyaSalama-TomTom:
Questions: Will the external developer need to import style specs from the new repo? or it will still be exported from the root `maplibre-gl` package and this change is only for your internal organizing?

aSalama-aya:
Questions: Will the external developer need to import style specs from the new repo? or it will still be exported from the root maplibre-gl package and this change is only for your internal organizing?

HarelM:
The public maplibre-gl-js "main" API shouldn't be affected by this change, if you used to import internal files/classes then yes, this might be a problem...
But this will be part of version 3, which will contain breaking changes...

birkskyum:
@aSalama-aya , the style-spec has now finalized the migration to https://github.com/maplibre/maplibre-gl-style-spec/, with only the documentation site still being work-in-progress, so feel free to open the PR there.

PR #2241, opened 2023-03-02
shouldPausePlacement short circuits when _forceFullPlacement

schlosna:
## Launch Checklist

When rendering many symbols on a map, there is a lot of time spent iterating through `continuePlacement()` and callbacks to `shouldPausePlacement()`. The current implementation of `shouldPausePlacement()` always invokes `browser.now()` to compute the elapsed time, even when `_forceFullPlacement == true`.

Given this can be a hot path when there are many symbols, we can check `_forceFullPlacement` and potentially short circuit without needing to compute the elapsed time, saving roughly ~30ns to ~50ns per invocation.

 - [x] Link to related issues.
     - None that I could find
     - N/A
     - N/A
 - [x] Post benchmark scores.


HarelM:
Can you add a changelog entry for this change as well?
Can you think of a test to cover this change?
Do you have a way to make sure this test actually improve performance? I see the code change and I understand why this is better, and the code itself isn't bad, so this change is obviously welcome, just want to make sure we are actually making a difference...

src/style/pauseable_placement.ts
```diff
@@ -97,8 +97,10 @@ class PauseablePlacement {
         const startTime = browser.now();
 
         const shouldPausePlacement = () => {
-            const elapsedTime = browser.now() - startTime;
-            return this._forceFullPlacement ? false : elapsedTime > 2;
+            if (this._forceFullPlacement) {
+                return false;
+            }
+            return browser.now() - startTime > 2;
```
schlosna, line 103: I'm not sure what style you all prefer, so curious if you would prefer ternary for conciseness?

```suggestion
            return this._forceFullPlacement ? false : (browser.now() - startTime) > 2;
```

schlosna (commented):
Thanks for reviewing! I have updated the changelog. I was finally able to get the benchmarks running locally and capture output for the ones that seem most relevant to this change around placement and symbol layout and attached below. The improvement does not seem statistically conclusive, but there are slight improvements in (20% trimmed) Mean compared to current `main` bc6d6f4.

I'm not super familiar with the maplibre-gl-js codebase, so if you have pointers on specific tests that should exercise the (pausable) placement bits, I can try to add some coverage there.

image

image

image

image

schlosna:
Hi @HarelM , I just wanted to check to see if there were any further comments, questions, or changes you'd like to see for this PR. Thanks!

HarelM (approved):

HarelM:
I'll wait for the tests to finish and merge. Thanks!

schlosna:
Excellent, thanks!

PR #2243, opened 2023-03-03
Change default original event type from undefined to unknown

neodescis:
## Launch Checklist

Currently, the default type for originalEvent on a MapLibreEvent is "undefined." This can lead to type errors when handling some events unless the type for originalEvent is specified on the handler, even when we may not care about the type of orginalEvent at all. For example, let's say I want to handle a zoom event. If I define an event handler like this:

onZoom(event: MapLibreEvent) {}

I will get a type error if I try to subscribe to a zoom event using that callback, because that event is typed as `MapLibreEvent<MouseEvent | TouchEvent | WheelEvent | undefined>`. The TypeScript compiler complains about the three non-undefined event types being unassignable to undefined.

Moreover, the type "unknown" is more correct here anyway, as the base case is not knowing what the type is. In the case where the handler doesn't care about originalEvent, using unknown as the default instead means you don't have to provide the type at all.

### DIFF
--- a/src/ui/map_events.test.ts
+++ b/src/ui/map_events.test.ts
@@ -2,7 +2,7 @@ import simulate, {window} from '../../test/unit/lib/simulate_interaction';
 import StyleLayer from '../style/style_layer';
 import {createMap, beforeMapTest} from '../util/test/util';
 import {MapGeoJSONFeature} from '../util/vectortile_to_geojson';
-import {MapLayerEventType} from './events';
+import {MapLayerEventType, MapLibreEvent} from './events';
 
 beforeEach(() => {
     beforeMapTest();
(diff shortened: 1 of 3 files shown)


HarelM:
Interesting. Thanks for taking the time to open this PR.
Can you think of a test to cover this change?

neodescis:
Well, I guess that depends on what you mean by "test." This change would not alter how anything actually runs, and it also wouldn't affect the way any tests would run. It does affect compilability though, so I could probably add a test that would build with this that wouldn't build before. It would be a simple event subscription, just like the above. Is that what you're looking for?

HarelM:
Yes, I think so. Or even changing an existing test to benefit from the change you just made.

src/ui/map_events.test.ts
```diff
@@ -155,6 +155,21 @@ describe('map events', () => {
         expect(spyB).toHaveBeenCalledTimes(1);
     });
 
+    test('Map#on calls an event listener with no type arguments, defaulting to \'unknown\' originalEvent type', () => {
+        const map = createMap();
+
+        const handler = {
+            onMove: function onMove(_event: MapLibreEvent) {}
```
neodescis, line 162: This does not build without the change to `unknown` above.  Instead, you would need to specify `MapLibreEvent<unknown>` or `MapLibreEvent<MouseEvent | TouchEvent | WheelEvent | undefined>`.

neodescis:
Test added!

HarelM (approved):

HarelM:
THANKS!!

PR #2244, opened 2023-03-03
Ensure features property is typed as MapGeoJSONFeature[] on MapLayerMouseEvent and MapLayerTouchEvent.

parkerziegler:
Currently, `MapLayerMouseEvent` and `MapLayerTouchEvent` are typed with their `features` property as `GeoJSON.Feature[]`. However, the features returned on either of these events are `MapGeoJSONFeature`s; they contain the additional properties of that interface (e.g., `layer`, `source`, `sourceLayer`, and `state`). In addition, it appears the unit tests already cast features returned on `MapLayerMouseEvent`s to `MapGeoJSONFeature[]`; see [here](https://github.com/maplibre/maplibre-gl-js/blob/main/src/ui/map_events.test.ts#L117-L118) for an example. This PR just corrects the types!

Looking through the CHANGELOG, it appears `MapGeoJSONFeature` is a fairly new addition (2.1.8-pre.2), so I think this may just be on spot where the types got missed!

## Launch Checklist

    - While these types are part of the TypeScript public API, they are not core to the library API (nor currently documented on the docs site).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## main
 
+- Fix the type of the `features` property on `MapLayerMouseEvent` and `MapLayerTouchEvent` to be `MapGeoJSONFeature[]` in lieu of `GeoJSON.Feature[]` ([#2244]((https://github.com/maplibre/maplibre-gl-js/pull/2244)))
+
 ### ✨ Features and improvements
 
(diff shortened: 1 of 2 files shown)


HarelM:
Thanks for taking the time to fix this!
The migration from flow to typescript contained a lot of changes, so it makes sense this was missed.

HarelM (approved):

PR #2258, opened 2023-03-08
Docs: Example for stylemissingimage

tordans:
This is supposed to improve the example at https://maplibre.org/maplibre-gl-js-docs/api/map/#map.event:styleimagemissing by…

- adding `imageId` to the logged output and implicitly explaining in the code that `event.id` is the image's id
- adding the TypeScript event type


HarelM (approved):

HarelM:
THANKS!

Issue #2170, opened 2023-02-12
`text-offset` on line-placed symbol layer causes gaps or bunching


1ec5:
**maplibre-gl-js version**: 3.0.0-pre.4 (but this issue has been reported since at least 2016)

**browser**: Firefox 109 on macOS

### Steps to Trigger Behavior

 1. Add a symbol layer to the style that has a `text-field` and `symbol-placement` set to `line`.
 2. Set the y component of `text-offset` to a value other than 0.

### Link to Demonstration

In this test case, there are five labels that say “Lorem ipsum dolor sit amet”. Each is rendered by a symbol layer that is identical except for a different `text-offset`:

https://jsbin.com/kafefin/edit?js,output

`showCollisionBoxes` is enabled, though this demo disables collision to show that it isn’t a collision issue.

### Expected Behavior
In each label, “IP” should be right next to “SUM” without any overlap and without any gap, even though there is an angle in the linestring geometry at that point.

### Actual Behavior
In the labels above the dotted line (with a negative y-offset), “IP” overlaps with “SUM”.
In the labels below the dotted line (with a positive y-offset), there is an unsightly gap betwen “IP” and “SUM”.

<img src="https://user-images.githubusercontent.com/1231218/218297995-c461d2ff-ee88-42e2-ab2e-b6cd9a83b55f.png" width="400" alt="Lorem ipsum">

### Impact
Style authors y-offset labels for various purposes, most notably:

* The name of a country on either side of a border
* The name of a street on one side of the street (at low zoom levels when the street is drawn as a simple line)
* The name of a river on one side of the river (at low zoom levels when the river is drawn as a simple line)
* The name of a railway

This issue makes such labels look visually unattractive and difficult to read. For a practical example in a real, deployed style, see https://github.com/ZeLonewolf/openstreetmap-americana/pull/753#discussion_r1096574641.

This issue was previously reported and confirmed in mapbox/mapbox-gl-js#3718.

HarelM:
This is only relevant for Firefox 109 on macOS? I think I see it on my chrome on windows as well...

1ec5:
It happens regardless of the browser. I just filled in that portion of the issue template based on what I was using at the time.

HarelM:
Fixed by #2260

Issue #2171, opened 2023-02-12
Collision boxes drift toward angled geometry despite `text-offset`


1ec5:
**maplibre-gl-js version**: 3.0.0-pre.4 (but this issue has been reported since at least 2017)

**browser**: Firefox 109 on macOS

### Steps to Trigger Behavior

 1. Add two symbol layers to the style that each have a `text-field` and `symbol-placement` set to `line`.
 2. Set the Y component of each layer’s `text-offset` to a different value to spread the layers apart.

### Link to Demonstration

In this test case, there are two labels that say “Lorem ipsum dolor sit amet”. Each is rendered by a symbol layer that is identical except for a different `text-offset`. `text-ignore-placement` remains disabled:

https://jsbin.com/zuxuyap/edit?js,output

`showCollisionBoxes` is enabled: the blue circles are the collision boxes of the label that GL JS decided to show, while the red circles are the collision boxes of the label that GL JS decided to hide.

### Expected Behavior
Both labels should be visible. The collision boxes should approximately align with the rendered glyphs.

### Actual Behavior
The second layer’s label collides out the first layer’s label, even though there’s plenty of room to show both. Both labels’ collision boxes drift toward the angle in the geometry instead of staying consistently offset:

<img src="https://user-images.githubusercontent.com/1231218/218298710-82509d31-9253-4d15-8efa-c7fe9d8b0045.png" width="400" alt="Lorem ipsum">

### Impact
Typically, you’d style an international border by labeling a different country on either side of the border. This requires `text-offset`. However, this issue means that, on a sufficiently curved border, only one of the two labels will ever appear. For a practical example in a real, deployed style, see https://github.com/ZeLonewolf/openstreetmap-americana/pull/753#issuecomment-1413315442.

Depending on the style, it may be possible to partially work around this issue by setting `text-ignore-placement` to `true`, disabling collision detection. However, the issue still reproduces, just less frequently. Moreover, in some styles, this may cause other layers’ labels to overlap, making them unreadable.

This issue was previously reported and confirmed in https://github.com/mapbox/mapbox-gl-js/issues/4798#event-1613841784.

HarelM:
Isn't this basically the same as #2170?

1ec5:
[...]

PR #2260, opened 2023-03-08
Improve line labels using text-offset

ChrisLoer:
## Summary
The status quo for line labels in MapLibre is that if a vertical offset is applied via `text-offset`, the glyph location is just moved away from its line in the direction of the normal vector. The problem with this approach is that at convex line vertices, this opens up gaps in the glyph placement, and at concave vertices, it causes glyphs to be jumbled on top of each other. (@1ec5 reported this as https://github.com/maplibre/maplibre-gl-js/issues/2170)

This PR improves the behavior by offsetting the actual line geometry and then placing glyphs along that geometry with no further offset. In practice, this means shifting the line segments based on the offset, and then extending or shortening adjacent segments towards their point of intersection. Here's a clumsy sketch:

Screenshot 2023-03-08 at 10 06 58 AM

There are no public API changes with this PR: I view this as functionality that's implicitly promised by the style spec but not yet delivered.

## Motivation

This change is important to us at Felt because we want to make it really easy to add line data to a map and style it well. Defaulting to having line labels offset from their lines is a much more flexible choice with this goal (you don't have to coordinate your text/halo color/width with your line color/width in order to be legible). Here are some examples:

**Felt railway layer w/ default `text-max-angle`:** line labels look bad at curves
Screenshot 2023-03-08 at 12 35 08 PM

**Felt railway layer w/ default `text-max-angle: 25`:** worst offenders removed, but we lose density (this is the compromise we're currently using)
Screenshot 2023-03-08 at 12 36 43 PM

**Felt railway layer w/ this PR:** 🎉 
Screenshot 2023-03-08 at 12 38 44 PM

## Testing and Performance
Look at the [test update commit diff](https://github.com/maplibre/maplibre-gl-js/pull/2260/commits/75819a79c41537801b7a0eed9863bc43a8a24c15) to see the change of behavior as captured in the tests. For a gently curving line, you can see the effect is just more even spacing. The "sharp" line tests expose some of the broken behavior of the previous implementation (although this behavior would normally be hidden by the default `text-max-angle` of 45 degrees). The point of the sharp tests is not that we want line labels that look like that, but just to exercise some extreme values. Note that the `line-sharp-outside-center` test is still showing the same behavior as before because the line extension algorithm doesn't handle the special case where the anchor point is _exactly_ on a line vertex. This is a solvable problem but doesn't fit in easily with the algorithm as currently written. I decided it was OK to leave as-is because (1) it's not a regression from current behavior, and (2) in most maps it should be pretty rare for anchor points to line up exactly with line vertices.

For non-offset line labels, this should have a pretty trivial effect on performance. For line labels with offsets, this definitely adds cost that scales with the number of label-covered line vertices, and this is a performance hot spot. Still, the costs should be pretty modest (some point geometry arithmetic and the line intersection math), on the order of what we currently incur for line vertex projection. For someone trying to use this functionality, I think it's a pretty good bet that they'd prefer to have the slightly slower but also less-broken behavior from this PR.

I ran the "Paint" benchmark, which uses MapTiler Streets, so in theory this should be a test of non-offset performance, which shouldn't be noticeably changed. I didn't see any red flags in the benchmark, but it also seemed to me like results weren't very reproducible between runs, so I don't have a ton of confidence in this as a signal.

Screenshot 2023-03-07 at 2 38 56 PM

## License Compliance
I was at Mapbox when @rreusser was working on an alternative solution to this problem in which actual offset curves are computed to place glyphs along as they go around line vertices. That approach definitely looks better, but also has more code complexity (and possibly perf impact? I haven't evaluated). My approach here is simpler, just trying to get legibility parity with non-offset line labels. In terms of license compliance -- it's been ~two years since I looked at his PR, and this is following a different approach, so I think this should be clean.

cc @ibesora

## Launch Checklist

 - [x] Link to related issues.
 - [x] Post benchmark scores.


ChrisLoer:
I just took a second look at https://github.com/maplibre/maplibre-gl-js/issues/2171

I'm not seeing that behavior with collision circles using this branch -- it looks like they're following what's expected:
Screenshot 2023-03-08 at 1 04 54 PM

I updated the "pathVertices" used for collision detection to follow the offset line, so that may have incidentally fixed that issue, but I haven't specifically tested the "two layers following same line" case.

ChrisLoer:
This PR also includes some un-related formatting changes that I _think_ are just coming from the .eslintrc settings in the project, but let me know if I should strip those out.

One thing -- I did a bunch of my interactive testing using my own local copy of `debug-pages/debug.html`, but it doesn't seem like it's in out-of-the-box working shape? (asks for fonts that aren't available, etc.) What are the go-to local debug options these days? Should I push some changes to `debug.html` to get it to render better, or put that in a separate PR?

In terms of which release to target, I assume it makes sense to merge this to main and then maybe if necessary cherry-pick to the 2.4 branch?

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ## main
 
-- Fix the type of the `features` property on `MapLayerMouseEvent` and `MapLayerTouchEvent` to be `MapGeoJSONFeature[]` in lieu of `GeoJSON.Feature[]` ([#2244]((https://github.com/maplibre/maplibre-gl-js/pull/2244)))
+- Fix the type of the `features` property on `MapLayerMouseEvent` and `MapLayerTouchEvent` to be `MapGeoJSONFeature[]` in lieu of `GeoJSON.Feature[]` ([#2244](<(https://github.com/maplibre/maplibre-gl-js/pull/2244)>))
```
HarelM, line 3: Not sure the < are needed, maybe there's an extra brackets here though.

HarelM:
I'll review it more thoroughly in the coming days.
A few minor comments from an initial scan:
Would be good to have some unit tests added as well to cover the code changes your made (unless I missed them...)
would be nice to improve projection cache typea bit more to indicate what the objects hold.

If I understand the idea correctly you take the offset vectors and cut them or prolong them as needed to create a new line that doesn't cut itself and doesn't have gaps, right? And only afterwards place the text.

ChrisLoer:
> Would be good to have some unit tests added as well to cover the code changes your made (unless I missed them...)

The added tests are all render tests under `text-offset`. Do you have a strong preference for "unit" vs "render" test here? My preference is "render", because:
- Way less "mocking" done, instead you use a real style
- Specification of the test is much more intelligible, both in the style.json and in the expected output
- It's still pretty well isolated to the specific thing we're testing, in the style of a unit test.

>would be nice to improve projection cache typea bit more to indicate what the objects hold.

:+1:

> If I understand the idea correctly you take the offset vectors and cut them or prolong them as needed to create a new line that doesn't cut itself and doesn't have gaps, right? And only afterwards place the text.

That's right. For historical background, in the original implementation there was no CPU projection (line label positioning and offsets were done entirely in the vertex shader). That's the constraint that led to the weird offset behavior in the first place, and we just never re-visited once we switched to CPU projection.

One thing about the algorithm is that it traverses the line from the anchor point for every single glyph to find a location. There's probably a somewhat more efficient algorithm in which we only the traverse the line once as we place glyphs, but IIRC when we did the original profiling the expensive part of re-traversing the line was just re-doing the projection work, so we just cached that as we went. Placing the first and last glyph before doing the rest of them is an important optimization, though, because it allows you to tell if the label is going to fit (and avoid collision) before you actually do the rest of the layout.

HarelM:
Render tests are great, but you can cover a lot more with unit tests. Especially with all this math logic and edge cases.
I think that this PR deserves both render and unit tests.
Generally speaking, I would like to get to around 80% code coverage, so we need to make sure we cover the files we change well...

1ec5 (commented):
Thanks for fixing this issue! It’s going to make a big difference for styles that adopt the North American convention of labeling roads, rivers, and boundaries as simple strokes with offset labels.

Before | After
----|----
[Sault Ste. Marie before](https://zelonewolf.github.io/openstreetmap-americana/#map=11/46.5025/-84.2927&language=en) | [Sault Ste. Marie after](https://zelonewolf.github.io/openstreetmap-americana/#map=11/46.5025/-84.2927&language=en)

When the map is tilted and the layer has `text-pitch-alignment` set to `viewport`, some symbols that were previously poorly kerned have better kerning, but individual characters sometimes get rotated incorrectly:

Before[^ireland-before] | After
----|----
[Ireland before](https://zelonewolf.github.io/openstreetmap-americana/#map=13/54.16078/-7.28633/0/45&language=en) | [Ireland after](https://zelonewolf.github.io/openstreetmap-americana/#map=13/54.16078/-7.28633/0/45&language=en)

The changes don’t seem to improve the kerning on layers that have `text-anchor` set to `bottom`:

[Mill Valley after](https://zelonewolf.github.io/openstreetmap-americana/#map=15/37.88445/-122.5264&language=en)

All the screenshots above link to coordinates in OSM Americana that you can play with, but let me know if you need a reduced test case.

[^ireland-before]: To see this in action, run: `map.setLayoutProperty("boundary_country_label_left", "text-pitch-alignment", "viewport"); map.setLayoutProperty("boundary_country_label_right", "text-pitch-alignment", "viewport")`

test/integration/render/tests/text-offset/line-sharp-outside-center/style.json
```diff
@@ -0,0 +1,55 @@
+{
+  "version": 8,
+  "metadata": {
+    "test": {
+      "description": "This test is tricky because the anchor point is exactly the point of inflection. Ideally we'd construct something like a synthetic anchor point with the offset, but for now we'll tolerate just allowing a gap to open here.",
```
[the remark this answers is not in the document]
ChrisLoer: Just to make sure it's clear: the problem this test exercises is that the anchor point for the label is directly on a vertex. The fact that it's a sharp angle just makes the break more dramatic. In this special case, we degenerate to the same behavior we had before the PR, for glyphs on either side of the starting vertex.

>Do you think there’s an opportunity to be consistent with word-wrapping on point-placed labels, trying to align these very sharp corners with word boundaries instead of arbitrary characters? If this isn’t straightforward, it would be a nice feature to track separately.

That's an interesting idea! The heuristic would be something like "if you hit a vertex that's sharper than `text-max-angle`, but can back up to a space that's less than X units away from the vertex, try breaking the glyph stream there, jumping ahead to past the vertex and then _if_ you can place at least one word before having to jump over vertices again, continue"?  But I think it definitely makes sense to treat as a separate feature.

src/symbol/projection.ts
```diff
@@ -347,6 +345,28 @@ function projectTruncatedLineSegment(previousTilePoint: Point, currentTilePoint:
     return previousProjectedPoint.add(projectedUnitSegment._mult(minimumLength / projectedUnitSegment.mag()));
 }
 
+function findIntersectionPoint(currentA: Point, currentB: Point, nextA: Point, nextB: Point): Point {
```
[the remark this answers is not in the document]
ChrisLoer: Yeah, I didn't see it in our existing code. It might make sense to promote to utils, but this version is slightly less general purpose because of its assumption that parallel line segments must also touch at currentB/nextA.

ChrisLoer:
>When the map is tilted and the layer has text-pitch-alignment set to viewport, some symbols that were previously poorly kerned have better kerning, but individual characters aren’t being rotated to match the new positions:

I'm not sure if "not being rotated" is the right description of what's going on? The glyphs are oriented to the normal of the offset line that they're place on. I think one thing that's changed is with a sharp convex angle, the intersection point is on the "inside" of the curve, which means switching to the next line happens sooner. This is the reason you see that last "D" rotated in the Ireland example -- it's shifted on to the next line. This behavior is the same as with glyphs placed along the centerline, but maybe it looks _worse_ because it's not directly on top of the visual reference that kind of explains why it's happening (it looks bad either way).

I experimented a couple years back with interpolating the glyph rotation as you approached a sharp angle, and I don't have the screenshots anymore, but I remember that it looked surprisingly _weird_. The better solution (for both offset and centered text) is what I mentioned @rreusser was working on -- define a curve that goes around every vertex, and place points along it. I think that might be the best solution! But it adds more complexity and cost, which might be part of why it never merged to Mapbox GL.

One "cheap" experiment I haven't tried is just boosting the spacing in proportion to how sharp a curve is (e.g. normal spacing up to 25 degrees, and then maybe linearly scale to 1em of extra spacing on the way to 90 degrees). I doubt it will look great, but it might lessen the worst parts of things like that I - RELAN - D with near-90-degree angles. I may try this tomorrow.

> The changes don’t seem to improve the kerning on layers that have text-anchor set to bottom:

Yeah unfortunately `text-anchor` has a totally different implementation -- the "offset" in that case is applied in the shaping process on the worker and is baked into the glyph quads (not the smaller x-only "glyphOffsetArray" that's passed to the foreground projection code). You could untangle this, but there would be some pretty annoying parts. I think this makes sense to treat as a separate feature as well (the shape of the feature would be something like "get the text-anchor effect for lines out of the shaping and passed forward to use in the foreground line offset behavior).

Thank you for the screenshots with links -- very clear!

1ec5:
> I just took a second look at #2171
> 
> I'm not seeing that behavior with collision circles using this branch -- it looks like they're following what's expected:

Yes, I can confirm in [this area](https://zelonewolf.github.io/openstreetmap-americana/#map=14/54.12535/-7.3072&language=en) that the collision circles are in the expected positions. Unfortunately, that leaves me searching for a new theory as to why the boundary edge labels in https://github.com/ZeLonewolf/openstreetmap-americana/pull/753#issuecomment-1413315442 keep going missing, but it’s unlikely to be due to what you’re touching.

> I'm not sure if "not being rotated" is the right description of what's going on? The glyphs are oriented to the normal of the offset line that they're place on. I think one thing that's changed is with a sharp convex angle, the intersection point is on the "inside" of the curve, which means switching to the next line happens sooner. This is the reason you see that last "D" rotated in the Ireland example -- it's shifted on to the next line. This behavior is the same as with glyphs placed along the centerline, but maybe it looks _worse_ because it's not directly on top of the visual reference that kind of explains why it's happening (it looks bad either way).

Ah, I missed that little kink in the line beside the “D”. I’m not sure how the symbol is being placed here, since that angle has got to be more than the 20&nbsp;degrees specified in `text-max-angle` for that layer. I can keep investigating and file a separate issue if you think it’s unrelated.

> Yeah unfortunately `text-anchor` has a totally different implementation -- the "offset" in that case is applied in the shaping process on the worker and is baked into the glyph quads (not the smaller x-only "glyphOffsetArray" that's passed to the foreground projection code). You could untangle this, but there would be some pretty annoying parts. I think this makes sense to treat as a separate feature as well (the shape of the feature would be something like "get the text-anchor effect for lines out of the shaping and passed forward to use in the foreground line offset behavior).

Makes sense. Fortunately, the workaround in many cases could be to keep `text-anchor` as `center` and instead increase the `text-offset` by roughly half of `text-size`.

src/symbol/collision_index.ts
```diff
@@ -2,19 +2,19 @@ import Point from '@mapbox/point-geometry';
 import clipLine from './clip_line';
 import PathInterpolator from './path_interpolator';
 
-import * as intersectionTests from '../util/intersection_tests';
-import GridIndex from './grid_index';
 import {mat4, vec4} from 'gl-matrix';
 import ONE_EM from '../symbol/one_em';
+import * as intersectionTests from '../util/intersection_tests';
```
HarelM, line 7: Can you please import specific stuff here instead of all the file? I would like to avoid as much as possible these types of imports.

HarelM: Also why are all these import here? I don't see changes in the code in this file that requires all these imports...

src/symbol/projection.ts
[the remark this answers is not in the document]
HarelM: I would also consider moving to this to utils or use the relevant part of turf even?
If this method has a flavor I would consider adding this as an optional parameter or handling it outside the method, maybe, IDK...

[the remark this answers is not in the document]
HarelM: Please also add jsdocs comments above the method if is has a special flavor.

src/symbol/projection.ts
```diff
@@ -347,6 +345,37 @@ function projectTruncatedLineSegment(previousTilePoint: Point, currentTilePoint:
...
+
+    const originDeltaY = currentA.y - nextA.y;
+    const originDeltaX = currentA.x - nextA.x;
+    const currentInterpolation = (nextDeltaX * originDeltaY - nextDeltaY * originDeltaX) / denominator;
+
+    // Find intersection by projecting out from origin of first segment
+    return new Point(currentA.x + (currentInterpolation * currentDeltaX), currentA.y + (currentInterpolation * currentDeltaY));
+}
+
+// We calculate label-plane projected points for line vertices as we place glyphs along the line
```
HarelM, line 370: Use jsdocs comments here instead.

src/symbol/projection.ts
```diff
@@ -387,50 +414,108 @@ function placeGlyphAlongLine(
 
     let current = anchorPoint;
     let prev = anchorPoint;
+
+    // offsetPrev and intersectionPoint are analagous to prev and current
+    // but if there's a line offset they are calculated in parallel as projection happens
+    let intersectionPoint;
```
HarelM, line 420: Can these new variables have type, please :-)

src/symbol/projection.ts
```diff
@@ -387,50 +414,108 @@ function placeGlyphAlongLine(
...
-        if (currentIndex < lineStartIndex || currentIndex >= lineEndIndex)
-            return null;
-
-        prev = current;
-        pathVertices.push(current);
-
-        current = projectionCache[currentIndex];
-        if (current === undefined) {
-            const currentVertex = new Point(lineVertexArray.getx(currentIndex), lineVertexArray.gety(currentIndex));
+    function projectVertexToViewport(index: number) {
```
HarelM, line 428: Can we move this function outside the other function? it creates a dependecy on variables that is later harder to untangle.

src/symbol/projection.ts
```diff
@@ -387,50 +414,108 @@ function placeGlyphAlongLine(
...
+            } else {
+                // This is the end of the line, no intersection to calculate
+                projectionCache.offsets[index] = offsetCurrent;
             }
+            return projectionCache.offsets[index];
         }
+    }
+
+    let currentLineSegment;
+    while (distanceToPrev + currentSegmentDistance <= absOffsetX) {
```
HarelM, line 474: Can this while also be moved to its own method? I'm having a hard time following this huge method...

src/symbol/projection.ts
```diff
@@ -387,50 +414,108 @@ function placeGlyphAlongLine(
...
-                current = projectTruncatedLineSegment(previousTilePoint, currentVertex, prev, absOffsetX - distanceToPrev + 1, labelPlaneMatrix, getElevation);
+                return projectTruncatedLineSegment(previousTilePoint, currentVertex, prev, absOffsetX - distanceToPrev + 1, labelPlaneMatrix, getElevation);
+            }
+        }
+    }
+
+    function findOffsetIntersectionPoint(index: number, prevToCurrentOffset: Point) {
+        if (projectionCache.offsets[index]) {
+            return projectionCache.offsets[index];
+        } else {
```
HarelM, line 453: No need for else if you return in the if, right?

src/symbol/projection.ts
```diff
@@ -387,50 +414,108 @@ function placeGlyphAlongLine(
...
     }
 
     // The point is on the current segment. Interpolate to find it.
     const segmentInterpolationT = (absOffsetX - distanceToPrev) / currentSegmentDistance;
-    const prevToCurrent = current.sub(prev);
-    const p = prevToCurrent.mult(segmentInterpolationT)._add(prev);
-
-    // offset the point from the line to text-offset and icon-offset
-    p._add(prevToCurrent._unit()._perp()._mult(lineOffsetY * dir));
+    const p = currentLineSegment.mult(segmentInterpolationT).add(offsetPrev || prev);
```
HarelM, line 518: use `_add` to improve performance a bit?

src/symbol/projection.ts
```diff
@@ -387,50 +414,108 @@ function placeGlyphAlongLine(
...
                     tileAnchorPoint :
                     new Point(lineVertexArray.getx(previousLineVertexIndex), lineVertexArray.gety(previousLineVertexIndex));
                 // Don't cache because the new vertex might not be far enough out for future glyphs on the same segment
-                current = projectTruncatedLineSegment(previousTilePoint, currentVertex, prev, absOffsetX - distanceToPrev + 1, labelPlaneMatrix, getElevation);
+                return projectTruncatedLineSegment(previousTilePoint, currentVertex, prev, absOffsetX - distanceToPrev + 1, labelPlaneMatrix, getElevation);
+            }
+        }
+    }
+
+    function findOffsetIntersectionPoint(index: number, prevToCurrentOffset: Point) {
```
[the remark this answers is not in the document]
HarelM: Also worth adding some jsdocs to this method to describe how it works and what it does.

HarelM:
@ChrisLoer thanks a lot for taking the time to do this!
I know I added some annoying comments but I believe this is for the greater good. :-)

src/symbol/collision_index.ts
ChrisLoer: All that changed here is that the import order is sorted. I thought this was picking up from the project's .eslintrc, but maybe this is coming from something local in my vscode setup.

src/symbol/projection.ts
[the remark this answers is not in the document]
ChrisLoer: I think https://www.npmjs.com/package/@turf/line-intersect would be the closest turf candidate? But in that case we would have to extend the lines ourselves and convert to LineString.

Move to util and removing the special case seems fine though -- the more general purpose response to "these lines are parallel" is to return a null value, and we could easily hook up the calling code to respond to null by choosing the known intersection point.

src/symbol/projection.test.ts
```diff
@@ -1,4 +1,4 @@
-import {project} from './projection';
+import {findIntersectionPoint, project} from './projection';
```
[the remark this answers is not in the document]
ChrisLoer: I agree the logic of this algorithm is hard to follow, and that writing unit tests would push hard towards refactoring. I'm reluctant to jump too deep into a significant refactor of the algorithm, but I think the obvious next incremental step is to factor out the currently inline function definitions, so I'll give that a try.

src/symbol/projection.ts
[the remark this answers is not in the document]
1ec5: Sorry for mentioning Turf, that was a bit of a distraction because this function is more specialized. There should be a comment explaining that nuance so that others don’t latch onto this function thinking it’s like line-intersect.

src/symbol/projection.ts
ChrisLoer: Can you point to an example of the preferred way to do jsdoc style comments for a type declaration? Is this asking for individual tags in the comment that describe `projections` and `offsets`?

HarelM: Sure, here is one:
https://github.com/maplibre/maplibre-gl-js/blob/2531d6a5cbd4a2e6521155bc82ab2ed23016a6bc/src/ui/events.ts#L267

src/symbol/projection.test.ts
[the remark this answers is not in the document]
ChrisLoer: I think the convex/concave/parallel unit tests for the intersection finding logic came out pretty decently -- I like that playing with them makes it easier to reason about the math involved step-by-step without just relying on whether the visual result looks right.

All of the refactoring so far was pretty mild (there are some small differences in what temporary objects get created, etc.), which I think makes sense. Trying to re-write significant portions of the entire algorithm seems definitely out-of-scope to me. While we have pretty good functional test coverage of this code and we have _some_ benchmarking (via the Paint benchmark), this is highly perf-sensitive code so I don't want to make major changes to the existing algorithm without being ready to invest a lot more time in evaluating the changes.

One downside of unit tests I meant to articulate earlier: they push somewhat towards ossifying internal data structures because updating unit tests increases the friction to making otherwise internal changes. I think this downside really starts biting when the surface area of the unit test gets too big.

src/symbol/projection.ts
ChrisLoer: Thanks! Updated accordingly.

ChrisLoer:
@1ec5 It's maybe worth noting along with the `text-anchor` gotcha -- `text-max-angle` is calculated entirely in tile space (see `check_max_angle.ts`), so if you're in a pitched map the angles you're seeing on the screen may be different than the ones used for the calculation.

HarelM (approved):

PR #2273, opened 2023-03-13
Keep discriminated union of MapGeoJSONFeature["layer"]

tomhicks:
Using `Omit` on a union usually results in a pretty unusable type. What you generally want to do is use a DistributiveOmit which performs the omit on each item in the union. This retains the discriminant across the union so you can still use it.

Here's a SO question on why `Omit` doesn't work in this case: https://stackoverflow.com/questions/67794339/why-doesnt-discriminated-union-work-when-i-omit-require-props

A specific example from our codebase:

```typescript
if (layer.type === "fill" && layer.paint?.["fill-opacity"] !== 0) return true
```

This isn't possible currently, because you cannot use the `type === "fill"` check to discern the `paint` type. As a result, you end up needing this instead:
```
if (f.layer.paint && (("fill-opacity" in f.layer.paint && f.layer.paint["fill-opacity"] !== 0) return true
```

There's a valid question as to where the DistributiveOmit should live. I've added it at the bottom of this file which is a bit meh. Other options include:
- `types/utils.ts`
- `import { DistributiveOmit } from "some-library"`

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
This looks great in general, except what you pointed about the location of the new type.
I would simply place it at the beginning of this file ATM.
But if there other omits in the code it's worth considering moving it to a common place.

HarelM:
Please fix failing PR checks...

HarelM (approved):

PR #2097, opened 2023-01-25
Throttle the image request queue while the map is moving to improve performance

zhangyiatmicrosoft:
## Launch Checklist

This is the same as https://github.com/maplibre/maplibre-gl-js/pull/1856 which will be abandoned by AdamS108.
Need transfer the ownership over to me.

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.
 - [x] Link to related issues.
 - [x] Include before/after visuals or gifs if this PR includes visual changes.
 - [x] Write tests for all new functionality.
 - [x] Document any changes to public APIs.
 - [x] Post benchmark scores.
 - [x] Add an entry to `CHANGELOG.md` under the `## main` section.

PR #2278, opened 2023-03-16
Fix perf regression introduced by PR 2097

zhangyiatmicrosoft:
## Launch Checklist

Fix ~50ms regression introduced by PR #2097 

- When previous set of requests are sent and not finished, image queue appears dirty so _render kicks of another frame and archives nothing. In fact, _imageQueueDirty flag is not needed because when images are received, other code (usually raster tile) will trigger rendering.

- When not throttled, calling processQueue does nothing but wasting time

Therefore, remove this._imageQueueDirty field and only call process queue when throttled (which means moving)

 - [x] Link to related issues.

### DIFF
--- a/src/ui/map.test.ts
+++ b/src/ui/map.test.ts
@@ -18,6 +18,7 @@ import Terrain, {} from '../render/terrain';
 import {mercatorZfromAltitude} from '../geo/mercator_coordinate';
 import Transform from '../geo/transform';
 import {StyleImageInterface} from '../style/style_image';
+import ImageRequest from '../util/image_request';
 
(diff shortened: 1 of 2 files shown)


HarelM (approved):

HarelM:
Changelog entry and a test are missing...

zhangyiatmicrosoft:
> Changelog entry and a test are missing...

Thanks for the reminder. Added UT.
This change is more or less internal, and previous PR 2097 is not released yet, so I skipped Changelog, keeping it cleaner and more meaningful for readers.

Issue #2030, opened 2023-01-03
Firefox WebGL warning (Alpha-premult and y-flip are deprecated for non-DOM-Element uploads).


jwoodwardtfx:
When running in Firefox we get the following warning:
**_WebGL warning: texImage: Alpha-premult and y-flip are deprecated for non-DOM-Element uploads._**

Google shows up the following:
https://bugzilla.mozilla.org/show_bug.cgi?id=1400077
https://github.com/mapbox/mapbox-gl-js/issues/5292

Is this an issue or a performance limitation we should be concerned about (as the Firefox developers say this path isn't optimized and should not be used)?

morgoth:
Since the issue is known at least from 2017, vide https://github.com/mapbox/mapbox-gl-js/issues/5292 is there any plan to fix/workaround/address this problem?

HarelM:
Can you please share a reproduction of this issue?
Generally speaking, Firefox has a very low usage statistics when it comes to browsers, but if someone would like to invest in fixing this, please do.
If this is a performance concern I might be able to assign a bounty to it, but I'll need a clear reproduction example first

morgoth:
You can just open this page in Firefox https://maplibre.org/maplibre-gl-js-docs/example/simple-map/ and look into the console.

I'm using Firefox 110.0 (64-bit) on Ubuntu

HarelM:
Would you be interested in working on this issue?

morgoth:
@HarelM I'm not even an user yet ;-) - just investigating future possibilities for the integration and that was the first "warning" to see the known issue opened since 2017 without addressing.

IvanSanchez:
I can have a look at this.

I'm having a look at the code paths triggering the warnings, and am currently considering skipping the creation of textures when the atlases are empty. This won't get rid of all the warnings, but it should avoid telling the GPU to allocate 1x1 px textures needlessly.

IvanSanchez:
@HarelM Sure thing!

Have a look at #2297 - it gets rid of all warnings *except for the first one*. I'll have to look at the root cause of that one.

HarelM:
I can't think of a strong argument to backport this into 2.4. we should be releasing 3.0 shortly (hopefully).

WebMechanic:
> I can't think of a strong argument to backport this into 2.4. we should be releasing 3.0 shortly (hopefully).

If 3.0 is virtually a seamless upgrade from 2.4 w/o any BC breaks, then sure.

[...]

PR #2297, opened 2023-03-23
Skip creating textures for empty atlases

IvanSanchez:
Indirectly fixes most of the deprecation warnings from #2030.

By making the tile worker return `undefined` instead of empty 1x1 atlases, the main thread skips creating textures for those. Those empty textures were the source of most of the WebGL warnings, AFAICS.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
Thanks for taking the time to open this PR!!
If this is still not complete, please use a draft PR.
Otherwise, please add changelog and tests.
Also would be nice to simply not initialize the image atlas and glyph atlas, if this is not possible I would feel better if empty would be a calculated property (getter only) according to the state of the class.

IvanSanchez:
Initialization of the `empty` property of atlases should be fine now. I don't think it's possible to properly skip atlas initialization, as the worker code refers to the positions inside the (empty) atlases. Since this is my first foray into the MapLibre code, I'd rather make minimal changes.

Regarding tests, I'm unfamiliar with MapLibre's test scaffolding. I'm not sure if it's even possible to run an automated puppeteer/karma/whatever browser test on a headless Firefox instance and check for WebGL warnings in the console.

IvanSanchez:
...and **of course** the unit tests assume that there are atlas images and textures for every tile, despite the `if (data.imageAtlas) {` line in `tile.ts` :angry: 

```
file:///home/runner/work/maplibre-gl-js/maplibre-gl-js/src/render/draw_fill_extrusion.ts:77
            tile.imageAtlasTexture.bind(gl.LINEAR, gl.CLAMP_TO_EDGE);
                                   ^
TypeError: Cannot read properties of undefined (reading 'bind')
```

IvanSanchez:
The changes impact the behaviour of one specific render test: extrusion pattern with missing texture. The current expectation is a transparent image, but removing the texture (and running the shader with no bound texture) creates a black extrusion.

@HarelM Would this change in behaviour be OK? Otherwise it'd be possible to create an empty transparent texture just for this edge case.

HarelM:
Can you fix the test so that I can see the change in the image?
I'm not sure I fully understand how it would look in this specific case.
Also maybe worth trying to find out when was this test added and the changes around it to maybe better understand the motivation behind this test and this behavior.

HarelM:
The code you added is fairly simple and I don't see big issues with it.
There should be either failing unit tests that needs to be fixed as part of this PR or new unit tests that needs to be added - not just render tests.

IvanSanchez:
> I'm not sure I fully understand how it would look in this specific case.

So the problematic render test is `tests/fill-extrusion-pattern/missing`. AFAICS, it checks that a *faulty* stylesheet (one referring to a pattern image that doesn't exist in the spritesheet) doesn't crash the library. It does call `console.warn` to inform about this, so I'm not worrying much about changing the behaviour (since IMO using a non-existing image pattern is undefined behaviour).

For reference, most of `tests/fill-extrusion-pattern/` looks like:

image

...except for `tests/fill-extrusion-pattern/missing`, which currently is a **blank image**.

After my changes, that specific render test looks like:

image

...this difference happens because, before this PR, tiles with no image atlas would allocate a transparent 1x1 texture - drawing an extrusion pattern with an invalid image would use that transparent 1x1 texture.

Since this PR skips the creation of those empty textures, the shader uses no texture - and the texture sampler somehow defaults to solid black.

Hope that helps?

HarelM:
Yes, it does! Sorry for the delay in response.
I don't think this edge case is interesting, but this is a breaking change so we need to document it.

HarelM:
Please add unit test to cover the new functionality (besides fixing the render tests).

HarelM:
Bummer, we didn't add it to version 3. :-(
Is there a way to keep this backward compatible somehow?
Also unit tests are still missing...

HandyMenny:
This is the PR that added `tests/fill-extrusion-pattern/missing`: https://github.com/mapbox/mapbox-gl-js/pull/4687/

I would say that trasparent texture is intended behaviour

johncarmack1984:
The warnings this was chasing (#2030) were fixed by #7128 in February. The empty 1x1 atlases are still built and uploaded on main, so the idea stands, but the unit tests and the backward-compatible answer for the missing-pattern render test never came. This one is closable.

HarelM:
If you think this is still worth solving please open a new PR. Thanks!

Issue #2035, opened 2023-01-04
setTerrain causes hillshading opacity to increase and detail to decrease


zdila:
**maplibre-gl-js version**: 2.4.0 or 3.0.0-pre.3

**browser**: Vivaldi, Chrome, Firefox

### Steps to Trigger Behavior

 1. render map with hillshading
 2. enable 3D terrain

### Link to Demonstration

https://codesandbox.io/s/maplibre-gl-js-stron-shading-issue-30pke4

### Expected Behavior

After pressing "Set terrain" button shading opacity should not change and hillshading detail should stay the same.

### Actual Behavior

After pressing "Set terrain" button the shading opacity rapidly increases and hillshading detail decreases.

HarelM:
Is this the same as #1317?

HarelM:
Probably not, this is strange...
Can you play around with the style to create a more simple reproduction?
I haven't seen this for other hillshade stuff, so it might be some configuration of the hillshare layer.
If possible, please create a reproduction that the entire style is in the codesandbox so that we'll be able to narrow down what's the cause of this issue.
Thanks!

acalcutt:
One thing that I notice is you are using the same 'terrain_rgb' source for both terrain and hillshade. In the past this was not recommended due to other issues like this (i'd have to look back and find the PR)... I'm not sure if that has changed.

I the demo style for https://demotiles.maplibre.org/terrain-tiles/ it uses the same source twice.
https://demotiles.maplibre.org/styles/osm-bright-gl-terrain/style.json
```
	"sources": {
		"openmaptiles": {
			"type": "vector",
			"url": "https://demotiles.maplibre.org/tiles-omt/tiles.json"
		},
		"TerrainSource": {
			"type": "raster-dem",
			"url": "https://demotiles.maplibre.org/terrain-tiles/tiles.json"
		},
		"HillshadeSource": {
			"type": "raster-dem",
			"url": "https://demotiles.maplibre.org/terrain-tiles/tiles.json"
		}
	},
```

acalcutt:
I think this is what I was thinking of https://github.com/maplibre/maplibre-gl-js/issues/1186#issuecomment-1100927180

rotu:
There are some warnings picked up by webgl-lint around terrain. Possibly related, possibly a red herring. https://osmus.slack.com/archives/C01G3D28DAB/p1671217859071759

acalcutt:
I added a second source for terrain to you example here 
[...]

PR #2298, opened 2023-03-23
Warnonce for Terrain Source

tempranova:
Issue described in to https://github.com/maplibre/maplibre-gl-js/issues/2035.

This change adds a `warnonce` when it detects that a user is using the same source for a hillshade and for 3D terrain. The check only runs when `setTerrain()` is called. I do not check for the visibility of the hillshade layer, and also do not conduct a similar check when a hillshade layer is turned on or added to the map.

I did not include a debug or test page for this, but happy to if required!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 - Update CONTRIBUTING.md with details on setting up on M1 mac ([#2196](https://github.com/maplibre/maplibre-gl-js/pull/2196))
 - Update default type of originalEvent in MapLibreEvent to be `unknown` ([#2243](https://github.com/maplibre/maplibre-gl-js/pull/2243))
 - Improve performance when forcing full symbol placement by short circuiting pause checks ([#2241](https://github.com/maplibre/maplibre-gl-js/pull/2241))
-- _...Add new stuff here..._
+- Adding a `warnonce` when terrain and hillshade source are the same
 
 ### 🐞 Bug fixes
 
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -1697,6 +1697,13 @@ class Map extends Camera {
             // add terrain
             const sourceCache = this.style.sourceCaches[options.source];
             if (!sourceCache) throw new Error(`cannot load terrain, because there exists no source with ID: ${options.source}`);
+            // Warn once if user is using the same source for hillshade and terrain
+            for(const index in this.style._layers) {
+              const thisLayer = this.style._layers[index]
+              if(thisLayer.type === 'hillshade' && thisLayer.source === options.source) {
+                warnOnce("You are using the same source for a hillshade layer and for 3D terrain. Please consider using two separate sources to improve rendering quality.");
+              }
+            }
             this.terrain = new Terrain(this.painter, sourceCache, options);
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -15,7 +15,7 @@
 - Update CONTRIBUTING.md with details on setting up on M1 mac ([#2196](https://github.com/maplibre/maplibre-gl-js/pull/2196))
 - Update default type of originalEvent in MapLibreEvent to be `unknown` ([#2243](https://github.com/maplibre/maplibre-gl-js/pull/2243))
 - Improve performance when forcing full symbol placement by short circuiting pause checks ([#2241](https://github.com/maplibre/maplibre-gl-js/pull/2241))
-- _...Add new stuff here..._
+- Adding a `warnonce` when terrain and hillshade source are the same
```
HarelM, line 18: Please add a link to PR and keep the `_...Add new stuff here..._` ATM.

src/ui/map.ts
```diff
@@ -1697,6 +1697,13 @@ class Map extends Camera {
             // add terrain
             const sourceCache = this.style.sourceCaches[options.source];
             if (!sourceCache) throw new Error(`cannot load terrain, because there exists no source with ID: ${options.source}`);
+            // Warn once if user is using the same source for hillshade and terrain
+            for(const index in this.style._layers) {
```
HarelM, line 1701: Can you add a test for this scenario?

tempranova: Sure. Is there a best place for me to add that? When I was testing myself I added a new HTML to the "debug-pages", but I'm not 100% sure if that's the right place for me to add a test?

HarelM: Add a unit test to the map.test.ts file.

tempranova: Hey there, think I've added a sufficient test now, let me know!

src/ui/map.test.ts
```diff
@@ -2622,6 +2622,35 @@ describe('Map', () => {
...
+          map.on('load', () => {
+            map.addSource('terrainrgb', {type: 'raster-dem', url: '/source.json'});
+            server.respond();
+            map.addLayer({id: 'hillshade', type: 'hillshade', source: 'terrainrgb'});
+            jest.spyOn(console, 'warn').mockImplementation(() => { });
+            map.setTerrain({
+                source: 'terrainrgb'
+            });
+            expect(console.warn).toHaveBeenCalledWith(
+              'You are using the same source for a hillshade layer and for 3D terrain. Please consider using two separate sources to improve rendering quality.'
```
HarelM, line 2648: I don't see a good reason to check the text, just that it was called. If the text changes I don't want the test to fall. Other than that looks great, thanks!

tempranova: OK, updated. I did move the whole test up as I was getting some unexpected errors with it being the last test in the file (really odd, may look deeper into that later). In any case, it's updated now to check for a single console warning regardless of the text!

HarelM (approved):

HarelM:
This looks good, once CI is green I'll merge this.

tempranova:
Sorry bout that, those linting errors should be fixed now.

PR #1022, opened 2022-02-24
Terrain 3D

wipfli:
@prozessor13 thank you for all the work you did to make 3D happen in MapLibre GL JS. 

This pull request is a continuation of #165 with the goal to merge the `terrain3d` branch into `main`.

[Demo](https://wipfli.github.io/maplibre-terrain3d/#11.49/46.4444/8.0585/10.4/66)

[flyTo example](https://wipfli.github.io/maplibre-terrain3d/flyto.html)

[](https://wipfli.github.io/maplibre-terrain3d/#11.49/46.4444/8.0585/10.4/66)

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!

PR #2304, opened 2023-03-24
avoid extra loops when terrain does not exist

zhangyiatmicrosoft:
## Launch Checklist

placeSymbol is a very hot code path if not the hottest. For most screen sizes, this function is called 1000 to 2000 times. 
When terrain does not exist, it is still doing a loop on 4-element array ['textBox', 'verticalTextBox', 'iconBox', 'verticalIconBox'] which means 4000 to 8000 extra loops.

Please review --- before my change, textBox had elevation of 0, and now it would be undefined. 

However, I cannot find anywhere the elevation of textBox is used for placement or collision. The previous code was added in #1022, @HarelM  @prozessor13, can you share more insights regarding where and how it is used?
There is no UT coverage either.

### DIFF
--- a/src/symbol/placement.ts
+++ b/src/symbol/placement.ts
@@ -496,16 +496,6 @@ export class Placement {
                 verticalTextFeatureIndex = collisionArrays.verticalTextFeatureIndex;
             }
 
-            // update elevation of collisionArrays
-            if (getElevation) {
-                for (const boxType of ['textBox', 'verticalTextBox', 'iconBox', 'verticalIconBox']) {
-                    const box = collisionArrays[boxType];
-                    if (box) {
-                        box.elevation = getElevation(box.anchorPointX, box.anchorPointY);
-                    }
-                }
-            }
-
             const textBox = collisionArrays.textBox;
             if (textBox) {
 
--- a/src/symbol/placement.ts
+++ b/src/symbol/placement.ts
@@ -460,6 +460,9 @@ export class Placement {
             bucket.deserializeCollisionBoxes(collisionBoxArray);
         }
 
+        const tileID = this.retainedQueryData[bucket.bucketInstanceId].tileID;
+        const getElevation = this.terrain ? (x: number, y: number) => this.terrain.getElevation(tileID, x, y) : null;
+
         const placeSymbol = (symbolInstance: SymbolInstance, collisionArrays: CollisionArrays) => {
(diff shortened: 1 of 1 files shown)


src/symbol/placement.ts
```diff
@@ -460,6 +460,9 @@ export class Placement {
             bucket.deserializeCollisionBoxes(collisionBoxArray);
         }
 
+        const tileID = this.retainedQueryData[bucket.bucketInstanceId].tileID;
+        const getElevation = this.terrain ? (x: number, y: number) => this.terrain.getElevation(tileID, x, y) : null;
```
IvanSanchez, line 464: Consider replacing `null` with a function that always returns zero, i.e. `()=>0`

zhangyiatmicrosoft: It would be even slower than the previous version because 1) still 4000 to 8000 loops, 2) calling a function instead of directly assigning 0.

HarelM:
I don't know what this was for unfortunately, maybe @prozessor13 remembers.
In any case, a single unit test to cover this scenario would be helpful to cover this case and allow merging.
I know there are missing unit tests around this area, which was related to the effort around 3D terrain, but I still think this is needed.

prozessor13:
After a quick look into the code, because i cannot remember, i think the whole `getElevtaion` logic at this place is no longer needed. I think it is a leftover from the days, where all elevation is calculated in CPU not in shader.
Originally i think it was needed to place the collision-box at correct height, today the `get_elevation` shader function is used for this.

Regards.

HarelM:
Thanks for the quick response @prozessor13!
Do we have a way to validate this assumption? If we remove this code, what do we need to check to make sure it's not needed?

HarelM:
@zhangyiatmicrosoft I think the best approach is to do what @prozessor13 suggested.
There's a debug page with terrain that I think you can use to try out things, render-test should also help us avoid breaking things as @prozessor13 mentioned.

zhangyiatmicrosoft:
> @zhangyiatmicrosoft I think the best approach is to do what @prozessor13 suggested. There's a debug page with terrain that I think you can use to try out things, render-test should also help us avoid breaking things as @prozessor13 mentioned.

Thank you all. Deleted the block.
getElevation function is still needed later but it does not depend on placeSymbol function so I moved it outside to have better performance.

HarelM (approved):

PR #2306, opened 2023-03-27
Lazy serialization of layers

zhangyiatmicrosoft:
## Launch Checklist

this._serializedLayers dictionary is only needed when the style itself is serialized OR queryRenderedFeatures is called.

Currently it is populated upon creation, which is an overhead for initial map load. This PR changes to lazy serialization. 
Based on our experiment, for 150 layers this change saves ~20ms at 75 percentile, and ~100ms at 90 percentile.

Additional changes:

- "_serializeLayers" is renamed to "_serializeByIds" and piggyback on this._serializedLayers dictionary for better performance. Previously it was doing fresh serialization each time.
- "setState" method is serializing current instance twice. Fixed.
- replace forEach with simple for/of loop in "setState". Better performance, easier to debug
- creating layers is the most time-consuming part of _load method (~80% of time), so split out to be its own method for easier profiling in Chrome performance tool.

### DIFF
--- a/src/style/style.test.ts
+++ b/src/style/style.test.ts
@@ -588,8 +588,8 @@ describe('Style#_load', () => {
             }]
         });
 
-        style._load(styleSpec, {validate: true});
-        expect(style._serializedLayers).toBeUndefined();
+        style._load(styleSpec, {validate: false});
+        expect(style._serializedLayers).toBeNull();
     });
 
     test('serialized layers should be corrected updated after adding/removing layers', done => {
@@ -631,7 +631,7 @@ describe('Style#_load', () => {
 
         map.on('style.load', () => {
             let serializedStyle = map.style.serialize();
-
+            expect(serializedStyle.layers).toHaveLength(1);
             expect(serializedStyle.layers[0].id).toBe('layerId0');
 
             const layer = {
@@ -642,6 +642,7 @@ describe('Style#_load', () => {
 
             // serialize again
             serializedStyle = map.style.serialize();
+            expect(serializedStyle.layers).toHaveLength(2);
             expect(serializedStyle.layers[1].id).toBe('background');
 
             // remove and serialize
--- a/src/style/style.test.ts
+++ b/src/style/style.test.ts
@@ -647,7 +647,7 @@ describe('Style#_load', () => {
             // remove and serialize
             map.removeLayer('background');
             serializedStyle = map.style.serialize();
-            expect(serializedStyle.layers.length).toBe(1);
+            expect(serializedStyle.layers).toHaveLength(1);
 
             done();
         });
--- a/src/style/style.test.ts
+++ b/src/style/style.test.ts
@@ -16,9 +16,14 @@ import {OverscaledTileID} from '../source/tile_id';
 import {fakeXhr, fakeServer} from 'nise';
 
 import EvaluationParameters from './evaluation_parameters';
-import {LayerSpecification, GeoJSONSourceSpecification, FilterSpecification, SourceSpecification} from '@maplibre/maplibre-gl-style-spec';
+import {LayerSpecification,
+    GeoJSONSourceSpecification,
+    FilterSpecification,
+    SourceSpecification,
+    StyleSpecification} from '@maplibre/maplibre-gl-style-spec';
 import {SourceClass} from '../source/source';
 import GeoJSONSource from '../source/geojson_source';
+import {createMap, beforeMapTest} from '../util/test/util';
 
 function createStyleJSON(properties?) {
     return extend({
@@ -585,6 +590,67 @@ describe('Style#_load', () => {
 
         style._load(styleSpec, {validate: true});
         expect(style._serializedLayers).toBeUndefined();
+    });
+
+    test('serialized layers should be corrected updated after adding/removing layers', done => {
+        beforeMapTest();
+        const map = createMap({deleteStyle: true});
+        const testStyle: StyleSpecification =   {
+            version: 8,
+            center: [-73.9749, 40.7736],
+            zoom: 12.5,
+            bearing: 29,
+            pitch: 50,
+            sources: {},
+            layers: []
+        };
+        map.setStyle(testStyle, {
+            diff: true,
+            transformStyle: (prevStyle, nextStyle) => {
+                expect(prevStyle).toBeUndefined();
+
+                return {
+                    ...nextStyle,
+                    sources: {
+                        maplibre: {
+                            type: 'vector',
+                            minzoom: 1,
+                            maxzoom: 10,
+                            tiles: ['http://example.com/{z}/{x}/{y}.png']
+                        }
+                    },
+                    layers: [{
+                        id: 'layerId0',
+                        type: 'circle',
+                        source: 'maplibre',
+                        'source-layer': 'sourceLayer'
+                    }]
+                };
+            }
+        });
+
+        map.on('style.load', () => {
+            let serializedStyle = map.style.serialize();
+
+            expect(serializedStyle.layers[0].id).toBe('layerId0');
+
+            const layer = {
+                id: 'background',
+                type: 'background'
+            } as LayerSpecification;
+            map.addLayer(layer);
+
+            // serialize again
+            serializedStyle = map.style.serialize();
+            expect(serializedStyle.layers[1].id).toBe('background');
+
+            // remove and serialize
+            map.removeLayer('background');
+            serializedStyle = map.style.serialize();
+            expect(serializedStyle.layers.length).toBe(1);
+
+            done();
+        });
     });
 });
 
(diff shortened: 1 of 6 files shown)


test/build/min.test.ts
```diff
@@ -39,7 +39,7 @@ describe('test min build', () => {
         const decreaseQuota = 4096;
 
         // feel free to update this value after you've checked that it has changed on purpose :-)
-        const expectedBytes = 756547;
```
zhangyiatmicrosoft, line 42: before this PR, the build size was already 757351.

src/style/style.ts
```diff
@@ -441,12 +445,44 @@ class Style extends Evented {
...
-        for (const id of ids) {
-            const layer = this._layers[id];
-            if (layer.type !== 'custom') {
-                serializedLayers.push(layer.serialize());
+    /**
+     * take an array of string IDs, and based on this._layers, generate an array of LayerSpecification
+     * @param ids an array of string IDs, for which serilized layers will be generated. If ommited, all serialized layers will be returned
+     * @returns {Array<LayerSpecification>} generated result
+     */
+    private _serializeById(ids?: Array<string>): Array<LayerSpecification> {
```
HarelM, line 453: _serializeByIds to indicate it's not for a single layer?

src/style/style.ts
```diff
@@ -441,12 +445,44 @@ class Style extends Evented {
...
+        return serializedLayers;
+    }
+
+    /**
+     * Lazy initialization of this._serializedLayers dictionary and return it
+     * @returns this._serializedLayers dictionary
+     */
+    private _getSerializedLayers(): {[_: string]: LayerSpecification} {
+        let serializedLayers = this._serializedLayers;
+        if (!serializedLayers) {
```
HarelM, line 478: Consider inverting this if to reduce indentation complexity.

src/style/style.test.ts
```diff
@@ -570,6 +570,22 @@ describe('Style#_load', () => {
         // cleanup
         stub.mockReset();
     });
+
+    test('layers are NOT serialized immediately after creation', () => {
```
HarelM, line 574: This test is great and important, but I would expect to see more tests here to cover the code changes that were made, for example, the "invalidation" of the layer serialization, update of data and check that the new serialization is up-to-date etc.

src/style/style.ts
```diff
@@ -164,7 +164,7 @@ class Style extends Evented {
     _request: Cancelable;
     _spriteRequest: Cancelable;
     _layers: {[_: string]: StyleLayer};
-    _serializedLayers: {[_: string]: any};
+    _serializedLayers: {[_: string]: LayerSpecification};
```
HarelM, line 167: Great!!! thanks!

src/style/style.ts
```diff
@@ -206,8 +206,8 @@ class Style extends Evented {
 
         this._spritesImagesIds = {};
         this._layers = {};
-        this._serializedLayers = {};
-        this._order  = [];
+
+        this._order = [];
```
HarelM, line 210: initialization to `null` would be a good practice here I think.

HarelM: ahh, it is done in the below method. might be better to have it here though, IDK...

src/style/style.ts
```diff
@@ -441,12 +445,44 @@ class Style extends Evented {
...
+    /**
+     * take an array of string IDs, and based on this._layers, generate an array of LayerSpecification
+     * @param ids an array of string IDs, for which serilized layers will be generated. If ommited, all serialized layers will be returned
+     * @returns {Array<LayerSpecification>} generated result
+     */
+    private _serializeById(ids?: Array<string>): Array<LayerSpecification> {
+
+        const serializedLayersDictionary = this._getSerializedLayers();
+        let serializedLayers: LayerSpecification[];
+        if (ids && ids.length > 0) {
```
HarelM, line 457: consider inverting the if and returning `Object.values(serializedLayersDictionary)` in order to reduce indentation and complexity.

src/style/style.ts
HarelM: The method name is a bit confusing as when reading it you don't see any method that is called serialize so it's hard to understand where serialization is done....

src/style/style.ts
```diff
@@ -441,12 +445,44 @@ class Style extends Evented {
...
+            serializedLayers  = Object.values(serializedLayersDictionary);
+        }
+        return serializedLayers;
+    }
+
+    /**
+     * Lazy initialization of this._serializedLayers dictionary and return it
+     * @returns this._serializedLayers dictionary
+     */
+    private _getSerializedLayers(): {[_: string]: LayerSpecification} {
```
HarelM, line 476: See my comment above about the method name, it somewhat applied here as well.

src/style/style.ts
```diff
@@ -894,7 +930,11 @@ class Style extends Evented {
         this._changed = true;
         this._removedLayers[id] = layer;
         delete this._layers[id];
-        delete this._serializedLayers[id];
+
+        const serializedLayers = this._serializedLayers;
```
HarelM, line 934: I would consider removing this extra variable, I don't see it helps besides removing 6 characters twice.

src/style/style.ts
```diff
@@ -1125,9 +1169,9 @@ class Style extends Evented {
             sprite: this.stylesheet.sprite,
             glyphs: this.stylesheet.glyphs,
             transition: this.stylesheet.transition,
-            sources: mapObject(this.sourceCaches, (source) => source.serialize()),
-            layers: this._serializeLayers(this._order)
-        }, (value) => { return value !== undefined; });
+            sources,
+            layers},
```
HarelM, line 1173: move the `},` to the next line

src/style/style.ts
```diff
@@ -1138,6 +1182,10 @@ class Style extends Evented {
             this._updatedSources[layer.source] = 'reload';
             this.sourceCaches[layer.source].pause();
         }
+
+        // upon updating, serilized layer dictionary should be reset.
+        // When needed, it will be populated with the correct copy again.
+        this._serializedLayers = null;
```
HarelM, line 1188: Add a test for this scenario and a scenario for init -> update -> serialize or something like that.

HarelM:
I've added a few comments.

src/style/style.ts
zhangyiatmicrosoft: _createLayers() will be called many times, so it has to set this._serializedLayers to null.
Adding another init statement here "feels" more complete but it is actually a waste. :)

src/style/style.ts
zhangyiatmicrosoft: renamed this one to _serializedAllLayers

src/style/style.ts
zhangyiatmicrosoft: added "s", since the other method is renamed, hope everything is easier to read.

src/style/style.ts
zhangyiatmicrosoft: thank you. Added

src/style/style.test.ts
zhangyiatmicrosoft: done

src/style/style.ts
```diff
@@ -1113,21 +1154,27 @@ class Style extends Evented {
     }
 
     serialize(): StyleSpecification {
+
+        const sources = mapObject(this.sourceCaches, (source) => source.serialize());
+        const layers = this._serializeByIds(this._order);
+        const myStyleSheet = this.stylesheet;
```
HarelM, line 1160: Same here, although less important as there is no change to the state of the class.

src/style/style.ts
```diff
@@ -441,14 +445,48 @@ class Style extends Evented {
...
+
+        return serializedLayers;
+    }
+
+    /**
+     * Lazy initialization of this._serializedLayers dictionary and return it
+     * @returns this._serializedLayers dictionary
+     */
+    private _serializedAllLayers(): {[_: string]: LayerSpecification} {
+        let serializedLayers = this._serializedLayers;
```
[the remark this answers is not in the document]
zhangyiatmicrosoft: this.foo is never minimized, so to reduce runtime footprint as much as possible, it is a good practice to assign frequently used "this.foo" to a local var. In addition accessing local variable is a slightly faster.
Of course, everything is in terms of microseconds, but I take what I can get ...

src/style/style.ts
zhangyiatmicrosoft: In this example, this.stylesheet is repeated 11 times, so we are talking about saving 110 bytes and 11 accesses to "this" object:
https://stackoverflow.com/questions/2936807/javascript-object-access-performance

HarelM: Thanks for this info! I did not know that.
Isn't there a typescript flag we can use to do this automatically?

HarelM (approved):

PR #2324, opened 2023-04-01
 Readability and code clean up in loadSprite

zhangyiatmicrosoft:
## Launch Checklist

Fixes and improvements:
- "sprite" is an array, let's name it correctly.
- A global "error" variable is redundant. Just directly "callback(err)" when detected -- this is what it was doing anyway
- Long lines with multiple function calls are not only hard to read, but also difficult to single step debug/profiling and set break point.
- Two request array (jsonRequests and imageRequests) are combined and just use map to avoid indexing over and over again. They have the same book-keeping purpose.
- Combine two callbacks as much as possible

Looks like existing test cases covered everything and no new ones needed.


src/style/load_sprite.ts
```diff
@@ -17,45 +17,46 @@ export default function loadSprite(
...
 ): Cancelable {
-    const sprite = coerceSpriteToArray(originalSprite);
+    const spriteArray = coerceSpriteToArray(originalSprite);
     const format = pixelRatio > 1 ? '@2x' : '';
 
-    let error;
-    const jsonRequests: Cancelable[] = [];
-    const imageRequests: Cancelable[] = [];
+    const combinedRequestsMap: {[requestKey: string]: Cancelable} = {};
+    const jsonsMap: {[id: string]: any} = {};
```
HarelM, line 24: Can these variable be removed or not used in the maybecomplete method?
I think this code changes are great, but I would like to use the `combinedRequestsMap` more in order to in be able to know the current state.

src/style/load_sprite.ts
```diff
@@ -17,45 +17,46 @@ export default function loadSprite(
...
 
-    function maybeComplete() {
+    /**
+     * @param id - id of the sprite whose callback has just been received
+     * @param requestKey - id_url (JSON or image) as the key of the network request
+     * @param dataMap - dataMap object (either jsonsMap or imagesMap dictionary)
+     * @param err - error object
+     * @param data - data object returned by JSON or image request
+     */
+    function maybeComplete(id: string, requestKey: string, dataMap:{[id: string]: any}, err: Error, data: any): void {
```
HarelM, line 48: Would be nice to have this method outside the `loadSprite` so that it doesn't inherit the state of the method and its variables.

src/style/load_sprite.ts
zhangyiatmicrosoft: Did not fully understand which variable you were asking? Current vars and their purposes:
combinedRequestsMap -- bookkeeping so they can be cancelled 
jsonsMap & imagesMap - data map for JSON and images respectively. They are used later to construct the result.

HarelM: See my next comment, fixing it would probably makes things more readable.

src/style/load_sprite.ts
```diff
@@ -28,59 +29,76 @@ export default function loadSprite(
...
+    resourceType: ResourceType,
+    jsonsMap:{[id: string]: any},
+    imagesMap:{[id: string]: (HTMLImageElement | ImageBitmap)},
+    err: Error,
+    data: any,
+    expectedResultCounter: number): void {
 
-        if (spriteArray.length === jsonsLength && jsonsLength === imagesLength) {
-            const result = {} as {[spriteName: string]: {[id: string]: StyleImage}};
+    if (err) {
```
HarelM, line 73: I think this part up until line 81 should be outside this method, much like what you did with the `delete combinedRequestsMap[...]`

src/style/load_sprite.ts
```diff
@@ -28,59 +29,76 @@ export default function loadSprite(
...
+    }
 
-            for (const spriteName in jsonsMap) {
-                result[spriteName] = {};
+    const jsonsLength = Object.values(jsonsMap).length;
+    const imagesLength = Object.values(imagesMap).length;
 
-                const context = browser.getImageCanvasContext(imagesMap[spriteName]);
-                const json = jsonsMap[spriteName];
+    if (expectedResultCounter === jsonsLength && jsonsLength === imagesLength) {
```
HarelM, line 86: Inverting this `if` would help a lot in readability and reduce complexity.
Having only this part of the code in this method will allow you to change its name to something like
`doOnceCompleted(...)` or `parseImagesWhenAllRequestsAreCompleted(...)` which is a lot better than `maybeComplete`, IMHO.

HarelM:
I've added a few more minor comments. I think this looks a lot better now, THANKS!

src/style/load_sprite.ts
```diff
@@ -17,79 +17,78 @@ export default function loadSprite(
...
-    let error;
-    const jsonRequests: Cancelable[] = [];
-    const imageRequests: Cancelable[] = [];
+    const combinedRequestsMap: {[requestKey: string]: Cancelable} = {};
+    const jsonsMap: {[id: string]: any} = {};
+    const imagesMap: {[id: string]: (HTMLImageElement | ImageBitmap)} = {};
 
-    const jsonsMap: {[baseURL: string]: any} = {};
-    const imagesMap: {[baseURL:string]: (HTMLImageElement | ImageBitmap)} = {};
+    for (const {id, url} of spriteArray) {
```
HarelM, line 28: What will happen if the user sets an "invalid" sprite array that has duplicate entries (id+url are not unique)?

HarelM (approved):
Overall looks great! Let me know what the outcome of the scenario I mentioned above would be.

src/style/load_sprite.ts
zhangyiatmicrosoft: Thanks for keen observation. It would be buggy code to begin with, however for the sake of completeness, I updated coerceSpriteToArray to dedup and cover this corner case.
.

HarelM: Cool, thanks.

PR #1586, opened 2022-08-30
Handle empty 20x raster tile response as empty tile.

cns-solutions-admin:
getImage (ajax.ts) now returns a null image, if there is no content (ArrayBuffer with length 0).
Raster source loadTile will store such an image as a tile with the new status 'empty' (to avoid loading it again, if there are cache headers).

See also #1579.

PR #2325, opened 2023-04-01
Better raster 204 handling

ibesora:
## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/util/image_request.ts
+++ b/src/util/image_request.ts
@@ -139,7 +139,7 @@ namespace ImageRequest {
         }
     };
 
-    const doArrayRequest = (itemInQueue: ImageRequestQueueItem): Cancelable => {
+    export const doArrayRequest = (itemInQueue: ImageRequestQueueItem): Cancelable => {
 
         const {requestParameters, callback} = itemInQueue;
 
--- a/src/source/raster_tile_source.test.ts
+++ b/src/source/raster_tile_source.test.ts
@@ -133,6 +133,36 @@ describe('RasterTileSource', () => {
         server.respond();
     });
 
+    test('handles no image data', done => {
+        server.respondWith('/source.json', JSON.stringify({
+            minzoom: 0,
+            maxzoom: 22,
+            attribution: 'MapLibre',
+            tiles: ['http://example.com/{z}/{x}/{y}.png'],
+            bounds: [-47, -7, -45, -5]
+        }));
+        server.respondWith('http://example.com/10/5/5.png', [204, {}, '']);
+
+        const source = createSource({url: '/source.json'});
+        source.on('data', (e) => {
+            if (e.sourceDataType === 'metadata') {
+                const tile = {
+                    tileID: new OverscaledTileID(10, 0, 10, 5, 5),
+                    state: 'loading',
+                    loadVectorData () {},
+                    setExpiryData() {}
+                } as any as Tile;
+                source.loadTile(tile, (err) => {
+                    expect(err).toBeNull();
+                    expect(tile.state).toBe('errored');
+                    done();
+                });
+                server.respond();
+            }
+        });
+        server.respond();
+    });
+
     test('cancels TileJSON request if removed', () => {
         const source = createSource({url: '/source.json'});
         source.onRemove();
--- a/src/util/image_request.test.ts
+++ b/src/util/image_request.test.ts
@@ -210,6 +210,54 @@ describe('ImageRequest', () => {
         server.respond();
     });
 
+    test('doArrayRequest handles request errors', done => {
+        server.respondWith(request => {
+            request.respond(404, undefined, '404 Not Found');
+        });
+
+        ImageRequest.doArrayRequest(
+            {
+                requestParameters: {url: 'http://example.com/test.json'},
+                cancelled: false,
+                completed: true,
+                cancel: () => {},
+                callback: async (error) => {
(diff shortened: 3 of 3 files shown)


src/util/image_request.ts
```diff
@@ -139,7 +139,7 @@ namespace ImageRequest {
         }
     };
 
-    const doArrayRequest = (itemInQueue: ImageRequestQueueItem): Cancelable => {
+    export const doArrayRequest = (itemInQueue: ImageRequestQueueItem): Cancelable => {
```
zhangyiatmicrosoft, line 142: It is not desirable to export a function just for unit testing.
Should be simple enough to just call the original entry point and let it flow here.

zhangyiatmicrosoft (changes requested):

src/util/image_request.ts
ibesora: Done!

src/source/raster_tile_source.ts
```diff
@@ -111,7 +111,8 @@ class RasterTileSource extends Evented implements Source {
             if (tile.aborted) {
                 tile.state = 'unloaded';
                 callback(null);
-            } else if (err) {
+            } else if (err || !img) {
+                // When an HTTP 204 response is received, there's no error message and no image data
                 tile.state = 'errored';
                 callback(err);
             } else if (img) {
```
zhangyiatmicrosoft, line 118: Please double check here.. !img is already checked at line 114, so no need to do it again?
No harm either just a little bit of verbose.

ibesora: You are right, I have updated this and fixed some failing tests.

xabbu42:
How does this relate to #1586?

src/util/image_request.ts
```diff
@@ -157,6 +157,11 @@ namespace ImageRequest {
                     }
                 };
                 arrayBufferToCanvasImageSource(data, decoratedCallback);
+            } else {
+                // HTTP 204 responses don't error and do not contain any data
+                // Calling the callback without any of them will
```
pramilk, line 162: This should be treated as an error. I think condition at line 149 and this one can be combined. In this case, 'err' will be null.

ibesora: > This should be treated as an error.

You mean that a 204 should rise an error as HTTPImageElement does?

pramilk: It should be treated as an error (not necessarily raise an error). Please see PR: https://github.com/maplibre/maplibre-gl-js/pull/2126. There will be some conflicts between these 2 PR's.  It will be better that irrespective of the underlying implementation to fetch the image, this function should treat them in the same way.

ibesora: HTMLImageElement will raise an error on 204s so, in order to mimic the same behaviour, I updated this to create an error and then to ignore it on source_cache. This should make the behaviour consistent when #2126 is merged

ibesora:
Can I get a rereview of this? I've addressed all the comments and it should now pass all the tests

src/source/source_cache.ts
```diff
@@ -265,9 +265,14 @@ class SourceCache extends Evented {
     _tileLoaded(tile: Tile, id: string, previousState: TileState, err?: Error | null) {
         if (err) {
             tile.state = 'errored';
-            if ((err as any).status !== 404) this._source.fire(new ErrorEvent(err, {tile}));
-            // continue to try loading parent/children tiles if a tile doesn't exist (404)
-            else this.update(this.transform, this.terrain);
+            if ((err as any).status !== 404 && (err as any).status !== 204) {
```
HarelM, line 268: This doesn't feel right - I know this was here before (probably from "before time"), but I would love if this was moved to the ajax layer and have here a specific error maybe, instead of relaying on http error codes...
Basically the `(err as any)` is a smell that I would like to remove if we are touching this code...

HarelM:
I've seen @pramilk comments so I guess he has looked into this, but just for completeness, the following PR is touching similar code I believe: #2126.
@zhangyiatmicrosoft can you elaborate of the request for changes or should this be dismissed?

pramilk:
Both these PR are touching same code path, but are independent. This can go through.

src/source/source_cache.ts
ibesora: I agree that that `(err as any)` looks really wrong. Since this is executed by all source types, we'll need to update all sources that implement `loadTile`.
Maybe I can work on this on another PR? It seems doing it here will greatly increase the scope of this PR and dillute the goal of it. Just double checking before doing extra work.

HarelM: The current scope isn't that big. If we could better define a ErrorWithStatus or something similar and throw that instead of a regular Erroro with this cast to any it would be much better, 
In a different PR we can replace these http status with something more meaningful, like an enum or string-enum...

HarelM:
@ibesora what's the status of this PR?

ibesora:
> @ibesora what's the status of this PR?

Sorry, I haven't been able to take another look at it

HarelM:
Note that there's a new PR related to this issue.
Also, I have changed the code dramatically to improve readability so hopefully this will be easier to solve now:
- #3428

HarelM:
Closing this in favor of the merged PR: #3428
If this is still relevant please open a new PR or ask me to reopen this one (although I think the conflicts will not be easy to solve).

PR #2354, opened 2023-04-06
Improve control performance by restricting worker count to a max of 1 except Safari

pramilk:
## Launch Checklist

Currently Worker Count is capped at 6. On Bing Maps A/B testing showed best results with 1 workerCount and thats what we in production.  Capping at 2 or 3 showed similar results. WorkerCount 1 was statistically 2% faster then 2 or 3 in our experiments. 

In this PR I am caping it to 2 instead of current 6.

 - [ ] Link to related issues.
 - [*] Document any changes to public APIs.
 - [ ] Post benchmark scores.


HarelM:
Can you share the performance results somehow?

pramilk:
> Can you share the performance results somehow?

Numbers that I have from our A/B testing is comparing 1 vs 2 vs 3 max workers and are about 1 year old. Let me run the tests again. Will take 1-2 weeks for me to have the results. Will reply back with the data.

HarelM:
Can you think of a reason why 2 would perform better than 3?

pramilk:
> Can you think of a reason why 2 would perform better than 3?

There are 2 aspects to performance
1. Initial loading performance. In our measurements its from nagivation start till first idle event (We call it page load time- PLT).
2. Interactivity performance. Zoom/Pan FPS, Zoom/pan load time (Time between zoom/pan stop and Subsequent idle event)  etc.

For #1, more the number of workers slower it gets. In our old experiments WorkerCount=1 was the fastest in this metrics (2+%). Reason I think is due to the time it takes to initialize all the workers and the fact that style has to be parsed on all of these workers. Style parsing is one of the most expensive operation.

For #2, FPS was almost same between 1 vs 2 vs 3 workers. But for Zoom load time, 3 Worker was the fastest (By 3.5% as compared to 2 worker) and slowest was 1 worker ( By 2.3% as compared to 2 worker). Pan load time had no impact.

All this data is from June 2022 A/B tests on Bing.com/maps. We finally decided to go with 1 worker as for us #1 was more important then Zoom load time. 

Anyway we will run a new A/B test next week to get fresh data. Will be comparing current Maplibre current default vs 2 vs 1 worker.

HarelM:
Thanks for the detailed response! This makes sense now.

pramilk:
Here are the results from the A/B test that I run
Notes about metrics:
1. Page Load Time - Time from Navigation start till first maps idle event
2. Zoom FPS - FPS during zoom. Measured by measuring number of move events between movestart and moveend. Any move <100ms in duration (between movestart and moveend) is ignored.
3. % Zoom Frame Drop - During Zoom (between movestart and moveend end), % of the total frames which took more then 16ms.
4. 3 Worker Count in the results means Math.min(3, navigator.hardwareConcurrency / 2). So its a max of 3 worker Count.
5. Default worker Count mean, maplibre current default which is Math.max(Math.min(navigator.hardwareConcurrency / 2, 6), 1)

Notes about results:
1. Statistical confidence was much higher for Chromium results then for Firefox and Safari browers likely due to the sample size
2. Pan showed no impact in any browser, so I have not added it in results

Conclusions:
1. Chromium and Firefox was clearly fastest with 1 worker count
2. Safari showed better results with max 3 worker count

Let me know, if you need any additional info. I think we can configure Worker Count by browser specifically for Safari.

image

HarelM (approved):

HarelM:
Thanks for the research. Just to make sure, this was done on vector tiles rendered on the client, right (not raster tiles or SSR vector tiled)?
@wipfli @birkskyum please share your input, I generally think this is a good change.

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 - Remove a deprecation warning by removing an empty texture that is no longer being used in the codebase ([#2299](https://github.com/maplibre/maplibre-gl-js/pull/2299))
 - Improve initial loading performance by lazy serializing layers only when needed. ([#2306](https://github.com/maplibre/maplibre-gl-js/pull/2306))
 - [Breaking] Cancel unloaded tile request on zooming in across multiple zoom. Previously these requests were not cancelled. ([#2377](https://github.com/maplibre/maplibre-gl-js/pull/2377))
+- Improve control performance by restricting worker count to a max of 1 except safari browser. ([#2354](https://github.com/maplibre/maplibre-gl-js/pull/2354))
```
HarelM, line 23: This is a breaking change I believe.

pramilk:
> Thanks for the research. Just to make sure, this was done on vector tiles rendered on the client, right (not raster tiles or SSR vector tiled)? @wipfli @birkskyum please share your input, I generally think this is a good change.

- On our site 95+ % of users come to road map style which is vector. Very few uses aerial which is totally raster.
- We detect users based on there GPU capability and break them in to 2 experience

1. Full vector: In this both the base tiles (Road/Parks/Land/water etc) and labels are MVT vector tiles. This constitute about 60% of our users
2. Partial Vector: In this for base tiles (Road/Parks/Land/water etc) we use raster tiles and labels is still rendered using MVT vector tiles. This constitutes remaining 40%. 

Numbers that I shared combines all of it. Not sure what is SSR vector tiles.

CHANGELOG.md
pramilk: Not sure what is classified as breaking. In this case there is no change in control rendering nor is there any change in public API. Let me know and I am fine adding it.

HarelM:
Server side rendering if the tiles.
I would be interested to concentrate the result of fully vector tiles scenario and understand if the performance of zoom-in-out is degraded, and if so by how much as this would be the downside of reducing workers I assume.

CHANGELOG.md
HarelM:
```suggestion
- [Breaking] Improve control performance by restricting worker count to a max of 1 except safari browser. ([#2354](https://github.com/maplibre/maplibre-gl-js/pull/2354))
```

pramilk:
> Server side rendering if the tiles. I would be interested to concentrate the result of fully vector tiles scenario and understand if the performance of zoom-in-out is degraded, and if so by how much as this would be the downside of reducing workers I assume.

Sure. I can provide more data for only full Vector. Will add more details about other zoom metrics that we capture. Will provide it by tomorrow.

HarelM:
Thanks!!

birkskyum:
Can we reasonably assume this to generalize to other hardware, like phones?

wipfli:
Looks like a good idea. How do you characterize slow vs fast gpu, @pramik?

wipfli:
Also, you are doing great work to improve performance, @pramik. Can we document somewhere the overall improvements between successive versions? Would be cool to be able to say at the end of the year that we made page load 10 percent faster or so...

pramilk:
> Server side rendering of the tiles. I would be interested to concentrate the result of fully vector tiles scenario and understand if the performance of zoom-in-out is degraded, and if so by how much as this would be the downside of reducing workers I assume.
How to interpret these results
1. These results are only for Full vector (Fast GPU)
2. This is for all browsers combined. Don't have per browser breakdown when looking at just Full Vector numbers.
3. Currently we have maplibre control on just PC and tablet (iPad/Android etc). So no mobile users in these results.
4. Control in each Column is workerCount =1 (Which is currently default on Bing)
5. Column 2 - Treatment (Worker Count 2) vs Control
6. Column 3 - Treatment (Worker Count 3) vs Control
7. Column 4 - Treatment (Maplibre default) vs Control
8. P-Move value shows the statistical confidence in the results. Higher the number more confidence it has in the results. Its shown with darker red/green color.
9. Metrics definition
  a. Page Load Time (PLT) - Time from Navigation start till first maps idle event.
  b. Zoom/Pan Time to glass (TTG) - This is the time taken from moveend till next idle event.
  c. Zoom/Pan FPS - FPS during zoom. Measured by measuring number of move events between movestart and moveend. Any move <100ms in duration (between movestart and moveend) is ignored.
  d. % Zoom/Pan Frame Drop - During Zoom (between movestart and moveend end), % of the total frames which took more then 16ms.
  
image

> Are these users a mix of phones and desktop hardware?
@birkskyum, we are still in the process of migrating over fully to maplibre. Our Mobile version currently does not use maplibre control. So these numbers are only for PC and tablet (iPad/Android etc).

> Looks like a good idea. How do you characterize slow vs fast gpu, @pramik?
@wipfli , We run a A/B test captured all users GPU agent (something like [this ](https://developer.mozilla.org/en-US/docs/web/api/webgl_debug_renderer_info)) along with CPU core and measured their performance. Based on this data classified users into 3 bucket (Fast/medium/slow). Fastest set of users are shown full vector experience.   Some of the obvious slow ones are "swiftshader", "software adapter", "llvmpipe" etc. Some of the faster ones are "uhd graphics", "iris(r) plus graphics", "iris(r) xe graphics", "vega 8 graphics" etc. In some cases (with relatively slower GPU) if user has > 8 CPU core we upgrade them to Fast.  This is all based on the data we captured.

> Also, you are doing great work to improve performance, @pramik. Can we document somewhere the overall improvements between successive versions? Would be cool to be able to say at the end of the year that we made page load 10 percent faster or so...
@wipfli. Thanks. There are tools available which can measure progress overtime. We internally do measure it for our products and have ongoing improvement targets.

HarelM:
This value can be changed by clients, so worst case someone can override this default.
The previous values are probably outdated in terms of relevance and I suspect were not updated recently.
The decision here is based on a good amount of data, which is the best way IMHO to take decision.
I generally think this should go in.

birkskyum (approved):

HandyMenny:
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/hardwareConcurrency) and [Can I Use](https://caniuse.com/hardwareconcurrency), Safari introduced support for navigator.hardwareconcurrency i safari 10.x and disabled it by default in safari 11, furthermore, safari on ios never returns more than 2 and 2/2 is 1. So the value set by maplibre for safari before and after this pull request is almost always 1. am I missing something? does it really make sense to have a different behaviour for such a rare case?

ChrisLoer:
I just noticed this change and I don't have a strong opinion on it, but I just wanted to point out a potential issue with optimizing against "zoom FPS" or "zoom dropped frames" here -- if you slow down tile loading, you will expect to increase FPS, because the tile load itself is frequently what causes dropped frames (b/c you need to upload all the tile buffers to the GPU). If you're zooming quickly enough, slower tile loads means more tiles will get cancelled before they load.

It's maybe a matter of perspective whether this is a win -- on the one hand, you have less jank, but on the other hand you're more likely to be showing blank tiles (they render really fast!) or heavily overzoomed tiles.

Where I expect the real cost of fewer web workers to show up is in "time to load many tiles", which happens during initial load (which you measured) but is obscured by all the other start up costs. It can also happen when you're adding new sources/layers to a map, or panning to an entirely new area. But I have to admit using a single worker seems to be performing fine! I only noticed because I was in the debugger and noticed that only one worker had launched.

HarelM:
@ChrisLoer can you replicate the results @pramilk shared? Can we identify the cause of this difference in performance?
Maybe we can document when should one use more workers in the docs.

HarelM:
It works be great to have a comment related to the worker count maplibre property saying something like:
When processing vector tiles with large amount of data it might be beneficial to increase this number.

Issue #2357, opened 2023-04-07
GeolocationControl._setupUI crashes when used with react-map-gl and mapStyle fails to load


bgeihsgt:
Please note that we don't have a consistent repro for this, but we have seen this crash occur in our production error logging.

**maplibre-gl-js version**: 2.4.0

**browser**: Safari/Mobile Safari

### Steps to Trigger Behavior

 1.  Use `react-map-gl` version `7.0.21`
 2.  In Safari, hit a case where mapStyle fails to fetch. In our case, we're using api.maptiler.com
 

### Link to Demonstration

We don't have a live demonstration because it requires a specific intermittent failure mode. But here is what we're seeing:

Our error log crashes with:

```
UnhandledPromiseRejectionWarning: TypeError: undefined is not an object (evaluating 'this._map._getUIString')
```

This is happening inside `_setupUI` within `GeolocateControl`. (Our code uses a structure like `<Map><GeolocateControl /></Map>` ). Each error log we see is accompanied by a failure to load the api.maptiler.com map style JSON.

In inspecting through the `react-map-gl` code, I noticed this code snippet:

https://github.com/visgl/react-map-gl/blob/v7.0.21/src/components/geolocate-control.ts#L80-L85

It looks like this is firing, but `this._map` is not defined due to the fetch failure.

### Expected Behavior

It doesn't crash in this failure case

### Actual Behavior

It crashes

HarelM:
Seems like the code there uses an internal method, which doesn't seem like a good idea.
I've also seen the following bug, which lacks tests and a PR in order to be resolved properly: #2294.
I think it's far easier to simply create a location control, but it might be just me :-)

bgeihsgt:
Thanks @HarelM. Agree that `react-map-gl` probably should not call internal methods like this :)

Note I attached a PR here that at least doesn't crash if `_setupUI` doesn't have a `_map` instance.

HarelM:
See my comments there - if the invocation is not done correctly, the fix should not be in this library...

bgeihsgt:
@HarelM fair enough. I'll file against `react-map-gl`

Pessimistress:
I just want to note here that react-map-gl does not call `_setupUI`. It simply wraps the method to work around a mapbox-gl bug *in case* it is called. I believe that #2358 would have resolved this issue.

HarelM:
Adding a null check without a test or a proper reproduction is a patch, not a fix.
[...]

PR #2358, opened 2023-04-07
Fix crash in GeolocateControl._setupUI

bgeihsgt:
Fixes https://github.com/maplibre/maplibre-gl-js/issues/2357

This obviously does not fix the root issue, but it at least should avoid a crash.

## Launch Checklist

 - [X] Link to related issues.

### DIFF
--- a/src/ui/control/geolocate_control.ts
+++ b/src/ui/control/geolocate_control.ts
@@ -371,6 +371,11 @@ class GeolocateControl extends Evented implements IControl {
     }
 
     _setupUI(supported: boolean) {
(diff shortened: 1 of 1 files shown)


HarelM:
Unfortunately, I don't think this is the right way to approach this.
There's need to be a test that represents this issue and an explanation on what can cause this.
If the cause is that this method is called from a wrapping library in an incorrect timing, I think the right location to fix this is in the wraping library.

bgeihsgt:
Agreed. Closing.

PR #2375, opened 2023-04-11
avoid creating dummy href element from document object

zhangyiatmicrosoft:
## Launch Checklist

Perf test:
https://jsbench.me/98lgbmptuk/1
image

 - [x] Post benchmark scores.


HarelM:
@zhangyiatmicrosoft is this waiting for something? Can we move this forward?

zhangyiatmicrosoft:
It is currently only used in getVideo so this change have very little benefit, and not exported for testing.
After #2126 it will be much more meaningful and also test friendly.

HarelM:
Not sure I follow... Please take out of draft if possible...

pramilk:
> It is currently only used in getVideo so this change have very little benefit, and not exported for testing. After #2126 it will be much more meaningful and also test friendly. Also string comparison is 2x faster than URL class.

@zhangyiatmicrosoft , https://github.com/maplibre/maplibre-gl-js/pull/2126 is now merged.

zhangyiatmicrosoft:
> It is currently only used in getVideo so this change have very little benefit, and not exported for testing. After #2126 it will be much more meaningful and also test friendly. Also string comparison is 2x faster than URL class.

Still using URL class now because it is very hard to make string comparisons stable. Added tests as well

src/util/ajax.test.ts
```diff
@@ -91,4 +92,27 @@ describe('ajax', () => {
...
+            host: 'somewhere.com'
+        } as any);
+
+        expect(sameOrigin('https://somewhere.com')).toBe(true);
+        expect(sameOrigin('https://somewhere.com/path')).toBe(true);
+        expect(sameOrigin('https://somewhere.com/path/?q=abc')).toBe(true);
+
+        expect(sameOrigin('HTTPS://somewhere.com')).toBe(true);
+        expect(sameOrigin('HTTP://somewhere.com')).toBe(false);
+        expect(sameOrigin('file:///c:/temp/foo.html')).toBe(false);
```
pramilk, line 108: Please add a test for invalid url to test the try/catch condition
expect(sameOrigin('./somewhere.com')).toBe(false);

pramilk: Actually './somewhere.com' should return true. Not sure, if its a valid supported format.

zhangyiatmicrosoft: > Please add a test for invalid url to test the try/catch condition expect(sameOrigin('./somewhere.com')).toBe(false);
Good catch!
Updated code to return true though -- the same behavior as before

src/util/ajax.ts
```diff
@@ -256,10 +256,17 @@ export const postData = function(requestParameters: RequestParameters, callback:
...
-    return a.protocol === window.document.location.protocol && a.host === window.document.location.host;
+export function sameOrigin(url: string) {
+    // URL class should be available everywhere
+    // https://developer.mozilla.org/en-US/docs/Web/API/URL
+    try {
+        const urlObj = new URL(url);
+        const locationObj = window.location;
+        return urlObj.protocol === locationObj.protocol && urlObj.host === locationObj.host;
+    } catch {
+        // all relative url or empty string will throw exception and it is considered as same origin
```
HarelM, line 267: Since this is a performance type of improvement, I would consider adding some `if` cases with basic string comparison to avoid using try-catch as those are probably very expensive, aren't they?

zhangyiatmicrosoft: Good call.

src/util/ajax.ts
```diff
@@ -256,10 +256,17 @@ export const postData = function(requestParameters: RequestParameters, callback:
...
-export function sameOrigin(url) {
-    const a: HTMLAnchorElement = window.document.createElement('a');
-    a.href = url;
-    return a.protocol === window.document.location.protocol && a.host === window.document.location.host;
+export function sameOrigin(inComingUrl: string) {
+    // URL class should be available everywhere
+    // https://developer.mozilla.org/en-US/docs/Web/API/URL
+    // In addtion, a relative URL "/foo" or "./foo" will throw exception in its ctor,
+    // try-catch is expansive so just use a heuristic check to avoid it
+    if (!inComingUrl || inComingUrl.indexOf('://') <= 0) {
```
pramilk, line 264: I think "foo.com/whatever", "://foo.com/whatever" are also valid inputs. Also, please update the comments.

src/util/ajax.test.ts
```diff
@@ -91,4 +92,48 @@ describe('ajax', () => {
...
         server.respond();
     });
+
+    test('sameOrigin method', () => {
+        jest.spyOn(window, 'location', 'get').mockReturnValue({
+            protocol: 'https:',
+            host: 'somewhere.com'
+        } as any);
+
+        expect(sameOrigin('https://somewhere.com')).toBe(true);
```
pramilk, line 102: Please add a test with port number.

zhangyiatmicrosoft: Good catch. Data URL previously was return false, but it surely should be sameOrigin, so return true now

src/util/ajax.ts
zhangyiatmicrosoft: both of them are considered as relative URLs (like before) and return true, meaning same origin.
The comment is correct.

pramilk: if user is on window.location is "abc.com" and tile url is "foo.com/whatever" then it's not relative. Similarly for "://foo.com/whatever".

pramilk: Not sure how to unresolve a resolved comment. So putting it here.

if user is on "abc.com" and tile url is "foo.com/whatever" then it's not relative. Similarly for "://foo.com/whatever". In these cases its not relative and should return false.

src/util/ajax.ts
```diff
@@ -256,10 +256,20 @@ export const postData = function(requestParameters: RequestParameters, callback:
...
-    a.href = url;
-    return a.protocol === window.document.location.protocol && a.host === window.document.location.host;
+export function sameOrigin(inComingUrl: string) {
+    // URL class should be available everywhere
+    // https://developer.mozilla.org/en-US/docs/Web/API/URL
+    // In addtion, a relative URL "/foo" or "./foo" will throw exception in its ctor,
+    // try-catch is expansive so just use a heuristic check to avoid it
+    // also check data URL
+    if (!inComingUrl ||
+        inComingUrl.indexOf('://') <= 0 || // relative URL
```
pramilk, line 266: On a website "abc.com"
1. if incoming url is "://foo.com/whatever", this code will return true. Which is incorrect.
2. if incoming url is "foo.com/whatever", this code will return true. Which is incorrect.

zhangyiatmicrosoft: Both cases were returning true before with HTMLAnchorElement, so that did not change ...

zhangyiatmicrosoft:
@HarelM ready to merge. Thanks

HarelM (approved):

HarelM:
Just to be on the same side, these tests pass before and after the implementation change, right?

zhangyiatmicrosoft:
> Just to be on the same side, these tests pass before and after the implementation change, right?

There was no direct test for sameOrigin function before. Maybe indirectly but not as comprehensive.
Manually validated each case in previous code and they all produced same results except for blob and base64 links, which should be sameOrigin indeed.

zhangyiatmicrosoft:
> I know there were no test, I was wondering if you can/did change the code back to how it was, ran the tests and made sure you get the same results - so you know there were no changes. If so, feel free to push the squash and merge button.

Tried that but previous code was using window.document.createElement --- in UT this is coming from jsdom, and the created HTMLAnchorElement always has http as protocol and "localhost" as host.
If I mock both window.document.createElement and window.document.location, I will be testing my mock against myself, and means nothing .. overall, it is quite a simple change and I manually verified all cases so should be ok.

Issue #94, opened 2021-03-06
Support ThreeJS for global coordinate


tan9710630:
MapboxGL has a [example](https://docs.mapbox.com/mapbox-gl-js/example/add-3d-model/) to render ThreeJS scene by using a CustomLayer. They just transforms the coordinate from global to local ,they used (0,0,0) for one particular latitude and longitude coordinate in world. I know this way they can avoid the accuracy lost in GPU, and it work every very in a small area. But if I want add  multiple 3D-Object to map in a very large scale, I have build multiple scene and renderer to CustomLayer. obviously that's very expensive and clearly not the best way.
I try to fix this problem by my own, but I don't have the ability to solve that, so I turn to you gays. If this feature request can't get in to roadmap, please at least show me how to do that, just show me a path is probable enough.
PS: I'm not a native English speaker, so if you see any grammatical errors, please just ignore that, I did my best to write these words.

JannikGM:
> If this feature request can't get in to roadmap,

It's more of a ThreeJS issue than maplibre / mapbox.

I don't think maplibre should take care of it, other than documenting the issue and providing samples for common frameworks like ThreeJS.

That said, a custom-layer type which is using the mapbox coordinate system and requests rendering for each tile separately could solve this issue partially.
However, there are more important shortcomings with the custom-layer interface.

> I have build multiple scene and renderer to CustomLayer. obviously that's very expensive and clearly not the best way.

maplibre / mapbox uses a tile-local coordinate system itself.

So this is one of the best solutions; although your implementation might not be the best. Someone could take a look at your implementation if you share your code.
I have no experience with ThreeJS, but you can probably set up clipping and render different tiles, too.

Other than tiling, I don't think there's any good alternative if the custom-layer renderer doesn't have high-precision coordinates (in vertices and matrices). It can be handled either on the CPU or the GPU.

>  I know this way they can avoid the accuracy lost in GPU, and it work every very in a small area

The ThreeJS sample is actually much better than the barebones custom-layer which shows significant amounts of jittering when zooming in: https://docs.mapbox.com/mapbox-gl-js/example/custom-style-layer/
*(In addition, that also uses a broken color for pre-multiplied alpha, which breaks on my Linux machine)*

>  please at least show me how to do that, just show me a path is probable enough.

[...]

Issue #128, opened 2021-04-07
Interactive area of symbols with interpolated size at fractional zoom levels


eriz0065:
**maplibre-gl-js version**: 1.14.0-rc.1

**browser**: chrome, edge

### Steps to Trigger Behavior

 1. Add a point source to map
 2. Add a symbol layer with icon-size determined by interpolation (e.g. "icon-size": ["interpolate", ["linear"], ["zoom"], 15, 0, 16, 0.5]) 
 3. The interactive area at fractional zoom levels (i.e. the box around the symbol for which mouse events are triggered) is larger than the actual icon size

### Link to Demonstration
https://codepen.io/eriz0065/pen/VwPMvdj
Open the console, mouseover and mouseout events at fractional zoom levels between 15 and 16 are triggered in a larger area with respect to actual icon size.

### Expected Behavior
Mouse events related are triggered considering the actual size of the layer symbol; mapbox from version 2.0.0 support this.

a-maggioni:
I confirm that I can still reproduce this problem with version 2.1.1

HarelM:
@drwestco I know you've been playing with this kind of stuff.
Any chance you can take a look at this issue?

drwestco:
The interactive area is determined by the symbol's collision box. Collision boxes are only recomputed at integer zoom levels; changing the size of a collision box affects layout and running a new layout pass at each fractional zoom level would incur a huge performance penalty.

Now it should be possible to address this specific case - modify queryRenderedFeatures to perform a filtering pass and check whether the point is within the actual symbol bounds.

HarelM:
Cool! Thanks for the info!
Can I close this one then?

drwestco:
Yeah, unless there's a strong use case for this, I'd close it.

PR #2376, opened 2023-04-11
Add CSS Color 4 syntax support - integration with changes from maplibre-style

kajkal:
Integration with changes from PR [#94](https://github.com/maplibre/maplibre-style/pull/94) from [maplibre-style](https://github.com/maplibre/maplibre-style).

- update color-related integration tests
- add color-interpolation render tests

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


test/integration/expression/expression.test.ts
```diff
@@ -1,6 +1,7 @@
 import path from 'path';
 import fs from 'fs';
 import glob from 'glob';
+import {printDiffOrStringify} from 'jest-matcher-utils';
```
kajkal, line 4: Anticipating potential questions: library `jest-matcher-utils` is already part of the project because it is one of the dependencies of `jest`, the new import in `package.json` is only for the ability to import a specific function in the code. The mentioned method is `jest` agnostic, it does not need the `jest` environment to work properly therefore it will not block a potential migration from `jest` to another testing framework.

Now when many tests from the `expression` suite fail it is difficult to assign the expected/actual to a specific test because they are displayed with console.log and, with many tests, they are easily mixed in with others (see [current_output.txt](https://github.com/maplibre/maplibre-gl-js/files/11195894/current_output.txt) file). With this change, the difference between expected/actual is displayed along with information about the failure of a particular test (see [new_output.txt](https://github.com/maplibre/maplibre-gl-js/files/11195891/new_output.txt) file).

HarelM:
@kajkal can you try and update the branch here and let me know if everything is working as expected in terms of updating the spec version?

kajkal:
Updated, colors (parsing/interpolation) are as expected, all tests are passing - it seems fine to me 👍

HarelM:
Any chance for a render test? So that this change could be visually seen?

kajkal:
Oh ok, those exist, I completely missed them and was only looking at the results of `jest` 😩

src/style/properties.ts
```diff
@@ -483,9 +483,10 @@ export class DataConstantProperty<T> implements Property<T, T> {
     }
 
     interpolate(a: T, b: T, t: number): T {
-        const interp: any = interpolateFactory(this.specification.type as any);
-        if (interp) {
-            return interp(a, b, t);
+        const interpolationType = this.specification.type as 'number' | 'color' | 'array' | 'padding';
```
drwestco, line 486: There should be a type for this in maplibre-gl-style-spec instead of hard-coding the possible values here.
```TypeScript
type interpolateType = keyof typeof interpolates;

File: src/style/properties.ts [hunk 3]
```diff
@@ -542,9 +543,11 @@ export class DataDrivenProperty<T> implements Property<T, PossiblyEvaluatedPrope
             return new PossiblyEvaluatedPropertyValue(this, {kind: 'constant', value: undefined}, a.parameters);
         }
 
-        const interp: any = interpolateFactory(this.specification.type as any);
-        if (interp) {
-            return new PossiblyEvaluatedPropertyValue(this, {kind: 'constant', value: interp(a.value.value, b.value.value, t)}, a.parameters);
+        const interpolationType = this.specification.type as 'number' | 'color' | 'array' | 'padding';
+        const interpolationFn = interpolates[interpolationType] as ((from: T, to: T, t: number) => T) | undefined;
```
drwestco, line 547: `interpolates` should have a proper type signature, so casting via `as` shouldn't be necessary.

kajkal:
As some render tests proved, there are cases in which an already parsed instance of `Color` class is passed to `Color.parse` function, contrary to [my assumptions](https://github.com/maplibre/maplibre-style-spec/pull/94#discussion_r1161059154).
I created PR [#128](https://github.com/maplibre/maplibre-style-spec/pull/128) (blocker for this PR) to fix that.

I added 2 render tests for color interpolation and added the ability to change the default sensitivity to color differences.
The tests in the previous configuration checked colors with a approximation: `rgb(46,215,46)` vs `rgb(0,255,0)` => pass.
By changing the sensitivity for tests directly related to colors, they are now able to detect a small change in the value of even one of the sRGB channels: `rgb(28,199,255)` vs `rgb(28,200,255)` => fail.

src/style/properties.ts
kajkal: Yes, you are right. This change would now require a separate PR to maplibre-gl-style-spec because I read your comment too late, so I'm not sure if it's worth it for those 2 repetitions - what do you think?

src/style/properties.ts
kajkal: I also think that `interpolates` should have a proper type signature in this case, but unfortunately sometimes TS doesn't want to cooperate :D
See [ts playground](https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABGEBbARgUwE4ApjZyoBcyaW2ANIlHKShjtVPeTgJSuPaIDeAUIiGJsmKCGxICRRAGoaiAFSJctRAFpE01OwDc-AL79+oSLASIAhtmyWAnvkIky3ANoBdZnRcUPzLhScPjgefILCouKSWk4AdKiWAA64uAAm1DDsiAC8AHzBeOk0cK4wnjTseobGEAgAzlCIMGBQOIlwADaWrXU5fAwU1Na2dgb6-FB2iZiIAMqTHZgACoTT2JOz0xAwwDAQ3fBgACpTM9mIAOQDOBeIAD6XDdjNAOa3DxfocJ2YlmDvl0wDABF1qHTg2BBw3sIMSllSqVeF3GEC6dV6ABFupYAML1KB-KArOBrSYAHiO+QEwhop1I8zsi2JpLsm0w212+3Mx1O+hptTATxA0AhqjpcwWy1WOA2Wx2ewOCBO0yy1JpwigAAsYHVYpNpn19Zg+cIjDTmq1sO0uq1cJZSEdqOgHf4CkEjmF1cIBQ0mi02p1FTyDectTq9acrL0rmxIfdLtC7Mjwl6fY0LQGbYcAGJIc4Zq2Bnqlf2FrNK07uE1enYqAvWoO51Upr0RMQSJD1os5sB2p3MKqtxAGRCYDp1GZqodCSIdqzV9Vm03VIA).

drwestco:
You can use the 'keyof typeof' syntax here instead of string literals, so
it keeps working when new interpolatable types are added. If we run into
more cases needing the interpolationType, we can then consider a change in
the style spec repo.

src/style/properties.ts
HarelM: I think it's worth it as I haven't released a new version yet...

test/integration/render/tests/background-color/colorSpace-hcl/style.json
```diff
@@ -3,7 +3,8 @@
   "metadata": {
     "test": {
       "width": 64,
-      "height": 64
+      "height": 64,
+      "threshold": 0.001
```
HarelM, line 7: Can we use an image that doesn't need this threshold change? I've struggled in the past to find this override, and I would like to avoid it as much as possible...

test/integration/render/tests/regressions/mapbox-gl-js#3107/style.json
```diff
@@ -45,8 +45,8 @@
       "type": "fill",
       "source": "geojson",
       "paint": {
-          "fill-outline-color": "hsla(0, 0, 0, 0)",
-          "fill-color": "hsla(0, 0, 0, 0)",
+          "fill-outline-color": "hsla(0, 0%, 0%, 0)",
```
HarelM, line 48: Is this mandatory now (the extra "%")? or will this work with the previous notation?

test/integration/expression/expression.test.ts
HarelM: The problem is that I'm already very much burden with updating dependencies, so I would like to avoid adding more work in that area.
If this is only for printing failing tests, I would rather avoid adding a dependency only for that, ATM.
My two cents at least...

test/integration/render/tests/background-color/colorSpace-hcl/style.json
kajkal: Now I see, I checked the results on Windows and on Ubuntu 22.04 (WSL), there the tests passed, here they have problems. 

I changed it after I decided that with their current threshold tests do not quite reliably check the colors, they allow too much margin, as I wrote above ^

> The tests in the previous configuration checked colors with a approximation: `rgb(46,215,46)` vs `rgb(0,255,0)` => pass. By changing the sensitivity for tests directly related to colors, they are now able to detect a small change in the value of even one of the sRGB channels: `rgb(28,199,255)` vs `rgb(28,200,255)` => fail.

Well, I will restore the previous threshold.

HarelM: My bad, I missed that comment. Can we tighten the threshold for all the tests? Will they still pass?

test/integration/render/tests/regressions/mapbox-gl-js#3107/style.json
HarelM: If this is the case we need to make sure we are documenting this more clearly in the changelog.

HarelM:
Seems like there's a render test that is failing. Note that I have recently added some instructions on how to get an image from the CI to get the test passing. There is a report that should be attached to the CI run, you can get it by going to the actions tab here in github, I believe.

kajkal:
Thanks, I was not aware that this was possible!
I adjusted the threshold for my new color interpolation tests - It should be working now.

test/integration/render/tests/regressions/mapbox-gl-js#3107/style.json
kajkal: Hmm... `hsl` without `%` is not a valid CSS color, Maplibre should not have supported it previously and there is nothing in the [documentation](https://maplibre.org/maplibre-style-spec/types/#color) about exceptional treatment of `hsl` without percentages. 
What changelog entry do you have in mind in this case?

test/integration/render/tests/background-color/colorSpace-hcl/style.json
kajkal: Unfortunately, no. More than 100 tests don't pass then, and I currently lack the knowledge of GL rendering differences between platforms, the Ubuntu machine at hand, and the time to address this.

test/integration/render/tests/regressions/mapbox-gl-js#3107/style.json
HarelM: In the same chengelog entry here we should add that "if you have an hcl color without % you should add it to create a valid one" or something similar - a kind of migration guide.

test/integration/expression/expression.test.ts
kajkal: I understand your concerns however this library is from under the `jest` umbrella and has the same model of releasing new versions. Its updates will take place together with updates of other libraries related to `jest`:
```json
"@types/jest": "^29.5.0",
"jest": "^29.5.0",
"jest-environment-jsdom": "^29.5.0",
"jest-matcher-utils": "^29.5.0",
```
So I think that one more `jest` library should not provide many new problems :)
Its use allows a quick and visually appealing (formatting, colors, etc.) assessment of what's wrong with a given test, which significantly contributes to the positive developer experience.

However, these are just my two cents added to this project, and if you still think this change is unnecessary, I will not insist on it.

test/integration/render/tests/regressions/mapbox-gl-js#3107/style.json
kajkal: Ok, so for `hsl` without `%` - breaking change. And what to do with the changelog entries added to maplibre-style-spec - repeat them in this repo changelog too?

HarelM: Just a single entry with the breaking change, migration guide and short description of the new functionality here.

test/integration/expression/expression.test.ts
HarelM: I tried to reduce the maintainance burden but failed, so for now, only critical libraries should be added, sorry...

kajkal: As you wish, removed.

HarelM (approved):

HarelM:
Generally approved, you'll need to set it for ready for review and resolve conflicts to have it merged.

PR #2381, opened 2023-04-11
Don't reload raster sources when RTL text plugin loads to avoid flicker.

### LINKED ISSUE #2380: Deferred loading of RTL text plugin causes flicker for raster layers
When a deferred load of the RTL text plugin completes, there's logic that causes the source caches to reload:

https://github.com/maplibre/maplibre-gl-js/blob/9f713cd2299a567e4f2a8000d984cd3f4ca14a19/src/style/style.ts#L226-L238

The purpose of this logic is to cause them to regenerate their symbol buckets with the new RTL shaping applied. If the plugin loads before the initial tile load, this will be a no-op, but if some tiles have already loaded it will reparse them. This is some extra work and the tile cache is cleared out, but the old tiles can still display until the new tiles are ready.

This logic is unnecessary for raster sources, since they don't have any symbols to regenerate. Also, raster tiles tend to load faster (because no parsing/bucket-generation is necessary), and are thus more likely to load before the RTL plugin loads. This generates extra work, but might not be that noticeable except for the fact that reloading a raster tile also resets the tile fade animation to the beginning.

The upshot of all this is an unsightly flicker during map load (in this example I'm just loading a single satellite layer and setting the rtl text plugin).

load_flicker

Stepping through in the debugger, you can see loaded tiles getting dropped one at a time as reloaded images arrive:

reloading_raster_tiles

And this is what the load looks like with the `reload()` call turned off for raster sources (PR coming soon):

load_no_flicker

ChrisLoer:
Fixes #2380

## Summary (from Issue)
When a deferred load of the RTL text plugin completes, there's logic that causes the source caches to reload:

https://github.com/maplibre/maplibre-gl-js/blob/9f713cd2299a567e4f2a8000d984cd3f4ca14a19/src/style/style.ts#L226-L238

The purpose of this logic is to cause them to regenerate their symbol buckets with the new RTL shaping applied. If the plugin loads before the initial tile load, this will be a no-op, but if some tiles have already loaded it will reparse them. This is some extra work and the tile cache is cleared out, but the old tiles can still display until the new tiles are ready.

This logic is unnecessary for raster sources, since they don't have any symbols to regenerate. Also, raster tiles tend to load faster (because no parsing/bucket-generation is necessary), and are thus more likely to load before the RTL plugin loads. This generates extra work, but might not be that noticeable except for the fact that reloading a raster tile also resets the tile fade animation to the beginning.

The upshot of all this is an unsightly flicker during map load (in this example I'm just loading a single satellite layer and setting the rtl text plugin).

load_flicker

Stepping through in the debugger, you can see loaded tiles getting dropped one at a time as reloaded images arrive:

reloading_raster_tiles

And this is what the load looks like with the `reload()` call turned off for raster sources (PR coming soon):

load_no_flicker

## Changes in this PR
This PR simply limits the sources that can reload to `vector` and `geojson`, the two types that contain symbol buckets. The bulk of the work was just setting up the mocking in the Style unit test to verify the new behavior.

cc @ibesora @HarelM 

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 - Fix LngLat `toArray` method return type to [number,number] ([#2233](https://github.com/maplibre/maplibre-gl-js/issues/2233))
 - Fix handling of text-offset with symbol-placement: line ([#2170](https://github.com/maplibre/maplibre-gl-js/issues/2170) and [#2171](https://github.com/maplibre/maplibre-gl-js/issues/2171))
 - Fix geolocate control permissions failure on IOS16 web view with fallback to `window.navigator.geolocation`
+- Prevent unnecessary reload of raster sources when RTL Text Plugin loads ([#2380](https://github.com/maplibre/maplibre-gl-js/issues/2380))
 - _...Add new stuff here..._
 
 ## 3.0.0-pre.4
--- a/src/source/rtl_text_plugin.ts
+++ b/src/source/rtl_text_plugin.ts
@@ -1,6 +1,6 @@
-import {Event, Evented} from '../util/evented';
 import {getArrayBuffer} from '../util/ajax';
 import browser from '../util/browser';
+import {Event, Evented} from '../util/evented';
 import {isWorker} from '../util/util';
 
 const status = {
@@ -56,6 +56,7 @@ export const registerForPluginStateChange = function(callback: PluginStateSyncCa
 export const clearRTLTextPlugin = function() {
     pluginStatus = status.unavailable;
     pluginURL = null;
+    _completionCallback = null;
 };
 
 export const setRTLTextPlugin = function(url: string, callback: ErrorCallback, deferred: boolean = false) {
--- a/src/style/style.test.ts
+++ b/src/style/style.test.ts
@@ -123,6 +123,54 @@ describe('Style', () => {
         sinonFakeServer.respond();
         new Style(getStubMap());
     });
+
+    test('RTL plugin load reloads vector source but not raster source', done => {
+        const style = new Style(getStubMap());
+        style.loadJSON({
+            'version': 8,
+            'sources': {
+                'raster': {
+                    type: 'raster',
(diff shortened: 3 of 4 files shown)


ChrisLoer:
Oops, looks like the tests don't work on latest `maplibre:main` -- I was working off an older version. I'll try to get them updated either later tonight or tomorrow.

HarelM:
Yea, that's not ideal, I agree. I had the same issue recently, I ended up introducing a default parameter to force clear cache... Wasn't too happy with that solution either...

HarelM:
Can you better describe the issue with testing? because I see there's a `clearRTLTextPlugin` method that is used in the tests to clear, and when you create a new style you can spy on `jest.spyOn(Style, 'registerForPluginStateChange');` can't you?

ChrisLoer:
> Can you better describe the issue with testing? 

At a high level -- it's callback hell. But going to describe the issues in more detail led me to some more satisfying fixes. The comment description explains it was basically two issues:

> - Make clearRTLTextPlugin also clear the _completionCallback
>   - Having a completion callback for an old plugin run is bad in general
>   - It specifically messed up the style tests by running during later tests
> - Call clearRTLTextPlugin when last plugin test is done
> - Set the stubMap style object -- this is a subtle one, but two frames _after_ this test runs, loadTileJSON triggered by this loadJSON will fail if the map doesn't have its style set, and the exception will get passed to a later style test.

I think the style unit tests will remain vulnerable to this kind of breakage though -- it's not just the global plugin logic but also the way the style defers work through `requestAnimationFrame` and can easily leave those requests pending even when any given test thinks it's already done. (in this case, we were marking ourselves `done()` in response to `style.load`, but loading the style had triggered source loading logic that would later fail)

HarelM (approved):

HarelM:
Thanks! Can you add a changelog entry for this bug/PR?

HarelM (approved):

PR #2390, opened 2023-04-13
Disable style validate via map Options

pramilk:
## Launch Checklist

Currently style validation is enabled by default. It's a very expensive operation.
Previously:
1. **_No_** way to enable/disable style validation if style is set via map options
2. Style Validation can be enabled/disabled if style is set via map.setStyle API. Default was to validateStyle

New changes:
1. Style validation can be enabled/disabled in both cases (Style set via setStyle or mapOptions)
2. Style Validation is disabled by default

For Maplibre sample page: https://maplibre.org/maplibre-gl-js-docs/example/add-a-marker/
 21 out of 66ms of style loading is spent on style validation.
image

Pending: Have not added tests. Waiting for comments before spending time on it.

## Launch Checklist
 - [ ] Link to related issues.
 - [*] Document any changes to public APIs.
 - [ ] Post benchmark scores.


HarelM:
This needs a broader input, I would recommend posting this PR in the slack channel and request for more feedback.
While I understand the motivation to change the default behavior, I tend to say that not many developers are actively turning this off to get better performance...

pramilk:
> This needs a broader input, I would recommend posting this PR in the slack channel and request for more feedback. While I understand the motivation to change the default behavior, I tend to say that not many developers are actively turning this off to get better performance...

Sure will post in Slack channel. Just to be clear, 
Previously:
1. **_No_** way to enable/disable style validation if style is set via map options
2. Style Validation can be enabled/disabled if style is set via map.setStyle API. Default was to validateStyle

New changes:
1. Style validation can be enabled/disabled in both cases (Style set via setStyle or mapOptions)
2. Style Validation is disabled by default

HarelM:
I generally think the first options is needed, the main question is about changing the default.

pramilk:
> I generally think the first options is needed, the main question is about changing the default.

I am fine changing the default, but then as you mentioned most users will leave it at default. Also, even with validation false and some invalid style, there will still be some error thrown, just not a nicely formatted error at style load.

wipfli:
Ah cool. Maybe we could change the error message to recommend turning style validation on?

ovivoda:
To me this looks like a good change. Of course the default behaviour needs a bit more input from the community.

maxammann:
I feel like this is an optimization which is beneficial for certain cases. I'm voting for opt-in, because new users definitely benefit from the style validation. I would suspect it is possible to trigger all kinds of weird crashes with validation disabled.

maxammann:
Maybe one other aspect: It might actually be, that style validation is a security feature for loading untrusted data. It might be that the validation protects against e.g. prototype pollution.

pramilk:
> Maybe one other aspect: It might actually be, that style validation is a security feature for loading untrusted data. It might be that the validation protects against e.g. prototype pollution.

I don't believe it might be the case, as anyone with ill intention can still disabled style Validation using map.setStyle API which  support  enabling/disabling validation (default enable).

pramilk:
Looking at different perspectives on this change. I will just let it be validateStyle=true by default.  It will be an easy change later if we want to make this change.

Will update this PR and will add tests. Thanks for the feedback.

HarelM (approved):

HarelM:
Thanks!

wipfli:
This is something we can highlight in the v3 release announcement. Turn off style validation and save some milliseconds on pageload...

PR #2393, opened 2023-04-14
Fix: Handle AddProtocol callback function returning a HTMLImageElement

pramilk:
## Launch Checklist

This PR fixes the issue with addProtocol, if the callback function returns a HTMLImageElement instead of ArrayBuffer.
Also added test cases for addProtocol

 - [ ] Link to related issues.
 - [*]Write tests for all new functionality.
 - [ ] Post benchmark scores.


pramilk:
> This generally looks great! Why would add protocol return an HTMLImageElement though?

If user want to have custom handling for raster image requests. Like Style sprite requests are currently made using Fetch which is a performance drag. We plan to use addProtocol to download it using HTMLImageElement. Can also be done for any Raster tiles which currently uses fetch.

Updated code to also support user returning ImageBitmap type for completeness.

src/util/ajax.test.ts
```diff
@@ -4,6 +4,8 @@ import {
     postData,
     AJAXError
 } from './ajax';
+import maplibre from '../index';
```
HarelM, line 7: Can we avoid importing this and only import and change the `config`?

src/util/image_request.test.ts
```diff
@@ -210,6 +213,47 @@ describe('ImageRequest', () => {
         server.respond();
     });
 
+    test('#addProtocol - returning arrayBuffer for getImage', done => {
+        maplibre.addProtocol('custom', (reqParam, callback) => {
```
HarelM, line 217: I don't think the image request tests should know about add protocol.
I would consider subbing the get image from ajax a bit more fully to not need the ajax class if possible and not know the internals of the implementation of a class that is not under tests.
i.e. improve the `stubAjaxGetImage` method or use a different one.

src/util/image_request.ts
```diff
@@ -148,6 +148,9 @@ namespace ImageRequest {
         return getArrayBuffer(requestParameters, (err?: Error | null, data?: ArrayBuffer | null, cacheControl?: string | null, expires?: string | null) => {
             if (err) {
                 callback(err);
+            } else if (data instanceof HTMLImageElement || data instanceof ImageBitmap) {
```
HarelM, line 151: The parameters for the callback are wrong as far as I understand as the callback expects an ArrayBuffer, which it doesn't get. Also the name is now a bit confusing, as it doesn't always gets an array buffer...

src/util/ajax.test.ts
pramilk: config contains the addProtocol implimentation details. addProtocol is the public API. I think we should directly test it instead of using config.REGISTERED_PROTOCOLS

src/util/image_request.test.ts
pramilk: Good point. getImage test should not know about the underlying implimentation. Have refactored code and have removed this test.
GetImage is not exposed via Ajax, thus I have to add these addProtocol test in this test file.

src/util/image_request.ts
pramilk: Refactored code to support it.

src/util/ajax.test.ts
HarelM: Then test it in the `index.test.ts` file not in the `ajax.test.ts` file.

HarelM:
Did you incorporate the other PR into this one? I'm confused...

pramilk:
> Did you incorporate the other PR into this one? I'm confused...

No its does not. This change need to  be done in order to support addProtocol for Images. In current code, only way for addProtocol to work for images was to return an ArrayBuffer (Which exposes the Implimentation details). Only workaround was to check for instanceof HtmlImageElement/ImageBitmap which was hacky as was pointed out in one of your comments. This change addresses your concerns.

src/util/ajax.test.ts
pramilk: Sure will move all tests to index.test.ts

pramilk: Updated

HarelM:
Thanks for fixing the code review comments!
Looking at the coverage there's still a need to cover a few more lines I think:
image
If this was the case of the coverage before this PR you can ignore this comment or add the relevant test :-)
Thanks again!

HarelM (approved):

pramilk:
> Thanks for fixing the code review comments! Looking at the coverage there's still a need to cover a few more lines I think: image If this was the case of the coverage before this PR you can ignore this comment or add the relevant test :-) Thanks again!

Thanks for through review. I don't know what I was thinking with test added all over the place instead of in index.test.ts.

I checked the test coverage report from the last merged checkin https://github.com/maplibre/maplibre-gl-js/commit/3df594864bdf3f8713eafa6b573251dda9d89061 and there is no additional coverage gap.

image

HarelM:
I think this PR added it, didn't it?

aliflux-omo:
I'm using the latest maplibre. When I return canvas in the addProtocol callback like this:

```
callback(null, canvas, null, null);
```

It throws an error on console:

```
Error: Could not load image because of The source image could not be decoded.. Please make sure to use a supported image type such as PNG or JPEG. Note that SVGs are not supported.\
    at util.ts:552:18
```

aliflux-omo:
The problem could be this block:

https://github.com/maplibre/maplibre-gl-js/blob/1cbb496f5cdf6751475986562eb0067c3ace7c4b/src/util/image_request.ts#L202

It only supports Image

HarelM:
canvas and HTMLImageElement are not the same thing as far as I know...

pramilk:
yes, HTMLCanvasElement directly is not currently supported But should be easy to add.

PR #2412, opened 2023-04-17
adapt links to a7e74509b1116207020626bf9942eac2563a7548

xabbu42:
This changes are necessary as index.html is no longer used for directories.

## Launch Checklist

### DIFF
--- a/test/bench/README.md
+++ b/test/bench/README.md
@@ -6,11 +6,11 @@ Benchmarks help us catch performance regressions and improve performance.
 
 Start the benchmark server with `npm run start-bench`.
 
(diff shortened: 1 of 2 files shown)


HarelM:
Isn't there an easier fix besides adding index.html to every url?

xabbu42:
I found no easy way to combine autogenerated directory lists and custom index.html with the st tool. We could bring back the index option but then loose the autogenerated index which are imho useful, for example for the debug pages. If somebody wants to invest the time, they could also try a custom st script or alternatives to st, but this is out of scope for me. In the meantime this change just adapts documentation to current reality.

HarelM (approved):

PR #2451, opened 2023-04-26
remove mapbox-gl-supported module

zhangyiatmicrosoft:
## Launch Checklist

#2452 
Net saving of 2186 bytes on minimized script from 761259 down to 759073

 - [x] Link to related issues.


src/ui/map.ts
```diff
@@ -2709,7 +2709,13 @@ class Map extends Camera {
     }
 
     _setupPainter() {
-        const attributes = extend({}, supported.webGLContextAttributes, {
+        const webGLContextAttributes = {
```
HarelM, line 2712: This assumes we are using modern browsers, doesn't it?
Is there an easy why to check web gl supported features without using mapbox-gl-supported?

HarelM:
I generally think this is a good direction, I don't like stale libraries, but we need to make sure we cover the relevant cases.

src/ui/map.ts
IvanSanchez: No, this just mimicks the behaviour of `@mapbox/mapbox-gl-supported`. No effort is made to detect support of `antialias`, `alpha`, `stencil` and `depth` (rather, see https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext#browser_compatibility - only Safari is a problem)

src/ui/map.ts
```diff
@@ -2709,7 +2709,13 @@ class Map extends Camera {
...
+            antialias: false,
+            alpha: true,
+            stencil: true,
+            depth: true
+        };
+        const attributes = extend({}, webGLContextAttributes, {
```
IvanSanchez, line 2718: I'd rather get rid of `extend()` and use a single object literal, something like

```
const attributes = {
            alpha: true,
            stencil: true,
            depth: true,
            failIfMajorPerformanceCaveat: this._failIfMajorPerformanceCaveat,
            preserveDrawingBuffer: this._preserveDrawingBuffer,
            antialias: this._antialias ?? false
};
```

IvanSanchez (changes requested):

src/ui/map.ts
HarelM: I don't like `extend` either, super hard to understand what you'll get back after using it...

src/ui/map.ts
zhangyiatmicrosoft: yes

src/ui/map.ts
zhangyiatmicrosoft: AGREE!

HarelM (approved):

pramilk:
@HarelM, @zhangyiatmicrosoft 
Does this documention need to be updated?
https://maplibre.org/maplibre-gl-js-docs/api/properties/#supported

HarelM:
I guess the example is either not relevant anymore or needs to be updated, I hope the supported will be removed automatically once the docs will be generated for version 3.
In any case, both issues should be addressed in the docs repo for now (until we move it here).

ambientlight:
@HarelM, @zhangyiatmicrosoft: should we fork the `@mapbox/mapbox-gl-supported`? since per webgl2 by default we would need to check for the existence of `webgl2` I guess now?

PR #2455, opened 2023-04-27
Remove non-fading raster tiles from the cache

### LINKED ISSUE #2445: Tiles with `raster-fade-duration: 0` inappropriately held for underzooming
When you zoom out on a raster layer, MapLibre holds on to underzoomed child tiles for the "raster fade duration" so that it can apply a cross fade. In the case where the fade-duration is 0, it shouldn't hold onto them at all once the parent is loaded, but its logic to do that incorrectly _always_ retains when fade-duration is 0:

https://github.com/maplibre/maplibre-gl-js/blob/c03607bad8a0d6574e9738b23fa84f3620713df5/src/source/source_cache.ts#L560

The tiles will stay in the pyramid potentially indefinitely, although only while in the viewport and within the "max underzooming" range. Here's an example of a satellite layer with higher-zoom (cloudy) tiles persistently drawn where we should have lower zoom tiles:

image

At Felt, we're working around this by using `raster-fade-duration: 1` which isn't visibly different but avoids the logic error.

I believe the correct logic should be:

```
if (!tile || !tile.fadeEndTime || tile.fadeEndTime <= browser.now()) continue;
```

a.k.a. "don't hold for fading, _if_ (1) the tile is already out of the pyramid, (2) the tile doesn't have a fade duration set, or (3) the tile has a fade duration set, but it's expired"

cc @ibesora

ibesora:
Fixes #2445 

As that issue describes, when raster-fade-duration was set to 0 the following condition failed and we kept non-fading rasters in the cache
https://github.com/maplibre/maplibre-gl-js/blob/c03607bad8a0d6574e9738b23fa84f3620713df5/src/source/source_cache.ts#L560

This PR checks for `tile.fadeEndTime` to be different than `undefined` because when that happens, tiles need to be retained but when `fadeEndTime` is 0 we want tiles to be removed from the cache.

There is a commit adding a failing test, another one fixing the issue and another one adding this to the changelog

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
THANKS!

HarelM:
Seems like this caused a render test to fail, can you please fix it?

ibesora:
> Seems like this caused a render test to fail, can you please fix it?

Updated the render test

HarelM:
Can you explain the change in the test? Send odd to me...

HarelM (approved):

HarelM:
Thanks for the explanation.

PR #2464, opened 2023-04-30
Reduce rendering calls on initial load

pramilk:
There is no reason to fire a render before map style is loaded. This change reduces number of render event fired on initial load.

Loading a Maptiler style invokes render 7 times, 1 before the style loaded and 6 after style loaded. After this change it saves the first render pass which is not required as there is nothing to render.

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -2806,7 +2806,7 @@ class Map extends Camera {
      * @private
      */
     _update(updateStyle?: boolean) {
-        if (!this.style) return this;
+        if (!this.style || !this.style._loaded) return this;
 
         this._styleDirty = this._styleDirty || updateStyle;
         this._sourcesDirty = true;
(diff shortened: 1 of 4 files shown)


src/ui/map.ts
```diff
@@ -2806,7 +2806,7 @@ class Map extends Camera {
      * @private
      */
     _update(updateStyle?: boolean) {
-        if (!this.style) return this;
+        if (!this.style || !this.style._loaded) return this;
```
HarelM, line 2809: Send like we are using a "private" field. Is there a better way to know if a style was loaded?

pramilk: style._loaded is also used in map.ts at line 1541. Agree its odd to use a private like variable outside the class. 

There is already a style.update() function which returns true,  if all style/sources/images are loaded. This function can't be used here.

HarelM (approved):

PR #2476, opened 2023-05-02
Lazy load style properties on demand

pramilk:
Currently object for all the default style properties is created on map control download. It has 2 issues.
1. It executed before worker thread creation, style download. So delaying these critical code paths.
2. It creates object for style properties which may not be used.

Generated code changes from:

```
const layout: Properties<FillLayoutProps> = new Properties({
     // Layout properties
});

const paint: Properties<FillPaintProps> = new Properties({
      // Paint properties
});

export default ({ paint, layout } as {
    paint: Properties<FillPaintProps>,
    layout: Properties<FillLayoutProps>
});

```

To:

```
let layout: Properties<FillLayoutProps>;
const getLayout = () => layout = layout || new Properties({
 // Layout properties
});

let paint: Properties<FillPaintProps>;
const getPaint = () => paint = paint || new Properties({
   // Paint properties
});

export default ({ getPaint, getLayout } as {
    getPaint: () => Properties<FillPaintProps>,
    getLayout: () => Properties<FillLayoutProps>
});
```
## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - [Breaking] Remove "mapbox-gl-supported" package from API. If needed, please reference it directly instead of going through MapLibre. ([#2451](https://github.com/maplibre/maplibre-gl-js/pull/2451))
 - Set fetchPriority for HTMLImageElement to help improve raster heavy scenarios ([#2459](https://github.com/maplibre/maplibre-gl-js/pull/2459))
 - Reduce rendering calls on initial load. No reason to try rendering before style is loaded. ([#2464](https://github.com/maplibre/maplibre-gl-js/pull/2464))
+- Lazy load default style properties to improve loading performance and reduce memory usage.
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 - [Breaking] Remove "mapbox-gl-supported" package from API. If needed, please reference it directly instead of going through MapLibre. ([#2451](https://github.com/maplibre/maplibre-gl-js/pull/2451))
 - Set fetchPriority for HTMLImageElement to help improve raster heavy scenarios ([#2459](https://github.com/maplibre/maplibre-gl-js/pull/2459))
 - Reduce rendering calls on initial load. No reason to try rendering before style is loaded. ([#2464](https://github.com/maplibre/maplibre-gl-js/pull/2464))
-- Lazy load default style properties to improve loading performance and reduce memory usage.
+- Lazy load default style properties to improve loading performance and reduce memory usage. ([#2476](https://github.com/maplibre/maplibre-gl-js/pull/2476))
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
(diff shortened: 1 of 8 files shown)


HarelM:
Can you elaborate on the motivation behind this PR.
Not sure I fully follow the value of introducing this extra complexity...?

pramilk:
> Can you elaborate on the motivation behind this PR. Not sure I fully follow the value of introducing this extra complexity...?

Main motivation is performance improvement and lower memory utilization. Here is the detailed explanation of what its trying to do.

Please refer to below snapshot of the profile for a simple loading of maplibre ([Link](https://perfmeasure.blob.core.windows.net/maplibre-sample/maplibre_old.html)).

1. Map control Js downloaded
2. Script compilation on main thread (includes style properties initialization before this change)
3. Client code execution which initializes Map control and start worker download.
4. Worker blob download
5. Script compilation on worker thread (includes style properties initialization  before this change)
6. Style parsing on main thread 
7. Style parsing on worker thread

**This change is trying to move the style properties initialization from step 2 to step 6 and also on worker from step 5 to step 7.**
 
image

This will have following advantages
1. All steps 3-7 can be executed a bit sooner. Specially client code in step 3.
2. Will reduce time between 5 and 7 (on Worker thread). Giving it a greater chance to be ready to receive styles when main thread broadcast it to worker. 
3. Since styles are now initialized with default values on demand (In step 6 & 7), For a map with say satelite imagery will not have to load default styles for Fill_extraction_style_layer or heatmap_style_layer etc. This will save both time and memory.

Here is the closer look at step 2
Before:
image

After:
image

 Here is the sample pages I am using for profiling
Current main branch: https://perfmeasure.blob.core.windows.net/maplibre-sample/maplibre_old.html
Current main branch with this change: https://perfmeasure.blob.core.windows.net/maplibre-sample/maplibre_new.html

HarelM:
Is it possible to change the implementation so that other part of the code are not aware of this underlaying change in "cache"/"lazy"?
i.e. use a getter instead of rename to `getPaint` etc?
From the image above, and I'm sure I'm getting it wrong so keep me honest here, this is a ~18ms improvement in load time? This doesn't sound like a lot, is this considered a good place to focus? I'm sure I'm getting it wrong, and regardless I would like to thanks you for all the amazing improvements you are doing to the performance of this library!

pramilk:
> Is it possible to change the implementation so that other part of the code are not aware of this underlaying change in "cache"/"lazy"? i.e. use a getter instead of rename to `getPaint` etc? From the image above, and I'm sure I'm getting it wrong so keep me honest here, this is a ~18ms improvement in load time? This doesn't sound like a lot, is this considered a good place to focus? I'm sure I'm getting it wrong, and regardless I would like to thanks you for all the amazing improvements you are doing to the performance of this library!

Thanks. Good point about using getter. Not sure why I don't used getter to begin with. This 18ms is on a high end machine. Slower machines gain the most with these optimizations. It will additionally save some memory (again small) and client code (map initialization or whatever client code is) can be executed a bit earlier.  Updated PR.

HarelM (approved):

HarelM:
Bottom line: do you think that the performance gain is worth the extra complexity?

pramilk:
> Bottom line: do you think that the performance gain is worth the extra complexity?

Yes it add little bit of complexity. But gains of 20-50ms that we are seeing on some of our metrics in our A/B testing makes its worth it. Most gains we are seeing are in Step 3 (please refer to my original explanation). This is allowing us to initiate our other client code a bit earlier. There is a small bit (0.5%) of memory usage improvement as well.

Issue #2411, opened 2023-04-17
Split tests run to optimize CI run time


HarelM:
Currently the split between the tests is not optimal and if a test fails the entire workflow needs to run again.
Also some workflows can run in parallel to reduce the total time the CI runs. the following is the longest run (on windows):
![image](https://user-images.githubusercontent.com/3269297/232467846-c6e42ec0-c2de-4837-bd88-c4efdffa5bb2.png)
@rotu, any chance you can take care of this?

HarelM:
Acceptance criteria:
Tests are split so that the longest run is a separate CI.
Assigning M bounty.
Requires #2482

birkskyum:
Did the github syntax wrong. Can this be assigned and closed?

birkskyum:
OC https://opencollective.com/maplibre/expenses/138080

PR #2495, opened 2023-05-08
Parallelize test workflow (splitting up jobs)

### LINKED ISSUE #2482: Have separate GitHub workflow for render tests
Having render tests in their own workflow will speed up the tests as it allow them to run in parallel with the integration/unit tests

birkskyum:
Close #2482, #2411

This splits up integration, unit, render and build tests. 

In this PR all these tests completed in 7m 57s, which is significantly shorter than a typical ~16 min run.

If i.e. a render test breaks, it only has the potential to cancel the other render test, but not the integration/unit/build tests which will keep running.

### DIFF
--- a/.github/workflows/test-all.yml
+++ b/.github/workflows/test-all.yml
@@ -42,6 +42,11 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version: lts/*
+      - name: Start display server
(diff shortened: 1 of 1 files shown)


HarelM (approved):
Interestingly, build dist takes most of the time right now.

PR #1622, opened 2022-09-07
More tests for sparse tilesets

xabbu42:
Tests for sparse raster tilesets and sparse tilesets with 204 status codes.

I also added some base tests on zoom 0 to show the zoom level where data is available. If this are considered redundant I can revert the last commit.

## Launch Checklist

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.

PR #2497, opened 2023-05-08
Render tests - Replace mocks with Puppeteer/Chrome

### LINKED ISSUE #2420: Add WebGL 2 capabilities to test suites
## Rationale

Both unit tests, integration tests and render tests currently rely heavily on the `gl` npm package, which is the undermaintained [headless-gl](https://github.com/stackgl/headless-gl). It rely on an unmaintained version of ANGLE, which is a distant fork of the  the [google repo](https://github.com/google/angle).

The `gl` package has been good enough up until now for development to run our tests in node, but unfortunately it has only WebGL 1 support, and with WebGL 2 having wide support, and WebGPU being released within a week, we need a more modern testing story. This lack of modern support has already proved problematic in pr's like #1891 , and it'll only get worse. The google ANGLE repo actually got WebGL 2 support, not to long ago, but there is slim hope that will ever be available in the ANGLE fork used by the `gl` package. There has been an open WebGL 2 ticket since 2017, so I wouldn't 
hold my breath on that one https://github.com/stackgl/headless-gl/issues/109#issuecomment-1374537584

### How does the current test suites work?

Currently, our unit tests are run with Jest in a JSDOM environment. In each unit test, the first thing we call is `beforeMapTest()`, which is a lot of browser mocking code we have in `/src/util/util.ts`. Part of that mocking process is creating a WebGL context, and that is where we pull in the `gl` package to help us out with that.

For render tests, we don't use JSDOM or `beforeMapTest`, but instead run the tests with the jest node environment, and here we additionally run the `/test/integration/render/mock_browser_for_node.ts` to create all the globals like global.navigator, 

### How could we do?

We could pull in a browser through either playwright or puppeteer, create an actual canvas, and a webgl/webgl2/webgpu context inside. This would allow us to delete all the mocking code for the integration/render tests, and have features as they get added to the evergreen browsers. The unit tests still need the mocks, because it's necessary to generate the coverage reports.

birkskyum:
Closes #2420

This PR replaces mocks of webgl1 with Chrome through Puppeteer in the render tests. This opens the door for future testing of webgl2/webgpu because it's available in Chrome but not in the mocks.

Upgrading the render tests effectively from integration (mocks) to e2e (full browser environment) did surface a few flaky tests (7 of 1238) that due to race conditions will have to be investigated. They have `expected-{os}-flaky{-n}.png` naming, and should eventually all be removed. There were also cases where the test expect to take a picture in the middle of an animation, but extracting glPixels isn't fast enough, so the picture we get is slightly delayed and the test flaky (i.e. `regression #2762` ), because there are no fakeTimers in Chrome.

Other cases where the different platforms are a few pixels off has been added as expected-{os}, to let pass.

The ubuntu and windows tests are running in parallel on 2 and 3 runners respectively, so 5 runners are occupied with render tests in order for them all to complete in ~11-17 min

 - [X] Link to related issues.
 - [X] Post benchmark scores.


test/integration/render/run_render_tests.ts
```diff
@@ -283,249 +243,432 @@ function getTestStyles(options: RenderOptions, directory: string): StyleWithTest
...
-function createFakeCanvas(document: Document, id: string, imagePath: string): HTMLCanvasElement {
-    const fakeCanvas = document.createElement('canvas');
-    const image = PNG.sync.read(fs.readFileSync(path.join(__dirname, '../assets', imagePath)));
-    fakeCanvas.id = id;
-    (fakeCanvas as any).data = image.data;
-    fakeCanvas.width = image.width;
-    fakeCanvas.height = image.height;
-    return fakeCanvas;
-}
+        class Tent3D {
```
HarelM, line 343: Is there a way to have this in a different file some how? I think this file should focus on running the tests.

test/integration/render/run_render_tests.ts
```diff
@@ -283,249 +243,432 @@ function getTestStyles(options: RenderOptions, directory: string): StyleWithTest
...
+            if (!operations || operations.length === 0) {
+                callback();
+
+            } else if (operation[0] === 'wait') {
+                if (operation.length > 1) {
+                    if (typeof operation[1] === 'number'
+                    ) {
+                        await new Promise((resolve) => {
+                            setTimeout(() => {
+                                resolve(10);
```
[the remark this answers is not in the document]
birkskyum: That is confusing yes - This can be anything. It just can't be empty.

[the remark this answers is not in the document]
birkskyum: Can do this

```ts
resolve(true); // Has to return something
```

test/integration/render/run_render_tests.ts
```diff
@@ -283,249 +243,432 @@ function getTestStyles(options: RenderOptions, directory: string): StyleWithTest
...
-                } else {
-                    options.operations = [['wait']];
-                }
+        // eslint-disable-next-line no-async-promise-executor
+        return new Promise(async (resolve, reject) => {
+            setTimeout(() => {
+                reject(new Error('Test timed out'));
+            }, options.timeout || 20000);
+
+            console.log('Starting gtest');
```
HarelM, line 581: gtest?

test/integration/render/run_render_tests.ts
```diff
@@ -283,249 +243,432 @@ function getTestStyles(options: RenderOptions, directory: string): StyleWithTest
...
+            if (options.addFakeCanvas) {
+                const fakeCanvas = await createFakeCanvas(document, options.addFakeCanvas.id, options.addFakeCanvas.image);
+                console.log('fakeCanvas', fakeCanvas);
```
HarelM, line 584: leftovers?

test/integration/render/run_render_tests.ts
birkskyum: Just a typo - should say 'test' - will remove the comment

test/integration/render/run_render_tests.ts
```diff
@@ -283,249 +243,432 @@ function getTestStyles(options: RenderOptions, directory: string): StyleWithTest
...
+                localIdeographFontFamily: options.localIdeographFontFamily || false as any,
+                crossSourceCollisions: typeof options.crossSourceCollisions === 'undefined' ? true : options.crossSourceCollisions
+            });
 
-                if (options.addFakeCanvas) {
-                    const fakeCanvas = window.document.getElementById(options.addFakeCanvas.id);
-                    fakeCanvas.parentNode.removeChild(fakeCanvas);
+            // Configure the map to never stop the render loop
+            map.repaint = typeof options.continuesRepaint === 'undefined' ? true : options.continuesRepaint;
+            //now = 0;
```
HarelM, line 616: remove these comments?

test/integration/render/run_render_tests.ts
birkskyum: Yes, I have removed this comment now

test/integration/render/run_render_tests.ts
```diff
@@ -283,249 +243,432 @@ function getTestStyles(options: RenderOptions, directory: string): StyleWithTest
...
+                    for (let i = 0, j = h - 1; i < j; i++, j--) {
+                        const start = i * stride;
+                        const end = j * stride;
+                        tmp.set(data.slice(start, start + stride), 0);
+                        data.set(data.slice(end, end + stride), start);
+                        data.set(tmp, end);
+                    }
+
+                    map.remove();
+                    //gl.getExtension('STACKGL_destroy_context').destroy();
```
HarelM, line 655: Is this needed or this comment can be removed?

test/integration/render/run_render_tests.ts
birkskyum: If this is taken out of page.evaluate(), we need to compile it to javascript and expose it somehow. I've tried a lot of ways before pulling all the needed functionality into the page.evaluate, with no luck

test/integration/render/run_render_tests.ts
```diff
@@ -642,8 +805,20 @@ function getReportItem(test: TestData) {
...
+    <div>
+    <p>Actual</p>
+    <img src="data:image/png;base64,${test.actual}" >
+    </div>
+    <div>
+    <p>Diff</p>
+    <img src="data:image/png;base64,${test.diff}">
+    </div>
+    <div>
+    <p>Closest expected</p>
```
HarelM, line 808: `data-alt-src` was used to allow clicking the image in order to see the diff one on top of the other.
I don't mind that the behavior is changed, but we need to change the relevant code as well, which I think is missing in this PR right now.

test/integration/render/run_render_tests.ts
birkskyum: It's obsolete like the rest of the stackgl

test/integration/render/run_render_tests.ts
HarelM: What about this SO answer:
https://stackoverflow.com/a/53762810
Something like: `await page.exposeFunction("Tent3D", Tent3D );`

HarelM:
Thanks for all the hard work around this!
I don't have a good way to compare the new images to those that already exist, so just please make sure that the new images are similar to those that exist.

test/integration/render/run_render_tests.ts
birkskyum: I've added data-alt-src to the actual/diff again

package.json
```diff
@@ -90,6 +91,7 @@
     "eslint-plugin-jest": "^27.2.1",
     "eslint-plugin-jsdoc": "^43.1.1",
     "eslint-plugin-react": "^7.32.2",
+    "fs-extra": "^11.1.1",
```
HarelM, line 94: Would love to avoid adding another dependency if possible.

birkskyum:
> Compare new images compared to those that already exist.

This is exactly why I try to be thorough around posting the result.zip, because the github 'Changed fies' is quite insufficient for reviewing here, because it just show a new image out of context.

test/integration/render/run_render_tests.ts
birkskyum: It can't handle typescript.

birkskyum: Just tried again, and it does break. It's very limited what can be sent to puppeteer like that. Maybe if all the types are removed, and it's stringified->sent to puppeteer -> .eval() and then it's there.

Screenshot 2023-05-10 at 16 12 09

https://stackoverflow.com/questions/51356212/pass-class-as-an-argument-to-page-evaluate-in-puppeteer

test/integration/render/run_render_tests.ts
```diff
@@ -1,40 +1,34 @@
...
-rtlTextPlugin['processStyledBidirectionalText'] = rtlText.processStyledBidirectionalText;
+// Move real-world fixtures
 
-let now = 0;
+const srcDir = path.join(__dirname, '../../../node_modules/@mapbox/mvt-fixtures/real-world');
+const destDir = path.join(__dirname, '../assets/mvt-fixtures/real-world');
+
+// To copy a folder or file, select overwrite accordingly
+try {
+    fse.copySync(srcDir, destDir, {overwrite: false});
```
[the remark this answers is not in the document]
birkskyum: I'll have a look at it

birkskyum:
These debug rectangles are flaky on all platforms - I wonder if it's a bug outside of the render tests.

Screenshot 2023-05-10 at 16 44 50

test/integration/render/run_render_tests.ts
HarelM: How well... We tried...

HarelM: Are these only used in render tests?
Maybe they can be removed from the code?

package.json
birkskyum: Got the server going.

test/integration/render/run_render_tests.ts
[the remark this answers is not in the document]
birkskyum: Got the server going.

test/integration/render/run_render_tests.ts
birkskyum: Yes, they were previously (still are i think) defined in test/integration/render/custom_layer_implementations.ts, so it's render test only

birkskyum:
So, the windows tests went through - It took 68 min., but there were 25 error tests that hit a timeout at 20s, so when they are fixed, it with the current setup it would take ~ 60 min for the windows runner. Ubuntu is < 20 min.

HarelM:
60 is a lot... Can we split the tests using different flows to make it parallel maybe? Or use multiple tabs on chrome?

birkskyum:
Splitting the windows test in 3 puts the whole test suite at ~30 min.

birkskyum:
More flaky debug boxes - because they are programmed to use different fonts on different platforms.

_Users_admin_Desktop_render%20tests_1%20win%20and%20ubuntu_results%204 html

birkskyum:
Fixing the windows errors took it down to 22 min.

birkskyum:
Splitting up ubuntu puts the test suite at just shy of 16 min.

I'll attempt to balance the windows splits to have equal runtime duration, because it could cut a minute - my target for this PR is 15 min.

.gitignore
```diff
@@ -12,6 +12,7 @@ coverage
 .nyc_output
 *_generated.js
 *_generated.js.map
+test/integration/assets/mvt-fixtures
```
HarelM, line 15: Is this still needed?

birkskyum: No, it's actually not

birkskyum:
The runtime is now consistently below 15 min, so I'll leave it at that.

There are ubuntu and win combined ~ 10 failing tests, 0 error/timeout, so it's getting close now.

birkskyum:
For the flaky tests, I add a expected-{relevant os}-flaky, so they can be looked into afterwards. I don't think puppeteer is to blame for these race conditions.

birkskyum:
Finally the tests are passing in a timely manner. I've update the PR description, and the changelog.

test/integration/render/run_render_tests.ts
```diff
@@ -283,249 +229,425 @@ function getTestStyles(options: RenderOptions, directory: string): StyleWithTest
...
-                    applyOperations(testData, map, operations.slice(1), callback);
-                });
-            }
-        } else {
-            const wait = function() {
-                if (map.loaded()) {
-                    applyOperations(testData, map, operations.slice(1), callback);
-                } else {
-                    map.once('render', wait);
+                console.log('IS A FUNCTION', typeof map[operation[0]] === 'function', operation[0]);
```
HarelM, line 528: Is this needed?

test/integration/render/run_render_tests.ts
```diff
@@ -283,249 +229,425 @@ function getTestStyles(options: RenderOptions, directory: string): StyleWithTest
...
-function getImageFromStyle(style: StyleWithTestData): Promise<Uint8Array> {
-    return new Promise((resolve, reject) => {
-        const options = style.metadata.test;
 
-        setTimeout(() => {
-            reject(new Error('Test timed out'));
-        }, options.timeout || 20000);
+        async function createFakeCanvas(document: Document, id: string, imagePath: string): Promise<HTMLCanvasElement> {
+            const fakeCanvas: HTMLCanvasElement = document.createElement('canvas');
+            console.log('document', document, fakeCanvas);
```
HarelM, line 539: Same here

test/integration/render/run_render_tests.ts
```diff
@@ -599,22 +721,76 @@ if (process.argv.length > 2) {
...
+);
+
+await new Promise<void>((resolve) => server.listen(2900, '0.0.0.0', resolve));
+await new Promise<void>((resolve) => mvtServer.listen(2901, '0.0.0.0', resolve));
 
 const directory = path.join(__dirname);
-const testStyles = getTestStyles(options, directory);
+let testStyles = getTestStyles(options, directory, (server.address() as any).port);
+
+if (process.env.PART_OF_TWO) {
```
HarelM, line 744: Can we use "total" and "current index" env variables instead?

HarelM:
Looks great overall! Thanks! I've added a few comments.

test/integration/render/run_render_tests.ts
```diff
@@ -599,22 +721,76 @@ if (process.argv.length > 2) {
...
+
+    testStyles = [first, second, third][parseInt(process.env.PART_OF_THREE)];
+}
+
 let index = 0;
+
+const page = await browser.newPage();
+page
+    .on('console', message =>
+        console.log(`${message.type().substr(0, 3).toUpperCase()} ${message.text()}`))
```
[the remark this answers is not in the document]
birkskyum: I've found it to be very useful, especially the network log which can say if an asset wasn't loaded.

test/integration/render/run_render_tests.ts
birkskyum: Sure, I've made a SPLIT_COUNT and a CURRENT_SPLIT_INDEX to have two env variables instead.

test/integration/render/run_render_tests.ts
birkskyum: Removed.

test/integration/render/run_render_tests.ts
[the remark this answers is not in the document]
birkskyum: It transfers the browser input to the workflow stdio, and we can't really access that information anywhere else.

xabbu42:
That pull request reversed my change in #1622 to return 204 for empty files. So the tileset sparse204 and the corresponding tests no longer test the behaviour of 204 responses as intended.

xabbu42:
My comment was about tests and not the logic in the library, so I'm not exactly sure where the confusion comes from.

To be more precise, I'm talking about the following change https://github.com/maplibre/maplibre-gl-js/pull/1622/files#diff-b894f8aa14661ba6783ba7c207b5f0a07226d46dce5f7976d673116a67061a5c. Together with the empty files in test/integration/assets/tiles/sparse204 the test server served 204 http status for this assets. Currently this assets are served as 200 with Content-Length 0. But the intention of the tests is to check the behavior of the library with 204 responses.

HarelM:
> That pull request reversed my change in #1622 to return 204 for empty files

This sentence sounds like a change of logic of a feature you added. 

In any case, can you please open a PR to address your concern? It will be a lot more productive than commenting on a closed PR...

HarelM:
Open an issue and link to this PR would be the best approach I believe.
I'm looking into adding this functionality to the tests.
I couldn't find a node static file server that supports 204 out of the box, I did manage to implement this using the build-in nodejs http server.
Do you know if the functionality was regressed in respect to 204? because I see that the tests are passing with my fix, but I'm not sure my fix is valid...
See the following PR:
- #5253

xabbu42:
The fix looks valid to me. Thank you!

I don't know about a regression with 204 in maplibre-gl-js. I looked into this again because I finally want to test what maplibre-native does in this cases as I have a suspicion that it is different. And this difference may be relevant to the discussion in https://github.com/maplibre/maplibre-style-spec/issues/938.

HarelM:
My concern regarding this feature is that even though I changed the code in the test harness now, no tests are failing, and when the PR that changed it got merged, no tests failed there.
From my understanding, it means that this feature is not covered in the render tests (it might be covered in other test suites, I'm not sure, maybe integration or unit tests, but not through render).
So, I'm not sure how valuable this change really is...
Is there a scenario I can use to make sure there's a test to cover this feature that is not part of the unit tests, or are the unit tests enough for this feature?

xabbu42:
What do you mean by this feature? Especially do you mean a feature of the library or of the tests?

Currently maplibre-gl-js handles 200 with Content-Length 0 the same as 204 (although I suspect the former will log some errors in the console), so its not surprising that the tests pass either way. Its just not the same thing that gets tested and I think given the amount of discussion especially about 204 results it is valuable to have some tests for that case. Thats was at least my reason why I added them.

HarelM:
I see, so there won't be any failing tests as maplibre-gl behaves the same for 200 and 204. I was under the impression there are some different code flows for 200 and 204, and those are not covered with tests.
If there is a console error in case of 204 it would be great to check that to know that we are testing 204 thoroughly, but that's nice to have I guess...

PR #2496, opened 2023-05-08
Add missing dependencies

jleedev:
Fixes these two build errors when using pnpm, yarn, or when using one of npm's newer (more correct) [install-strategy]s.

> [!] (plugin typescript) RollupError: @rollup/plugin-typescript: Could
> not find module 'tslib', which is required by this plugin. Is it
> installed?
> (!) Plugin typescript: @rollup/plugin-typescript TS2307: Cannot find
> module 'expect' or its corresponding type declarations.

Note that the [plugin says] "The `tslib` helper module always must be used", so it's needed to run even though we're using `importHelpers: false`.

[install-strategy]: https://docs.npmjs.com/cli/v9/commands/npm-install#install-strategy
[plugin says]: https://www.npmjs.com/package/@rollup/plugin-typescript#ignored-options

## Launch Checklist

 - [ ] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [ ] Briefly describe the changes in this PR.
 - [ ] Link to related issues.
 - [ ] Include before/after visuals or gifs if this PR includes visual changes.
 - [ ] Write tests for all new functionality.
 - [ ] Document any changes to public APIs.
 - [ ] Post benchmark scores.
 - [ ] Add an entry to `CHANGELOG.md` under the `## main` section.

PR #2504, opened 2023-05-10
Add missing deps (remake)

birkskyum:
Remake of #2496 , due to lots of deps updates since.

Closes #2496


HarelM (approved):
Approved. Why is this needed again?

birkskyum:
Apparently, it's troublesome to use some of the alternative package managers because they are more strict than npm.

PR #2510, opened 2023-05-12
Release 3.0.0-pre.6

birkskyum:
It seems like a good time for a pre-release, with quite a lot happening since pre.5.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 ## main
 
 ### ✨ Features and improvements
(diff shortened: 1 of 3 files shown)


HarelM (approved):

HarelM:
If this pre-release is good and there's no comment on the sky we should think about creating an official release.
I'll check with the ngx-maplibre library to see if there's anything special there...

birkskyum:
I agree. It would be great to get back on track with a normal release cycle asap.

PR #2511, opened 2023-05-12
Fix for blurry raster tiles due to raster tiles requests stuck in queue

pramilk:
On fast mouse wheel zoom across multiple zoom level raster tile requests were getting stuck in the image request queue.

There were 2 main issues causing it
1. When image was cancelled. ImageRequestQueueItem.cancel method might not have been attached. Please see added comment in imageRequest.ts. This caused a lot of image which were cancelled to continue to be downloaded.
2. Combination of when to move the queue and when not to was error prone and caused this issue.

This change now significantly simplify the image throttling code. Also does not require a lot of tests associated with it. 
Test coverage for Image_request.ts is at 100% after these changes.

Leaving it as draft while we validate the fix for any additional issues. On dev box testing can't repo any issue and can now additionally see a lot more images getting cancelled on a fast mouse wheel scroll. 

Repro:
1. Open https://perfmeasure.blob.core.windows.net/customerissue/blurry-tiles/maplibre_aerial_WithoutFix.html (latest main branch code - Pre.6)
2. Do a fast long mouse wheel scroll (might have to do it 4-5 times to repro)

With fix: https://perfmeasure.blob.core.windows.net/customerissue/blurry-tiles/maplibre_aerial_Fixed.html

Repro images:
image
image
image

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


src/util/image_request.ts
```diff
@@ -122,23 +124,31 @@ namespace ImageRequest {
...
             supportImageRefresh,
             callback,
             cancelled: false,
             completed: false,
+            cancel: () => {
+                if (!request.completed && !request.cancelled) {
+                    request.cancelled = true;
 
-            // Just a place holder. The real one will be assigned during processQueue()
-            cancel: () => {}
```
pramilk, line 133: Previously, if the image request was cancelled before image was actually taken out of the queue then it would be a no- op. This was causing a lot of images to be downloaded even though they were cancelled.

src/ui/map.ts
```diff
@@ -2894,15 +2894,6 @@ class Map extends Camera {
         if (this.terrain) this.terrain.sourceCache.update(this.transform, this.terrain);
         this.transform.updateElevation(this.terrain);
 
-        // a bit of counter intuitive:
-        // - when map is moving (throttled) image queue does not auto advance so need manually process it for each render
-        // or it may miss raster tiles.
-        // - when not moving (initial load or changing styles), image queue is self-driven to finish. manual process
-        // is not doing anything but wasting time.
-        if (this.isMoving()) {
-            ImageRequest.processQueue();
```
pramilk, line 2903: Turns out this was not needed as when map start moving then this._sourcesDirty is dirty and line 2890 `this.style._updateSources(this.transform);` will trigger a getImage request.

src/util/image_request.ts
```diff
@@ -122,23 +124,31 @@ namespace ImageRequest {
...
-        };
-        imageRequestQueue.push(queued);
+                    // Only reduce currentParallelImageRequests, if the image request was issued.
+                    if (request.innerRequest) {
+                        request.innerRequest.cancel();
+                        currentParallelImageRequests--;
+                    }
 
-        if (!isThrottled()) {
```
pramilk, line 137: Any of these checks for move the queue are not needed. `processQueue` method has the code to limit the number of requests to be made.

src/util/image_request.ts
```diff
@@ -207,43 +217,27 @@ namespace ImageRequest {
...
 
         // limit concurrent image loads to help with raster sources performance on big screens
-
         for (let numImageRequests = currentParallelImageRequests;
             numImageRequests < maxImageRequests && imageRequestQueue.length > 0;
             numImageRequests++) {
 
             const topItemInQueue: ImageRequestQueueItem = imageRequestQueue.shift();
             if (topItemInQueue.cancelled) {
+                numImageRequests--;
```
pramilk, line 240: This was causing loop to run lesser number of times then intended.

src/util/image_request.ts
```diff
@@ -207,43 +217,27 @@ namespace ImageRequest {
...
-            if (!isThrottled()) {
-                processQueue();
-            }
+            processQueue();
         }
     };
 
     /**
      * Process some number of items in the image request queue.
-     * @param {number} maxImageRequests The maximum number of request items to process. By default, up to {@link Config.MAX_PARALLEL_IMAGE_REQUESTS} will be processed.
```
pramilk, line 218: Removed since it was not used from any code.

src/util/image_request.ts
```diff
@@ -207,43 +217,27 @@ namespace ImageRequest {
...
-     * @returns {number} The number of items remaining in the queue.
      */
-    export const processQueue = (
```
pramilk, line 221: `processQueue` method does not need to be exported anymore.

src/util/image_request.test.ts
```diff
@@ -219,7 +176,50 @@ describe('ImageRequest', () => {
         }, false);
     });
 
-    test('getImage request cancelled for HTTPImageRequest', done => {
+    test('Cancel: getImage cancelling frees up request for maxParallelImageRequests', done => {
```
HarelM, line 179: Can this be moved up so that it will be easier to compare the diffs?

HarelM (approved):
I'm having a hard time following the tests logic. Did you add a test that simulates the issue which was failing before the code changes and now is passing?
If you have such a test, running it against old code shouldn't be an issue to see that it "used to fail" and now it passes.

src/util/image_request.test.ts
pramilk: Done.

pramilk:
> I'm having a hard time following the tests logic. Did you add a test that simulates the issue which was failing before the code changes and now is passing? If you have such a test, running it against old code shouldn't be an issue to see that it "used to fail" and now it passes.

I have added test which would have failed previously. Here is the breakdown of new/dropped tests

New test cases:
Cancel: Cancellation of an image which has not yet been requested
throttling: image queue will process MAX_PARALLEL_IMAGE_REQUESTS_PER_FRAME if throttling control returns true

Dropped test cases (Dropped as now queue does not have to be explicitly moved and will move automatically):
throttling: getImage queues requests for later processing
throttling: do NOT advance to next item when one of them is completed
throttling: DO advance to next item when one of them is canceled
throttling: process next item only when processQueue is called again

birkskyum (approved):
Thanks!

birkskyum:
Fixes #2577

PR #2512, opened 2023-05-13
Sunset webgl1

birkskyum:
This PR removes webgl1 from the distributions, but allow our coverage/unit tests to continue using a nodejs webgl1 mock for now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### ✨ Features and improvements
 - _...Add new stuff here..._
+- [Breaking] Remove WebGL1 support ([#2555](https://github.com/maplibre/maplibre-gl-js/pull/2512))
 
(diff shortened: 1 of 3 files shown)


HarelM (approved):

clementmas:
Does this mean Safari < 15 is not supported with MapLibre GL JS v3?

If so, I'll hold off on the upgrade because I still have about 3% of users using older versions of Safari.

HarelM:
Yes, webgl 2 is supported from safari 15 and onwards.

birkskyum:
@tpneumat , the reasoning behind sunsetting webgl1 is that it allow us to pursue a range of performance optimizations that would be very hard to implement using conditional logic. Webgl2 allow for a more modern render pipeline architecture of the graphics code, which in practice means more parallelization and doing more work up front instead of per frame.

Issue #2515, opened 2023-05-14
Remove WebGL1 usages and complete the migration to WebGL2


birkskyum:
There is a lot of conditional logic in the codebase, which is there to accommodate WebGL1 which isn't supported anymore.

All this code should be cleaned up, while allowing the unit tests enough mocking to run with the StackGL (webgl1 node implementation)

HarelM:
Adding M bounty.
Direction: https://github.com/maplibre/maplibre/issues/270

birkskyum:
I believe the existing PR's cover all cases of WebGLRenderingContext / webgl1 extensions.

HarelM:
THANKS!

birkskyum:
OC https://opencollective.com/maplibre/expenses/139066

PR #2516, opened 2023-05-14
Remove isWebGL2 type guard

birkskyum:
This is the first in a range of cleanup PR's related to #2515


src/util/test/util.ts
```diff
@@ -39,6 +39,74 @@ export function equalWithPrecision(test, expected, actual, multiplier, message,
     return test.equal(expectedRounded, actualRounded, message, extra);
 }
 
+export function setupMockWebGLContext(webglContext: any) {
```
HarelM, line 42: I would consider moving all this webgl mocking code to a separate file.

HarelM (approved):
Nice!

PR #2533, opened 2023-05-15
Remove DX items from changelog

birkskyum:
I think it makes sense to separate DX improvements from the features/improvements section, because it's not so relevant if you're a consumer of this library, whereas it can be quite informative for contributors to know if there has been changes to the stack / reorganizations / refactors / benchmarks / other changes to tests.


HarelM (approved):

birkskyum:
Sure, that was easy now that I've already filtered the list.

Issue #1545, opened 2022-08-24
Ability to use Map as a stateless component


Pessimistress:
## Motivation

Most modern web applications use a reactive/MVC UI framework, such as React, Angular and Vue. When building a complex application that contains a map component, it is very rare that the map library itself provides all the necessary functionalities. Some typical use cases include:

- Synchronize multiple Map instances, side-by-side, minimap, etc.
- Synchronize the Map camera with an external overlay, e.g. [deck.gl](https://deck.gl) or custom UI that moves with the map
- Match the Map with non-geospatial UI, e.g. a list of visible points of interests.

These scenarios all concern two-way state sharing between Map instances and other UI components. In the reactive programming paradigm, an exemplary implementation manages the camera state at the app's top level with a state store. Any UI component can request a change to that state. When the state does change, a rerender is triggered that flows down to the map and other components.

The design that maplibre-gl inherited from mapbox-gl does not make this easy. When the map receives user input, it immediately modifies the camera (transform) and rerenders. A `move` event is fired **after the fact**. Say the application updates its state based on a `move` event listener. When the other React components get rendered, they are always one animation frame behind the map.

I have maintained [react-map-gl](https://github.com/visgl/react-map-gl) for over five years. There had been different attempts to address this issue, all concerning a) blocking the map from updating its own camera b) forcing the map to redraw when React rerenders. Prior to v6, the wrapper sets `interactive: false` and implements its own input handlers. In v7, the wrapper swaps out `map.transform` with a "shadow transform" that matches the React props just before repaint. Both approaches have painful shortcomings including performance and a creeping amount of hacks.

[mapbox-gl issue regarding the stateful nature of input handling](https://github.com/mapbox/mapbox-gl-js/issues/3746)
[mapbox-gl issue regarding the state dependency of transform](https://github.com/mapbox/mapbox-gl-js/issues/12040)

### Proposal

Whenever `map.transform` is about to be modified (any operation that leads to a `move` event), fire a `premove` event with the proposed changes: `zoom`, `pitch`, `bearing`, `center`. The event listeners are allowed to mutate the proposed camera parameters. `map.transform` is then updated to the "approved" values. In pseudo code:

```js
/// Current
const tr = map.transform;
tr.bearing = newBearing;
tr.pitch = newPitch;
tr.zoom = newZoom;
tr.setLocationAtPoint(around, aroundPoint);

fireMoveEvents();
```

```js
/// Proposed
const tr = map.transform;
[...]

PR #2535, opened 2023-05-16
Add transformCameraUpdate callback

Pessimistress:
For #1545 

## Change List
- Add `transformCameraUpdate` option to `Map`
- Inject `_startCameraUpdate` and `_endCameraUpdate` into the following methods where `Camera.transform` is modified:
  + `Camera.jumpTo`
  + `Camera.easeTo`
  + `Camera.flyTo`
  + `HandlerManager._updateMapTransform`
  See inline comments
- A helper class `HandlerBase` for handlers to access the correct camera state

There should be no behavioral change or performance overhead if `transformCameraUpdate` is not assigned.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/geo/transform.ts
+++ b/src/geo/transform.ts
@@ -89,11 +89,11 @@ class Transform {
 
     clone(): Transform {
         const clone = new Transform(this._minZoom, this._maxZoom, this._minPitch, this.maxPitch, this._renderWorldCopies);
-        clone.copy(this);
+        clone.apply(this);
         return clone;
     }
 
-    copy(that: Transform) {
+    apply(that: Transform) {
         this.tileSize = that.tileSize;
         this.latRange = that.latRange;
         this.width = that.width;
--- a/src/geo/transform.ts
+++ b/src/geo/transform.ts
@@ -89,20 +89,24 @@ class Transform {
 
     clone(): Transform {
         const clone = new Transform(this._minZoom, this._maxZoom, this._minPitch, this.maxPitch, this._renderWorldCopies);
-        clone.tileSize = this.tileSize;
-        clone.latRange = this.latRange;
-        clone.width = this.width;
-        clone.height = this.height;
-        clone._center = this._center;
-        clone._elevation = this._elevation;
-        clone.zoom = this.zoom;
-        clone.angle = this.angle;
-        clone._fov = this._fov;
-        clone._pitch = this._pitch;
-        clone._unmodified = this._unmodified;
-        clone._edgeInsets = this._edgeInsets.clone();
-        clone._calcMatrices();
+        clone.copy(this);
         return clone;
+    }
+
+    copy(that: Transform) {
+        this.tileSize = that.tileSize;
+        this.latRange = that.latRange;
+        this.width = that.width;
+        this.height = that.height;
+        this._center = that._center;
+        this._elevation = that._elevation;
+        this.zoom = that.zoom;
+        this.angle = that.angle;
+        this._fov = that._fov;
+        this._pitch = that._pitch;
+        this._unmodified = that._unmodified;
+        this._edgeInsets = that._edgeInsets.clone();
+        this._calcMatrices();
     }
 
     get minZoom(): number { return this._minZoom; }
--- a/src/ui/camera.ts
+++ b/src/ui/camera.ts
@@ -127,6 +127,18 @@ export type AnimationOptions = {
     freezeElevation?: boolean;
 };
 
+export type CameraUpdateTransformFunction =  (next: {
+    center: LngLat;
+    zoom: number;
+    pitch: number;
+    bearing: number;
+}) => {
+    center?: LngLat;
+    zoom?: number;
+    pitch?: number;
+    bearing?: number;
+};
+
 abstract class Camera extends Evented {
     transform: Transform;
     terrain: Terrain;
@@ -161,7 +173,7 @@ abstract class Camera extends Evented {
 
     _transformSource?: Transform;
     _transformInProgress?: Transform;
-    _transformCameraUpdate: ((nextTransform: Transform, currentTransform: Transform) => Transform) | null;
+    _transformCameraUpdate: CameraUpdateTransformFunction | null;
 
     abstract _requestRenderFrame(a: () => void): TaskID;
     abstract _cancelRenderFrame(_: TaskID): void;
@@ -747,6 +759,7 @@ abstract class Camera extends Evented {
     jumpTo(options: JumpToOptions, eventData?: any) {
         this.stop();
 
+        this._startCameraUpdate();
         const tr = this.transform;
         let zoomChanged = false,
             bearingChanged = false,
@@ -774,6 +787,7 @@ abstract class Camera extends Evented {
         if (options.padding != null && !tr.isPaddingEqual(options.padding)) {
             tr.padding = options.padding;
         }
+        this._endCameraUpdate();
 
         this.fire(new Event('movestart', eventData))
             .fire(new Event('move', eventData));
@@ -1021,8 +1035,18 @@ abstract class Camera extends Evented {
         if (!this._transformCameraUpdate) return;
 
         this.transform = this._transformSource;
-        const tr = this._transformCameraUpdate(this._transformInProgress.clone(), this._transformSource);
-        this._transformSource.copy(tr);
+        const nextTransform = this._transformInProgress.clone();
+        const {
+            center,
+            zoom,
+            pitch,
+            bearing
+        } = this._transformCameraUpdate(nextTransform);
+        if (center) nextTransform.center = center;
+        if (zoom !== undefined) nextTransform.zoom = zoom;
+        if (pitch !== undefined) nextTransform.pitch = pitch;
+        if (bearing !== undefined) nextTransform.bearing = bearing;
+        this._transformSource.copy(nextTransform);
     }
 
     _fireMoveEvents(eventData?: any) {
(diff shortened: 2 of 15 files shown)


src/ui/camera.ts
```diff
@@ -992,6 +1017,44 @@ abstract class Camera extends Evented {
...
+     * Called before the camera is manipulated.
+     * If `_transformCameraUpdate` is specified, a copy of the current transform is created to track the accumulated changes.
+     * This underlying transform represents the "desired state" proposed by input handlers / animations / UI controls.
+     * It may differ from the state used for rendering.
+     * @returns Transform to apply changes to
+     */
+    _startCameraUpdate(): Transform {
+        if (!this._transformCameraUpdate) return this.transform;
+
+        if (!this._transformInProgress) {
```
birkskyum, line 1030: Maybe _transformInProgres could be renamed to _requestedCameraEndState or something like that, because I thought it was a bool at first and it's described to be the "desired state".

src/ui/camera.ts
```diff
@@ -159,6 +171,13 @@ abstract class Camera extends Evented {
     // so the linear interpolation between start and target keeps smooth and without jumps.
     _elevationStart: number;
 
+    /** Used to track accumulated changes during continuous interaction */
+    _transformInProgress?: Transform;
+    /** A callback used to defer camera updates or apply arbitrary constraints.
+     * If specified, this Camera instance can be used as a stateless component in React etc.
+     */
+    _transformCameraUpdate: CameraUpdateTransformFunction | null;
```
birkskyum, line 179: Think "transform" is overloaded in this codebase, to both be the camera state and the action of modifying the camera state. Maybe we could have two different terms, like _modifyCameraUpdate, to make it clear that this isn't a transform object.

src/ui/camera.ts
```diff
@@ -992,6 +1017,44 @@ abstract class Camera extends Evented {
...
+
+        if (!this._transformInProgress) {
+            this._transformInProgress = this.transform.clone();
+        }
+        return this._transformInProgress;
+    }
+
+    /**
+     * Called after the camera is done being manipulated.
+     * @param {Transform} tr - the proposed camera state
```
birkskyum, line 1038: Is this proposed by "modifyCameraUpdate" or proposed by handler/animation like in startCameraUpdate?

birkskyum:
Do we only allow setting center and zoom on the camera after the map is initialized, or is there a way to control pitch, bearing etc.?

birkskyum:
To see how this worked in practice, I made a debug page here (it can be dumped next to the other, if this PR branch is checked out):
https://github.com/birkskyum/maplibre-gl-js/blob/transform-camera-update/test/debug-pages/reactive.html

I think it works as expected, with minimal change to our public API. Here is a small clip of the left map driving the right.
https://github.com/maplibre/maplibre-gl-js/assets/74932975/355460ee-fc6d-46e4-94bf-797005e10c91

birkskyum:
I think the main function here transformCameraUpdate has so much utility that we should keep it a part of the public API, and not append a "_".

src/ui/map.ts
```diff
@@ -446,6 +450,7 @@ class Map extends Camera {
         this._locale = extend({}, defaultLocale, options.locale);
         this._clickTolerance = options.clickTolerance;
         this._pixelRatio = options.pixelRatio ?? devicePixelRatio;
+        this._transformCameraUpdate = options.transformCameraUpdate;
```
birkskyum, line 453: I think the main function here transformCameraUpdate has so much utility that we should keep it a part of the public API, and not append a "_".

birkskyum (approved):

birkskyum:
This little hook has surprisingly great utility - should not be overlooked.

- Without a js framework (like the debug-pages i made) it's possible to use gl js in a reactive way.
- All JS frameworks (react, vue, svelte, solid, angular etc.), especially where a wrapper isn't available, will get better ergonomics. You can finally get the direct camera values synced to a store, and they are not lagging 1 frame behind - terrain makes this even more important when the camera 'jumps' over a peak as that camera value wasn't possible to get at all.
- Advanced wrappers like react-map-gl, ngx-maplibre-gl, deckgl, maptiler-sdk-js, will have a much easier time dealing with the camera.
@wipfli , fyi

src/ui/handler/handler-base.ts
```diff
@@ -0,0 +1,42 @@
+import type Map from '../map';
+import type {PointLike} from '../camera';
+import type Transform from '../../geo/transform';
+import Point from '@mapbox/point-geometry';
+import LngLat from '../../geo/lng_lat';
+
+/**
+ * Shared utilities for the Handler classes to access the correct camera state
+ */
+class HandlerBase {
```
HarelM, line 10: Can this be done using composition instead of inheritance?

src/ui/camera.ts
```diff
@@ -992,6 +1017,44 @@ abstract class Camera extends Evented {
...
     }
 
+    /**
+     * Called before the camera is manipulated.
+     * If `_transformCameraUpdate` is specified, a copy of the current transform is created to track the accumulated changes.
+     * This underlying transform represents the "desired state" proposed by input handlers / animations / UI controls.
+     * It may differ from the state used for rendering.
+     * @returns Transform to apply changes to
+     */
+    _startCameraUpdate(): Transform {
```
HarelM, line 1027: startCameraUpdate doesn't sound like a method that will return a transform, I'm not sure what should be the right name though...

src/geo/transform.ts
```diff
@@ -89,22 +89,26 @@ class Transform {
...
-        clone._fov = this._fov;
-        clone._pitch = this._pitch;
-        clone._unmodified = this._unmodified;
-        clone._edgeInsets = this._edgeInsets.clone();
-        clone._calcMatrices();
+        clone.copy(this);
         return clone;
     }
 
+    copy(that: Transform) {
```
HarelM, line 96: `apply` instead of `copy`? I would expect `copy` to copy this into that and not the other way around.

src/ui/camera.ts
birkskyum: I personally don't mind a bit long names if that's what it takes to remove ambiguity.

src/ui/map.ts
```diff
@@ -1527,6 +1533,20 @@ class Map extends Camera {
         return this;
     }
 
+    /**
+     *  Updates the map's transform camera update with a new function
```
HarelM, line 1537: [Optional] Can we mark it as experimental for now? In theory, I would like the camera to be a separate component in the map and not that a map inherits from the camera, but I'm not sure when this will happen.

HarelM:
@Pessimistress thanks for taking the time to open this PR!
I think there are still some naming and small fixes left to do, but this looks great in general.

src/ui/camera.ts
Pessimistress: This transform tracks what's proposed by handler/animation. When we call `_transformCameraUpdate` we pass a clone of this transform to prevent it being mutated by the user.

birkskyum: Okay, so it's the same as the one in startCameraUpdate, and not the modified output from transformCameraOutput

src/ui/handler/handler-base.ts
Pessimistress: What do you have in mind? I don't see a way to do it without being incredibly verbose.

Pessimistress:
Some renames according to review comments:

- `Transform.copy(that: Transform)` -> `Transform.apply(that: Transform)`
- `Camera._transformCameraUpdate` -> `Camera.transformCameraUpdate`
- `Camera._transformInProgress;` -> `Camera._requestedCameraState`
- `Camera._startCameraUpdate(): Transform` -> `Camera._getTransformForUpdate(): Transform`
- `Camera._endCameraUpdate(tr: Transform)` -> `Camera._applyUpdatedTransform(tr: Transform)`

I don't think `transformCameraUpdate` is necessarily confusing because the internal term "transform" as a camera/viewport is not exposed to the end user. There's already an option `transformRequest` that does a similar thing (intercept arguments and modify them before use).

How concerned are you about the bundle size?

src/ui/handler/handler-base.ts
HarelM: Instead of this being a base class, define something like `PositionProvider` or `TransformProvider`, initialize it in every handler and use it when needed.

Partially related, since all handlers now require a map to be initialized it might be worth considering the removal of the `map` parameter in `cameraAnimation`. I haven't looked deeply into it, but "it feels" like this might not be needed anymore.

src/ui/camera.ts
```diff
@@ -992,6 +1019,46 @@ abstract class Camera extends Evented {
...
+
+        const nextTransform = tr.clone();
+        const {
+            center,
+            zoom,
+            pitch,
+            bearing,
+            elevation
+        } = this.transformCameraUpdate(nextTransform);
+        if (center) nextTransform.center = center;
```
HarelM, line 1054: Why is the center check different from the other checks (other use !== underfined)?

HarelM:
> How concerned are you about the bundle size?

Not really concerned. Please update the size in the test.
Overall looks good. THANKS!

src/ui/camera.ts
birkskyum: Because center isn't a primitive, it's the only of these where this check work. I.e. if bearing = 0 that would be false, so it would not be set with if(value). The center condition could be made like the other ones, but I don't know what the advantage of that would be except for aesthetic

birkskyum:
Thanks for the renaming, i find this easier to reason about. We need a changelog entry, but apart from that I don't have further inquiries.

HarelM (approved):

birkskyum (approved):

birkskyum:
Great job!

PR #2543, opened 2023-05-17
Fix event.isSourceLoaded to reflect the state of source loading for sourcedata event

### LINKED ISSUE #2539: sourcedata's event.isSourceLoaded is only true for last emission when the tiles for all sources are loaded
**maplibre-gl-js version**: 3.0.0-preview.7 as of 66fc5fedf0cee27a8d9aa1c706f972827f15bef7

sourcedata's `event.isSourceLoaded` may be incorrect, as it doesn't actually reflect the load state of the source (of `event.sourceId`) if your style contains more than one source or the sprites are not loaded. Observe that the event actually takes https://github.com/maplibre/maplibre-gl-js/blob/e43e6ef2e71e93bf3b23bca2c4600f069370136a/src/style/style.ts#L744 which will only be set to true on the complete load of all sources (including tiles for all source and sprites). In tests, this normally corresponds to only the last invocation (provided that sprites are not loading), something like:
```
...
source1: false
source2: false
source2: false
source1: false
source1: true
```

Judging from https://github.com/maplibre/maplibre-gl-js/commit/bacb46d0afbc8f2a7db8a30aa3ac3e6fd20a8f34 which is from https://github.com/mapbox/mapbox-gl-js/pull/4254, it has been flipped from the expected `sourceCache.loaded()`. However the original issue that the PR was deemed to resolve doesn't look to be relevant anymore. Flipping the value of the isSourceLoaded to  `sourceCache.loaded()` resolves the issue, where all last loads will be set to true for each source, the above becomes expected:

```
...
source1: false
source2: false
source2: true
source1: false
source1: true
```

### Steps to Trigger Behavior

 1. instantiate the map with a style that contains more then one source.
 2. log the sourcedata events: 
    ```js
    map.on('sourcedata', event => console.log(`${event.sourceId}: ${event.isSourceLoaded}`));
    ```
 3. browse around beyond the visible viewport, observe the logs in the console where only last sourcedata emission has `isSourceLoaded` set to true.

### Link to Demonstration

Navigate to https://glasssh.blob.core.windows.net/repro/event_repro.html, open devtools' console, browse around, check the log messages printed.

### Expected Behavior
Event emission from the each source's last pending tile load from https://github.com/maplibre/maplibre-gl-js/blob/22909c817d55e2a1e172b12a43d792db3e23b40b/src/source/source_cache.ts#L281 should result in `event.isSourceLoaded` set to true. The value of `event.isSourceLoaded` should be logically equivalent to `map.isSourceLoaded(event.sourceId)`.

### Actual Behavior
Event emission from tile load result in `event.isSourceLoaded` set to `true` only once the entire style's https://github.com/maplibre/maplibre-gl-js/blob/e43e6ef2e71e93bf3b23bca2c4600f069370136a/src/style/style.ts#L437 returns true

### Comment
This sounds like the trivial fix unless I am missing something. However this is a breaking change, and the current state may be feel confusingly two-fold (when the entire `style.loaded()` is true, it also implies that any individual source is loaded), however given that the event-driven logic like: 
```
map.on('sourcedata', event => event.sourceId === 'something' && event.isSourceLoaded ? doSomethingWithTileData() : doNothing());
```
could be a common pattern for people to do something with tile data once all tiles in viewport are loaded for a given source, I think we should fix this.

ambientlight:
Fixes #2539 

* use `sourceCache.loaded()` for `event.isSourceLoaded`
* `sourceCache.loaded()` is true, if `used` or `usedForTerrain` in set to false (after source loaded)
* do not consider source cache as loaded before update is called (because we don't know if we need to fetch tiles yet). 
* added new `idle` **sourcedatatype** used in event emissions to propagate the `source_cache.loaded()` change to `true` for cases when aren't any tile loads pending: on `sourceCache.update()` called when viewport is outside of source bounds, on image/video/canvas `source.prepare()`

## Context
The goal of this change is to make sure we have a single `sourcedata` invocation with `event.isSourceLoaded` **true** for each source at the end after initial load and subsequent interactions that result in tile fetches. Current behavior has used `style.loaded()` for the value of `event.isSourceLoaded` and this would result in whatever source tile load that comes last to have this set. Flipping the https://github.com/maplibre/maplibre-gl-js/blob/e43e6ef2e71e93bf3b23bca2c4600f069370136a/src/style/style.ts#L744 to `sourceCache.loaded()` is not enough because we will have `event.isSourceLoaded` set to true after the source is ready (tilejson fetched), but prior to any tiles actually being fetched. 

We may only know whether the source will have tiles to fetch once the source receives transform (on update on render), and thus `source_cache.loaded()` would now require update to be called at least once on the source to be considered loaded. 

Additionally, because the `sourcedata` event emissions are not tied to `source_cache.loaded()`, we are not guaranteed we will actually receive the emission with `isSourceLoaded` set to `true`. This applies to cases, when:
*  we don't have any tiles to fetch when our source has bounds outside of the viewport (and all `sourcedata` invocations may happen prior to the first `source_cache.update()`) 
* we use `image_source`, `canvas_source`, `video_source` that will not set the tile state to `loaded` on its load, and will only set the tile state later on `prepare()` on `map.painter.render()`.

To compensate for this, additional `sourcedata` event invocations with **sourcedatatype** `idle` have been added.

## Launch Checklist

 - [X] Link to related issues.
 - [x] Handle the source with bounds outside of the viewport, where on initial load there may be no events with `isSourceLoaded` true until the camera is moved to tileBounds.


HarelM:
Note that we will release 3.0 soon. If you would like to have this in please finish the changes and request for review.
Thanks!!

ambientlight:
@HarelM: yes, I would like to have this in, sure, need to add the tests now, probably need a few hours, thanks.

birkskyum:
just fyi - the breaking render test ( /debug/collision ) is known to be a bit flaky since the migration from node mocks to puppeteer, so don't mind it.

birkskyum:
@ambientlight , is this ready?

ambientlight:
@birkskyum: yes, it is ready for review.

HarelM:
It's a bit more complicated than I expected, so I'll need a bit more time to review this, hopefully tonight.

src/source/video_source.ts
```diff
@@ -177,13 +177,19 @@ class VideoSource extends ImageSource {
             gl.texSubImage2D(gl.TEXTURE_2D, 0, 0, 0, gl.RGBA, gl.UNSIGNED_BYTE, this.video);
         }
 
+        let willLoad = false;
         for (const w in this.tiles) {
             const tile = this.tiles[w];
             if (tile.state !== 'loaded') {
                 tile.state = 'loaded';
```
birkskyum, line 184: How is this to be understood? If the tile isn't loaded, it's loaded?

ambientlight: vector / raster / raster-dem sources will set the `loaded` status in load callback (either in fetch callback, or tile parsing callback to load/reload tile from web worker), but in those sources that inherit from ImageSource, note that `_finishLoading` will not update the tile state and the tile state will only get update on the first render that will call the painter.render and then fill the tile with the texture from here in prepare()

src/ui/map.test.ts
```diff
@@ -542,6 +542,24 @@ describe('Map', () => {
...
         });
 
+        test('Map#isSourceLoaded (equivalent to event.isSourceLoaded)', done => {
+            const style = createStyle();
+            const map = createMap({style});
+
+            map.on('load', () => {
+                map.on('data', (e) => {
+                    if (e.dataType === 'source') {
+                        expect(map.isSourceLoaded('geojson')).toBe(e['isSourceLoaded']);
```
HarelM, line 552: Can we force `e` to a type instead of calling the property? It will make refactoring and find usages better for others.

src/ui/events.ts
```diff
@@ -292,7 +292,7 @@ export type MapLibreZoomEvent = {
  * @property {boolean} [isSourceLoaded] True if the event has a `dataType` of `source` and the source has no outstanding network requests.
  * @property {Object} [source] The [style spec representation of the source](https://maplibre.org/maplibre-style-spec/#sources) if the event has a `dataType` of `source`.
  * @property {string} [sourceDataType] Included if the event has a `dataType` of `source` and the event signals
- * that internal data has been received or changed. Possible values are `metadata`, `content` and `visibility`.
+ * that internal data has been received or changed. Possible values are `metadata`, `content`, `visibility` and `idle`.
```
HarelM, line 295: Not sure if this is the right place, but I would be interested to know what each one represents (metadata vs content vs visibility vs idle)...

src/source/video_source.ts
HarelM: I'm not sure I fully understand the explanation and reason for this code as well...

src/source/source_cache.test.ts
```diff
@@ -1529,6 +1530,10 @@ describe('source cache loaded', () => {
 
         sourceCache.on('data', (e) => {
             if (e.sourceDataType === 'metadata') {
+                const tr = new Transform();
```
HarelM, line 1533: Can you elaborate on why this code was added to the tests?

HarelM:
Can you also check the code coverage for the new code in the report here in the CI to make sure the new code is covered?

src/source/source_cache.test.ts
ambientlight: transform is a parameter to `source_cache.update()`
source_cache.loaded() now requires `update()` to be called at least once on the source to be considered loaded. (we may only know whether the source will have tiles to fetch once the source receives transform)

src/ui/events.ts
ambientlight: for now the generated docs don't look too clunky so I guess it should ok, maybe even we can put in the bullet-list after `dataType`'s bullet-list:

image

src/source/video_source.ts
ambientlight: you mean why do we initialize the texture and assign it to the tiles on `prepare()` and the tile state is not changed to loaded prior to first iteration over the tile in `prepare()`? Like if the texture could be assigned to tile on https://github.com/maplibre/maplibre-gl-js/blob/7db0e8111f05770d938611375104dbc75e1b1ae0/src/source/image_source.ts#L157 and then on subsequent https://github.com/maplibre/maplibre-gl-js/blob/7db0e8111f05770d938611375104dbc75e1b1ae0/src/source/image_source.ts#L246
I don't know now tbh, may need to trace the history to see how this evolved.

or why the texture was assigned `once` in https://github.com/maplibre/maplibre-gl-js/blob/7db0e8111f05770d938611375104dbc75e1b1ae0/src/source/video_source.ts#L183 (on `loading` -> `loaded` state flip). I also don't know, I see the https://github.com/maplibre/maplibre-gl-js/commit/1d86f681b9df8d0088de3098c4e715b907846f73, based on the commit message, I can only speculate it has something to do with the texture lifecycle with regards to rendering, on every `prepare()` that is called on each render, we will recreate the texture instance, if we don't reassign it to the tile, maybe there is smart behavior / optimization at play from webgl where texture will be cached idk...

birkskyum: What trips me up is this

```ts
        let willLoad = false;
        for (const w in this.tiles) {
            const tile = this.tiles[w];
            if (tile.state !== 'loaded') {
                tile.state = 'loaded';
                tile.texture = this.texture;
                willLoad = true;
            }
        }
```

As I understand, willLoad will be true if of the tile aren't loaded (tile.state !== 'loaded'), which makes a lot of sense to me. If willLoad is set, work is still ongoing, and the 'idle' event will not be fired yet.

So far, so good. What then confuses me is this line `tile.state = 'loaded';`, because we just concluded that it wasn't loaded, and that more work has to happen, so why would it's set to be loaded here puzzles me. If I'm not mistaken, all tiles will have tile.state = `loaded` after going through this loop once.

src/ui/events.ts
birkskyum: So the list is `metadata, content, visibility and idle`. If I understand Harel's question is it what the definition for each of these four states are.

src/source/video_source.ts
birkskyum: Aaah, I think I'm finally starting to get it now. So `willLoad` indicate if new tiles were added (loaded) to the source, which will set `willLoad` flag that cause the `idle` event to fire from the source. When the source says it's `idle`, what is it that will then load? I assume something will load because the `willLoad` flag is what triggers it..

ambientlight: willLoad is just a flag so that I know when to fire an idle event. Maybe the name is not ideal. It is just:
```js
const isAnyTileNotLoadedYet = Object.values(this.tiles).reduce((notLoaded, tile) => loaded || tile.state != 'loaded', false)
```
Idle will only fire once, until the transform changes and if there are any new tiles added (map is moved, scaled etc...). 

Again, note, raster / vector / raster_dem sources don't need this since every tile load will result in the emissions at https://github.com/maplibre/maplibre-gl-js/blob/608a53a8c640c1fed35e792d24564e022c874020/src/source/source_cache.ts#L281, and because the `tile.state = 'loaded'` will be set in the tile load callback, when this invocation happens for the last pending tile, the `event.isSourceLoaded` will be set to true.

birkskyum: I assume it's the source_cache.loaded() that gets propagated by it, so would it make sense to rename `willLoad` to maybe `newTilesLoaded`?

src/ui/events.ts
ambientlight: @birkskyum: added description per https://github.com/maplibre/maplibre-gl-js/pull/2543/commits/755567777a4f97a1697a96d5a0aca5a462f650d0

 * `'metadata'`: indicate that any necessary source metadata has been loaded (such as TileJSON) and it is ok to start loading tiles
 * `'content'`: indicates the source data has changed (such as when source.setData() has been called on GeoJSONSource)
 * `'visibility'`: send when the source becomes used when at least one of its layers becomes visible in style sense (inside the layer's zoom range and with layout.visibility set to 'visible')
 * `'idle'`: indicates that no new source data has been fetched (but the source has done loading)

birkskyum (approved):
Thanks @ambientlight !

PR #2545, opened 2023-05-17
Generate report by default and hide debug messages in render tests.

zhangyiatmicrosoft:
## Launch Checklist

- Render tests are generating a lot of messages and when running >1200 of them together it is difficult to find the failed ones, so hide them by default add a <code>--debug</code> switch to turn on as needed.
- Puppeteer tests takes ~10 min to run them all, so generate reports by default for easy troubleshooting. This can be skipped by <code>--skip-report</code>
- Clean up and cosmetic changes: make a main "executeRenderTests" function as entry point to improve readability.
- Update readme.md with instructions

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.

PR #2549, opened 2023-05-18
Remove --report parameter

HarelM:
Let's see if removing this parameter fixes the CI...

Related to comments in here: #2545


birkskyum (approved):
It seems like it. - The render tests run for a more realistic amount of time at least

PR #2552, opened 2023-05-18
Capture initial resize event

### LINKED ISSUE #2551: `moveend` is fired without a reason due to resize observer
The following browser test fails as of version 3.0.0.pre.5:
It shows that moveend is fired before load on a map that doesn't move.
This is due to the change to `ResizeObserver` as it fire an event right after subsribing to it.
See #2157 for more info.
```js
    test('No moveend before load', async () => {
        let loadWasFirst = await page.evaluate(() => {
            const map2 = new maplibregl.Map({
                center: [10, 10], // starting position [lng, lat]
                zoom: 10 // starting zoom
            });
            return new Promise<boolean>((resolve, _reject) => {
                map2.once('moveend', () => resolve(false));
                map2.once('load', () => resolve(true));
            });
        });
        expect(loadWasFirst).toBeTruthy();
    }, 20000);
```

birkskyum:
Closes #2551 

Captures the initial resize event from the resize observer, so it's more similar to the previous implementation that didn't fire on setup.

Adds the browser test that checks that the 'load' event comes before 'resize' and 'moveend'

## Launch Checklist

### DIFF
--- a/src/ui/map.test.ts
+++ b/src/ui/map.test.ts
@@ -803,7 +803,7 @@ describe('Map', () => {
             expect(spyC).not.toHaveBeenCalled();
         });
 
-        test('do resize if trackResize is true (default)', () => {
+        test('do resize if trackResize is true (default)', async () => {
             let observerCallback: Function = null;
             global.ResizeObserver = jest.fn().mockImplementation((c) => ({
                 observe: () => { observerCallback = c; }
@@ -814,6 +814,7 @@ describe('Map', () => {
             const spyA = jest.spyOn(map, '_update');
             const spyB = jest.spyOn(map, 'resize');
 
+            observerCallback(); // The initial "observe" event fired by ResizeObserver is muted in the map constructor
             observerCallback();
 
             expect(spyA).toHaveBeenCalled();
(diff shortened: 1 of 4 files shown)


HarelM:
Seems like our unit tests are covering is 😀
I guess there's a need to add another call to trigger the observer. But this test can remain to make sure resize doesn't get called on first event.

birkskyum:
Updated the unit test to observe twice

src/ui/map.test.ts
```diff
@@ -814,6 +814,10 @@ describe('Map', () => {
             const spyA = jest.spyOn(map, '_update');
             const spyB = jest.spyOn(map, 'resize');
 
+            // Call twice, because the initial "observe" event fired by ResizeObserver
+            // is muted in the map constructor
+
+            observerCallback();
```
HarelM, line 820: Can you add another test or and expectations between the first and second callback?

birkskyum: Great idea, I added these so it now says:

```ts
            // The initial "observe" event fired by ResizeObserver should be captured/muted
            // in the map constructor

            observerCallback();
            expect(spyA).not.toHaveBeenCalled();
            expect(spyB).not.toHaveBeenCalled();

            // Following "observe" events should fire a resize / _update

            observerCallback();
            expect(spyA).toHaveBeenCalled();
            expect(spyB).toHaveBeenCalled();
            ```

HarelM (approved):

PR #2581, opened 2023-05-23
Expose map options.maxTileCacheZoomLevels to allow better control of cache

pramilk:
Currently map options has maxTileCacheSize option to control the max number of tiles to cache. This has following limitations
1. Raster tiles are typically 256 vs Vector are 512.
2. User might have different screen sizes
3. Map size might change
Because of above 3 scenario, its difficult to set a maxTileCacheSize value.

Currently control cache 5 zoom level of tiles by default which is quite high. Each vector tile takes about 500KB of memory. 

**Pending**: Will add unit tests, if the change looks good.

We ran a A/B test to see its impact when 1/2/5 Zoom worth of tiles are cached and here is the result
Control: Default 5 Zoom level
Treatment (TileCache1v9):  1 Zoom level (Left Column)
Treatment (TileCache2v9):  2 Zoom level (Right Column)

image

Notes about metrics:
1. Memory Usage: Its a value reported by performance.memory.usedJSHeapSize immediately after initial load
2. Long session Memory: Its a value reported by performance.memory.usedJSHeapSize after 21 min of user session. Browser only allow reporting updated data after 20 min of first call to performance.memory.usedJSHeapSize
3. Zoom/Pan FPS: Number of animation frame between move start and end.
4. Zoom/Pan TTG: Time between move end and Idle event.

Also, current default 5 zoom is quite high, if you agree we should consider decreasing it to a low value may be 3 or 4.

## Launch Checklist

 - [ ] Link to related issues.
 - [x] Post benchmark scores.


src/source/source_cache.ts
```diff
@@ -445,7 +448,7 @@ class SourceCache extends Evented {
         const widthInTiles = Math.ceil(transform.width / this._source.tileSize) + 1;
         const heightInTiles = Math.ceil(transform.height / this._source.tileSize) + 1;
         const approxTilesInView = widthInTiles * heightInTiles;
-        const commonZoomRange = 5;
+        const commonZoomRange = typeof this._maxTileCacheZoomLevels === 'number' ? this._maxTileCacheZoomLevels : 5;
```
HarelM, line 451: Can't we simply initialize it to 5 instead of this check?

HarelM:
Can you clarify the results? How much memory was saved compared to performance degradation...?

src/source/source_cache.ts
pramilk: Yes, its not needed. Updated.

pramilk:
> Can you clarify the results? How much memory was saved compared to performance degradation...?

I have only tested 1/2 vs default value of 5 zoom levels.  Since I don't have data for 3/4 zoom levels, we can keep the current default of 5 zoom levels. I will run additional A/B test to see if 3/4 has memory usage benefits without impacting zoom/pan performance. 

Regarding current tests, memory usage with 1 zoom level of caching showed a 22% improvement in memory usage but it had some zoom/pan performance implications. 2 zoom level of caching showed 14.8% improvement in memory usage.

Our use case, is to set this property to 2 or 3 for low end devices and will leave it to default value for faster devices. Current option.maxTileCacheSize has a bunch of challenges as pointed out in this PR description.

src/source/source_cache.ts
```diff
@@ -445,9 +448,8 @@ class SourceCache extends Evented {
         const widthInTiles = Math.ceil(transform.width / this._source.tileSize) + 1;
         const heightInTiles = Math.ceil(transform.height / this._source.tileSize) + 1;
         const approxTilesInView = widthInTiles * heightInTiles;
-        const commonZoomRange = 5;
 
-        const viewDependentMaxSize = Math.floor(approxTilesInView * commonZoomRange);
+        const viewDependentMaxSize = Math.floor(approxTilesInView * this._maxTileCacheZoomLevels);
```
HarelM, line 452: Is there a way for this to be null?
In previous implementation this wasn't possible, but I'm not so confident about this new implementation.

HarelM:
Can you please add a unit test that makes sure they value won't be null?

HarelM:
Send like a unit test is rightfully failing, maybe fixing it would be enough.

src/source/source_cache.ts
pramilk: updated.

HarelM (approved):

PR #2593, opened 2023-05-25
Document that dblclick is preceded by click events

dabreegster:
I spent at least a few minutes puzzling through the fact that `dblclick` handlers don't magically suppress `click` events. It makes sense in hindsight, but I want to try and save the next person a little bit of time. See https://osmus.slack.com/archives/C01G3D28DAB/p1685016451984109 for some discussion.

## Launch Checklist

### DIFF
--- a/src/ui/events.ts
+++ b/src/ui/events.ts
@@ -564,6 +564,8 @@ export type MapEvent =
      * If `layerId` is included as the second argument in {@link Map#on}, the event listener will fire only
      * when the point that is clicked twice contains a visible portion of the specified layer.
      *
+     * **Note:** Two {@link Map.event:click} events will precede this event.
+     *
      * @event dblclick
      * @memberof Map
      * @instance
--- a/src/ui/events.ts
+++ b/src/ui/events.ts
@@ -564,7 +564,7 @@ export type MapEvent =
      * If `layerId` is included as the second argument in {@link Map#on}, the event listener will fire only
      * when the point that is clicked twice contains a visible portion of the specified layer.
      *
-     * **Note:** Two {@link Map.event:click} events will precede this event.
+     * **Note:** Under normal conditions, this event will be preceded by two {@link Map.event:click} events.
      *
      * @event dblclick
      * @memberof Map


IvanSanchez:
@dabreegster Have you seen "Getting Touchy"? ( https://archive.org/details/Patrick_H_Lauke_Getting_Touchy_FOWA_London_24.10.2013 ) - it explains that the order of mouse/pointer events is unreliable, so actually there might be cases where the second `click` event might trigger after the `dblclick` event.

(I might have spent a bit too much time worrying about the `dblclick` emulation in Leaflet)

dabreegster:
No... that's quite upsetting, if this is still true in modern browsers. :\

If this is the case, then the wording should be something closer to "Two {@link Map.event:click} events will also fire along with this event. The order is undefined."

IvanSanchez:
> [...] if this is still true in modern browsers. :\

Modern evergreen browsers *should* have reliable behaviour.

Unfortunately, the behaviour of `dblclick` events **is underspecified in the DOM specs** (see https://w3c.github.io/uievents/#event-type-dblclick). See how the specs allow for a `mousedown`→`mouseup`→`click`→`mousedown`→`mouseup`→`dblclick`→`click` sequence of events. There will be also edge cases where the user moves the pointer 1px between the first and the second click and the OS might still fire a `dblclick` event, but each of the `click`s might have gone to individual DOM elements. And touchscreen delays are *fun*.

I would suggest chaning the wording to `Under normal conditions, this event will be preceded by two {@link Map.event:click} events.`

HarelM:
The following comment probably won't help, but I'll write it anyway:
From a UX point of view, double click are discouraged since they affect accessibility in a negative way, so the guidelines are not to use it.
As for right click, this is also not a great experience in browsers, especially for sites that are also accessible from mobile devices.
I have designed an entire site based on single (left) click only. It's a great experience for the users as they don't need to guess if there are places that supports other types of clicks.
You'll need to be creative at some points, but when you start thinking mobile first UX you create a far easier and intuitive app.

Having said that, I don't mind updating the docs. 😀

dabreegster:
> the behaviour of dblclick events is underspecified in the DOM specs

Thanks for the reference to this! Indeed quite unfortunate. I'll dig through other drawing libraries like https://github.com/JamesLMilner/terra-draw to see if they have a nice workaround. In the meantime, I've changed the wording to your suggestion.

> From a UX point of view, double click are discouraged since they affect accessibility in a negative way, so the guidelines are not to use it.

I can see the reasons for that. In my particular case, several participants in a UX study expected double click to make a drawing tool finish, because QGIS and other tools they're used to do that. It's not the only way to make that tool finish, but the others we put in (a button or pressing enter) weren't as obvious to them! We'll keep thinking through alternatives.

HarelM (approved):

Issue #2590, opened 2023-05-25
Firefox: Heatmap doesn't appear on the map since 3.0.0


FloChehab:
Hello,
I found it weird that the heatmap was not working in 3.0.0, after debugging as far as I can (I don't speak webgl), I have identified what seems to be a Firefox specifc bug related to the heatmap.

**maplibre-gl-js version**: 3.0.0 **(to be more specific, I have identified locally that the bug appears since https://github.com/maplibre/maplibre-gl-js/commit/7a1d5b5355f9b1da31ced9d6f1731b9828f8e057 from  @rotu )**

**browser**: Firefox 113.0.1

### Steps to Trigger Behavior

N.A.

### Link to Demonstration

Go to https://maplibre.org/maplibre-gl-js-docs/example/heatmap-layer/ with Firefox

### Expected Behavior

The heatmap is shown.

### Actual Behavior

The heatmap is not shown and I have those warnings in my console, if this can help.

![image](https://github.com/maplibre/maplibre-gl-js/assets/20242665/f0572b48-a644-400d-ada1-15a32526a7a3)

birkskyum:
Thanks for reporting this @FloChehab , I can reproduce this too and have posted a fix for it in #2595 . It would appear that after the migration to WebGL2 and addition extension for these color types is needed.

FloChehab:
Awesome @birkskyum, can't wait to use that maplibregljs in prod :fire:

FloChehab:
@birkskyum Any idea regarding when that fix will be officially out ? (3.0.1?)

birkskyum:
@FloChehab , probably in an hour

birkskyum:
@FloChehab , it's 3.0.1  - should be on npm by now.

birkskyum:
Let me know if it resolves your issue

FloChehab:
Everything looks fine @birkskyum, awesome, thanks a lot!

PR #2595, opened 2023-05-25
Load float webgl2 extension to support i.e. RGBA16F

birkskyum:
Fixes https://github.com/maplibre/maplibre-gl-js/issues/2590 

Adds the `gl.getExtension('EXT_color_buffer_float');` to support [all these types](https://developer.mozilla.org/en-US/docs/Web/API/EXT_color_buffer_float).


birkskyum:
The heatmap example can be used to verify the fix

HarelM (approved):

PR #2596, opened 2023-05-26
migrate background-pattern to glsl 3

### LINKED ISSUE #2576: Allow the use of GLSL 3.0 shaders for gradual migration
In order to gain the performance benefits WebGL2 has to offer through ie UBOs, the first thing that has to happen is updating all the shaders to [GLSL ES 3.0](https://registry.khronos.org/OpenGL/specs/es/3.0/GLSL_ES_Specification_3.00.pdf).

Some initial work has to happen to the shader compilation setup before the approx 60 shaders can be ported, and that initial work is what this ticket is all about.

We currently have a GLSL 1.0 _prelude, and some checks for GLSL 1.0 `attributes` etc. in the boilerplate, that makes it hard to do a gradual shader migration.

The aim of this ticket is to make our boilerplate (fragment/vertex compilation, program linking, prelude etc.) support GLSL 3.0 _and_ GLSL 1.0 shaders at the same time, so that we can migrate the shaders little by little instead of a big swap.

This is a prerequisite for actually porting the shaders  ( see #2562 )

In order to complete this ticket, it should be possible to add `#version 300 es` in the top of a shader and get access to GLSL 3.0 in that particular shader, while still allowing the remaining GLSL 1.0 shaders to compile and link.

It's possible to achieve this by modifying shaders.ts and program.ts - no changes to the _prelude's should be needed.

birkskyum:
This code contains the port of 1 shader pair, the **background-pattern** to GLSL 3.0, to prove the feasibility of a migration. When going through the rest of the shaders, there will likely surface more adjustments.

Closes #2576 

### DIFF
--- a/src/render/program.ts
+++ b/src/render/program.ts
@@ -18,6 +18,14 @@ import Terrain from '../render/terrain';
 
 export type DrawMode = typeof WebGL2RenderingContext.LINES | typeof WebGL2RenderingContext.TRIANGLES | typeof WebGL2RenderingContext.LINE_STRIP;
 
+function isGLSL3Shader(input: string):boolean {
+    return input.split('\n')[0] === '#version 300 es';
+}
+
+function removeFirstLine(input: string):string {
+    return input.substring(input.indexOf('\n') + 1);
+}
+
 function getTokenizedAttributesAndUniforms(array: Array<string>): Array<string> {
     const result = [];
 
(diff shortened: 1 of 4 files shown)


HarelM:
Interesting, thanks!!

HarelM:
Let's make sure we open an issue to clean up the "is 3" at the end so we won't forget.

HarelM (approved):

PR #2603, opened 2023-05-28
Try close and open the page when retrying.

HarelM:
I've seen that the retry doesn't really help, so I added some close and open a page.
For my environment locally it seems to solve one of the tests.
Let's monitor this and see if more improvements is needed there...


HarelM:
I did a retry, let's see if the tests still pass...

birkskyum:
It seems like the "["sleep", "1000"]" make a positive difference

birkskyum (approved):

birkskyum:
3 full runs without breaking is more than I've seen for a while, so it appears that this is helping

PR #2623, opened 2023-06-02
Add strict typecheck to generated d.ts file

### LINKED ISSUE #2588: Error at build

HarelM:
Fixes #2588.

This adds a tsconfig file that is specific to the `d.ts` file in the dist folder in order to make sure that the generated file is following the strict mode.

### DIFF
--- a/.github/workflows/test-all.yml
+++ b/.github/workflows/test-all.yml
@@ -27,7 +27,7 @@ jobs:
         if: success() || failure()
       - run: npm run lint-docs
         if: success() || failure()
-      - run: npm generate-typings
+      - run: npm run generate-typings
         if: success() || failure()
       - run: npm run typecheck
         if: success() || failure()
--- a/.github/workflows/test-all.yml
+++ b/.github/workflows/test-all.yml
@@ -27,6 +27,8 @@ jobs:
         if: success() || failure()
       - run: npm run lint-docs
         if: success() || failure()
+      - run: npm generate-typings
+        if: success() || failure()
       - run: npm run typecheck
         if: success() || failure()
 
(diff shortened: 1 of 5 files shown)


HarelM:
I've added the following comment to the dts-bundle-generator repo to suggest a way to do it without this "hack".
https://github.com/timocov/dts-bundle-generator/issues/187#issuecomment-1574070995

neodescis:
I like this approach! However, I noticed that the typecheck commands are not excluding the "staging" folder. I'm guessing we should exclude that folder in the base tsconfig?

HarelM:
Yes, staging should be ignored, but when I added it to the exclude list it didn't work, not sure why...

birkskyum (approved):

PR #2655, opened 2023-06-09
Inform better if shaders doesn't compile

birkskyum:
If for some reason the shaders doesn't compile, or the program doesn't link correctly, we should throw an informative message.


HarelM:
Seems like some tests are covering the compilation stuff, can you add tests to cover the other exceptions?

birkskyum:
Interestingly enough, this change actually break 6 unit test suites that are attempting (and failing) to compile webgl2 shaders that they don't need in order to check the logic they test for.

HarelM:
@birkskyum where are we with this PR? Do you want to push it forward?

birkskyum:
I'll have a look at rebasing it, as it should be compatible with the current state of the repo.

HarelM:
Any updates on this?
Should this be closed or do you plan to work on it?

birkskyum:
I'll pick it up again when we find a way to sunset webgl1 for good - unit tests and all

birkskyum:
These three small additions are all there is to it. Resolved the merge conflict

HarelM:
Do we want to use some flag to allow skipping this?
In any case, this is approved.

HarelM (approved):

birkskyum:
I don't think a flag is needed here, because the compute of the status has already happened and this just prints it if there is an issue.

PR #2656, opened 2023-06-10
Enable webgl1 fallback

birkskyum:
This is a fix to the regression caused by the complete sunset of webgl1. This will be discussed in the TSC meeting Wed June 14th.

This reverts shaders to GL ES 1.0, and adds conditional logic and union typings to allow webgl1 as a fallback while keeping the majority of users on webgl2.


src/gl/context.ts
```diff
@@ -101,18 +102,37 @@ class Context {
...
-        gl.getExtension('EXT_color_buffer_float');
-        this.RGBA16F = gl.RGBA16F;
-        this.RGB16F = gl.RGB16F;
+        const extSRGB = gl.getExtension('EXT_sRGB');
+        if (!extSRGB) {
+            console.warn('EXT_sRGB extension not supported');
+        }
+
+        if (isWebGL2(gl)) {
+            console.log('WEBGL2');
```
HarelM, line 123: `console.log` is probably not needed...

src/render/draw_heatmap.ts
```diff
@@ -16,6 +16,7 @@ import type SourceCache from '../source/source_cache';
 import type HeatmapStyleLayer from '../style/style_layer/heatmap_style_layer';
 import type HeatmapBucket from '../data/bucket/heatmap_bucket';
 import type {OverscaledTileID} from '../source/tile_id';
+import {isWebGL2} from '../gl/webgl2';
```
HarelM, line 19: Not needed?

HarelM:
Is there a way to keep the new shaders code and use the rollup to translate it to "old" code?
So that if and when we move to the new shaders we would only need to remove this part?
This is a rough thought, I don't know how possible it is.

birkskyum:
> Is there a way to keep the new shaders code and use the rollup to translate it to "old" code? So that if and when we move to the new shaders we would only need to remove this part? This is a rough thought, I don't know how possible it is.

There might be a away by introducing a lot of find/replace logic, but there is already a lot of find/replace logic with all the "pragma", and adding more would just add complexity. I think we might as well just keep the commit before this PR, and the changes in this PR, as a reference to make it trivial to get to this point again... the planned refactors of shaders/logic that would have most impact on benchmarks are not possible within the limitations of webgl1 anyway.

HarelM:
LGTM! Thanks! Yea, this is what I meant.

HarelM (approved):

HandyMenny:
the downgrade is done at build time, right? what about downgrading at runtime? Indeed that would be a (small?) penality for webgl1 users, but it clearly states that webgl2 is the path that maplibre is optimizing for.

birkskyum:
@HandyMenny , yes, this is done a build time. It's an interesting idea to consider runtime transpiring - i'll bring that suggestion to the tsc meeting discussion as well.

Issue #2650, opened 2023-06-08
High pixel ratio causes distortions


HandyMenny:
Hi, I have noticed that under certain conditions the map is distorted, specifically stretched, translated or over-zoomed.
Thi issue occurs with mobile devices with very high resolutions e.g. 2k, 4k. But it's also present with devices with modest resolutions (so-called FHD+) in "desktop site" mode. 
I then started experimenting with `options.pixelRatio` and by increasing it (artificially) the problem appeared on more devices (including desktops), by limiting it (to 1-2) the problem disappeared.

Note: I was only able to reproduce this easily on Chrome and Edge, with Firefox it was more difficult, I only achieved some translation with values close to 20 which are completely unrealistic.

### Steps to Trigger Behavior

 1. Create a maplibre map which occupies the entire viewport
 2. Set the pixelratio option to a very high value (I suggest >= 3.5 for mobile, >= 7 for desktop)
 3. Interact with the map and see if it's distorted.

### Link to Demonstration
You can see a demo here: https://stackblitz.com/edit/maplibre-pixelratio
To maximise the effect, open the preview in a new tab and enable the 'desktop site' option (on mobile).

### Screenshots

![image](https://github.com/maplibre/maplibre-gl-js/assets/3463028/c3133c93-5506-4901-9512-9d82de623c5d)
![image](https://github.com/maplibre/maplibre-gl-js/assets/3463028/97914ffa-d37d-4ed0-b099-b760f4823929)
![image](https://github.com/maplibre/maplibre-gl-js/assets/3463028/a94f3d14-a9b1-4edb-9a27-a1866851ed79)

### Environment info

**maplibre-gl-js version**: 3.0.1 and 2.4.0

**Browser**: Edge 114 and Chrome 114

**OS**: Android and Windows

HandyMenny:
As a followup, one of these conditions is to run MapLibre with Leaflet (using maplibre-gl-leaflet bindings) on a retina browser. 
In that case, leaflet doubles the size of the canvas, thus effectively doubling the pixel ratio value. 
So actually, ignoring the "desktop site" mode, this issue doesn't impact 2k mobile devices, but may impact 4k devices with higher dpi than the default.

HandyMenny:
closed by #2674

PR #2657, opened 2023-06-10
Reduce pixel ratio when hitting drawing buffer limit

HandyMenny:
This is an attempt to address #2650.

According to [specification](https://registry.khronos.org/webgl/specs/latest/1.0/#5.14.1), if the implementation is unable to satisfy the requested width or height, **drawing buffer size** may be different from **canvas size**. When this happens, the map may be distorted as described in #2650.

A way to workaround this issue is to reduce pixel ratio when canvas size and drawing buffer size diverge.
This is the same approach used by luma.gl:
https://github.com/visgl/luma.gl/blob/39394e419b879ea714acfb7305d50b490aee7c90/modules/api/src/adapter/canvas-context.ts#L256-L265
And so as a side effect compatibility with deck.gl is improved.

The main caveats of this approach are:
1. This check must be performed each time the canvas size changes, or at least each time the height, width or their multiplication increases. 
To achieve this, I added it to the map.resize() function, but I'm not so sure that all resizes happens there (I "fixed" one in this pull request). If there are better places let me know.
3. This workaround triggers a canvas resize, so it can cause loops. I have added a "basic protection" with a flag, but if it's concerning there are certainly better solutions. For example, trigger the check only if the canvas grows and apply the reduce factor only if it really is a reduce.
4. The original pixel ratio is lost, so the pixel ratio will not return to its original value if within the limits (for example after a window resize).
5. GL mocks must update the drawing buffer size each time the canvas changes (implemented in this pull request).

You can find a demo here: https://stackblitz.com/edit/maplibre-pixelratio-reduce

### DIFF
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -718,12 +718,8 @@ class Map extends Camera {
      * @param {number} pixelRatio The pixel ratio.
      */
     setPixelRatio(pixelRatio: number) {
-        const [width, height] = this._containerDimensions();
-
         this._pixelRatio = pixelRatio;
-
-        this._resizeCanvas(width, height, pixelRatio);
-        this.painter.resize(width, height, pixelRatio);
+        this.resize();
     }
 
     /**
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -687,6 +687,7 @@ class Map extends Camera {
         this.transform.resize(width, height);
         this._requestedCameraState?.resize(width, height);
         this.painter.resize(width, height, this.getPixelRatio());
+        this._reducePixelRatioIfNeeded();
 
         const fireMoving = !this._moving;
         if (fireMoving) {
@@ -723,6 +724,19 @@ class Map extends Camera {
 
         this._resizeCanvas(width, height, pixelRatio);
         this.painter.resize(width, height, pixelRatio);
+    }
+
+    /**
+     * Check if we have hit client limit for drawing buffer width or height.
+     * In those case they get clamped causing distortions. To avoid that, reduce pixel ratio.
+     */
+    _reducePixelRatioIfNeeded() {
+        const {width, height} = this._canvas;
+        const {drawingBufferWidth, drawingBufferHeight} = this.painter.context.gl;
+        if (drawingBufferWidth !== width || drawingBufferHeight !== height) {
+            const reduceFactor = Math.min(drawingBufferWidth / width, drawingBufferHeight / height);
+            this.setPixelRatio(this.getPixelRatio() * reduceFactor);
+        }
     }
 
     /**
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -671,6 +671,8 @@ class Map extends Camera {
      * @param eventData Additional properties to be passed to `movestart`, `move`, `resize`, and `moveend`
      * events that get triggered as a result of resize. This can be useful for differentiating the
      * source of an event (for example, user-initiated or programmatically-triggered events).
+     * @param disablePixelRatioReduce Pixel Ratio will automatically be reduced if we have hit drawing buffer limit,
+     * setting this to true will disable this behaviour.
      * @returns {Map} `this`
      * @example
      * // Resize the map when the map container is shown
@@ -678,7 +680,7 @@ class Map extends Camera {
      * var mapDiv = document.getElementById('map');
      * if (mapDiv.style.visibility === true) map.resize();
      */
-    resize(eventData?: any) {
+    resize(eventData?: any, disablePixelRatioReduce?: Boolean) {
         const dimensions = this._containerDimensions();
         const width = dimensions[0];
         const height = dimensions[1];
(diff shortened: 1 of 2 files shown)


HarelM:
Can you add a render test to show how this would look before and after the fix?

HarelM:
Thanks for taking the time to fix this issue!

HandyMenny:
> Can you add a render test to show how this would look before and after the fix?

I did try, but without success. For images above 2048x2048 puppeter just crashes. I tried adding options, disabling headless mode, adding debugging, but no luck... After the rendering is complete chrome closes and I get this in the console: 

```
1/1: errored tests\pixel-ratio\veryhigh ProtocolError: Protocol error (Runtime.callFunctionOn): Internal error
    at new Callback (file:///Github/maplibre-gl-js/node_modules/puppeteer-core/src/common/Connection.ts:66:12)
    at CallbackRegistry.create (file:///Github/maplibre-gl-js/node_modules/puppeteer-core/src/common/Connection.ts:127:22)
    at Connection._rawSend (file:///Github/maplibre-gl-js/node_modules/puppeteer-core/src/common/Connection.ts:278:22)
    at CDPSessionImpl.send (file:///Github/maplibre-gl-js/node_modules/puppeteer-core/src/common/Connection.ts:537:29)
    at ExecutionContext._ExecutionContext_evaluate (file:///Github/maplibre-gl-js/node_modules/puppeteer-core/src/common/ExecutionContext.ts:314:44)      
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async ExecutionContext.evaluate (file:///Github/maplibre-gl-js/node_modules/puppeteer-core/src/common/ExecutionContext.ts:189:12)
    at async getImageFromStyle (file:///Github/maplibre-gl-js/test/integration/render/run_render_tests.ts:284:28)
    at async runTests (file:///Github/maplibre-gl-js/test/integration/render/run_render_tests.ts:757:26)
    at async executeRenderTests (file:///Github/maplibre-gl-js/test/integration/render/run_render_tests.ts:822:5)
```

this is the test I was trying to add: https://gist.github.com/HandyMenny/25c2270c1992853385528c2571e76c67

pixelratio 33 because texture max size is 16834 in the testing environment (on my pc it's 8192 and on my smartphone it's 4096)

HarelM:
You say that the issue happens even if the pixel ratio is not that high, can't this be reproduced in a "simpler" test? i.e. with pixel ratio that is lower, say 5 or 10?

HandyMenny:
> You say that the issue happens even if the pixel ratio is not that high, can't this be reproduced in a "simpler" test? i.e. with pixel ratio that is lower, say 5 or 10?

yes, but chrome mobile has a hard limit that is 1/4 of this test environment*

What I was trying now is to transfer only part of the image, that should be ok?


\* Actually mobile devices seem to limit even more, this is the viewport of one of the devices that has this problem, S23 with minimum zoom:
image

HandyMenny:
I understood why I could trigger it so easily, among other tests I was also using maplibre in conjunction with leaflet, leaflet was duplicating the canvas size on retina browsers, so a 3.5x was actually a 7x

HandyMenny:
Added a render test, as explained in the commit message, width and pixel ratio were chosen to exceed 16834px. 
As I anticipated in the other messages, I came up with a way to limit the size of the resulting PNG. I have added two properties, reportWidth and reportHeight to the test data interface, these have priority over the rendering dimensions

HarelM:
THANKS!!
Can you place the before the fix and after images of the test in the initial PR so that someone in the future that will read this PR would have an easy time understanding this change?

Also please add a changelog entry.

src/ui/map.ts
```diff
@@ -671,14 +671,16 @@ class Map extends Camera {
      * @param eventData Additional properties to be passed to `movestart`, `move`, `resize`, and `moveend`
      * events that get triggered as a result of resize. This can be useful for differentiating the
      * source of an event (for example, user-initiated or programmatically-triggered events).
+     * @param disablePixelRatioReduce Pixel Ratio will automatically be reduced if we have hit drawing buffer limit,
```
HarelM, line 674: What's the default?

HandyMenny: It's enabled by default, I've updated the description to clarify this.
Another approach, that doesn't cause loops, would be to cap canvas size to gl.max_texture_size.
It isn't very clear to me how these limitations works... I mean I had 8192 until yesterday, now I have 16834 and nothing has changed on my PC. But checking each time that value should be safe.
Another idea might be to cap canvas height/width to 4096px, but this might impact users who want to use this library to make prints.

HandyMenny:
> Can you place the before the fix and after images of the test in the initial PR [...]?
> Also please add a changelog entry.

Done. The problem is that the test fails if MAX_TEXTURE_SIZE is different from 8192 (I updated it, locally I had 16834, but in the CI it's 8192)

HandyMenny:
@HarelM I let my users play (literrally) with a maplibre stack and it turns out that a fixed pixelratio is a bad idea (open maplibre.org, set zoom to 15%, reload and set zoom to 100%) and that being close to max_texture_size can cause crashes/lags in browser or OS.

So I'll close this, sorry for wasting your time. I'll get back with a better approach (and probably a new issue).
In the meantime I can extract "Resize map after setting pixel ratio" from this pull, if you are OK with that change.

HarelM:
Sure, no worries 😀 the more people that knows the code, the better we are.

PR #2671, opened 2023-06-12
Set worker count to 2 for Safari browser

HandyMenny:
After merging #2354 workercount is set to half the number of CPU cores capped at 3 for Safari.

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/hardwareConcurrency) and [Can I Use](https://caniuse.com/hardwareconcurrency) data, all versions of Safari except 10.1 don't support navigator.hardwareConcurrency.
When hardwareConcurrency isn't available the current logic sets 2. 
So just use 2 for Safari and get rid of all that logic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 - Fix regression - Add webgl1 fallback to accomondate users without webgl2 support
 - _...Add new stuff here..._
 
+### ⚠️ Potentially breaking changes
+- Worker count statically set to 2 for Safari browser as it doesn't support by default.
+- _...Add new stuff here..._
+
 ## 3.0.1
 
 ### ✨ Features and improvements
--- a/src/index.ts
+++ b/src/index.ts
@@ -109,7 +109,7 @@ const exported = {
 
     /**
      * Gets and sets the number of web workers instantiated on a page with GL JS maps.
-     * By default, workerCount is 1 except for Safari browser where it is set to half the number of CPU cores (capped at 3).
+     * By default, workerCount is 1 except for Safari browser where it is set to 2.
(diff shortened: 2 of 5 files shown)


HarelM:
@pramik can you please comment on this PR?

HandyMenny:
The change was made under the assumption that MDN and Can I use were up-to-date, in reality they were not. I tried a few versions of Safari and it indeed disappeared with Safari 11, but then returned with Safari 15.x (I think 15.4 or 15.5).
It was added to webkit with this change: https://bugs.webkit.org/show_bug.cgi?id=233381
And looking at the code it returns 4 for devices with < 8 cores and 8 for those >= 8 cores https://github.com/WebKit/WebKit/blob/708f1326df9c5680178a0ff89e9cb13fd0a5f79e/Source/WebCore/page/NavigatorBase.cpp#L179
So, translated (ignoring Safari 10.1), workercount = 2 for Safari clients that don't support hardwareConcurrency or have less than 8 cores, workercount 3 for Safari clients that have 8 cores or more.

Although I like the idea of simplifying the logic (min, max, half...), Safari actually supports it so at the very least the descriptions and changelog should be updated (if you like the idea too)

pramilk:
I am fine caping it to 2 instead of current 3. Just the experimentation data that we got showed best performance when we caped it to 3. 

image

HandyMenny:
@pramilk Thanks, it would be interesting to know how many of those safari sessions actually supported hardwareConcurrency (the first versions to support it again should date back to April or May 2022). 
But since the assumptions of this pull were found to be incorrect, this pull request can be closed.

PR #294, opened 2021-09-02
[Typescript] Remove the usage of devicePixelRatio as a getter from browser.ts

fredj:
fixes #265

PR #2674, opened 2023-06-13
Introduce max canvas size option and automatically limit pixel ratio

HandyMenny:
This supersedes #2657 and addresses the issue #2650.

## The issue

Browsers have limits on how many pixels they can render through webgl. One of these limits is `MAX_TEXTURE_SIZE`, a hard limit on the width or height of the drawing buffer.
Exceeding these limits results in the canvas size no longer corresponding to the size of the underlaying drawing buffer. This can cause distortions (stretching, over-zooming, translations).

## Affected clients

Devices where this issue might be noticeable:

- 4k mobile device with DPI higher than default
- 2k mobile devices using _Maplibre GL Leaflet_ with a generous padding (e.g. 0.2-0.35) 
- FHD mobile devices in _desktop site_ mode
- Desktop devices while zooming out (the entire viewport) with a fixed pixel ratio (the default since #294)

Basically, the higher the resolution and pixelratio/dpi the more noticeable this problem is.

## Example of distortion

I've added a render test with a very high resolution, that test only pass if canvasMaxSize and GL.MAX_TEXTURE_SIZE are 8192.
<details>
    <summary>Result before these changes</summary>

before-fix

</details>
<details>
    <summary>Result after these changes</summary>

after-fix

</details>

## Proposed Solution

The proposed solution consists of two parts:
1. Introduce an option, maxCanvasSize, to limit the maximum width and height of the canvas. This limit is applied by dynamically limiting the pixel ratio. This is an internal thing, the value of getPixelRatio() will not change. The default value is 4096 as MAX_TEXTURE_SIZE is typically at least 4096px. 
2. At each resize check if drawing buffer size is the one requested, otherwise we scale down pixel ratio further.

Why both?
The former isn't enough to cover all cases, with a very high maxCanvasSize other limits can be exceeded.
The latter is enough to cover all cases but its being "trial and error" makes it not easily testable. Furthermore, IMHO we should never request such large drawing buffers, they can cause hangs or even crashes.

## List of related changes

- Round down painter and canvas dimensions so they're consistent
- Resize map after setting pixel ratio
- rendering test: implement cropping of png result
- mock webgl: update drawing buffer size when viewport changes

Let me know if they need to be mentioned in changelog or need a separate pull request.

## Checklist

 - [x] Link to related issues.


HandyMenny:
Draft because I have not yet added tests for all the changes made and have not added lines to the changelog.
But I would appreciate feedback on the proposed solution in the meantime

src/ui/map.ts
```diff
@@ -117,6 +117,7 @@ export type MapOptions = {
     pitchWithRotate?: boolean;
     pixelRatio?: number;
     validateStyle?: boolean;
+    maxCanvasSize?: number;
```
HarelM, line 120: I would expect this to have two dimensions, maybe `maxCanvasBufferSize`? IDK if this the right term...

src/ui/map.ts
```diff
@@ -683,10 +691,21 @@ class Map extends Camera {
...
+        let clampedPixelRatio = this._getClampedPixelRatio(width, height, this._maxCanvasSize);
+        this._resizeCanvas(width, height, clampedPixelRatio);
+        this.painter.resize(width, height, clampedPixelRatio);
+
+        // check if we've reached GL limits, in that case further clamps pixelRatio
+        if (this.painter.overLimit()) {
+            const gl = this.painter.context.gl;
+            const maxSize = Math.max(gl.drawingBufferWidth, gl.drawingBufferHeight);
+            clampedPixelRatio = this._getClampedPixelRatio(width, height, maxSize);
+            this._resizeCanvas(width, height, clampedPixelRatio);
```
HarelM, line 703: Seems like we are resizing twice in this case, is there a way to know that up front and resize only once?

src/ui/map.ts
```diff
@@ -159,6 +160,9 @@ const defaultMaxPitch = 60;
 // use this variable to check maxPitch for validity
 const maxPitchThreshold = 85;
 
+// GL MAX_TEXTURE_SIZE is usually at least 4096px
+const defaultMaxCanvasSize = 4096;
```
HarelM, line 164: Do we actually need this const? Seems like we are using it only as part of the default options.

HarelM:
I've looked at the before and after visuals of this PR and comparing them by opening two browser tabs and switching allowed me to see the stretch gone from the before to the after image.
If possible, and I know this might be annoying to do, please create a render test that should draw a perfect square after the fix compared to a rectangle before the fix, I think it can really emphasize the fix well.
THANKS!!

src/ui/map.ts
HandyMenny: I can change it to have two dimensions, I put one because MAX_TEXTURE_SIZE has one.
I think the name is correct, what we do is limit the maximum size of the canvas to prevent the browser from limiting the size of the drawing buffer, but it can also be used for other purposes, for example I don't want to have the map rendered at 4k on mobile devices.

src/ui/map.ts
HandyMenny: Unfortunately no, I found no easily usable limits other than MAX_TEXTURE_SIZE. 
During the various tests I've done, I've noticed that some browsers limit the canvas area to ~32-33 million pixels. it's a very large value but with maxCanvasSize = MAX_TEXTURE_SIZE = 16834 (which should be "safe") and zomming out the window (with a fixed pixel ratio) it can be exceeded. In this case the aspect ratio remains almost intact, the problem is that if you have other overlays on top of maplibre they will misalign.

src/ui/map.ts
HandyMenny: I saw other const there and added this one too, it's quite useless yes

HandyMenny:
I have updated this pull: 
- rebased on top of main
- reorganised the commits
- added unit tests for maxCanvasSize, gl 'overlimit' and canvas size rounding.
- updated the changelog
- modified the rendertest to draw a black square as suggested
- made maxCanvasSize bi-dimensional and removed the unnecessary const

src/ui/map.test.ts
```diff
@@ -858,6 +858,19 @@ describe('Map', () => {
             expect(spyB).toHaveBeenCalled();
         });
 
+        test('width and height correctly rounded', () => {
```
HarelM, line 861: Shouldn't this test be in `painter` rather than `map`?

src/ui/map.ts
```diff
@@ -117,6 +117,7 @@ export type MapOptions = {
     pitchWithRotate?: boolean;
     pixelRatio?: number;
     validateStyle?: boolean;
+    maxCanvasSize?: number[];
```
HarelM, line 120: `[number, number]`?

src/ui/map.ts
```diff
@@ -202,7 +203,9 @@ const defaultOptions = {
     transformCameraUpdate: null,
     fadeDuration: 300,
     crossSourceCollisions: true,
-    validateStyle: true
+    validateStyle: true,
+    // Because GL MAX_TEXTURE_SIZE is usually at least 4096px
```
HarelM, line 207: Use `/**` comment as they appear in intellisense.

src/ui/map.ts
```diff
@@ -285,6 +288,7 @@ const defaultOptions = {
...
  * @param {Object} [options.locale=null] A patch to apply to the default localization table for UI strings, e.g. control tooltips. The `locale` object maps namespaced UI string IDs to translated strings in the target language; see `src/ui/default_locale.js` for an example with all supported string IDs. The object may specify all UI strings (thereby adding support for a new translation) or only a subset of strings (thereby patching the default translation table).
  * @param {number} [options.pixelRatio] The pixel ratio. The canvas' `width` attribute will be `container.clientWidth * pixelRatio` and its `height` attribute will be `container.clientHeight * pixelRatio`. Defaults to `devicePixelRatio` if not specified.
+ * @param {number} [options.maxCanvasSize] The canvas' `width` and `height` max size. The values are passed as an array where the first element is max width and the second element is max height. You shouldn't set this above WebGl `MAX_TEXTURE_SIZE`. Defaults to [4096, 4096].
```
HarelM, line 291:
```suggestion
 * @param {[number,number]} [options.maxCanvasSize] The canvas' `width` and `height` max size. The values are passed as an array where the first element is max width and the second element is max height. You shouldn't set this above WebGl `MAX_TEXTURE_SIZE`. Defaults to [4096, 4096].
```

src/ui/map.ts
```diff
@@ -357,6 +361,7 @@ class Map extends Camera {
     _removed: boolean;
     _clickTolerance: number;
     _pixelRatio: number;
+    _maxCanvasSize: number[];
```
HarelM, line 364: See my comment above

src/ui/map.ts
HarelM: I still have problem with this code... but I don't know how to solve them - I'm worried that if there's an event fired for example from the painter on resize the event will fire twice in this case...

HarelM:
The comments are minor, the only thing I find lacking is the second loop to resize in case of an overflow. If there was a way to know this before applying, or before running all the resize logic it would be great.

src/ui/map.test.ts
HandyMenny: Do you mean in a new section or new test suite? I didn't find one dedicated to Painter

src/ui/map.ts
HandyMenny: Yes, you're right, I'll update it

HarelM:
Can we use a "dummy" canvas? one that we create just for this specific test maybe?

src/ui/map.test.ts
HarelM: Ah, that's disappointing, yea, painter doesn't have a file of its own. :-(
Never mind then...

src/ui/map.ts
```diff
@@ -2720,8 +2752,8 @@ class Map extends Camera {
 
     _resizeCanvas(width: number, height: number, pixelRatio: number) {
         // Request the required canvas size taking the pixelratio into account.
-        this._canvas.width = pixelRatio * width;
-        this._canvas.height = pixelRatio * height;
+        this._canvas.width = Math.floor(pixelRatio * width);
```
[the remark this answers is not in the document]
HarelM: I see, OK. thanks.

src/ui/map.ts
HandyMenny: I can modify the code to avoid resizing the painter twice, but the canvas could still be resized twice (if the test fails)

HarelM:
@pramilk, any chance for feedback on this? I know you guys are trying to shave milliseconds...

src/ui/map.ts
```diff
@@ -202,7 +203,9 @@ const defaultOptions = {
     transformCameraUpdate: null,
     fadeDuration: 300,
     crossSourceCollisions: true,
-    validateStyle: true
+    validateStyle: true,
+    /** Because GL MAX_TEXTURE_SIZE is usually at least 4096px. */
+    maxCanvasSize: [4096, 4096]
```
pramilk, line 208: Is it possible to use the actual MAX_TEXTURE_SIZE rather then setting it to a fixed 4096 * 4096? This looks constraining for high end machines and might even be buggy on some low end machines.

pramilk, line 208: If we can set this value based on the device actual MAX_TEXTURE_SIZE, I don't see a need to expose this option.

src/ui/map.ts
```diff
@@ -683,10 +689,21 @@ class Map extends Camera {
         const width = dimensions[0];
         const height = dimensions[1];
 
-        this._resizeCanvas(width, height, this.getPixelRatio());
+        let clampedPixelRatio = this._getClampedPixelRatio(width, height, this._maxCanvasSize[0],  this._maxCanvasSize[1]);
+        this._resizeCanvas(width, height, clampedPixelRatio);
+        this.painter.resize(width, height, clampedPixelRatio);
+
+        // check if we've reached GL limits, in that case further clamps pixelRatio
+        if (this.painter.overLimit()) {
```
pramilk, line 697: This code is only executed on map resize. Map resize event does not fire on initial map load. 
Does this code also need to be executed on initial map load?

pramilk, line 697: Optimization: 
1. Once we know that the limit has reached, any subsiquent re-sizing of canvas should use know machine max limit and avoid any double resizing.
2. Instead of using a constant default 4096, if we use MAX_TEXTURE_SIZE to calculate the clampedPixelratio, then we should not require to check for overLimit and thus can avoid any double resizing. Please see if this will be true.

pramilk:
> @pramilk, any chance for feedback on this? I know you guys are trying to shave milliseconds...

Sorry for the delay. I was on vacation and didn't got chance to look at it before. Added some comments.

src/ui/map.ts
HandyMenny: Yes, it's possible, but we can't query that value before attaching a gl context to canvas, so initially it will be unbounded and then eventually resized at the first resize() (which is part of map initialisation anyway).

But I don't really agree with the motivations.
MAX_TEXTURE_SIZE doesn't seem to depend on the performance of the device. On the same smartphone that can barely render complex maps at 2400px * 1080px, one browser has a limit of 4096px and another 16834px.
For low-end devices this limit is surely less demanding than the current implementation, and in any case there is the overlimit() check.
For high end devices, we're talking about resolutions above 4k, I would have some doubts that "common" devices would be able to sustain the load.

HandyMenny: I added this option in the first place to ease testing, some of the tests I added fail if there is a different limit. For unit tests it can probably be bypassed, but for the rendering test I couldn't find a way to force chrome to use a custom MAX_TEXTURE_SIZE.
Then I liked having a way to lower this value for example to make the map less demanding on mobile devices. You can do the same thing with setPixelRatio, but then the pixel ratio becomes fixed so you have to listen for resize events... It's more complex

src/ui/map.ts
HandyMenny: It's already called in the initialization:
https://github.com/maplibre/maplibre-gl-js/blob/3fd4727b53d495807c09c4d08203fb8711fe4b0a/src/ui/map.ts#L545

HandyMenny: The problem is that MAX_TEXTURE_SIZE is the most obvious limit, but there are other limits.
On the machine I'm writing on now MAX_TEXTURE_SIZE is 16834 (sometimes 8192) but the canvas area is limited to about 32Mpx (less than 16834^2 and 8192^2, but above 4096^2).
And I can't figure out how to retrieve that limit, this's what webgl reports: <details><summary>webglreports</summary> 

image
</details>

I agree to remove this overlimit logic, but only if we have a safe default limit (like 4096*4096) and write in the documentation that raising it may result in bad rendering

pramilk: As mentioned in optimization number 1, Once the overSize limit is reached first time, can we store the max limits and use it each subsiquent time. This will ensure that we are not doing double work each time.

pramilk:
Overall looks good to me. I have left a comment to optimize a bit so that this double resizing is only fired once. I dont see any major performance concern, as
1. Will only happen only on a small % of machines. 
2. Even with multiple resize, render will only happen once as its async and gets queued.

src/ui/map.ts
HandyMenny: Updated the pull with that optimization, I implemented it in a simple way, each time overlimit() is true the internal maxCanvasSize is updated. We don't have a get or a set so it should be OK.

HarelM (approved):
THANKS!!!

PR #2680, opened 2023-06-13
Revert to typescript 5.0.4

HarelM:
Not sure why this was merged... But it broke the CI.
I've opened an issue with the typescript repo here:
https://github.com/microsoft/TypeScript/issues/54639
The original PR that failed was #2638

PR #2681, opened 2023-06-13
Improve typings for typescript 5.1.3

HarelM:
Fixes stuff that got broken in typescript 5.1.3.
See #2680
https://github.com/microsoft/TypeScript/issues/54639
The original PR that failed was #2638

### DIFF
--- a/src/render/texture.ts
+++ b/src/render/texture.ts
@@ -41,7 +41,7 @@ class Texture {
         x: number;
         y: number;
     }) {
-        const {width, height} = image;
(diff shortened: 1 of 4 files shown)


birkskyum (approved):

birkskyum:
Could there be other types from the CanvasImageSource that are worth keeping?

HarelM:
I looked at the calling code to that method, and this code can only send these two, so it is aligned with the calling code basically...

PR #2707, opened 2023-06-20
Removed `bindall`, added some types, changed source.ts factory

HarelM:
Superficial refactoring to remove bindall and introduce the typescript `() =>` way of doing things in modern code.

### DIFF
--- a/src/source/canvas_source.ts
+++ b/src/source/canvas_source.ts
@@ -11,10 +11,10 @@ import type Dispatcher from '../util/dispatcher';
 import type {Evented} from '../util/evented';
 
 export type CanvasSourceSpecification = {
-    'type': 'canvas';
-    'coordinates': [[number, number], [number, number], [number, number], [number, number]];
-    'animate'?: boolean;
-    'canvas': string | HTMLCanvasElement;
+    type: 'canvas';
+    coordinates: [[number, number], [number, number], [number, number], [number, number]];
+    animate?: boolean;
+    canvas?: string | HTMLCanvasElement;
 };
 
 /**
@@ -107,7 +107,7 @@ class CanvasSource extends ImageSource {
      * @memberof CanvasSource
      */
 
-    load() {
+    load = () => {
         this._loaded = true;
         if (!this.canvas) {
             this.canvas = (this.options.canvas instanceof HTMLCanvasElement) ?
@@ -137,7 +137,7 @@ class CanvasSource extends ImageSource {
         };
 
         this._finishLoading();
-    }
+    };
 
     /**
      * Returns the HTML `canvas` element.
@@ -174,7 +174,7 @@ class CanvasSource extends ImageSource {
     // */
     // setCoordinates inherited from ImageSource
 
-    prepare() {
+    prepare = () => {
         let resize = false;
         if (this.canvas.width !== this.width) {
             this.width = this.canvas.width;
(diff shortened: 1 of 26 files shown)


birkskyum:
Why are arrow functions preferred in object/class methods?

HarelM:
They preserve "this" when used with event listeners. They do the same as bindall basically.
There's is a different solution to this by adding () => when registering the events. Not sure which is better...

birkskyum (approved):

rotu:
> They preserve "this" when used with event listeners. They do the same as bindall basically. There's is a different solution to this by adding () => when registering the events. Not sure which is better...

I'm a fan of the [method definition syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Method_definitions) because: 

1. I find the arrow function harder to read (`name = (...args) =>`), and the return type before the arrow break my brain)
2. I can use `super`, which cannot be used with arrow function syntax.
3. It has more natural debugger support. Using the arrow definition syntax makes a property enumerable, so it shows up in debugger output where a method wouldn't: image

---

The advantage of late-binding (when registering the callback) is you don't have to look at the implementing class to tell whether `this` is bound. You can always assume it isn't (as is the case with builtin classes). I personally prefer late-binding in general.

The problem with late-binding which can make trouble:
1. It's an implementation detail you can easily forget.
2. It's harder to unsubscribe an event listener - the caller needs to keep a reference to the bound function (similar to how you need to save the return value of `setInterval` if you ever want to call `clearInterval`).

---

Another option to consider is [decorators](https://www.typescriptlang.org/docs/handbook/decorators.html). But I don't think we should be using them. It's still too exotic a feature.

---

It seems suspicious that many properties of [`Source`](https://github.com/maplibre/maplibre-gl-js/blob/682b90de692be0d1fd539d14e005e8ee98269292/src/source/source.ts#L44-L73) are bound by the implementing class - that's a major gotcha that we haven't represented in the type system.

e.g. `prepare` is typed as `() => void` but, if I'm reading the code correctly, implementers must instead conform to the stricter type `(this:unknown) => void`.

clementmas:
Once clear disadvantage of these arrow functions is that it breaks inheritance.

Am I missing something or is it not possible to extend class methods anymore using MapLibre v3?

HarelM:
I don't like these arrow function in general, but I dislike bind more, so I think this is a lesser evil.
I prefer composition over inheritance any time, but there are complicated cases here for composition.
If you have a better way to remove these arrow function and keep the code without these bind stuff, please open a PR and I'll be happy to review it.
One option I guess would be to save the registration callbacks as private members in order to unsubscribe them, but that's a bit messy as well.
In any case, any improvement is welcomed.

neodescis:
I agree with @clementmas. I also strongly dislike bind, but I'm not sure breaking inheritance is better? Also, I believe this is a breaking change if it breaks inheritance, especially in the controls. Breaking inheritance there seems especially onerous.

HarelM:
Looking deeply at this PR, most of the methods are prefixed with `_` meaning private, so these are not part of the public API.
I think the methods in the sources do not require the change and the bind all there was redundant, but this is a bit besides the point, and we might consider fixing it there.
There is the navigation control methods which I think should be prefixed with `_` and it's probably a mistake that they are not.
Bottom line, I agree that this is a breaking change in theory, in practice it's very much contained.
Nevertheless, I'm thinking about version 4, so it might come sooner than later...

I do think we should create some helper method to allow registering an event with an anonymous method and returning something you can unsubscribe from, like rxjs does.
But this require a bit more change.
I'll write it down for version 4 breaking changes.

PR #2708, opened 2023-06-20
Improve handler manager types 

### LINKED ISSUE #1941: Remove "as any" from handler_manager code
## Motivation
handler_manager assumes all kind of things about the handlers it registers.
This code is super confusing and isn't type safe.
It should be.
This was discussed and uncovered as part of #1908 
There's a need to improve typing, reduce string concatination and make the code more readable, even if it's longer, or less dry.

Also, there's a usage of bind and bindall which are not needed in typescript at all.
The more we remove them, the better we are at.

HarelM:
Fixes #1941 - remove as any from handler manager.
Added proper types to things that needed it.
Also improved the typings of the handlers when they are registered.


birkskyum (approved):
Much better, thanks!

PR #2711, opened 2023-06-21
Remove default import in favor of named

HarelM:
As part of looking into the documentation alternatives I've came across a problem in the docs that default exports are "loosing" their name.
While there is a work around for the documentation for it, the author of the docs insisted that this isn't the "right" way.
I also think this might improve the way forward, even just by a bit, to exporting stuff in ESM.
This is a very large PR, but there is very little code changes (two objects that were changed to switch cases).
Other than that it's all imports and exports.


src/render/painter.ts
```diff
@@ -489,7 +475,38 @@ class Painter {
         if (layer.type !== 'background' && layer.type !== 'custom' && !(coords || []).length) return;
         this.id = layer.id;
 
-        draw[layer.type](painter, sourceCache, layer as any, coords, this.style.placement.variableOffsets);
+        switch (layer.type) {
```
HarelM, line 478: This is one place where I added a switch case.

src/style/create_style_layer.ts
```diff
@@ -1,34 +1,39 @@
...
-};
-
-export default function createStyleLayer(layer: LayerSpecification | CustomLayerInterface) {
+export function createStyleLayer(layer: LayerSpecification | CustomLayerInterface) {
     if (layer.type === 'custom') {
         return new CustomStyleLayer(layer);
-    } else {
-        return new subclasses[layer.type](layer);
+    }
+    switch (layer.type) {
```
HarelM, line 18: This is the other one...

birkskyum (approved):

HarelM:
@birkskyum thank you so much for reviewing this!!
I'm sure this was a pain to go through.

PR #2768, opened 2023-06-30
edited map.ts

krishnaglodha:
fixed map.ts

### DIFF
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -54,6 +54,7 @@ import type {
     LayerSpecification,
     FilterSpecification,
     StyleSpecification,
+    SpriteSpecification,
     LightSpecification,
     SourceSpecification,
     TerrainSpecification
@@ -140,13 +141,7 @@ export type MapOptions = {
 };
 
 /**
- * An options object for the gesture settings
- * @example
- * let options = {
- *   windowsHelpText: "Use Ctrl + scroll to zoom the map",
- *   macHelpText: "Use ⌘ + scroll to zoom the map",
- *   mobileHelpText: "Use two fingers to move the map",
- * }
+ * An options object for the gesture settings.
  */
 export type GestureOptions = {
     windowsHelpText?: string;
@@ -155,13 +150,22 @@ export type GestureOptions = {
 };
 
 // See article here: https://medium.com/terria/typescript-transforming-optional-properties-to-required-properties-that-may-be-undefined-7482cb4e1585
+/**
+ * A complete version of `MapOptions` where all properties are required.
+ */
 type Complete<T> = {
     [P in keyof Required<T>]: Pick<T, P> extends Required<Pick<T, P>> ? T[P] : (T[P] | undefined);
 }
 
 // This type is used inside map since all properties are assigned a default value.
+/**
+ * Complete version of `MapOptions` where all properties are required.
+ */
 export type CompleteMapOptions = Complete<MapOptions>;
 
+/**
+ * Options for querying rendered features on the map.
+ */
 type QueryRenderedFeaturesOptions = {
     layers?: Array<string>;
     filter?: FilterSpecification;
@@ -171,14 +175,15 @@ type QueryRenderedFeaturesOptions = {
 
 const defaultMinZoom = -2;
 const defaultMaxZoom = 22;
-
 // the default values, but also the valid range
 const defaultMinPitch = 0;
 const defaultMaxPitch = 60;
-
 // use this variable to check maxPitch for validity
 const maxPitchThreshold = 85;
 
+/**
+ * Default options for configuring the MapLibre GL JS map.
+ */
 const defaultOptions = {
     center: [0, 0],
     zoom: 0,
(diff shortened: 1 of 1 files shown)


src/ui/map.ts
```diff
@@ -1533,42 +1538,69 @@ export class Map extends Camera {
...
+     * the update will be delayed until the previous style is loaded.
+     *
+     * If `style` is `null`, the current style will be removed.
+     *
+     * @example
+     * map.updateStyle('https://example.com/style.json');
+     * map.updateStyle({ version: 8, sources: { ... }, layers: [ ... ] });
+     * map.updateStyle(null);
+     */
+    _updateStyle(style: StyleSpecification | string | null, options?: StyleSwapOptions & StyleOptions): Map {
```
HarelM, line 1595: "_" prefixed method are usually private, I would focus on documenting the public ones, those that doesn't start with "_"

src/ui/map.ts
```diff
@@ -140,13 +141,7 @@ export type MapOptions = {
 };
 
 /**
- * An options object for the gesture settings
```
HarelM, line 143: This info is important, where did you place it?

src/ui/map.ts
krishnaglodha: this will be private ?

HarelM:
Thanks for looking into it!
Above the Map xlass there is a huge comment that need to be split and move the comments to the mapoptions object.

src/ui/map.ts
HarelM: Yes, _updateStyle is private I believe. There's probably another one that is public somewhere in this long file 😀

krishnaglodha:
> Thanks for looking into it! Above the Map xlass there is a huge comment that need to be split and move the comments to the mapoptions object.

Can you please share more about it

HarelM:
See https://github.com/maplibre/maplibre-gl-js/blob/254d57df96901d2213438da6001af9b2aa40eeb4/src/ui/map.ts#L239L297

You can also see my commit in this branch on the map.ts file starting the work to break this comment.

krishnaglodha:
> See https://github.com/maplibre/maplibre-gl-js/blob/254d57df96901d2213438da6001af9b2aa40eeb4/src/ui/map.ts#L239L297
> 
> You can also see my commit in this branch on the map.ts file starting the work to break this comment.

ahh so you want to split this comments into different positions?

HarelM:
Yes, add they need to be documented in the "right place".

src/ui/map.ts
```diff
@@ -73,6 +74,9 @@ export type MapOptions = {
     interactive?: boolean;
     /**
      * The HTML element in which MapLibre GL JS will render the map, or the element's string `id`. The specified element must have no children.
+     * @param {number} [options.bearingSnap=7] The threshold, measured in degrees, that determines when the map's
```
HarelM, line 77: this line should be right above `bearingSnap` no above `container`, it also shouldn't have the `{number}` too.
The comments below also needs fixing.

src/ui/map.ts
```diff
@@ -100,6 +104,33 @@ export type MapOptions = {
     minPitch?: number | null;
     /**
      * The maximum pitch of the map (0-85). Values greater than 60 degrees are experimental and may result in rendering issues. If you encounter any, please raise an issue with details in the MapLibre project.
+     * @param {number} [options.clickTolerance=3] The max number of pixels a user can shift the mouse pointer during a click for it to be considered a valid click (as opposed to a mouse drag).
```
HarelM, line 107: All this part is incorrect, as mentioned above...

src/ui/map.ts
```diff
@@ -140,13 +174,7 @@ export type MapOptions = {
...
 
 /**
- * An options object for the gesture settings
- * @example
- * let options = {
- *   windowsHelpText: "Use Ctrl + scroll to zoom the map",
- *   macHelpText: "Use ⌘ + scroll to zoom the map",
- *   mobileHelpText: "Use two fingers to move the map",
- * }
+ * An options object for the gesture settings.
```
HarelM, line 177: I mean the example code and the rest of the comment, not just this specific line.

src/ui/map.ts
```diff
@@ -155,13 +183,22 @@ export type GestureOptions = {
 };
 
 // See article here: https://medium.com/terria/typescript-transforming-optional-properties-to-required-properties-that-may-be-undefined-7482cb4e1585
+/**
+ * A complete version of `MapOptions` where all properties are required.
```
HarelM, line 187: Not sure this class is even exported, so I'm not sure it's worth adding a comment for it, but your call...

src/ui/map.ts
```diff
@@ -171,14 +208,18 @@ type QueryRenderedFeaturesOptions = {
...
 // the default values, but also the valid range
 const defaultMinPitch = 0;
 const defaultMaxPitch = 60;
-
 // use this variable to check maxPitch for validity
 const maxPitchThreshold = 85;
 
+/**
+ * Default options for configuring the MapLibre GL JS map.
+ * @param {boolean} [options.attributionControl=true] If `true`, an {@link AttributionControl} will be added to the map.
```
HarelM, line 219: This is incorrect too...

src/ui/map.ts
```diff
@@ -418,6 +417,12 @@ export class Map extends Camera {
     /**
      * The map's {@link TwoFingersTouchPitchHandler}, which allows the user to pitch the map with touch gestures.
      * Find more details and examples using `touchPitch` in the {@link TwoFingersTouchPitchHandler} section.
+     * @param {boolean} [options.maplibreLogo=false] If `true`, the MapLibre logo will be shown.
```
HarelM, line 420: Not sure why this is here...

HarelM:
I've stopped the review in the middle as this was far from the right approach.
Please let me know when the concept is clear and all the JSDocs comments are migrated to TSDocs comments in this file.

The idea is that types written like `@param {type} paramName` should not appear in the comments and only the param name should be there. If there are places with `[options.paramName]` the comment should appear above the `paramName` in the class/type/interface where it is defined and not in the class that uses this type/interface.

Hope this is clearer now.

src/ui/map.ts
```diff
@@ -2664,7 +2736,13 @@ export class Map extends Camera {
         return [width, height];
     }
 
-    _setupContainer() {
+    /**
+     * Sets up the map container by adding necessary classes, creating the canvas element,
+     * and attaching event listeners.
+     * @param {boolean} [options.interactive=true] If `false`, no mouse, touch, or keyboard listeners will be attached to the map, so it will not respond to interaction.
```
birkskyum, line 2742: remove type

src/ui/map.ts
```diff
@@ -2508,6 +2573,8 @@ export class Map extends Camera {
      * @param {string} [feature.sourceLayer] (optional) *For vector tile sources, `sourceLayer` is required.*
      * @param {Object} state A set of key-value pairs. The values should be valid JSON types.
      *
+     * @returns {void}
```
birkskyum, line 2576: remove type

src/ui/map.ts
```diff
@@ -2508,6 +2573,8 @@ export class Map extends Camera {
      * @param {string} [feature.sourceLayer] (optional) *For vector tile sources, `sourceLayer` is required.*
      * @param {Object} state A set of key-value pairs. The values should be valid JSON types.
```
birkskyum, line 2574: remove type

src/ui/map.ts
```diff
@@ -2508,6 +2573,8 @@ export class Map extends Camera {
      * @param {string} [feature.sourceLayer] (optional) *For vector tile sources, `sourceLayer` is required.*
```
birkskyum, line 2573: remove type

src/ui/map.ts
```diff
@@ -2481,7 +2546,7 @@ export class Map extends Camera {
     /**
      * Returns the value of the light object.
      *
-     * @returns {LightSpecification} light Light properties of the style.
+     * @returns {LightSpecification} Light properties of the style.
```
birkskyum, line 2549: remove type

src/ui/map.ts
```diff
@@ -2463,16 +2527,17 @@ export class Map extends Camera {
...
 
     /**
-     * Sets the any combination of light values.
+     * Sets any combination of light values.
      *
-     * @param light Light properties to set. Must conform to the [MapLibre Style Specification](https://maplibre.org/maplibre-style-spec/#light).
+     * @param {LightSpecification} light Light properties to set. Must conform to the [MapLibre Style Specification](https://maplibre.org/maplibre-style-spec/#light).
      * @param {StyleSetterOptions} [options] Options object.
      * @param {boolean} [options.validate=true] Whether to check if the filter conforms to the MapLibre Style Specification. Disabling validation is a performance optimization that should only be used if you have previously validated the values you will be passing to this function.
      * @returns {Map} `this`
```
birkskyum, line 2535: remove type

src/ui/map.ts
```diff
@@ -2463,16 +2527,17 @@ export class Map extends Camera {
     }
 
     /**
-     * Sets the any combination of light values.
+     * Sets any combination of light values.
      *
-     * @param light Light properties to set. Must conform to the [MapLibre Style Specification](https://maplibre.org/maplibre-style-spec/#light).
+     * @param {LightSpecification} light Light properties to set. Must conform to the [MapLibre Style Specification](https://maplibre.org/maplibre-style-spec/#light).
      * @param {StyleSetterOptions} [options] Options object.
```
birkskyum, line 2533: remove type

src/ui/map.ts
```diff
@@ -2463,16 +2527,17 @@ export class Map extends Camera {
     }
 
     /**
-     * Sets the any combination of light values.
+     * Sets any combination of light values.
      *
-     * @param light Light properties to set. Must conform to the [MapLibre Style Specification](https://maplibre.org/maplibre-style-spec/#light).
+     * @param {LightSpecification} light Light properties to set. Must conform to the [MapLibre Style Specification](https://maplibre.org/maplibre-style-spec/#light).
```
birkskyum, line 2532: remove type

src/ui/map.ts
```diff
@@ -2427,32 +2490,33 @@ export class Map extends Camera {
...
-    getSprite() {
+    getSprite(): SpriteSpecification | undefined {
         return this.style.getSprite();
     }
 
     /**
      * Sets the value of the style's sprite property.
      *
-     * @param spriteUrl Sprite URL to set.
+     * @param {string | null} spriteUrl The sprite URL to set.
```
birkskyum, line 2511: remove type

src/ui/map.ts
```diff
@@ -2380,33 +2443,33 @@ export class Map extends Camera {
...
         return this.style.getGlyphsUrl();
     }
 
     /**
      * Adds a sprite to the map's style.
      *
      * @param {string} id The ID of the sprite to add. Must not conflict with existing sprites.
-     * @param {string} url The URL to load the sprite from
+     * @param {string} url The URL to load the sprite from.
      * @param {StyleSetterOptions} [options] Options object.
```
birkskyum, line 2466: remove type

src/ui/map.ts
```diff
@@ -2380,33 +2443,33 @@ export class Map extends Camera {
...
+    getGlyphs(): string | null {
         return this.style.getGlyphsUrl();
     }
 
     /**
      * Adds a sprite to the map's style.
      *
      * @param {string} id The ID of the sprite to add. Must not conflict with existing sprites.
-     * @param {string} url The URL to load the sprite from
+     * @param {string} url The URL to load the sprite from.
```
birkskyum, line 2465: remove type

PR #2777, opened 2023-07-02
Js ts controls

HarelM:
This removed the jsdocs plugin which doesn't help at this point.
It adds the tsdocs plugin but without using it as it reports around 4K warnings :-(
It fixes the docs of all the controls, which look pretty neat!

This PR is against the tsdocs branch obvioudly.


birkskyum (approved):
I think this stack is working really well

birkskyum:
It seems like the inheritDoc's was supposed to be automatically added for the methods implementing an interface (https://github.com/TypeStrong/typedoc/issues/1498), but I can confirm that it unfortunately doesn't work if they are removed.

PR #2790, opened 2023-07-03
Migrated popup and marker

HarelM:
Migrated Popup and Marker and changed the Map group to "Main":
It seems that setting the `## Events` section below the examples helps with the grouping for some reason...
image

### DIFF
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -76,7 +76,7 @@ export type MapOptions = {
      * An additional string may optionally be provided to indicate a parameter-styled hash,
      * e.g. http://path/to/my/page.html#map=2.59/39.26/53.07/-24.1/60&foo=bar, where foo
      * is a custom parameter and bar is an arbitrary hash distinct from the map hash.
-     * @devalueValue false
+     * @defaultValue false
      */
     hash?: boolean | string;
     /**
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -420,7 +420,7 @@ const defaultOptions = {
  * object.
  *
  * @group Main
- * 
+ *
  * @example
  * var map = new maplibregl.Map({
  *   container: 'map',
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -68,7 +68,7 @@ import {config} from '../util/config';
 import type {QueryRenderedFeaturesOptions, QuerySourceFeatureOptions} from '../source/query_features';
 
 const version = packageJSON.version;
-/* eslint-enable no-use-before-define */
+
 export type MapOptions = {
     /**
      * If `true`, the map's position (zoom, center latitude, center longitude, bearing, and pitch) will be synced with the hash fragment of the page's URL.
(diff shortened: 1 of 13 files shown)


birkskyum:
It seems like that if the only comment to a return tag is `this`, then it's actually cleaner to not have it because it's reflected from the code, which is impressive because this specific addTo function from hash.ts doesn't even state `: Hash`.

Screenshot 2023-07-03 at 23 27 45
Screenshot 2023-07-03 at 23 27 22

src/ui/marker.ts
```diff
@@ -12,39 +12,74 @@ import type {LngLatLike} from '../geo/lng_lat';
 import type {MapMouseEvent, MapTouchEvent} from './events';
 import type {PointLike} from './camera';
 
+type Alignment = 'map' | 'viewport' | 'auto';
```
birkskyum, line 15: TSDoc comment above this will include it in the docs

src/ui/marker.ts
```diff
@@ -12,39 +12,74 @@ import type {LngLatLike} from '../geo/lng_lat';
...
+
 type MarkerOptions = {
```
birkskyum, line 17: TSDoc comment above this will include it in the docs

birkskyum:
It seems like the already migrated NavigationControl also lack a TSDoc comment above the NavigationOptions.

src/ui/marker.ts
HarelM: I wasn't sure what to write here honestly...

HarelM:
Thanks for the review!

src/ui/marker.ts
birkskyum: Just the name of the type is sufficient to make typedoc included it

HarelM: I added something...

src/ui/marker.ts
birkskyum: I unresolved this, because it's still missing

Screenshot 2023-07-04 at 00 02 20

HarelM: ahh, my bad, sorry... I thought it was a duplicate comment.

birkskyum: I've found that even adding just the following will include the respective class/methods/function/type , so if it's not always clear what to write, that is a fast improvement

```
/**
 */
 ```

HarelM:
@birkskyum note that I removed a non-documented part of the `Marker` constructor.
I'm not sure it qualifies for a major version release, and if so I'll revert it, but this basically is a non documented behavior, I've mainly aligned the code with the docs.
Let me know what you think.
Other comments were addressed.

birkskyum (approved):

src/ui/map.ts
birkskyum, line 1: Replace `devalueValue` with `defaultValue`

birkskyum:
> @birkskyum note that I removed a non-documented part of the `Marker` constructor. I'm not sure it qualifies for a major version release, and if so I'll revert it, but this basically is a non documented behavior, I've mainly aligned the code with the docs. Let me know what you think. Other comments were addressed.

I see, the legacyOptions - I think it's reasonable to remove it like this.

HarelM:
I really like how less verbose and in the right place the new comments are, instead of lay around in places that are not relevant.
Merging as CI will run on the draft PR.

PR #2794, opened 2023-07-03
Add MapOptions to docs

birkskyum:
The table of map options is currently missing


HarelM:
I've fixed that manually in the branch...

PR #2796, opened 2023-07-03
Add tsdoc hyphens to param in map.ts

birkskyum:
TSDoc requires a hyphen after the `@param name - Description`, which jsdoc didn't need


HarelM:
Something is off with the branch, I'm not sure why...
Both tests are failing and main seems to be diverged...
I'll look at it tomorrow as I'm too tired and I'm probably making mistakes...

HarelM:
In any case, I would like to do it after all the docs are converted in a single pass, hopefully automatically...

PR #2801, opened 2023-07-04
JSDocs TSDocs leftovers

HarelM:
Removed most of the leftovers.
I think most of the comments are OK now.

### DIFF
--- a/src/geo/lng_lat_bounds.ts
+++ b/src/geo/lng_lat_bounds.ts
@@ -292,7 +292,7 @@ export class LngLatBounds {
      *
      * @param center center coordinates of the new bounds.
      * @param radius Distance in meters from the coordinates to extend the bounds. default 0
-     * @returns {LngLatBounds} A new `LngLatBounds` object representing the coordinates extended by the `radius`.
+     * @returns A new `LngLatBounds` object representing the coordinates extended by the `radius`.
      * @example
      * var center = new maplibregl.LngLat(-73.9749, 40.7736);
      * maplibregl.LngLatBounds.fromLngLat(100).toArray(); // = [[-73.97501862141328, 40.77351016847229], [-73.97478137858673, 40.77368983152771]]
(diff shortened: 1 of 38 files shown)


src/geo/lng_lat_bounds.ts
```diff
@@ -279,8 +291,8 @@ export class LngLatBounds {
      * Returns a `LngLatBounds` from the coordinates extended by a given `radius`. The returned `LngLatBounds` completely contains the `radius`.
      *
      * @param center center coordinates of the new bounds.
-     * @param {number} [radius=0] Distance in meters from the coordinates to extend the bounds.
-     * @returns {LngLatBounds} A new `LngLatBounds` object representing the coordinates extended by the `radius`.
+     * @param radius Distance in meters from the coordinates to extend the bounds. default 0
```
birkskyum, line 294: I don't think we need the `default 0` here, because it's extracted from the function arguments.

Screenshot 2023-07-04 at 16 35 55
Screenshot 2023-07-04 at 16 35 36

HarelM: Good point! I didn't know what will happen, so I left it there just in case...

src/source/tile.ts
```diff
@@ -225,7 +227,7 @@ export class Tile {
 
     /**
      * Release any data or WebGL resources referenced by this tile.
-     * @returns {undefined}
+     * @returns
```
birkskyum, line 230: This `@returns` can be removed, and it'll still show up as:
Screenshot 2023-07-04 at 16 41 13

birkskyum:
in the top of `properties.ts`, these two types lack tsdoc comment.

type TimePoint = number;

export type CrossFaded<T> = {
    to: T;
    from: T;
};

Screenshot 2023-07-04 at 16 45 45

birkskyum (approved):

src/source/tile.ts
HarelM: removed.

HarelM:
Not sure I know what to write about it.

PR #2803, opened 2023-07-05
Add tsdoc.json

birkskyum:
This adds all the TypeDoc specific tags to TSDoc (see more [here](https://typedoc.org/guides/tags/) and [here](https://typedoc.org/options/configuration/)), thus removing lots of lint warning for tags like `@group, @private` etc. that the linter says aren't configured because it doesn't exist in base [tsdoc](https://tsdoc.org/).

1726 -> 1377 problems

If this doesn't appear to change anything, removing the `cache` flag from `npm run lint` can do wonders


HarelM:
This is part of #2804.
I'll close this.
Thanks for the tip!

PR #2807, opened 2023-07-06
fix compile errors

Namue:
## Launch Checklist

 - Fix compile errors (Not all code paths return a value /  Property '...' comes from an index signature, so it must be accessed with ['...'].) 
 - [ ] Link to related issues.
 - no issues
 - no changes
 - no new functionality
 - no changes
 - [ ] Post benchmark scores.
 - no changes


HarelM:
Thanks for taking the time to contribute!!
Isn't `return;` and `return undefined` are the same?
I personally dislike the use of `['member']` as the IDE doesn't know about it when you refactor your code.
Is there a better way to solve this?

Also I didn't see any CI change to make sure this won't break in the future - I would expect to see a tsconfig.json change or something similar...

Namue:
Yes, from outside the function `return `and `return undefined` is the same. Even a missing return would also result in an `undefined`.
I think it's recognized as an error, because the compiler can not know, if you wrote (or omitted) the code because you know that, or if you didn't see this line and undefined will be totally wrong.

Yes, I like the point access to members more too, but is it possible to rename the properties at the moment?
For example one class in uniform_bindin.ts  is `export type UniformLocations = {[_: string]: WebGLUniformLocation};`
This means, every string can be used as key.
An error was shown for example in the second line at locations.u_matrix:
```
const backgroundUniforms = (context: Context, locations: UniformLocations): BackgroundUniformsType => ({
    'u_matrix': new UniformMatrix4f(context, locations.u_matrix),
```
Can an IDE change a property name, when it is not defined in its type? This is the reason the error was shown to me. If the key had been existing in the type, everything would be fine.

Yes, I would prefer a change in a configuration file too. I don't see any possibility at the moment, to change the behaviour of the compiler für your code, but not for the code of the surrounding application.
I will check if I missed a possibility.

HarelM:
If there are missing types or definitions, I suggest to add them.
The current PR at it's current state only degrade the code in this library, I don't see the benefit though...
So I can't approve it as is, sorry...

PR #2813, opened 2023-07-07
Main docs readme links fixes

HarelM:
I've aligned the group to the original doc's structure and fixed the linked in the main readme.
I also changed the main exported const to be a static class - this is mainly so it would show up in the docs with all the relevant information - I'm not super excited about this change so I'll need a second opinion here.

Added also the build to the CI, although I'm not sure what it actually checks...

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -121,6 +121,19 @@ jobs:
           destination_dir: benchmarks/${{ steps.prepare_release.outputs.version_tag }}
           commit_message: 'deploy: ${{ steps.prepare_release.outputs.version_tag }}'
 
+      - name: Build the Docs
+        run: |
+          npm run generate-docs
+          docker run --rm -v ${PWD}:/docs squidfunk/mkdocs-material build
+
+      - name: Publish the docs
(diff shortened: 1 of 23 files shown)


HarelM:
@birkskyum, I have published an initial version of the docs to the following address (using this repo's github pages under the `docs` folder):
https://maplibre.org/maplibre-gl-js/docs/

Feel free to review.
I still need to work on making the CI run as expected, but this shouldn't be an issue.

Any feedback is welcomed.

birkskyum:
Do you have an example of where this static class changes the Docs?

HarelM:
The default.md looks very different if it's not a static class

birkskyum:
I think it all looks good, except the `static`'s which are quite annoying. Does it i.e. make a change to use the `@next` version of the typedoc markdown plugin? It's quite a big update, so it could be.

HarelM:
There's a need to tests and see how it looks I guess...

HarelM:
Generally speaking, these methods are static in nature basically. So it's not "wrong", just weird...

birkskyum (approved):

PR #2827, opened 2023-07-09
Fix 3d docs

HarelM:
I still need to figure out how to replace the 34M_17.bin file with a valid one.
If I change that locally git will ignore it and won't let me push it :-/


HarelM:
These fixes are needed regardless of the binary fix in #2830.
After that we can release pre.3 and if everything works we can release 3.2.

birkskyum (approved):

PR #2828, opened 2023-07-09
Fix 3d example bin file - PR 1/2

birkskyum:
Remove the existing .bin file. Will in another PR add again the correct one from maplibre-gl-js-docs


HarelM (approved):

HarelM:
I think we'll need to use something like:
https://stackoverflow.com/questions/26831494/delete-all-previous-versions-of-file-in-git-repo

PR #2830, opened 2023-07-09
Fix 3d example bin file - PR 2/2

birkskyum:
Add .bin file again


birkskyum:
This doesn't work. Wrong file size.

HarelM:
You can test to see if it worked on the branch, before pushing to main.
Spoiler, I tried that and it didn't work.
We need to remove the file from git history for this to work I believe.

birkskyum:
Tried rename, it still has the altered file size.

HarelM (approved):

PR #2835, opened 2023-07-09
Bump version to 3.2.0

HarelM:
Release version 3.2.

I believe I fixes most of the docs issues.
I'm fairly pleased with it.
I've added a condition to the release.yml to avoid publishing the docs when publishing a pre-release.


birkskyum (approved):
Awesome!

PR #2836, opened 2023-07-09
Remove not working Showcase URLs

KristjanESPERANTO:


HarelM:
There's actually working, we are just in the middle of migrating the docs and this was a typo...

PR #2837, opened 2023-07-10
Fix setting line-gradient to null/undefined

### LINKED ISSUE #2683: Is there a method to delete layer style attributes
**maplibre-gl-js version**: 2.4.0

**browser**:

### Steps to Trigger Behavior

 I briefly describe the needs .
 According to some attributes of the api map, they have priority among each other .
 For example: line-color and line-pattern .
 line-pattern has higher priority than line-color .
 When I create a layer and set the line-pattern property .
```
		map.addLayer({
			type: 'line',
			source: 'line',
			id: 'line',
			paint: {
				'line-color': 'red',
				'line-width': 14,
				// 'line-gradient' must be specified using an expression
				// with the special 'line-progress' property
				'line-gradient': [
					'interpolate',
					['linear'],
					['line-progress'],
					0,
					'blue',
					0.1,
					'royalblue',
					0.3,
					'cyan',
					0.5,
					'lime',
					0.7,
					'yellow',
					1,
					'red'
				]
			},
			layout: {
				'line-cap': 'round',
				'line-join': 'round'
			}
		});
```
Now I want to use the line-color property .
How should I remove the effect of the line-gradient attribute .
I tried the following .

```
map.setPaintProperty('line', 'line-gradient', undefined)
map.setPaintProperty('line', 'line-color', 'red')
```

In this way, the console will report an error .
I checked the api and couldn't find the corresponding method .
Is it possible to provide a method to delete an attribute .
For example : 
`map.removePaintProperty('line', 'line-gradient')`

### Link to Demonstration

https:maplibre.org/maplibre-gl-js-docs/example/line-gradient/

### Expected Behavior

Use the following method to delete an attribute in a layer

map.removePaintProperty(layerId, 'line-gradient')

### Actual Behavior

tangerine-orange:
This PR fixes #2683

## Problem
A `StyleLayer` is added to a map by calling [`map.addLayer`](https://maplibre.org/maplibre-gl-js/docs/API/classes/maplibregl.Map/#addlayer). In particular, we can add a `LineStyleLayer` that adds a color gradient (`line-gradient`) as follows:
```
    map.addLayer({
        type: 'line',
        source: 'line',
        id: 'line',
        paint: {
            'line-color': 'red',
            'line-width': 14,
            'line-gradient': [
                'interpolate',
                ['linear'],
                ['line-progress'],
                0,
                'blue',
                1,
                'red'
            ]
        }
    });
```
Our problem comes when we try to unset the `line-gradient` using [`Map.setPaintProperty`](https://maplibre.org/maplibre-gl-js/docs/API/classes/maplibregl.Map/#setpaintproperty):
```
map.setPaintProperty('line', 'line-gradient', null)
```
Note the third argument, `null`, which is meant to unset the property. (Note that `null` and `undefined` have same behavior in this case)
Our app crashes and we see an error in console:
Screen Shot 2023-07-09 at 3 59 12 PM

## Explanation
When we call `map.setPaintProperty('line', 'line-gradient', null)`, our callstack is:

-->[`LineStyleLayer._handleSpecialPaintPropertyUpdate`](https://github.com/maplibre/maplibre-gl-js/blob/main/src/style/style_layer/line_style_layer.ts#L61)
->[`StyleLayer.setPaintProperty`](https://github.com/maplibre/maplibre-gl-js/blob/main/src/style/style_layer.ts#L150)
[`Map.setPaintProperty`](https://maplibre.org/maplibre-gl-js/docs/API/classes/maplibregl.Map/#setpaintproperty)

In `LineStyleLayer._handleSpecialPaintPropertyUpdate`, there is a special case for `line-gradient`.
```
        if (name === 'line-gradient') {
            const expression: ZoomConstantExpression<'source'> = (this._transitionablePaint._values['line-gradient'].value.expression as any);
            this.stepInterpolant = expression._styleExpression.expression instanceof Step;
            this.gradientVersion = (this.gradientVersion + 1) % Number.MAX_SAFE_INTEGER;
        }
```
You can see that this code assumes the existence of `expression._styleExpression`. This assumption is broken when we're setting the property to `null`.

## Solution
Check for the existence of `_styleExpression` before trying to access it.

## Bonus
`Map.setPaintProperty` expects you to pass `null` when unsetting the existing value, not `undefined`. This is what we test for. I added this info to the documentation for this method.

## Potential future work
1. Do we want to enforce the use of `null` over `undefined` for unsetting? I'd imagine we would do this using Typescript. If so I can make ticket.
2.  [Casting as `ZoomConstantExpression`](https://github.com/maplibre/maplibre-gl-js/blob/main/src/style/style_layer/line_style_layer.ts#L63) doesn't seem ideal, especially since `ZoomConstantExpression` isn't even one of the possible types for `this._transitionablePaint._values['line-gradient'].value.expression`. We should probably clean up the typing here? If so I can make ticket.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/style/style_layer.test.ts
+++ b/src/style/style_layer.test.ts
@@ -6,7 +6,7 @@ import {EvaluationParameters} from './evaluation_parameters';
 import {TransitionParameters} from './properties';
 import {BackgroundStyleLayer} from './style_layer/background_style_layer';
 import {SymbolStyleLayer} from './style_layer/symbol_style_layer';
-import { LineStyleLayer } from './style_layer/line_style_layer';
+import {LineStyleLayer} from './style_layer/line_style_layer';
 
 describe('StyleLayer', () => {
     test('instantiates the correct subclass', () => {
@@ -392,7 +392,7 @@ describe('StyleLayer->LineStyleLayer', () => {
                 ]
             }
         }, layer);
-    };
+    }
 
     test('updating with valid line-gradient updates this.gradientVersion', () => {
         const lineLayer = createStyleLayer(createLineLayer()) as LineStyleLayer;
@@ -417,4 +417,4 @@ describe('StyleLayer->LineStyleLayer', () => {
         lineLayer.setPaintProperty('line-gradient', null);
         expect(lineLayer.gradientVersion).toBeGreaterThan(gradientVersion);
     });
-})
+});
--- a/src/style/style_layer.test.ts
+++ b/src/style/style_layer.test.ts
@@ -6,7 +6,6 @@ import {EvaluationParameters} from './evaluation_parameters';
 import {TransitionParameters} from './properties';
 import {BackgroundStyleLayer} from './style_layer/background_style_layer';
 import {SymbolStyleLayer} from './style_layer/symbol_style_layer';
-import {LineStyleLayer} from './style_layer/line_style_layer';
 
 describe('StyleLayer', () => {
     test('instantiates the correct subclass', () => {
@@ -369,52 +368,5 @@ describe('StyleLayer#serialize', () => {
         const layer = createStyleLayer({type: 'fill'} as LayerSpecification);
         // paint is never undefined
         expect(layer.paint).toBeTruthy();
-    });
-});
-
-describe('StyleLayer->LineStyleLayer', () => {
-    function createLineLayer(layer?) {
-        return extend({
-            type: 'line',
-            source: 'line',
-            id: 'line',
-            paint: {
-                'line-color': 'red',
-                'line-width': 14,
-                'line-gradient': [
-                    'interpolate',
-                    ['linear'],
-                    ['line-progress'],
-                    0,
-                    'blue',
-                    1,
-                    'red'
-                ]
-            }
-        }, layer);
-    }
-
-    test('updating with valid line-gradient updates this.gradientVersion', () => {
-        const lineLayer = createStyleLayer(createLineLayer()) as LineStyleLayer;
-        const gradientVersion = lineLayer.gradientVersion;
-
-        lineLayer.setPaintProperty('line-gradient', [
-            'interpolate',
-            ['linear'],
-            ['line-progress'],
-            0,
-            'red',
-            1,
-            'blue'
-        ]);
-        expect(lineLayer.gradientVersion).toBeGreaterThan(gradientVersion);
-    });
-
-    test('updating with invalid line-gradient updates this.gradientVersion', () => {
-        const lineLayer = createStyleLayer(createLineLayer()) as LineStyleLayer;
-        const gradientVersion = lineLayer.gradientVersion;
-
-        lineLayer.setPaintProperty('line-gradient', null);
-        expect(lineLayer.gradientVersion).toBeGreaterThan(gradientVersion);
     });
 });
--- a/src/style/style_layer.test.ts
+++ b/src/style/style_layer.test.ts
@@ -6,6 +6,7 @@ import {EvaluationParameters} from './evaluation_parameters';
 import {TransitionParameters} from './properties';
 import {BackgroundStyleLayer} from './style_layer/background_style_layer';
 import {SymbolStyleLayer} from './style_layer/symbol_style_layer';
+import { LineStyleLayer } from './style_layer/line_style_layer';
 
 describe('StyleLayer', () => {
     test('instantiates the correct subclass', () => {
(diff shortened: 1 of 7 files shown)


src/style/style_layer/line_style_layer.ts
```diff
@@ -60,8 +60,13 @@ export class LineStyleLayer extends StyleLayer {
 
     _handleSpecialPaintPropertyUpdate(name: string) {
         if (name === 'line-gradient') {
-            const expression: ZoomConstantExpression<'source'> = (this._transitionablePaint._values['line-gradient'].value.expression as any);
-            this.stepInterpolant = expression._styleExpression.expression instanceof Step;
+            if (Object.prototype.hasOwnProperty.call(this.gradientExpression(), '_styleExpression')) {
```
tangerine-orange, line 63: It seems like the codebase has a preference for this `hasOwnProperty` syntax: https://github.com/search?q=repo%3Amaplibre%2Fmaplibre-gl-js%20Object.prototype.hasOwnProperty&type=code

tangerine-orange:
@HarelM This is my first time contributing to this project -- do I need to do something to start the CI jobs?

HarelM:
I need to start it, I changed the settings now so that it won't happen in the future...

src/style/style_layer/line_style_layer.ts
```diff
@@ -60,8 +60,12 @@ export class LineStyleLayer extends StyleLayer {
...
+                const expression: ZoomConstantExpression<'source'> = this.gradientExpression() as any;
```
HarelM, line 64: Not really related to this PR, but can we remove this "as any" by any chance?

src/style/style_layer/line_style_layer.ts
HarelM: Can we call this.gradientExpression before the if condition to calculate it only once?

src/style/style_layer.test.ts
```diff
@@ -370,3 +371,50 @@ describe('StyleLayer#serialize', () => {
         expect(layer.paint).toBeTruthy();
     });
 });
+
+describe('StyleLayer->LineStyleLayer', () => {
+    function createLineLayer(layer?) {
```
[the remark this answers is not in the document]
HarelM: I believe it will make more sense to have a line_style_layer.test.ts file.
But it's not too critical for me...

HarelM:
I think setting null and undefined should be possible and should behave the same, nice to have though.
If the typings aren't great I suggest to improve them as part of this PR.
If you prefer to have it under a different PR that completely fine too.

src/style/style_layer/line_style_layer.ts
tangerine-orange: Yes I considered doing this. It would look like:
```
            const expression: ZoomConstantExpression<'source'> = this.gradientExpression() as any;
            this.stepInterpolant = !!Object.prototype.hasOwnProperty.call(expression, '_styleExpression')
                && expression._styleExpression.expression instanceof Step;
            this.gradientVersion = (this.gradientVersion + 1) % Number.MAX_SAFE_INTEGER;
```
The reason I decided against: in this version, we are casting `expression: ZoomConstantExpression<'source'>` before we check the existence of `expression._styleExpression`. That is, we cast to `ZoomConstantExpression` before we know that it actually is of that type.

Here's another possibility:
```
            const expression = this.gradientExpression() as any;
            this.stepInterpolant = !!Object.prototype.hasOwnProperty.call(expression, '_styleExpression')
                && expression._styleExpression.expression instanceof Step;
            expression as ZoomConstantExpression<'source'>;
            this.gradientVersion = (this.gradientVersion + 1) % Number.MAX_SAFE_INTEGER;
```

src/style/style_layer/line_style_layer.ts
tangerine-orange: We do this:
```
const expression = this.gradientExpression() as  ZoomConstantExpression<'source'>;
```
We have to do some casting no matter what, since our `expression` is of type `StylePropertyExpression = ConstantExpression | SourceExpression | CameraExpression | CompositeExpression`. Notably excluding the type we expect, `ZoomConstantExpression`.

tangerine-orange:
>  If you prefer to have it under a different PR that completely fine too.

I think I prefer to do it separate. I will make a ticket for it and link this PR. I suspect it might get pretty messy so I don't want to hold up this bugfix.

tangerine-orange:
> > If you prefer to have it under a different PR that completely fine too.
> 
> I think I prefer to do it separate. I will make a ticket for it and link this PR. I suspect it might get pretty messy so I don't want to hold up this bugfix.

https://github.com/maplibre/maplibre-style-spec/issues/267

src/style/style_layer/line_style_layer.ts
HarelM: That looks better, so should we wait for the other PR first?

tangerine-orange: IMO we ship this now for the bugfix, then probably we change the the typing again after the [style spec issue](https://github.com/maplibre/maplibre-style-spec/issues/267#issuecomment-1632273603) is resolved.

tangerine-orange:
@HarelM Are you okay with this branch as is or did you want me to do anything else before shipping?

src/style/style_layer/line_style_layer.ts
HarelM: Can we do the casting outside the if and avoid the "hasOwnProperty" and simply check if (expression._style...) ...?

HarelM:
See my last comment. Otherwise approved.

src/style/style_layer/line_style_layer.ts
tangerine-orange: kind of..
1. Doing the casting outside of the `if`
     - yes technically we can do this, but I prefer not. I chose to do casting after checking existence of `._styleExpression` for consistency between types and code logic. `expression` is a `ZoomConstantExpression` only if `._styleExpression` exists. So to me it makes sense to cast to `ZoomConstantExpression` only after checking existence of `._styleExpression`
2. check `expression._style...` instead of `hasOwnProperty`
     - I'm totally ok with doing this. I used `hasOwnProperty` because it seems like it's used a lot around the codebase, but I personally have no preference.

HarelM: I would consider adding a type guard then.
Something like:
`isZoomExpression(expr: ...) expr is ZoomExpression {...`
It will do the relevant casting and check the relevant property.
I think it will be the cleanest.

HarelM:
@tangerine-orange where are we with this PR?

tangerine-orange:
@HarelM Sorry for the delay. I've added the type guard you suggest to https://github.com/maplibre/maplibre-style-spec/pull/334. I integrated the updated maplibre-style-spec locally and it works as intended.

tangerine-orange:
@HarelM okay, I integrated the new version of @maplibre/maplibre-gl-style-spec and am using the type guard.

HarelM (approved):

HarelM:
Am I missing a change in the package.json file? I see the luck file change, but not the package...?

tangerine-orange:
> Am I missing a change in the package.json file? I see the luck file change, but not the package...?

my bad, added.

PR #2854, opened 2023-07-10
Fix typo

chekos:
Going through the docs I saw this typo.

I don't usually contribute to projects in Javascript so I wouldn't know how to run the tests and check other stuff but hopefully this is a small enough change that it doesn't actually require that.

### DIFF
--- a/build/generate-docs.ts
+++ b/build/generate-docs.ts
@@ -13,7 +13,7 @@ function generateAPIIntroMarkdown(lines: string[]): string {
     let intro = `# Intro
 
 This file is intended as a reference for the important and public classes of this API.
-We reccomend looking at the [expamples](../examples/index.md) as they will help you the most to start with MapLibre.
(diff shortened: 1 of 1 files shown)


build/generate-docs.ts
```diff
@@ -13,7 +13,7 @@ function generateAPIIntroMarkdown(lines: string[]): string {
     let intro = `# Intro
 
 This file is intended as a reference for the important and public classes of this API.
-We reccomend looking at the [expamples](../examples/index.md) as they will help you the most to start with MapLibre.
+We reccomend looking at the [examples](../examples/index.md) as they will help you the most to start with MapLibre.
```
birkskyum, line 16: might as well fix the `reccomend` as well :)

HarelM:
```suggestion
We recommend looking at the [examples](../examples/index.md) as they will help you the most to start with MapLibre.
```

Issue #2861, opened 2023-07-12
Warn against using underscore-prefixed symbols


1ec5:
The new documentation pages include entries for formerly private methods that are prefixed with an underscore, such as [`_applyUpdatedTransform`](https://maplibre.org/maplibre-gl-js/docs/API/classes/maplibregl.Map/#_applyupdatedtransform). To my knowledge, these methods are not intended to be used by consumers of the library in general, although some plugins use them by necessity. They should either be hidden by default or appropriately marked, so that developers unfamiliar with the underscore convention won’t inadvertently depend on an unstable part of the API. External usage beyond plugins could make it more difficult to ensure backwards compatibility going forward. Even prefacing the description with the word “Internal” would be sufficient I think.

HarelM:
The right thing to do from a code perspective is to mark those as private.
Having said that, there was a request not to do it as part of #333.
The other option is to add `@hidden`, `@internal` or `@private` to all these method/properties in order to exclude them from the docs.

Setting M bounty.
Bounty direction:
https://github.com/maplibre/maplibre/issues/191

visitskyworld:
@HarelM can I take this one?

HarelM:
I think @ZeLonewolf is currently working on it.
But there's a lot to solve in the docs besides this issue.
Go ahead and explore them so you could suggest other improvements.

1ec5:
#2861 addresses my concern with respect to the underscore-prefixed symbols by hiding them. I wasn’t suggesting a runtime warning, just a ⚠️ warning message in the documentation, but hiding the symbols from the documentation has the same effect as far as downstream developers are concerned.

#2916 is a follow-up issue for internal types.

HarelM:
Resolved by #2863
@ZeLonewolf you are welcome to submit an expense in open collective to collect your bounty.
If you don't plan to do so, please let me know and I'll remove the bounty from this issue.
THANKS!

ZeLonewolf:
I do not intend to seek a bounty.

PR #2863, opened 2023-07-12
Improve documentation on Map and Camera classes

ZeLonewolf:
This improves the Map/Camera class documentation in the following ways:
- Private functions and properties are suppressed from the output
- Event annotations are constrained only to event listeners, and lists of events are added to each method that fires events.
- Some events were added to the Camera class where they seemed to be missing/wrong/copy-paste-errored, for example, adding pitch and rotate to functions that pitch and rotate the map.

If the table takes up too much white space to list the Camera class events, it can be changed to a simple, more compact list.

Related to #2861, however, this changes documentation only and does not fire warnings.

## Launch Checklist

### DIFF
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -505,7 +505,7 @@ export class Map extends Camera {
 
     /**
      * @hidden
-     * image queue throttling handle. To be used later when clean up 
+     * image queue throttling handle. To be used later when clean up
      */
     _imageQueueHandle: number;
 
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -818,14 +818,7 @@ export class Map extends Camera {
      * This method must be called after the map's `container` is resized programmatically
      * or when the map is shown after being initially hidden with CSS.
      *
-     * Events triggered:
-     *
-     * | Event       |
-     * |-------------|
-     * | `move`      |
-     * | `moveend`   |
-     * | `movestart` |
-     * | `reside`    |
+     * Triggers the following events: `movestart`, `move`, `moveend`, and `resize`.
      *
      * @param eventData - Additional properties to be passed to `movestart`, `move`, `resize`, and `moveend`
      * events that get triggered as a result of resize. This can be useful for differentiating the
@@ -1862,9 +1855,7 @@ export class Map extends Camera {
      *
      * Events triggered:
      *
-     * | Event                  |
-     * |------------------------|
-     * | `source.add`           |
+     * Triggers the `source.add` event.
      *
      * @param id - The ID of the source to add. Must not conflict with existing sources.
      * @param source - The source object, conforming to the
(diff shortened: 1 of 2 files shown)


src/ui/map.ts
```diff
@@ -1071,6 +1115,12 @@ export class Map extends Camera {
      * If the map's current pitch is higher than the new maximum,
      * the map will pitch to the new maximum.
      *
+     * Events triggered:
```
HarelM, line 1118: I'm not sure this actually raises an event, looking at the code it seems that this method throws an exception, which I think is different.
I've seen method "firing" and `error` event, which I believe is different.

ZeLonewolf:
Map before and after:

image
image

flyTo before and after:

image
image

Second example is what I mean by compacting the events into a plain list.  The table takes up too much space and I think I want to change that.

src/ui/map.ts
```diff
@@ -1890,6 +1955,14 @@ export class Map extends Camera {
 
     /**
      * Loads a 3D terrain mesh, based on a "raster-dem" source.
+     *
+     * Events triggered:
+     *
+     * | Event     | Condition              |
+     * |-----------|------------------------|
+     * | `error`   | Failed to load source  |
```
HarelM, line 1963: Does it? I see only "warnOnce"...

ZeLonewolf: Line 1989 says:
```typescript
if (!sourceCache) throw new Error(`cannot load terrain, because there exists no source with ID: ${options.source}`);
```

ZeLonewolf: Oh, I see, that's an ordinary error, not an event.

ZeLonewolf: Fixed in c3483f4e6

ZeLonewolf:
Some of the camera class functions fire as many as 10 events which will take up quite a bit of vertical space even as a plain markdown list.  I was thinking just an in text list like:

Fires the following events: `event1`, `event2`, and `event3`.

HarelM:
Sounds good too, this was my initial thought :-)

src/ui/map.ts
```diff
@@ -810,6 +818,15 @@ export class Map extends Camera {
      * This method must be called after the map's `container` is resized programmatically
      * or when the map is shown after being initially hidden with CSS.
      *
+     * Events triggered:
```
HarelM, line 821: Didn't you prefer a list instead of a table?

src/ui/map.ts
```diff
@@ -1830,11 +1860,16 @@ export class Map extends Camera {
     /**
      * Adds a source to the map's style.
      *
+     * Events triggered:
```
HarelM, line 1863: List instead of table?

src/ui/map.ts
```diff
@@ -1890,6 +1927,13 @@ export class Map extends Camera {
 
     /**
      * Loads a 3D terrain mesh, based on a "raster-dem" source.
+     *
```
HarelM, line 1930: Save here

ZeLonewolf:
I removed the three tables noted.  There's still one function with a table, but it's two-column so I think it makes sense to keep that one.

HarelM (approved):

PR #2894, opened 2023-07-20
Exposed the DOM util class to the public

TheSpartaPT:
Very small change that only exposes the DOM util class to the public. This will be useful for people wanting to make custom controls, and will allow them to use the same DOM manipulation util class as the bundled controls.

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -15,6 +15,7 @@ import {LngLatBounds} from './geo/lng_lat_bounds';
 import Point from '@mapbox/point-geometry';
 import {MercatorCoordinate} from './geo/mercator_coordinate';
 import {Evented} from './util/evented';
+import {DOM} from './util/dom';
 import {config} from './util/config';
 import {Debug} from './util/debug';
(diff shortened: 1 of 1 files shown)


HarelM:
I'm sure there are better DOM utilities packages in npm, I'm not sure this is the best one or there.
But if this is to be exposed, every public method should be documented using tsdocs comments.

HarelM:
I've looked through the file, I don't think it will be a good candidate for a public API kinf if class...

TheSpartaPT:
yeah, there probably are better, more in-depth libraries that only focus on dom manipulation, but the idea here was to be able to use the same utils maplibre is using for custom controls and such

HarelM:
Should we close this PR?
The alternative is to really invest in the DOM utils class to have it publicly, meaning good docs, well though of API, etc.
This is currently not the case with this PR.
Please decide which direction you would like to go.

PR #2897, opened 2023-07-21
Don't draw halo pixels "underneath" text pixels

ChrisLoer:
This prevents unexpected color blending when labels are partially transparent. This also allows fully transparent text with an opaque halo for an "outline" effect.

We came across the previous behavior at Felt because our UI gives users control over `text-color` and `text-halo-color` and allows setting the alpha channel. I think if you have a partially transparent label, you don't expect the of the text to be blended with the halo color.

Here's the previous behavior with transparent text-color and a black halo:
Screenshot 2023-07-21 at 12 37 58 PM

Here's the previous behavior with text-opacity: 0.5 and blue text with a yellow halo -- note how it becomes green:
Screenshot 2023-07-21 at 12 37 35 PM

This adds a tiny amount of extra work to the symbol_sdf fragment shader, but in my experience a little extra math in shaders rarely has measurable perf impact. I don't see anything jump out from the benchmark results:

Screenshot 2023-07-21 at 11 51 01 AM
Screenshot 2023-07-21 at 11 45 39 AM
Screenshot 2023-07-21 at 11 45 33 AM

cc @makella @ibesora 

## Launch Checklist

 - [ ] Link to related issues.
 - [x] Post benchmark scores.


ChrisLoer:
Oops, my first take at this didn't pass tests because marking the "inside" of the halo as the edge of the text left a small overlap area where even in the fully opaque case, partially transparent text (from the edge anti-aliasing) could be drawn over a partially transparent inner halo edge, and you'd get some slight unexpected color mixing from beneath the halo. I fixed that by just making the halo wider on the inside by one `gamma` increment -- one side effect of this is that if you intended to use this intentionally to get an "outline" text effect, you'd probably have to make your font slightly bigger than you'd expect to get the effect you want.

The `icon-halo-color/opacity` test is I think actually a good example of what was broken before -- in the previous test expectation, the middle of the dot is slightly red, which doesn't really seem like what you'd expect.

HarelM:
Can you please add a change log entry?
Otherwise, this looks good.

wipfli:
@louwers is there an issue in Native for this?

wipfli:
As a rule of thumb I think whenever somebody edits a shader in MapLibre GL JS, we need an issue in Native. And vice versa...

louwers:
Some edits to shaders have no discernible changes to rendering output. A better rule of thumb would probably be when someone adds or edits a render test. Even better would be if this was automated somehow.

I'll have a look and create an issue.

louwers:
I don't see any green in the picture in the first post. Here is the current behavior of MapLibre Native:


actual

Issue #2896, opened 2023-07-21
MapLibre 3.2.0 Typing Errors


neodescis:
I'm trying to upgrade our project from 3.1.0 to 3.2.0, but I'm running into a few typing problems and cannot build against the new version.

1. The maplibre-gl.d.ts typings file now references supercluster. The supercluster package does not provide its own typings, and MapLibre only lists @types/supercluster as a dev dependency, so the typings for it are not installed with MapLibre. We have noImplicitAny set to true for our project, so this import fails to build. Can we elevate the @types/supercluster package to a regular dependency since it is now part of the public API?
2. In the exported Hash class, there is now a member function typed like this: `_updateHash: () => NodeJS.Timeout;`. Not only does this not build for me, but the typing looks incorrect since this code isn't running on Node. I'm not sure where the type is coming from though.
3. The generated typing for event handlers is broken because the order of the signatures of the Map's "on" function changed, with the most generic version now being listed first. In some cases I'm now seeing the type of my handler function as "any", and in some cases I'm seeing it as all-the-things. Manually changing the declaration order in the .d.ts file back to what it was fixes this.

**maplibre-gl-js version**: 3.2.0

**browser**: N/A

neodescis:
I'm happy to work a PR for these. @HarelM , would you have any idea where the typing for #2 above is coming from?

HarelM:
No clue, maybe an issue with dts bundle generator or the tsconfig, but I'm just guessing...

HarelM:
Feel free to push this forward, I'm planning a new release soon, let me know if you would like to push this through fast so it will be included.

neodescis:
I can at least knock out 1 and 3. I'll investigate 2 further and see if I can figure it out.

HarelM:
I don't see these issues in any of the libraries I maintain, so it's worth making sure this is covered in tests somehow, much like I did with the tsconfig for the dist `d.ts` file.

neodescis:
I'll have a PR up in a sec and we can discuss there... not sure how 1 and 2 would be covered by tests.

PR #2898, opened 2023-07-21
Fix typings issues in 3.2.0

neodescis:
Closes https://github.com/maplibre/maplibre-gl-js/issues/2896.

- Fixed inferred typing of event data in map.on(), map.off() and map.once()
- Elevated @types/supercluster to a regular dependency
- Specified the typing for _updateHash rather than let it be inferred (since it is inferred to `() => NodeJS.Timeout`)

## Launch Checklist

 - [X] Link to related issues.
 - [ ] Post benchmark scores.


package.json
```diff
@@ -25,6 +25,7 @@
     "@types/mapbox__point-geometry": "^0.1.2",
     "@types/mapbox__vector-tile": "^1.3.0",
     "@types/pbf": "^3.0.2",
+    "@types/supercluster": "^7.1.0",
```
neodescis, line 28: This should be an actual dependency since it's now part of the API surface.

src/ui/hash.ts
```diff
@@ -133,6 +133,5 @@ export class Hash {
     /**
      * Mobile Safari doesn't allow updating the hash more than 100 times per 30 seconds.
      */
-    _updateHash = throttle(this._updateHashUnthrottled, 30 * 1000 / 100);
-
+    _updateHash: () => ReturnType<typeof setTimeout> = throttle(this._updateHashUnthrottled, 30 * 1000 / 100);
```
neodescis, line 136: This was being inferred to () => NodeJS.Timeout during typing generation, because MapLibre installs @types/node. That package overrides the type of the return value of setTimeout from `number` to NodeJS.Timeout. If using a project without @types/node, though, this breaks. Specifying the type here instead of inferring it allows it to work either way.

src/ui/map.ts
```diff
@@ -1411,13 +1411,13 @@ export class Map extends Camera {
...
      * @see [Create a draggable marker](https://maplibre.org/maplibre-gl-js/docs/examples/drag-a-point/)
      */
-    on(type: keyof MapEventType | string, listener: Listener): this;
     on<T extends keyof MapLayerEventType>(
         type: T,
         layer: string,
         listener: (ev: MapLayerEventType[T] & Object) => void,
     ): Map;
     on<T extends keyof MapEventType>(type: T, listener: (ev: MapEventType[T] & Object) => void): this;
+    on(type: keyof MapEventType | string, listener: Listener): this;
```
neodescis, line 1420: Reordering so that the most generic version comes last.

src/ui/map_events.test.ts
```diff
@@ -4,6 +4,14 @@ import {createMap, beforeMapTest} from '../util/test/util';
 import {MapGeoJSONFeature} from '../util/vectortile_to_geojson';
 import {MapLayerEventType, MapLibreEvent} from './events';
 
+type IsAny<T> =
```
neodescis, line 7: Source: https://stackoverflow.com/questions/58778637/assert-type-is-not-any

This seems like a very useful test utility. Should it go somewhere else?

src/ui/hash.ts
neodescis: It seems like we should have a check somewhere that the generated typings don't reference NodeJS at all?

src/ui/map_events.test.ts
```diff
@@ -4,6 +4,10 @@ import {createMap, beforeMapTest} from '../util/test/util';
 import {MapGeoJSONFeature} from '../util/vectortile_to_geojson';
 import {MapLayerEventType, MapLibreEvent} from './events';
 
+type IsAny<T> = 0 extends T & 1 ? T : never;
```
neodescis, line 7: Source: https://medium.com/frontend-canteen/type-challenges-implement-the-isany-t-utility-type-6a4babcf1da7

Seems like a useful utility. Should it go somewhere else?

src/ui/map_events.test.ts
```diff
@@ -170,6 +174,20 @@ describe('map events', () => {
         expect(handler.onMove).toHaveBeenCalledTimes(1);
     });
 
+    test('Map#on allows a listener to infer the event type ', () => {
+        const map = createMap();
+
+        const spy = jest.fn();
+        map.on('mousemove', (event) => {
+            assertNotAny(event);
```
neodescis, line 182: This does not build if "event" is of type "any"

src/ui/map.ts
HarelM: The reorder here was for the docs site, in order to have the method seen as a readable. 
Not sure what's the solution here...

src/ui/map.ts
```diff
@@ -1460,13 +1460,13 @@ export class Map extends Camera {
...
      * @returns `this` if listener is provided, promise otherwise to allow easier usage of async/await
      */
-    once(type: keyof MapEventType | string, listener?: Listener): this | Promise<any>;
     once<T extends keyof MapLayerEventType>(
         type: T,
         layer: string,
         listener?: (ev: MapLayerEventType[T] & Object) => void,
     ): this | Promise<MapLayerEventType[T] & Object>;
     once<T extends keyof MapEventType>(type: T, listener?: (ev: MapEventType[T] & Object) => void): this | Promise<any>;
+    once(type: keyof MapEventType | string, listener?: Listener): this | Promise<any>;
```
HarelM, line 1469: Same here... The tsdoc comment should be on the relevant method definition.

src/ui/map_events.test.ts
HarelM: We can move it if it's needed somewhere what I guess, for now it's OK here.

src/ui/hash.ts
HarelM: Can we configure the tsconfig.dist not to include node types or something like that?

HarelM:
Overall this looks good, thanks!
Added a few minor comments.

HarelM:
Also can you ping me privately on maplibre's Slack ("Harel Mazor" is my handle there)?

build/generate-typings.ts
```diff
@@ -8,7 +8,7 @@ if (!fs.existsSync('dist')) {
 
 console.log('Starting bundling types');
 const outputFile = './dist/maplibre-gl.d.ts';
-childProcess.execSync(`dts-bundle-generator --umd-module-name=maplibregl -o ${outputFile} ./src/index.ts`);
+childProcess.execSync(`dts-bundle-generator --umd-module-name=maplibregl -o ${outputFile} ./src/index.ts  --project tsconfig.typegen.json`);
```
neodescis, line 11: Specify new tsconfig for generating typings

src/ui/hash.ts
neodescis: Addressed! I did have to make another tsconfig though...

neodescis: I did test using the new tsconfig by reverting this change, and it produces "number" as I'd expect (the return type for a browser's setTimeout()). I think this change should stay though, as this is the typing for a bunch of other setTimeout return values in this file.

src/ui/map.ts
neodescis: Well, it wasn't on the relevant method definition though, since it was referring to a param that doesn't exist on that definition (layer). It looks like tsdoc itself supports documentation on each overload though, so I've gone through and add them for on, off and once.

Also, I should note that while tsdoc itself supports documentation on each overload (confirmed by it showing for each signature in vscode), it appears that our document generation only picks up the first one. However, I think the wording on the first signature of each of these functions does a good enough job explaining that the second parameter is optional.

tsconfig.typegen.json
```diff
@@ -0,0 +1,9 @@
+{
```
neodescis, line 1: New tsconfig for generating typings, which does not include node or jest that are in the base config

src/ui/map.ts
neodescis: Also, I did manage to dig this up:  https://github.com/microsoft/TypeScript/issues/1860

The TypeScript compiler picks up the first matching signature, so we do need to have the generic signature last.

build/generate-typings.ts
HarelM: There are two spaces here - should be a single one.
Can't we use the tsconfid.dist.json for this?

src/ui/hash.ts
HarelM: Sure, being explicit is better when it comes to types I believe.

src/ui/map.ts
HarelM: The docs sites picks all of them, I was hoping to make the first `on` and `off` documentation as the more simple definition, that's why I moved it...
Can you explain why the "more complicated" one should be the first?

Here are the docs:
https://maplibre.org/maplibre-gl-js/docs/API/classes/maplibregl.Map/#on

build/generate-typings.ts
HarelM: I now see that you renamed it, it's not a different one.
Can you clarify why?
I named it originally since it only checks the d.ts file that is part of the dist folder.

HarelM: No, this is confusing... :-(

tsconfig.typegen.json
HarelM: If we are adding a tsconfig I would prefer it the other way around - create a config for the tests and keep the "main" config without `jest` and `node`, so that dts-bundle-generator uses the same one as when building the project, and when running all kind of scripts and tests use a dedicated one.
I think I saw it in some projects and I think it makes more sense. It a thought, I'm not too attached to that solution.

build/generate-typings.ts
HarelM: See my comments on the thread in the new tsconfig file.

HarelM:
Added a few more comments, thanks for bearing with me.

src/ui/map.ts
neodescis: Sorry, I guess I didn't read far enough! You're right, it does pick them all up.

It's not that the "most complicated" signature should be first, it's just that TypeScript picks up the first matching signature during a build. So, if you have a more generic signature (e.g. one with "any") first, then TypeScript will use that instead of one further down the list with better typing (as mentioned in the TypeScript issue I referenced). This is what I'm seeing in the current state of things.

tsconfig.typegen.json
neodescis: Sure, I agree that would be better. I'll go down that path and see how it works. I couldn't just use the existing "dist" tsconfig though, because it has strict: true in it.

tsconfig.dist.json
```diff
@@ -1,7 +1,7 @@
 {
```
neodescis, line 1: Undid the rename on this file; just fixed the extraneous comma now

tsconfig.json
```diff
@@ -21,15 +21,18 @@
       "DOM.Iterable"
     ],
     "types": [
```
neodescis, line 23: Removed both "node" and "jest." I think ts-jest must add in "jest" if it's not there, because tests still work.

tsconfig.json
```diff
@@ -43,5 +46,7 @@
     "build/rollup",
     "test/bench/rollup",
     "test/bench/**/*generated*",
+    "**/*.test.ts",
```
neodescis, line 49: Exclude test files and utils from the main build.

tsconfig.json
```diff
@@ -21,15 +21,18 @@
...
-      "jest",
       "geojson",
       "offscreencanvas"
     ]
   },
   "ts-node": {
     "experimentalSpecifierResolution": "node",
-    "esm": true
+    "esm": true,
+    "compilerOptions": {
```
neodescis, line 31: Add in the "node" types just for running ts-node. I think this is great, because it truly walls off the Node types from everything else.

tsconfig.json
neodescis: Also... I was afraid that excluding these here would yield "errors" in these files when pulling them up in an editor, but that is not the case for me with vscode; I still see jest typings when editing them.

HarelM: Does this mean that the typecheck run in the package.json won't check that the tests files are typed correctly?
The test you added are basically a typecheck, right? There is also other tests in the style spec expression that checks only typings.

tsconfig.typegen.json
neodescis: Gone!

tsconfig.json
HarelM: Looks good!

build/generate-typings.ts
neodescis: Reverted

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 - Incorrect distance field of view calculation for negative elevation, fixed by storing min elevation for the tile in view ([#1655](https://github.com/maplibre/maplibre-gl-js/issues/1655), [#2858](https://github.com/maplibre/maplibre-gl-js/pull/2858))
 - Fix reloadCallback not firing on VectorTileWorkerSource.reloadTile ([#1874](https://github.com/maplibre/maplibre-gl-js/pull/1874))
 - Fix RasterDEMTileSource not serializing its options correctly ([#2895](https://github.com/maplibre/maplibre-gl-js/pull/2895))
+- Remove node and jest from tsconfig types, fix map event and other typing problems ([https://github.com/maplibre/maplibre-gl-js/pull/2898](https://github.com/maplibre/maplibre-gl-js/pull/2898))
```
HarelM, line 14:
```suggestion
- Remove node and jest from tsconfig types, fix map event and other typing problems ([#2898](https://github.com/maplibre/maplibre-gl-js/pull/2898))
```

HarelM (approved):
Overall approved, thanks!

tsconfig.json
neodescis: The "typecheck" script currently only checks the generated .d.ts file. However, I would expect that ts-jest would error on typescript errors, but that doesn't appear to be happening. I'm investigating now.

HarelM: It runs tsc with no emit, so that checked the tests files' typescript correctness before this change I believe.

neodescis: Ah, you are correct. I missed that. It seems that ts-jest is not spitting out typing errors because of 'isolatedModules': true, but turning that off is yielding some errors that don't yet make sense to me. That feels like the right place to do the type checking for tests, though, if I can get it to work. What do you think? An alternative would be to create a tsconfig for tests and add that to the typecheck script.

HarelM: Angular way is what I had in mind to some extent, but when you put it like that it seems too complicated.
I would recommend finding the simplest solution at this point in time (maybe a tsconfig for tests only) and then continue to try and find a more elegant solution if one exists so that we don't hold this PR hostage.

tsconfig.dist.json
```diff
@@ -2,6 +2,10 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "strict": true,
+    "types": [
```
neodescis, line 5: Only allow these global types in the dist type check

tsconfig.json
```diff
@@ -21,15 +21,19 @@
       "DOM.Iterable"
     ],
     "types": [
-      "node",
-      "jest",
       "geojson",
-      "offscreencanvas"
+      "offscreencanvas",
+      "jest"
```
neodescis, line 26: Leaving jest in here at least for now (and I undid the change to exclude tests from the tsconfig). If we want to split out into build and test tsconfigs, we can still do so, but I think maybe it's not necessary given my change to tsconfig.dist.json.

neodescis:
I've tested everything out I can think of:
- NodeJS global types are not included in the generated .d.ts file
- Even if I put a NodeJS global type in manually, the typecheck script fails
- The typecheck script fails on type errors in tests, including the assertNotAny() check I've added

HarelM:
Thanks! Feel free to create a version bump PR so we can release a new version with these fixes.

PR #2910, opened 2023-07-24
Feature/support cache parameter in fetch request

marucjmar:
## Launch Checklist

In my case, I need it to use a **force-cache** strategy in **fetch** API when downloading map tiles. 

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - Remove cooperative gesture screen from the accessibility tree since screenreaders cannot interact with the map using gestures
 - Add `cooperated gestures` example to the doc.([#2860](https://github.com/maplibre/maplibre-gl-js/pull/2860))
-- Add `fetchCache` to [`RequestParameters`](https://maplibre.org/maplibre-gl-js/docs/API/types/maplibregl.RequestParameters/)
+- Add `cache` parameter to [`RequestParameters`](https://maplibre.org/maplibre-gl-js/docs/API/types/maplibregl.RequestParameters/) ([#2910](https://github.com/maplibre/maplibre-gl-js/pull/2910))
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Remove cooperative gesture screen from the accessibility tree since screenreaders cannot interact with the map using gestures
 - Add `cooperated gestures` example to the doc.([#2860](https://github.com/maplibre/maplibre-gl-js/pull/2860))
+- Add `fetchCache` to [`RequestParameters`](https://maplibre.org/maplibre-gl-js/docs/API/types/maplibregl.RequestParameters/)
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
--- a/src/util/ajax.test.ts
+++ b/src/util/ajax.test.ts
@@ -7,7 +7,7 @@ import {
 } from './ajax';
 
 import {fakeServer, FakeServer} from 'nise';
-import {fetchMock, RequestMock, setupFetchMock} from './test/mock_fetch';
+import {RequestMock, setupFetchMock} from './test/mock_fetch';
 
 function readAsText(blob) {
     return new Promise((resolve, reject) => {
--- a/src/util/ajax.test.ts
+++ b/src/util/ajax.test.ts
@@ -7,6 +7,7 @@ import {
 } from './ajax';
 
 import {fakeServer, FakeServer} from 'nise';
+import {setupFetchMock} from './test/mock_fetch';
 
 function readAsText(blob) {
     return new Promise((resolve, reject) => {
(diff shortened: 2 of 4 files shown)


HarelM:
Can you please add a unit test for that?

marucjmar:
ok, I added unit tests. See if such mocking of the fetch intefrejs is ok. 
I wanted to use [jest-fetch-mock](https://www.npmjs.com/package/jest-fetch-mock) library but this library uses [whatwg-fetch](https://github.com/JakeChampion/fetch), which does not support all Fetch standard options.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Remove cooperative gesture screen from the accessibility tree since screenreaders cannot interact with the map using gestures
 - Add `cooperated gestures` example to the doc.([#2860](https://github.com/maplibre/maplibre-gl-js/pull/2860))
+- Add `fetchCache` to [`RequestParameters`](https://maplibre.org/maplibre-gl-js/docs/API/types/maplibregl.RequestParameters/)
```
HarelM, line 7: Can you link the PR like the other entries?

src/util/ajax.ts
```diff
@@ -49,6 +49,10 @@ export type RequestParameters = {
      * If `true`, Resource Timing API information will be collected for these transformed requests and returned in a resourceTiming property of relevant data events.
      */
     collectResourceTiming?: boolean;
+    /**
+     * Parameters supported only by browser fetch API. Property of the Request interface contains the cache mode of the request. It controls how the request will interact with the browser's HTTP cache. (https://developer.mozilla.org/en-US/docs/Web/API/Request/cache)
+     */
+    fetchCache?: RequestCache;
```
HarelM, line 55: shouldn't this be called `cache` to keep the naming consistent?

src/util/test/mock_fetch.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    constructor(input: RequestInfo | URL, init?: RequestInit) {
+        this.cache = typeof input === 'object' && 'cache' in input ? input.cache : init.cache;
+        this.method = typeof input === 'object' && 'method' in input ? input.method : init.method;
+        this.url = typeof input === 'object' && 'url' in input ? input.url : input.toString();
+        this.headers = typeof input === 'object' && 'headers' in input ? new Headers(input.headers) : new Headers(init.headers || {});
+    }
+}
+
+class AbortController {
```
HarelM, line 19: `AbortControllerMock`...?

src/util/ajax.test.ts
```diff
@@ -43,6 +44,31 @@ describe('ajax', () => {
...
     });
 
+    [
+        {functionName: 'getJSON', targetFunction: getJSON},
+        {functionName: 'getArrayBuffer', targetFunction: getArrayBuffer}
+    ].forEach(({targetFunction, functionName}) => {
+        test(`${functionName} set correct request parameters`, (done) => {
+            setupFetchMock();
+
+            jest.spyOn(global, 'fetch').mockImplementation((requestInfo: Request) => {
```
HarelM, line 54: This looks odd - if you setup fetch mock two lines above, why would you need to set a spy.
Shouldn't the `setupFetchMock` return a `spy`?

src/util/ajax.test.ts
```diff
@@ -43,6 +44,31 @@ describe('ajax', () => {
         server.respond();
     });
 
+    [
```
HarelM, line 47: I would prefer to keep the tests a bit more "moist" and add two tests here instead of doing this foreach.

src/util/test/mock_fetch.ts
```diff
@@ -0,0 +1,39 @@
...
+export function setupFetchMock(): jest.Mock<Promise<Response>, [input: RequestInfo | URL, init?: RequestInit], any> {
+    const fetchMock = jest.fn(async (_input: RequestInfo | URL, _init?: RequestInit): Promise<Response> => {
+        return <Response>{
+            json: async () => null,
+            arrayBuffer: async () => null,
+            text: async () => null,
+        };
+    });
+
+    global.AbortController = AbortControllerMock;
```
HarelM, line 34: Does this part requires an "uninitialize" method to revert stuff to how they used to be?
i.e. will this might create a side effect for other unit tests?

marucjmar: Good catch, I added a function that reset state to default after use `setupFetchMock()`

src/util/ajax.test.ts
```diff
@@ -25,6 +26,19 @@ describe('ajax', () => {
     });
     afterEach(() => {
         server.restore();
+        destroyFetchMock();
```
HarelM, line 29: If this is done in the `afterEach`, I would expect the opposite to happen in the `beforeEach`.
Alternatively, I believe you can create a before and after each for specific tests that use this kind of mocking.

HarelM:
According to the following it seems that not all parts of the mock files are needed:
https://app.codecov.io/gh/maplibre/maplibre-gl-js/pull/2910?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=maplibre
image

HarelM (approved):
THANKS!

HarelM:
Oh, forgot about the coverage part... Can you check that too?

marucjmar:
I checked coverage.

The `abort()` method in `AbortControllerMock` class is required by `AbortController` type. if I wanted to remove it I would have to force override type of `AbortControllerMock` class.

The `signal` getter in `RequestMock` class is required because this [line](https://github.com/maplibre/maplibre-gl-js/blob/d27b43de876441c4ac517e014f6531a73da0ba16/src/util/ajax.ts#L256) check this property exists in prototype.

The rest of the death code has been removed.

HarelM:
Ok, cool.

PR #2914, opened 2023-07-25
Implement text-variable-anchor-offset

drwestco:
## Launch Checklist

Consume new text-variable-anchor-offset property from style spec 19.3.0 (https://github.com/maplibre/maplibre-style-spec/issues/112). Add methods to treat text-variable-anchor and text-variable-anchor-offset in a consistent manner. Add render tests (copied/modified from existing text-variable-anchor render tests) to verify consistency.

 - [X] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
Uh... That will take me a few days to go though... 😀

test/integration/render/tests/text-variable-anchor-offset/all-anchors/style.json
```diff
@@ -0,0 +1,122 @@
...
+  "layers": [
+    {
+      "id": "text",
+      "type": "symbol",
+      "source": "point",
+      "layout": {
+        "text-field": "x",
+        "text-size": 20,
+        "text-justify": "auto",
+        "text-variable-anchor-offset": [
```
HarelM, line 95: The original request was for an icon that is placed above a point and to have the label surround it properly.
I don't see this kind of test here, this is the closest I found.

test/examples/variable-offset-label-placement.html
```diff
@@ -0,0 +1,132 @@
+<!DOCTYPE html>
```
HarelM, line 1: There is a need to add an image to the example, there's a script that generates this image, or you can open this page in the browser and do a screen shot.

src/symbol/symbol_layout.ts
```diff
@@ -789,9 +714,9 @@ function addSymbol(bucket: SymbolBucket,
         useRuntimeCollisionCircles,
         0,
         textBoxScale,
-        textOffset0,
-        textOffset1,
-        collisionCircleDiameter);
+        collisionCircleDiameter,
```
HarelM, line 717: Looking at the original code, the order of the parameters here is different, is this intended?

src/style/style_layer/symbol_style_layer.ts
```diff
@@ -193,3 +195,162 @@ export function getIconPadding(layout: PossiblyEvaluated<SymbolLayoutProps, Symb
...
+
+// The radial offset is to the edge of the text box
+// In the horizontal direction, the edge of the text box is where glyphs start
+// But in the vertical direction, the glyphs appear to "start" at the baseline
+// We don't actually load baseline data, but we assume an offset of ONE_EM - 17
+// (see "yOffset" in shaping.js)
+const baselineOffset = 7;
+export const INVALID_TEXT_OFFSET = Number.POSITIVE_INFINITY;
+
+export function evaluateVariableOffset(anchor: TextAnchor, offset: [number, number]): [number, number] {
```
HarelM, line 221: Why was this moved here?
If this is a "subject" of its own, I would consider moving it to a separate file all together.

HarelM: I think overlap mode is somewhat similar in terms of having its own file, doesn't it?

src/style/style_layer/symbol_style_layer.ts
```diff
@@ -193,3 +195,162 @@ export function getIconPadding(layout: PossiblyEvaluated<SymbolLayoutProps, Symb
         values[3] * pixelRatio,
     ];
 }
+
+export enum TextAnchorEnum {
```
HarelM, line 199: Why do we need an enum here? Are we passing this number to the shaders?

src/symbol/placement.ts
```diff
@@ -952,11 +964,12 @@ export class Placement {
...
         if (bucket.hasTextCollisionBoxData()) bucket.textCollisionBox.collisionVertexArray.clear();
 
-        const layout = bucket.layers[0].layout;
+        const layer = bucket.layers[0];
+        const layout = layer.layout;
         const duplicateOpacityState = new JointOpacityState(null, 0, false, false, true);
         const textAllowOverlap = layout.get('text-allow-overlap');
         const iconAllowOverlap = layout.get('icon-allow-overlap');
-        const variablePlacement = layout.get('text-variable-anchor');
+        const hasVariablePlacement = layer._unevaluatedLayout.hasValue('text-variable-anchor') || layer._unevaluatedLayout.hasValue('text-variable-anchor-offset');
```
HarelM, line 972: Why are we using `_unevaluatedLayout` here?

src/symbol/placement.ts
```diff
@@ -561,47 +566,54 @@ export class Placement {
...
-                                textBoxScale, rotateWithMap, pitchWithMap, textPixelRatio, posMatrix,
-                                collisionGroup, overlapMode, symbolInstance, bucket, orientation, variableIconBox, getElevation);
-
-                            if (result) {
-                                placedBox = result.placedGlyphBoxes;
-                                if (placedBox && placedBox.box && placedBox.box.length) {
-                                    placeText = true;
-                                    shift = result.shift;
-                                    break;
+                        let placementPasses = (textOverlapMode === 'never') ? 1 : 2;
```
HarelM, line 582: Are there unit tests I'm missing that are covering this part? If not, I suggest to try and add some, this is clearly something unit tests can check well, with all kinds of edge cases that render tests will probably not cover.

src/style/style_layer/symbol_style_layer.ts
HarelM: All this code also need to be covered with unit tests.

src/data/bucket/symbol_attributes.ts
```diff
@@ -100,8 +100,9 @@ export const symbolInstance = createLayout([
     {type: 'Uint16', name: 'useRuntimeCollisionCircles'},
     {type: 'Uint32', name: 'crossTileID'},
     {type: 'Float32', name: 'textBoxScale'},
-    {type: 'Float32', components: 2, name: 'textOffset'},
     {type: 'Float32', name: 'collisionCircleDiameter'},
+    {type: 'Uint16', name: 'textAnchorOffsetStartIndex'},
```
HarelM, line 104: Where are those used? is this definition used for serialization to the worker or used for the shaders?

HarelM:
I have more questions than answers unfortunately.
Maybe a short design diagram would help understand how all this works together - which class is passing which data to where and who uses this data.
THANKS!!

src/data/bucket/symbol_attributes.ts
drwestco: Worker serialization.

src/render/draw_symbol.ts
```diff
@@ -281,7 +279,7 @@ function drawLayerSymbols(
 
     const depthMode = painter.depthModeForSublayer(0, DepthMode.ReadOnly);
 
-    const variablePlacement = layer.layout.get('text-variable-anchor');
+    const hasVariablePlacement = layer._unevaluatedLayout.hasValue('text-variable-anchor') || layer._unevaluatedLayout.hasValue('text-variable-anchor-offset');
```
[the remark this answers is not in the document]
drwestco: This was how the original code was testing for the existence of a property. Depending on how the defaults in the style spec are set, I think using `layout.get` will return something else?
```TypeScript
if (layer._unevaluatedLayout.getValue('text-radial-offset') === undefined) { ... }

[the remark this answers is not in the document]
drwestco: Also, I was keen to avoid creation of the anchor offset collection for a simple existence check, though that may not happen until you actually call `.evaluate`?

src/style/style_layer/symbol_style_layer.ts
drwestco: It gets passed to/from the worker. I didn't see an obvious way to pass the existing anchor values as strings.

test/integration/render/tests/text-variable-anchor-offset/all-anchors/style.json
drwestco: Added

src/symbol/symbol_layout.ts
drwestco: Previously, `text-offset` and `text-radial-offset` was read and converted from EM to PX, and stored in symbol_bucket at lines 605-612. It was stored as two numeric properties - textOffset0, textOffset1. These values were only read back out if `text-variable-anchor` is also in play. Now, since there may be a separate offset per variable anchor, a single pair of numeric properties doesn't suffice. So there's a new serialized array that stores the anchor+offset pairs.

src/style/style_layer/symbol_style_layer.ts
drwestco: Sure. Will move to new file and add unit tests.

src/symbol/placement.ts
drwestco: I think in the case of `text-variable-anchor` and `text-variable-anchor-offset`, using `layout.get` vs `_unevaluatedLayout.hasValue` is equivalent, since neither have a default value and will thus return `undefined`. I ran into existing code that failed to differentiate between an explicit 0 value in the style vs implicit 0 value for an undefined property. Using _unevaluatedLayout seems more reliable here.

src/symbol/placement.ts
drwestco: I don't see any unit tests for placement.ts. Refactoring this class so the internals are exposed for unit testing is well beyond the scope of this PR.

HarelM: 😔

drwestco:
Added the preview image for the new sample. I think that's all the PR comments addressed?

HarelM:
Change log entry is missing.
I think that it.

drwestco:
> Change log entry is missing. I think that it.

Should I also bump version number in package.json? 3.3.0 maybe?

HarelM:
In a different PR please.

HarelM (approved):

birkskyum:
Does this close  https://github.com/maplibre/maplibre-style-spec/issues/112 ?

HarelM:
Depends on how we want to track native parity.

birkskyum:
Aaaaah, gotcha. Was there a repro of this feature made for testing that can be converted into a docs example or to present it in the newsletter?

HarelM:
Yes, see the example file in this PR:
https://maplibre.org/maplibre-gl-js/docs/examples/variable-offset-label-placement/

PR #1874, opened 2022-11-25
VectorTileWorkerSource: change reloadTile behavior to reload right away instead of waiting for previous parse to complete

ambientlight:
As pointed out in https://github.com/maplibre/maplibre-gl-js/pull/1805#issuecomment-1315978349, the issue occurred where after adding a geojson source, a bunch of symbol layers and new spritesheets in a `load` callback, the sprites would be not shown for geojson's source tiles already loaded (which are loaded before the sprites are loaded), as a result of tiles not being reparsed:

https://github.com/maplibre/maplibre-gl-js/blob/870c2a38550a363ca62833a2f401f599d2720f97/src/source/worker_tile.ts#L64

not getting called inside the web-worker and so the https://github.com/maplibre/maplibre-gl-js/blob/870c2a38550a363ca62833a2f401f599d2720f97/src/source/worker_tile.ts#L169 would not fire to grab the new loaded image assets. 

This issue would not happen if browser have cached sprite assets, hinting on some async behavior weirdness. Tracing the flow of https://github.com/maplibre/maplibre-gl-js/blob/870c2a38550a363ca62833a2f401f599d2720f97/src/style/style.ts#L503

happened to show that https://github.com/maplibre/maplibre-gl-js/blob/870c2a38550a363ca62833a2f401f599d2720f97/src/source/vector_tile_worker_source.ts#L157-L167 done callback that would perform reparsing and would be set if the workerTile is already in the `parsing` state would not fire. This happened because the workerTile.reloadCallback would only happen if the `workerTile.parse` (that sets the status to `parsing`) would itself be called from `reloadTile`, but would not from its own `loadTile` logic that could happen prior to this `reloadTile` call.

This PR fixes that by checking the reloadCallback inside the `loadTile`'s `workerTile.parse` callback.

# **UPDATE: 2023/07/15**

reload callback has been eliminated per conversation in favor of reloading right away and cancelling in flight dependency sprite / glyphs dependency loading during parse at https://github.com/maplibre/maplibre-gl-js/blob/main/src/source/worker_tile.ts#L141:L177. 
## Launch Checklist

 - [X] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [X] Briefly describe the changes in this PR.
 - [X] Link to related issues.
 - [x] Write tests for all new functionality.
 - [x] Add an entry to `CHANGELOG.md` under the `## main` section.

PR #2941, opened 2023-08-05
VectorTileWorkerSource: fix reload for original's load parse would not pass the rawTileData and meta

### LINKED ISSUE #2923: `queryRenderFeature` emits `Out of bounds.` Error in 3.2.1 version
Hi,

I noticed that in version 3.2.1 there is an "Out of bounds" error. The error does not always occur because race condition occurs in the code. After a brief investigation, I determined that a change from this PR #1874 caused the problem. 

**maplibre-gl-js version**: 3.2.1

**browser**: all

### Steps to Trigger Behavior

 1. Init map with vector styles
 2. Add empty GeoJSON source to map after load event
 3. Add symbol layer with created source
 4. Add mouse event listener
 5. Update the source by `setData` method
 6. Trigger mouse event on symbol from this layer or use `queryRenderedFeatures` on layer

### Link to Demonstration

[fiddle](https://jsfiddle.net/fpt35enj/)

https://github.com/maplibre/maplibre-gl-js/assets/5881368/df665528-bdd5-470d-9f35-9b1d9961eb7b

### Expected Behavior
No error

### Actual Behavior
`Out of bounds.` Error

ambientlight:
## Launch Checklist

Fixes #2923.

An unfortunate regression has been introduced in #1874, where when `reloadTile`  would get called before the original load completes, the `rawTileData` won't get passed to the main thread, leading to #2923. The base `loadTile` will extend the parse result with previously fetched data at https://github.com/maplibre/maplibre-gl-js/blob/01e5f51cf15421cb40ad0ff4798f8f536f816a0c/src/source/vector_tile_worker_source.ts#L130

This PR will keep the original fetching state until the first parse completes and will pass it along as expected.

 - [X] Link to related issues.

### DIFF
--- a/src/source/vector_tile_worker_source.ts
+++ b/src/source/vector_tile_worker_source.ts
@@ -24,6 +24,12 @@ export type LoadVectorTileResult = {
     resourceTiming?: Array<PerformanceResourceTiming>;
 } & ExpiryData;
 
+type FetchingState = {
+    rawTileData: ArrayBuffer;
+    cacheControl: ExpiryData;
+    resourceTiming: any;
+}
+
 /**
  * The callback when finished loading vector data
  */
@@ -52,12 +58,6 @@ function loadVectorTile(params: WorkerTileParameters, callback: LoadVectorDataCa
         request.cancel();
         callback();
     };
-}
-
-interface FetchingState {
-    rawTileData: ArrayBuffer;
-    cacheControl: ExpiryData;
-    resourceTiming: any;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


HarelM:
Can you fix the lint warning?

src/source/vector_tile_worker_source.ts
```diff
@@ -54,6 +54,12 @@ function loadVectorTile(params: WorkerTileParameters, callback: LoadVectorDataCa
     };
 }
 
+interface FetchingState {
```
HarelM, line 57: Can this be a type instead? I aim to use interfaces for things people can inherit publicly.

HarelM:
I see that a "else" part of this fix is not covered by a test, can you add one (see the linked coverage report)?

HarelM:
Thanks a lot for taking care of it in such a short notice!!

ambientlight:
@HarelM: done, thanks!

HarelM (approved):

Issue #2971, opened 2023-08-11
White flickering on resize in version 3.x.x


predirector:
**maplibre-gl-js version**:
3.3.0 and 3.2.0
**browser**:
Firefox & Chrome
### Steps to Trigger Behavior

 1. Resize your browser with a full screen map

### Expected Behavior
Does not flicker. 
Especially because the behaviour in 2.4.0 on resize was better. Only the resized area was rerendered but not the complete map. 

### Actual Behavior
In Firefox is flickers very heavy on resize. In Chrome it gets completely white for a short time.

Thank you

HarelM:
This has changed recently to better support container resize and not just window resize.
The following PR is the relevant one I believe: #2157.
It might be that there's a need to throttle this event, IDK.
Feel free to suggest a solution.

neodescis:
It does seem to behave a lot nicer if a debounce is used; I tried it with 50 milliseconds. Should it be configurable?

HarelM:
I'm not sure this should be configurable, but then hardcoded stuff tend to make their way to being configurable... IDK...

neodescis:
I agree. I'll start with 25 or 50, whatever seems like it works best, and we can go from there.

PR #2973, opened 2023-08-12
Debounce ResizeObserver's call to map.resize()

neodescis:
- Closes https://github.com/maplibre/maplibre-gl-js/issues/2971
- Debounces the map's ResizeObserver's call to map.resize(), to 30 milliseconds
- Leaves ResizeObserver's leading edge calls (i.e. initial calls before debouncing) to map.resize() in place, meaning that the first resizes before debouncing are immediate
- Resizing in Chrome and Firefox behaves considerably better, with much less flicker

## Launch Checklist

 - [X] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -648,19 +648,33 @@ export class Map extends Camera {
         if (typeof window !== 'undefined') {
             addEventListener('online', this._onWindowOnline, false);
             let initialResizeEventCaptured = false;
-            let resizeDebounceTimer: ReturnType<typeof setTimeout>;
+            let resizeDebounceTimerId: ReturnType<typeof setTimeout> | null = null;
             const resizeDebounceTime = 30;
+            const shouldResize = () => this._trackResize && !this._removed;
             this._resizeObserver = new ResizeObserver((entries: ResizeObserverEntry[]) => {
                 if (!initialResizeEventCaptured) {
                     initialResizeEventCaptured = true;
                     return;
                 }
 
-                if (resizeDebounceTimer) {
-                    clearTimeout(resizeDebounceTimer);
+                // Debounce resize call, but with leading edge call invoked immediately
+                let resizedImmediately = false;
+                if (resizeDebounceTimerId) {
+                    clearTimeout(resizeDebounceTimerId);
+                } else if (shouldResize()) {
+                    resizedImmediately = true;
+                    this.resize(entries)._update();
+                    resizeDebounceTimerId = setTimeout(() => {
+                        resizeDebounceTimerId = null;
+                    }, resizeDebounceTime);
                 }
-                if (this._trackResize && !this._removed) {
-                    resizeDebounceTimer = setTimeout(() => this.resize(entries)._update(), resizeDebounceTime);
+                if (!resizedImmediately && shouldResize()) {
+                    resizeDebounceTimerId = setTimeout(() => {
+                        if (shouldResize()) {
+                            this.resize(entries)._update();
+                        }
+                        resizeDebounceTimerId = null;
+                    }, resizeDebounceTime);
                 }
             });
             this._resizeObserver.observe(this._container);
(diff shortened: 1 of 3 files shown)


src/ui/map.ts
```diff
@@ -638,14 +638,33 @@ export class Map extends Camera {
         if (typeof window !== 'undefined') {
             addEventListener('online', this._onWindowOnline, false);
             let initialResizeEventCaptured = false;
-            this._resizeObserver = new ResizeObserver((entries) => {
+            let resizeDebounceTimerId: ReturnType<typeof setTimeout> | null = null;
+            const resizeDebounceTime = 30;
+            const shouldResize = () => this._trackResize && !this._removed;
```
neodescis, line 643: Calling resize() is no longer synchronous, so need to check if the map was removed each time

src/ui/map.ts
```diff
@@ -638,14 +638,33 @@ export class Map extends Camera {
         if (typeof window !== 'undefined') {
             addEventListener('online', this._onWindowOnline, false);
             let initialResizeEventCaptured = false;
-            this._resizeObserver = new ResizeObserver((entries) => {
+            let resizeDebounceTimerId: ReturnType<typeof setTimeout> | null = null;
+            const resizeDebounceTime = 30;
```
neodescis, line 642: Hard-coded value, at least for now

src/ui/map.ts
```diff
@@ -638,14 +638,33 @@ export class Map extends Camera {
...
                     return;
                 }
 
-                if (this._trackResize) {
+                // Debounce resize call, but with leading edge call invoked immediately
+                let resizedImmediately = false;
+                if (resizeDebounceTimerId) {
+                    clearTimeout(resizeDebounceTimerId);
+                } else if (shouldResize()) {
+                    resizedImmediately = true;
```
neodescis, line 655: If nothing was queued, then resize immediately and force the next resize to wait

src/ui/map.ts
```diff
@@ -638,14 +638,33 @@ export class Map extends Camera {
...
                     this.resize(entries)._update();
+                    resizeDebounceTimerId = setTimeout(() => {
+                        resizeDebounceTimerId = null;
+                    }, resizeDebounceTime);
+                }
+                if (!resizedImmediately && shouldResize()) {
+                    resizeDebounceTimerId = setTimeout(() => {
```
neodescis, line 662: If an immediate resize didn't happen, then queue one

HarelM:
I think there's a throttle utility somewhere in the code.

neodescis:
There is, but throttle and debounce are not the same, and debounce seemed more appropriate here as it will wait while the user is resizing. If you think throttle would be better though, I could switch it, but the throttle utility would need an update to handle arguments to the function.

What do you think about using lodash for these?

HarelM:
I would like to avoid adding lodash just for this although it is super useful in general.
Throttle seems fairly reasonable to use here I believe, I don't have strong feelings either way, it might be that we need to remove all kinds of utility functions and add the relevant losadh ones, as they are probably better maintained there, IDK, I'm on the fence here...

neodescis:
I can try using throttle instead and let you know. The time might need to be higher than for debounce too though, since it's "guaranteed" to fire that often.

neodescis:
There are a few problems with the existing throttle() implementation in the codebase, but I think they can be overcome.

1. As stated above, there is no support for the throttled function to have arguments. This is easy to add, and I've got something working for it.
2. The function's return type is overridden to return a timer ID from setTimeout(). This isn't a problem for throttling ResizeObserver, but could be a problem in the future.
3. It does not support leading-edge (immediate) execution. Even the first call is made to wait. I do think we would want that to happen here, so that a programmatic resize of the container would resize the map immediately.

The only other usage of throttle() so far is in the Hash class. I think that might benefit from a leading-edge execution as well, so I can go ahead and change throttle() to invoke immediately the first time, unless you think that's a problem.

HarelM:
As long as the code is written in the utilities file I'm fine with the solution 🙂

PR #2974, opened 2023-08-13
Sky fog test

acalcutt:
Test of splitting https://github.com/maplibre/maplibre-gl-js/pull/1713 into separate style spec changes and maplibre-gl-js changes

Testing sky/fog style spec at @acalcutt/maplibre-style-spec
https://github.com/wifidb/maplibre-style-spec/tree/sky_fog_test
https://github.com/maplibre/maplibre-style-spec/pull/298

Testing maplibre changes at 
https://github.com/wifidb/maplibre-gl-js/tree/sky_fog_test


acalcutt:
Can someone change this to a draft. I meant to open it as one.

src/style/sky.ts
```diff
@@ -0,0 +1,119 @@
...
+};
+
+type PropsPossiblyEvaluated = {
+    'sky-color': Color;
+    'fog-color': Color;
+    'fog-blend': number;
+    'horizon-blend': number;
+};
+
+const properties: Properties<Props> = new Properties({
```
HarelM, line 28: I think this part of the file should be automatically generated from the v8 definition file, shouldn't it?

HarelM:
Thanks for taking the time to push this forward!
Note that the reason the other PR was not merged is that tests were missing. So let's make sure we add the relevant tests.
The first step is obviously to finish with the spec in order to continue with this PR as we'll need a new version of the spec.

src/style/sky.ts
acalcutt: To me this looks similar to what is done in light.ts at https://github.com/maplibre/maplibre-gl-js/blob/main/src/style/light.ts#L79-L84 , so it doesn't seem that out of place to me.  Do you have an example where a file like this is generated from the v8 file?

I use light/LightSpecification as an example because it proceeds Sky in most places in the StyleSpec changes, so it is used similarly.

HarelM: You might be right, light is also not generated, but other types are, for example all the layers (circle, background, etc).
I would keep this file as "simple" as possible and not introduce GPU logic there, but preferably in another file and keep this file for definition mainly, much like the light, where there isn't a lot of logic there.

HarelM:
@acalcutt is there anything I can help with to push this forward?

acalcutt:
I was waiting to see progress on https://github.com/wifidb/maplibre-style-spec/tree/sky_fog_test before I put to much effort.

PR #2991, opened 2023-08-16
Docs: Replace broken iframe with image

SnailBones:
Iframes aren't supported in GH Markdown

# Before:

image

# After:

image

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -4,8 +4,7 @@ MapLibre GL JS is a TypeScript library that uses WebGL to render interactive map
 
 
 ## Quickstart
-
-<iframe src="./examples/simple-map.html" width="100%" style="border:none"></iframe>
+![Display a map](assets/examples/simple-map.png)	
 
 ```html
 <div id="map"></div>


HarelM:
This page is part of the docs site where this iFrame is working, I'm not sure I want to change this:
https://maplibre.org/maplibre-gl-js/docs/

SnailBones:
@HarelM Of course! I didn't realize this was used on the site.

I was confused by the name "README", which seem unusual for a file that affects the deployed docs (I picture a README as a guide for developers). Thoughts on changing it back to `index.md`?

HarelM:
I also would love to have a readme here for how to write the docs, but it will be part of the published docs site, and I'm not sure it should be there.
Index.md sounds good to me.
I don't have strong feelings when it comes to docs, feel free to change to what is most intuitive to you.

SnailBones:
> I also would love to have a readme here for how to write the docs, but it will be part of the published docs site, and I'm not sure it should be there.

Looking at the [mkdocs docs](https://www.mkdocs.org/user-guide/writing-your-docs/#index-pages), I think that this won't be the case so long as we also have an `index.md`:

> If both an `index.md` file and a `README.md` file are found in the same directory, then the `index.md` file is used and the `README.md` file is ignored.

I'll open another PR making that change if you don't mind!

HarelM:
Sure

PR #3003, opened 2023-08-20
Improve cooperative gesture example

HarelM:
This is to show how cooperative gestures behave on fullscreen mode.
Related to #2152 and #1488.


neodescis (approved):
Looks good to me!

PR #3005, opened 2023-08-20
Improvements to organization of developer docs

SnailBones:
In this PR, I rename the documentation landing page `docs/README.md` to `docs/index.md`, a more accurate name that avoids the confusion described in #2991. Then, I move the docs writing guide `documents_and_diagrams/README.md` to `docs/README.md`. This should make it easier for new contributors to discover it.

The README won't be visible on the site thanks to this behavior from the [mkdocs docs](https://www.mkdocs.org/user-guide/writing-your-docs/#index-pages):

> If both an `index.md` file and a `README.md` file are found in the same directory, then the `index.md` file is used and the `README.md` file is ignored.

Thus, the docs site is unchanged.

I also rename the folder `documents_and_diagrams` to `developer guides` to be more specific.

to do:

- [x] Update links


docs/README.md
```diff
@@ -1,84 +1,85 @@
...
+### MapLibre GL JS Documentation
 
-MapLibre GL JS is a TypeScript library that uses WebGL to render interactive maps from vector tiles in a browser. The customization of the map comply with the [MapLibre Style Spec](https://maplibre.org/maplibre-style-spec). It is part of the [MapLibre ecosystem](https://github.com/maplibre), with a pendant for Mobile, Desktop, Servers called [MapLibre Native](https://maplibre.org/projects/maplibre-native/).
+This directory contains the source for the [MapLibre GL JS docs](https://maplibre.org/maplibre-gl-js/docs/) hosted on the MapLibre site.
 
+Besides this README, each other `.md` file in this directory corresponds to a site page. Each is converted into an `.html` file by [MkDocs](https://www.mkdocs.org/).
+## Setting up Docker access
 
-## Quickstart
+If you're on Linux, before you can run the docs you need to give your user permission to run `docker` commands without `sudo`, as explained [here in the Docker docs](https://docs.docker.com/engine/install/linux-postinstall/).
```
HarelM, line 8: Maybe move this part to the linux getting started readme? There's a whole section for these kind of instructions...?

docs/index.md
```diff
@@ -0,0 +1,84 @@
...
+
+As a mitigation for Cross-Site Scripting and other types of web security vulnerabilities, you may use a [Content Security Policy (CSP)](https://developer.mozilla.org/en-US/docs/Web/Security/CSP) to specify security policies for your website. If you do, MapLibre GL JS requires the following CSP directives:
+
+```
+worker-src blob: ;
+child-src blob: ;
+img-src data: blob: ;
+```
+
+Requesting styles from Mapbox or other services will require additional directives. For Mapbox, you can use this `connect-src` directive:
```
HarelM, line 50: This whole section about mapbox should be here, interesting that I haven't noticed this before.
Basically, a style/tile provider should have this in their docs, not here...

HarelM:
I've left s few comments, otherwise this looks great, thanks!

HarelM:
It might be good to move the contributing.md to this new folder...?

SnailBones:
> It might be good to move the contributing.md to this new folder...?

Then should we also move `ARCHICETURE.md'? It's not obvious to me which should go where, but I think it's especially important for`CONTRIBUTING.md` to be easy to find.

HarelM (approved):
Thanks!! Feel free to open another PR if you have more thoughts on where things should be.

Issue #2990, opened 2023-08-16
Improve accuracy of CJK glyphs with higher resolution TinySDF textures


bdon:
## User Story

As a viewer of the map in Chinese or Japanese language, the quality of rendered text should look as good as Latin text and other scripts. 

## Rationale

* All text displayed in MapLibre GL, web or native, is rendered using [Signed Distance Fields](https://en.wikipedia.org/wiki/Signed_distance_function#Applications) (SDFs). This means text can be smoothly rotated and rescaled between different sizes as a function of zoom level, and text halos - a common visual design for map labels - are computationally "free". 

<img src="https://upload.wikimedia.org/wikipedia/commons/thumb/7/74/Signed_distance_field_duck.svg/1920px-Signed_distance_field_duck.svg.png" width="200" height="200">

* The drawback to the SDF approach is that it isn't the format fonts are commonly stored in. Fonts are defined as a collection of Bézier curves in a .TTF or .OTF file. In order to display them as SDFs they need to be "baked" into a texture via preprocessing with a program like [font-maker](https://github.com/maplibre/font-maker). 

* The conversion from .TTF to SDFs requires the choice of a specific font size for rasterization into a bitmap via FreeType. Early on in GL JS development this was determined to be a [**24 point font**](https://github.com/maplibre/font-maker/blob/main/main.cpp#L21) as a compromise between text sharpness and SDF bitmap size. SDF bitmaps are then stored in ranges of 256 glyphs and served over HTTP. A "font stack" is 256 files of 256 glyphs each, covering the entire 65,536 codepoint [Basic Multilingual Plane](https://en.wikipedia.org/wiki/Plane_(Unicode)#Basic_Multilingual_Plane)

* SDF 24 point font is good enough to render most Latin fonts with a small amount of rounding. The issue arises when rendering glyphs with more internal detail. These are common where CJK is used with "traditional" scripts - mainly Japan, Taiwan, Hong Kong and Macau. From the MapLibre [example page](https://maplibre.org/maplibre-gl-js/docs/examples/fullscreen/) I screenshotted a few[^1]:

![font size](https://github.com/maplibre/maplibre-gl-js/assets/77501/933bb71e-b05c-49cb-be57-629ea952d88f)

[...]

Issue #1051, opened 2022-03-05
Character alignment issue when CJK glyphs are mixed with other characters.


peaceshi:
**maplibre-gl-js version**: 2.1.7

**browser**: Microsoft Edge 99.0.1150.30（latest stable）

**I use `localIdeographFontFamily` to load the font.**

```
{
            type: "FeatureCollection",
            features: [
              {
                type: "Feature",
                properties: {
                  description: "璃月港A8璃月港\nGgLlIi\n寄了JjPpQqYy寄了"
                },
                geometry: {
                  type: "Point",
                  coordinates: [0, 0]
                }
              }
            ]
          }
```

![image](https://user-images.githubusercontent.com/10380828/156879765-92b3d495-d24c-4dbc-9c8b-7883f14bc5c8.png)

Seems like mapbox fixed it in a way. But I can find it.
https://www.mapbox.com/blog/efficient-pitched-tile-loading-precise-cjk-labels-and-js-promises-mapbox-gl-js-v2-1-1
![image](https://user-images.githubusercontent.com/10380828/156879832-36e1e254-e59d-455e-8dc0-80ac8c9c2c10.png)

Some useful link:
https://github.com/mapbox/mapbox-gl-js/pull/10390
https://github.com/mapbox/mapbox-gl-js/issues/8041
https://github.com/mapbox/mapbox-gl-js/issues/8041#issuecomment-473858431

HarelM:
We can't copy code from mapbox...
I'm not sure how to properly solve this, a PR would help a lot in this case.

peaceshi:
https://github.com/mapbox/mapbox-gl-js/pull/10298
I will try PR later.

HarelM:
@peaceshi any updates on this?

spatialillusions:
Ping on this. 
I have built an offline web app that you can run from an USB stick, and if would be awesome if I could use local fonts instead of including all font ranges.
Maybe it would be as simple as returning true from _doesCharSupportLocalGlyph if a localFont property is set.

HarelM:
I'm not sure I understand how local fonts are related to this issue... Can you please explain?

spatialillusions:
Sorry, I might have been unclear. 

[...]

PR #3006, opened 2023-08-21
Improve CJK text rendering [#2990]

bdon:
Addresses proposal in https://github.com/maplibre/maplibre-gl-js/issues/2990

* Double TinySDF-rendered glyphs with new texScale glyph property.

* Change topAdjustment/leftAdjustment to make 2x glyphs match 1x positions.

Video clip of before/after:

https://github.com/maplibre/maplibre-gl-js/assets/77501/b4009321-fba1-4f66-a997-793851e608c0

This required adding `0.5` to both topAdjustment and leftAdjustment, otherwise the alignment was slightly off from the old glyphs, potentially being a "breaking" visual change (an extremely small one).

## Launch Checklist

 - [x] Link to related issues.
 - [x] ~~Document any changes to public APIs~~ - no changes to public APIs.
 - [ ] Post benchmark scores.


src/style/style_glyph.ts
```diff
@@ -2,13 +2,15 @@ import type {AlphaImage} from '../util/image';
...
  * The glyph's metrices
+ * texScale = 1 for a texture at 24pt, 2 = 48pt, etc
  */
 export type GlyphMetrics = {
     width: number;
     height: number;
     left: number;
     top: number;
     advance: number;
+    texScale: number;
```
HarelM, line 13: Does this mean "textureScale"? if so, please rename.

bdon: OK, will do.

src/render/glyph_manager.test.ts
```diff
@@ -110,31 +110,32 @@ describe('GlyphManager', () => {
     test('GlyphManager generates CJK PBF locally', done => {
         const manager = createGlyphManager('sans-serif');
 
+        // character 平
         manager.getGlyphs({'Arial Unicode MS': [0x5e73]}, (err, glyphs) => {
             expect(err).toBeFalsy();
-            expect(glyphs['Arial Unicode MS'][0x5e73].metrics.advance).toBe(1);
```
bdon, line 115: need to double-check where this `1` value comes from (a real canvas text implementation in the test runner?)

HarelM:
Is it possible to fix the alignment as part of this PR as well (you mentioned changing it)? see #1051.

bdon:
> Is it possible to fix the alignment as part of this PR as well (you mentioned changing it)? see #1051.

I don't believe so. I would intend this PR to make it into a minor version release or even a patch (no API enhancements). 

If we change those hardcoded alignment numbers, we will almost certainly change the visual appearance of multilingual labels, which would be a "breaking" change for v4 possibly. We're in a bit of a bind from what I understand, because those numbers are hardcoded to be appropriate for the fonts Mapbox provided as part of their API - DIN Pro for latin and Arial Unicode for fallbacks. 

Because MapLibre users are using whatever fonts they like, we don't have data to support an improvement there. The best we could we do is make recommendations on a fallback font - see https://github.com/maplibre/font-maker/issues/16 - and then make multilingual labels work well for that recommendation.

In any case, it's beyond the scope of this PR.

HarelM:
Ok, thanks for the info!

bdon:
@HarelM what's the best way example for the `Post benchmark scores.` task?

HarelM:
Try reading this:
https://github.com/maplibre/maplibre-gl-js/blob/95d6f43ed396555ca710445cb9cb25a378c84cc4/test/bench/README.md
I hope it's up to date...
You should be able to run this from command line and also through the browser.

bdon:
Testing https://bdon.github.io/maplibre-cjk/ visually on:

- [ ] macOS WebKit
- [ ] macOS Chromium
- [ ] macOS Gecko
- [ ] iOS WebKit
- [ ] Android Chromium
- [ ] Windows Edge
- [ ] Windows Chromium
- [ ] Windows Gecko

HarelM:
I think there are a few failing render tests due to this change.

HarelM:
This looks like a fairly small change, so I think it can go in, I don't see and issues in the code.
I don't get a notification when changing a PR from draft to ready for review, so please request a review once you change the status of this PR.

bdon:
I'm currently experimenting with a refactoring the code a bit to move `textureScale` out of `metrics`, because that breaks all the test fixture JSONS in the integration tests.

wipfli:
Made a demo: https://github.com/wipfli/double-resolution-cjk-maplibre

wipfli:
I added to my demo what would happen if we quadruple the resolution. Quite fun, it would allow MapLibre GL JS to render some quite large text labels without too much round corners

wipfli:
I don't want to bring this pull request off track, but it is very appealing what you can do with local glyph generation. Will think about the options

src/style/style_glyph.ts
```diff
@@ -2,13 +2,15 @@ import type {AlphaImage} from '../util/image';
 
 /**
  * The glyph's metrices
+ * isDoubleResolution = true for 48px textures
```
HarelM, line 5: Move this above the relevant field.
If you know what the other fields are for, feel free to add comment above them as well.
i.e.
```ts
/**
 * This is my field!!
 */
myField: number
```

HarelM (approved):
I've left a small comment, approved otherwise.

HarelM:
THANKS!!

src/render/glyph_manager.test.ts
HarelM: @bdon is this still relevant?

bdon: I figured out it comes from https://github.com/hustcc/jest-canvas-mock - nothing to see here.

HarelM:
Can you resolve conflicts for the changelog file please?

HarelM:
THANKS!
I would expect the CJK browser test to fail, shouldn't it?
https://github.com/maplibre/maplibre-gl-js/blob/0e9a4843c91b137c3681aa8379eb2e3c8a85fe6f/test/integration/browser/browser.test.ts#L168C7-L168C7
It basically checks image equality of what is rendered in the browser, but I don't see it failing here and I don't see a change in the test as part of this PR.
What am I missing?

bdon:
not sure about this test case - at least locally I can delete the entire code block at https://github.com/maplibre/maplibre-gl-js/blob/0e9a4843c91b137c3681aa8379eb2e3c8a85fe6f/test/integration/browser/browser.test.ts#L170 and the test still passes. `compareByPixelmatch` always returns 0 at locally (for all platforms). Can you verify you see the same?

bdon:
Is this `parse` used in other test suites? 

https://github.com/maplibre/maplibre-gl-js/blob/0e9a4843c91b137c3681aa8379eb2e3c8a85fe6f/test/integration/browser/browser.test.ts#L240

maybe we need to use these? https://github.com/pngjs/pngjs#sync-api

wipfli:
Maybe the pixel diff threshold of the test is too low to detect the improvement of this pull request ...

HarelM:
Nope, the CJK map is empty, so nothing is drawn.
I'm not sure how much time this test is broken. :-(

HarelM:
Wait, not sure it is empty, I'll check if I can add a background...

HarelM:
I'll open a PR shortly with the relevant fixes. Seem like this test is broken for a long time... :-(

HarelM:
I've opened a PR: #3162.
Looking at what I did there and the fact that we are running render test in the browser now, I would expect to see some failures in render tests that are using CJK characters.
If this is not the case, it might be because the threshold is OK as this is a subtle change in the font.
Can you run the render test on this branch with UPDATE env parameter to see how these changes look in the render test actual vs expected pngs?
If the change is subtle, as can be seen in the PR comment I can rest assure that we are testing what we expect.
Sorry for nagging, I just want to make sure our tests are actually testing what they should.

HarelM:
I've decided to remove the CJK test there and count on the render tests.
So the only question left for this PR is if the render tests related to CJK are behaving as expected - please verify this and I'll merge this PR.
Thanks for all the hard work!

bdon:
@HarelM all the render tests may be passing because the test harness disables local ideograph rendering (the exact code path this PR affects) by default: 

https://github.com/maplibre/maplibre-gl-js/blob/main/test/integration/render/run_render_tests.ts#L538

I don't think there are any instances where this code is tested by render tests. If we enable the option in a render test, the exact pixel output will depend on platform.

HarelM:
Then I think we should enable it so that the render tests will test these code paths.
In the past, the tests ran on node environment, and it might have made sense to disable it.
We have the infrastructure to have multiple images for every render test so that it passes for every platform, we should utilize this here as well I believe.
The bottom line, is that I want to make sure we cover these important use cases in the render tests as well and that CJK is not an after thought but rather a core part of this library.

Let me know if my help is needed to achieve this.

bdon:
Created a new test using the old code at https://github.com/maplibre/maplibre-gl-js/pull/3166 we can merge first before this PR

HarelM:
Sure, why not.

wipfli:
Thanks bringing higher resolution CJK to MapLibre GL JS, @bdon. Can you open an issue in MapLibre Native that we remember to do the same there?

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ### ✨ Features and improvements
 
+- Improve error message when a tile can't be loaded ([#3130](https://github.com/maplibre/maplibre-gl-js/pull/3130))
```
HarelM, line 5: This is duplicated and should be removed from here...

HarelM:
Can we push this forward?
I would like to have this in, in the next version.
Let me know what else is needed here in order to merge this.

HarelM:
@bdon any chance to fix that last few comments?

CHANGELOG.md
bdon: fixed

bdon:
The default allowed diff of 0.00025 is not sensitive enough to catch the difference here: 

visually it looks like the correct change in one of the 3 platforms (not sure what's going on in the other 2) with a diff of `0.00006614583333333334`

(gif attached)


https://github.com/maplibre/maplibre-gl-js/assets/77501/cda8e528-036b-4fa8-9477-fe2684af2a3a

bdon:
I changed the test to use much larger text sizes over the default, which should make it more sensitive to pixel changes

HarelM (approved):

wipfli:
> Can you open an issue in MapLibre Native that we remember to do the same there?

@bdon did you open a ticket in MapLibre Native?

bdon:
@wipfli I haven't. I can open one when I have a plan to work on it, but I don't right now.

wipfli:
Ah easy. @louwers can you open a ticket for double resolution cjk in native? It is just that we keep track of parity tasks. @bdon you don't have to work on this in native, it is just for us to track...

PR #3021, opened 2023-08-23
Copy LICENSE.txt to dist folder so it's included in 3rdpartylicenses by webpack

### LINKED ISSUE #2865: MapLibre license is not getting picked up by webpack
**maplibre-gl-js version**: 3.2

**browser**: Any

### Steps to Trigger Behavior

 1. Use maplibre with webpack (for example with angular)
 2. Check out the 3rdpartylicenses.txt it generates
 3. --> MapLibre license is not there

### Expected Behavior
License text should be there

### Actual Behavior
License text isn't there.

### Additional Info
This was mentioned here: #2737.
The cause of this is due to the fact that the repo' main package.json is with `type: module` but the package.json in the dist folder is generated with `type: commonjs`.
It also has a weird name I think `@maplibre/distfiles`.

One possible solution is to simply copy the license file into the dist folder as part of the build process.
I'm not sure there's actually another way...

cc: @rotu @birkskyum What do you guys think?

We should also change the license's year...

miccou:
Fixes #2865 

This PR consists of three changes:
- [when creating dist/package.json, also copy LICENSE.txt from repo root](https://github.com/maplibre/maplibre-gl-js/commit/6b827479cad87a87cd04d6212fcfdd2a58c402c0)
  - This is achieved by modifying `generate-dist-package.js`. I assume this file must be run during build, but couldn't see where in the GitHub actions. This seemed like the best place without adding a rollup plugin or relying on an OS specific command.
  - I tested with local npm installation and upon `ng build` the license was included in 3rdpartylicenses.txt.
- [set name in dist/package.json to maplibre-gl](https://github.com/maplibre/maplibre-gl-js/commit/9213b1f0678f1d26d4e4409484640e515496b6ec)
  - This will ensure that the package name as listed in 3rdpartylicenses.txt is correct, though I am not sure if this carries any risk - local testing was fine.
- [bump license year to 2023](https://github.com/maplibre/maplibre-gl-js/commit/a921a5a7faf113765751cd08dbbc5c8bb0c24162)
  - Since the website mentions Copyright © 2023 Maplibre, I took the liberty of bumping the year here.

Please let me know if there's any feedback or a way I can test more thoroughly than via `npm install ~/repos/maplibre-gl-js`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### ✨ Features and improvements
 
+- Copy LICENSE.txt to dist folder so it's included in 3rdpartylicenses.txt by webpack ([#3021](https://github.com/maplibre/maplibre-gl-js/pull/3021))
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,4 +1,4 @@
-Copyright (c) 2020, MapLibre contributors
+Copyright (c) 2023, MapLibre contributors
 
 All rights reserved.
 
--- a/build/generate-dist-package.js
+++ b/build/generate-dist-package.js
@@ -3,7 +3,7 @@
 /// This package.json ensures that node imports from outside see them as such
 // https://nodejs.org/api/packages.html#type
 
-import { writeFile, mkdir } from "node:fs/promises"
+import { writeFile, mkdir, copyFile } from "node:fs/promises"
 
 async function ensureDist() {
   const dist = new URL("../dist/", import.meta.url);
@@ -18,4 +18,9 @@ await writeFile(
     type: "commonjs",
     deprecated: "Please install maplibre-gl from parent directory instead",
   })
+)
+
+await copyFile(
+  "./LICENSE.txt",
+  new URL("LICENSE.txt", await ensureDist())
 )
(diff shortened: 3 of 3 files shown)


HarelM (approved):

HarelM:
LGTM, THANKS!

HarelM:
Can you add a changelog entry with the info about this fix please?

miccou:
Thanks for your review @HarelM, I just pushed an update to the changelog to include this: a3add2c

HarelM:
@miccou I see the following warning in the jest unit test run:
```
Running one project: unit
jest-haste-map: Haste module naming collision: maplibre-gl
  The following files share their name; please adjust your hasteImpl:
    * <rootDir>/package.json
    * <rootDir>/dist/package.json
```
Can you see if you can take care of this somehow?

miccou:
@HarelM I'll take a look. The name from dust/package.json is used for the 3rdpartylicenses.txt file, maybe we can change that to something that doesn't conflict? What's your preference?

miccou:
@HarelM 

My understanding is that the tests don't need to load via `require()` anything in dist, and that the build tests just use `fs` to check the contents of the dist folder.

Given this, it's probably best to not allow the contents of the dist folder to be discoverable by jest. This can be achieved through configuration.

I think this is the best solution. Renaming the module in package.json will mean that npm doesn't match 3rdpartylicenses.txt. 

Here's what that configuration change would look like, let me know if you want a PR: https://github.com/maplibre/maplibre-gl-js/commit/9c2dd150b7e9cfde7974568f5ff0e995446c853e

HarelM:
I'm fine with either way we solve this.
Feel free to submit a PR with the linked change.

PR #3050, opened 2023-08-30
Replace st with http-server

birkskyum:
This replace our dev static file server ([st](https://github.com/isaacs/st)) with a more mainstream alternative ([http-server](https://github.com/http-party/http-server)).

Usage stats https://npmtrends.com/http-server-vs-st


birkskyum:
[http-server](https://github.com/http-party/http-server) actually looks unmaintained as well - what does people use for a modern simple static file server these days?

jingsam:
live-server

PR #3051, opened 2023-08-31
Reset scroll handler state properly

### LINKED ISSUE #2709: Interrupting a scroll zoom causes the next scroll zoom to return to the prior zoom level
**maplibre-gl-js version**: 3.1.0

**browser**: Chrome

### Steps to Trigger Behavior
 1. Initiate a mouse wheel zoom
 2. Interrupt the zoom by calling map.stop(), map.flyTo(), map.jumpTo(), map.setZoom(), etc
 3. Change the zoom with another method, such as map.flyTo()
 4. Initiate another mouse wheel zoom

### Link to Demonstration

https://jsbin.com/gituvujaga/edit?html,js,output

### Expected Behavior

1. Scroll once. The camera zooms very far in.
2. Scroll again. The camera zooms in/out relative to where it currently is.

### Actual Behavior

1. Scroll once. The camera zooms very far in.
2. Scroll again. The camera snaps back to where it was when the zoom got interrupted.

As a workaround, deleting the `map.scrollZoom._targetZoom` property after interrupting the zoom yields the expected behavior.

HarelM:
Fixes #2709

This basically resets the scroll handler state better which solves the above bug.
I couldn't find a way to test this using the existing unit test, as they are more of integration tests.
The browser test was a last resort, but it reproduces the bug perfectly.

### DIFF
--- a/src/ui/handler/scroll_zoom.ts
+++ b/src/ui/handler/scroll_zoom.ts
@@ -256,7 +256,6 @@ export class ScrollZoomHandler implements Handler {
 
         // if we've had scroll events since the last render frame, consume the
         // accumulated delta, and update the target zoom level accordingly
-        console.log(`fr, ${this._delta} ${this._targetZoom}`);
         if (this._delta !== 0) {
             // For trackpad events and single mouse wheel ticks, use the default zoom rate
             const zoomRate = (this._type === 'wheel' && Math.abs(this._delta) > wheelZoomDelta) ? this._wheelZoomRate : this._defaultZoomRate;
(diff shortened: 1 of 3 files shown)


src/ui/handler/scroll_zoom.ts
```diff
@@ -351,5 +351,11 @@ export class ScrollZoomHandler implements Handler {
 
     reset() {
         this._active = false;
+        this._zooming = false;
+        delete this._targetZoom;
```
neodescis, line 355: It seems strange to me that these are deleted instead of set to undefined, but I guess it's not a big deal.

neodescis (approved):

src/ui/handler/scroll_zoom.ts
HarelM: This is how it's used in other parts of this class.
I personally don't see a big difference.

neodescis: Nope, not a big difference. And I did look at the rest of the class. Just thought it a bit strange is all. But certainly not a problem.

Issue #2238, opened 2023-03-01
Momentum scrolling seems instant on MacOS with reduced motion


joepio:
**maplibre-gl-js version**: 2.4.0 (also [posted this on MapBoxGL](https://github.com/mapbox/mapbox-gl-js/issues/12602))

**browser**: Chrome, safari (on MacOS)

### Steps to Trigger Behavior

1. Turn on reduced motion on macOS accessibility settings
 1. Open a map on MacOS (e.g. https://maplibre.org/)
 2. Drag over the map with the mouse, let go while making the dragging motion (don't wait with letting go until movement has stopped)
 3. You will see the map skips over an area, dependent on your speed (see video below)

### Link to Demonstration

Note how the map instantly teleports to a new location when the cursor changes from 'dragging' to 'normal'.

https://user-images.githubusercontent.com/2183313/222200309-4769a50b-d649-4a4b-bdf7-bd73c2c4d901.mov

### Expected Behavior

There are two ways dragging and letting go while having momentum should work:

1. Freeze the position when the mouse button is let go. 
2. Momentum scrolling, even when reduced motion is on.

### Actual Behavior

Map 'teleports' to position further then where the mouse is let go. I suspect that it instantly teleports to the calculated 'momentum' position (where it should come to a stop), but the animation is 0ms long. But this is just a guess coming from someone who doesn't know the internals of course.

HarelM:
I don't think maplibre has versions 2.12, 2.13...
Can you clarify the following please:
"Turn on reduced motion on macOS accessibility settings"?
Can this be tested using the following media query?
https://a11y-101.com/development/reduced-motion#:~:text=On%20a%20Mac%20you%20go,turn%20on%22Reduce%20Motion%22.

joepio:
@HarelM 

> I don't think maplibre has versions 2.12, 2.13...

You're right, it's the latest version (if that's the one on maplibre.org).

> Can you clarify the following please: "Turn on reduced motion on macOS accessibility settings"?

On a mac, go to accessibility settings, then to `display` then tick `reduce motion`

> Can this be tested using the following media query?

And not sure what you mean with can this be tested?

Anyways, I'm pretty sure this is the culprit: https://github.com/maplibre/maplibre-gl-js/blob/04ea214c92518c8245fcc2552bd0da8b3cde9d3c/src/ui/camera.ts#L1098

HarelM:
[...]

PR #3068, opened 2023-09-10
fix: [issue-2238] stop camera animation on mac when reduced motion is on

sebastianoscarlopez:
To fix issue #2238, this PR modifies the camera animation behavior on Mac when the 'reduced motion' setting is on.

### Explanation:
 The **camera.easeTo** method is called after the map movement happens, in order to add if needed a kind of "inertia movement". The *easeTo* method is called with the **around** property option with the position that should be the end of the camera ease movement.
Currently, when the 'reduced motion' setting is on, the duration is changed by 0. This does not avoid the final result, it only avoids the animation, because the final result will be at the **around** position, so it looks like an unexpected jump.

Example of the current behaviour, another example can be found here #2238

https://github.com/maplibre/maplibre-gl-js/assets/10349981/df381d05-a92b-4561-a677-9cde719a1549

### Approaches to fix

- My first approach was to just `return this;` if the reduced motion was on, but that approach didn't work well because it broke the zoom when the method was called by a double click.

- Another approach can be to reassign the **around** property to the center, because as was said before the movement was already done, and the center is in the current position, so the following code in the method will try as usual to do a movement, but its start position will be equal to the end position, so no movement will occur at all.

- This PR approach avoids calling this method when reduced motion is on

Result with this approach

https://github.com/maplibre/maplibre-gl-js/assets/10349981/c8b8b3f0-c2d7-4789-8044-95e1a2fa49ce

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Fix unit test warning about duplicate module names ([#3049](https://github.com/maplibre/maplibre-gl-js/pull/3049))
 - Correct marker position when switching between 2D and 3D view ([#2996](https://github.com/maplibre/maplibre-gl-js/pull/2996))
 - Fix error thrown when unsetting line-gradient [#2683]
+- Avoiding inertia animation on Mac when reduced motion is on ([#3068](https://github.com/maplibre/maplibre-gl-js/pull/3068))
 - _...Add new stuff here..._
 
 ## 3.3.1
--- a/src/ui/handler_manager.ts
+++ b/src/ui/handler_manager.ts
@@ -17,6 +17,7 @@ import {DragPanHandler} from './handler/shim/drag_pan';
 import {DragRotateHandler} from './handler/shim/drag_rotate';
 import {TwoFingersTouchZoomRotateHandler} from './handler/shim/two_fingers_touch';
 import {extend} from '../util/util';
+import {browser} from '../util/browser';
 import Point from '@mapbox/point-geometry';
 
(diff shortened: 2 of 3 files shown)


HarelM:
Can you elaborate why setting arond to center solves this issue?
Also can you add a before and after videos to better show the change/fix?
What will happen if the user uses easeTo not around the center with reduced motion? Is it a possibility?

sebastianoscarlopez:
Render tests (windows-latest, 2) failed, I thought it was related to my code, but now I'm not sure.

sebastianoscarlopez:
Hi @HarelM, thanks for the feedback. The description was updated. While I was writing it, another approach came to my mind.
Please let me know your thoughts about it.

HarelM:
I think I would like to avoid calling easeTo in case of reduced motion, seems cleaner to me...

sebastianoscarlopez:
> I think I would like to avoid calling easeTo in case of reduced motion, seems cleaner to me...

@HarelM as you said, avoiding easeTo (inertia animation) results in cleaner code.

note: some tests failed, but it seems like something unrelated to this pr.

HarelM:
Unfortunately, the tests are flaky, not sure how to solve this... :-(
Can you add a unit test to make sure this behavior will not break in the future?
Otherwise, this looks great.

sebastianoscarlopez:
Now the test drag map check behavior with and without reduced motion feature

HarelM:
Looks great and super clean, THANKS!

PR #3077, opened 2023-09-13
Add lock file version to `.npmrc`

coliff:
maplibre-gj-js uses lock file version 3 (https://github.com/maplibre/maplibre-gl-js/blob/main/package-lock.json) so we should specify here in the `.npmrc` to help prevent changes. I added spaces between the = to make it easier to read.

Info: https://docs.npmjs.com/cli/v10/using-npm/config#lockfile-version 

As this is so minor, I haven't changed the CHANGELOG.md.

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,1 +1,2 @@
-engine-strict=true
+engine-strict = true
+lockfileVersion = 3


HarelM:
Not that I mind adding this, but it's specified in the lock file, why repeat this information?
What I'm trying to ask is if we decide to change the version wouldn't we need to change it in two places instead of one?

coliff:
If someone with an older version of npm installs then that may change the version to lockfileversion in package-lock.json to v2 (or even v1!). so specifying it in the `.npmrc` is specifying the correct version.

HarelM (approved):

HarelM:
Thanks for taking the time to contribute this!!

PR #3087, opened 2023-09-16
Use raster dem encoding

ibesora:
This is uses [maplibre-style-spec#337](https://github.com/maplibre/maplibre-style-spec/pull/337) to allow rendering raster-dem sources with an arbitrary encoding.

As an example, the first image is a QGIS reference render of a raster-dem with a different encoding, the second one is how  it is rendered currently in Maplibre and the third one is how it's rendered using this PR.

Screenshot 2023-09-16 at 18 22 16
Screenshot 2023-09-16 at 17 17 45
Screenshot 2023-09-16 at 17 17 52

Tests are obviously failing because this is pointing to a maplibre-style-spec version that doesn't contain the needed changes

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


src/source/worker_source.ts
```diff
@@ -37,7 +37,11 @@ export type WorkerDEMTileParameters = TileParameters & {
         w: number;
     };
     rawImageData: RGBAImage | ImageBitmap;
-    encoding: 'mapbox' | 'terrarium';
+    encoding: 'mapbox' | 'terrarium' | 'custom';
```
HarelM, line 40: It's better to define this enum somewhere instead of repeating this definition.

HarelM:
This looks great!! THANKS!

HarelM:
Otherwise this looks good.

src/data/dem_data.ts
```diff
@@ -70,29 +99,22 @@ export class DEMData {
...
-        return ((r * 256 * 256 + g * 256.0 + b) / 10.0 - 10000.0);
-    }
-
-    _unpackTerrarium(r: number, g: number, b: number) {
-        // unpacking formula for mapzen terrarium:
-        // https://aws.amazon.com/public-datasets/terrain/
-        return ((r * 256 + g + b / 256) - 32768.0);
+    unpack(r: number, g: number, b: number) {
+        const value = (r * this.redFactor + g * this.greenFactor + b * this.blueFactor - this.baseShift);
+        const significantDigits = 5;
```
HarelM, line 116: Can you clarify this?

ibesora:
@HarelM I need your take on the last fixes and the error that is still showing [here](https://github.com/maplibre/maplibre-gl-js/actions/runs/6409566839) 

1. I was getting a bunch of errors on all the hillshade tests. I have been looking at what was happening and it was a rounding error. Using the previous behavior we'd get a value like `1243` while with the new approach we get `1243.000000006` and that was causing the differences. I fixed this by adding a rounding step to 5 decimal places [here](https://github.com/maplibre/maplibre-gl-js/pull/3087/commits/884039ce6e19452b86ece9e038e3e6b3b0e56a15) but that's totally made up. Alternatively we could just update the test images.
2. The style in tests/terrain/field-of-view was wrong. It contained a `"format"` key that's not valid according to the RasterDEM spec. This was not a problem before because RasterDEM was validated as an Object in maplibre-style-spec and that allowed for random keys. The update to the last version of the style spec changed the validation from an Object to a proper validation so it fails when there are invalid properties. This was fixed [here](https://github.com/maplibre/maplibre-gl-js/pull/3087/commits/a95f6bd1162270daf339c445955cc021759b606c)
3. And lastly, I had to increase the min size to make the build test pass
4. On the error that's still showing, do you think we can just update the test result image? It seems like it's only a precision problem again

HarelM:
If the change in the images is not significant, I would consider simply updating the images.
It's weird that using variables instead of hardcoded numbers creates these rounding errors, I'm not sure rounding is the right approach though, but if we can't figure out the root cause of these small changes I guess rounding is OK.

ibesora:
> It's weird that using variables instead of hardcoded numbers creates these rounding errors, I'm not sure rounding is the right approach though, but if we can't figure out the root cause of these small changes I guess rounding is OK.

It's not that, we are essentially doing more floating point operations than before. The previous Mapbox encoding formula was `(r*256*256 + g*256 + b) * 0.1 - 10000.0` while the new one is `r*6553.6 + g*25.6  b*0.1 - 10000.0`. The first one results in `832150.4`while the second one results in `832150.4000000001`.

Anyway, it seems that the errors I was seeing were not related to that so I just removed the rounding and the tests are still passing so I think this is ready to merge now!

HarelM (approved):

HarelM:
Damn, changelog entry is still missing, can you please open a new PR with the addition of this feature in the changelog?

PR #3162, opened 2023-10-03
Remove CJK browser test

HarelM:
This removes the CJK test - the png parse was never observed, the diff was too strict and using base64 text isn't a great way to see what is the expected images.
In the past the render test were ran in node, but now they are running in the browser, so this test is not needed anymore.


test/integration/browser/browser.test.ts
```diff
@@ -165,97 +173,9 @@ describe('Browser tests', () => {
...
-
-        // At least one platform should be identical
-        expect(minDiff).toBe(0);
-
-    }, 20000);
-
     test('Marker: correct position', async () => {
         const markerScreenPosition = await page.evaluate(() => {
-            const markerMapPosition = [11.40, 47.30];
+            const markerMapPosition = [11.40, 47.30] as [number, number];
```
neodescis, line 178: Maybe this doesn't matter, but I would usually prefer to write this as follows to avoid a "cast":

`const markerMapPosition: [number, number] = [11.40, 47.30];`

neodescis (approved):
Looks good to me. One minor comment, but feel free to ignore :-)

PR #3178, opened 2023-10-05
Bumps version to 3.4

ibesora:
This also updates the changelog to add info about the custom raster-dem encodings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,18 +2,10 @@
 
 ### ✨ Features and improvements
 
-- Improve error message when a tile can't be loaded ([#3130](https://github.com/maplibre/maplibre-gl-js/pull/3130))
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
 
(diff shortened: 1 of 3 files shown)


neodescis (approved):
Looks good to me. @HarelM , are we ready to bump version and do a release?

CHANGELOG.md
```diff
@@ -16,6 +16,23 @@
 - 3d buildings example doesn't work as expected ([#3165](https://github.com/maplibre/maplibre-gl-js/pull/3165))
 - _...Add new stuff here..._
 
+## 3.3.2
+
+### ✨ Features and improvements
```
HarelM, line 21: This is duplicated... no need to duplicate, just add empty sections under "main" instead.

HarelM:
Sure, if someone wants this version out I see no reason to keep it.
We can always release another one soon after.

CHANGELOG.md
ibesora: I thought I had removed them 🤦

ibesora:
Wondering if this should be a new minor version instead of a patch one. 
There are things in the changelog that seem to be new features under patch versions so I'm not sure (i.e. adding `cache` to `RequestParameters`in 3.2.2) but happy to follow more strict semver rules if needed

HarelM:
Yes, let's do 3.4.

ibesora:
> Yes, let's do 3.4.

Done!

HarelM (approved):

PR #3184, opened 2023-10-06
Improve geojson source typings

HarelM:
This basically improves the types of geojson source.
I was trying to find a bug there and found out the types are not great in certain places and there was a function that used "this" without being part of a class, which really confused me.

So I fixed what I found, reduced some duplicate types when I saw them.


src/source/geojson_worker_source.ts
```diff
@@ -32,57 +32,22 @@ export type LoadGeoJSONParameters = {
...
 
-export interface GeoJSONIndex {
-    getTile(z: number, x: number, y: number): any;
-    // supercluster methods
-    getClusterExpansionZoom(clusterId: number): number;
-    getChildren(clusterId: number): Array<GeoJSON.Feature>;
-    getLeaves(clusterId: number, limit: number, offset: number): Array<GeoJSON.Feature>;
-}
-
-function loadGeoJSONTile(params: WorkerTileParameters, callback: LoadVectorDataCallback): (() => void) | void {
```
HarelM, line 52: this method wasn't removed, it just moved to be a part of the class so that reference to "this" will make more sense, I hope...

src/source/geojson_worker_source.ts
```diff
@@ -101,12 +66,41 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
...
     constructor(actor: Actor, layerIndex: StyleLayerIndex, availableImages: Array<string>, loadGeoJSON?: LoadGeoJSON | null) {
-        super(actor, layerIndex, availableImages, loadGeoJSONTile);
+        super(actor, layerIndex, availableImages);
+        this.loadVectorData = this.loadGeoJSONTile;
         if (loadGeoJSON) {
             this.loadGeoJSON = loadGeoJSON;
         }
     }
 
+    loadGeoJSONTile(params: WorkerTileParameters, callback: LoadVectorDataCallback): (() => void) | void {
```
HarelM, line 76: Here it is, without any changes.

src/source/geojson_worker_source.ts
```diff
@@ -101,12 +66,41 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
      * See {@link GeoJSONWorkerSource#loadGeoJSON}.
      */
     constructor(actor: Actor, layerIndex: StyleLayerIndex, availableImages: Array<string>, loadGeoJSON?: LoadGeoJSON | null) {
-        super(actor, layerIndex, availableImages, loadGeoJSONTile);
+        super(actor, layerIndex, availableImages);
+        this.loadVectorData = this.loadGeoJSONTile;
```
HarelM, line 70: Since this is what is being done in the constructor of the base class, doing it here explicitly is better I believe.

src/source/geojson_worker_source.ts
```diff
@@ -264,7 +258,7 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
         clusterId: number;
     }, callback: Callback<number>) {
         try {
-            callback(null, this._geoJSONIndex.getClusterExpansionZoom(params.clusterId));
+            callback(null, (this._geoJSONIndex as Supercluster).getClusterExpansionZoom(params.clusterId));
```
HarelM, line 261: These cast are now required due to proper typings of `_geoJSONIndex`. (x3)

src/source/geojson_wrapper.ts
```diff
@@ -40,7 +30,7 @@ class FeatureWrapper implements VectorTileFeature {
         // vector tile spec only supports integer values for feature ids --
         // allowing non-integer values here results in a non-compliant PBF
         // that causes an exception when it is parsed with vector-tile-js
-        if ('id' in feature && !isNaN(feature.id)) {
+        if ('id' in feature && !isNaN(feature.id as any)) {
```
HarelM, line 33: id was `any` and now it is `string | number`, but isNaN only receives number unfortunately...

HarelM:
This requires the following PR to be merged I think:
https://github.com/maplibre/maplibre-style-spec/pull/365
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66951
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66952

Let's see if the above are merged and published if this will pass all the tests...
Code wise, this is ready though.

src/source/geojson_worker_source.ts
neodescis: This function wasn't exposed before. Do we need to make it private?

HarelM: We can, but maplibre doesn't use private a lot.
The worker code runs on the worker thread so calling this method is almost impossible, but I don't mind marking it as private, let me know which direction you want to go...

src/source/geojson_worker_source.ts
neodescis: I think I disagree, as this change relies on the base class continuing to use the loadVectorData function in exactly the same way. If the constructor of the base class were to change how it uses or registers the function passed in, this may cease to work.

src/source/geojson_worker_source.ts
neodescis: Should GeoJSONIndex be `ReturnType<typeof geojsonvt> & Supercluster` instead? I believe you wouldn't need the cast here with that change.

HarelM: But it's not both - it's only one of the according to the code when assigning this variable value.

src/source/geojson_worker_source.ts
HarelM: Fair enough.

HarelM: Oh, wait, I can't pass a member function to the base class constructor as the class is still being created.
I can do the following instead, but I'm not sure it's better:
`super(actor, layerIndex, availableImages, (...args) => this.loadGeoJSONTile(...args));`

I think the bottom line, is that `loadVectorData` is a member of the base class, regardless of what is being done in the base constructor, for this derived class to work there's a need to override its behavior.
This is what is being done here.
Feel free to suggest a different way to solve this.

src/source/geojson_worker_source.ts
neodescis: Well, it's a good way to avoid expanding the API in general. Understood that it's on the worker thread, but I would always tend toward making something private unless there's a need for it to be public.

src/source/geojson_worker_source.ts
neodescis: That's funny because I almost suggested passing in `(...args) => this.loadGeoJSONTile(...args)` :-D

Actually, I also almost suggested that the base class have loadVectorData in it as a class method that could just be overridden here. That seems better, but maybe beyond the scope of what you were hoping to change.

src/source/geojson_worker_source.ts
neodescis: Thanks for the explanation. Sounds good to me.

HarelM:
@neodescis this is now ready for review and merge.
The definitely packages are updated and so is the style-spec.
I want to introduce this to improve the situation in #3233 for geojson source.
I'll probably change a few more things in the linked PR and I'm changing a lot of the infra between the workers and the main thread.

src/source/geojson_worker_source.ts
HarelM: I'll work on it as part of the #3233 PR.

src/source/geojson_worker_source.ts
HarelM: I'll address that as part of the #3233 PR.

birkskyum (approved):

PR #3185, opened 2023-10-07
Fix Raster DEM decoding in safari private browsing mode

### LINKED ISSUE #3110: Hillshade and 3d terrain look wrong in iOS 17 private browsing mode
I upgraded my iPhone 15 to iOS 17.0.1 and a few of the maplibre examples using terrain look wrong:

https://maplibre.org/maplibre-gl-js/docs/examples/3d-terrain/

image

https://maplibre.org/maplibre-gl-js/docs/examples/contour-lines/

image

And I occasionally get this warning at the top of the page:

image

The contour line problem is my issue over in the maplibre-contour plugin but hillshading and terrain are maplibre-gl-js. 

I suspect that it's related to this line from https://developer.apple.com/documentation/safari-release-notes/safari-17-release-notes

> Added noise to fingerprintable web APIs. (99360413)

msbarry:
Fix #3110 by conditionally using a [webcodecs](https://w3c.github.io/webcodecs/) `VideoFrame` API to get pixels from terrain RGB images when the browser mangles images to mitigate fingerprinting. This makes hillshading work for me on iOS 17 with enhanced privacy protections.

I think we should still prefer `OffscreenCanvas.getImageData` when it works since the image can come back in many [formats](https://w3c.github.io/webcodecs/#pixel-format), and we only handle RGB/BGR here.  There is a proposal to ask the browser to convert pixel formats for us: https://github.com/w3c/webcodecs/issues/92 - maybe when that's implemented more broadly we can use this by default when available.

I added console.log timing the performance of this method, and it takes around 0.5-1ms for a 512px tile on desktop chrome (M1 macbook pro) and about 5-10ms on an iPhone 14.

<table>
<tr>
<td>

terrain before

IMG_0967

</td>
<td>

terrain after

IMG_0966

</td>
<td>

contours before

IMG_0968

</td>
<td>

contours after

IMG_0965

</td>
<tr>
</table>

## Launch Checklist

 - [x] Link to related issues.
 - [x] Post benchmark scores.


src/source/raster_dem_tile_worker_source.ts
```diff
@@ -18,17 +19,22 @@ export class RasterDEMTileWorkerSource {
         this.loaded = {};
     }
 
-    loadTile(params: WorkerDEMTileParameters, callback: WorkerDEMTileCallback) {
+    async loadTile(params: WorkerDEMTileParameters, callback: WorkerDEMTileCallback) {
         const {uid, encoding, rawImageData, redFactor, greenFactor, blueFactor, baseShift} = params;
         // Main thread will transfer ImageBitmap if offscreen decode with OffscreenCanvas is supported, else it will transfer an already decoded image.
-        const imagePixels = isImageBitmap(rawImageData) ? this.getImageData(rawImageData) : rawImageData as RGBAImage;
+        const imagePixels = isImageBitmap(rawImageData) ? await this.getImageData(rawImageData) : rawImageData as RGBAImage;
```
msbarry, line 25: Can we use async/await for new code?

msbarry:
Looks like the bundle size goes up by 1344 bytes from 753k to 754k raw, and 498 bytes gzipped and that tripped the bundle size check.  Think that's too much?

src/util/offscreen_canvas_mangled.ts
```diff
@@ -0,0 +1,34 @@
+import {offscreenCanvasSupported} from './offscreen_canvas_supported';
+
+let manglesOffscreenCanvas: boolean;
+
+export function offscreenCanvasMangled(): boolean {
+    if (manglesOffscreenCanvas == null) {
+        manglesOffscreenCanvas = false;
+        // browser can mangle canvas#getImageData results when enhanced privacy protectsion are enabled (see #85)
+        if (offscreenCanvasSupported()) {
+            const canvas = new OffscreenCanvas(10, 10);
```
msbarry, line 10: do we think this needs tests?

msbarry:
> Add an entry to CHANGELOG.md under the ## main section.

Does this need a new entry in changelog?

src/source/raster_dem_tile_worker_source.ts
```diff
@@ -6,13 +6,16 @@ import type {
     WorkerDEMTileCallback,
     TileParameters
 } from './worker_source';
-import {isImageBitmap} from '../util/util';
+import {isImageBitmap, readImageUsingVideoFrame} from '../util/util';
+import {offscreenCanvasMangled} from '../util/offscreen_canvas_mangled';
+import {Callback} from '../types/callback';
+
+let offscreenCanvas: OffscreenCanvas;
+let offscreenCanvasContext: OffscreenCanvasRenderingContext2D;
```
msbarry, line 14: this reduces bundle size by a couple hundred bytes since these get minified to single-character variables but they need to be fully-expanded when they are members of the class.

src/util/util.ts
```diff
@@ -517,3 +517,73 @@ export function arrayBufferToImage(data: ArrayBuffer, callback: (err?: Error | n
...
+            }).then(() => {
+                if (swapBR) {
+                    for (let i = 0; i < result.length; i += 4) {
+                        const tmp = result[i];
+                        result[i] = result[i + 2];
+                        result[i + 2] = tmp;
+                    }
+                }
+                callback(null, result);
+            }).catch(e => callback(e)).finally(() => frame.close());
```
msbarry, line 585: This is a lot cleaner with promises using async/await:

```js
async readImageUsingVideoFrame() {
  try {
    // do work...
  } catch(e) {
    return undefined;
  } finally {
    frame.close();
  }
```

src/source/raster_dem_tile_worker_source.ts
birkskyum: That should be possible

birkskyum:
Will be interesting to see if this approach also fixes a longstanding issue with spikes in the terrain mesh that in some circumstances show up in the Brave browser.

- https://github.com/brave/brave-browser/issues/25149

src/source/raster_dem_tile_worker_source.ts
msbarry: The concern would be browsers that support the things we already require (webgl, promises) but not async/await. Looks like about 1% of users whose browser supports webgl and 0.7% of users whose browsers support promises don't support async/await:
- chrome 32-54 (ok since dec 2016)
- edge 12-14 (oc since oct 2016)
- safari 7.1-10.1 (ok since july 2018)
- firefox 27-51 (ok since march 2017)
- opera 19-41 (ok since dec 2016)

https://caniuse.com/webgl,promises,async-functions

So we might lose a little bit of browser compatibility here. I'm not sure if there are other features we use that might bring the difference down further though?

birkskyum: if async/await is problematic, we can just adjust the typescript compile target until the last percent is there

msbarry: OK sounds good, thanks! I put async/await back in.

CHANGELOG.md
```diff
@@ -24,6 +24,8 @@
 - Update raster tile end points in documentation
 - Avoiding inertia animation on Mac when reduced motion is on ([#3068](https://github.com/maplibre/maplibre-gl-js/pull/3068))
 - 3d buildings example doesn't work as expected ([#3165](https://github.com/maplibre/maplibre-gl-js/pull/3165))
+- Fix Raster DEM decoding in safari private browsing mode ([#3185](https://github.com/maplibre/maplibre-gl-js/pull/3185))
```
HarelM, line 27: 3.4 was released, please move this up under main.

src/source/raster_dem_tile_worker_source.ts
HarelM: If this is needed only to support reducing bundle size then let's **not** do it, having global variables isn't a good solution IMHO.
Unless you want to move this to a different file in order to improve performance, so that they don't need to be in the class but rather in the file, there are other examples of this thought-out the code.

src/source/raster_dem_tile_worker_source.ts
```diff
@@ -6,43 +6,54 @@ import type {
...
         callback(null, dem);
     }
 
-    getImageData(imgBitmap: ImageBitmap): RGBAImage {
+    async getImageData(imgBitmap: ImageBitmap) {
+        const width = imgBitmap.width;
+        const height = imgBitmap.height;
+        const newWidth = width + 2;
+        const newHeight = height + 2;
+        if (offscreenCanvasMangled()) {
```
HarelM, line 38: `isOffscreenCanvasMangled`? `isOffscreenCanvasDistorted`?

src/source/raster_dem_tile_worker_source.ts
```diff
@@ -6,43 +6,54 @@ import type {
...
-        const imagePixels = isImageBitmap(rawImageData) ? this.getImageData(rawImageData) : rawImageData as RGBAImage;
+        const imagePixels = isImageBitmap(rawImageData) ? await this.getImageData(rawImageData) : rawImageData as RGBAImage;
         const dem = new DEMData(uid, imagePixels, encoding, redFactor, greenFactor, blueFactor, baseShift);
         this.loaded = this.loaded || {};
         this.loaded[uid] = dem;
         callback(null, dem);
     }
 
-    getImageData(imgBitmap: ImageBitmap): RGBAImage {
+    async getImageData(imgBitmap: ImageBitmap) {
```
HarelM, line 33: Add return value - `Promise<RGBAImage>`

src/source/raster_dem_tile_worker_source.ts
```diff
@@ -6,43 +6,54 @@ import type {
...
+        const width = imgBitmap.width;
+        const height = imgBitmap.height;
+        const newWidth = width + 2;
+        const newHeight = height + 2;
+        if (offscreenCanvasMangled()) {
+            const parsed = await readImageUsingVideoFrame(imgBitmap, -1, -1, newWidth, newHeight);
+            if (parsed) return new RGBAImage({width: newWidth, height: newHeight}, parsed);
```
HarelM, line 40: What happens if this is not parsed? do we return to "regular" flow? Should we throw an exception here?

src/source/raster_dem_tile_worker_source.ts
```diff
@@ -6,43 +6,54 @@ import type {
...
+        }
+
         // Lazily initialize OffscreenCanvas
-        if (!this.offscreenCanvas || !this.offscreenCanvasContext) {
+        if (!offscreenCanvas || !offscreenCanvasContext) {
```
HarelM, line 44: This was per class, and now it is global - I'm not sure if this can affect inheritance and suff like that...
I would consider moving this logic to a different class maybe - `OffscreenCanvasProvider` or something like that, that will manage the lifecycle of this canvas...
IDK...

src/source/raster_dem_tile_worker_source.ts
```diff
@@ -6,43 +6,54 @@ import type {
...
             // Dem tiles are typically 256x256
-            this.offscreenCanvas = new OffscreenCanvas(imgBitmap.width, imgBitmap.height);
-            this.offscreenCanvasContext = this.offscreenCanvas.getContext('2d', {willReadFrequently: true});
+            offscreenCanvas = new OffscreenCanvas(width, height);
+            offscreenCanvasContext = offscreenCanvas.getContext('2d', {willReadFrequently: true});
         }
 
-        this.offscreenCanvas.width = imgBitmap.width;
-        this.offscreenCanvas.height = imgBitmap.height;
+        offscreenCanvas.width = width;
```
HarelM, line 50: I would consider moving this to a different file much like the video frame - `readImageUsingOffscreenCanvas(...)`
This would make this method a lot cleaner.

src/util/util.ts
```diff
@@ -517,3 +517,76 @@ export function arrayBufferToImage(data: ArrayBuffer, callback: (err?: Error | n
...
+ * @param y - top-left y coordinate to read from the image
+ * @param width - width of the rectangle to read from the image
+ * @param height - height of the rectangle to read from the image
+ * @returns a promise containing the parsed RGBA pixel values of the image, or undefined if an error occurred
+ */
+export async function readImageUsingVideoFrame(
+    image: HTMLImageElement | HTMLCanvasElement | ImageBitmap | OffscreenCanvas,
+    x: number, y: number, width: number, height: number
+): Promise<Uint8ClampedArray> {
+    if (typeof VideoFrame !== 'undefined') {
```
HarelM, line 535: Invert if to reduce indentation and improve readability.

src/util/util.ts
```diff
@@ -517,3 +517,76 @@ export function arrayBufferToImage(data: ArrayBuffer, callback: (err?: Error | n
...
+        const frame = new VideoFrame(image, {timestamp: 0});
+        try {
+            const format = frame?.format;
+            if (format && (format.startsWith('BGR') || format.startsWith('RGB'))) {
```
HarelM, line 539: Is there an else to this statement? Consider inverting this as well.

src/util/util.ts
```diff
@@ -517,3 +517,76 @@ export function arrayBufferToImage(data: ArrayBuffer, callback: (err?: Error | n
...
+                const swapBR = format.startsWith('BGR');
+                /*
+                Offset is the byte offset in the dest image that the first pixel appears at
```
HarelM, line 542: Consider moving this to the method comment instead of here.
This is OK as well, I just prefer not having comments inside a method, it makes is harder to read.
You can maybe extract it to a different method and have that comment on that method.

src/util/util.ts
```diff
@@ -517,3 +517,76 @@ export function arrayBufferToImage(data: ArrayBuffer, callback: (err?: Error | n
...
+                    }
+                }
+                return result;
+            }
+        } catch (e) {
+            // return undefined on failure
+        } finally {
+            frame.close();
+        }
+    }
```
HarelM, line 591: I would consider returning something explicitly in case of error/problem etc.
Reading this method it is not clear what is returned in case of an error or problem.

src/util/util.test.ts
```diff
@@ -1,5 +1,6 @@
 import Point from '@mapbox/point-geometry';
-import {arraysIntersect, asyncAll, bezier, clamp, clone, deepEqual, easeCubicInOut, extend, filterObject, findLineIntersection, isClosedPolygon, isCounterClockwise, isPowerOfTwo, keysDifference, mapObject, nextPowerOfTwo, parseCacheControl, pick, uniqueId, wrap} from './util';
+import {arraysIntersect, asyncAll, bezier, clamp, clone, deepEqual, easeCubicInOut, extend, filterObject, findLineIntersection, isClosedPolygon, isCounterClockwise, isPowerOfTwo, keysDifference, mapObject, nextPowerOfTwo, parseCacheControl, pick, readImageUsingVideoFrame, uniqueId, wrap} from './util';
+import {promisify} from 'util';
```
HarelM, line 3: Please remove this

src/util/util.test.ts
```diff
@@ -346,3 +347,153 @@ describe('util findLineIntersection', () => {
         expect(intersection).toBeNull();
     });
 });
+
+describe('util readImageUsingVideoFrame', () => {
```
HarelM, line 351: GREAT!! THANKS!

src/util/offscreen_canvas_mangled.ts
```diff
@@ -0,0 +1,31 @@
+import {offscreenCanvasSupported} from './offscreen_canvas_supported';
+
+let manglesOffscreenCanvas: boolean;
+
+export function offscreenCanvasMangled(): boolean {
+    if (manglesOffscreenCanvas == null) {
+        manglesOffscreenCanvas = false;
+        // browser can mangle canvas#getImageData results when enhanced privacy protectsion are enabled (see #85)
```
HarelM, line 8:
```suggestion
        // browser can mangle canvas#getImageData results when enhanced privacy protection is enabled (see #3185)
```

Or did you mean `#3110`?

src/source/raster_dem_tile_worker_source.ts
```diff
@@ -6,43 +6,54 @@ import type {
...
     loaded: {[_: string]: DEMData};
-    offscreenCanvas: OffscreenCanvas;
-    offscreenCanvasContext: OffscreenCanvasRenderingContext2D;
 
     constructor() {
         this.loaded = {};
     }
 
-    loadTile(params: WorkerDEMTileParameters, callback: WorkerDEMTileCallback) {
+    async loadTile(params: WorkerDEMTileParameters, callback: WorkerDEMTileCallback) {
```
HarelM, line 23: There are missing tests to this new flow of loadTile I think, can you please add them?

src/source/raster_dem_tile_worker_source.ts
HarelM: I would consider moving all the code from here to the utils class - i.e. move `getImageData` to utils class so that it can be used as is to read images.

src/source/raster_dem_tile_worker_source.ts
HarelM: Another though - now that the method is async it can be interuppted in the middle of the flow - this needs to be taken into consideration, I think, not sure I have a lot of insight here to what it may do.

HarelM:
I've reviewed this file and added my comments.
Mainly nit-picking...
THANKS!!!

birkskyum:
Hold merging this for a sec - the await/async isn't resolved.
I tried setting the compile target here #3188 , but it broke a ton of stuff.

We have three options:
- use two tsconfig.json, one for running tests, one for compiling the library
- not use await/async here.
- let await/async through to the browser, utilizing the 94.9% browser support, but giving up on on Safari < 10.

HarelM:
Removing top level await doesn't sound too complicated.
Not using async await is not a solution I'm willing to entertain.
I'll wait for the other PR.
Option 4 is also something I'm willing to consider, we can't support both safari 10 and 17 in the same version, it's too much...

birkskyum:
I tried getting option 2 going, but the render tests are tricky, because I can't see any usage of top-level await, but it still complains about it.

It feels a bit wrong to be introducing clutter all over the test suite to avoid a bit of clutter here.

I'm now leaning more towards option 4. The support we let go here are distributed on browsers all dating back to at least 2016:
Android browser <55 = 0.34% usage
Safari <10 = 0.19% usage
Samsung internet <6 = 0.11% usage
Chrome <55 = 0.10% usage

src/source/raster_dem_tile_worker_source.ts
msbarry: now that this is extracted into a util do you think this still applies? I tried forcing a unit test to take this path, but wasn't able to easily because you need the input to be an `ImageBitmap` and the `Canvas` doesn't quite work with it the same way that the browser API does.

msbarry:
Thanks for the review, I moved `getImageData` into `util.ts` and made applied some of the other suggestions as well.

I also vote for option 4 eventually since the transpiler will likely expand async/await to a lot of extra code that everyone needs to download for the few users on 5+ year old browsers.  Maybe the microsoft/bing team has some input here?

In the short term, is option 1 that difficult?

src/source/raster_dem_tile_worker_source.ts
HarelM: This wasn't covered before your code changes, so I guess we can forfeit it:
image

src/source/raster_dem_tile_worker_source.ts
```diff
@@ -6,46 +6,30 @@ import type {
...
         const {uid, encoding, rawImageData, redFactor, greenFactor, blueFactor, baseShift} = params;
         // Main thread will transfer ImageBitmap if offscreen decode with OffscreenCanvas is supported, else it will transfer an already decoded image.
-        const imagePixels = isImageBitmap(rawImageData) ? this.getImageData(rawImageData) : rawImageData as RGBAImage;
+        const width = rawImageData.width + 2;
+        const height = rawImageData.height + 2;
+        const imagePixels = isImageBitmap(rawImageData) ?
+            new RGBAImage({width, height}, await getImageData(rawImageData, -1, -1, width, height)) :
+            rawImageData as RGBAImage;
```
HarelM, line 26: There is a way in typescript to make a boolean method force the type so that you won't require `as RGBAImage`, do you think it's possible to do it here:
i.e. write `isRGBAImage` instead of `isBitmapImage` and force the type to be `RBGAImage`?

msbarry: would this work instead:

```js
        const imagePixels = rawImageData instanceof RGBAImage ? rawImageData :
            new RGBAImage({width, height}, await getImageData(rawImageData, -1, -1, width, height));
```

HarelM: If it transiples OK then I think it's a good replacement, yes.

src/util/util.ts
```diff
@@ -1,6 +1,8 @@
 import Point from '@mapbox/point-geometry';
 import UnitBezier from '@mapbox/unitbezier';
 import type {Callback} from '../types/callback';
+import {isOffscreenCanvasDistorted} from './offscreen_canvas_distorted';
+import {Size} from './image';
```
HarelM, line 5: `import type {Size} ...`?

src/util/offscreen_canvas_distorted.ts
```diff
@@ -0,0 +1,31 @@
+import {offscreenCanvasSupported} from './offscreen_canvas_supported';
+
+let offscreenCanvasDistorted: boolean;
+
+export function isOffscreenCanvasDistorted(): boolean {
```
HarelM, line 5: It's probably worth explaining what this method is doing - if I understand correctly, it draws a rectangle and check that all the pixels are with the same color...?

src/source/raster_dem_tile_worker_source.ts
msbarry: looks like it transpiles to:

```js
        const imagePixels = rawImageData instanceof performance.RGBAImage ? rawImageData :
            new performance.RGBAImage({ width, height }, await performance.getImageData(rawImageData, -1, -1, width, height));
```

in `dist/maplibre-gl-dev.js`

HarelM: I mean that if you don't get typescript errors then it's fine I think.

HarelM (approved):
THANKS!!!

src/source/raster_dem_tile_worker_source.ts
msbarry: Turns out we didn't need any change - `isImageBitmap` returned type `image is ImageBitmap` so typescript was smart enough to know in the else case that rawImageData is an `RGBAImage` without any casting.

HarelM: Great!

HarelM:
@birkskyum let me know if you would like to do something before merging this PR.
Generally speaking, it is ready to be merged.

msbarry:
Thanks @HarelM !  Also I noticed that if `ImageRequest.getImage` returns an HTML image then it uses `browser.ts#getImageData` instead which still mangles the result.  What do you think about having `raster_tem_tile_source.ts` (or maybe `browser.ts`) try using `util.ts#getImageData` before falling-back to DOM canvas image decoding?

HarelM:
I would consider it as a different PR...?
We need to measure the performance impact here I believe...

msbarry:
> I would consider it as a different PR...? We need to measure the performance impact here I believe...

Good point, we could limit scope to just raster dem tile source, and only when offscreen canvas is distorted?

```diff
-                const rawImageData = transfer ? img : browser.getImageData(img, 1);
+                const rawImageData = transfer ? img :
+                    isOffscreenCanvasDistorted() ? new RGBAImage({width: img.width + 2, height: img.height + 2}, await readImageUsingVideoFrame(img, -1, -1, img.width + 2, img.height + 2)) :
+                        browser.getImageData(img, 1);
```

msbarry:
I applied that change in https://github.com/maplibre/maplibre-gl-js/pull/3185/commits/f12470cf22e06c2db2446a0b2e82ad676df29afb - I think it should have minimal performance implication, and only change behavior when the image gets mangled but let me know if you have any concerns.

msbarry:
> we can't support both safari 10 and 17 in the same version, it's too much...

Another datapoint here is that Apple's support for Safari version 10 ended August 2017 according to https://en.wikipedia.org/wiki/OS_X_Yosemite

birkskyum:
We'll still be supporting Safari >= 10.1 , so it's actually only 10.0 that's unsupported.

msbarry:
Ugh unfortunately it looks like that pushed this PR just over the 1024 byte increase limit from 770837 - should I bump it?

msbarry:
> I wonder if there's a way for rollup to minify names shared across the chunks...?

Yep... there sure is:

```diff
--- a/rollup.config.ts
+++ b/rollup.config.ts
@@ -22,7 +22,8 @@ const config: RollupOptions[] = [{
         format: 'amd',
         sourcemap: 'inline',
         indent: false,
-        chunkFileNames: 'shared.js'
+        chunkFileNames: 'shared.js',
+        minifyInternalExports: true
     },
     onwarn: (message) => {
         console.error(message);
```

That brings this from a 1220 byte increase to a 9397 decrease.    Think we should pursue that?

HarelM:
Yes, but not in this PR.
Let's have this merged, including size increase, release a version and have that in the next version so that we'll have a way to pinpoint issues that these two separate features can create...

msbarry:
Sounds good, I bumped the quota and opened #3194 to track.

src/source/raster_dem_tile_source.ts
```diff
@@ -90,6 +92,19 @@ export class RasterDEMTileSource extends RasterTileSource implements Source {
             }
         }
 
+        async function readImageImageNow(img: ImageBitmap | HTMLImageElement): Promise<RGBAImage | ImageData> {
```
HarelM, line 95: can you explain the name? It is a mistake?

src/source/raster_dem_tile_source.ts
```diff
@@ -90,6 +92,19 @@ export class RasterDEMTileSource extends RasterTileSource implements Source {
...
+            if (typeof VideoFrame !== 'undefined' && isOffscreenCanvasDistorted()) {
```
HarelM, line 96: Is `browser.getImageData(...)` used elsewhere that wouldn't require the video frame to be part of the implementation?
It looks to me like this code should be moved inside the `browser.getImageData` method, but I might be missing something obvious here.

HarelM:
I've added a few last comments, overall, this looks good. THANKS!!!

src/source/raster_dem_tile_source.ts
msbarry: It's used in `map.ts` as well in add/update image - I don't think it's strictly necessary there because the imperceptible noise that that safari adds to getImageData isn't noticeable to the naked eye in a regular image, just when we depend on exact pixel values to get the elevation.

It adds a bit of complexity too because it woudl turn browser.getImageData into an async function which would require changes to those methods in `map.ts`

HarelM: I see, OK then.

src/source/raster_dem_tile_source.ts
```diff
@@ -57,7 +59,7 @@ export class RasterDEMTileSource extends RasterTileSource implements Source {
         tile.request = ImageRequest.getImage(this.map._requestManager.transformRequest(url, ResourceType.Tile), imageLoaded.bind(this), this.map._refreshExpiredTiles);
 
         tile.neighboringTiles = this._getNeighboringTiles(tile.tileID);
-        function imageLoaded(err: Error, img: (HTMLImageElement | ImageBitmap) & ExpiryData) {
+        async function imageLoaded(err: Error, img: (HTMLImageElement | ImageBitmap) & ExpiryData) {
```
HarelM, line 62: If we are changing this method can we also change how we call it?
I don't like the `imageLoaded.bind(this)` part.
I also don't know if `ImageRequest.getImage` can handle a promise as a return type...?

msbarry: Sounds good, I changed it a bit to inline imageLoaded like image_source does it. I also noticed that expiry got broken-out into a 3rd argument so it couldn't have been working as-is but should be fixed now.

What I'd really like to do is change the entire function to use async/await without any callbacks, something like

```js
try {
  tile.request = ImageRequest.getImage(request, this.map._refreshExpiredTiles);
  const image = await tile.request.image;
  const transfer = isImageBitmap(img) && offscreenCanvasSupported();
  const rawImageData = transfer ? img : await readImageNow(img);
  const params = { ... };
  if (!tile.actor || tile.state === 'expired') {
      const data = await tile.actor.send('loadDEMTile', params);
      if (data) {
          tile.dem = data;
          tile.needsHillshadePrepare = true;
          tile.needsTerrainPrepare = true;
          tile.state = 'loaded';
          callback(null);
      }
  }
} catch (err) {
  tile.state = 'errored';
  callback(err);
}
```

But that would require some larger refactoring to `ImageRequest` and the actor API.  I implemented a similar API in maplibre-contour, except all promise-based, we might want to use some utilities to make promise-based code a little easier to write here, like [CancelablePromise](https://github.com/onthegomap/maplibre-contour/blob/8c000c1b32daa9fefe05dd93350ee46e9516e8d0/src/types.ts#L23-L29) and a promise-based [actor implementation](https://github.com/onthegomap/maplibre-contour/blob/main/src/actor.ts) that checks the remote method being called and its argument types when you do `actor.send('method', args);`

HarelM (approved):

HarelM:
P.S. if you would like to have a version released, please open a PR with version bump and I'll merge and carve out a release.
Thanks for all the hard work invested here in this!!

msbarry:
Thanks @HarelM ! Do you think we should do a release that potentially drops support for a small percentage of browsers because of async/await? Or @birkskyum should we try to get the transpiler working before a release?

HarelM:
I'm OK with releasing it as is, and if there's a requirement later on to address this, the person/company that needs this can push it forward like you did in this case.

birkskyum:
Just to give a bit of context. Compiling it down to ES2016 adds just 1 kb to maplibre-gl.js, so it's 772 -> 773 kb, so that concern is not too big. It's a bit cumbersome to do though, and for anyone interested in undertaking this, the most feasible path seems to be having 2 tsconfig.json files, one for all tests and build scripts with >= es2017 target (due to usage of top-level await), and one for library source files with <= es2016 target.

I'm also in favor of releasing it.

msbarry:
OK thanks, I opened #3197 to do a patch version bump. I double-checked onthegomap traffic over the last week and it looks like only about 0.02% of users would fall into the safari/chrome/firefox version range that supports webgl but not async/await so on my end I'm fine with this without transpilation.

birkskyum:
This fixed Brave support as well 👍

PR #3188, opened 2023-10-08
Set TS compile target to es2016

birkskyum:
This should compile await/async down, maintaining a very high level of browser support.

Preparation for
- https://github.com/maplibre/maplibre-gl-js/pull/3185

Context 
- https://github.com/maplibre/maplibre-gl-js/pull/3185#discussion_r1349429451
- https://github.com/maplibre/maplibre-gl-js/pull/3185#discussion_r1349518420

Removed usage of top-level await from build scripts, because it won't allow us to have a compile target below es2017, thus forcing us to ship await/async to browsers and loosing Safari <10 support (0.19% browser usage).

### DIFF
--- a/build/generate-dist-package.js
+++ b/build/generate-dist-package.js
@@ -11,16 +11,20 @@ async function ensureDist() {
   return dist
 }
 
-await writeFile(
-  new URL("package.json", await ensureDist()),
-  JSON.stringify({
-    name: "maplibre-gl",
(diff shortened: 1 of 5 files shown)


birkskyum:
oh my, this would require removing all top-level awaits from tests, or having two tsconfig.json

birkskyum:
I'm attempting to remove top-level await in this branch as well.

Issue #3180, opened 2023-10-06
addProtocol callback is not using cacheControl and expires


emanueleperuffo:
**maplibre-gl-js version**: 3.3.1

**browser**: Chrome

### Steps to Trigger Behavior

 1. Add a new protocol with `addProtocol`
 2. Call the call back with the array buffer and with `cacheControl` and `expires` arguments
 3. You'll notice that any argument after `data` is ignored in the `addProtocol` response callback

### Link to Demonstration

I provide a screenshot, because the issue is in the implementation of the callback passed to the `loadFn` function in `addProtocol`.

![Screenshot from 2023-10-06 15-21-41](https://github.com/maplibre/maplibre-gl-js/assets/5085029/9e4c604c-1e31-4306-bd81-39b8e73b7133)

### Expected Behavior

The tile will reload after it expires, also because the type `ResponseCallback` is defined as `(error?: Error | null, data?: T | null, cacheControl?: string | null, expires?: string | null) => void;`

### Actual Behavior

The tile does not reload after it expires. As you can see from the screenshot, the stringified callback is not using the 3rd and 4th arguments of the callback.

HarelM:
Please provide a jsbin/codepen/stackblitz that reproduce this issue.
Please make sure you create a reproduction without any wrapper like react-map etc as it might be a wrapper issue.
Add protocol basically passes the callback it received, note that cache is not relevant for geojson source.

emanueleperuffo:
Hi @HarelM, I have created this https://stackblitz.com/edit/typescript-ey8qy2.

The custom protocol replies with a `null` because I wouldn't know how else to create a vector tile with a short validity, but I would expect anyway the protocol function to be called every 10 seconds.

HarelM:
I'm not sure, since the current method isn't returning anything it might be that the cache control isn't "activated".
You can simply `fetch` a tile, read it as `arraybuffer` and return it to test your theory.

emanueleperuffo:
Thanks! I updated the code to make it fetching real vector tile data and making it display only the centroids layer as circles.

johncarmack1984:
The callback form of addProtocol was replaced by the promise form in #3433 (v4.0.0), whose response carries cacheControl and expires. This can be closed.

PR #3196, opened 2023-10-10
Fix actor not passing additional callback arguments

emanueleperuffo:
Before this change, `Actor` was not forwarding additional callback arguments and assumed the callback is always in the form `(err, data) => {}`.

There are examples where the callback has more than 2 arguments. For example, in the `loadVectorTile` function implementation, the callback that is passed directly to the actor is defined as `(err?: Error | null, data?: ArrayBuffer | null, cacheControl?: string | null, expires?: string | null) => {}`. In this case, `Actor` only passes `data`, *forgetting* `cacheControl` and `expires`.

Linked issue: fixes https://github.com/maplibre/maplibre-gl-js/issues/3180

## Launch Checklist

 - [x] Link to related issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### 🐞 Bug fixes
 
 - _...Add new stuff here..._
-- Fix actor passing only the first 2 arguments to callbacks
+- Fixed actor passing only the first 2 arguments to callbacks ([#3130](https://github.com/maplibre/maplibre-gl-js/pull/3196))
 
 ## 3.5.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### 🐞 Bug fixes
 
 - _...Add new stuff here..._
+- Fix actor passing only the first 2 arguments to callbacks
 
 ## 3.5.2
 
--- a/src/util/actor.ts
+++ b/src/util/actor.ts
@@ -213,13 +213,13 @@ export class Actor {
                 if (task.error) {
                     callback(deserialize(task.error));
                 } else {
-                    callback(null, deserialize(task.data));
+                    callback(null, ...(deserialize(task.data) as any[]));
                 }
             }
         } else {
             let completed = false;
             const buffers: Array<Transferable> = [];
-            const done = task.hasCallback ? (err: Error, data?: any) => {
+            const done = task.hasCallback ? (err: Error, ...data: any[]) => {
                 completed = true;
                 delete this.cancelCallbacks[id];
                 const responseMessage: MessageData = {
--- a/src/util/actor.ts
+++ b/src/util/actor.ts
@@ -31,26 +31,34 @@ export interface GlyphsProvider {
     );
 }
 
-export type MessageType = '<response>' | '<cancel>' |
-'geojson.getClusterExpansionZoom' | 'geojson.getClusterChildren' | 'geojson.getClusterLeaves' | 'geojson.loadData' |
+export type MessageType = 'geojson.getClusterExpansionZoom' | 'geojson.getClusterChildren' |
+'geojson.getClusterLeaves' | 'geojson.loadData' |
 'removeSource' | 'loadWorkerSource' | 'loadDEMTile' | 'removeDEMTile' |
 'removeTile' | 'reloadTile' | 'abortTile' | 'loadTile' | 'getTile' |
 'getGlyphs' | 'getImages' | 'setImages' |
 'syncRTLPluginState' | 'setReferrer' | 'setLayers' | 'updateLayers';
 
-export type MessageData = {
+type BasicMessage = {
     id: string;
-    type: MessageType;
-    data?: Serialized;
-    targetMapId?: string | number | null;
-    mustQueue?: boolean;
-    error?: Serialized | null;
-    hasCallback?: boolean;
+    targetMapId?: string | null;
     sourceMapId: string | number | null;
 }
 
-export type Message = {
-    data: MessageData;
+type RequestMessage = BasicMessage & {
+    type: MessageType;
+    hasCallback: boolean;
+    mustQueue: boolean;
+    params: Serialized;
+}
+
+type ResponseMessage = BasicMessage & {
+    type: '<response>';
+    error?: Serialized | null;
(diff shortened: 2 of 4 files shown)


HarelM:
Can you please write a test that this issue solves? While the unit test is great, I would like to see the actual bug this PR solves...

HarelM:
P.S. thanks for taking the time to open this PR!

emanueleperuffo:
Hi @HarelM, can you help me here? How could I write such test?
I have added a stackblitz that I shared in the linked issue #3180, is it enough?

HarelM:
Since this requires the usage of actors and in between thread commination, the only thing I can think of is a e2e test.
We have those under browser.test.ts - these test loads the browser and you can write some code to be executed.
You can mock the backend to return the relevant tiles I believe and change the expiry date to be minimal - a second or even less.
I can't think of a better way to test this...

emanueleperuffo:
I have added the e2e test.

HarelM:
Thanks!! Can you please verify that the test fails without the fix?

emanueleperuffo:
I verified it and it was failing. I guess though that trust is not ehough?

HarelM:
Trust is enough 😀 you can create a PR with the test only to show that it is failing, but it's not necessary.

Thanks for all the hard work!

test/integration/browser/browser.test.ts
```diff
@@ -248,4 +248,65 @@ describe('Browser tests', () => {
...
+                });
+
+                setTimeout(() => {
+                    resolve(callCountByUrl);
+                }, 5000);
+            });
+        });
+
+        for (const [, count] of  Object.entries(callCountByUrl)) {
+            expect(count).toBeGreaterThan(1);
```
HarelM, line 309: Greater than 3-4? If the timeout is for 5 sec and the cache is for 1 sec?

src/util/actor.ts
```diff
@@ -175,13 +175,13 @@ export class Actor {
                 if (task.error) {
                     callback(deserialize(task.error));
                 } else {
-                    callback(null, deserialize(task.data));
+                    callback(null, ...(deserialize(task.data) as any[]));
```
HarelM, line 178: I still don't fully understand how this solves the problem... I mean task.data can be anything, so why cache control and expiry are not part of it?
Also if we used to send the data as a single object, now it is spread, isn't it?

emanueleperuffo: `task.data` is the 1st argument after the `err` argument of the callback.

As you can see from the piece of code in `Actor.ts` in the below screenshot, the `done`  callback is what is passed to the real function, and probably in some other places that I ignore right now.

Actor in fact proxies to-from the web worker and keeps a cache of the _real_ callbacks to call on the other side of the message (postMessage and message listener). It calls the function that has the logic with its new callback (here the `done`) instead of the _real_ callback. Then it passes the result back to the _real_ callback.

Prior to this change, `...data` was just `data`. Any other argument after `data` was just dropped. So if in my function I called `callback(null, data, cacheControl, expires)`, it was exactly like calling `done(null, data)` (here data is the array buffer of the vector tile in the example I gave with addProtocol). This because the `callback` is this `done` that you can see in `Actor`.

As you can see also the spread is both in the actor while it calls the _real_ callback, but also in the `done` callback where it acquires the arguments, so it's in both places. Probably to make things more clear we could rename `task.data` to `task.args`.

Screenshot from 2023-10-12 14-41-22

HarelM: args makes more sense yes.
What do you mean by other places?

emanueleperuffo: I have seen some other mention of cacheControl and expiry in the arguments of the callback passed to the function loading the geojson if I recall correctly.

emanueleperuffo: For geojson sources I mean.

emanueleperuffo: I have tried to take a look at renaming `data` to args, but I wouldn't feel safe in doing so because `data` and `task` are typed as `any` in a lot of places, and a refactor is not that easy... Maybe this is a task for the future when typing is a bit more strict in this repo.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### 🐞 Bug fixes
 
 - _...Add new stuff here..._
+- Fix actor passing only the first 2 arguments to callbacks
```
HarelM, line 10: Please add a PR link.

emanueleperuffo: Done

src/util/actor.ts
HarelM: After reading the code a bit more and this change it seems too risky for me, as I don't see the full picture yet, and I'm not certain that this change doesn't break anything...
I would recommend first creating a PR which only add types so that we know we don't break anything (adding types doesn't change the code), and after that we can introduce this fix.
See the following comment for more information, there's also some implementation with types as far as I've seen in the contour repo: https://github.com/maplibre/maplibre-gl-js/pull/3185#discussion_r1352433448
cc: @msbarry 

Sorry for dragging this, but better safe than sorry...

HarelM:
No need to close this, it can wait for the other PR I believe...

emanueleperuffo:
I have added a new commit improving the typings in Actor.

HarelM:
Due to this issue and PR, I have started to replace all the actor communication layer with types and promises.
You can take a look here:
https://github.com/maplibre/maplibre-gl-js/pull/3269
Latest code in this branch is the most advanced one.
I suggest opening a PR against these branches to fix what you have seen in this issue.
The types, messages and everything there is well defined and you should be able to add the missing data in the right place.
I would recommend closing this PR.

emanueleperuffo:
Ok I understand.

However, would you mind giving a look at the last commit in my PR?

Actually I think that the PR is not up to date with the branch in my forked repo because of a force push, so the branch to look at is the PR branch on my repo instead of this PR "file changes" section.

I have implemented some more strict typing for the messages: one for the request, one for the response and one for the cancel, instead of having a generic Message that doesn't distinguish and contains all the fields.

HarelM:
I took it a step forward and added promises etc.
Take a look at what I did, you can even try my branch.
If there's a missing data that is not send somewhere you should be able to add it easily, most of the worker code is not typed well, I hope...

PR #3208, opened 2023-10-11
Add RasterTileSource setTiles method

### LINKED ISSUE #3207: Add setTiles capability to RasterTileSource
## User Story

Add ability to dynamically `setTiles` on a RasterTileSource, similar to [what's already possible for VectorTileSource](https://github.com/maplibre/maplibre-gl-js/blob/9dcb8715d82107f415fddc0eea932a348a973b39/src/source/vector_tile_source.ts#L155). 

## Rationale

Our map application has the ability to select dates and update layers on-the-fly with geospatial data from the date selected. We're using GeoJSON, vector tile sources, and would like to add imagery as well, but we're currently blocked by the inability to update tile sources with the lastest URL & params. The workaround is to completely remove the source & associated layers and then re-add it as a new source with an updated tiles field, but it's not a smooth experience.

## Impact

Vector and vector tile sources are certainly where a lot of modern development is happening, but there is a lot of satellite imagery captured and served in raster format. Having the capability to show this imagery over time is really powerful.

wagewarbler:
Resolves #3207. Adds a `setTiles` method to `RasterTileSource`. Changes made:

- Add `setTiles` to `RasterTileSource`
- Add unit test to confirm tiles update
- Changelog entry

[Screencast from 2023-10-11 13-43-31.webm](https://github.com/maplibre/maplibre-gl-js/assets/133472204/9ee52951-5289-4710-a77a-2418da89039a)

### DIFF
--- a/src/source/raster_tile_source.test.ts
+++ b/src/source/raster_tile_source.test.ts
@@ -164,6 +164,17 @@ describe('RasterTileSource', () => {
         server.respond();
     });
 
+    test('supports updating tiles', () => {
+        const source = createSource({url: '/source.json'});
+        source.setTiles(['http://example.com/{z}/{x}/{y}.png?updated=true']);
+
+        source.on('data', (e) => {
+            if (e.sourceDataType === 'metadata') {
+                expect(source._options.tiles[0]).toBe('http://example.com/{z}/{x}/{y}.png?updated=true');
+            }
+        });
+    });
+
     test('cancels TileJSON request if removed', () => {
         const source = createSource({url: '/source.json'});
         source.onRemove();
--- a/src/source/raster_tile_source.test.ts
+++ b/src/source/raster_tile_source.test.ts
@@ -170,7 +170,7 @@ describe('RasterTileSource', () => {
 
         source.on('data', (e) => {
             if (e.sourceDataType === 'metadata') {
-                expect(source._options.tiles[0]).toBe('http://example.com/{z}/{x}/{y}.png?updated=true');
+                expect(source.tiles[0]).toBe('http://example.com/{z}/{x}/{y}.png?updated=true');
             }
         });
     });
--- a/src/source/raster_tile_source.ts
+++ b/src/source/raster_tile_source.ts
@@ -131,7 +131,7 @@ export class RasterTileSource extends Evented implements Source {
      * @returns `this`
      */
     setTiles(tiles: Array<string>): this {
-        this._options.tiles = tiles;
+        this.tiles = tiles;
         this.onRemove();
         this.load();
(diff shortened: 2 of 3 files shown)


src/source/raster_tile_source.test.ts
```diff
@@ -164,6 +164,17 @@ describe('RasterTileSource', () => {
         server.respond();
     });
 
+    test('supports updating tiles', () => {
+        const source = createSource({url: '/source.json'});
+        source.setTiles(['http://example.com/{z}/{x}/{y}.png?updated=true']);
+
+        source.on('data', (e) => {
+            if (e.sourceDataType === 'metadata') {
+                expect(source._options.tiles[0]).toBe('http://example.com/{z}/{x}/{y}.png?updated=true');
```
HarelM, line 173: Can we test something that is not an internal property.

src/source/raster_tile_source.ts
```diff
@@ -124,6 +124,20 @@ export class RasterTileSource extends Evented implements Source {
...
 
+    /**
+     * Sets the source `tiles` property and re-renders the map.
+     *
+     * @param tiles - An array of one or more tile source URLs, as in the raster tiles spec (See the [Style Specification](https://maplibre.org/maplibre-style-spec/)
+     * @returns `this`
+     */
+    setTiles(tiles: Array<string>): this {
+        this._options.tiles = tiles;
+        this.onRemove();
```
HarelM, line 135: Is there a better way other than remove and load again?

wagewarbler: This appears to be how it's done in [vector_tile_source](https://github.com/maplibre/maplibre-gl-js/blob/9dcb8715d82107f415fddc0eea932a348a973b39/src/source/vector_tile_source.ts#L155), but calling `onRemove` directly does feel a little sloppy. I've updated the shape of the calls to be consistent with VectorTileSource.

HarelM (approved):

HarelM:
If you need this released please open a PR with version bump and I'll carve out a release.

dabreegster:
Apologies if I should open an issue instead, but how can this new API call be used from regular code? `map.getSource("some-raster-source").setTiles([...])` fails, because `getSource` doesn't seem to return a `RasterTileSource` object, and casting doesn't do anything.

HarelM:
Which version are you using?
Might be a typings issue in theory, but casting should work...

dabreegster:
Huh, false alarm -- sorry. I'm using 3.5.1. I just tried the same code again and it's fine. I suspect yesterday my browser cache was pulling in something old, despite force-refreshing.

PR #3225, opened 2023-10-16
Improve `Actor` typings

HarelM:
## Launch Checklist

This mainly type changes (a few minor code changes).
I needed a break from the angular/webpack/babel stuff.

@msbarry can you take a look and suggest what can be done better?
Going forward I would like every message type to have a data type associated with it so that when you send a message you know what data you send with it, but that can be done as a later step.

### DIFF
--- a/src/util/actor.ts
+++ b/src/util/actor.ts
@@ -203,6 +203,7 @@ export class Actor {
             }
         } else {
             let completed = false;
+            const options: WindowPostMessageOptions = isSafari(this.globalScope) ? undefined : {transfer: []};
             const done = task.hasCallback ? (err: Error, data?: any) => {
                 completed = true;
                 delete this.cancelCallbacks[id];
@@ -213,7 +214,6 @@ export class Actor {
                     error: err ? serialize(err) : null,
                     data: serialize(data)
                 };
-                const options: WindowPostMessageOptions = isSafari(this.globalScope) ? undefined : {transfer: []};
                 this.target.postMessage(responseMessage, options);
             } : (_) => {
                 completed = true;
--- a/src/util/actor.ts
+++ b/src/util/actor.ts
@@ -1,7 +1,8 @@
-import {isWorker, isSafari} from './util';
+import {isWorker} from './util';
 import {serialize, deserialize, Serialized} from './web_worker_transfer';
 import {ThrottledInvoker} from './throttled_invoker';
 
+import type {Transferable} from '../types/transferable';
 import type {Cancelable} from '../types/cancelable';
 import type {WorkerSource} from '../source/worker_source';
 
@@ -96,7 +97,7 @@ export class Actor {
         if (callback) {
             this.callbacks[id] = callback;
         }
-
+        const buffers: Array<Transferable> = [];
         const message: MessageData = {
             id,
             type,
@@ -104,10 +105,10 @@ export class Actor {
             targetMapId,
             mustQueue,
             sourceMapId: this.mapId,
-            data: serialize(data)
+            data: serialize(data, buffers)
         };
-        const options: WindowPostMessageOptions = isSafari(this.globalScope) ? undefined : {transfer: []};
-        this.target.postMessage(message, options);
+
+        this.target.postMessage(message, {transfer: buffers});
         return {
             cancel: () => {
                 if (callback) {
@@ -203,7 +204,7 @@ export class Actor {
             }
         } else {
             let completed = false;
-            const options: WindowPostMessageOptions = isSafari(this.globalScope) ? undefined : {transfer: []};
+            const buffers: Array<Transferable> = [];
             const done = task.hasCallback ? (err: Error, data?: any) => {
                 completed = true;
                 delete this.cancelCallbacks[id];
@@ -212,9 +213,9 @@ export class Actor {
                     type: '<response>',
                     sourceMapId: this.mapId,
                     error: err ? serialize(err) : null,
-                    data: serialize(data)
+                    data: serialize(data, buffers)
                 };
-                this.target.postMessage(responseMessage, options);
+                this.target.postMessage(responseMessage, {transfer: buffers});
             } : (_) => {
                 completed = true;
             };
--- a/src/util/actor.ts
+++ b/src/util/actor.ts
@@ -1,4 +1,4 @@
-import {isWorker} from './util';
+import {isWorker, isSafari} from './util';
 import {serialize, deserialize, Serialized} from './web_worker_transfer';
 import {ThrottledInvoker} from './throttled_invoker';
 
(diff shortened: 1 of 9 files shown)


src/util/actor.ts
```diff
@@ -69,40 +96,35 @@ export class Actor {
         if (callback) {
             this.callbacks[id] = callback;
         }
-        const buffers: Array<Transferable> = isSafari(this.globalScope) ? undefined : [];
-        this.target.postMessage({
+
```
HarelM, line 99: I have removed the isSafari part here as MDN signature is not compatible with this and I think it's not needed, but it needs to be verified with Safari I guess...

src/util/actor.ts
```diff
@@ -1,9 +1,42 @@
...
+export interface WorkerSourceProvider {
+    getWorkerSource(mapId: string, sourceType: string, sourceName: string): WorkerSource;
+}
+
+export type MessageType = '<response>' | '<cancel>' |
+'geojson.getClusterExpansionZoom' | 'geojson.getClusterChildren' | 'geojson.getClusterLeaves' | 'geojson.loadData' |
+'removeSource' | 'loadWorkerSource' | 'loadDEMTile' | 'removeDEMTile' |
+'removeTile' | 'reloadTile' | 'abortTile' | 'loadTile' | 'getTile' |
+'getGlyphs' | 'getImages' | 'setImages' |
+'syncRTLPluginState' | 'setReferrer' | 'setLayers' | 'updateLayers';
```
msbarry, line 24: I think you can also do `keyof Worker` here instead of enumerating these to get them automatically from the methods defined in `worker.ts`.

The `geojson.*` ones wouldn't come for free, but we could probably either do some fancy typescript work to create them automatically from a list of WorkerSource types, or just make explicit methods on worker `geojsonGetClusterLeaves` and `getjsonLoadData` since there's only 2 of them.

src/util/actor.ts
```diff
@@ -55,7 +82,7 @@ export class Actor {
      * @param targetMapId - A particular mapId to which to send this message.
      */
     send(
-        type: string,
+        type: MessageType,
         data: unknown,
```
msbarry, line 86: Is it in scope for this PR to make data argument typed so you know you are passing the right data into it?  Here's what I landed on in maplibre-contour:

https://github.com/onthegomap/maplibre-contour/blob/6abe4628329227337ae536c9b3eb6dc9658560c2/src/actor.ts#L107-L117

src/util/actor.ts
HarelM: I agree, but this is not the way I would like to move forward, I don't think that a message type should be a method name, it's not refactoring-safe, what I would like to do going forward is "register" an even type with a specific method, much like pub-sub and introduce promises to this entire mess.
I think this would be the first step toward reducing the number of callbacks from the code.
But I want this initial step to have more clarity on how to change things and how to proceed.

src/util/actor.ts
HarelM: I think this is OK for this PR, I'm thinking about creating a `sendAsync` method that will get a message, each message will have a type and a payload that is specific to that message, this way you know what you are passing and what you are expecting in return.
This will return a promise with a single data object that will have all the relevant data.
It will be a slow process to fix all the places, but once done, I'll simply remove this `send` method.

src/util/actor.ts
HarelM: The geojson part seems like a hack I'll be removing once this will be refactored to something that is easy to read and maintain.

HarelM:
There's also this which might reduce the need to write any of this:
https://github.com/GoogleChromeLabs/comlink

HarelM:
Doesn't look like comlink will be very helpful due to the spaghetti of a code that currently exist around the actor code.
I'll need to simplify it, refactor it and only then I might be able to use comlink, which in that point would probably not worth the effort...

birkskyum (approved):

src/util/actor.ts
msbarry: It might help to see some pseudocode of the end state you're looking for. Is it something like:

```js
// in shared code: define  a message object/type
class LoadDemTileRequest { ... }

// in worker: register handler
actor.register(LoadDemTileRequest, demWorker.loadTile);

// in main: send using shared message object/type
actor.send(new LoadDemTileRequest(...))
```

?

I think it's possible to get the existing "method of worker" approach to work with promises, but you've got a good point that we've got a chance to rethink this now so don't need to feel bound by the old implementation.

src/util/actor.ts
msbarry: That makes sense - the typing gets pretty gnarly to make that work, makes sense to set up the typing we want on the new clean-slate implementation and clean things up as we migrate to it.

msbarry:
> There's also this which might reduce the need to write any of this: https://github.com/GoogleChromeLabs/comlink

I'd also hesitate to to remove the task queue/throttled invoker in actor. I know a lot of work has went into optimizing how those work to make the map as responsive as possible with lots of messages going back and forth. It seems safe to swap-out the message types and how they are dispatched though.

src/util/actor.ts
msbarry: Wait don't we actually still want to keep this? It makes message transfers faster by listing out the objects where ownershi can be transferred to the other worker without copying the contents

https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage#transfer

src/util/actor.ts
HarelM: Yes, something similar to what you wrote, I still don't have a good picture of how this will look as I need to understand most of the places this is used and how to best define it.
For example I leaned that `Style` class is registering itself to provide `getGlyphs` method to the workers. It took me too long to figure out why and where the registration occures and how this works. 
It shouldn't be that complicated...
So I need to map the events and see how I can slowly progress and to it bit by bit...

src/util/actor.ts
HarelM: I couldn't make typescript happy with this parameter, I'll try again now, thanks for the link!

HarelM: I understand why I removed this parameter - the worker API for `postMessage` is different than the window API for `postMessage`. The actor servers both window (main thread) and the web worker.
I'm not sure what is the outcome of sending an empty array when sending a message from the main thread.
Maybe I can use the `transfer` parameter of options as it seems both are supporting it.
https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage

HarelM: This is the original issue.
https://github.com/mapbox/mapbox-gl-js/issues/8771
Seems that I removed all the transferable object by mistake, I need to revert this.
According to the above last comment the issue in safari is now fixed and it should be safe to add the transferable objects again.

HarelM:
OK, I think the code now is a lot more close to what it used to be.
I removed the check for Safari as it seems like the memory issue was resolved there.
So mainly types additions.

Issue #3209, opened 2023-10-12
Issue with Angular v16 - Uncaught ReferenceError: _asyncToGenerator is not defined


adevinwild:
**maplibre-gl-js version**:  `"maplibre-gl": "^3.4.1"`

**browser**: `Chrome`

### Steps to Trigger Behavior

 1. Create an Angular Map component that imports `Map` from `maplibre-gl`
 2. Create a new instance of the `Map` in the `ngAfterViewInit`
 3. Map shows without styles and error pops in the console  : 

### Expected Behavior
It should work ?
### Actual Behavior
Triggers this error 
 
![image](https://github.com/maplibre/maplibre-gl-js/assets/68734155/275ff020-23cb-4ac2-bad9-6eea251d919f)

HarelM:
Please provide a stackblitz reproduction.
Also I advise to use the angular wrapper - ngx-maplibre-gl.
I suspect the problem is with your code as I maintain the ngx wrapper and I know for a fact it is working great.

PR #3228, opened 2023-10-17
Remove async/await from bundle

birkskyum:
Async/await causes trouble in the bundle, so this PR compiles it down to es2016 promises.

Related to
- https://github.com/maplibre/maplibre-gl-js/discussions/3210#discussioncomment-7304396
- #3209

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -14,7 +14,7 @@
     "skipLibCheck": false,
     "sourceMap": true,
     "strict": false,
-    "target": "ES2019",
+    "target": "ES2016",
     "lib": [
       "ESNext",
       "DOM",
@@ -30,6 +30,7 @@
     "experimentalSpecifierResolution": "node",
     "esm": true,
     "compilerOptions": {
+      "target": "es2019",
       "types": [
         "node"
       ]
@@ -37,15 +38,10 @@
   },
   "include": [
     "rollup.config.*",
-    "src/**/*",
-    "test/bench/**/*",
-    "build/**/*"
+    "src/**/*"
   ],
   "exclude": [
     "node_modules",
-    "dist",
-    "build/rollup",
-    "test/bench/rollup",
-    "test/bench/**/*generated*"
+    "dist"
   ]
 }
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -38,10 +38,16 @@
   },
   "include": [
     "rollup.config.*",
-    "src/**/*"
+    "src/**/*",
+    "test/bench/**/*"
   ],
   "exclude": [
     "node_modules",
-    "dist"
+    "dist",
+    "build/rollup",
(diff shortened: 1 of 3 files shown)


birkskyum:
the test in map.test.ts seems a bit flaky. it would be great to have that figured out before merging this.

tsconfig.json
```diff
@@ -38,14 +39,15 @@
...
-    "test/bench/**/*",
-    "build/**/*"
+    "test/bench/**/*"
   ],
   "exclude": [
     "node_modules",
     "dist",
     "build/rollup",
     "test/bench/rollup",
+    "test/bench/run-benchmarks.ts",
```
HarelM, line 49: Why are these files excluded? Top level await?

tsconfig.json
```diff
@@ -38,14 +39,15 @@
   "include": [
     "rollup.config.*",
     "src/**/*",
-    "test/bench/**/*",
```
HarelM, line 41: Why was build removed?

tsconfig.json
birkskyum: yep - we run them directly with ts-node, so they don't have to be included when run `tsc`

tsconfig.json
birkskyum: Otherwise rollup will attempt to compile them to es2016, which they can't because they have top-level-await. We compile them directly with ts-node, which will set the target to es2019

tsconfig.json
HarelM: But tsc also checks for valid typescript syntax, not that these scripts change frequently, but still...

tsconfig.json
birkskyum: it's all these .ts/.js files that doesn't have to be in the include list

"benchmark": "npm run tsnode test/bench/run-benchmarks.ts",
    "gl-stats": "npm run tsnode test/bench/gl-stats.ts",
        "generate-dist-package": "npm run tsnode build/generate-dist-package.js",
    "generate-shaders": "npm run tsnode build/generate-shaders.ts",
    "generate-struct-arrays": "npm run tsnode build/generate-struct-arrays.ts",
    "generate-style-code": "npm run tsnode build/generate-style-code.ts",
    "generate-typings": "npm run tsnode build/generate-typings.ts",
    "generate-docs": "typedoc && npm run tsnode build/generate-docs.ts",
    "generate-images": "npm run tsnode build/generate-doc-images.ts",

HarelM (approved):
THANKS!!

tsconfig.json
birkskyum: Unless we add `--transpile-only` to ts-node, it should check for typescript syntax as well, so the CI would be breaking.

birkskyum:
it's ready to be merged

Issue #2498, opened 2023-05-09
Don't show cooperativeGestures messages when scrollZoom is disabled 


htrex:
When scrollZoom is programmatically enabled/disabled, cooperativeGestures should behave consequently.

When using MapLibre 2.4.0 with scrollZoom disabled and cooperativeGestures enabled the map still shows the cooperativeGestures messages when the user scrolls  over the map, but in that case CTRL+scrollwheel  triggers the web browser page zoom, not the map zoom.

HarelM:
Interesting. Feel free to open a PR to fix this issue.

PR #3240, opened 2023-10-20
#2498: Don't show cooperativeGestures messages when scrollZoom is dis…

arekgotfryd:
## Launch Checklist

 CooperativeGesture message will not get show when scrollZoom is disbaled
 - [x] Link to related issues.
 [https://github.com/maplibre/maplibre-gl-js/issues/2498](https://github.com/maplibre/maplibre-gl-js/issues/2498)


src/ui/handler/scroll_zoom.ts
```diff
@@ -149,6 +149,7 @@ export class ScrollZoomHandler implements Handler {
     disable() {
         if (!this.isEnabled()) return;
         this._enabled = false;
+        this._map._destroyCooperativeGestures();
```
HarelM, line 152: This must be symmetric - if you destroy it here you must create it in the `enable` method.

arekgotfryd: Done

src/ui/handler/scroll_zoom.ts
```diff
@@ -136,6 +136,7 @@ export class ScrollZoomHandler implements Handler {
         if (this.isEnabled()) return;
         this._enabled = true;
         this._aroundCenter = !!options && (options as AroundCenterOptions).around === 'center';
+        this._map._setupCooperativeGestures();
```
HarelM, line 139: Wouldn't this cause the backdrop to be created twice?
I'm not super fond of calling an internal map method as part of this class.
Is there another way to solve this? The fact that this logic is spread in two classes is not great, I would prefer to centralize the logic of managing the cooperative gestures state...

arekgotfryd: I moved logic related to cooperativeGestures to ScrollZoomHandler class

arekgotfryd:
Ready for review

src/ui/handler/scroll_zoom.ts
```diff
@@ -358,4 +362,45 @@ export class ScrollZoomHandler implements Handler {
             delete this._finishTimeout;
         }
     }
+
+    setupCooperativeGestures() {
```
HarelM, line 366: I don't think this logic should be here... scroll_zoom is about user interaction, not HTML building.
If you need to find a place for this logic I think a `CooperativeGestureControl` might be in place, and maybe communicate with it using events? I would try and see what other controls have implemented as they have HTML rendering part as well.

src/ui/map.test.ts
```diff
@@ -567,6 +567,7 @@ describe('Map', () => {
                 });
                 map.addSource('geojson', createStyleSource());
                 expect(map.isSourceLoaded('geojson')).toBe(false);
+                done();
```
HarelM, line 570: I'm not sure I understand why this was added here... Can you elaborate?

src/ui/map.test.ts
```diff
@@ -586,6 +587,7 @@ describe('Map', () => {
...
```
HarelM, line 590: I'm not sure I understand why this was added here... Can you elaborate?

src/ui/map.test.ts
arekgotfryd: Removed

src/ui/map.test.ts
arekgotfryd: Removed

src/ui/handler/scroll_zoom.ts
arekgotfryd: Do you mean to create additional `CooperativeGestureControl` which would be located under `src/ui/control`
directory? Is that right?

arekgotfryd: Sounds good to me. Thanks!

arekgotfryd:
Ready for review.

src/ui/control/cooperative_gesture_contol.ts
```diff
@@ -0,0 +1,111 @@
...
+        const mapCanvasContainer = this._map.getCanvasContainer();
+        const cooperativeGestures = this._map.getCooperativeGestures();
+        this._container = DOM.create('div', 'maplibregl-cooperative-gesture-screen', this._map.getContainer());
+        let desktopMessage = typeof cooperativeGestures !== 'boolean' && cooperativeGestures.windowsHelpText ? cooperativeGestures.windowsHelpText : 'Use Ctrl + scroll to zoom the map';
+        if (this._metaKey === 'metaKey') {
+            desktopMessage = typeof cooperativeGestures !== 'boolean' && cooperativeGestures.macHelpText ? cooperativeGestures.macHelpText : 'Use ⌘ + scroll to zoom the map';
+        }
+        const mobileMessage = typeof cooperativeGestures !== 'boolean' && cooperativeGestures.mobileHelpText ? cooperativeGestures.mobileHelpText : 'Use two fingers to move the map';
+        this._container.innerHTML = `
+            <div class="maplibregl-desktop-message">${desktopMessage}</div>
```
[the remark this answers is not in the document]
HarelM: Are there better ways to do this that is more safe?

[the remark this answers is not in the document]
arekgotfryd: Done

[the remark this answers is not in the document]
arekgotfryd: Done.

arekgotfryd:
Ready for review.

src/ui/control/control.ts
```diff
@@ -4,7 +4,7 @@ import type {Map} from '../map';
  * A position defintion for the control to be placed, can be in one of the corners of the map.
  * When two or more controls are places in the same location they are stacked toward the center of the map.
  */
-export type ControlPosition = 'top-left' | 'top-right' | 'bottom-left' | 'bottom-right';
+export type ControlPosition = 'top-left' | 'top-right' | 'bottom-left' | 'bottom-right' | 'full';
```
HarelM, line 7: I'm not sure how I feel about this, since there isn't another control that is full and I'm not sure what would be the expected behavior.
Can this control simply ignore this input maybe?
IDK...

src/ui/handler/scroll_zoom.ts
```diff
@@ -147,14 +150,15 @@ export class ScrollZoomHandler implements Handler {
      * ```
      */
     disable() {
+        this._map.setCooperativeGestures(false);
```
HarelM, line 153: Wouldn't this disable cooperative gesture even if it was enabled to begin with when disabling scroll zoom?

src/ui/handler/touch_pan.ts
```diff
@@ -31,27 +31,13 @@ export class TouchPanHandler implements Handler {
...
-            this._cancelCooperativeMessage = false;
-        }, 200);
     }
 
     touchstart(e: TouchEvent, points: Array<Point>, mapTouches: Array<Touch>) {
         return this._calculateTransform(e, points, mapTouches);
     }
 
     touchmove(e: TouchEvent, points: Array<Point>, mapTouches: Array<Touch>) {
-        if (this._map._cooperativeGestures) {
```
HarelM, line 46: I haven't seen this code moved to the controller, am I missing something?

src/ui/map.ts
```diff
@@ -1176,9 +1157,14 @@ export class Map extends Camera {
     setCooperativeGestures(gestureOptions?: GestureOptions | boolean | null): Map {
         this._cooperativeGestures = gestureOptions;
         if (this._cooperativeGestures) {
-            this._setupCooperativeGestures();
+            if (this._cooperativeGesturesControl) {
```
HarelM, line 1160: Is there a more elegant way instead of remove and then add - simply keep it as is, or call some enable method...?

src/ui/handler/scroll_zoom.ts
arekgotfryd: I don't follow... 
What is the point of cooperative gestures message when zooming is disabled?

src/ui/control/control.ts
arekgotfryd: The reason for 'full' position to be there is following method in map.ts
https://github.com/arekgotfryd/maplibre-gl-js/blob/3da57effa8c864fd7cad9682f5a84816eed42f33/src/ui/map.ts#L717
if I remove `getDefaultPosition` implementation from `CooperativeGestureControl` class `addControl` method will assume top-right position for it.

src/ui/handler/touch_pan.ts
arekgotfryd: I can't see that line in touch_pan.ts file.
https://github.com/arekgotfryd/maplibre-gl-js/blob/3da57effa8c864fd7cad9682f5a84816eed42f33/src/ui/handler/touch_pan.ts#L40

src/ui/map.ts
arekgotfryd: I couldn't come up with better way to handle this without changing public interface of Map class but I'm open to any suggestions.

arekgotfryd:
@HarelM I'm wondering if there is a chance to get a bounty assigned to following issue. Even the smallest one? 
The thing is I applied for developer role and I got an email with question: What ticket would you want to work on first?

src/ui/handler/scroll_zoom.ts
HarelM: You can enable and disable scroll zoom after the map is initialized, this will "forget" the initial state I think.

src/ui/handler/touch_pan.ts
HarelM: Because you removed it...
https://github.com/maplibre/maplibre-gl-js/blob/838bb63d20d547b881d2b691044ba7bb903224d3/src/ui/handler/touch_pan.ts#L45

src/ui/handler/scroll_zoom.ts
arekgotfryd: Ah yes get it now.

arekgotfryd: Ok I get it now but in that case I would need two additional methods on Map class:
enableCooperativeGestures
disableCooperativeGestures
Are you fine with that?

arekgotfryd: Having these separate methods would avoid cluttering of setCooperativeGestures method.

src/ui/control/control.ts
HarelM: While not "accurate" the default seems to be OK I think.
Otherwise we'll need to support configuring the scale control, or zoom control to be full screen, which is not what we want I believe.

HarelM:
@arekgotfryd Are you no longer interested in this feature?

arekgotfryd:
@HarelM I just don't have time at the moment to look into this. I decided to close it in order to give someone else possibility to pick it up.

PR #3269, opened 2023-10-24
More improvement to async `Actor` and replace more callbacks with promises 

HarelM:
I want to make sure that I don't mess things up so I'm keeping this PR open to see that the tests are passing.
I think I'll do most of the leftover improvements in this PR.
Let's see how big it gets.

### DIFF
--- a/src/render/glyph_manager.ts
+++ b/src/render/glyph_manager.ts
@@ -2,12 +2,12 @@ import {loadGlyphRange} from '../style/load_glyph_range';
 
 import TinySDF from '@mapbox/tiny-sdf';
 import {unicodeBlockLookup} from '../util/is_char_in_unicode_block';
-import {asyncAll} from '../util/util';
 import {AlphaImage} from '../util/image';
 
 import type {StyleGlyph} from '../style/style_glyph';
 import type {RequestManager} from '../util/request_manager';
 import type {Callback} from '../types/callback';
+import type {GetGlyphsResponse} from '../util/actor_messages';
 
 type Entry = {
     // null means we've requested the range, but the glyph wasn't included in the result.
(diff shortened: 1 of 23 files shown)


HarelM:
This is now ready for review.
I've tested as much as possible, seems like things are working as expected.
The more I dig into this code the more weird things I find, and when I convert the code from callbacks to async functions I'm finding out a lot of hacks that were made a long the years, some of them I can't solve apparently.

CC: @smellyshovel @msbarry @ChrisLoer 

This PR is against the actor async branch.
There is only one `actor.send` call left inside the ajax class, which will not part of this PR, I'm not even sure I would like to have it as part of this change in general as it might be too intrusive.

Most of the TODOs are done, there are a few left for the next days.
I hope this change will be able to be merged within a week or so.

HarelM:
OK, I figured out the problem with the return promise, now fixed.
Aborting a request to a worker will create a pending promise, not sure what should be the right solution here, but for now I'm keeping the old behavior, when all the code is converted to promises it will be a lot easier to understand when something should return and what.
But this is out of scope of this PR.
This PR is now ready for review.

neodescis (approved):
Great changes. Keep them coming, @HarelM!

src/render/glyph_manager.ts
```diff
@@ -47,117 +47,105 @@ export class GlyphManager {
...
     }
 
-    getGlyphs(glyphs: {
-        [stack: string]: Array<number>;
-    }, callback: Callback<{
-        [stack: string]: {
-            [id: number]: StyleGlyph;
-        };
-    }>) {
+    async getGlyphs(glyphs: {[stack: string]: Array<number>}): Promise<GetGlyphsResponse> {
```
wipfli, line 50: Oh finally someone is making this code readable! Thanks for working on this @HarelM.

PR #3272, opened 2023-10-25
Fix null feature properties in resolve_tokens

mattesCZ:
There is a bug in `resolve_tokens`. When the feature properties are null, the code breaks with `TypeError: Cannot use 'in' operator to search for 'name' in null`. This can easily happen with GeoJSON source, that has some feature with `properties: null`, which is valid value according to GeoJSON specification.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🐞 Bug fixes
 
+- Fix null feature properties in resolve_tokens ([#3272](https://github.com/maplibre/maplibre-gl-js/pull/3272))
 - _...Add new stuff here..._
 
 ## 3.5.2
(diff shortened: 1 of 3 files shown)


HarelM:
npm run lint -- --fix.

Can you link to the geojson spec, I think the code assumes in a lot of places that properties: null is not valid...

mattesCZ:
It's described here https://datatracker.ietf.org/doc/html/rfc7946#section-3.2 in the last bullet point. The specification is linked at https://geojson.org/

The command `npm run lint -- --fix` passes, it seems to be some issue with initialization of github checks more than failing some new tests.

HarelM (approved):

HarelM:
Yea, I tried to rerun but it's still failing, I'll try to rerun later tonight maybe.

HarelM:
Can you merge from main? It's related to node LTS change...

mattesCZ:
Merged, works much better 👌 But somehow one check failed again. It's probably not related to my code changes.

HarelM (approved):
THANKS!

PR #3274, opened 2023-10-26
Use node version from .nvmrc file in CI

HarelM:
Node LTS just changed from 18 to 20 making the CI fail.
While there's a need to fix this and update node, this will make a simple place to change the version for all the CI related flows and to developers using nvm (which they should).


louwers (approved):
Looks good!

neodescis (approved):

PR #3233, opened 2023-10-18
Async `Actor`

HarelM:
The aim of this PR is to remove a lot of the callback hell that currently resides in the code.
It does it by exposing the low communication layer i.e. ajax and actor using the promise API.
Cancellation is done using `AbortController`.
In order to run a method in a worker or in the main thread there's a need to register to a specific event.

This will introduce a lot of breaking changes so it will be mandatory to create version 4 after this is merged.

Open tasks:
- [x] Make sure API docs are correct
- [x] ~Create a 3.6 version with optional promises~ I don't think it's worth the effort...
- [x] ~Create a working example for add custom source~ too complicated for an example :-(
- [x] Ask plugin owners that exist in the examples to make sure they can use this new code
- [x] Think about other breaking changes related stuff.
- [x] Present it in the monthly meeting
- [x] Run benchmarks

Needed changes after this PR:
1. make style.ts more async await
2. addSourceType should be for the global object
3. `updateImage` should return a promise instead of this 
4. tileJson - improve `any`
5. Change addProtocol?

PR #3301, opened 2023-10-30
Remove TODOs, tidy up last things

HarelM:
Hopefully the last PR agains #3233 
It should remove the last things that were not removed in previous PRs.


HarelM:
OK, this is the last PR I think.
I've removed a lot of places that were using callbacks.
The "public" API of both ajax and actor are using promises and `AbortContoller`.
I've removed most of the TODOs I added while doing the refactoring (a few minor ones left do decide on which are related to changes of the public API).

I need to think what should be done about public methods that are using callbacks and if to create a 3.6 version that supports both callbacks and promises to ease the migration.
IDK...

CC: @smellyshovel, @msbarry 
Once the other PR is merged to main I'll create a pre-release version I think.

src/source/geojson_worker_source.ts
```diff
@@ -60,20 +57,7 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
...
-    constructor(actor: IActor, layerIndex: StyleLayerIndex, availableImages: Array<string>, loadGeoJSON?: LoadGeoJSON | null) {
-        super(actor, layerIndex, availableImages);
-        this.loadVectorData = this.loadGeoJSONTile;
-        if (loadGeoJSON) {
-            this.loadGeoJSON = loadGeoJSON;
-        }
-    }
-
-    async loadGeoJSONTile(params: WorkerTileParameters, _abortController: AbortController): Promise<LoadVectorTileResult | null> {
+    override async loadVectorTile(params: WorkerTileParameters, _abortController: AbortController): Promise<LoadVectorTileResult | null> {
```
neodescis, line 60: Nice! So much better.

src/source/vector_tile_source.test.ts
```diff
@@ -191,6 +192,64 @@ describe('VectorTileSource', () => {
         server.respond();
     });
 
+    test('loads an tile even in case of 404', done => {
```
neodescis, line 195:
```suggestion
    test('loads a tile even in case of 404', done => {
```

src/util/util.ts
```diff
@@ -642,3 +645,23 @@ export async function getImageData(
     }
     return readImageDataUsingOffscreenCanvas(image, x, y, width, height);
 }
+
+// HM TODO: remove this when done
+export function cancelableToPromise<TRequest, TResponse>(method: (requestParameters: TRequest, callback: Callback<TResponse>) => Cancelable): (requestParameters: TRequest, abortController: AbortController) => Promise<TResponse> {
```
neodescis, line 650: Why are there two copies of this?

src/source/geojson_worker_source.ts
HarelM: Yup 👍

src/util/util.ts
HarelM: Can you link to the other one?
I left this for general purpose and created one specifically for addProtocol conversation.
I'm thinking of removing both and use promises everywhere, this will force a version 4 though, but I think the current changes already forces that...

neodescis: Unless I'm mistaken, it looks like there's another copy in src/util/ajax.ts.

HarelM: Not exactly the same, but I want to keep both a bit longer to see if I need them or not.

neodescis:
Looks good to me. I'll be happy to beta test for you when it's ready!

neodescis (approved):

PR #3312, opened 2023-11-03
Fix zoom around 180 meridian (#3048)

### LINKED ISSUE #3048: Map zooms to the wrong area when double clicking
When trying to zoom in by double click, sometimes I am zoomed into a location on the other side of the world (for example, clicking australia it might zoom into africa or the altlantic ocean). This seems to only happen when 3d terrain is enabled.  I am told in slack maybe this has to do with the antimeridian line, which makes sense since it seems to occur when the map center is to the right of this line and I am clicking to the left of it.

**maplibre-gl-js version**:

**browser**:

### Steps to Trigger Behavior

 1. Go to https://stackblitz.com/edit/web-platform-whmqdc?file=index.html
 2. Double click australia to zoom in. when it zooms in it brings you to the altlantic ocean

### Link to Demonstration
Here is an example video using the [3d terrain example page](https://maplibre.org/maplibre-gl-js/docs/examples/3d-terrain/)

https://github.com/maplibre/maplibre-gl-js/assets/3792408/949b8110-0a49-406d-a10b-5daa15e4cb84

### Expected Behavior

Double clicking austrailia should zoom into austrailia 

### Actual Behavior

Double clicking austrailia zooms into the atlantic ocean

sbachinin:
Fixes #3048.

Zoom was incorrect when mouse position and map center were on the opposite sides of 180 meridian. (With terrain 3d enabled).
To solve this, mercator coordinate has to be outside [0,1] range:
a) negative if mouse was on the left of 180
b) >1 if mouse was on the right of 180
(It's how it works in no-3d mode).

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/render/terrain.test.ts
+++ b/src/render/terrain.test.ts
@@ -15,10 +15,12 @@ import {LngLat} from '../geo/lng_lat';
 
 describe('Terrain', () => {
     test('pointCoordiate should not return null', () => {
+        expect.assertions(1);
         const painter = {
             context: new Context(gl(1, 1) as any),
             width: 1,
-            height: 1
+            height: 1,
+            transform: {center: {lng: 0}}
         } as any as Painter;
         const sourceCache = {} as SourceCache;
         const getTileByID = (tileID) : Tile => {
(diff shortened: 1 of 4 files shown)


HarelM:
Can you link to the place you mentioned it already exists?
Is there some code these two places can share somehow?
What about the `wrap` method, can it be used somehow?

sbachinin:
> Can you link to the place you mentioned it already exists?

https://github.com/maplibre/maplibre-gl-js/blob/7519ac91a88224ea4f9daa32173f64a73a49821b/src/geo/transform.ts#L609
At this line **no**-3d algorithm yields a mercator coordinate that overflows (when necessary) to negative or >1 value.

> Is there some code these two places can share somehow?

I don't think so. Currently 3d and no-3d branches are totally different.
3d retreives its coordinate from coords texture. No-3d does it by transforming the vectors. I can't figure out how to mate them.

> What about the `wrap` method, can it be used somehow?

Well, I tried it and certainly `wrap` can be used but it doesn't make a more straightforward or concise code.

There might be a more elegant solution but that's what I can produce at the moment.

src/render/terrain.ts
```diff
@@ -441,4 +444,17 @@ export class Terrain {
...
     }
+
+    _allowMercatorOverflow(p: Point, mercatorX: number, width: number, centerLng: number): number {
+        const inLeftHalf = p.x < (width / 2);
+        let lng = lngFromMercatorX(mercatorX);
+        if (
+            (inLeftHalf && Math.sign(lng) > 0 && Math.sign(centerLng) < 0) ||
+            (!inLeftHalf && Math.sign(lng) < 0 && Math.sign(centerLng) > 0)
+        ) {
+            lng = 360 * Math.sign(centerLng) + lng;
```
HarelM, line 455: Is there a way to do that without passing width and centerLng parameters? Maybe somekind of normalization or "%"?

src/render/terrain.test.ts
```diff
@@ -45,7 +45,7 @@ describe('Terrain', () => {
         terrain._fboCoordsTexture.texture = imageTexture.texture;
         terrain.coordsIndex.push('abcd');
 
-        const coordinate = terrain.pointCoordinate(new Point(0, 0));
+        const coordinate = terrain.pointCoordinate(new Point(0, 0), 1, 0);
```
HarelM, line 48: Please add more test to make sure this is covering the changes you added.

HarelM:
Sorry for the delay in response, I've added a few minor comments.

src/render/terrain.ts
sbachinin: Right, turns out terrain already had this knowledge

src/render/terrain.test.ts
sbachinin: Done.

HarelM (approved):

src/render/terrain.test.ts
HarelM: It's not, the terrain object it too complicated and not well tested, unfortunately...

test/build/min.test.ts
```diff
@@ -36,7 +36,7 @@ describe('test min build', () => {
         const decreaseQuota = 4096;
 
         // feel free to update this value after you've checked that it has changed on purpose :-)
-        const expectedBytes = 761628;
+        const expectedBytes = 761698;
```
```suggestion
        const expectedBytes = 761777;
```

birkskyum:
there is a similar zoom-in-the-wrong-direction error when in high pitch and zooming with the cursor above the horizon - don't know if any lessons from this work can be utilized to fix that as well.

sbachinin:
> there is a similar zoom-in-the-wrong-direction error when in high pitch with terrain3d enabled and zooming with the cursor above the horizon - don't know if any lessons from this work can be utilized to fix that as well.

@birkskyum
I tried it and it seems to be a problem **not only** when terrain3d is enabled.
What's the expected behaviour here? Basically when you zoom in to the sky, you have to "fly upwards", that is zoom out. And it looks like a paradox. I know little about technical details but it must be really difficult to provide the intuitive behaviour in such case. Maybe this kind of scrolling should be ingnored (disabled) somehow.

HarelM:
I just tried it, in 2D it "scroll faster": meaning for the same scroll clicks it zoom in/out more.
This is only relevant for above 60 pitch, which I'm not sure how well supported this is ATM.
But I think I generally agree that disabling scrolling when the mouse or fingers are above the horizon is the right approach.

PR #3326, opened 2023-11-07
Benchmark improvements

HarelM:
Added better titles to versions,
Show graphs only when all benchmarks are completed to avoid performance degradation.
Improve code readability a bit.
This is still a hot mess of code which I hope is the last time I'm touching... it's just horrible...
Before the fix:
image

After the fix:
image

### DIFF
--- a/test/bench/components/BenchmarkRow.tsx
+++ b/test/bench/components/BenchmarkRow.tsx
@@ -15,8 +15,8 @@ export type BenchmarkRowProps = {
 export const BenchmarkRow = (props: BenchmarkRowProps) => {
     const endedCount = props.versions.filter(version => version.status === 'ended').length;
 
-    let main;
-    let current;
+    let main: Version;
+    let current: Version;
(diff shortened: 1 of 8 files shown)


test/bench/components/BenchmarkRow.tsx
```diff
@@ -96,8 +97,8 @@ export const BenchmarkRow = (props: BenchmarkRowProps) => {
                     {pInferiority && <tr><td colSpan={3}>{pInferiority}</td></tr>}
                 </tbody>
             </table>
-            {endedCount > 0 && <StatisticsPlot versions={props.versions}/>}
-            {endedCount > 0 && <RegressionPlot versions={props.versions}/>}
+            {props.finishedAll && <StatisticsPlot versions={props.versions}/>}
```
HarelM, line 100: This is the actual fix to workerTransfer delay...

neodescis (approved):

PR #3329, opened 2023-11-08
Fix Actor XSS, introduce subscribe

### LINKED ISSUE #3239: Security: `Actor` should be protected against XSS attacks in `postMessage` / `onmessage`
*This is a theoretical situation only - there's no PoC and the described attack might not work in practice on maplibre. maplibre is not less secure than many other npm packages / libs - these (targeted) attacks are unlikely, but they'll gain significance as maplibre is being used in more products.*

---

## User Story

A user opens multiple tabs.

One of them is using maplibre in a context where security should be expected:

- online-banking which has a "location finder" on a map (maplibre).
- some social network which might contain private messages + event location on a map (maplibre)
- ...

Another tab contains a hostile website which aims to extract as much data as possible from the other tab that use maplibre (including potential credentials, personal information or similar).

That data gets stolen (and potentially abused).

## Rationale

The (probably) affected code is (among other locations) here:

https://github.com/maplibre/maplibre-gl-js/blob/fbdb32dc439f41c37d1a6dc2d695f10446b72016/src/util/actor.ts#L88
https://github.com/maplibre/maplibre-gl-js/blob/fbdb32dc439f41c37d1a6dc2d695f10446b72016/src/util/actor.ts#L232

There are many XSS attacks on `postMessage` / `onMessage` which is used in the maplibre `Actor` to communicate with the web-workers.

Examples of attacks and words of caution:

- https://book.hacktricks.xyz/pentesting-web/postmessage-vulnerabilities
- https://udn.realityripple.com/docs/Web/API/Window/postMessage#Security_concerns
- https://html.spec.whatwg.org/multipage/web-messaging.html#security-postmsg

It's fairly easy to add additional security measures.
Recently, a similar PR for mapbox has taken steps to secure their `Actor` implementation.

## Impact

There are many attacks and even if there's no practical PoC, the fact that this might be exploitable should be taken seriously until proven otherwise.
With maplibre being widely used (including large companies like Meta, Microsoft and others), someone will eventually try to attack it and if there is a vulnerability, it will be found and exploited.

## Notes

On OSM Slack ( https://osmus.slack.com/archives/C01G3D28DAB/p1697724774588149 ), I proposed to generate random UUID when maplibre is loaded, which is then attached to each message from/to the webworkers, so they know it's from the map (initial communication of the UUID could happen in the worker source).
Thereby there'd be a trusted channel as no code in another window/tab could know the UUID to communicate with the worker.

I'm unable to find any thoughts on this UUID technique. Most documentation seems to only check the origin (which I consider to be flawed, especially as the origin might not be known ahead of time or there might be multiple pages on the same origin).
So, my proposal with the UUID for the trusted channel might also have flaws.

I also noticed that the `throttled_invoker` uses https://developer.mozilla.org/en-US/docs/Web/API/Channel_Messaging_API which might be intended for this case (I'm not sure). However, even the standard for that uses an untrusted `postMessage`: https://html.spec.whatwg.org/multipage/web-messaging.html#introduction-13 (which needs to be protected *somehow* to build a chain of trust).

Someone with a better understanding of `postMessage` and `onmessage` / XSS attacks and web standards in this area is probably needed to assess the situation and propose a proven method to secure maplibre.

HarelM:
Fixes #3239.

It also builds the infrastructure need in order to resolve the problems described here: https://github.com/maplibre/maplibre-gl-js/pull/2707#issuecomment-1797847191

cc: @neodescis, @clementmas, @JannikGM 

This is an addition on top of the async actor branch ATM as I assume it will be merged soon.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### 🐞 Bug fixes
 
 - Fix null feature properties in resolve_tokens ([#3272](https://github.com/maplibre/maplibre-gl-js/pull/3272))
+- Fixes a security issue in `Actor` against XSS attacks in postMessage / onmessage ([#3239](https://github.com/maplibre/maplibre-gl-js/pull/3239))
 - _...Add new stuff here..._
 
 ## 3.5.2
--- a/src/util/actor.test.ts
+++ b/src/util/actor.test.ts
@@ -172,4 +172,18 @@ describe('Actor', () => {
 
         expect(spy).not.toHaveBeenCalled();
     });
+
+    test('should not process a message with the wrong origin', async () => {
+        const worker = workerFactory() as any as WorkerGlobalScopeInterface & ActorTarget;
+        const actor = new Actor(worker, '1');
+
+        const spy = jest.fn().mockReturnValue(Promise.resolve({}));
+        worker.worker.actor.registerMessageHandler('getClusterExpansionZoom', spy);
+
+        actor.target.postMessage({type: 'getClusterExpansionZoom', data: {} as any, origin: 'https://example.com'});
+
+        await sleep(100);
+
+        expect(spy).not.toHaveBeenCalled();
+    });
 });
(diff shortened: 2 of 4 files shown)


neodescis (approved):
Nice. Like the subscribe() function!

PR #3330, opened 2023-11-09
Removes manual steps in creating a version

HarelM:
This does the following steps in the release workflow:
1. Add a required input of the version number
2. Updates the changelog and package.json (and lock) files.
3. Commits the changes to main branch
4. Continues with the regular flow of the release.

This removes the need to create a PR with a version bump which is very manual and redundant.

The changelog changes is done in a new script I added as part of this PR.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,12 @@
 name: Release
 
 on:
-  workflow_dispatch
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Version to change to.
+        required: true
+        type: string
 
 jobs:
   release:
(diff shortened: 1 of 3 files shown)


HarelM:
I hope the commit and push will work, I don't know if there's extra permissions needed.
I'll run this after this is merged to main to make sure.

wipfli:
Only people with write access can trigger this, right?

HarelM:
Right.

.github/workflows/release.yml
```diff
@@ -22,27 +27,29 @@ jobs:
...
-        id: package-version
-        uses: martinbeentjes/npm-get-version-action@v1.3.1
-
-      # - name: Check tag does not exist yet
-      #   run: if git rev-list v${{ steps.package-version.outputs.current-version }}; then echo "Tag already exists. Aborting the release process."; exit 1; fi
-
+      - name: Bump version
+        run: |
+          npm version --commit-hooks false --git-tag-version false ${{ inputs.version }}
+          ./build/bump-version-changelog.js
```
wipfli, line 33: Don't you need a git add before the commit?

wipfli (approved):
Approved in general

wipfli:
Ui did't expect the automatic merge. Let's see if it works

.github/workflows/release.yml
HarelM: No new files are added, so no, I believe.

acalcutt:
I'm curious to see if you get this working. it is similar to what I did in https://github.com/maplibre/maplibre-native/blob/main/.github/workflows/node-release.yml#L29-L48 , but in my testing the GITHUB_TOKEN didn't have enough permission and we ended up using a NODE_PRE_GYP_GITHUB_TOKEN on checkout which was a PAT with more permission.

Last time i tried it though the NODE_PRE_GYP_GITHUB_TOKEN didn't work anymore, but that might have been org permission changes.

HarelM:
Yes, I'm interested to know if this will work as well. If not, I might need to add some permissions like I gave the auto merge dependabot workflow.

The current action-bot already creates a tag in the repo when releasing, so seems like it has similar permissions, IDK...

Issue #3328, opened 2023-11-08
map.jumpTo(...) while map.easeTo(...) is in progress throws a TypeError: this._onEaseFrame is not a function


ToHold:
**maplibre-gl-js version**: 3.5.2

**browser**: Chrome, Edge

### Steps to Trigger Behavior

 1. Do an easeTo 
 2. Then do a jumpTo during the ease animation
 
### Link to Demonstration

https://codepen.io/ToHold/pen/mdvWwPg?editors=1111

### Expected Behavior

I wanted to set map pitch at 40 when the user zoom upper than level 14, then get it back to 0 if the user unzoom under 14 level. And I also wanted to move camera at a specific zoom and position.

### Actual Behavior

When the camera ease animation goes upper 14 zoom level the jumpTo is call and cause a crash because easeTo is not finished.

ToHold:
So after reading source code, building and testing, the Camera._stop method is called after the TaskQueue.run method.
And so the task.cancelled is not set to false when TaskQueue.run check if it is. So it calls task's callback that we wanted to stop. But as we delete Camera.onEaseFrame in Camera.onEaseFrame._stop, it is undefined when Camera._renderFrameCallback is called and crash the app.

TaskQueue cancelled check in TaskQueue.run :
https://github.com/maplibre/maplibre-gl-js/blob/ba72cb973840ce2482ccfba30c9f922ea937d75c/src/util/task_queue.ts#L48-L52

Camera._cancelRenderFrame call in Camera._stop :
https://github.com/maplibre/maplibre-gl-js/blob/ba72cb973840ce2482ccfba30c9f922ea937d75c/src/ui/camera.ts#L1386-L1390

TaskQueue.remove that sets task.cancelled to true :
https://github.com/maplibre/maplibre-gl-js/blob/ba72cb973840ce2482ccfba30c9f922ea937d75c/src/util/task_queue.ts#L29-L38

Camera.onEaseFrame call :
https://github.com/maplibre/maplibre-gl-js/blob/ba72cb973840ce2482ccfba30c9f922ea937d75c/src/ui/camera.ts#L1429

ToHold:
An easy way to fix, but I'm not sure if it's the best way, is to remove "delete this._onEaseFrame" from code and let the last _renderFrameCallback passed.

HarelM:
This code is complicated, especially with all the callbacks.
Having said that, isn't adding a setTimeout in your code enough to overcome this issue?

[...]

PR #3332, opened 2023-11-09
Fix TypeError: this._onEaseFrame is not a function

ToHold:
## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

In case of stopping an Camera ease animation, it appears that Camera._renderFrameCallback() of a Task can be called before the Camera._cancelRenderFrame(TaskID) of this Task and so Camera._onEaseFrame(...) is undefined and crash the app.

Fix this issue : https://github.com/maplibre/maplibre-gl-js/issues/3328


src/ui/camera.ts
```diff
@@ -1370,7 +1370,9 @@ export abstract class Camera extends Evented {
     }
 
     isEasing() {
-        return !!this._easeFrameId;
```
ToHold, line 1373: This can return true even after a Camera._stop call

src/ui/camera.ts
```diff
@@ -1370,7 +1370,9 @@ export abstract class Camera extends Evented {
...
+        // currently only prepareEase sets _moving to true
+        // and afterEase sets _moving to false
+        return this._moving;
```
HarelM, line 1375: The name of the method is now a bit confusing I think...

src/ui/camera.ts
```diff
@@ -1425,12 +1427,15 @@ export abstract class Camera extends Evented {
...
     // Callback for map._requestRenderFrame
     _renderFrameCallback = () => {
-        const t = Math.min((browser.now() - this._easeStart) / this._easeOptions.duration, 1);
-        this._onEaseFrame(this._easeOptions.easing(t));
-        if (t < 1) {
-            this._easeFrameId = this._requestRenderFrame(this._renderFrameCallback);
-        } else {
-            this.stop();
+        // avoid callback if _afterEase has been called, in case of _stop called after task callback
+        if (this._moving) {
```
HarelM, line 1431: While this might be a solution to the issue, it seems that this should be fixes elsewhere, I don't know where, but I would like renderFrameCallback to be a bit more "stupid" if possibe.
Do you think there might be a different way to solve this?

HarelM:
Please add the relevant test, either a unit test or browser test to simular this behavior.
I would also say that committing a test that fails and only then apply the fix would be an ideal demonstration that the code solves the problem...

ToHold:
Thanks to the tests, I now know more about the crash. The crash appears only when we do a jumpTo on these events : "move", "zoom", "rotate", and "pitch" depending on what easeTo have done just before. It appears because Camera._fireMoveEvents(...) calls Camera.fire("eventname"), and we are registering our jumpTo on one of these events so we are cancelling a task that is already running.

I've done the test, let me know what you think about it, it's my first time here.
And I also propose a new fix.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### 🐞 Bug fixes
 
 - Fix null feature properties in resolve_tokens ([#3272](https://github.com/maplibre/maplibre-gl-js/pull/3272))
+- Fix undefined _onEaseFrame call in Camera._renderFrameCallback() while doing Camera.jumpTo during an Camera.easeTo ([#3332](https://github.com/maplibre/maplibre-gl-js/pull/3332))
```
HarelM, line 12:
```suggestion
- Fix `undefined` `_onEaseFrame` call in `Camera._renderFrameCallback()` while doing `Camera.jumpTo` during a `Camera.easeTo` ([#3332](https://github.com/maplibre/maplibre-gl-js/pull/3332))
```

src/ui/camera.ts
```diff
@@ -1428,7 +1428,7 @@ export abstract class Camera extends Evented {
         const t = Math.min((browser.now() - this._easeStart) / this._easeOptions.duration, 1);
         this._onEaseFrame(this._easeOptions.easing(t));
         if (t < 1) {
-            this._easeFrameId = this._requestRenderFrame(this._renderFrameCallback);
+            this._easeFrameId = this._easeFrameId ? this._requestRenderFrame(this._renderFrameCallback) : this._easeFrameId;
```
HarelM, line 1431: Can this be done using `if` `else`? While I think I understand what is written here, I think an if will be a lot more readable...?

src/ui/camera.test.ts
```diff
@@ -974,6 +975,39 @@ describe('#easeTo', () => {
...
+                        requestAnimationFrame(render);
+                        camera.simulateFrame();
+                    } catch (error) {
+                        reject(error);
+                    }
+                }, 0);
+            };
+
+            camera.on('moveend', (e: Event & {done?: true}) => {
+                if (('done' in e)) {
```
HarelM, line 995: Extra `(`, `)` are not needed

src/ui/camera.test.ts
```diff
@@ -974,6 +975,39 @@ describe('#easeTo', () => {
         assertTransitionTime(done, camera, 0, 10);
         camera.easeTo({center: [100, 0], zoom: 3.2, bearing: 90, duration: 1000});
     });
+
+    test('jumpTo on("zoom") during easeTo', async () => {
+        return new Promise((resolve, reject) => {
```
HarelM, line 980: I would consider using `done` instead of creating a promise that is not awaited, or in general a test that doesn't use `await`.
consider using `once` which returns a promise and leveraging it.

HarelM:
Thanks! looks like a pretty small fix :-) I'm glad the test helped.
I've added a few comments related to readability.

src/ui/camera.test.ts
```diff
@@ -974,6 +975,39 @@ describe('#easeTo', () => {
         assertTransitionTime(done, camera, 0, 10);
         camera.easeTo({center: [100, 0], zoom: 3.2, bearing: 90, duration: 1000});
     });
+
+    test('jumpTo on("zoom") during easeTo', async () => {
```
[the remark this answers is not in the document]
ToHold: Thanks, indeed I only need two simulateFrame calls.

src/ui/camera.test.ts
ToHold: Good idea for once, I used it to register jumpTo on event. By the way I also added tests for move, pitch, and rotate events that also crash without fix.

ToHold:
All done, I hope you'll find it ok :D

src/ui/camera.test.ts
```diff
@@ -974,6 +975,86 @@ describe('#easeTo', () => {
...
+        camera.easeTo({zoom: 20, bearing: 90, pitch: 60, duration: 500}, {done: true});
+        camera.once('move', () => {
+            camera.jumpTo({pitch: 40});
+        });
+
+        camera.simulateFrame();
+        camera.simulateFrame();
+    });
+
+    test('jumpTo on("zoom") during easeTo', (done) => {
```
HarelM, line 999: Is this test has the same name as the previous one?

src/ui/camera.test.ts
```diff
@@ -974,6 +975,86 @@ describe('#easeTo', () => {
         assertTransitionTime(done, camera, 0, 10);
         camera.easeTo({center: [100, 0], zoom: 3.2, bearing: 90, duration: 1000});
     });
+
+    test('jumpTo on("move") during easeTo', (done) => {
```
HarelM, line 979:
```suggestion
    test('jumpTo on("move") during easeTo with zoom, pitch, etc', (done) => {
```

HarelM (approved):

HarelM (approved):

HarelM:
THANKS! I fixed the last issues manually.

PR #3333, opened 2023-11-09
Use commit and push GitHub action

HarelM:
This uses the commit and push action instead of writing my own code...


HarelM:
@wipfli was right, I needed to use git add before committing... this should remove the need for that...

louwers (approved):

PR #3335, opened 2023-11-09
Automate release

HarelM:
This changes the release process to do the following steps:
1. Allow creating a bump version PR from the workflows manual run menu, allowing to set the version number.
2. Checks every commit to see if the `package.json` file had changed, if so, it will publish a version to npm and github automatically.

I tried to do it in one step but there are problems with permissions etc and I didn't want to create a PAT just for this.
While it's not ideal, we'll be able to push version more easily now.


louwers (approved):
Cool solution! Hope it will work.

Issue #3106, opened 2023-09-20
fill-pattern does not display when changing patterns using "step", "zoom" 


sequel63:
Using different fill-patterns and switching them does not display any pattern at step zoom level.

"fill-pattern": [
            "step", ["zoom"],
            "zoo_11", 4,
            "volcano_11"
        ]
    }

**maplibre-gl-js version**: 2.4.0

**browser**: Chrome

### Steps to Trigger Behavior

 1. https://jsbin.com/cuxetuxexu/edit?html,output
 2. Zoom gradually to sea area.
 3. At Zoom level 4 sea fill (zoo) disappears.
 4. At Zoom level 5 new symbols (volcanoes) appear correctly.
 5. Subsequent zooming in/out does not replicate the issue, only on initial load and zoom.

### Link to Demonstration
https://jsbin.com/cuxetuxexu/edit?html,output

### Expected Behavior
Zoom level 4 should have zoo fill pattern.
### Actual Behavior
Zoom level 4 has no fill pattern.

HarelM:
Interesting... Looks likes some rounding issue probably...
Feel free to investigate and open a PR to fix this.

sequel63:
Don't have skill required to investigate unfortunately. Willing to put a moderate bounty on this to help expedite...

1ec5:
@msbarry also noticed this issue in mapbox/mapbox-gl-js#9724.

CraigglesO:
@sequel63 - are you still interested in expediting this problem? I might have time to solve this.

sequel63:
@CraigglesO yes that would be great. Not familiar with how I can put a bounty on this for you, but if you can assist it would be appreciated!

HarelM:
I would recommend connecting directly and negotiate the terms :-)

CraigglesO:
Cool. Well, I solved the bug. I'm pretty happy with the way I resolved this as well.

@sequel63 -
my discord: craiggleso
slack via OpenStreetMap: Craig OConnor
If you would like to connect and share what you're willing to pay that would be great, we can resolve this ASAP.

@HarelM - I ONLY address this bug and I do it in a way that fixes the bug but the whole area of code is actually broken and problematic. I hope it's ok that although I fix this bug and specific symptoms, that the real fundamental issue is unaddressed? I can explain more when I create a PR.

HarelM:
Yes, let's discuss this in the retirement PR.
[...]

PR #3339, opened 2023-11-10
Bug Fix: updatePatternPositions fix

CraigglesO:
Fixes https://github.com/maplibre/maplibre-gl-js/issues/3106 and most likely https://github.com/mapbox/mapbox-gl-js/issues/9724

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.
 
 
 
 ## Case Definition

Take the example from this gist:
https://gist.github.com/CraigglesO/2c9a3f741c6f3c7d276dde282aa82e6e

Zoom from starting zoom 2 to zoom 4
BEFORE:
Screenshot 2023-11-10 at 1 54 36 PM

AFTER:
Screenshot 2023-11-10 at 1 52 41 PM

## The Problem

Assume we just zoomed to zoom 4:

When data is shipped over and it's time to render, the fill will check the [current pattern property](https://github.com/maplibre/maplibre-gl-js/blob/bf6d06d3648b41597505ef58d2402f3ca87cef04/src/render/draw_fill.ts#L82). Here is the console log:

Screenshot 2023-11-10 at 1 57 35 PM

This is WRONG. At zoom 4, it should be volcano_11 for both cases.

Meawhile the reason this is actually an issue is that the [WorkerTile parsing the laye](https://github.com/maplibre/maplibre-gl-js/blob/bf6d06d3648b41597505ef58d2402f3ca87cef04/src/source/worker_tile.ts#L67)r information at the [bucket population](https://github.com/maplibre/maplibre-gl-js/blob/bf6d06d3648b41597505ef58d2402f3ca87cef04/src/source/worker_tile.ts#L130) checks for [patterns](https://github.com/maplibre/maplibre-gl-js/blob/bf6d06d3648b41597505ef58d2402f3ca87cef04/src/data/bucket/fill_bucket.ts#L77) and when it finally runs [addPatternDependencies](https://github.com/maplibre/maplibre-gl-js/blob/bf6d06d3648b41597505ef58d2402f3ca87cef04/src/data/bucket/pattern_bucket_features.ts#L34) we see this:
Screenshot 2023-11-10 at 1 57 46 PM

The reason is because the WorkerTile constantly [updates the style/layer states](https://github.com/maplibre/maplibre-gl-js/blob/bf6d06d3648b41597505ef58d2402f3ca87cef04/src/source/worker_tile.ts#L246) because it's always running [maybePrepare](https://github.com/maplibre/maplibre-gl-js/blob/bf6d06d3648b41597505ef58d2402f3ca87cef04/src/source/worker_tile.ts#L200) before studyng the layers to ship back to the main thread.

The easiest fix (albeit a bandaid to a larger problem) is to edit updatePatternPositionsInProgram to render SOMETHING (basically just avoid complete failure). Since image dimensions are shared on demand for each tile, the WorkerTile never shared anything about zoo, so when it can't find any zoo data it completely fails. Whereas now, the function will find a way to succeed even if it can't find the data.

Perhaps other sections of the codebase actually correct for this behavior but the BEST solution is to find a way to force incremental updates from the styler more frequently. This would fix any other layer's having similar properties from not updating properly as well, but after a decent effort into digging I couldn't figure out the best way to update layers more frequently without potentially breaking some things.


src/render/update_pattern_positions_in_program.ts
```diff
@@ -33,9 +33,16 @@ export function updatePatternPositionsInProgram(
     let posTo = patternPositions[constantPattern.to.toString()];
     let posFrom = patternPositions[constantPattern.from.toString()];
 
+    // BUG: Sometimes the layer properties are not properly updated to the zoom on the client side unlike the TileWorker.
```
HarelM, line 36: Reading the comments on this method, don't they describe the issue you are describing here in this PR?

src/render/update_pattern_positions_in_program.ts
```diff
@@ -33,9 +33,16 @@ export function updatePatternPositionsInProgram(
...
+    // This is a workaround to ignore patterns that aren't relavent/accessable to the current zoom but upon fixing
+    // the root cause this should be removed.
+    if (!posTo && posFrom) posTo = posFrom;
+    if (!posFrom && posTo) posFrom = posTo;
+
     // try again in case patternPositions has been updated by worker
     if (!posTo || !posFrom) {
         const transitioned = layer.getPaintProperty(propertyName) as string;
+        // BUG: key could be an array like ['step', Array(1), 'zoo_11', 4, 'volcano_11'] its not guaranteed to be a string
```
HarelM, line 45: I don't understand what this comment means, what should I do with this information if I read it a year from now?

src/render/update_pattern_positions_in_program.test.ts
```diff
@@ -0,0 +1,70 @@
...
+import type {FillLayerSpecification, ResolvedImage} from '@maplibre/maplibre-gl-style-spec';
+import type {ProgramConfiguration} from '../data/program_configuration';
+import type {ImagePosition} from './image_atlas';
+
+describe('updatePatternPositionsInProgram', () => {
+
+    test('geojson tile', () => {
+        const config = constructMockProgramConfiguration();
+        const tile = new Tile(new OverscaledTileID(3, 0, 2, 1, 2), undefined);
+        // @ts-expect-error we only need the image atlas
```
HarelM, line 15: Can we use `as any` instead?

src/render/update_pattern_positions_in_program.test.ts
```diff
@@ -0,0 +1,70 @@
...
+
+        // @ts-expect-error we added this property to just see what the update looks like
+        expect(config.patternPositions).toEqual({
+            posFrom: {x: 0, y: 0, w: 0, h: 0},
+            posTo: {x: 0, y: 0, w: 0, h: 0}
+        });
+    });
+});
+
+function constructMockProgramConfiguration(): ProgramConfiguration {
```
HarelM, line 40: In most tests, helper methods are at the top of the file, before the describe.
I think it's a good convention...

HarelM: Also, if possible, use jest mocks more.

src/render/update_pattern_positions_in_program.ts
CraigglesO: Yes, my thought process was if someone comes across this code in the future, and the fundamental issue is addressed, they would know its safe to remove this. Should I create a bug report of the fundamental issue and link it instead with a `NOTE` rather then `BUG`?

src/render/update_pattern_positions_in_program.ts
CraigglesO: If the goal is to eventually have proper typescript support (turn `strict:true` on), transitioned is not necessarily a key and the method of "fetching" `from` and `to` should be improved upon. I don't see that as in scope for this bug fix but if you think I should remove the comment or write something different let me know.

HarelM: I tried to turn strict on, it's a lost cause...
If the types are incorrect, either correct them as I often do when seeing that they are wrong or ignore it, this comment will not solve it I believe.

src/render/update_pattern_positions_in_program.ts
HarelM: Yes, opening an issue would be a better way to do it, after opening the issue you can link to it from the code, this way the comment is a lot shorter...

CraigglesO:
Hey, I think I addressed all your issues. I rarely use Jest, so I'm assuming this is correct? Let me know if I missed anything.

src/render/update_pattern_positions_in_program.ts
```diff
@@ -33,6 +33,10 @@ export function updatePatternPositionsInProgram(
     let posTo = patternPositions[constantPattern.to.toString()];
     let posFrom = patternPositions[constantPattern.from.toString()];
 
+    // https://github.com/maplibre/maplibre-gl-js/issues/3377
+    if (!posTo && posFrom) posTo = posFrom;
```
HarelM, line 37: Shouldn't this part be below the next part where there is an attempt to populate posTo and posFrom?
I don't know what this code is doing, but reading it a bit this was my initial hunch...

CraigglesO: That was my initial gut reaction too. But if I do it after it actually fails.
The case below doesn't work except if it's a static styled asset. So for example:
```json
{
        "id": "patternbug",
        "type": "fill",
        "source": "openmaptiles",
        "source-layer": "water",
        "layout": {},
        "paint": {
          "fill-pattern": [
            "step",
            ["zoom"],
            "zoo_11",
            4,
            "volcano_11"
          ]
        }
      }
```
The code you're referring to specifically the line `const transitioned = layer.getPaintProperty(propertyName) as string;`: `transitioned` becomes: `['step', ['zoom'], 'zoo_11', 4, 'volcano_11']`, but the code expects a string something like `"zoo_11"`. In other words, the added code and the code below are trying to solve different problems and if it goes after the properties become broken before I can fix it.

Also, I'm fairly certain the code is not only buggy but  doesn't work or add value even if it was a string `fill-pattern` property, but I'm too scared to touch it and create new bugs. It's why I originally added a bug comment but it's probably better to just leave it be.

HarelM: Ok, thanks for the detailed explanation!

HarelM (approved):

HarelM:
Forget to ask for a changelog entry before I merge.

CraigglesO:
Sure. I'll add as soon as I get the bounty so you know you can merge.

HarelM:
Ok, ping me.
I want to start working on a pre release of version 4, so if this needs to be included in version 3.x this needs to be done soon.

HarelM (approved):

PR #3341, opened 2023-11-10
Add registry url to allow publishing npm package

HarelM:
Hopefully this would solve the issue.
If now, I can try and revert to the previous definition of the node version, it might help, IDK...

Based on this:
https://github.com/npm/cli/issues/6184#issuecomment-1544496444


HarelM:
I think this is it:
Failed release:
image
Successful release:
image

Take away message: Always make sure you have good defaults... :-/

louwers (approved):

PR #3365, opened 2023-11-14
Allow release to run on every commit to main

HarelM:
This allows the release to run on main every time there's a push.
It will check if the current version is the same, if not, it will publish a release to npm and all the other release stuff.
I've tested this to work on ngx-maplibre-gl, so I'm confidant that this is possible now.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,11 +1,39 @@
 name: Release
 
 on:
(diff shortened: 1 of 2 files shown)


louwers (approved):

HarelM:
Yay! This is working!! Thanks @louwers!

PR #3369, opened 2023-11-15
Remove symbolic links and just copy test assets

zhangyiatmicrosoft:
## Launch Checklist

 On Windows machine if environment is not created with symbolic link, two render tests will fail and it is time consuming to debug. This PR removes the symbolic links and just copy the test assets instead


HarelM:
Are you sure this is the problem?
The CI passes on windows and I haven't seen this flag there...?

zhangyiatmicrosoft:
> Are you sure this is the problem? The CI passes on windows and I haven't seen this flag there...?

Not sure how CI machines were setup, maybe they have a special flag?
All win machines I saw and all co-workers who had issues running render tests, this was the issue. When sym links were not there, these "json" files have a one-line text which is the name of the link target

HarelM:
Can't we simply remove this symlinks? This extra step and test seems like a patch, not a solution to the root cause...

zhangyiatmicrosoft:
> Can't we simply remove this symlinks? This extra step and test seems like a patch, not a solution to the root cause...

These tests were setup this way since Mapbox; I am guessing the original idea was for several tests to share the same assets and no need to worry about sync them. We could just copy them instead of using sym link. That would a solution as well ..

HarelM:
Yup, also using wsl is a faster option when working on windows...

zhangyiatmicrosoft:
> Yup, also using wsl is a faster option when working on windows...

done.

zhangyiatmicrosoft:
> Yup, also using wsl is a faster option when working on windows...

Oh I guess that is why:
Error: ENAMETOOLONG: name too long, stat '/home/runner/work/maplibre-gl-js/maplibre-gl-js/test/integration/assets/sprites/icon-text-fit-1x@2x.json'

Shall we go back to my original fix?

zhangyiatmicrosoft:
> > Yup, also using wsl is a faster option when working on windows...
> 
> Oh I guess that is why: Error: ENAMETOOLONG: name too long, stat '/home/runner/work/maplibre-gl-js/maplibre-gl-js/test/integration/assets/sprites/[icon-text-fit-1x@2x.json](mailto:icon-text-fit-1x@2x.json)'
> 
> Shall we go back to my original fix?

@HarelM any suggestions? The name above does not look very long at all, only 138 char.
I am deeply surprised

HarelM:
This is strange.
I think the OS tries to use the content of the json file as a link...?
I would try and remove the file and add it regularly, maybe there are some flags somewhere...

HarelM:
The part that runs the lint is on a linux system where the 256 path limitation does not exist, so it has to be some flags on the file itself...

HarelM:
This is what I get on my mac:
```
harelmazor@Harels-MBP maplibre-gl-js % ls -lt test/integration/assets/sprites/icon-text-fit-1x@2x.json
lrwxr-xr-x  1 harelmazor  staff  18 Jul  7 23:43 test/integration/assets/sprites/icon-text-fit-1x@2x.json -> icon-text-fit.json
```

So this must be configured somewhere and checked in, much like when you `chmod` a file I believe...

zhangyiatmicrosoft:
Should be good to go now @HarelM

HarelM (approved):

PR #3371, opened 2023-11-15
Async actor remove cancelables

HarelM:
Removes more places that use cancelable in favor of abort controller.
Changes getVideo to return a promise instead of callback/cancelable.
Improve usage in async in tests and added a spy.


neodescis (approved):
LGTM!

PR #3372, opened 2023-11-16
Remove callback from render tests

HarelM:

This simply removes the callback in the apply operation callback, remove the recursion there, and uses more await to simplify the code.


lseelenbinder (approved):
LGTM

Issue #3052, opened 2023-08-31
Breaking changes MapLibre GL JS v4


wipfli:
Let's start tracking breaking changes for MapLibre GL JS v4 in the thread below.

wipfli:
I would like to suggest that we start showing a Leaflet-style "MapLibre" in source attribution by default starting with MapLibre GL JS v4.

acalcutt:
@wipfli I was thinking as an alternate to that is we could enable the maplibre logo by default. in the past we had discussed possibly doing that.

Though adding to attribution instead migjt be less disruptive.

HarelM:
I'm working on the code to reduce as much as possible the callback hell that currently exists.
As part of that, I think there will be a need to change the `GeoJSONSource` API so that it will return a promise instead of a callback.
Currently all the methods are chainable (returning `this`) so I can't even ,as an initial step, return a `Promise` in case a callback is not supplied, which is what I had done with the `once`.
Generally speaking, it might be a good idea to remove as many of the callbacks as possible from the public API as part of this breaking change version.

HarelM:
The relevant PR of most changes can be found here, it's mostly complete, there are a few things I want to improve in terms of docs, examples, etc, but nothing major ATM:
- #3233

After digging into this deeply as I have, I'm thinking that there might be a way to improve the `addProtocol` method.
The current problem with the implementation is that it "ping-pongs" from worker thread to main thread in order to execute the add protocol method.
While this might be OK, I think I have an idea how to extend this capability to allow running things in the worker thread code and avoid this extra "ping-pong".

CC: @abelvm, @msbarry, @smellyshovel, @bdon.

This is also an opportunity to add some other breaking changes, I'll look into issues and PRs to see if I find a mentioning of things that waited for a breaking changes version and add them here.

bdon:
Can we take advantage of breaking changes to clean up this? https://github.com/maplibre/maplibre-style-spec/issues/311

HarelM:
I think it's a good idea. Can you push this forward?

wipfli:
Just to clarify: a breaking version of MapLibre GL JS means that some public APIs have breaking changes, but the style spec should not have breaking changes. So https://github.com/maplibre/maplibre-style-spec/issues/311 can come also with a minor release of GL JS.

HarelM:
This is a default value change basically, so it's also considered a breaking change, just like the attribution in this thread.

HarelM:
[...]

PR #3374, opened 2023-11-16
Async actor callback and promise

HarelM:
This PR is replacing callbacks with promises in the sources to allow easier maintenance.
Since source cache is using the promise versions now, there's no point in keeping the old callback API.

It also reverts the change from method to properties in the sources which caused an unwanted breaking change.
Also adds docs and type improvements.

Any input would be valuable.
Relevant issue: #3052


src/source/canvas_source.ts
```diff
@@ -107,7 +107,7 @@ export class CanvasSource extends ImageSource {
         this.animate = options.animate !== undefined ? options.animate : true;
     }
 
-    load = () => {
+    async load() {
```
neodescis, line 110: I don't see anything in this function that necessitates the async?

neodescis: Ah, yep, that's it. ImageSource's load() is async. Should we just make all of the load() functions async for consistency then?

src/util/actor.test.ts
```diff
@@ -64,7 +64,7 @@ describe('Actor', () => {
         expect(received).toBeFalsy();
     });
 
-    test('cancel a request of a cancelable registered callback will cancel it', async () => {
+    test('abort a request of will abort it', async () => {
```
neodescis, line 67:
```suggestion
    test('aborting a request will successfully abort it', async () => {
```

src/source/canvas_source.ts
HarelM: I think I did, I'll check.

HarelM: Added in the last one. 
It's worth remembering that `load` is not part of the interface or the public API, but more of a common method that is used in all these classes. But, I like consistency, so why not? :-)

PR #3375, opened 2023-11-16
Fix the canvas-source example

Caerbannog:
The following example looks broken, at least on Firefox. This PR fixes it.
https://maplibre.org/maplibre-gl-js/docs/examples/canvas-source/

Before
image

After
image

### DIFF
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -103,7 +103,7 @@ export type MapOptions = {
      */
     attributionControl?: boolean;
     /**
-     * Attribuition text to show in an {@link AttributionControl}. Only applicable if `options.attributionControl` is `true`.
+     * Attribution text to show in an {@link AttributionControl}. Only applicable if `options.attributionControl` is `true`.
      */
     customAttribution?: string | Array<string>;
     /**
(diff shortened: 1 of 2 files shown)


test/examples/canvas-source.html
```diff
@@ -19,6 +19,7 @@
     //Animation from https://javascript.tutorials24x7.com/blog/how-to-draw-animated-circles-in-html5-canvas
     const canvas = document.getElementById('canvasID');
     const ctx = canvas.getContext('2d');
+    canvas.style.display = 'none';
```
HarelM, line 22: Can you elaborate on why this fixes the issue?

src/ui/map.ts
```diff
@@ -103,7 +103,7 @@ export type MapOptions = {
      */
     attributionControl?: boolean;
     /**
-     * Attribuition text to show in an {@link AttributionControl}. Only applicable if `options.attributionControl` is `true`.
+     * Attribution text to show in an {@link AttributionControl}. Only applicable if `options.attributionControl` is `true`.
```
HarelM, line 106: THANKS! :-)

HarelM (approved):

test/examples/canvas-source.html
Caerbannog: We declare a 400x400 (mostly blank) canvas at line 16. It always sits above the map.
But it should not be visible, it is only used as a source for the map.
The canvas used to take the space reserved for the map, and push it out of sight.

Issue #2601, opened 2023-05-26
Mapbox-gl-draw no longer works in maplibre v3.0.0


erikvullings:
**maplibre-gl-js version**: 3.0.0

**browser**: Edge

### Steps to Trigger Behavior

 1. Go to your own [example](https://maplibre.org/maplibre-gl-js-docs/example/mapbox-gl-draw/)
 2. Note that the drawing tools, which are shown in the example gallery, are no longer there (probably, they have become very small in the top-right).
 3. Even when visible, the draw controls no longer react on mouse click events.

### Link to Demonstration

You can use your own [jsfiddle](https://jsfiddle.net/api/post/library/pure) to have a demo.

### Expected Behavior

I can draw on the map.

### Actual Behavior

Nothing happens.

neodescis:
I believe this is because MapLibre no longer has the Mapbox CSS classes applied. Mapbox Draw depends on them being there, and not just for CSS. There are checks in the Mapbox Draw code for them as well.

neodescis:
I've been able to work around the issue by adding the Mapbox classes back. Here's what I did.

First, add the Mapbox CSS classes to the MapLibre map:
```
// MapboxDraw requires the canvas's class order to have the class 
// "mapboxgl-canvas" first in the list for the key bindings to work
map.getCanvas().className = 'mapboxgl-canvas maplibregl-canvas';
map.getContainer().classList.add('mapboxgl-map');
const canvasContainer = map.getCanvasContainer();
canvasContainer.classList.add('mapboxgl-canvas-container');
if (canvasContainer.classList.contains('maplibregl-interactive')) {
  canvasContainer.classList.add('mapboxgl-interactive');
}
```

Also, when the draw controls are created, the MapLibre classes need to be added to what MapboxDraw creates:
```
const originalOnAdd = draw.onAdd.bind(draw);
draw.onAdd = (map) => {
  const controlContainer = originalOnAdd(map);
  controlContainer.classList.add('maplibregl-ctrl', 'maplibregl-ctrl-group');
  return controlContainer;
};
```

HarelM:
I'll close this issue as this is not a msplibre-gl bug. 
You can use the css from before version 3 asv it has both css classes, it might also have some stuff missing.

neodescis:
So... by closing this, are you saying that MapLibre no longer wants to support MapboxDraw at all? That seems strange to me, as does leaving a broken demo/example.

HarelM:
[...]

PR #3394, opened 2023-11-20
Fix the example as per the discussion in #2601 issue

dBitech:
Fixes the example as per the discussion in #2601


HarelM:
THANKS!!! Can you add a changelog entry for this?

HarelM (approved):

Issue #333, opened 2021-09-12
Make all `_` fields private


HarelM:
## Motivation

The motivation is to properly define all methods and members of classes with their appropriate modifier (public, private mainly).
This is a big challenge that requires a lot of work.
Having said that, doing it for `_` prefixed fields and methods should be relatively simple and an easy win.

rbrundritt:
I wonder if this would break any workarounds people have implemented in production.

HarelM:
This is what 2.0 is for...

drwestco:
Unless the project is using a minifier that mangles private class member names (not common practice), marking members as private shouldn't affect such debug console hacks.

douglasg14b:
This would break tons of workaround that people use.... I even have some of these kicking around.

douglasg14b:
Oh boy, stale bot.

HarelM:
Unfortunately, due to the large echo system around maplibre I think this won't help anyone but only cause pain.
Having said that, I urge people who are using these private fields to open issues with API request changes/improvements so that they will be able to use fields that they need without the fear of things breaking...

PR #3418, opened 2023-11-28
Refactor RTL plugin loading code

HarelM:
## Launch Checklist

This is a refactoring of the RTL plugin loading code.
It splits the code into what is done in the worker vs what is done in the main thread (which was a complicated single class).
Moves all the logic into the class, using extending `evented` in the main thread no notify when the plugin is loaded or the state changes.

This also removes all the callback hell that was in this class which is not a lot more readable.

It changes the `setRTLTextPlugin` public API and removed the callback from it.


HarelM:
Main, these render tests are a pain... :-S

src/source/rtl_text_plugin_main_thread.ts
```diff
@@ -0,0 +1,70 @@
...
+import {Dispatcher} from '../util/dispatcher';
+import {getGlobalWorkerPool} from '../util/global_worker_pool';
+
+class RTLMainThreadPlugin extends Evented {
+    pluginStatus: RTLPlginStatus = 'unavailable';
+    pluginURL: string = null;
+    dispatcher: Dispatcher = new Dispatcher(getGlobalWorkerPool(), 'rtl-text-plugin-dispacher');
+    queue: PluginState[] = [];
+
+    async _sendPluginStateToWorker() {
```
neodescis, line 14: Should this be private?

src/source/rtl_text_plugin_main_thread.ts
```diff
@@ -0,0 +1,70 @@
...
+        this.pluginURL = browser.resolveURL(url);
+        this.pluginStatus = 'deferred';
+        await this._sendPluginStateToWorker();
+        if (!deferred) {
+            //Start downloading the plugin immediately if not intending to lazy-load
+            await this._downloadRTLTextPlugin();
+        }
+    }
+
+    async _downloadRTLTextPlugin() {
```
neodescis, line 41: Same here... private?

src/source/rtl_text_plugin_status.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * `deferred`: The plugin URL has been specified, but loading has been deferred.
+ *
+ * `loading`: request in-flight.
+ *
+ * `loaded`: The plugin is now loaded
+ *
+ *  `error`: The plugin failed to load
+ */
+export type RTLPlginStatus = 'unavailable' | 'deferred' | 'loading' | 'loaded' | 'error';
```
neodescis, line 14:
```suggestion
export type RTLPluginStatus = 'unavailable' | 'deferred' | 'loading' | 'loaded' | 'error';
```

src/source/rtl_text_plugin_main_thread.ts
HarelM: I'm not sure if I should start adding `private`... while I generally agree that this is the right approach, it should probably be a wider usage of these modifiers, so I'm not sure when will be a good time to introduce this...

src/style/style.ts
```diff
@@ -320,6 +288,18 @@ export class Style extends Evented {
         });
     }
 
+    _rtlTextPluginStateChange = () => {
```
neodescis, line 291: Private?

neodescis:
This is soooo much better. Thank you for doing this!

src/source/rtl_text_plugin_main_thread.ts
neodescis: Well, the way I look at it is anything that isn't `private` is adding to the API, even if it has an underscore.

neodescis: If it makes sense, especially since we're gearing up for 4.x, I'd be happy to go through map.ts and style.ts to mark stuff as private.

HarelM: I agree, but there are plugins that uses these underscore methods and the code itself which uses those, so this is more of a convention than an actual way to block usage... IDK...
Also see here: 
- #333

HarelM:
Yup, it's a lot clearer now which part runs where and what it actually do. Took me a few hours to reverse engineer this... :-/

src/source/rtl_text_plugin_main_thread.ts
HarelM: Was writing while you were, so my message is answering you first and not second.
I would advise to get some feedback from the community around this before doing so - changing stuff to private, using our slack channel.
While I generally agree this is the right solution, I'm not sure what it would solve on one hand, but I know it will cause some pain on the other - so the value here is low I believe.

HarelM:
@neodescis can we merge this and decide later on the modifier stuff?

src/source/rtl_text_plugin_main_thread.ts
neodescis: OK, if you don't think it's worth the pain, I'll stop harping, haha

neodescis:
> @neodescis can we merge this and decide later on the modifier stuff?

Certainly!

neodescis (approved):

PR #3420, opened 2023-11-28
Move `addSourceType` to global object

HarelM:
This moves the `addSourceType` to be a member of the global object.
This method is not per map as it add code to all the workers which are shared between all maps in the current website.
This method is similar to addProtocl in a lot of aspects and should be part of the global object.

I'm still not sure if I want to add an example to the examples pages for this method.


src/index.ts
```diff
@@ -231,6 +231,15 @@ class MapLibreGL {
     static removeProtocol(customProtocol: string) {
         delete config.REGISTERED_PROTOCOLS[customProtocol];
     }
+
```
neodescis, line 234: Should we expose getSourceType here as well?

HarelM: sounds right, but what would you use if for?

neodescis: One thing that comes to mind is allowing the source type to be changed at runtime. Allowing retrieval of the type would allow for that, without having to keep track of it outside of MapLibre. Of course, I realize changing some of the things about a source type at runtime would be dangerous...

HarelM: If anything, remove would be the opposite of add. but I'm not sure how to "remove" code that was added to the worker.
And yes, changing a source is super risky...
I'm not sure either way... I think this API isn't really in use as it is so complicated to get it right...
In general you can mess with the prototype of the source object without the need for this method basically.

neodescis: Yes, understood. I'm also perfectly fine if we leave it as-is for now. Just food for thought.

neodescis (approved):

PR #3423, opened 2023-11-28
Improve `loadTileJson` types

HarelM:
This improves the loadTileJson method types according to what's really is returned.

### DIFF
--- a/src/source/load_tilejson.ts
+++ b/src/source/load_tilejson.ts
@@ -12,7 +12,7 @@ export async function loadTileJson(
     requestManager: RequestManager,
     abortController: AbortController,
 ): Promise<TileJSON> {
-    let tileJSON: any = options;
+    let tileJSON: TileJSON | typeof options = options;
     if (options.url) {
         const response = await getJSON<TileJSON>(requestManager.transformRequest(options.url, ResourceType.Source), abortController);
         tileJSON = response.data;
@@ -26,9 +26,8 @@ export async function loadTileJson(
             ['tiles', 'minzoom', 'maxzoom', 'attribution', 'bounds', 'scheme', 'tileSize', 'encoding']
         );
 
-        if (tileJSON.vector_layers) {
-            result.vectorLayers = tileJSON.vector_layers;
-            result.vectorLayerIds = result.vectorLayers.map((layer) => { return layer.id; });
+        if ('vector_layers' in tileJSON) {
+            result.vectorLayerIds = tileJSON.vector_layers.map((layer) => { return layer.id; });
         }
 
         return result;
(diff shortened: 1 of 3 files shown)


src/source/load_tilejson.ts
```diff
@@ -7,28 +7,39 @@ import type {RequestManager} from '../util/request_manager';
 import type {TileJSON} from '../types/tilejson';
 import type {RasterDEMSourceSpecification, RasterSourceSpecification, VectorSourceSpecification} from '@maplibre/maplibre-gl-style-spec';
 
+export type LoadTileJsonResponse = {
+    tiles: Array<string>;
+    minzoom: number;
+    maxzoom: number;
+    attribution: string;
+    bounds: RasterDEMSourceSpecification['bounds'];
+    scheme: VectorSourceSpecification['scheme'];
```
neodescis, line 16: Why is the type for bounds coming from RasterDEM, while scheme is coming from Vector? I realize they're the same type, but this reads weird.

src/source/load_tilejson.ts
```diff
@@ -7,28 +7,39 @@ import type {RequestManager} from '../util/request_manager';
...
+        const result: LoadTileJsonResponse = pick(
             // explicit source options take precedence over TileJSON
             extend(tileJSON, options),
             ['tiles', 'minzoom', 'maxzoom', 'attribution', 'bounds', 'scheme', 'tileSize', 'encoding']
         );
 
-        if (tileJSON.vector_layers) {
-            result.vectorLayers = tileJSON.vector_layers;
-            result.vectorLayerIds = result.vectorLayers.map((layer) => { return layer.id; });
+        if ('vector_layers' in tileJSON) {
```
neodescis, line 41: vector_layers could still be null/undefined, even if the property exists in tileJSON. Might want to check that too.

HarelM: Good point.

src/source/load_tilejson.ts
HarelM: It's a superset, so I'll need to take something from both in any case...

HarelM: Another option is the following
```
    bounds: RasterSourceSpecification['bounds'];
    scheme: RasterSourceSpecification['scheme'];
    tileSize: number;
    encoding: RasterDEMSourceSpecification['encoding'];
```
But it's not fundamentally different...
There is no one type of source specification that has all these definitions...

src/source/load_tilejson.ts
HarelM: Done

src/source/load_tilejson.ts
neodescis: OK, understood. I would say go with whatever you think is easiest on the eyes.

neodescis (approved):

Issue #160, opened 2021-05-28
Suppress error messages for missing raster tiles


mactrem:
Returning http status code 204 (no content) with no body for missing tiles is used in many mapserver implementations like TileServer GL (https://github.com/maptiler/tileserver-gl/issues/46) and also proposed by the OGC API Tiles (https://github.com/opengeospatial/ogcapi-tiles/blob/master/core/standard/requirements/core/REQ_tc-success.adoc).    
This is particularly the case for raster tile sources  where heterogenous maximum zoom levels are common.   
MapLibre doesn't explicitly handle the 204 status for raster tiles which leads to a large number of error messages in the dev console.   

### Expected Behavior
No errors are thrown when a mapserver returns 204 for a missing raster tile and therefore no error messages in the dev console

### Actual Behavior
![image](https://user-images.githubusercontent.com/75912142/119965946-19e9d600-bfab-11eb-98a9-da84b275d39e.png)

frank-unovica:
+1

HarelM:
The following PR might be relevant to this issue:
- #8335

PR #3428, opened 2023-11-30
Handle loading of empty raster tiles (204 No Content)

petrsloup:
This PR tries to solve the issue of multiple errors in case of loading raster tiles from server responding 204s with no content.

Related to #160 and #1586, but a different approach.

There are currently two ways of loading the image data (ArrayBuffer):
- `arrayBufferToImage` ("legacy" way which uses `Image` under the hood) 
- the newer `arrayBufferToImageBitmap` (which requires [createImageBitmap](https://developer.mozilla.org/en-US/docs/Web/API/createImageBitmap)).

The legacy path has [always](https://github.com/maplibre/maplibre-gl-js/commit/f183b416b152868c2f787dbfa78f173fccfadb78#diff-bb4eb119021870bc4f2e7fcd6b535235bad4b6316c2d23ff218b5f5f2614fbabR67) been able to handle 204s correctly by load a transparent image instead: https://github.com/maplibre/maplibre-gl-js/blob/859b1a7c1d88a111031f3938f1c329604120f1e0/src/util/util.ts#L494

There is no such check for the `ImageBitmap` version and it just fails with an error (it's not as easy to mock the transparent image here).

This PR adds a simple check so that empty ArrayBuffers (= 204 no content responses) always use the legacy path.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### 🐞 Bug fixes
 - Fix zooming outside the central globe when terrain 3D is enabled ([#3425](https://github.com/maplibre/maplibre-gl-js/pull/3425))
+- Handle loading of empty raster tiles (204 No Content) ([#3428](https://github.com/maplibre/maplibre-gl-js/pull/3428))
 - _...Add new stuff here..._
 
 ## 4.0.0-pre.1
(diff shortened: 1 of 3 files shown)


HarelM:
cc: @ibesora can you take a look at this?
The following PR also does some work in this area, please let me know how relevant it will be after this PR is merged.
- #2325 
If it is not relevant anymore, I'll close this other one.
Overall, making the behavior of createImage and the other flow identical is a good solution from my perspective.
THANKS!

Please add a changelog entry.

HarelM (approved):

PR #3430, opened 2023-11-30
Cooperative gestures redesign

### LINKED ISSUE #2498: Don't show cooperativeGestures messages when scrollZoom is disabled

HarelM:
This was inspired by the following PR 
- #3240

Fixes #2498 

I've updated the event simulator to be better typed and allow find all references.
I've created a handler for the cooperative gestures so it feels like the other handler
This handler creates and destroys the relevant UI (HTML) part, and is the only handler doing that.
I thought that a control would be a better fix, but control has add and remove, while the handler has enable disable and reacts to touch and scroll events.
So this handler is somewhat both...

 - [x] Link to related issues.
 - [x] Post benchmark scores.


HarelM:
cc: @arekgotfryd

src/ui/handler/cooperative_gestures.ts
```diff
@@ -0,0 +1,128 @@
...
+    _container: HTMLElement;
+    _metaKey: 'metaKey' | 'ctrlKey' = navigator.userAgent.indexOf('Mac') !== -1 ? 'metaKey' : 'ctrlKey';
+    _enabled: boolean;
+
+    constructor(map: Map, options: boolean | GestureOptions = {}) {
+        this._map = map;
+        this._options = options;
+        this._enabled = false;
+    }
+    isActive(): boolean {
```
neodescis, line 53: I guess I'm not sure what `isActive()` is used for, but why is this always false instead of returning _enabled?

neodescis: Ah, looks like you added a comment below explaining what isActive() is for. I take it this handler should never interfere with others then?

HarelM: Yes, it might be that this can be better defined so that the cross interaction is defined in the handler manager and not in each handler or better define the active flag so that it will block the interaction of other handlers, but I'm not sure I want to solve this as part of this PR as I'm not sure I know exactly how to define the active state right, maybe given a number of fingers or scroll zoom... IDK...
I'll open an issue to handle it in the future if it's interesting enough.

HarelM: I started writing the issue, and the decided that I should try and implement this, but I don't think it possible (easily), the interaction logic is more complicated than just true/false as there are types of interaction that requires two fingers, one finger and three fingers and there's no good way in the cooperative gestures handler to know how to set active properly.
The current logic of checking which is enabled between the handlers creates coupling, but it's very simple to reason with.

Issue #2870, opened 2023-07-16
Marker opacity changes before it's behind the mountain


jashanbhullar:
**maplibre-gl-js version**: latest/main branch 

**browser**: Mozilla Firefox 114.0 20230605223242 20230605223242 

### Steps to Trigger Behavior

 1.On your local branch `npm start`
 2.Open `3d-Terrain.html` 
 3.Add the following marker 
```javascript 
        const marker = new maplibregl.Marker()
            .setLngLat([11.39067, 47.305513])
            .addTo(map);
```
4. Change the angle so are somewhat looking at horizon, move close the mountain
5. The opacity of the marker changes even before it's behind the mountains

### Link to Demonstration

https://output.jsbin.com/vatekax#11.9/47.28354/11.40553/0/52
https://jsbin.com/vatekax/2/edit?html,output

### Expected Behavior
The marker visibility should only change when completely behind the mountain

### Actual Behavior
The opacity of the marker changes even before it's behind the mountains

[Screencast from 16-07-23 04:37:27 PM IST.webm](https://github.com/maplibre/maplibre-gl-js/assets/20891087/07e65570-4ea6-434f-9454-3789494684cb)

Found when working on this: https://github.com/maplibre/maplibre-gl-js/issues/1134 
Might be related

SnailBones:
@HarelM I'm interested in this

HarelM:
Good luck!

SnailBones:
Thanks @prozessor13! I've implemented a second visibility check as you suggest in https://github.com/maplibre/maplibre-gl-js/pull/3431.

Issue #2869, opened 2023-07-16
Marker opacity doesn't change when switching from 3D to 2D


jashanbhullar:
**maplibre-gl-js version**: latest/main branch 

**browser**: Mozilla Firefox 114.0 20230605223242 20230605223242  

### Steps to Trigger Behavior

 1.On your local branch `npm start`
 2.Open `3d-Terrain.html` 
 3.Add the following marker 
```javascript 
        const marker = new maplibregl.Marker()
            .setLngLat([11.39067, 47.305513])
            .addTo(map);
```
4. Go behind the mountain, the marker will be less visible 
5. Switch to 2D, the marker still remains it's original opacity 

### Link to Demonstration

https://output.jsbin.com/vatekax#11.9/47.28354/11.40553/0/52
https://jsbin.com/vatekax/2/edit?html,output

### Expected Behavior
The marker should be opaque when switched to 2D from 3D

### Actual Behavior
The marker opacity doesn't change when switching from 3D to 2D

[Screencast from 16-07-23 04:14:11 PM IST.webm](https://github.com/maplibre/maplibre-gl-js/assets/20891087/fa346c44-1e1f-4cfc-88aa-4ff501790bd3)

Found when working on this: https://github.com/maplibre/maplibre-gl-js/issues/1134 
Might be related

visitskyworld:
@HarelM can I take this issue?

HarelM:
Same as the other one, sorry...

SnailBones:
Is anyone working on this? I'd be interested unless @visitskyworld or @jashanbhullar still wants to go for it!

> Same as the other one, sorry...

@HarelM Is this the same issue as #1134? If so perhaps this one should be closed so they can be tracked in one place?

HarelM:
No, same response as #2868 

I'm not aware of any current effort to solve these bugs.
The previous bug was closed, but there are still more bugs in this area.

Since I didn't see any progress lately, feel free to take it.

SnailBones:
https://opencollective.com/maplibre/expenses/176000

Issue #1134, opened 2022-03-30
Change marker-visibility logic from coordinate comparison to depth-buffer check


prozessor13:
In [marker.ts](https://github.com/maplibre/maplibre-gl-js/blob/terrain3d/src/ui/marker.ts#L477) the visibility check of a marker is done via:

* get the screen-coordinate of the marker anchor
* get the lnglat-coordinate in terrain for the marker screen-pixel
* check if marker-lnglat & screen-lnglat are equal
* if yes show, else hide the marker

this implementation is from a state before i implemented the depth-framebuffer. So this needs to be changed to the depth-buffer check as well. In my opinion a minor-task.

HarelM:
There's another question here related `getBounds` - in case of pitch > 60 this might return an area that is way too big, wouldn't it?
Maybe the marker logic can use this method to check for visibility, I.e. If the marker is within the map bounds, if this function is relatively accurate it can be very useful.

HarelM:
Assigned S bounty.
Link to parent Bounty: https://github.com/maplibre/maplibre/issues/189

SnailBones:
@HarelM I'm interested in this too! Seems very related to https://github.com/maplibre/maplibre-gl-js/issues/2870, I'm hoping the switch can solve that issue.

HarelM:
Yes, all these marker bugs are related to some extent, I believe.
Assigning to you.

PR #3431, opened 2023-12-01
Improvements to marker transparency

SnailBones:
Fixes https://github.com/maplibre/maplibre-gl-js/issues/2869. Markers now update when removing terrain.

[disable-terrain.webm](https://github.com/maplibre/maplibre-gl-js/assets/14878684/6c25be2f-4f17-46d1-b1b5-ac666829c541)

Fixes https://github.com/maplibre/maplibre-gl-js/issues/1134.

Instead of checking the position of the marker and the unprojected screenpoint in LngLat space, now we sample the depth texture (with function `terrain.depthAtPoint`) and convert the marker's position to depth in GL space (with function `transform.lngLatToCameraDepth`) and compare these values. ~~For more precision, I've adjusted the near plane. This means that buildings clip slightly sooner but it doesn't seem to be an issue.~~ I reverted this change as it caused incorrect render order with [DeckGL and extrusions](https://docs.maptiler.com/deck-gl/interleaved/). Some precision is lost but given the following double-check it's rarely noticeable. 

Fixes https://github.com/maplibre/maplibre-gl-js/issues/2870

If the initial check at the marker's lng/lat finds a colission we check again at a position with elevation adjusted to the center of the marker. This resolves this issue where markers were being occluded by a small hill at the base that only covers a small portion of the marker.

[fade-behind-hill.webm](https://github.com/maplibre/maplibre-gl-js/assets/14878684/399fa755-10f6-478b-8404-5e5884357190)

## Launch Checklist

 - [X] Link to related issues.
 - [ ] Post benchmark scores.


src/geo/transform.ts
```diff
@@ -833,10 +835,10 @@ export class Transform {
         // - the more depth precision is available for features (good)
         // - clipping starts appearing sooner when the camera is close to 3d features (bad)
         //
-        // Smaller values worked well for mapbox-gl-js but deckgl was encountering precision issues
-        // when rendering it's layers using custom layers. This value was experimentally chosen and
-        // seems to solve z-fighting issues in deckgl while not clipping buildings too close to the camera.
-        const nearZ = this.height / 50;
+        // With a smaller value, we see precision issues detecting when markers are behind terrain.
```
HarelM, line 838: What about the z fighting stuff?

src/geo/transform.ts
```diff
@@ -955,4 +957,18 @@ export class Transform {
...
+     * Return the distance to the camera in GL coordinates from a LngLat.
+     * This can be compared to the value from the depth buffer (terrain.depthAtPoint)
+     * to determine whether a point is occluded.
+     * @param lngLat - the point
+     * @param elevation - the point's elevation
+     * @returns value between 0 and 1
+     */
+    lngLatToCameraDepth(lngLat: LngLat, elevation: number) {
+        const coord = this.locationCoordinate(lngLat);
+        const p = [coord.x * this.worldSize, coord.y * this.worldSize, elevation, 1] as vec4;
```
HarelM, line 970: Is there a need to wrap coordinates here in regards to 180 here?

src/render/terrain.ts
```diff
@@ -348,6 +348,23 @@ export class Terrain {
         );
     }
 
+    /**
+     * Reads the depth value from the depth-framebuffer at a given screen pixel
+     * @param p - Screen-Coordinate
+     * @returns depth value for a given screen pixel
```
HarelM, line 354: In what units with this return?

HarelM:
Looks good in general, can you add some videos with 2D and 3D after the fix?

src/geo/transform.ts
SnailBones: My understanding is that the deck-gl z-fighting issues occurred with smaller values. Making this value bigger improves precision, so it shouldn't cause this issue. But I'll test it with deck.gl to be sure!

src/geo/transform.ts
SnailBones: I don't think so. Marker lngLats are already wrapped via `smartWrap`. And there's not much terrain to be seen along the 180th meridian!

src/geo/transform.ts
HarelM: Ok, let me know what you found out.

HarelM:
I see that some tests are failing...
Also I'm not sure about the accuracy of the coverage, but it should be higher...

src/geo/transform.ts
SnailBones: Good catch @HarelM, I tested with deck.gl and it seems that this value needs to be _exactly_ this value in order to render deckgl layers and maplibregl buildings in the correct order. So I'm reverting this change. We lose some precision in the marker depth check, but with the second check (and adjusted forgiveness) it's barely noticeable.

SnailBones:
Thanks for the review @HarelM! I think the failing tests are all fixed :crossed_fingers: 

> Also I'm not sure about the accuracy of the coverage, but it should be higher...

Just added a unit test for `transform.lngLatToCameraDepth`! Unfortunately I don't think there's a simple way to unit test`terrain.depthAtPoint` .

src/ui/map.test.ts
```diff
@@ -570,7 +570,7 @@ describe('Map', () => {
             map.addSource('geojson', createStyleSource());
             expect(map.isSourceLoaded('geojson')).toBe(false);
             await promise;
-        });
+        }, 20000); // 20 second time out instead of the default 5 to prevent flaking.
```
HarelM, line 573: I'll be surprised if this would help, this test is super annoying, I have spent some time to debug it, and it seems that the order of events is different when the test fails, but I still didn't get to the bottom of this.
I would suggest to revert this change and leave this test to a different PR as I think there is a real big here.

src/geo/transform.ts
```diff
@@ -458,13 +458,15 @@ export class Transform {
     zoomScale(zoom: number) { return Math.pow(2, zoom); }
     scaleZoom(scale: number) { return Math.log(scale) / Math.LN2; }
 
+    // Convert from LngLat to world coordinates (Mercator coordinates scaled by 512)
```
HarelM, line 461: Can you use tsdoc style comments? This way you get it with intellisense in the IDE.

src/ui/marker.ts
```diff
@@ -506,6 +506,43 @@ export class Marker extends Evented {
         return this;
     }
 
+    _updateOpacity(force: boolean = false) {
+        const terrain = this._map.terrain;
+        if (!terrain) {
+            this._element.style.opacity = '1';
```
HarelM, line 512: Should updating the opacity be optional? What if someone would like to set the opacity of a marker regardless of the state of the terrain?
In the previous code, this was only done in case of 3D, but in this case it is set to 1 even if terrain is off, which might cause issues for 2D use cases...?

HarelM: I've tested an example I have that is somewhat transparent and it seems like I'm using the background color transparency and not the opacity, nevertheless, this is worth thinking about.

HarelM:
In general this looks good. See my last comment above.

HarelM (approved):
THANKS!!

SnailBones:
> 516, 520, and 534-535 are part of the changes made here.
> Let me know if you would like to write tests to cover those, I don' know how critical those are, but it's probably easier to write now than later...

@HarelM Just expanded the test, I think it should cover most of those!

How are you seeing the list of lines not covered? Going to https://app.codecov.io/gh/maplibre/maplibre-gl-js/pull/3431 I see "No Files covered by tests were changed"

PR #3432, opened 2023-12-01
Improve `extend` typings

HarelM:
This adds proper typings to the `extend` method.
To be honest, I'm not sure if this method is the same as `Object.assign`, so I'm not sure it's actually needed.

But, just in case it's not, I improved the typings on it.

### DIFF
--- a/src/source/geojson_source.ts
+++ b/src/source/geojson_source.ts
@@ -17,6 +17,7 @@ import type {GeoJSONWorkerOptions, LoadGeoJSONParameters} from './geojson_worker
 export type GeoJSONSourceOptions = GeoJSONSourceSpecification & {
     workerOptions?: GeoJSONWorkerOptions;
     collectResourceTiming?: boolean;
+    data: GeoJSON.GeoJSON | string;
 }
 
 export type GeoJsonSourceOptions = {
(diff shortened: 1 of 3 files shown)


HarelM:
I went ahead and changed all the usage of extend to Object.assign which seems to fail a few tests.
Seems like the problem is that the following is not the same, I'm not entirely sure why:
```js
const event =extend({}, e, {ctrlKey: true, preventDefault: () => e.preventDefault()});
const event2 =extend({}, {...e}, {ctrlKey: true, preventDefault: () => e.preventDefault()});
const event3 = Object.assign({}, e, {ctrlKey: true, preventDefault: () => e.preventDefault()});
```
event1 != event2 = event3, which might be related to the fact that we are trying to copy data from an event, but generally speaking, I'm not sure I understand why... :-/

HarelM:
Here it is in the browser console - weird...
image

neodescis:
It looks like the difference here is [`for...in`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in) includes inherited properties, whereas [`Object.assign`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) only looks at "own" properties.

PR #3434, opened 2023-12-02
Fix popup.trackPointer pointer cursor bug

### LINKED ISSUE #3405: Popup .trackPointer() changes cursor to pointer forever
When adding "trackPointer()" to the popup, it changes cursor to pointer when hovering over a feature. It is then impossible to change it back to "default", even if setting it in the code.

This results in cursor being a pointer even if its not hovering any of the features and popup is not showing or is removed.

Replication (the code is taken straight from the documentation, just added the method to the popup on line 159): https://jsfiddle.net/Lynxe7tu/

As I understand, "trackPointer" sets cursor to pointer in the library itself. But then if I hide the popup on "mouseleave" it does not allow to set cursor to "default". User setting should be preferred.

simondriesen:
Fixes #3405

### Bug description
The ".maplibregl-track-pointer" CSS class get's added to the map when a popup uses its `trackPointer` method. This makes sure the map displays a pointer cursor. The issue is that when you disable the popup using its `remove` method it doesn't remove this CSS class from the map causing the cursor to be shown as a pointer indefinitely.

### Changes
When using a popup's `remove` method it will make sure to remove the ".maplibregl-track-pointer" CSS class from the map.

Before:

After:

## Launch Checklist

 - [x] Link to related issues.
 - [x]  Write tests for all new functionality.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ### 🐞 Bug fixes
 - Fix zooming outside the central globe when terrain 3D is enabled ([#3425](https://github.com/maplibre/maplibre-gl-js/pull/3425))
+- Fix cursor being shown indefinitely as a pointer when removing a popup with its `trackPointer` method active ([#3434](https://github.com/maplibre/maplibre-gl-js/pull/3434))
 - _...Add new stuff here..._
 
 ## 4.0.0-pre.1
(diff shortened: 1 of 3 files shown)


HarelM:
Can you add some assert to a test to make sure this won't break in the future?

PR #3452, opened 2023-12-05
Upgrade node to version 20

HarelM:
Upgrades node to version 20, also removed unwanted console log.
Removed `gl` package as it's an overkill for the unit test and jest can mock most of it.
Replaced `jest-canvas-mock` with `jest-webgl-canvas-mock` which is a fork that allows to mock some of the webgl methods.
If the following is merged some mocking can be removed.
- https://github.com/Adamfsk/jest-webgl-canvas-mock/pull/5


HarelM:
Forgot to disable auto merge... :-/
@birkskyum feel free to take a look at this as I have remove the `gl` package which should simplify some stuff when running on linux I believe.

PR #3455, opened 2023-12-06
Update CHANGELOG.md with node upgrade

HarelM:
In continue to #3452, where I forgot to add this...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## main
 
 ### ✨ Features and improvements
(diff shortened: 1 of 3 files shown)


louwers (commented):
Not really relevant for end-users?

HarelM:
No, more for developers and for me to know when things changed so I can backtrack issues without looking at the git history which is filled with dependencies updates.

HarelM:
@louwers let me know if you want me to remove this changelog comment or not, I don't have strong feelings either way.

louwers (approved):
Ah OK, sure. Maybe a new category for this makes sense, since probably a lot more users that are not developers read the changelog?

PR #3459, opened 2023-12-06
Allow `addProtocol` in the worker code

HarelM:
This allows adding a protocol in the worker thread by exposing an addProtocol method in the workers code and allowing to import a code into the workers.

One issue I noticed is that the `raster` source is not using the worker thread at all since all it does it fetch an image and convert it to pixels, so there's not much work in the works anyway, which means that if you add a protocol in the worker thread but use a raster source it won't be called.
This means that the developer needs to know where to add the method, which is OK, as the default is in the main thread and only if you'd like to optimize things you can move it to the worker thread.

The rest is the same.

I've also removed the import of the script when adding a source type, so that importing a script has only one entry point.

cc: @bdon, @msbarry 

## Launch Checklist

 - [x] Link to related issues.
 - [x] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## main
 
 ### ✨ Features and improvements
+- ⚠️ Add the ability to import a script in the worker thread and call `addProtocol` and `removeProtocol` there ([#3459](https://github.com/maplibre/maplibre-gl-js/pull/3459)) - this also changed how `addSrouceType` works since now you'll need to load the script with `maplibregl.importScriptInWorkers`.
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
--- a/src/index.ts
+++ b/src/index.ts
@@ -15,7 +15,7 @@ import {LngLatBounds} from './geo/lng_lat_bounds';
 import Point from '@mapbox/point-geometry';
 import {MercatorCoordinate} from './geo/mercator_coordinate';
 import {Evented} from './util/evented';
-import {AddProtocolAction, config} from './util/config';
+import {config} from './util/config';
 import {Debug} from './util/debug';
 import {isSafari} from './util/util';
 import {rtlMainThreadPluginFactory} from './source/rtl_text_plugin_main_thread';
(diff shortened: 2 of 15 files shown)


HarelM:
To overcome this limitation, I think I can think of a way to register the worker to `getResource` actor message so that the main thread can try and call the worker thread so that it doesn't matter where you registered the protocol - there's a need to make sure this doesn't cause an endless loop, but I think it's possible without a lot of changes - this will reduce the need to know where to add the protocol, and you can fine tune the performance if you decide to (opt-in kind of performance).


Regarding shared tiles, this is a different story and I'm not sure this is related to this PR, we also have the same issue with the terrain and hillshding - i.e. you have to use different sources so that things are not colliding (tile fetch aborting etc). I'm not sure I have an easy solution for this...

HarelM:
Passing a actor to the `makeRequest` method when in the main thread proved harder than I though...
I'm not sure I have a good solution for this ATM, and the more I look at it the less I'm convinced this is the right direction.
I'll try again and see if I can find an elegant solution, but I think that the need to know where to run your code is problematic and kinda a deal breaker from my perspective as a developer.
IDK... I'll sleep on it.

HarelM:
I had the following idea:
Instead of creating a public API to add a protocol to the worker thread I can let the plugin owner decide on the implementation while giving them all the relevant building blocks by providing the following two new APIs:
1. The ability to load a script in the worker thread
2. The ability to add a protocol method in the worker thread

This way it is up to the developers to properly test everything and make sure it's working as expected instead of trying to solve everything for them.

I'll change this PR so you can take a look at it.

msbarry:
I think that makes more sense. Seems like from a users perspective they just want to say `maplibregl.addPlugin(plugin)` then the contract is between the plugin writer and maplibre. It seems like this moves us more in that direction, or maybe if the api is `plugin.addTo(maplibregl)` then we just need to expose APIs like this for the plugin to hook into.

HarelM:
I guess a guide about how to write a plugin that uses these abilities would create a common plugins look and feel, hopefully. Generally speaking I'm not sure what API a plugin should implement, so I'm not sure how to define this API.
Also there are a lot of types of plugins, so I'm not sure all should conform to the same API.
@msbarry let me know if you think the code in it's current state is OK and I'll merge it.

Thanks again for the input!

HarelM:
@birkskyum @neodescis can you guys approve this?
I would like to merge this to main.

PR #3515, opened 2023-12-28
Set terrain from diff

### LINKED ISSUE #3463: When updating the style with `diff:true` and terrain change the map doesn't change
There is a way to have the map look for changes in the style and apply them automatically, which is a great feature, but it doesn't take the terrain part of the spec into consideration.

Here's an example:
https://jsbin.com/muriqipaba/edit?html,output
When switching from `diff: true` to `diff:false` the 3D is removed, but it should also work for diff mode of update style.

Motivation:
I tried to add the terrain configuration to maputnik only to discover this doesn't work due to this issue.
Also maputnik doesn't use typescript, so upgrading to version 3/4 from the current version (2.4) will be a nightmare...

HarelM:
## Launch Checklist

Fixes #3463 - add support for terrain in diff
- #3463

This depends on the following PR to be merged and published first:
-  https://github.com/maplibre/maplibre-style-spec/pull/460

Also added some types to `pick` method.

 - [x] Link to related issues.
 - [x] Post benchmark scores.


HarelM:
The 3 missing lines in the coverage report are related to a situation where the diffStyle function from the style spec package returns a different operation than expected.
I'm not sure it's worth the extra effort of trying to mock the diffStyle method in order to cover this code, it's mostly a safeguard.

This is now ready for review.

src/util/util.test.ts
```diff
@@ -11,7 +11,7 @@ describe('util', () => {
     expect(keysDifference({a: 1}, {a: 1})).toEqual([]);
     expect(extend({a: 1}, {b: 2})).toEqual({a: 1, b: 2});
     expect(pick({a: 1, b: 2, c: 3}, ['a', 'c'])).toEqual({a: 1, c: 3});
-    expect(pick({a: 1, b: 2, c: 3}, ['a', 'c', 'd'])).toEqual({a: 1, c: 3});
+    expect(pick({a: 1, b: 2, c: 3}, ['a', 'c', 'd'] as any)).toEqual({a: 1, c: 3});
```
birkskyum, line 14: why is `as any` needed here, but not on the line above?

HarelM: because `d` is not part of `{a, b, c}` object.

birkskyum: I was wondering if it's better to change the KeysOfUnion code to allow for attempting to pick non-existing keys, but i guess that would just remove useful mis-match information.

birkskyum (approved):

neodescis (approved):

PR #3528, opened 2024-01-01
Update plugins.md: add jekyll integration

rriemann:
This is only adding a reference to the documentation concerning the maplibre integrations


HarelM (approved):
I think you need to know a lot about maplibre, how it works and how to configure it, but everything is there, so kudos!

Issue #1591, opened 2022-08-30
Add option to force integer-only zoom


zdila:
## Motivation

Raster maps are often slightly blurred if non-integer zoom is used. MapLibre should allow option to force integer-only zoom levels.

## Design Alternatives

Add `options.forceIntegerZoom`. An alternative is to guard it programatically outside of MapLibre.

Duoquote:
I am using map with;

```js
map.scrollZoom.setZoomRate(10);
map.scrollZoom.setWheelZoomRate(10);
```

But still, at random times the zoom level gets changed to a float value. I think most straightforward thing can be done is this right now;

```js
map.on("zoomend", () => {
  const zoom = map.getZoom();
  if (zoom !== Math.round(zoom)) {
    map.setZoom(Math.round(zoom));
  }
});
```

Not elegant but works.

1ec5:
I suspect most users would expect the map to mostly track their zoom gesture even if it eventually snaps to an integer zoom level. I took https://github.com/maplibre/maplibre-gl-js/issues/1591#issuecomment-2748319483 and [made a few modifications](https://jsbin.com/fodosuc/6/edit?html,output) to feel more natural on a trackpad or mouse capable of incremental scrolling or on a touch screen. It eases to the snapped zoom level with a 300-millisecond animation, pinning the last touch/scroll event’s coordinates to the same location on screen. This imitates Leaflet’s behavior on mobile browsers. However, it still feels a bit unresponsive, because the snapping starts on the `zoomend` event, which only fires after the inertia animation that follows the actual zoom gesture. I’ve tested it in Firefox with a multitouch trackpad. It doesn’t work yet in MobileSafari.

Prior to the fork, https://github.com/mapbox/mapbox-gl-js/issues/1695#issuecomment-478392580 prototyped a “rachet” effect that would snap the zoom level while the zoom gesture is still ongoing. I think the idea is that the map would tear away from the gesture and catch back up repeatedly. This would be closer to Leaflet’s behavior on desktop browsers when using a scroll wheel or scrolling gesture. Unfortunately, [this animated version](https://jsbin.com/tefeqal/3/edit?html,output) doesn’t seem to work reliably on my incremental zoom–enabled trackpad. Anyways, I prefer the behavior of only snapping after the zoom gesture.

CommanderStorm:
To update this issue:
- we now have `zoomSnap` as an API, this allows you to snap to a certain zoom
[...]

PR #3532, opened 2024-01-02
Snap to integer zoom level

quinncnl:
Add an option to snap to integer zoom level: https://github.com/maplibre/maplibre-gl-js/issues/1591

Example: https://tracestrack.s3-website.nl-ams.scw.cloud/dist/
Example: https://tracestrack.s3-website.nl-ams.scw.cloud/dist/#12.40/11.7933/53.4402
jsfiddle: https://jsfiddle.net/142rftnx/1/

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
Quick question, What about touch gestures?
Also not that there's a similar behavior to snap bearing to north, might be worth getting inspiration from, maybe...

quinncnl:
Thanks for the note! I'll check those :)

quinncnl:
I've done some changes to `ScrollZoom`. For `TagDragZoom` and `TwoFingersTouchZoom` it requires more investigation and probably a bigger change. Can we ignore the touch zoom for now?

HarelM:
I don't think we can ignore the touch events without a flag that is specifying to which gestures the interger zoom applies I guess.
Also I'm not sure the animation of the zoom scroll in the demo is working well...
Can you add hash control to the demo so that zoom level will be easy to see?

quinncnl:
OK will add hash control. For the touch event it'll take some time since it's quite complicated.

HarelM:
What I meant to say is that if you decide not to implement touch handling the flag specifying the integer zoom should be something like:
snapToIntegerZoom: {touch: false, scroll: true, controls: true}
And not a simple boolean, this will on one hand allow more granularity and also provide a way to implement just the relevant part.
When all are implemented, true and false can be added to indicate "apply to all" kind of configuration.

HarelM:
Any updates on this?

quinncnl:
Hey @HarelM sorry I got busy. Will try to followup on it this weekend. I'll break it into smaller pieces as you suggested.

quinncnl:
Hi @HarelM I've updated the example map and added a jsfiddle. Could you please check it and provide feedback? Once it looks good I'll fix the tests and docs.

src/ui/map.ts
```diff
@@ -1019,6 +1027,8 @@ export class Map extends Camera {
      */
     getMaxZoom(): number { return this.transform.maxZoom; }
 
+    getSnapToIntegerZoomOptions(): SnapToIntegerZoomOptions { return this._snapToIntegerZoomOptions; }
```
HarelM, line 1030: Consider getting a string of the relevant option instead of returning the object, this way in the future you can check if the value of this is a boolean and return that value, obfuscating the internal implementation.

src/ui/map.ts
```diff
@@ -1008,6 +1014,8 @@ export class Map extends Camera {
         } else throw new Error('maxZoom must be greater than the current minZoom');
     }
 
+    setSnapToIntegerZoom(snap: SnapToIntegerZoomOptions): Map { this._snapToIntegerZoomOptions = snap; return this; }
```
HarelM, line 1017: Use multiple line instead of a single one.

src/ui/handler/tap_zoom.ts
```diff
@@ -57,7 +57,7 @@ export class TapZoomHandler implements Handler {
             return {
                 cameraAnimation: (map: Map) => map.easeTo({
                     duration: 300,
-                    zoom: tr.zoom + 1,
+                    zoom: Math.round(tr.zoom + 1),
```
HarelM, line 60: This should be using the check?

src/ui/handler/tap_zoom.ts
```diff
@@ -68,7 +68,7 @@ export class TapZoomHandler implements Handler {
             return {
                 cameraAnimation: (map: Map) => map.easeTo({
                     duration: 300,
-                    zoom: tr.zoom - 1,
+                    zoom: Math.round(tr.zoom - 1),
```
HarelM, line 71: This should be using the check?

HarelM:
I think this direction is OK.

src/ui/map.ts
quinncnl: Could you elaborate what kind of string? When we move to single boolean toggle we can still use this options object right?

HarelM: Use `type SnapType = "box" | "scroll" ...` as the input parameter to this method.

sbachinin:
Noticed that in your [jsfiddle](https://jsfiddle.net/L2zdp3eq/1/) if you zoom out to 3.0, you can't zoom in back

src/ui/map.ts
quinncnl: Done

src/ui/map.ts
quinncnl: Done

src/ui/handler/tap_zoom.ts
quinncnl: Done

src/ui/handler/tap_zoom.ts
quinncnl: Done

src/ui/map.ts
```diff
@@ -61,6 +61,8 @@ import type {QueryRenderedFeaturesOptions, QuerySourceFeatureOptions} from '../s
 
 const version = packageJSON.version;
 
+export type SnapToIntegerZoomType = 'boxZoom' | 'scrollZoom' | 'clickZoom' | 'tapZoom';
```
HarelM, line 64: I think you can use `keyof SnapToIntegerZoomOptions`...?

src/ui/map.ts
```diff
@@ -1019,6 +1032,19 @@ export class Map extends Camera {
      */
     getMaxZoom(): number { return this.transform.maxZoom; }
 
+    shouldSnapToIntegerZoom(type: SnapToIntegerZoomType): boolean {
```
HarelM, line 1035: `SnapToIntegerZoomType` contains 4 values, this switch case contains 3, is it intentional?

src/ui/map.ts
quinncnl: Nice! I'm not expert of typescript so your help is useful :)

src/ui/map.ts
qingcai-tomtom: I missed it! Thanks

src/ui/map.ts
quinncnl: Fixed

src/ui/map.ts
quinncnl: Fixed

src/ui/camera.ts
```diff
@@ -113,6 +138,11 @@ export type CameraForBoundsOptions = CameraOptions & {
      * The maximum zoom level to allow when the camera would transition to the specified bounds.
      */
     maxZoom?: number;
+
+    /**
+     * Whether the map should snap to integer zoom levels during the transition.
+     */
+    shouldSnapToIntegerZoom?: boolean;
```
quinncnl, line 145: Note: this is for BoxZoom only.

src/ui/map.ts
HarelM: Since SnapToIntegerZoomType is keyof SnapToIntegerZoomOptions I think you can simply use return `this._snapToIntegerZoomOptions[type]` can't you?

test/build/min.test.ts
```diff
@@ -29,7 +29,7 @@ describe('test min build', () => {
 
         // Need to be very frugal when it comes to minified script.
         // Most changes should increase less than 1k.
-        const increaseQuota = 1024;
+        const increaseQuota = 1600;
```
HarelM, line 32: Better to increase the expectedBytes instead of the quota.

src/ui/map.ts
quinncnl: Yes indeed!

test/build/min.test.ts
quinncnl: Indeed. Updated

src/ui/map.ts
```diff
@@ -1036,6 +1031,35 @@ export class Map extends Camera {
...
+     * @example
+     * ```ts
+     * map.setSnapToIntegerZoom({boxZoom: true});
+     * ```
+     */
+    setSnapToIntegerZoom(options: SnapToIntegerZoomOptions | null): Map {
+        if (options === null || options === undefined) {
+            this._snapToIntegerZoomOptions = {boxZoom: false, clickZoom: false, scrollZoom: false, tapZoom: false};
+        } else {
+            this._snapToIntegerZoomOptions = options;
```
HarelM, line 1049: I think this will reset other snap options, wouldn't it? i.e. if I set scroll to true in the first call and set box to true in the second call, I'll reset scroll configuration.
Consider doing a for loop here maybe? Not sure if it's a great idea though...

HarelM: Or maybe consider the API to be similar to the get/should

src/ui/handler/scroll_zoom.ts
```diff
@@ -193,6 +197,24 @@ export class ScrollZoomHandler implements Handler {
...
+            const pos = DOM.mousePos(this._map.getCanvas(), e);
+            e.preventDefault();
+            let zoomTarget = value > 100 ? this._map.getZoom() - 2 :
+                value > 0 ? this._map.getZoom() - 1 :
+                    value > -100 ? this._map.getZoom() + 1 :
+                        value < -100 ? this._map.getZoom() + 2 : 0;
+
+            zoomTarget = Math.round(zoomTarget);
+
+            this._map.easeTo({
```
HarelM, line 210: It seems like there is a explicit animation code for regular zoom, how come this can be avoided using easeTo instead of writing the animation code (using frames etc)?

HarelM:
@quinncnl did you check/fix the issue reported by @sbachinin?

src/ui/map.ts
quinncnl: I see your point. I think replacing the entire options is more logical here because we are setting `SnapToIntegerZoomOptions` instead of `type: SnapToIntegerZoomType`. It's also same like other setters like `style: StyleSpecification`.

> Or maybe consider the API to be similar to the get/should

Something like `setSnapToIntegerZoom(type: SnapToIntegerZoomType, value: boolean)`? Normally this options is set once at the map init. Using this kind of setter is not very convenient in this common case. As you might find codes like:
```
setSnapToIntegerZoom("boxScroll", true);
setSnapToIntegerZoom("clickScroll", true);
```
instead of simple setOptions({boxScrolll: true, clickScroll: true})

src/ui/handler/scroll_zoom.ts
quinncnl: The explicit animation is about smoothOutEasing which doesn't have a targetZoom. The end zoom is based on an easing function which is hard to ease to an integer level. So I just skip that and use a simple `easeTo` instead.

quinncnl:
> @quinncnl did you check/fix the issue reported by @sbachinin?

I can't seem to reproduce the scenario. @sbachinin could you try again? Maybe it was before my refator to scrollZoom.

sbachinin:
This particular [jsfiddle](https://jsfiddle.net/L2zdp3eq/1/) shows the same bad behaviour.
It could be broken because of using some outdated bundle, I don't know.

Here's what I get when I try to zoom in:

https://github.com/maplibre/maplibre-gl-js/assets/11789697/c8fd5837-7b4a-4fcd-87ea-03a7c8e044c7

quinncnl:
Hi @sbachinin I updated the demo. Could you try again? It seems we have to round the zoom at the end of inertia animation. If this is not acceptable can we have a separate toggle for it, like `trackpadScroll` and do it later? @HarelM

HarelM:
I would expect the last commit to be added with a test to make sure this is the expected behavior.

sbachinin:
@quinncnl, all the demos given here still demonstrate zooming problems for me. (Chrome 122.0.6261.129)
When I roll mousewheel forward:
first several "clicks" are ignored, then I'm jumping to zoom 3.
When I roll mousewheel backward:
first several "clicks" are ignored, then begins a normal zoom out.

quinncnl:
I tried a different computer on Windows and there it jumps zoom by 2 with single scroll. Looks like I need to test with different browser/OS/scroll device(trackpad/mouse) as they have different scroll values.

quinncnl:
Hi @HarelM , do you think it's okay to move on without scroll_zoom? At least then people can build upon it.

HarelM:
There are missing tests and reported errors. How do you suggest to address them?

quinncnl:
> There are missing tests and reported errors. How do you suggest to address them?

I will revert all changes related to scrollZoom as it's hard to fix with my knowledge. It does not pre-calculate a targetZoom making it hard to achieve integer final zoom.

I'll add some tests to supported handlers.

HarelM:
@quinncnl are you still working on this?
Seems like this needs more work than you originally signed up for, and I think the implementation is too lacking to be considered as a full feature.
I'm trying to clean up stale PRs.
Let me know if I can close this one.

quinncnl:
Hi @HarelM indeed it requires more work still. I'm quite occupied and like to park it. If someone can pick it up that would nice!

HarelM:
I don't think there is ATM.
Closing this for now, if anyone would like to work on it in the future they would be able to find it here.

PR #3533, opened 2024-01-02
Fix upload action breaking changes

HarelM:
Improves the ci build by removing unneeded complexity in strategy and OS matric, and fix the upload artifact behavior change in version 4.

### DIFF
--- a/.github/workflows/test-all.yml
+++ b/.github/workflows/test-all.yml
@@ -77,11 +77,12 @@ jobs:
   render-tests-ubuntu:
     name: Render tests
     env:
-      SPLIT_COUNT: 2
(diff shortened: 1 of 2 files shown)


louwers (approved):
I can highly recommend [actionlint](https://github.com/rhysd/actionlint). It type checks YAML workflows and catches these things.

HarelM:
The problem was not the yaml, but the breaking changes of version 4 of the upload...
Github is using this when edition in the browser I think.

HarelM:
These "required" are fragile... :-/
In any case, I removed a lot of the duplication in yaml for the render tests now.

Issue #3504, opened 2023-12-19
geolocate control loses "locked" behavior with window resize and zoom adjustment.


michaelkirk:
**maplibre-gl-js version**: present in both v3.6.2 and main@7971b3a00

**browser**: Safari Version 17.1, Chrome 120.0.6099.109

### Context (how the geolocate control UI works) 

The geolocate_control has 3 states that I'm interested in. Here's their description with the associated geolocate control icon: 

**OFF**: user has not clicked geolocate control yet
<img width="55" alt="Screenshot 2023-12-19 at 11 25 13" src="https://github.com/maplibre/maplibre-gl-js/assets/217057/dbe31b52-e2ea-4c19-9dfe-8a0a0aa00e9c">

---

**BACKGROUND**: user location is being tracked, you can see their location as a blue dot on the map.

<img width="53" alt="Screenshot 2023-12-19 at 11 25 32" src="https://github.com/maplibre/maplibre-gl-js/assets/217057/90aeb6f3-517c-4e23-9691-27e85a9b79db">

---

**ACTIVE_LOCK**: user location is being tracked, you can see their location as a blue dot on the map, and the map camera is "locked" onto their position — e.g. if the user is on a bus, the map will remain centered on their location, "following" them as they are driven around.

<img width="46" alt="Screenshot 2023-12-19 at 11 25 21" src="https://github.com/maplibre/maplibre-gl-js/assets/217057/92eeb902-bbec-4de1-9b85-df2c03123f5e">

---

When you are in ACTIVE_LOCK state, panning the map changes the state to BACKGROUND - meaning you'll still see the blue dot, but since you've indicated that you want to look at other parts of the map, the map camera is no longer "locked" on your current location blue dot. This all seems to work as I expect.

### Steps to Trigger Behavior

1. click on geolocate control
2. resize window.

or

1. click on geolocate control
2. click zoom or pinch to zoom

### Link to Demonstration

Clicking and dragging is expected to break the camera lock, but resizing or zooming the browser is not. 

https://github.com/maplibre/maplibre-gl-js/assets/217057/8167a00c-d488-47ce-9a85-8fd1218b7382

#### This part works as expected:

1. click geolocate control, see that location is locked.
2. click and drag to move location, location is still tracked (you can see the blue dot), but the viewport no longer follows the current location.

Since the user has explicitly panned away from their location, they have indicated they want the map to be somewhere else. 

#### This part is unexpected:

[...]

PR #3535, opened 2024-01-04
strict lints for two_fingers_touch.ts

michaelkirk:
## Launch Checklist

 - [x] Briefly describe the changes in this PR:

I intended to only add some type declarations, but then there were some logical consequences that required some behavioral changes. I believe these strictly make things more robust in the "should never happen" case of state being undefined. I'll call them out below with inline comments momentarily.

As mentioned in https://github.com/maplibre/maplibre-gl-js/issues/3504#issuecomment-1875927741, I wanted to start adding more thorough and correct type information. It's my first commit to the project, so please let me know if something seems off.

 - [n/a?] Add an entry to `CHANGELOG.md` under the `## main` section.

### DIFF
--- a/src/ui/handler/two_fingers_touch.ts
+++ b/src/ui/handler/two_fingers_touch.ts
@@ -38,7 +38,7 @@ abstract class TwoFingersTouchHandler implements Handler {
     }
 
     abstract _start(points: [Point, Point]): void;
-    abstract _move(points: [Point, Point], pinchAround: Point | null, e: TouchEvent): HandlerResult | undefined;
+    abstract _move(points: [Point, Point], pinchAround: Point | null, e: TouchEvent): HandlerResult | void;
 
     touchstart(e: TouchEvent, points: Array<Point>, mapTouches: Array<Touch>): void {
         //log('touchstart', points, e.target.innerHTML, e.targetTouches.length ? e.targetTouches[0].target.innerHTML: undefined);
@@ -53,7 +53,7 @@ abstract class TwoFingersTouchHandler implements Handler {
         this._start([points[0], points[1]]);
     }
 
-    touchmove(e: TouchEvent, points: Array<Point>, mapTouches: Array<Touch>): ReturnType<typeof TwoFingersTouchHandler.prototype._move> {
+    touchmove(e: TouchEvent, points: Array<Point>, mapTouches: Array<Touch>): HandlerResult | void {
         if (!this._firstTwoTouches) return;
 
         e.preventDefault();
@@ -166,7 +166,7 @@ export class TwoFingersTouchZoomHandler extends TwoFingersTouchHandler {
         this._startDistance = this._distance = points[0].dist(points[1]);
     }
 
-    _move(points: [Point, Point], pinchAround: Point | null): ReturnType<typeof TwoFingersTouchHandler.prototype._move> {
+    _move(points: [Point, Point], pinchAround: Point | null): HandlerResult | void {
         const lastDistance = this._distance!;
         this._distance = points[0].dist(points[1]);
         if (!this._active && Math.abs(getZoomDelta(this._distance, this._startDistance!)) < ZOOM_THRESHOLD) return;
@@ -206,7 +206,7 @@ export class TwoFingersTouchRotateHandler extends TwoFingersTouchHandler {
         this._minDiameter = points[0].dist(points[1]);
     }
 
-    _move(points: [Point, Point], pinchAround: Point | null, _e: TouchEvent): ReturnType<typeof TwoFingersTouchHandler.prototype._move> {
+    _move(points: [Point, Point], pinchAround: Point | null, _e: TouchEvent): HandlerResult | void {
         const lastVector = this._vector!;
         this._vector = points[0].sub(points[1]);
 
@@ -285,7 +285,7 @@ export class TwoFingersTouchPitchHandler extends TwoFingersTouchHandler {
         }
     }
 
-    _move(points: [Point, Point], center: Point | null, e: TouchEvent): ReturnType<typeof TwoFingersTouchHandler.prototype._move> {
+    _move(points: [Point, Point], center: Point | null, e: TouchEvent): HandlerResult | void {
         // If cooperative gestures is enabled, we need a 3-finger minimum for this gesture to register
         if (this._map.cooperativeGestures.isEnabled() && this._currentTouchCount < 3) {
             return;
--- a/src/ui/handler/two_fingers_touch.ts
+++ b/src/ui/handler/two_fingers_touch.ts
@@ -291,8 +291,8 @@ export class TwoFingersTouchPitchHandler extends TwoFingersTouchHandler {
             return;
         }
 
-        const vectorA = points[0].sub(this._lastPoints![0]!);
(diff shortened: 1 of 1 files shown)


src/ui/handler/two_fingers_touch.ts
```diff
@@ -20,27 +20,27 @@ export type AroundCenterOptions = {
...
-    reset() {
+    reset(): void {
         this._active = false;
         delete this._firstTwoTouches;
     }
 
-    abstract _start(points: [Point, Point]);
-    abstract _move(points: [Point, Point], pinchAround: Point, e: TouchEvent);
+    abstract _start(points: [Point, Point]): void;
+    abstract _move(points: [Point, Point], pinchAround: Point | null, e: TouchEvent): { zoomDelta: number; pinchAround: Point | null } | { bearingDelta: number; pinchAround: Point | null } | { pitchDelta: number } | void;
```
HarelM, line 41: I think this has a type defined somewhere, something like `HandlerResults` or something similar.

src/ui/handler/two_fingers_touch.ts
```diff
@@ -152,20 +152,24 @@ function getZoomDelta(distance, lastDistance) {
...
 
-    _start(points: [Point, Point]) {
+    _start(points: [Point, Point]): void {
         this._startDistance = this._distance = points[0].dist(points[1]);
     }
 
-    _move(points: [Point, Point], pinchAround: Point) {
+    _move(points: [Point, Point], pinchAround: Point | null): ReturnType<typeof TwoFingersTouchHandler.prototype._move> {
+        if (typeof this._distance == 'undefined' || typeof this._startDistance === 'undefined')  {
+            console.assert(false, `failed precondition - _distance: ${this._distance}, startDistance: ${this._startDistance}`);
```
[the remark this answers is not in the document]
HarelM: I don't see any benefit in adding `console.assert`.
You can also use `!` instead of adding this `if`, since, as you said, you know _distance and _startDistance are set in the method before.

src/ui/handler/two_fingers_touch.ts
```diff
@@ -191,21 +195,26 @@ function getBearingDelta(a, b) {
...
 
-    _start(points: [Point, Point]) {
+    _start(points: [Point, Point]): void {
         this._startVector = this._vector = points[0].sub(points[1]);
         this._minDiameter = points[0].dist(points[1]);
     }
 
-    _move(points: [Point, Point], pinchAround: Point) {
+    _move(points: [Point, Point], pinchAround: Point | null, _e: TouchEvent): ReturnType<typeof TwoFingersTouchHandler.prototype._move> {
+        if (typeof this._vector === 'undefined') {
```
HarelM, line 213: Same as above.
If you still want to add initialization, add another method called `initVector` so that the logic of this will be shared between both methods.

HarelM:
The addition of the types is great and appreciated, I think the extra `if`s are not so appealing and I would consider using the `!` operator in these cases.

src/ui/handler/two_fingers_touch.ts
```diff
@@ -117,21 +117,21 @@ abstract class TwoFingersTouchHandler implements Handler {
...
      */
-    isActive() {
-        return this._active;
+    isActive(): boolean {
+        return !!this._active;
     }
 }
 
-function getTouchById(mapTouches: Array<Touch>, points: Array<Point>, identifier: number) {
+function getTouchById(mapTouches: Array<Touch>, points: Array<Point>, identifier: number): Point | void {
```
neodescis, line 134: In my opinion, it would be better to specify `| undefined` instead, and explicitly `return undefined` at the end of the function. There's even a lint rule for that: [consistent-return](https://eslint.org/docs/latest/rules/consistent-return)

src/ui/handler/two_fingers_touch.ts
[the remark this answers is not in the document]
michaelkirk: I've removed the logging assertions and changed all the graceful fallback to null assertions as requested.

michaelkirk:
Thanks for the feedback. I think I've addressed everything. Please take another look.

src/ui/handler/two_fingers_touch.ts
```diff
@@ -253,46 +254,45 @@ const ALLOWED_SINGLE_TOUCH_TIME = 100;
  */
 export class TwoFingersTouchPitchHandler extends TwoFingersTouchHandler {
 
-    _valid: boolean | void;
-    _firstMove: number;
-    _lastPoints: [Point, Point];
+    _valid?: boolean;
```
HarelM, line 257: Can we initialize booleans and numbers instead of defining them as optional?

src/ui/handler/two_fingers_touch.ts
```diff
@@ -253,46 +254,45 @@ const ALLOWED_SINGLE_TOUCH_TIME = 100;
...
-    _move(points: [Point, Point], center: Point, e: TouchEvent) {
+    _move(points: [Point, Point], center: Point | null, e: TouchEvent): HandlerResult | void {
         // If cooperative gestures is enabled, we need a 3-finger minimum for this gesture to register
         if (this._map.cooperativeGestures.isEnabled() && this._currentTouchCount < 3) {
             return;
         }
 
-        const vectorA = points[0].sub(this._lastPoints[0]);
-        const vectorB = points[1].sub(this._lastPoints[1]);
+        const vectorA = points[0].sub(this._lastPoints![0]!);
```
HarelM, line 294: Do we really need the two "!" here?

HarelM:
Added a few minor comments, looks great, thanks!

src/ui/handler/two_fingers_touch.ts
michaelkirk: We could, but what value should we assign? Most of the undefined variables are initialized with the call to `_start` which has access to the touches. Values from those touches are used to initialize the variables.

Without the touches, we'll need to try to pick some good "default" nullish value, which might inadvertently change behavior. That's why in my [previous version](https://github.com/maplibre/maplibre-gl-js/pull/3535/commits/3278a5940c9055c4b1a508ddea40456b32511c65) I waited until we had access to touches to set reasonable defaults on the off chance that one of these vars was not set.

In the case of `_valid` we actually do discriminate on `false` vs `undefined`, so at least that one's not a trivial change:
```
 if (this._valid !== undefined) return this._valid;
```

I agree though: It's not nice to have all these "maybe undefined types", but that's not new, it's just making the reality of the existing types explicit. 

My hope is that you're OK to gradually add more type information without requiring fixing all the existing shortcomings that are made visible by surfacing those existing types. Of course if it's trivial, I'm all for improvements to behavior, but I feel like we're getting a bit beyond trivial here.

If you feel strongly that this should be addressed, let me know and I can think about possible approaches.

src/ui/handler/two_fingers_touch.ts
michaelkirk: Fixed - thanks!

src/ui/handler/two_fingers_touch.ts
HarelM: Fair enough, let's do it on step at a time.

HarelM (approved):

PR #3545, opened 2024-01-07
Fixing labels not appearing when enabling terrain at high zoom

SnailBones:
Fixes #3454

## Before:
[broken.jpg](https://github.com/maplibre/maplibre-gl-js/assets/14878684/b2590d5b-617f-40b2-8d89-d32c09c69fa6)

## After:
[fixed.webm](https://github.com/maplibre/maplibre-gl-js/assets/14878684/69f94f64-fedc-484d-bfdd-49b7cfd49d80)

## Launch Checklist

 - [X] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.
 - [X] Link to related issues.
 - [X] Include before/after visuals or gifs if this PR includes visual changes.
 - [x] Write tests for all new functionality.
 - [x] Add an entry to `CHANGELOG.md` under the `## main` section.

PR #3561, opened 2024-01-10
Fix failing browser test by updating tiles only when adding new terrain

SnailBones:
As noted by @HarelM in https://github.com/maplibre/maplibre-gl-js/pull/3558#issuecomment-1883814307, the change in https://github.com/maplibre/maplibre-gl-js/pull/3545 (presumably when combined with the changes in https://github.com/maplibre/maplibre-gl-js/pull/3515) resulted in a failing browser test, where "idle" events aren't being correctly fired.

This is caused by terrain tiles getting stuck in the "updating" state.  This happens when `sourceCache.reload()` is called needlessly, i.e. when terrain is set on a map that already has terrain, as happens in the render test.

This fix ensures that the tile reloading only happens when terrain is added to a map that previously had no terrain.

## Launch Checklist

 - [X] Link to related issues.

### DIFF
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -1928,7 +1928,8 @@ export class Map extends Camera {
             // add terrain
             const sourceCache = this.style.sourceCaches[options.source];
             if (!sourceCache) throw new Error(`cannot load terrain, because there exists no source with ID: ${options.source}`);
-            sourceCache.reload();
+            // Update terrain tiles when adding new terrain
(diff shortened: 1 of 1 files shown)


HarelM:
This sounds like an easy fix, thanks!
Do you think the reload method should be smarter to prevent future mistakes?

HarelM (approved):
Ok, I won't hold this PR hostage, let me know if this is something you can do later on

PR #3573, opened 2024-01-12
Fix bug with FullscreenControl when map is in shadow DOM

### LINKED ISSUE #3572: Fullscreen control not working properly when MapLibre exists within a shadow root
**maplibre-gl-js version**: 3.6.2

**browser**: Firefox & Chrome (probably all?)

### Steps to Trigger Behavior

 1. create a map with `FullscreenControl` within a custom element
 2. click full screen button
 3. notice that the button doesn't switch to "Exit Fullscreen" and doesn't work

### Link to Demonstration

https://files.trailstash.net/maplibre-shadow-root-fs-bug/

### Expected Behavior
can exit full screen the normal way

### Actual Behavior
you can't

### Probable fix

this is because `window.document.fullscreenElement` is set to the custom `map-libre` element, not the actual container of the map.

I haven't been able to setup the maplibre development environment to test this, but I'm fairly confident changing `const fullscreenElement...` to `let fullscreenElement` [here](https://github.com/maplibre/maplibre-gl-js/blob/0414c6cae6fd20ee51298ce8d8eb575c0ff03e40/src/ui/control/fullscreen_control.ts#L112) and adding this after it will fix it:
```javascript
while (fullscreenElement.shadowRoot) {
  fullscreenElement = fullscreenElement.shadowRoot.fullscreenElement;
}
````
`while` loop instead of a simple `if` in order to support maplibre being nested more than 1 shadow root deep.

I'm not sure if prefixed variants are necessary in this case, as custom elements / shadow DOM support are considerably newer than fullscreen support.

### PR DESCRIPTION
**Pull request by dschep**
This fixes #3572 by checking if the selected `fullscreenElement` has a `shadowRoot` that contains an element that is actually the element that is fullscreen.

It uses a `while` loop instead of a simple `if` in order to support the fullscreened element being more than 1 shadow root deep.

It does not check prefixed versions of `fullscreenElement` because the ShadowRoot API is much newer than the Fullscreen API.

Demo of the fix here: https://files.trailstash.net/maplibre-shadow-root-fs-bug-bugfix/

### CODE REVIEW

**Comment by HarelM** [comment]
Can you add s test to cover this scenario?

**Comment by dschep** [comment]
Ignore that test I added, it passed on main too 😬

**Comment by dschep** [comment]
AFAICT shadow DOM doesn't behave the same in jsdom as in real browsers, so this can't be tested that way 😞

**Comment by HarelM** [comment]
Use the browser integration tests then.

**Comment by HarelM** [comment]
@dschep any updates on this? Would be nice to merge this in before the release of 4.0.

**Comment by dschep** [comment]
sorry, @HarelM, haven't had time to dig into adding an integration test yet. I'll try to get to it this weekend.

**Comment by HarelM** [comment]
@dschep any chance to finalize this?

File: test/integration/browser/browser.test.ts [hunk 1]
```diff
@@ -278,7 +278,66 @@ describe('Browser tests', () => {
...
+            let fullscreenButton = document.getElementsByTagName("map-libre")[0].shadowRoot.querySelector(".maplibregl-ctrl-fullscreen") as HTMLButtonElement;
+            fullscreenButton.click();
+            await sleepInBrowser(1000);
+
+            return fullscreenButton.title;
+        });
+
+        expect(fullscreenButtonTitle).toBe('Exit fullscreen');
+
+      test('Marker: correct opacity after resize with 3d terrain', async () => {
```
HarelM, line 340:
```suggestion
    test('Marker: correct opacity after resize with 3d terrain', async () => {
```

test/integration/browser/browser.test.ts
```diff
@@ -278,6 +278,65 @@ describe('Browser tests', () => {
...
+            await map.once('idle');
+            let fullscreenButton = document.getElementsByTagName("map-libre")[0].shadowRoot.querySelector(".maplibregl-ctrl-fullscreen") as HTMLButtonElement;
+            fullscreenButton.click();
+            await sleepInBrowser(1000);
+
+            return fullscreenButton.title;
+        });
+
+        expect(fullscreenButtonTitle).toBe('Exit fullscreen');
+
```
HarelM, line 339:
```suggestion
    }, 20000);

```

test/integration/browser/browser.test.ts
```diff
@@ -278,6 +278,66 @@ describe('Browser tests', () => {
...
+                                source: 'osm'
+                            }]
+                        }
+                    });
+                    map.addControl(new maplibregl.FullscreenControl());
+                }
+            }
+            customElements.define('map-libre', MapLibre);
+            document.body.innerHTML = '<map-libre></map-libre>';
+
```
HarelM, line 327:
```suggestion

```

test/integration/browser/browser.test.ts
```diff
@@ -278,6 +278,65 @@ describe('Browser tests', () => {
...
+                    });
+                    map.addControl(new maplibregl.FullscreenControl());
+                }
+            }
+            customElements.define('map-libre', MapLibre);
+            document.body.innerHTML = '<map-libre></map-libre>';
+            await sleepInBrowser(100);
+
+            await map.once('idle');
+            const fullscreenButton = document.getElementsByTagName("map-libre")[0].shadowRoot.querySelector(".maplibregl-ctrl-fullscreen") as HTMLButtonElement;
```
HarelM, line 330:
```suggestion
            const fullscreenButton = document.getElementsByTagName('map-libre')[0].shadowRoot.querySelector('.maplibregl-ctrl-fullscreen') as HTMLButtonElement;
```

HarelM:
Surprisingly enough I was able to push code to your branch, didn't know this was possible.
In any case, I've added the relevant test and made sure it fails without the fix.
There are a few last lint errors that needs to be fixed, it's a bit annoying to do it from the web editor...

test/integration/browser/browser.test.ts
```diff
@@ -278,6 +278,65 @@ describe('Browser tests', () => {
         expect(markerScreenPosition.y).toBeCloseTo(378.1);
     }, 20000);
 
+    test('Fullscreen control should work in shadowdom as well', async () => {
+        const fullscreenButtonTitle = await page.evaluate(async () => {
+            function sleepInBrowser(milliseconds: number) {
+                return new Promise(resolve => setTimeout(resolve, milliseconds));
+            };
```
HarelM, line 285:
```suggestion
            }
```

sbachinin (approved):
Looks ok to me

dschep:
>Surprisingly enough I was able to push code to your branch, didn't know this was possible.

Yep, that's an option that can be enabled on PRs, I guess it was on by default in this repo bc I don't think I went out of my way to enable it.

> In any case, I've added the relevant test and made sure it fails without the fix.

Thank you!! Sorry I didn't get around to making a working test myself 😅

PR #3589, opened 2024-01-15
Investigate coverage reports

HarelM:
I hope this will work...
This is not to be merge...


HarelM:
I've opened the following bug:
https://github.com/istanbuljs/v8-to-istanbul/issues/238
I'll see what I get from there, hopefully this will help solve this issue.

PR #3591, opened 2024-01-15
Bugfix: _cameraForBoxAndBearing not fitting bounds properly when using asymettrical camera viewport and bearing 

fcwheat:
I work with rotated maps a lot and noticed that fitting to bounds was not functioning as expected for some cases. In digging into this came to realize that the NW-SE points on the bounding box were always being fit to even in cases when fitting to these points leads to the NE-SW points being out of frame.

The code was updated to consider all corners of the bounding box when rotating and fitting.

In performing the update it highlighted two existing test cases which were incorrect and so these were updated as well. A new test case was added which is similar to the case presented in the JS Fiddle.

I also added a small `degreesToRadians` utility function since it felt cleaner to use this than repeatedly typing `degrees * Math.PI / 180`

JS Fiddle with a demo of the bug: https://jsfiddle.net/z4xma2wk/3/

Actual result when fitting bounds with 45deg rotation:
too_zoomed

Approx. expected result when fitting bounds with 45deg rotation:
correct_zoom

## Launch Checklist

 - [x] Update changelog


HarelM:
Can you check if merging from main improve the code coverage percentage?
Otherwise, assuming this is well covered with the new tests, I think it can be merged.

HarelM:
Please also add a changelog entry.

HarelM (approved):

PR #3593, opened 2024-01-18
Fix example button position

JaeSeoKim:
## describe

some examples has overlay button, but they are render with wrong position.

so fixed the wrong positioning of the overlay buttons in the example.

#### before

image

https://maplibre.org/maplibre-gl-js/docs/examples/language-switch/

#### after

language-switch-alt


HarelM (approved):
Thanks!! Interesting how you are the first one to comment on this 😀

PR #3600, opened 2024-01-21
Fixing initializing with esm exports

timocov:
Related to https://github.com/timocov/dts-bundle-generator/discussions/291

cc @HarelM 

There are 2 commits and it works on both of them. Feel free to drop the last one if you don't feel comfortable with the changes in it (it is a tiny refactoring).

image

### DIFF
--- a/build/rollup/bundle_prelude.js
+++ b/build/rollup/bundle_prelude.js
@@ -13,9 +13,10 @@ function define(_, chunk) {
 
         var sharedChunk = {};
         shared(sharedChunk);
-        maplibregl = chunk(sharedChunk);
+        maplibregl = chunk(sharedChunk, sharedChunk);
+        maplibregl = sharedChunk;
         if (typeof window !== 'undefined') {
-            maplibregl.workerUrl = window.URL.createObjectURL(new Blob([workerBundleString], { type: 'text/javascript' }));
+            sharedChunk.setWorkerUrl(window.URL.createObjectURL(new Blob([workerBundleString], { type: 'text/javascript' })));
         }
     }
 }
--- a/build/rollup/bundle_prelude.js
+++ b/build/rollup/bundle_prelude.js
@@ -1,22 +1,32 @@
 /* eslint-disable */
 
-var shared, worker, maplibregl;
-// define gets called three times: one for each chunk. we rely on the order
-// they're imported to know which is which
-function define(_, chunk) {
-    if (!shared) {
-        shared = chunk;
-    } else if (!worker) {
-        worker = chunk;
-    } else {
-        var workerBundleString = 'var sharedChunk = {}; (' + shared + ')(sharedChunk); (' + worker + ')(sharedChunk);'
+var maplibregl = {};
 
-        var sharedChunk = {};
-        shared(sharedChunk);
-        maplibregl = chunk(sharedChunk, sharedChunk);
-        maplibregl = sharedChunk;
-        if (typeof window !== 'undefined') {
-            sharedChunk.setWorkerUrl(window.URL.createObjectURL(new Blob([workerBundleString], { type: 'text/javascript' })));
+var define = (function() {
+    var modules = {};
+
+    return function(moduleName, dependencies, moduleFactory) {
+        modules[moduleName] = moduleFactory;
+
+        // to get the list of modules see generated dist/maplibre-gl-dev.js file (look for `define(` calls)
+        if (moduleName !== 'index') {
+            return;
         }
-    }
-}
+
+        // we assume that when an index module is initializing then other modules are loaded already
+        var workerBundleString = 'var sharedModule = {}; (' + modules.shared + ')(sharedModule); (' + modules.worker + ')(sharedModule);'
+
+        var sharedModule = {};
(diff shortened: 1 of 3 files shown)


rollup.config.ts
```diff
@@ -39,7 +42,7 @@ const config: RollupOptions[] = [{
     output: {
         name: 'maplibregl',
         file: outputFile,
-        format: 'umd',
+        format: 'iife',
```
timocov, line 45: I don't think this is related to the fix, but it feels more what you want? It doesn't feel like you want to have UMD modules when you can simply return object you've initialized and store in the global variable. But happy to revert it, it should still work with UMD I think.

rollup.config.ts
```diff
@@ -23,6 +23,9 @@ const config: RollupOptions[] = [{
         sourcemap: 'inline',
         indent: false,
         chunkFileNames: 'shared.js',
+        amd: {
+            autoId: true,
```
timocov, line 27: this generated `moduleName` in `define` function calls: `define(moduleName, dependencies, factory)` (instead of `define(dependencies, factory)`) so you don't need to rely on the order of calls of `define` function.

build/rollup/maplibregl.js
```diff
@@ -6,42 +6,17 @@
...
 // which outputs them as AMD modules.
 
-// Shared dependencies, i.e.:
-/*
-define(['exports'], function (exports) {
-    // Code for all common dependencies
-    // Each module's exports are attached attached to 'exports' (with
-    // names rewritten to avoid collisions, etc.)
-})
-*/
```
timocov, line 16: This comment part feels confusing and it can get outdated easily so I decided to remove them, feel free to bring it back.

build/rollup/bundle_prelude.js
```diff
@@ -1,21 +1,32 @@
...
+
+        // we assume that when an index module is initializing then other modules are loaded already
+        var workerBundleString = 'var sharedModule = {}; (' + modules.shared + ')(sharedModule); (' + modules.worker + ')(sharedModule);'
+
+        var sharedModule = {};
+        // the order of arguments of a module factory depends on rollup (it decides who is whose dependency)
+        // to check the correct order, see dist/maplibre-gl-dev.js file (look for `define(` calls)
+        // we assume that for our 3 chunks it will generate 3 modules and their order is predefined like the following
+        modules.shared(sharedModule);
+        modules.index(maplibregl, sharedModule);
```
timocov, line 24: to make it even more stable this define function probably should go over `dependencies` and initialize them in the correct order and then pass in the correct order for each factory. but I think its fine for now ¯\_(ツ)_/¯

timocov, line 24: not sure if the first argument to factories that require `exports` chunkId` should be a common object `maplibregl` or not. `index` re-exports everything from `shared` anyway so it shouldn't be a problem but technically might be more accurate to do this instead. up to you

HarelM:
Thanks!! I need to understand what went wrong with the RTL text tests, as it seems like this is breaking them.
Might be related to the code in the worker thread which might be a bit different than what the RTL plugin is expecting, IDK...
What's the difference between UMD and IIFE BTW? These build systems confuse the hell out of me.. :-/

build/rollup/bundle_prelude.js
```diff
@@ -1,21 +1,32 @@
...
-        shared = chunk;
-    } else if (!worker) {
-        worker = chunk;
-    } else {
-        var workerBundleString = 'var sharedChunk = {}; (' + shared + ')(sharedChunk); (' + worker + ')(sharedChunk);'
-
-        var sharedChunk = {};
-        shared(sharedChunk);
-        maplibregl = chunk(sharedChunk);
+var maplibregl = {};
```
HarelM, line 3: Is this what's basically creates the global object?

HarelM:
Do I still need the `standalone.ts` file after this change?

HarelM:
I've merged this so I can try and continue working on this branch.

timocov:
> Do I still need the standalone.ts file after this change?

Well, not really. Your `index.ts` is kinda standalone (it includes everything into one bundle). Initially I thought that the default version is the one that you call csp (so a user have to upload it and provide the library the URL to the worker), but apparently you have other way around.

build/rollup/bundle_prelude.js
timocov: So practically only the rollup config matters as it wraps all other modules and exposes what you export.

timocov:
> What's the difference between UMD and IIFE BTW? These build systems confuse the hell out of me.. :-/

[IIFE](https://developer.mozilla.org/en-US/docs/Glossary/IIFE) doesn't generate any modules, it just merged all modules into one huge function and then it exports everything you export from the main entry point:

image

UMD is trying to detect what modules system a customer might use and based on that assigned exported value:

image

@HarelM I'm checking now and it seems that it would be better to either switch back to UMD, or remove `export default maplibregl` in `build/rollup/maplibregl.js` and add `(window as any).maplibregl = maplibregl;` (from standalone) to `build/rollup/bundle_prelude.js` to make sure that you assign an export to a proper "global" variable).

Issue #3458, opened 2023-12-06
`setOpacity` and opacity options for `Marker`


HarelM:
## User Story

As a user I would like to be able to set the opacity of a marker when terrain 3D is on in case I override it with css or for whatever other reason

## Rationale

When terrain is on and the marker is behind terrain the opacity is set to 0.2 and returned to 1.0 when not behind terrain.
If I set this value in the css I'll get a behavior that is not intended, and therefore there's a need to disable it or define other values beside 0.2 and 1.

## Impact
The is a workaround as one can programmatically change this so this is a "nice to have" API for the marker.

sbachinin:
This has to be closed I guess

HarelM:
Fixed by #3620

PR #3620, opened 2024-01-25
Add Marker's "opacity" option and "setOpacity" method

sbachinin:
Implements https://github.com/maplibre/maplibre-gl-js/issues/3458

A small caveat here.
This solution means that all users have to provide an array of 2 strings. (2nd string is the reduced opacity when marker is behind the terrain).
But in most cases (without 3d terrain) this second string will be useless and just an incovenience.
Perhaps it makes sense to split this single option and method to:
'opacity/setOpacity' and 'reducedOpacity/setReducedOpacity'

### DIFF
--- a/src/ui/marker.ts
+++ b/src/ui/marker.ts
@@ -159,9 +159,7 @@ export class Marker extends Evented {
         this._rotationAlignment = options && options.rotationAlignment || 'auto';
         this._pitchAlignment = options && options.pitchAlignment && options.pitchAlignment !== 'auto' ?  options.pitchAlignment : this._rotationAlignment;
         this.setOpacity(); // set default opacity
-        if (options) {
-            this.setOpacity(options.opacity, options.opacityWhenCovered);
-        }
+        this.setOpacity(options?.opacity, options?.opacityWhenCovered);
 
         if (!options || !options.element) {
             this._defaultMarker = true;
--- a/src/ui/marker.ts
+++ b/src/ui/marker.ts
@@ -74,10 +74,15 @@ type MarkerOptions = {
      */
     pitchAlignment?: Alignment;
     /**
-     * First string will be used as a Marker's opacity when it's in clear view. Second string - when a Marker is behind the terrain (with 3D terrain enabled).
-     * @defaultValue ['1', '0.2']
+     * Marker's opacity when it's in clear view (not behind 3d terrain)
+     * @defaultValue 1
      */
-    opacity?: [string, string];
+    opacity?: number;
+    /**
+     * Marker's opacity when it's behind 3d terrain
+     * @defaultValue 0.2
+     */
+    opacityWhenCovered?: number;
 };
 
 /**
@@ -132,8 +137,9 @@ export class Marker extends Evented {
     _rotation: number;
     _pitchAlignment: Alignment;
     _rotationAlignment: Alignment;
-    _opacity: [string, string];
     _originalTabIndex: string; // original tabindex of _element
+    _opacity: string;
+    _opacityWhenCovered: string;
     _opacityTimeout: ReturnType<typeof setTimeout>;
 
     /**
@@ -152,7 +158,8 @@ export class Marker extends Evented {
         this._rotation = options && options.rotation || 0;
         this._rotationAlignment = options && options.rotationAlignment || 'auto';
         this._pitchAlignment = options && options.pitchAlignment && options.pitchAlignment !== 'auto' ?  options.pitchAlignment : this._rotationAlignment;
-        this._opacity = options && options.opacity ? options.opacity : ['1', '0.2'];
+        this._opacity = options && options.opacity !== undefined ? String(options.opacity) : '1';
+        this._opacityWhenCovered = options && options.opacityWhenCovered !== undefined ? String(options.opacityWhenCovered) : '0.2';
 
         if (!options || !options.element) {
             this._defaultMarker = true;
@@ -516,7 +523,7 @@ export class Marker extends Evented {
     _updateOpacity(force: boolean = false) {
         const terrain = this._map.terrain;
         if (!terrain) {
-            if (this._element.style.opacity !== this._opacity[0]) { this._element.style.opacity = this._opacity[0]; }
+            if (this._element.style.opacity !== this._opacity) { this._element.style.opacity = this._opacity; }
             return;
         }
         if (force) {
@@ -538,7 +545,7 @@ export class Marker extends Evented {
 
         const forgiveness = .006;
         if (markerDistance - terrainDistance < forgiveness) {
-            this._element.style.opacity = this._opacity[0];
+            this._element.style.opacity = this._opacity;
             return;
         }
         // If the base is obscured, use the offset to check if the marker's center is obscured.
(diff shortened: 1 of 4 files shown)


neodescis:
Also, these values should really be numbers instead of strings.

HarelM:
In theory, opacity can be 50% instead of 0.5, but I agree, a number would be better from an API perspective.

sbachinin:
Now setOpacity accepts 2 numbers.
If any of them is undefined, a respective state value (_opacity or _opacityWhenCovered) remains unchanged.
If called without arguments, does nothing.

src/ui/marker.ts
```diff
@@ -797,4 +812,20 @@ export class Marker extends Evented {
     getPitchAlignment(): Alignment {
         return this._pitchAlignment;
     }
+
+    /**
+     * Sets the `opacity` property of the marker.
+     * @param opacity - Sets the `opacity` property of the marker.
+     * @returns `this`
+     */
+    setOpacity(opacity?: number, opacityWhenCovered?: number): this {
```
neodescis, line 821: How does someone revert back to the default values? Or is that not something we care about?

Also, it seems a little weird to me that both arguments are optional. Calling `setOpacity()` would do nothing.

sbachinin: I honestly don't know if a user needs to revert to defaults. Do you want it to be implemented?
If yes, then I suppose we need a separate method like resetOpacity or smth. (Since it's not right to have both arguments of setOpacity optional). Is it what you mean?

src/ui/marker.ts
```diff
@@ -128,6 +138,8 @@ export class Marker extends Evented {
     _pitchAlignment: Alignment;
     _rotationAlignment: Alignment;
     _originalTabIndex: string; // original tabindex of _element
+    _opacity: string;
```
HarelM, line 141: I think the options object and the internal representation should both be numbers.
Is there a reason not to?
This will reduce the need to convert from number to string in multiple places.

src/ui/marker.ts
HarelM: If this method can receive a call with no parameters I suggest it to be the reset method.
If we document it in the method call it might be enough. i.e.:
`setOpacity(0.8)` - sets the "normal opacity"
`setOpacity(0.8, 0.3)` - sets the normal and behind terrain
`setOpacity()` - will recet.
It's a hack, but I'm not sure will need to reset once you set it.

HarelM:
Can you also merge from main branch?
I want to see that the new code coverage is working as expected.
There should not be a message saying that 3 lines are missing coverage as you have added tests to cover the code you added.
The missing lines are ending brackets, main is now updated with this fix, but I want to see it actually fixes this issue.

sbachinin:
Done.

Merged from main. After that **npx jest** stopped working for me locally. (I use it because the doc suggests using it to run individual tests).
The only output in the console for me is "Determining test suites to run...".
(Previously I saw results of passed and failed tests).
Managed to run tests using **npm test-unit**, it seems ok.

src/ui/marker.ts
sbachinin: The reason was exactly that saving _opacity as a string slightly reduced the amount of conversions. But just a tiny bit. So yes, perhaps it's better to use consistent types here.

HarelM:
Yes, it's due to the new reported I added.
You should be able to use `npx jest --reporter=default ...`
I should probably update the docs somewhere...

src/ui/marker.ts
```diff
@@ -146,6 +158,8 @@ export class Marker extends Evented {
         this._rotation = options && options.rotation || 0;
         this._rotationAlignment = options && options.rotationAlignment || 'auto';
         this._pitchAlignment = options && options.pitchAlignment && options.pitchAlignment !== 'auto' ?  options.pitchAlignment : this._rotationAlignment;
+        this._opacity = options && options.opacity !== undefined ? options.opacity : 1;
```
HarelM, line 161: Maybe you can use `setOpacity` here instead of duplicating the defaults?

src/ui/marker.ts
HarelM: Since `style.opacity` is string it might be better to have the entire API use strings, including the internal members and the public API to align with CSS definitions.
I don't know, it's not very important, just a though...

src/ui/marker.ts
sbachinin: Done.

`npx jest --reporter=default` doesn't work either. (Same output)

HarelM: Sorry, missed an "s": should be `--reporters=default`

HarelM:
Can you update the build test with the extra size?
Although this is strange that this change adds a significant bundle size...

src/ui/marker.ts
```diff
@@ -146,6 +158,10 @@ export class Marker extends Evented {
         this._rotation = options && options.rotation || 0;
         this._rotationAlignment = options && options.rotationAlignment || 'auto';
         this._pitchAlignment = options && options.pitchAlignment && options.pitchAlignment !== 'auto' ?  options.pitchAlignment : this._rotationAlignment;
+        this.setOpacity(); // set default opacity
+        if (options) {
+            this.setOpacity(options.opacity, options.opacityWhenCovered);
```
HarelM, line 163: is it possible to call `this.setOpacity(options?.opacity, options?.opacityWhenCovered);
And remove this if completely and the above setOpacity to stay with just one call?

src/ui/marker.ts
sbachinin: Agree with this thought. That's why I made them all strings in the beginning )

HarelM: They were not all strings after the setOpacity API change which created the extra convesions.
If you now replace everything to string and remove the `String(...)` I think it will be OK.

src/ui/marker.ts
sbachinin: `if (options)` can be removed of course.
But in one call unlikely.
If only 1 opacity is specified, the second will be set to undefined. Currently omitted stuff is just left as is.
I think it will be counterintuitive to reset omitted opacity to default (when the 2nd opacity is specified).

HarelM: Right...

src/ui/marker.ts
```diff
@@ -797,4 +812,27 @@ export class Marker extends Evented {
     getPitchAlignment(): Alignment {
         return this._pitchAlignment;
     }
+
+    /**
+     * Sets the `opacity` and `opacityWhenCovered` properties of the marker.
```
HarelM, line 817: Can you add a note here about what will happen if none of them is provided?

HarelM:
@sbachinin first and foremost thanks for all the work and the review changes, I know I can by annoying sometimes...
I see in the code coverage report that one of the lines is partially covered.
According to the tests you added I think this is not a correct report - i.e. I think the coverage report tool is not functioning well with the `?` statement.
Can you make sure both options are covered before I submit a bug report (using logging, or however you think is the fastest option)?

CC: @cenfun

HarelM (approved):

sbachinin:
Seems to be a problem with my test. In all test cases marker "is invisible" though I expected it to be visible in one case.
So yes, coverage report is right.
I'll fix the test.

HarelM:
Ok, great! This means that the report is working as expected and found a place that is not covered! Yay!

cenfun:
@HarelM I'm glad to know this.

sbachinin:
In order to make this line fully covered, I need a very delicate scenario when the marker's base is invisible but its center is visible. I'm not sure if it makes sense to test this.

HarelM:
I'll merge this as this feature wasn't introduced by you.

@SnailBones can you please open a PR with a test that covers this scenario?
I believe this is the only way to make sure the behavior is kept in the future.

SnailBones:
> @SnailBones can you please open a PR with a test that covers this scenario? I believe this is the only way to make sure the behavior is kept in the future.

I don't think this is possible with a unit test, since AFAIK we don't have a way to set real terrain in unit tests. Maybe a browser test could work?

HarelM:
I believe you can mock what you need, can't you?

sbachinin:
If you don't mind I have an idea how to cover the case of semi-visible marker.
It feels really hacky to me though.
https://github.com/sbachinin/maplibre-gl-js/commit/f49d0a2d54a39d3b18e946c5c26d362a2459624d

PR #3621, opened 2024-01-25
Move all string to locale class

HarelM:
## Launch Checklist

Related:
- #3052 

This moves all the strings to the same place and also aligns their names.
This is basically a refactoring, but also a breaking change, so it should be part of 4.0.
The changes in public API is for the cooperative gestures configuration when initializing the map.
I have updated the relevant example.


src/ui/map.ts
```diff
@@ -468,7 +468,7 @@ export class Map extends Camera {
     _localIdeographFontFamily: string;
     _validateStyle: boolean;
     _requestManager: RequestManager;
-    _locale: any;
+    _locale: typeof defaultLocale;
```
neodescis, line 471: Would it be better for this to be `Partial<typeof defaultLocale>`? Otherwise you'd have to define every string, even for controls not being used.

HarelM: Sounds good thank!

HarelM:
@neodescis can I merge this? I've changed the type according to your input, although I think the member itself is always full due to the extend initialization, but it's not critical, anything is better than "any" from my point of view.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ### ✨ Features and improvements
 
+- ⚠️ Changed cooperative gesture config and revmoed the string for it in favor of the locale variable, Also changed the terrain enable disable locale key to match the other keys style ([#3621](https://github.com/maplibre/maplibre-gl-js/issues/3621))
```
louwers, line 5:
```suggestion
- ⚠️ Changed cooperative gesture config and removed the string for it in favor of the locale variable, Also changed the terrain enable disable locale key to match the other keys style ([#3621](https://github.com/maplibre/maplibre-gl-js/issues/3621))
```

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ### ✨ Features and improvements
 
+- ⚠️ Changed cooperative gesture config and removed the string for it in favor of the locale variabl., Also changed the terrain enable disable locale key to match the other keys' styles ([#3621](https://github.com/maplibre/maplibre-gl-js/issues/3621))
```
louwers, line 5:
```suggestion
- ⚠️ Changed cooperative gesture config and removed the string for it in favor of the locale variable., Also changed the terrain enable disable locale key to match the other keys' styles ([#3621](https://github.com/maplibre/maplibre-gl-js/issues/3621))
```

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ### ✨ Features and improvements
 
+- ⚠️ Changed cooperative gesture config and removed the string for it in favor of the locale variable., Also changed the terrain enable disable locale key to match the other keys' styles ([#3621](https://github.com/maplibre/maplibre-gl-js/issues/3621))
```
louwers, line 5: Nitpick: but I think this shoulld be two separate points?

Does this need migration instructions, or will people figure it out?

louwers (approved):
I did not try it out myself, but looks good.

CHANGELOG.md
HarelM: I updated the example for full screen which showed how the locale can be updated, I haven't seen a single case of maplibre locale usage anywhere, this is the first time I happen to see how locale is handled in maplibre, so I sincerely doubt it is in use.

HarelM: I'll split this point never the less.

neodescis:
> @neodescis can I merge this? I've changed the type according to your input, although I think the member itself is always full due to the extend initialization, but it's not critical, anything is better than "any" from my point of view.

Sorry, I mostly don't look at these on the weekend unless I have some spare time. Next time I can approve in advance if you like! I think you're right about the extend() initialization though; sorry I missed that.

PR #3645, opened 2024-01-30
Sky Implementation according to spec

prozessor13:
This PR is a successor of https://github.com/maplibre/maplibre-gl-js/pull/1713 which is working with the actual maplibre-codebase. Writing Tests is still an open issue for which i currently do not have any time :/


HarelM:
I've opened the following issue for the spec as I have forgot to add the diff part of the sky to the relevant place in the code.
https://github.com/maplibre/maplibre-style-spec/issues/517

src/shaders/terrain.fragment.glsl
```diff
@@ -1,7 +1,17 @@
...
+uniform float u_fog_blend;
+uniform float u_fog_blend_opacity;
 
 in vec2 v_texture_pos;
+in float v_fog_depth;
 
 void main() {
-    fragColor = texture(u_texture, v_texture_pos);
+    vec4 color = texture2D(u_texture, v_texture_pos);
+    if (v_fog_depth > u_fog_blend) {
```
HarelM, line 11: Do `v_fog_depth` and `u_fog_blend` have default values? What will happen if I don't set the sky?

src/style/sky.ts
```diff
@@ -0,0 +1,120 @@
...
+const properties: Properties<Props> = new Properties({
+    'sky-color': new DataConstantProperty(styleSpec.sky['sky-color'] as StylePropertySpecification),
+    'fog-color': new DataConstantProperty(styleSpec.sky['fog-color'] as StylePropertySpecification),
+    'fog-blend': new DataConstantProperty(styleSpec.sky['fog-blend'] as StylePropertySpecification),
+    'horizon-blend': new DataConstantProperty(styleSpec.sky['horizon-blend'] as StylePropertySpecification),
+});
+
+const TRANSITION_SUFFIX = '-transition';
+
+export class Sky extends Evented {
```
HarelM, line 38: Does it make sense to inherit from `evented`? If so, why?

src/style/sky.ts
```diff
@@ -0,0 +1,120 @@
...
+    _validate(validate: Function, value: unknown) {
+        return emitValidationErrors(this, validate.call(validateStyle, extend({
+            value,
+            // Workaround for https://github.com/mapbox/mapbox-gl-js/issues/2407
+            style: {glyphs: true, sprite: true},
+            styleSpec
+        })));
+    }
+
+    // Currently fog is a very simple implementation, and should only used
```
HarelM, line 107: Use `/**` style comments for methods.

src/style/style.ts
```diff
@@ -1453,6 +1464,28 @@ export class Style extends Evented {
         this.light.updateTransitions(parameters);
     }
 
+    getSky() {
+        return this.sky && this.sky.getSky();
```
HarelM, line 1468: Should sky be serialized somewhere?

src/geo/transform.ts
```diff
@@ -31,10 +31,12 @@ export class Transform {
     rotationMatrix: mat2;
     pixelsToGLUnits: [number, number];
     cameraToCenterDistance: number;
+    cameraToSeaLevelDistance: number;
```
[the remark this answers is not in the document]
prozessor13: Dont think this is necesarry, because recalculated before every render-call

src/ui/map.ts
```diff
@@ -2688,6 +2689,30 @@ export class Map extends Camera {
         return this.style.getLight();
     }
 
+    /**
+     * Loads a Sky onto the map.
```
HarelM, line 2693: I think it should also mention that this will work only when terrain is on and over 60 degrees pitch, right?

test/examples/3d-terrain.html
```diff
@@ -60,7 +60,8 @@
             terrain: {
                 source: 'terrainSource',
                 exaggeration: 1
-            }
+            },
+            sky: {}
```
HarelM, line 64: Does terrain still works as expected with no sky?

src/shaders/terrain.fragment.glsl
prozessor13: v_fog_depth is calculated in vertex-shader
v_fog_blend should have a default invalue in style-spec

src/style/sky.ts
prozessor13: Dont know the reason, i orientated this code from light.ts

src/style/style.ts
prozessor13: getSky already uses sky._transitionable.serialize()

HarelM:
Thanks for putting this together again!!
@acalcutt do you want to open a PR against this branch to add tests instead of this PR: 
- #2974 

@ibesora do we want to wait with this PR for GSoC or can we change plans in case of this being completed before the program starts?

src/ui/map.ts
prozessor13: yep. works only in terrain. The logic is documented in sky.calculateFogBlendOpacity.

HarelM: This is the main public API for sky, we should document its "limitations" here as well.

src/style/sky.ts
HarelM: I see. OK.

test/examples/3d-terrain.html
prozessor13: Yep.

src/geo/transform.ts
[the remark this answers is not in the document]
HarelM: I've seen matrix calculation that are done as part of the `clone` (when setting zoom) so that failing to clone a member has caused issues.
It might not be the case here, IDK...

src/style/sky.ts
prozessor13: fixed!

src/style/style.ts
HarelM: Where is `getSky` being used?
It should be added to serialization of the style I believe, here:
https://github.com/maplibre/maplibre-gl-js/blob/1e7f6b1082f1efe8c729f27df1bf25e5c649b842/src/style/style.ts#L1269

prozessor13: Ah. you are right!

prozessor13: fixed!

acalcutt:
> Thanks for putting this together again!! @acalcutt do you want to open a PR against this branch to add tests instead of this PR:
> 
> * [Sky fog test #2974](https://github.com/maplibre/maplibre-gl-js/pull/2974)


I've never been the best at adding test, but I added a few expect tests in 
https://github.com/maplibre/maplibre-gl-js/pull/3648

HarelM:
@prozessor13 can I delete `sky` branch and leave only `sky2`?

HarelM:
@acalcutt feel free to fix the unit tests as well while you are at it :-)

acalcutt:
I'm not sure I understand those errors

```
Error: TypeError: painter.transform.calculateFogMatrix is not a function

    at drawTerrain (src/render/draw_terrain.ts:88:45)
    at RenderToTexture.renderLayer (src/render/render_to_texture.ts:[18](https://github.com/maplibre/maplibre-gl-js/actions/runs/7742575082/job/21112014472?pr=3645#step:6:19)7:24)
    at Object.<anonymous> (src/render/render_to_texture.test.ts:119:20)
```

isn't it a function?
https://github.com/maplibre/maplibre-gl-js/blob/32ae12bb550ea8ae196dd755f4a3a4b889561a46/src/geo/transform.ts#L724-L736

HarelM:
The tests are mocking stuff,  if the mock doesn't have a function that the test need you'll see this error.
I'll take a look later this week.

ibesora:
> @ibesora do we want to wait with this PR for GSoC or can we change plans in case of this being completed before the program starts?

I don’t think it’s worth waiting for GSoC if this is good to go… but we’ll need to find something else for GSoC. I didn’t realize this was so close to being done

HarelM:
I've merged the tests fix, and merged main branch.
Let's see what the coverage is reporting and decide how to continue.

test/build/min.test.ts
```diff
@@ -36,7 +36,7 @@ describe('test min build', () => {
         const decreaseQuota = 4096;
 
         // feel free to update this value after you've checked that it has changed on purpose :-)
-        const expectedBytes = 773333;
+        const expectedBytes = 777777;
```
```suggestion
        const expectedBytes = 778888;
```

HarelM:
>but we’ll need to find something else for GSoC

Increasing coverage is always an option.
Helping the globe effort might be a good idea too.
Supporting multiple coordinate systems is always high up.
Improving performance.
Ping me on slack if you want to brain storm.

birkskyum:
> > but we’ll need to find something else for GSoC
> 
> Increasing coverage is always an option. Helping the globe effort might be a good idea too. Supporting multiple coordinate systems is always high up. Improving performance. Ping me on slack if you want to brain storm.

https://github.com/maplibre/maplibre-gl-js/issues?q=is%3Aissue+is%3Aopen+sort%3Areactions-%2B1-desc

HarelM:
@prozessor13 I moved the gl logic from sky class to draw_sky method, I'm not sure this is a correct approach, but generally speaking I think gl stuff should not be in the sky.ts file.
I also added some tests and fixes to the serialization and setStyle diff implementation.
I'll try and think about other places which might need fixing.
Feel free to review my changes and let me know if they are OK or not.

src/geo/transform.ts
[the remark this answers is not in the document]
HarelM: This variable is used in the same method and doesn't need to be a member of the class as far as I can tell.
Am I missing anything?

HarelM:
I went over the last coverage report just now, I believe most of the relevant code is covered in tests.
There might be some bugs or things that will need to solve later on, but currently I can't think of any thing from the top of my head.
Waiting for @prozessor13 response for some of the comments here before I merge this.
@acalcutt let me know if you have any inputs on this PR.
Once merged I'll delete the `sky` branch as well as `sky2` branch.

src/geo/transform.ts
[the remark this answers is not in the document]
neodescis: I was just going to ask the same question. Seems this doesn't need to be a member variable.

neodescis (approved):

acalcutt:
I liked this in my test page at https://wifidb.net/demo/sky/terrain.html . In practice I don't know if I would use fog, but it is a cool effect.

One thing is it is a bit unclear that fog color is the second color in the gradient.

Right now I am using a css linear-gradient based approach for sky, which as a nice sunset color look.
```
#map {
	background: linear-gradient(0deg, rgba(36,0,0,1) 45%, rgb(249 209 89) 62%, rgb(81 151 173) 85%, rgb(2 92 179) 95%);
	height: 100%;
	width: 100%;
}
```
image

to get that same effect, you have to set fog to minimal + the color you want as the fog color, like
```
  "sky": {
    "sky-color": "#199EF3",
    "fog-color": "#FFA500",
    "horizon-blend": 0.6,
    "fog-blend": 1
  }
```
image


Its not really an issue, and it is fine, but I just wasn't clear on that until I started testing it.

acalcutt:
I was thinking though, if i was actually using a "fog-blend" less than 1, that orange color would probably be kind of strange, since a orange fog would cover the surface. 

which reminds me, the [style spec](https://maplibre.org/maplibre-style-spec/sky/) recommend a [green color ](https://www.colorhexa.com/00ff00)for fog, which did not look good at all.
image

HarelM:
The green color is an example, the default is white/gray. I guess the example is green so that you'd know what it looks like and how it affects using a "weird" color.
Feel free to open a PR to change the soec to a better example.
If you think the spec needs to change in order to accommodate for a two color gradient for the sky that is unrelated to the fog color, now would the time to do so, the spec can change as long as there is no implementation, once an implementation exists we'll have a hard time changing it.

acalcutt:
I was thinking maybe and addition color, maybe named "horizon-color" could be added.

then maybe it is this code where that would be used in place of 'u_fog_color'
https://github.com/maplibre/maplibre-gl-js/blob/3726ee9ea1d7b10959fe0b3983ade8cf59679341/src/shaders/sky.fragment.glsl#L11

HarelM:
I would suggest to open/continue the discussion in the style-spec repo.

src/render/draw_sky.ts
```diff
@@ -16,7 +19,21 @@ export function drawSky(painter: Painter, sky: Sky) {
     const colorMode = painter.colorModeForRenderPass();
     const program = painter.useProgram('sky');
 
+    const vertexArray = new PosArray();
```
prozessor13, line 22: Hi Harel, i dont think this is a good idea, because with your change on every render call the vertexarray is uploaded to GPU, but one time is sufficient

src/geo/transform.ts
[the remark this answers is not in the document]
prozessor13: Yep. you are right. Technically not needed to be a class member. But please can you change this?, i am currently quite busy in my dayjob.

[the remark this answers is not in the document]
HarelM: Sure, I'll fix that, don't worry.

src/render/draw_sky.ts
HarelM: OK, I'll see if I can fix that, thanks for the info, this is exactly why I wanted you to take a look :-)

HarelM: I've added a cache variable in latest commit, and since this is always hard coded I guess it doesn't need to change? I don't know what would happen when using more than one map though as this seems like a global variable for all map...
I need to test it I guess...

HarelM: Ok, I used the technic used by the other drawing logic of keeping the gl stuff in the bucket, here I used the sky like you did, but I only create those the first time the sky object is sent to rendering, it is very similar to what you did, but moved the gl logic to the draw method, which is a bit better I think, IDK...

HarelM:
@acalcutt what's your take about the horizon color? Should we add something to the sky spec? Do you need my help to push this forward?
I want to move forward with this PR and feature and I think this is the only open issue ATM...

acalcutt:
In my opinion I think it would be nice to separate the fog color from the sky color, because I don't know that they should always be the same. but I also don't know what that would actually look like. I was more wondering what other people thought here.

I'm not sure I have the time right now to submit a style spec PR.

src/render/draw_sky.ts
prozessor13: Hi Harel, i think your code should work fine! What is your intention to create the new class `Mesh`? Do you plan do reuse it with other mesh'es in future?

prozessor13:
The original code from Bas

prozessor13:
And that the fog is drawn over the sky has the intention to create an atmospheric effect for the fog. This is in my opinion completely separated from any (yet not existing) sky gradient.

HarelM:
> and create new features later on?

While I don't mind postponing the implementation of sky gradient, I do want to make sure we define it in the spec, or st least make sure it doesn't break the current definition.
If it can be done as an increment and not as a breaking change, I have no objections to wait with it.

src/render/draw_sky.ts
HarelM: Yes, for the globe branch, I borrowed the implementation from there.

prozessor13:
I simply would suggest an additional sky-gradient property :)

HarelM:
How would the spec be then?
sky: { ...?

prozessor13:
It should be easy to implement a second sky color to create a 2-color gradient, and i think i can find time for this the next weeks.

proposal:

* sky-color: same as current definition
* horizon-color: the second sky color at the horizon, default is sky-color
* horizon-sky-blend: a value between 0 and 1. 0 is the horizon, 1 is map-height / 2
* fog-color: same as current definition
* horizon-fog-blend: same as horizon-blend in current definition
* fog-blend: same as current definition

I am really not sure how to name the parameters. please make better suggestions. For sure this proposal is very limited in features, but may sufficient in most scenarios? What do you think?

wipfli:
Can you explain what `horizon-sky-blend` does?

HarelM:
This is getting interesting :-)
image

HarelM:
@prozessor13 @Pheonor @kubapelc can you take a look at what I did in the `terrain.fragment.glsl`?
I'm not sure I did the right mixing between the colors - I wanted the fog-ground-blend to indicate "how far the fog is seen" and horizon-fog-blend to indicate how to blend the fog color and the horizon color - it should be gradually blended more as you are further away from the horizon, and at the horizon the color should be the horizon's color.

I hope I was able to explain my self...

I've added a very basic test with a terrain that has all zeros to allow seeing the colors of the horizon, sky and fog as clear as possible.

Overall, this can probably be improved, but I think it's already in a good shape and might be ready to be merged to main.
The logic here is fairly simple, it uses all the spec variables that were defined in the discussion and does something with the colors :-)

src/shaders/terrain.fragment.glsl
```diff
@@ -1,7 +1,20 @@
...
 in vec2 v_texture_pos;
+in float v_fog_depth;
 
 void main() {
-    fragColor = texture(u_texture, v_texture_pos);
+    vec4 color = texture2D(u_texture, v_texture_pos);
+    if (v_fog_depth > u_fog_ground_blend) {
+        vec4 fog_horizon_color = mix(u_horizon_color, u_fog_color, 1.0 - u_horizon_fog_blend - v_fog_depth);
+        float a = (v_fog_depth - u_fog_ground_blend) / (1.0 - u_fog_ground_blend);
+        gl_FragColor = mix(color, fog_horizon_color, pow(a * u_fog_ground_blend_opacity, 2.0));
```
kubapelc, line 16: I think `u_fog_ground_blend_opacity` should be outside the `pow`, because if I were to set `u_fog_ground_blend_opacity` to 0.5 I would expect greatest fog opacity to be 0.5, right now it would end up being 0.5^2=0.25.

src/shaders/terrain.vertex.glsl
```diff
@@ -1,15 +1,17 @@
...
+    float ele = get_elevation(a_pos3d.xy);
     float ele_delta = a_pos3d.z == 1.0 ? u_ele_delta : 0.0;
-    v_texture_pos = a_pos3d.xy / extent;
-    gl_Position = u_matrix * vec4(a_pos3d.xy, get_elevation(a_pos3d.xy) - ele_delta, 1.0);
-    v_depth = gl_Position.z / gl_Position.w;
-}
+    v_texture_pos = a_pos3d.xy / 8192.0;
+    gl_Position = u_matrix * vec4(a_pos3d.xy, ele - ele_delta, 1.0);
+    vec4 pos = u_fog_matrix * vec4(a_pos3d.xy, ele, 1.0);
+    v_fog_depth = pos.z / pos.w * 0.5 + 0.5;
```
kubapelc, line 16: Computing depth for fog in this way makes fog dependent on the depth range of the `u_fog_matrix`. Is this the intention? I would intuitively expect fog to be dependent on (unscaled) distance from camera instead. It looks like the depth range isn't a constant, see https://github.com/maplibre/maplibre-gl-js/blob/main/src/geo/transform.ts#L873

I've noticed that `u_fog_matrix` has a constant near plane, but the far plane is still dependent on `farZ`.

I think the current setup would make the fog change its density when pitching the camera, I'm not sure if that would look good (I haven't actually ran the code). But please ignore this comment if the current setup gives good visual results :)

src/shaders/terrain.fragment.glsl
```diff
@@ -1,7 +1,20 @@
...
 
 in vec2 v_texture_pos;
+in float v_fog_depth;
 
 void main() {
-    fragColor = texture(u_texture, v_texture_pos);
+    vec4 color = texture2D(u_texture, v_texture_pos);
+    if (v_fog_depth > u_fog_ground_blend) {
+        vec4 fog_horizon_color = mix(u_horizon_color, u_fog_color, 1.0 - u_horizon_fog_blend - v_fog_depth);
+        float a = (v_fog_depth - u_fog_ground_blend) / (1.0 - u_fog_ground_blend);
```
kubapelc, line 15: `a` should also be clamped, add `float a = max(..., 0)`

src/shaders/terrain.fragment.glsl
```diff
@@ -1,7 +1,20 @@
...
+uniform float u_horizon_fog_blend;
 
 in vec2 v_texture_pos;
+in float v_fog_depth;
 
 void main() {
-    fragColor = texture(u_texture, v_texture_pos);
+    vec4 color = texture2D(u_texture, v_texture_pos);
+    if (v_fog_depth > u_fog_ground_blend) {
+        vec4 fog_horizon_color = mix(u_horizon_color, u_fog_color, 1.0 - u_horizon_fog_blend - v_fog_depth);
```
kubapelc, line 14: The third parameter of the mix can be negative if `u_horizon_fog_blend > 0` and `v_fog_depth` is near 1, which would produce unexpected fog color. If fog should gradually blend its color to horizon color when `v_fog_depth` is in range `u_horizon_fog_blend`..1, then I think this line should be something like:

```glsl
vec4 fog_horizon_color = mix(u_fog_color, u_horizon_color, max((v_fog_depth - u_horizon_fog_blend) / (1.0 - u_fog_ground_blend), 0));
```

(similar to how `a` is computed)

kubapelc:
BTW the usual way to do fog in games is exponential fog, eg. `float fog_opacity = exp(-distance_from_camera * fog_density)`. The fancy name for this is Beer's law.

src/shaders/terrain.fragment.glsl
HarelM: You can't set it, it is calculated base on pitch in the sky class. The more pitch the map is, the more fog you see. 
But it probably needs to be outside.

src/shaders/terrain.vertex.glsl
HarelM: If there's an easier way to show fog that doesn't require an extra matrix, I'm all for it, this code was donated in a previous PR, I'm just trying to get it in...

src/shaders/terrain.fragment.glsl
HarelM: Right, thanks.

HarelM:
Feel free to create a PR against this branch to update the fog shaders and calculation, I'll be happy to update the relevant tests. 
Also if the spec is not a good fit to the relevant parameters let me know and I'll open a PR/proposal to update the spec as well. 

Bottom line, this should be fairly simple and I want to push this in basically.

src/shaders/terrain.vertex.glsl
kubapelc: Based on your other comment, it sounds like `u_fog_ground_blend_opacity` is calculated to specifically counteract the change in fog density due to it being dependent on the far plane. But let's leave this as it is for now, I might try to simplify this in the future.

kubapelc:
Changing the fog to use `exp` should be straightforward, I'll try to do the PR later today.

HarelM:
I've added the following example page to allow playing with the colors, blends etc:
I think it really helps understanding how each parameter is affecting the behavior, and also help me find a bug when updating the sky.
image

HarelM:
I've added the relevant unit test to make sure the bug won't happen in the future.
I also moved a sky test under a sky folder.
@louwers note that merging this to main, which should happen soon might create an issue with the parity render tests, this is expected, I have created a folder named `sky` to hold other sky related tests so you can ignore them until the native parity is completed.

src/render/painter.ts
```diff
@@ -407,6 +408,9 @@ export class Painter {
         this.context.clear({color: options.showOverdrawInspector ? Color.black : Color.transparent, depth: 1});
         this.clearStencil();
 
+        // draw sky first to not overwrite symbols
+        if (this.style.sky) drawSky(this, this.style.sky);
```
Pheonor, line 412: Is it necessary to check this.style.sky ?

src/render/draw_sky.ts
```diff
@@ -0,0 +1,44 @@
...
+        indexArray.emplaceBack(0, 2, 3);
+
+        sky.mesh = new Mesh(
+            context.createVertexBuffer(vertexArray, posAttributes.members),
+            context.createIndexBuffer(indexArray),
+            SegmentVector.simpleSegment(0, 0, vertexArray.length, indexArray.length)
+        );
+    }
+
+    program.draw(context, gl.TRIANGLES, depthMode, stencilMode, colorMode,
```
Pheonor, line 41: Perhaps it could be avoided to call the drawing function when the value of the  sky parameters are set to 0.

src/style/style.ts
```diff
@@ -338,6 +341,8 @@ export class Style extends Evented {
 
         this.light = new Light(this.stylesheet.light);
 
+        if (this.stylesheet.sky) this.setSky(this.stylesheet.sky);
```
Pheonor, line 344: As discuss during the atmosphere PR, the sky is alway set.

src/style/style.ts
```diff
@@ -522,6 +527,10 @@ export class Style extends Evented {
             return true;
         }
 
+        if (this.sky && this.sky.hasTransition()) {
```
Pheonor, line 530: Same here, test against sky is probably not necessary.

src/style/style.ts
```diff
@@ -580,6 +589,7 @@ export class Style extends Evented {
             }
 
             this.light.updateTransitions(parameters);
+            if (this.sky) this.sky.updateTransitions(parameters);
```
Pheonor, line 592: Same here, the sky test is probably not necessary.

src/style/style.ts
```diff
@@ -624,6 +634,7 @@ export class Style extends Evented {
         }
 
         this.light.recalculate(parameters);
+        if (this.sky) this.sky.recalculate(parameters);
```
Pheonor, line 637: Same here, the sky test is propably not necessary.

src/style/style.ts
```diff
@@ -1474,6 +1490,38 @@ export class Style extends Evented {
         this.light.updateTransitions(parameters);
     }
 
+    getSky() {
+        return this.sky?.getSky();
+    }
+
+    setSky(skyOptions?: SkySpecification) {
```
Pheonor, line 1497: I rewrite this method in the Globe Atmosphere PR to be like the Light one with the deep equal check. Perhaps you could use it here ?

Pheonor (commented):
Adding the fog call systematically, even when not used could propably avoided with a simple chech in the painter class.

src/style/style.ts
HarelM: I'll take a look, thanks for the review!

src/style/style.ts
HarelM: Actually, I looked at the code again and sky can be set to null using set sky, so sky is not always available, thus the checks are needed.
It might be that there's a need to define a transparent sky by default or something to avoid all these ifs, IDK, I don't have a good solution right now...

src/style/style.ts
HarelM: See my comment above.

src/style/style.ts
HarelM: Updated. Thanks for the tip, I also found a bug in the implementation.

src/style/style.ts
HarelM: Updated.

src/render/painter.ts
HarelM: Updated to check the stylesheet state.

src/render/draw_sky.ts
HarelM: The entire method can be avoided if the sky is not defined.

HarelM:
This is now ready to be merged. I'll merge it tomorrow if there's no new review changes requirements.

docs/guides/large-data.md
```diff
@@ -63,7 +63,7 @@ Once the data is loaded, to ensure a smooth user experience, it's essential to o
 
 One simple approach is to visualise fewer points. If we are using a GeoJSON source (i.e. not vector tiles), we can use 'clustering' to group nearby points together. This approach reduces the number of features displayed on the map, improving rendering performance and maintaining map readability.
 
-To do this, when we add the data, we can adjust the [cluster options](/maplibre-gl-js/docs/API/type-aliases/SetClusterOptions/). For example:
+To do this, when we add the data, we can adjust the [cluster options](../API/type-aliases/SetClusterOptions.md). For example:
```
coliff, line 66: This URL doesn't look right to me - shouldn't link to markdown files (.md). I think the URL you need is:
https://maplibre.org/maplibre-gl-js/docs/API/type-aliases/SetClusterOptions/

HarelM: No, this link creates a warning in the documentation creation process. I've verified that and thus made this change.

HarelM:
Good luck to us all.
I'm pushing the button.

src/ui/map_tests/map_style.test.ts
```diff
@@ -475,4 +475,44 @@ describe('#getStyle', () => {
         expect(spy).not.toHaveBeenCalled();
     });
 
+    describe('#setSky', () => {
```
chrisgervang, line 478: How should one unset the sky? (revert back to the default white). It seems calling `setSky({})` doesn't do it, and `setSky()` isn't allowed by the typing _but works_.. is the type just wrong?

HarelM: Does setSky(undefined) works?

chrisgervang: `setSky()` and `setSky(undefined)` partially work. It works to toggle the sky off, but then calling setSky with the same object as before doesn't reapply the sky until the sky value is different.

https://github.com/user-attachments/assets/59f6d525-d14a-439f-8d16-581e6896a6d2

HarelM: Feel free to submit a PR to fix this, probably a missing implementation for this case...

Uranbold:
Hello guys. Will that fog-ground-blend work on 2D? I'm using the fog ground but it looks like this. 
Screenshot 2024-09-11 at 10 04 24

HarelM:
Fog was only added to the terrain...

Uranbold:
a Vector map with Fog will look cool. Please consider it :)

HarelM:
Sure will, you should consider taking the time to implement this and open a PR :-)

Issue #3646, opened 2024-01-31
Marker loses opacity on window resize with 3d terrain


sbachinin:
With 3d terrain enabled, resizing a browser window causes the marker to become "semi-transparent", i.e. to attain reduced opacity that has to be applied only when marker is hidden behind the terrain.

It happens not every time but in about 80% cases for me.

https://github.com/maplibre/maplibre-gl-js/assets/11789697/56f50cda-1548-48ae-bcbb-a92708a10b89

**browser**: Chrome 121.0.6167.85

**Expected Behavior**: marker doesn't change opacity.

HarelM:
Can you provide a reproduction for future readers of this issue?

PR #3656, opened 2024-02-03
Fix marker opacity after resize

sbachinin:
Fixes https://github.com/maplibre/maplibre-gl-js/issues/3646

https://github.com/maplibre/maplibre-gl-js/assets/11789697/4a68acdd-a41e-4c36-9088-138eb6cd6604

### Explanation
On resize Marker was updated **before** map._render.
When marker is updated, it reads the depthTexture to figure out the terrainDistance.
At this moment depthTexture was **empty** because it was just re-initialized due to resize. It was populated only later as part of map._render.
So the order of events on each resize was: _resize -> destroy depthTexture -> read it -> populate it_.

My idea is to delay opacity-related code using requestAnimationFrame (because map render is delayed so). It ensures that marker's opacity is updated **after** map render and after drawDepth.

From my understanding, this fix is untestable because ResizeObserver in jest doesn't really work.

### DIFF
--- a/src/ui/marker.ts
+++ b/src/ui/marker.ts
@@ -142,7 +142,6 @@ export class Marker extends Evented {
     _opacity: string;
     _opacityWhenCovered: string;
     _opacityTimeout: ReturnType<typeof setTimeout>;
-    _frameRequest: AbortController;
 
     /**
      * @param options - the options
@@ -597,9 +596,7 @@ export class Marker extends Evented {
 
         DOM.setTransform(this._element, `${anchorTranslate[this._anchor]} translate(${this._pos.x}px, ${this._pos.y}px) ${pitch} ${rotation}`);
 
-        this._frameRequest = new AbortController();
-        browser.frameAsync(this._frameRequest).then(() => { // Run _updateOpacity only after painter.render and drawDepth
-            this._frameRequest = null;
+        browser.frameAsync(new AbortController()).then(() => { // Run _updateOpacity only after painter.render and drawDepth
             this._updateOpacity(e && e.type === 'moveend');
         }).catch(() => {});
     };
--- a/src/ui/marker.ts
+++ b/src/ui/marker.ts
@@ -1,4 +1,5 @@
 import {DOM} from '../util/dom';
+import {browser} from '../util/browser';
 import {LngLat} from '../geo/lng_lat';
 import Point from '@mapbox/point-geometry';
 import {smartWrap} from '../util/smart_wrap';
@@ -141,6 +142,7 @@ export class Marker extends Evented {
     _opacity: string;
     _opacityWhenCovered: string;
     _opacityTimeout: ReturnType<typeof setTimeout>;
+    _frameRequest: AbortController;
 
     /**
      * @param options - the options
@@ -520,7 +522,7 @@ export class Marker extends Evented {
         return this;
     }
 
-    _updateOpacity = (force: boolean = false) => {
+    _updateOpacity(force: boolean = false) {
         const terrain = this._map?.terrain;
         if (!terrain) {
             if (this._element.style.opacity !== this._opacity) { this._element.style.opacity = this._opacity; }
@@ -556,7 +558,7 @@ export class Marker extends Evented {
         // Display at full opacity if center is visible.
         const centerIsInvisible = markerDistanceCenter - terrainDistanceCenter > forgiveness;
         this._element.style.opacity = centerIsInvisible ? this._opacityWhenCovered : this._opacity;
-    };
+    }
 
     _update = (e?: { type: 'move' | 'moveend' | 'terrain' | 'render' }) => {
         if (!this._map) return;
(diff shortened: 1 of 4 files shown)


HarelM:
The explanation seems good, thanks!
Let's fix the existing tests first.
Maybe we can test it as part of integration tests where the tests are ran in the browser, there is a resize test there exactly because of the reason you specified.

sbachinin:
Fixed old tests, added 1 integration test

src/ui/marker.ts
```diff
@@ -520,8 +520,8 @@ export class Marker extends Evented {
         return this;
     }
 
-    _updateOpacity(force: boolean = false) {
-        const terrain = this._map.terrain;
+    _updateOpacity = (force: boolean = false) => {
```
HarelM, line 523: I'm not sure the change here to arrow function is needed, is it?

src/ui/marker.ts
```diff
@@ -594,7 +594,9 @@ export class Marker extends Evented {
         }
 
         DOM.setTransform(this._element, `${anchorTranslate[this._anchor]} translate(${this._pos.x}px, ${this._pos.y}px) ${pitch} ${rotation}`);
-        this._updateOpacity(e && e.type === 'moveend');
+
+        // Run it after painter.render and drawDepth
+        requestAnimationFrame((_) => this._updateOpacity(e && e.type === 'moveend'));
```
HarelM, line 599: I think you can use `frameAsync` from here: https://github.com/maplibre/maplibre-gl-js/blob/2ff14cbd5075644d6bc73b40a9396a043d66b0e2/src/util/browser.ts#L19
This should allow you to mock it more easily I think.
Not mandatory, but I think this is how it is done in other places/tests.

sbachinin:
Right, arrow function for _updateOpacity is not necessary now.

I used frameAsync instead of requestAnimationFrame though honestly I don't see why it's better. It requires much more code. Does _updateOpacity need a way to be aborted?

Also changed my mind about requestAnimationFrame in tests. Mocking it is not necessary (and it no longer helps, after introduction of frameAsync).

src/ui/marker.ts
```diff
@@ -141,6 +142,7 @@ export class Marker extends Evented {
     _opacity: string;
     _opacityWhenCovered: string;
     _opacityTimeout: ReturnType<typeof setTimeout>;
+    _frameRequest: AbortController;
```
HarelM, line 145: Doesn't look like this is used to cancel the frame, I would advise not to add this variable.

test/integration/browser/browser.test.ts
```diff
@@ -277,4 +277,53 @@ describe('Browser tests', () => {
         expect(markerScreenPosition.x).toBeCloseTo(386.5);
         expect(markerScreenPosition.y).toBeCloseTo(378.1);
     }, 20000);
+
+    test('Marker: correct opacity after resize with 3d terrain', async () => {
```
HarelM, line 281: Can you make sure this test fails without the fix?

sbachinin: Yes, I'm sure it fails

sbachinin:
Btw I was able to run **test-integration** only after adding --reporters=default

src/ui/marker.ts
sbachinin: Removed

HarelM: Did you push this change?

HarelM (approved):
THANKS!!

HarelM:
Is there a need to add something to the changelog?

sbachinin:
Other things about Marker that look wrong to me.
Are they serious enough to dig into?

1) Markers are updated even when **outside viewport**.
On each render a new style.transform is set on a marker element. I think this operation is not necessarily cheap, even for elements outside the viewport.
Opacity-related stuff is throttled but still with a large amount of markers that will be a lot of extra work. 
I think it must be fairly simple to fix.

2) Marker gets updated **twice** for each map **resize**. And for each resize event it updates opacity with "force" (=> without throttling).
Double update happens because map emits both move & moveend events on each resize. It looks strange but that's how it works since very old times.
And Marker is subscribed to both move & moveend.
It's possible to ignore moveend when it's caused by resize.

3) Marker._updateOpacity should be throttled **with trailing**. (Instead of [this](https://github.com/maplibre/maplibre-gl-js/blob/main/src/ui/marker.ts#L534)).
Currently there's no guarantee that opacity will be updated in the very end of some "move". In theory everything must be already ok because Marker is subscribed to "moveend" but for some reason it doesn't always work. For example, sometimes it leads to wrong opacity when mouse button isn't released immediately after the end of move. ([Video](https://github.com/maplibre/maplibre-gl-js/assets/11789697/22e8a572-dd21-4ab4-9fe7-5a3308e9f584)).
Also in VERY rare cases opacity is incorrect on page load. (It's full when marker is behind terrain). Could also be related to this throttling but I'm not sure yet.

HarelM:
Seems like you are working a lot with these markers :-)
If you think it's worth fixing feel free to submit a PR, I'll be happy to review any fix you make.
If you don't, you can consider opening an issue, but I don't think someone will bother fixing it.
In any case, your call :-)

PR #3660, opened 2024-02-05
Tweak banner license comment to include @license.

forgotPassword:
UglifyJS is a popular js minifer. To preserve comments (without using regex) it expects (JSDoc-style) multiline comment with the keyword `@license, @preserve` etc.


HarelM (approved):

HarelM:
Thanks!

Issue #3680, opened 2024-02-08
Vector tiles not requested by maplibre-gl 4.0.0 running in file:// URL (works in 3.6.2)


lrowe:
**maplibre-gl-js version**: 4.0.0 (works in 3.6.2)

**browser**: Chrome and Safari (works in Firefox)

### Steps to Trigger Behavior

 1. Place the html below in a file on your filesystem.
 2. Open the html file with either Chrome or Safari from your filesystem as a file:// URL

### Repro demonstration

```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
    <title>Display a map on a webpage</title>
    <link rel="icon" type="image/png" href="data:image/png;base64,iVBORw0KGgo=" />
    <meta name="viewport" content="initial-scale=1,maximum-scale=1,user-scalable=no" />
    <link rel="stylesheet" href="https://unpkg.com/maplibre-gl@4.0.0/dist/maplibre-gl.css" />
    <script src="https://unpkg.com/maplibre-gl@4.0.0/dist/maplibre-g-devl.js"></script>
    <style>
      body {
        margin: 0;
        padding: 0;
      }
      #map {
        position: absolute;
        top: 0;
        bottom: 0;
        width: 100%;
      }
    </style>
  </head>
  <body>
    <div id="map"></div>
    <script>
      var map = new maplibregl.Map({
        container: "map", // container id
        style: "https://demotiles.maplibre.org/style.json", // style URL
        center: [0, 0], // starting position [lng, lat]
        zoom: 1, // starting zoom
      });
      map.on("error", e => console.error("error", e));
      map.on("load", e => console.log("load", e));
    </script>
  </body>
</html>
```

### Expected Behavior

1. Map displays as it does when opened over http:// e.g. https://jsbin.com/waxofufeza/edit, opened as a file:// in Firefox or in any browser when changing maplibre-gl dependency to 3.6.2.
2. Load event logged in console.

### Actual Behavior

1. Map does not display and tiles are requested in network panel of browser developer tools.
2. Load event not logged in console.
3. No error events logged in console.

HarelM:
Strange... Seems like this was introduced as part of pre-release 1 of 4.0, I'm not sure I can think of something that could've caused this...
I did some major changes there, so I'm not sure I'd know where to start looking...

sbachinin:
I believe the problem is in Actor.receive:
[...]

PR #3681, opened 2024-02-09
Fix vector tiles not loading when html is opened via file://

sbachinin:
Fixes https://github.com/maplibre/maplibre-gl-js/issues/3680

### DIFF
--- a/src/util/actor.ts
+++ b/src/util/actor.ts
@@ -129,7 +129,7 @@ export class Actor implements IActor {
     receive(message: {data: MessageData}) {
         const data = message.data;
         const id = data.id;
-        if (data.origin !== location.origin) {
+        if (data.origin !== 'file://' && location.origin !== 'file://' && data.origin !== location.origin) {
             return;
         }
         if (data.targetMapId && this.mapId !== data.targetMapId) {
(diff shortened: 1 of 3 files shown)


HarelM:
Can you please add a changelog item?
I'm not sure this is worth a test, but the coverage report rightfully reports that this change is only partially covered, your call.

Please keep the template next time you open a PR, it has reminders of things that should be done as part of the PR, mainly for me to remember to review.

src/util/actor.ts
```diff
@@ -129,7 +129,7 @@ export class Actor implements IActor {
     receive(message: {data: MessageData}) {
         const data = message.data;
         const id = data.id;
-        if (data.origin !== location.origin) {
+        if (data.origin !== 'file://' && location.origin !== 'file://' && data.origin !== location.origin) {
```
HarelM, line 132: Seems like the test is not covering all options of this condition, also I think this condition can be simplified, can't it?
The check is that both are not file protocol and are not the same, I tend to think it's enough to check one of them for file and not the same, not sure though...
Maybe not, IDK...

sbachinin: I really don't know how this line can be partly covered. There is at least one test case where all parts of the condition are true, therefore the line is fully executed.

This solution works only with both `!== file://` checks

HarelM: Ok, thanks for taking the time to check this!

HarelM (approved):

PR #3700, opened 2024-02-14
Adds RasterTileSource#setUrl method

manuelroth:
This PR adds the `setUrl` function to RasterTileSource:
- Add `setUrl` to RasterTileSource
- Add unit test to confirm tilejson url update
- Changelog entry

### DIFF
--- a/src/source/raster_tile_source.ts
+++ b/src/source/raster_tile_source.ts
@@ -156,14 +156,14 @@ export class RasterTileSource extends Evented implements Source {
      * @param url - A URL to a TileJSON resource. Supported protocols are `http:` and `https:`.
      * @returns `this`
      */
-        setUrl(url: string): this {
-            this.setSourceProperty(() => {
-                this.url = url;
-                this._options.url = url;
-            });
+    setUrl(url: string): this {
+        this.setSourceProperty(() => {
+            this.url = url;
+            this._options.url = url;
+        });
 
-            return this;
-        }
+        return this;
+    }
 
     serialize() {
         return extend({}, this._options);
--- a/src/source/raster_tile_source.ts
+++ b/src/source/raster_tile_source.ts
@@ -150,6 +150,21 @@ export class RasterTileSource extends Evented implements Source {
         return this;
     }
 
+    /**
+     * Sets the source `url` property and re-renders the map.
+     *
+     * @param url - A URL to a TileJSON resource. Supported protocols are `http:` and `https:`.
+     * @returns `this`
+     */
+        setUrl(url: string): this {
+            this.setSourceProperty(() => {
+                this.url = url;
+                this._options.url = url;
+            });
+
+            return this;
+        }
+
     serialize() {
         return extend({}, this._options);
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,11 @@
 ## main
 
 ### ✨ Features and improvements
-- _...Add new stuff here..._
+
+- Add `setUrl` method to RasterTileSource to dynamically update existing TileJSON resource. ([3700](https://github.com/maplibre/maplibre-gl-js/pull/3700))
 
 ### 🐞 Bug fixes
+
 - Fix Marker losing opacity after window resize ([#3656](https://github.com/maplibre/maplibre-gl-js/pull/3656))
 - Fix vector tiles not loading when html is opened via "file://" ([#3681](https://github.com/maplibre/maplibre-gl-js/pull/3681))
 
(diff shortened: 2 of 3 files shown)


src/source/raster_tile_source.ts
```diff
@@ -150,6 +150,21 @@ export class RasterTileSource extends Evented implements Source {
...
 
+    /**
+     * Sets the source `url` property and re-renders the map.
+     *
+     * @param url - A URL to a TileJSON resource. Supported protocols are `http:` and `https:`.
+     * @returns `this`
+     */
+    setUrl(url: string): this {
+        this.setSourceProperty(() => {
+            this.url = url;
```
HarelM, line 161: Comparing this to `tiles` I don't see that the value of `this.tiles` is updated, can you explain the difference?

HarelM:
I've added a minor comment.
Did you test that the tiles are refreshed in case of a URL change?

src/source/raster_tile_source.ts
manuelroth: Yes sure. As I understand the difference between `this.tiles` and `this.url` is that this.tiles contains an array of one or more tile source URLs ([see style spec](https://maplibre.org/maplibre-style-spec/sources/#raster-tiles)) and url contain an URL to a TileJSON resource ([see style spec](https://maplibre.org/maplibre-style-spec/sources/#raster-url)). If the url is specified instead of the tiles array, Maplibre will load the TileJSON and read the tiles property directly from the json. Therefore the tiles property of RasterTileSource is not updated as it will be loaded later on in the [`loadTileJson` function](https://github.com/manuelroth/maplibre-gl-js/blob/743eb0157a384e15c59d5450439cf9a0a8b39ed4/src/source/load_tilejson.ts#L22)

HarelM: I meant that in your code you set this.url, in the setTiles the code doesn't set this.tiles, which I find odd as I would expect a similar/symmetric behavior in both cases...

manuelroth: Ah I see. I think it's actually not necessary to set this.url, because only the private property this._options.url is passed to loadTileJson function. I did it like this because the VectorTileSource does it as well -> https://github.com/manuelroth/maplibre-gl-js/blob/743eb0157a384e15c59d5450439cf9a0a8b39ed4/src/source/vector_tile_source.ts#L174. I can remove it. What do you think?

HarelM: I see, so it is similar to another source, so both options are OK I guess, your call...

HarelM (approved):

src/source/raster_tile_source.ts
manuelroth: Okey, I think I am gonna leave it like this

manuelroth:
I was able to locally test if the tiles are refreshed in case of an URL change using the `setUrl` function. I used the following html page to verify it. It points to the local development server running this branch. Do you know a better way to manually test a PR?

```
<!DOCTYPE html>
<html lang="en">

<head>
  <title>Dynamically update RasterTileSource using setUrl Function</title>
  <meta property="og:description" content="Add a third-party raster source to the map." />
  <meta charset='utf-8'>
  <link rel='stylesheet' href='https://unpkg.com/maplibre-gl@4.0.0/dist/maplibre-gl.css' />
  <script src='http://localhost:9966/dist/maplibre-gl-dev.js'></script>
  <style>
    body {
      margin: 0;
      padding: 0;
    }

    html,
    body,
    #map {
      height: 100%;
    }
  </style>
</head>

<body>
  <div id="map"></div>
  <script>
      style: {
        'version': 8,
        'sources': {
          'raster-tiles': {
            'type': 'raster',
            'url': 'https://tiles.stadiamaps.com/data/stamen_watercolor_v1.json',
            'tileSize': 256,
            'attribution': 'Map tiles by <a target="_blank" href="http://stamen.com">Stamen Design</a>; Hosting by <a href="https://stadiamaps.com/" target="_blank">Stadia Maps</a>. Data &copy; <a href="https://www.openstreetmap.org/about" target="_blank">OpenStreetMap</a> contributors'
          }
        },
        'layers': [{
          'id': 'simple-tiles',
          'type': 'raster',
          'source': 'raster-tiles',
          'minzoom': 0,
          'maxzoom': 22
        }]
      },
      center: [-74.5, 40], // starting position
      zoom: 2 // starting zoom
    });
    map.on('load', () => {
      const source = map.getSource("raster-tiles");
      if (source) {
        source.setUrl("https://tiles.stadiamaps.com/data/terrarium.json");
      }
    });
  </script>
</body>
</html>
```

manuelroth:
@HarelM thank you so much for looking at this and making a release. An open-source contribution couldn't go smoother. I want you to know that it is much appreciated. I hope I get a chance to contribute more in the future. I wish you a nice day.

HarelM:
Thanks for the warm words @manuelroth!
image

wipfli:
Good stuff

Issue #2871, opened 2023-07-16
Marker flies off in the distance when chaning view angles 


jashanbhullar:
**maplibre-gl-js version**: latest/main branch 

**browser**: Mozilla Firefox 114.0 20230605223242 20230605223242 

### Steps to Trigger Behavior

 1.On your local branch `npm start`
 2.Open `3d-Terrain.html` 
 3.Add the following marker 
```javascript 
        const marker = new maplibregl.Marker()
            .setLngLat([11.39067, 47.305513])
            .addTo(map);
```
4. Change the angle so are somewhat looking at horizon
5. The marker flies off to the horizon 
6. Comes back when looking at more perpendicular view angle 

### Link to Demonstration

https://output.jsbin.com/vatekax#11.9/47.28354/11.40553/0/52
https://jsbin.com/vatekax/2/edit?html,output

### Expected Behavior
The marker should not fly off but should remain at the same place

### Actual Behavior
The marker flies off to some random distance 

[marker-flies-off-low.webm](https://github.com/maplibre/maplibre-gl-js/assets/20891087/dcaaf7cd-2b02-4e74-a200-2e44515b1814)

Found when working on this: https://github.com/maplibre/maplibre-gl-js/issues/1134 
Might be related

tbo47:
I can not reproduce with Firefox 115.3.0esr (64-bit). Is it still a bug?

HarelM:
@sbachinin let me know if you would like to collect the bounty.

sbachinin:
@HarelM, yes

HarelM:
Make sure to follow this guide:
https://maplibre.org/roadmap/step-by-step-bounties-guide/

sbachinin:
https://opencollective.com/maplibre/expenses/191778

HarelM:
THANKS!
FYI @wipfli.

PR #3704, opened 2024-02-15
Fix marker flying off near horizon

sbachinin:
Fixes https://github.com/maplibre/maplibre-gl-js/issues/2871

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.
 - [ ] 

The problem here is smartWrap.
It's not designed for 3d world.
It measures distances between points **based on their screen position**. The goal is to provide smooth transition of a Marker/Popup from central globe to the side globes.
This technique is safe only when there is no terrain and no high pitch/bearing.
Now there are many scenarios in which smartWrap will cause quirky behaviour. (Though they are very unlikely to happen for a sensible user).
https://github.com/maplibre/maplibre-gl-js/issues/2871 is only one of such cases. Another one for [example](https://github.com/maplibre/maplibre-gl-js/assets/11789697/9042dcfb-895d-46c6-919f-194f03af5b47).

All these quirks are reproducible for **Popups too**.

They mostly happen when you "look" rather eastwards or westwards and the marker is (almost) at the same latitude as the camera.

In 3d world smartWrap seems to be conceptually wrong. Sometimes a very distant point (-360 or +360 deg) appears very close (in terms of screen pixels) to a current point due to high pitch or whatever. SmartWrap decides that the marker belongs to that distant globe and assigns a new lng accordingly. This is the reason for the current bug.

Ideal solution would mean a complete rewrite of smartWrap. And it has to be something extra smart. But considering how small the bugs are, it's better to fix some minor mistakes. 

In this PR I suggest 2 small improvements that are simple and just make sense. They also look safe in terms of preserving the existing smartWrap behavior.

Current bug remains in theory possible but after these changes I can't reproduce it at all. Also the 2nd bug from video above disappears too.

What these fixes do.
1) SmartWrap compared an **elevated** "prior position" with **non-elevated** "possible new positions". For this reason priorPos is often considered closer to the "very distant" position than it actually is.
To fix this, I pass a non-elevated prior position to smartWrap. Unfortunately I had to introduce a separate property on a Marker class to save this flatPos.
2) Smart-wrapped position should be applied only if it's below horizon.

[Before and after](https://github.com/maplibre/maplibre-gl-js/assets/11789697/474017b2-ef39-4367-b8ef-c2c21d840729)

As for tests, I have yet no idea how to make it.

BTW existing smartWrap tests were written only for Popup but not for Marker. Don't know if it's a problem.

Let me know if I'm writing too much.

### DIFF
--- a/src/ui/popup.test.ts
+++ b/src/ui/popup.test.ts
@@ -334,9 +334,11 @@ describe('popup', () => {
 
     test('Popup is repositioned at the specified LngLat', () => {
         const map = createMap({width: 1024}); // longitude bounds: [-360, 360]
-
+        map.terrain = {
+            getElevationForLngLatZoom: () => 0
+        } as any;
         const popup = new Popup()
-            .setLngLat([270, 0])
+            .setLngLat([70, 0])
             .setText('Test')
             .addTo(map)
             .setLngLat([0, 0]);
--- a/src/util/smart_wrap.ts
+++ b/src/util/smart_wrap.ts
@@ -17,6 +17,7 @@ import type {Transform} from '../geo/transform';
  * should wrap just enough to avoid doing so.
  */
 export function smartWrap(lngLat: LngLat, priorPos: Point, transform: Transform): LngLat {
+    const originalLngLat = lngLat;
     lngLat = new LngLat(lngLat.lng, lngLat.lat);
 
     // First, try shifting one world in either direction, and see if either is closer to the
@@ -47,5 +48,11 @@ export function smartWrap(lngLat: LngLat, priorPos: Point, transform: Transform)
         }
     }
 
-    return lngLat;
+    // Apply the change only if new coord is below horizon
+    if (lngLat.lng !== originalLngLat.lng &&
+        transform.locationPoint(lngLat).y > (transform.height / 2 - transform.getHorizon())) {
+        return lngLat;
+    }
+
+    return originalLngLat;
 }
--- a/src/util/smart_wrap.ts
+++ b/src/util/smart_wrap.ts
@@ -17,7 +17,7 @@ import type {Transform} from '../geo/transform';
  * should wrap just enough to avoid doing so.
  */
 export function smartWrap(lngLat: LngLat, priorPos: Point, transform: Transform): LngLat {
-    const originalLngLat = lngLat;
+    const originalLngLat = new LngLat(lngLat.lng, lngLat.lat);
     lngLat = new LngLat(lngLat.lng, lngLat.lat);
 
     // First, try shifting one world in either direction, and see if either is closer to the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
+- Fix marker flying off near horizon ([3704](https://github.com/maplibre/maplibre-gl-js/pull/3704))
 - _...Add new stuff here..._
 
 ## 4.0.1
--- a/src/ui/marker.test.ts
+++ b/src/ui/marker.test.ts
@@ -845,7 +845,7 @@ describe('marker', () => {
         const marker = new Marker()
             .setLngLat([0, 0])
             .addTo(map);
-        await sleep(100);
+        await sleep(500);
         expect(marker.getElement().style.opacity).toMatch('1');
         map.remove();
     });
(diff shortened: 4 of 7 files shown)


src/ui/popup.test.ts
```diff
@@ -334,9 +334,11 @@ describe('popup', () => {
 
     test('Popup is repositioned at the specified LngLat', () => {
         const map = createMap({width: 1024}); // longitude bounds: [-360, 360]
-
+        map.terrain = {
```
sbachinin, line 337: This is actually added only for coverage

src/util/smart_wrap.ts
```diff
@@ -47,5 +48,11 @@ export function smartWrap(lngLat: LngLat, priorPos: Point, transform: Transform)
...
     }
 
-    return lngLat;
+    // Apply the change only if new coord is below horizon
+    if (lngLat.lng !== originalLngLat.lng &&
+        transform.locationPoint(lngLat).y > (transform.height / 2 - transform.getHorizon())) {
+        return lngLat;
+    }
+
+    return originalLngLat;
```
HarelM, line 57: Shouldn't this return a cloned object? Seems like other paths return a clone

src/util/smart_wrap.ts
```diff
@@ -47,5 +48,11 @@ export function smartWrap(lngLat: LngLat, priorPos: Point, transform: Transform)
         }
     }
 
-    return lngLat;
+    // Apply the change only if new coord is below horizon
+    if (lngLat.lng !== originalLngLat.lng &&
```
HarelM, line 52: Is this comparison enough?

src/util/smart_wrap.ts
```diff
@@ -17,6 +17,7 @@ import type {Transform} from '../geo/transform';
  * should wrap just enough to avoid doing so.
  */
 export function smartWrap(lngLat: LngLat, priorPos: Point, transform: Transform): LngLat {
+    const originalLngLat = lngLat;
```
HarelM, line 20: I would expect to see unit tests for smaetWrap method that are not relevant to the marker/popup code

src/util/smart_wrap.ts
sbachinin: Done

src/util/smart_wrap.ts
sbachinin: Not sure I understand the question. The whole condition (together with the next line) checks if 1) lngLat has changed and 2) if new lngLat is below horizon.
I think it's enough.

src/util/smart_wrap.ts
sbachinin: SmartWrap was introduced exlusively for popups and markers. And we _can_ test popup/marker behavior that also confirms the proper functioning of smartWrap. So I think it's better to test popup/marker behavior than the util function alone.
I assume it's always better to test something more high-level, more "integration". Just my opinion.

src/util/smart_wrap.ts
HarelM: The comparison is only on lng and not both lng and lat, why?

src/util/smart_wrap.ts
HarelM: In terms of coverage, I prefer to relay on unit tests. When checking a behavior, yes, integration tests are better, but there should be less of them.
In this case for example making sure smartWrap is working by writing test both for marker and popup might be redundant.
https://theqalead.com/topics/testing-pyramid/#:~:text=The%20testing%20pyramid%20is%20a,QA%20specialists%20achieve%20higher%20quality.

src/util/smart_wrap.ts
sbachinin: It's because smartWrap is meant to modify only lng.
No lat change is possible here.

HarelM: Ahh, ok, thanks.

HarelM (approved):

HarelM:
Changelog is missing, otherwise this looks good.

sbachinin:
Shall I add smartWrap tests in this pr?

HarelM:
Your call, you are well familiar with this method now, so adding unit tests to it should be fairly simple for you.
Having said that, this method is fully covered according to codecov, so it might be an overkill.
I leave the decision to you.

sbachinin:
Ok, I'll add smartWrap tests here

HarelM:
THANKS!

PR #3711, opened 2024-02-17
Fix Style.setState ignoring validate flag #3709

### LINKED ISSUE #3709: Style.setState ignores validation options
**4.0.0**:

### Steps to Trigger Behavior

 1. Create a map with `{ validateStyle: false }`
 2. Update the style with `map.setStyle(..., { validate: false })`
 3. The style is validated, despite both settings.

### Link to Demonstration
E.g. something like this should not cause a validation error: 

```lang=js
  container: 'map',
  style: {
    version: 8,
    sources: {},
    layers: []
  },
  center: [12.550343, 55.665957],
  zoom: 8,
  validateStyle: false
});

map.addSource("invalid", { type: "invalid" });

map.setStyle({ ...map.getStyle() }, { validate: false });

```

### Expected Behavior
Style is not validated.

### Actual Behavior
Console error if the style is invalid. The bug appears to be in `style.ts` where the validation option is not checked: https://github.com/maplibre/maplibre-gl-js/blob/3bea9a10acf741caf0913eb3792ef0a268f6fbe3/src/style/style.ts#L662-L667

Would be happy to submit a PR with a fix :thumbsup:

dsshep:
## Launch Checklist

Adds a test for 
- #3709 

Fixes #3709

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### 🐞 Bug fixes
 
-- Fix Style.setState ignoring validate flag.
+- Fix `Style.setState` ignoring validate flag ([#3709](https://github.com/maplibre/maplibre-gl-js/pull/3709))
 - Fix marker flying off near horizon ([3704](https://github.com/maplibre/maplibre-gl-js/pull/3704))
 - _...Add new stuff here..._
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
-- _...Add new stuff here..._
+- Fix Style.setState ignoring validate flag.
 
 ## 4.0.1
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,6 @@
 ### 🐞 Bug fixes
 
 - Fix Style.setState ignoring validate flag.
-
 - Fix marker flying off near horizon ([3704](https://github.com/maplibre/maplibre-gl-js/pull/3704))
 - _...Add new stuff here..._
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,6 @@
 - Fix marker flying off near horizon ([3704](https://github.com/maplibre/maplibre-gl-js/pull/3704))
 - _...Add new stuff here..._
 
-
 ## 4.0.1
 
 ### ✨ Features and improvements
--- a/src/style/style.test.ts
+++ b/src/style/style.test.ts
@@ -883,6 +883,22 @@ describe('Style#setState', () => {
         expect(style.stylesheet.layers[0].id).toBe(initialState.layers[0].id);
         expect(style.stylesheet.layers).toHaveLength(1);
     });
+
+    it.each([[true, false, [false, true], [false, false]]])('ignores validation when validate:false set', async (validateStyle, validateState) => {
+        const style = new Style(getStubMap(), {validate: validateStyle});
+        const styleSpec = createStyleJSON();
+        style.loadJSON(styleSpec);
+
+        await style.once('style.load');
+
+        style.addSource('abc', createSource());
+        const nextState = {...styleSpec};
+        nextState.sources['def'] = {type: 'geojson'} as GeoJSONSourceSpecification;
+
+        const didChange = style.setState(nextState, {validate: validateState} as any); // `StyleSwapOptions` should be `StyleSwapOptions & StyleSetterOptions`
+
+        expect(didChange).toBeTruthy();
+    });
 });
 
 describe('Style#addSource', () => {
--- a/src/style/style.test.ts
+++ b/src/style/style.test.ts
@@ -884,7 +884,7 @@ describe('Style#setState', () => {
         expect(style.stylesheet.layers).toHaveLength(1);
     });
 
-    test('Style#setState ignores validation style when validate false', async () => {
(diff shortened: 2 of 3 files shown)


src/style/style.test.ts
```diff
@@ -883,6 +883,22 @@ describe('Style#setState', () => {
         expect(style.stylesheet.layers[0].id).toBe(initialState.layers[0].id);
         expect(style.stylesheet.layers).toHaveLength(1);
     });
+
+    it.each([[true, false, [false, true], [false, false]]])('ignores validation when validate:false set', async (validateStyle, validateState) => {
```
HarelM, line 887: Can you make this into describe with beforeEach? I find it now readable that each test has a title.

HarelM:
Looks good, thanks!
Added a minor comment.

src/style/style.test.ts
dsshep: Not the prettiest test (also seen there is an error, should be `[[true, false], [false, true], [false, false]]...`). Can you expand on what you mean with the describe and beforeEach?

HarelM: Instead of using `it.each` use 
```ts
describe('setStyle', () => {
    beforeEach(() => {
        // do all the initialization needed here
    });
    test('true false', () => { // but here and below put a better name
    });
    test('false true', () => {
    });
});
```

HarelM (approved):

CHANGELOG.md
```diff
@@ -5,9 +5,13 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
+
+- Fix Style.setState ignoring validate flag.
+
```
HarelM, line 10:
```suggestion
```

CHANGELOG.md
```diff
@@ -5,9 +5,12 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
+
+- Fix Style.setState ignoring validate flag.
 - Fix marker flying off near horizon ([3704](https://github.com/maplibre/maplibre-gl-js/pull/3704))
 - _...Add new stuff here..._
 
+
```
HarelM, line 13:
```suggestion
```

HarelM:
I think the test are not passing... :-/

dsshep:
Couple of issues:

1. async on describe, oops.
2. When `map.setStyle(...)` is called, the options from the map get passed down, including validateStyle flag. However, the test is calling `style.setState` directly.

I've fixed the test, which has simplified it. Ideally we could test the `Map.setStyle` option as well, something like this:

```ts
describe.each([
    [true, {validate: false}],
    [false, undefined]])(
    'Map#setStyle ignores validation style when',
    (validate, setStyleOptions) => {
        let map: Map;
        let nextState: any;

        beforeEach(async () => {
            map = createMap({validate});
            const style = new Style(map);
            const styleSpec = createStyle();
            style.loadJSON(styleSpec);

            await style.once('style.load');

            style.addSource('abc', createStyleSource());
            nextState = {...styleSpec};
            nextState.sources['def'] = {type: 'geojson'} as GeoJSONSourceSpecification;
        });

        test(`validate ${validate}`, () => {
            map.setStyle(nextState, setStyleOptions); // capture validation errors somehow?!

            expect(validationErrors).toHaveLength(0);
        });
    });
```
But I couldn't see any prior art for capturing the validation errors. Any pointers?

HarelM:
If the current test captures the bug, I don't see any reason to complicate it.
There are duplicated tests between map and style, but I prefer simplicity if possible.

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
+
+- Fix Style.setState ignoring validate flag.
```
HarelM, line 9:
```suggestion
- Fix `Style.setState` ignoring validate flag ([#3709](https://github.com/maplibre/maplibre-gl-js/pull/3709))
```

src/style/style.test.ts
```diff
@@ -883,6 +883,22 @@ describe('Style#setState', () => {
         expect(style.stylesheet.layers[0].id).toBe(initialState.layers[0].id);
         expect(style.stylesheet.layers).toHaveLength(1);
     });
+
+    test('Style#setState ignores validation style when validate false', async () => {
```
HarelM, line 887: Shouldn't the test validate that `validateStyle` is not called or that there are no error that are being emitted?
A similar (?) question is basically if this test fails without the fix.

dsshep: `setState` returns false if any validation errors are emitted:
https://github.com/maplibre/maplibre-gl-js/blob/3bea9a10acf741caf0913eb3792ef0a268f6fbe3/src/style/style.ts#L667

Do you think it is worth mocking `validateStyle` instead of depending on the return value? I've push the branch with the change commented out to show it failing again.

HarelM: > setState returns false if any validation errors are emitted

True, but it can also return false for other reasons.
I'm not sure it's worth the extra effort, but it would make the test a bit more correct I believe.
I'm good with either way, your call.

dsshep: I'd be happy to leave as is as the overhead of mocking either `emitValidationErrors` or `validateStyle` seems over the top.

HarelM: Ok, sure, no problem.

HarelM:
Let me know how urgent it is in terms of releasing a version (would a month from now be OK?).

dsshep:
As soon as possible would be handy, thank you for the help and getting it merged 👍

PR #3728, opened 2024-02-20
Overhaul of RTL plugin loading sequence

zhangyiatmicrosoft:
## Launch Checklist

<b>Symptom</b>: we were previously using 3.6.1 and recently tried to upgrade to 4.x, and noticed that map load speed is slowed down by 50ms at percentile 90. I found that it took one or two extra frames when setRTLTextPlugin is called before map creation, based on official instruction: https://maplibre.org/maplibre-gl-js/docs/examples/mapbox-gl-rtl-text/.
This problem appears to be introduced by PR https://github.com/maplibre/maplibre-gl-js/pull/3418

The following diagrams apply when the app calls setRTLTextPlugin for the very first time and "deferred" parameter is true
<b>3.6.1</b>
```mermaid
sequenceDiagram
    participant App
    participant MapLibre Main
    participant MapLibre Worker
    App->>MapLibre Main: setRTLTextPlugin(...) "deferred" is true
    MapLibre Main->>MapLibre Main: sendPluginStateToWorker
    MapLibre Main->>MapLibre Main: Fire pluginStateChange event
    MapLibre Main->>MapLibre Main: (Style.ts) _rtlTextPluginCallback 
    MapLibre Main->>MapLibre Worker: dispatcher.broacast "syncRTLPluginState"<br /> (pluginStatus is "deferred")
    MapLibre Worker->>MapLibre Worker: syncRTLPluginState does nothing (!!!) <br /> because globalRTLTextPlugin.isLoaded() is false.
```
<b>4.0.2 (after PR #3418), notice 2, 3, 4 were not there in 3.6.1</b>
The "broadcasting-reload" cycle is done twice (loading, loaded) when lazyLoadRTLTextPlugin is called and it is downloading the plugin. Only one is enough.
```mermaid
sequenceDiagram
    participant App
    participant MapLibre Main
    participant MapLibre Worker
    App->>MapLibre Main: setRTLTextPlugin(...) "deferred" is true
    MapLibre Main->>MapLibre Main: 1- (rtl_text_plugin_main_thread.ts) <br/> await this._sendPluginStateToWorker
    MapLibre Main->>MapLibre Worker: dispatcher.broacast "syncRTLPluginState"<br /> (pluginStatus is "deferred")
    MapLibre Worker->>MapLibre Worker: 2- _syncRTLPluginState returns false <br /> because state.pluginStatus === 'loaded' is false.
   MapLibre Worker->>MapLibre Main:  message 
    MapLibre Main->>MapLibre Main: 3- (rtl_text_plugin_main_thread.ts) _sendPluginStateToWorker <br />fire event "pluginStateChange"
    MapLibre Main->>MapLibre Main: 4- (style.ts) _rtlTextPluginStateChange <br />reload all vector and geojson sources <br> which means ALL vector tiles !!
```
<b>This PR</b>
State diagram -- add "requested" state, so they are:
'unavailable' | 'deferred' | 'requested' | 'loading' | 'loaded' | 'error'

```mermaid
  graph TD;
      unavailable -- lazyLoadRTLTextPlugin --> requested;
      unavailable -- setRTLTextPlugin --> checkDefer{param<br/>defer?};
      checkDefer-- true --> deferred;
      checkDefer-- false --> loading;
      deferred -- lazyLoadRTLTextPlugin -->loading;
      requested -- setRTLTextPlugin --> loading;
      loading -- To worker and back --> checkSuccess{success?};
      checkSuccess -- true --> loaded;
      loaded --> finish[Fire event RTLPluginLoaded];
      checkSuccess -- false --> error;
```
Cosmetic changes to improve readability/robustness/performance:
- symbol_bucket.ts: once RTL is detect for this bucker, all future detections can be skipped
- index.ts: formatting for reading pleasure
- rtl_text_plugin_status.ts: define event name and message name to be used by other files
- worker.ts to respond with PluginState object instead of boolean, which can be too ambiguous to process
- worker_tile.ts: formatting and commenting for readability
- ajax.ts: I personally found nested tertiary operations are really hard to read and error prone. Made them to simple "if". The minimizer will do its job and end result will be the same anyway.
- Some method names are shortened

 - [x] Link to related issues.


HarelM:
I'm not sure I fully understand the change in this PR.
I also am not fully sure the refactoring I did was good, but it was impossible to understand what went back from and to the worker before my changes, and now I think the code is simpler to read.
I would recommend trying to write the code related to RTL plugin better instead of patching my mistakes.
As far as I could tell, in the worker "loaded" and applyArabicShaping are equivalent as in the past setting loaded was only after this was set.
Again, the code before my changes was unreadable and the fact that you couldn't tell if you are running in the worker or the main thread was causing a headache.
So I would recommend a deeper investigation here/refactoring/design here...

zhangyiatmicrosoft:
> I'm not sure I fully understand the change in this PR. I also am not fully sure the refactoring I did was good, but it was impossible to understand what went back from and to the worker before my changes, and now I think the code is simpler to read. I would recommend trying to write the code related to RTL plugin better instead of patching my mistakes. As far as I could tell, in the worker "loaded" and applyArabicShaping are equivalent as in the past setting loaded was only after this was set. Again, the code before my changes was unreadable and the fact that you couldn't tell if you are running in the worker or the main thread was causing a headache. So I would recommend a deeper investigation here/refactoring/design here...

Understand this is a difficult PR on top of previously unreadable code (100% agree with you!)
Please try to read my before/after trace a little bit, allow me to summarize it:
before -- worker never responed.
after -- worker respond with false and "pluginStateChange" event is fired and result in all tiles reload one extra round
fix -- fire "pluginStateChange" **only** when at least one plugin state is actually changed.

zhangyiatmicrosoft:
> > I'm not sure I fully understand the change in this PR. I also am not fully sure the refactoring I did was good, but it was impossible to understand what went back from and to the worker before my changes, and now I think the code is simpler to read. I would recommend trying to write the code related to RTL plugin better instead of patching my mistakes. As far as I could tell, in the worker "loaded" and applyArabicShaping are equivalent as in the past setting loaded was only after this was set. Again, the code before my changes was unreadable and the fact that you couldn't tell if you are running in the worker or the main thread was causing a headache. So I would recommend a deeper investigation here/refactoring/design here...
> 
> Understand this is a difficult PR on top of previously unreadable code (100% agree with you!) Please try to read my before/after trace a little bit, allow me to summarize it: before -- worker never responded. after -- worker responds with false and "pluginStateChange" event is fired, and results in all tiles reload one extra round fix -- fire "pluginStateChange" **only** when at least one plugin state is actually changed.

While I agree the ultimate fix is to take a deeper look at RTL plugin and its loading logic, that efforts can take many weeks which is almost impossible  ...

HarelM:
I see. Thanks for the info!
Any chance you could change the "before" and "after" in the initial post to use [Github's Mermain markdown](https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/)
I think it will help me understand the change better.
If you could create 3 charts for the following:
1. 3.6.1, 
2. 4.0 
3. Your proposed change 
 
It would be super!

zhangyiatmicrosoft:
> I see. Thanks for the info! Any chance you could change the "before" and "after" in the initial post to use [Github's Mermain markdown](https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/) I think it will help me understand the change better. If you could create 3 charts for the following:
> 
> 1. 3.6.1,
> 2. 4.0
> 3. Your proposed change
> 
> It would be super!

Very interesting tool --- did not know it exists --- quite helpful and now it is clear that my previously proposed change can be optimized further. I think 3.6.1 and old were confusing as hell because at certain times Mapbox was probably running these on both main and worker, correct? 
Now if we only running in main/work separated, a simpler and more intuitive solution can be made. Can you confirm?

HarelM:
Yes, I agree, the original code was running in both main thread and in the worker, I have split the code to improve readability and maintainability.
Both options are valid from my point of view:
1. Don't send sync state to worker if plugin load is deferred
2. Don't raise an event of state changed if the state didn't change

Which ever you find more intuitive for you.

zhangyiatmicrosoft:
> Yes, I agree, the original code was running in both main thread and in the worker, I have split the code to improve readability and maintainability. Both options are valid from my point of view:
> 
> 1. Don't send sync state to worker if plugin load is deferred
> 2. Don't raise an event of state changed if the state didn't change
> 
> Which ever you find more intuitive for you.

Proposed new design --- have to use 3 charts as it is impossible to draw on one, and only the 3rd charts involves worker.
The key is that we have to solve the race condition between Tile (calling lazyLoadRTLTextPlugin) and the hosting app (calling setRTLTextPlugin), and ensure whichever comes in first the end result is the same.
Previously both 3.6.1 and 4.0.2 were solving that problem by keep calling _sendPluginStateToWorker, which fires <code>        pluginStateChange</code> event, at least 2 and up to 3 times; and that was a performance problem

HarelM:
Can you draw a state machine chart with the new/updated states?
Currently there's a mix between states and method calls.
Are you suggesting to remove the loading state and add a different one called RTLRequested?

zhangyiatmicrosoft:
> Can you draw a state machine chart with the new/updated states? Currently there's a mix between states and method calls. Are you suggesting to remove the loading state and add a different one called RTLRequested?

cleaned up and merged to just two charts; hope it makes better sense now

src/source/worker.ts
```diff
@@ -177,7 +177,12 @@ export default class Worker {
     private async _syncRTLPluginState(map: string, state: PluginState): Promise<boolean> {
         rtlWorkerPlugin.setState(state);
         const pluginURL = rtlWorkerPlugin.getPluginURL();
-        if (state.pluginStatus === 'loaded' && !rtlWorkerPlugin.isParsed() && pluginURL != null) {
+
+        const isLoaded =
+            state.pluginStatus === 'loaded' || // Main Thread: loaded if the completion callback returned successfully
+            rtlWorkerPlugin.applyArabicShaping != null; // Web-worker: loaded if the plugin functions have been compiled
```
HarelM, line 183: I believe `applyArabicShaping != null` is equivalent to `isParsed()`, so I tend to think this code is incorrect...?

HarelM:
Thanks for the updated diagram!
A few comments on it:
I'm not sure I fully support sending the event only when the results from the worker arrive.
the event name is `pluginStateChange` and I tend to think it should be fired on every state change, at least that would make sense to me, it might be worth (if relevant) to check that we don't fire a state change if the previous state and the current one are the same, but I'm not sure it's possible right now.
As far as I understand the code and your diagram, `loaded` state can only be achieved when the script is fetch successfully in the main thread and loaded in the worker thread, so this probably is the right time to repaint the map - so I would consider adding this logic to `style.ts`.
I don't oppose to adding the `requested` state, it makes sense, but I think you can solve your race condition by simply calling `setRTLTextPlugin` before initializing the map object like it is done here:
https://maplibre.org/maplibre-gl-js/docs/examples/mapbox-gl-rtl-text/
In any case, I think we have a good solution in general, feel free to continue with implementation and tests.

src/source/worker.ts
zhangyiatmicrosoft: correct...I was trying to mimic old code and later also realized that the old is too messy to begin with. Will fix it

zhangyiatmicrosoft:
> Thanks for the updated diagram! A few comments on it: I'm not sure I fully support sending the event only when the results from the worker arrive. the event name is `pluginStateChange` and I tend to think it should be fired on every state change, at least that would make sense to me, it might be worth (if relevant) to check that we don't fire a state change if the previous state and the current one are the same, but I'm not sure it's possible right now. As far as I understand the code and your diagram, `loaded` state can only be achieved when the script is fetch successfully in the main thread and loaded in the worker thread, so this probably is the right time to repaint the map - so I would consider adding this logic to `style.ts`. I don't oppose to adding the `requested` state, it makes sense, but I think you can solve your race condition by simply calling `setRTLTextPlugin` before initializing the map object like it is done here: https://maplibre.org/maplibre-gl-js/docs/examples/mapbox-gl-rtl-text/ In any case, I think we have a good solution in general, feel free to continue with implementation and tests.

- regarding 'pluginStateChange', yes the event name implies it is intended for every state change. However I feel this name was too generic to begin with. the only listener is in style ts to reload all tiles. And come to think of it, reloading tiles for state change from unavailable to deferred makes no sense and it is a big waste. I am going to change that event to be very specific 'RTLpluginLoaded'
- Not sure what do you mean by "adding this logic to style.ts.": continue from the point above, it is already there. <b>note</b> main thread does not need to fetch the script, only worker needs to import. This was another bug in existing code.
- Yes, setting setRTLTextPlugin before init will make everything work, which is what we were doing but that was the initial problem as illustrated in 4.0.2 diagram --- compared to 3.6.1, it triggers 1 or 2 extra frames and delays the map loading.

HarelM:
Go ahead. Ping me when you set it to ready for review as GitHub doesn't notify me on this change.

zhangyiatmicrosoft:
> @zhangyiatmicrosoft any updates on this? Do you need any help?

Sorry for the delay --- took a lot of testing and debugging to make sure I did not mess up anything. Ready now.

src/source/rtl_text_plugin_status.ts
```diff
@@ -5,18 +5,32 @@
...
 export type PluginState = {
     pluginStatus: RTLPluginStatus;
     pluginURL: string;
+
+    /** Optional error object that carries error from worker to main thread */
+    error?: Error;
 };
+
+export const RTLPluginLoadedEventName = 'RTLPluginLoaded';
+export const SyncRTLPluginStateMessageName = 'syncRTLPluginState';
```
HarelM, line 36: I prefer to keep all the messages names in the same file.
I believe the definitions are good enough to protect against "invalid" changes.

src/source/rtl_text_plugin_main_thread.test.ts
```diff
@@ -3,12 +3,13 @@ import {rtlMainThreadPluginFactory} from './rtl_text_plugin_main_thread';
 import {sleep} from '../util/test/util';
 import {browser} from '../util/browser';
 import {Dispatcher} from '../util/dispatcher';
-
+import {RTLPluginStatus, PluginState, SyncRTLPluginStateMessageName} from './rtl_text_plugin_status';
```
HarelM, line 6:
```suggestion
import {PluginState, SyncRTLPluginStateMessageName} from './rtl_text_plugin_status';
```

src/source/rtl_text_plugin_main_thread.ts
```diff
@@ -1,60 +1,93 @@
...
-            this.pluginStatus = 'loading';
-            await this._sendPluginStateToWorker();
-            await getArrayBuffer({url: this.pluginURL}, new AbortController());
-            this.pluginStatus = 'loaded';
-        } catch {
-            this.pluginStatus = 'error';
+    }
+
+    /** Download RTL plugin by sending a message to worker and process its response */
+    async _download() : Promise<void> {
```
HarelM, line 61: This doesn't download anything anymore.
I'm wondering what was the original idea behind downloading the plugin code.
In any case, I think a better name to this method is a good idea.

src/source/rtl_text_plugin_main_thread.ts
```diff
@@ -1,60 +1,93 @@
...
+                this.fire(new Event(RTLPluginLoadedEventName));
+            } else {
+                // failed scenario: returned, but bad status
+                if (workerResult.error) {
+                    throw workerResult.error;
+                } else {
+                    throw new Error(`worker failed to load ${this.url}`);
+                }
+            }
+        } else {
```
HarelM, line 78: I don't think this edge case is possible, if the worker doesn't respond the callback will not get hit and the promise will never be resolved.
I think this else is redundant...

src/source/rtl_text_plugin_status.ts
```diff
@@ -5,18 +5,32 @@
...
 
 /**
  * The RTL plugin state
  */
 export type PluginState = {
     pluginStatus: RTLPluginStatus;
     pluginURL: string;
+
+    /** Optional error object that carries error from worker to main thread */
+    error?: Error;
```
HarelM, line 32: Is this really needed? I think the actor throws when there's an error, I don't think the error should be part of the payload...

src/source/rtl_text_plugin_main_thread.test.ts
```diff
@@ -17,6 +18,32 @@ describe('RTLMainThreadPlugin', () => {
         broadcastSpy = jest.spyOn(Dispatcher.prototype, 'broadcast').mockImplementation(() => { return Promise.resolve({} as any); });
     });
 
+    function broadcastMockSuccess(message: string, payload: PluginState): Promise<PluginState[]> {
```
HarelM, line 21: `message: string` can be better typed since this is an actor message and the messages are "strict".

src/source/rtl_text_plugin_main_thread.ts
HarelM: I've done some digging to understand why a download was done on the main thread first and only then send the message to the worker and the reason is that it will allow the browser to cache the response and avoid a race condition in the workers.
The following are the relevant PRs that change this.
The original PR was sending the script to the worker as a blob, but caused a bug in the csp version:
- https://github.com/mapbox/mapbox-gl-js/pull/8865
The next PR removed the blob part, but kept the download of the plugin:
- https://github.com/mapbox/mapbox-gl-js/pull/9122

So I guess the name show stay download and bring back the getArrayBuffer part, it's worth adding a comment above the getArrayBuffer to say why it is being downloaded and ignored.

HarelM:
I think I wend through most/all of the code, I've added my comments.

src/source/rtl_text_plugin_status.ts
zhangyiatmicrosoft: which file do these belong to?
"rtl_text_plugin_main_thread.test.ts" or "actor_messages.ts"
Noice one is event name and the other is actor message.
And what do you think about the idea of moving RTLPluginStatus to const enum, instead of just strings (see my comment from last week)

src/source/rtl_text_plugin_main_thread.ts
zhangyiatmicrosoft: After looking at 9122, I still don't understand why it was done this way. Maybe back then there was a race somewhere. 

Anyway, here is what we have now --- Worker thread importScript is a synchronous API, and because of the messaging system promises already done in 4.0.2, there is no race problem anymore. 
It is further guaranteed by the mini state machine in this PR and reducing chatting between main/worker.

In addition by checking Chrome Network tab., <code>importScript</code> is a "javascript" request under the hood
So there is no CSP issues what-so-ever since we are not using blobs anywhere.

I think it is very safe and helps perf to eliminate the redundant getArrayBuffer in main thread.

zhangyiatmicrosoft:
> I think I wend through most/all of the code, I've added my comments.

@HarelM resolved all but one that needs more clarification. Please take a look thanks

src/source/rtl_text_plugin_status.ts
HarelM: The actor message.

src/source/rtl_text_plugin_main_thread.ts
```diff
@@ -68,17 +68,8 @@ class RTLMainThreadPlugin extends Evented {
             if (workerResult.pluginStatus === 'loaded') {
                 this.fire(new Event(RTLPluginLoadedEventName));
             } else {
-                // failed scenario: returned, but bad status
-                if (workerResult.error) {
-                    throw workerResult.error;
-                } else {
-                    throw new Error(`worker failed to load ${this.url}`);
-                }
+                throw new Error(`worker failed to load ${this.url}, worker status=${workerResult.pluginStatus}`);
```
HarelM, line 71: This only takes care of the first response, I would consider checking all the results if one of them is not loaded I guess. Or better yet, throw an error in the worker and simply try-catch here, the infrastructure around this should work.

src/source/worker.ts
```diff
@@ -183,17 +183,27 @@ export default class Worker {
...
-            const complete = rtlWorkerPlugin.isParsed();
-            if (complete) {
-                resultState.pluginStatus = 'loaded';
-                resultState.pluginURL = incomingState.pluginURL;
-                rtlWorkerPlugin.setState(resultState);
-            }
-        } else {
+        // Parsed plugin cannot be changed;
+        // and cannot do anything if url is blank so both cases just return its current state.
+        if (rtlWorkerPlugin.isParsed() || !incomingState.pluginURL) {
```
HarelM, line 188: I think `incomingState.pluginURL` is always defined as it is being checked when calling `setRTLTextPlugin`, right? So no need to check for it here...?

src/source/rtl_text_plugin_main_thread.ts
```diff
@@ -48,17 +48,17 @@ class RTLMainThreadPlugin extends Evented {
 
             } else {
                 // immediate download
-                return this._download();
+                return this._requestImport();
```
HarelM, line 51: The comment above this method is outdated, I would consider removing it.

src/source/rtl_text_plugin_main_thread.ts
```diff
@@ -87,7 +78,7 @@ class RTLMainThreadPlugin extends Evented {
         if (this.status === 'unavailable') {
             this.status = 'requested';
         } else if (this.status === 'deferred') {
-            this._download();
+            this._requestImport();
```
HarelM, line 81: Can lazy load be called twice? I'm asking since I don't see handling of status requested in this `lazyLoad` method.

src/source/worker.ts
```diff
@@ -183,17 +183,27 @@ export default class Worker {
...
+
+        if (incomingState.pluginStatus !== 'loading') {
             // simply sync and done
             rtlWorkerPlugin.setState(resultState);
+            return resultState;
+        }
+
+        this.self.importScripts(incomingState.pluginURL);
+        const complete = rtlWorkerPlugin.isParsed();
+        if (complete) {
```
HarelM, line 203: How about throwing an error in case complete is false?
See my other comment in the trl main thread code.

src/source/rtl_text_plugin_main_thread.test.ts
```diff
@@ -18,7 +18,7 @@ describe('RTLMainThreadPlugin', () => {
         broadcastSpy = jest.spyOn(Dispatcher.prototype, 'broadcast').mockImplementation(() => { return Promise.resolve({} as any); });
     });
 
-    function broadcastMockSuccess(message: string, payload: PluginState): Promise<PluginState[]> {
+    function broadcastMockSuccess(message: typeof SyncRTLPluginStateMessageName, payload: PluginState): Promise<PluginState[]> {
```
HarelM, line 21: I meant something like actor_message's `MessageType`.

src/source/rtl_text_plugin_main_thread.ts
```diff
@@ -1,60 +1,84 @@
...
+                // fire and forget: in this case it does not need wait for the broadcasting result
+                // it is important to sync the deferred status once because
+                // symbol_bucket will be checking it in worker
+                this._syncState(this.status);
 
-    async _downloadRTLTextPlugin() {
-        if (this.pluginStatus !== 'deferred' || !this.pluginURL) {
-            throw new Error('rtl-text-plugin cannot be downloaded unless a pluginURL is specified');
+            } else {
+                // immediate download
```
HarelM, line 50: The comment is outdated, I would consider removing it.

src/source/rtl_text_plugin_main_thread.ts
```diff
@@ -1,60 +1,84 @@
...
+
+    /** Send a message to worker which will import the RTL plugin script */
+    async _requestImport() : Promise<void> {
+        const workerResults = await this._syncState('loading');
+        if (workerResults.length > 0) {
+            const workerResult: PluginState = workerResults[0];
+            this.status = workerResult.pluginStatus;
+
+            // expect worker to return 'loaded'
+            if (workerResult.pluginStatus === 'loaded') {
```
HarelM, line 68: This only takes care of the first response, I would consider checking all the results if one of them is not loaded I guess. Or better yet, throw an error in the worker and simply try-catch here, the infrastructure around this should work.

src/source/rtl_text_plugin_main_thread.ts
```diff
@@ -1,60 +1,84 @@
...
+            }
         }
-        await this._sendPluginStateToWorker();
     }
 
-    async lazyLoadRTLTextPlugin() {
-        if (this.pluginStatus === 'deferred') {
-            await this._downloadRTLTextPlugin();
+    /** Start a lazy loading process of RTL plugin */
+    lazyLoad(): void {
```
HarelM, line 77: Can lazyLoad be called twice? I'm asking since I don't see handling of status requested in this method.

zhangyiatmicrosoft: yes it can, and existing code already handles it.
First call goes to _requestImport -> this._syncState('loading') -> now the status is loading, and all future lazyLoad() will  be no-op

src/source/rtl_text_plugin_main_thread.ts
zhangyiatmicrosoft: Excellent point!
Throughout extensive research we concluded that 1 worker has the best perf overall so I have completely forgotten the case of multiple workers. :)
Added code to load in main if more than one to have the best of both cases.

src/source/rtl_text_plugin_main_thread.ts
zhangyiatmicrosoft: Excellent point. Fixed and added test case

zhangyiatmicrosoft:
> Sorry for nitpicking, but since we are rewriting it I prefer to give it extra TLC as I guess this code won't be touched in the future.

no prob at all. Deeply appreciate the thorough review --- only through which we can archive the best code quality!

src/source/rtl_text_plugin_status.ts
HarelM: This is still not resolved.

src/util/actor_messages.ts
```diff
@@ -80,6 +80,9 @@ export type GetGlyphsResponse = {
  */
 export type GetImagesResponse = {[_: string]: StyleImage}
 
+/** the message name to sync RTL Plugin state. Shared by main thread and worker */
+export const SyncRTLPluginStateMessageName = 'syncRTLPluginState';
```
HarelM, line 84: Either do that for all other messages or remove this line.
Since typescript and the IDE are auto-completing I would recommend simply removing this line.

src/source/worker.ts
```diff
@@ -181,18 +186,34 @@ export default class Worker {
...
-            if (complete) {
-                return complete;
-            }
-            throw new Error(`RTL Text Plugin failed to import scripts from ${pluginURL}`);
+    private async _syncRTLPluginState(mapId: string, incomingState: PluginState): Promise<PluginState> {
+        const resultState = incomingState;
+
+        // Parsed plugin cannot be changed, so just return its current state.
+        if (rtlWorkerPlugin.isParsed()) {
+            return {
```
HarelM, line 194: Since there's a `setState` you might want to simply add `getState` for symmetry and return that, reducing the need to know how the state is structured internally.

src/source/worker.ts
```diff
@@ -17,8 +17,13 @@ import type {
 
 import type {WorkerGlobalScopeInterface} from '../util/web_worker';
 import type {LayerSpecification} from '@maplibre/maplibre-gl-style-spec';
-import type {PluginState} from './rtl_text_plugin_status';
-import type {ClusterIDAndSource, GetClusterLeavesParams, RemoveSourceParams, UpdateLayersParamaeters} from '../util/actor_messages';
+import {
```
HarelM, line 20: Considering my other comment, I think this change won't be needed.

src/source/rtl_text_plugin_main_thread.ts
```diff
@@ -1,60 +1,105 @@
...
+
         try {
-            this.pluginStatus = 'loading';
-            await this._sendPluginStateToWorker();
-            await getArrayBuffer({url: this.pluginURL}, new AbortController());
-            this.pluginStatus = 'loaded';
-        } catch {
-            this.pluginStatus = 'error';
+            const workerResults = await this._syncState('loading');
+            if (workerResults.length > 0) {
```
HarelM, line 75: In what case `workerResults.length` won't be greater than 0?

src/source/rtl_text_plugin_main_thread.ts
```diff
@@ -1,60 +1,105 @@
...
+
+                // expect all of them to be 'loaded'
+                const expectedStatus = 'loaded';
+                const failedToLoadWorkers = workerResults.filter((workerResult) => {
+                    return workerResult.pluginStatus !== expectedStatus;
+                });
+
+                if (failedToLoadWorkers.length > 0) {
+                    throw failedToLoadWorkers[0].pluginStatus;
```
HarelM, line 84: Is this a valid scenario? what flow can cause this?

src/source/rtl_text_plugin_main_thread.ts
HarelM: Generally speaking, I would like to simplify the logic here as much as possible.

src/source/rtl_text_plugin_main_thread.ts
```diff
@@ -1,60 +1,105 @@
...
-            throw new Error('rtl-text-plugin cannot be downloaded unless a pluginURL is specified');
+    /** Send a message to worker which will import the RTL plugin script */
+    async _requestImport() : Promise<void> {
+
+        // Reference PR: https://github.com/mapbox/mapbox-gl-js/pull/9122
+        // if have more than 1 workers, it is better to load once in main thread to warm up browser cache
+        // so all workers can use it --- even though the result of getArrayBuffer is not being used here.
+        // Otherwise, just let worker importScript once.
+        if (WorkerPool.workerCount > 1) {
+            await getArrayBuffer({url: this.url}, new AbortController());
```
HarelM, line 70: Nice! The comments here are great.

src/source/rtl_text_plugin_main_thread.ts
```diff
@@ -1,60 +1,105 @@
...
+                    throw failedToLoadWorkers[0].pluginStatus;
+                } else {
+                    // all success
+                    this.status = expectedStatus;
+                    this.fire(new Event(RTLPluginLoadedEventName));
+                }
+            }
+        } catch (e) {
+            this.status = 'error';
+            throw new Error(`worker failed to load ${this.url}, worker status is ${e.toString()}`);
```
HarelM, line 93: I'm not sure the original code threw an error here, are you sure about this?

src/source/rtl_text_plugin_main_thread.ts
HarelM: Thinking about this again, why do we care if there is 1 worker or more? downloading the plugin code is async in terms of main thread in any case, isn't it?
So while the comment about why we download the code is informative, I'm not sure the if is needed, isn't it?

src/source/rtl_text_plugin_main_thread.ts
zhangyiatmicrosoft: original code did not, but I think it is needed because otherwise RTL main thread status is error but no details --- RTL text just won't render well and devs are in the dark with no clue where to start debugging, which is a scenario I myself hate the most. :)

src/source/rtl_text_plugin_main_thread.ts
zhangyiatmicrosoft: It really meant to address "all success" vs "all failure", which can happen if network is bad, or the RTL script is somehow invalid. It would just pick the 1st failure and throw.

src/source/rtl_text_plugin_main_thread.ts
HarelM: The question here is who will receive this error that is being thrown.
If no one is expecting this error then throwing is not a great solution.
I would consider warnOnce or console.error even.

zhangyiatmicrosoft: console.error is good enough; all I wanted to fix was the previous logic of catch and swallow which is usually a debugging nightmare ...

src/source/worker.ts
```diff
@@ -181,18 +185,32 @@ export default class Worker {
...
+                pluginStatus: 'loaded',
+                pluginURL: urlToLoad
+            };
+            rtlWorkerPlugin.setState(loadedState);
+            return loadedState;
         }
-        return false;
+
+        // error case
+        throw new Error(`RTL Text Plugin failed to import scripts from ${urlToLoad}`);
```
HarelM, line 213: I think that in case of an error, the state of the worker is not updated, not sure how important that is, as the main thread state will be updated, but it's worth thinking about I guess...

src/source/rtl_text_plugin_main_thread.ts
HarelM: What I meant is that I think the if is redundant in terms of performance as the download will happen once either using the main thread or the worker thread, so this if doesn't really improve performance (correct me if I'm wrong).
So if it doesn't improve performance, why do we need it?

src/source/rtl_text_plugin_main_thread.ts
```diff
@@ -1,60 +1,102 @@
...
-            await this._sendPluginStateToWorker();
-            await getArrayBuffer({url: this.pluginURL}, new AbortController());
-            this.pluginStatus = 'loaded';
-        } catch {
-            this.pluginStatus = 'error';
+            const workerResults = await this._syncState('loading');
+
+            // expect all of them to be 'loaded'
+            const expectedStatus = 'loaded';
+            const failedToLoadWorkers = workerResults.filter((workerResult) => {
```
HarelM, line 77: Use `some` or `every` instead.

src/source/rtl_text_plugin_main_thread.ts
zhangyiatmicrosoft: I don't have data to prove one way or another ... because the current popular Mapbox RTL plugin is 44kb, which roughly means 40ms for client to download and parse, and the delta of two approaches is too small to measure. 
Not to mention RTL user base is very tiny to begin with and we only use one worker.
See my previous response --- conceptually it could be downloaded more than once but that was still my speculation, which can be wrong. 

Bottom line I don't think the code will break either way, to be on the safe side, sticking with original MB code is better.
To favor less and cleaner code, go back to my original implementation. Suggestions?

HarelM: Just remove the if and keep the download on the main thread.
Later on, if we can prove addind this if improves performance, we can add this if.

src/source/rtl_text_plugin_main_thread.ts
zhangyiatmicrosoft: changed to find so it can see the failed status. Thanks

src/source/worker.ts
zhangyiatmicrosoft: Set the state to error now. Probably does not matter much ...The control has done its job, either the user needs to refresh or devs need to step in and find out why.

src/source/rtl_text_plugin_main_thread.ts
zhangyiatmicrosoft: Very good discussion, and it drove me to another round of exhaustive testing. Although no 100% subjective data from users, it is fairly simple to gather semi-subjective ones from dev machine and test environment. The result is interesting --- main thread download <b>does</b> help workers, but overall, it is bad because main thread is already busy with all other things, and its time is far more precious than all workers combined. Here are the data on existing <b>4.0.2</b>:

<table>
<tr>
<td>main download + 1 worker download</td>
<td> total of 295ms (main 294, worker 1)</td>
</tr>
<tr>
<td> Just 1 worker download</td>
<td> 142ms</td>
</tr>
<tr>
<td>main download + 8 workers download</td>
<td> total of 312ms (main 300, worker 12)</td>
</tr>
<tr>
<td> Just 8 workers download</td>
<td> 150ms</td>
</tr>
</table>

Since there is no more race condition bugs now, I am removing the main download for all scenarios

src/source/rtl_text_plugin_main_thread.ts
```diff
@@ -1,60 +1,93 @@
...
+            const workerResults = await this._syncState('loading');
+
+            // expect all of them to be 'loaded'
+            const expectedStatus = 'loaded';
+            const failedToLoadWorker = workerResults.find((workerResult) => {
+                return workerResult.pluginStatus !== expectedStatus;
+            });
+
+            if (failedToLoadWorker) {
+                throw failedToLoadWorker.pluginStatus;
```
HarelM, line 73: This throws a status object?
Looking at the worker code, the only way to get here is if we are trying to send to the worker a status that is not loading, in this case this status will be returned.
But just before we get the results we sync `loading` status, so I'm not sure how this code path can be reached.

src/source/rtl_text_plugin_main_thread.ts
HarelM: I like data driven discussion! great findings! I can't argue with facts.

HarelM:
Note that there are lint warnings in last commit related to unused imports, should be a few sec to fix I believe.

src/source/rtl_text_plugin_main_thread.ts
zhangyiatmicrosoft: Actually do not need this try-catch at all and should just let promise itself handle it

src/source/rtl_text_plugin_main_thread.ts
zhangyiatmicrosoft: you are right we dont need throw here at all. Promise can handle everything

HarelM: There's still a need to try - catch - setState in main thread I believe.

src/source/rtl_text_plugin_main_thread.ts
```diff
@@ -1,60 +1,85 @@
...
+    async _requestImport() : Promise<void> {
+        const workerResults = await this._syncState('loading');
+
+        // expect all of them to be 'loaded'
+        const expectedStatus = 'loaded';
+        const failedToLoadWorkerExist: boolean = workerResults.some((workerResult) => {
+            return workerResult.pluginStatus !== expectedStatus;
+        });
+
+        if (failedToLoadWorkerExist) {
```
HarelM, line 68: I believe that in a case of error, since you are throwing in the worker, this code won't be reached but instead, you'll need to try-catch the `_syncState` call.
Feel free to verify me.

zhangyiatmicrosoft: For clear error cases (networking, bad script, failed to parse ..), worker will throw an exception. On console it looks like:
<code>Uncaught (in promise) Error: RTL Text Plugin failed to import scripts from https://somehost/foo.js</code>
or
<code>uncaught (in promise) DOMException: Failed to execute 'importScripts' on 'WorkerGlobalScope': The script at 'https://foo/rtl/foo.js' failed to load.</code>

Catching or not catching in <code>_syncState</code> would not make any difference, because it can only set error state and throw again, not swallowing the error. In reality it does not help dev or end users in anyway.

<code>return workerResult.pluginStatus !== expectedStatus;</code> is NOT intended to handle anything above.  It is only looking for anything that is not 'loaded', so just a "catch all" to process anything unforeseeable in messaging and worker, which means for whatever unknown reason, some or all workers failed to reach 'loaded' state.

If we take out this part, it means main will fire RTLPluginLoaded without checking the result from worker, the next dev to debug this code will be left in dark and very confused of that what happened.

HarelM: In any case of a worker error, the worker will throw, but the main thread state won't be error, which I don't think is a good idea.
If there is any problem in the wroker, it will throw, which is great, so this code is dead code, unless you can tell me how to reach it...

I would consider addidng tests that involve both the main thread and the worker code to make sure we can reproduce the scenarios we are talking about, as the current tests mock half the code at every scenario, and the flows we are talking about are more in the integration realm...

zhangyiatmicrosoft: Moved catch to _syncState now. I have not add/modified any test cases --- let's agree on the code logic first.

HarelM: This looks good now. thanks!
You can use `try { ... } catch` instead of throwing inside the `catch` callback as I'm not sure it will throw to the "right place".

zhangyiatmicrosoft: .catch block is correct. Experimented with <code>try/catch</code> and it did not catch anything .

HarelM: That doesn't make any sense, try catch with await and .catch are equivalent.

zhangyiatmicrosoft: > That doesn't make any sense, try catch with await and .catch are equivalent.

syncState is returning a promise, not doing await because:
- "setRTLTextPlugin" is using syncState for 'deferred' status and that is "fire and forget" and no need for await. A minor perf improvement
- "requestImport" is doing await on the promise returned by syncState
Both cases, .catch will capture and everything behaves correctly

HarelM: This isn't critical. Let's finish with this PR...

test/integration/browser/browser.test.ts
```diff
@@ -385,4 +385,21 @@ describe('Browser tests', () => {
...
     }, 20000);
+
+    test('Load map with RTL plugin should throw exception for invalid URL', async () => {
+
+        const rtlPromise = page.evaluate(() => {
+            // console.log('Testing start');
+            return maplibregl.setRTLTextPlugin('badURL', false);
+        });
+
+        await rtlPromise.catch(e => {
```
HarelM, line 396: I think you can use `await expect(promise).rejects...` or something similar.
Also you can simplify the string matching to be less specific.

zhangyiatmicrosoft:
> Note that there are lint warnings in last commit related to unused imports, should be a few sec to fix I believe.

lint problem fixed

test/integration/browser/browser.test.ts
zhangyiatmicrosoft: cool

zhangyiatmicrosoft:
I don't have write access anymore. Can some please merge. Thanks

PR #3734, opened 2024-02-22
Fix memory leak in Worker when map is removed (#3733)

### LINKED ISSUE #3733: map.remove() doesn’t clean up data inside Worker class -> memory leak
**maplibre-gl-js version**: 4.0.2

**browser**: Chrome 122.0.6261.57

DEMO:  https://jsbin.com/moburomadu/edit?html,output

The `Worker` class has several caches (layerIndexes, availableImages, workerSources, demWorkerSources) that use mapId as a key. It looks like they won’t get cleaned up by `map.remove()` (or anything else except terminating the worker), and thus leak memory if you remove and create new maps often (e.g. inside a single-page app).

You can observe this in Chrome DevTools "Memory" tab (memory usage for the worker keeps growing), or directly by looking at the data (select the worker in "Console" tab and print `Object.keys(worker.layerIndexes)`).

I guess the way to fix this would be to send a `removeMap` message to the Worker? I can do a PR for that next...

pasieronen:
Fixes #3733 

 - [X] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/ui/map.test.ts
+++ b/src/ui/map.test.ts
@@ -1313,15 +1313,9 @@ describe('Map', () => {
 
     test('#remove broadcasts removeMap to worker', () => {
         const map = createMap();
-        let dispatchType;
-        const _broadcastSpyOn = jest.spyOn(map.style.dispatcher, 'broadcast')
-            .mockImplementation((type, _data) => {
-                dispatchType = type;
-                return Promise.resolve({} as any);
-            });
+        const _broadcastSpyOn = jest.spyOn(map.style.dispatcher, 'broadcast');
         map.remove();
-        expect(_broadcastSpyOn).toHaveBeenCalled();
-        expect(dispatchType).toBe('removeMap');
+        expect(_broadcastSpyOn).toHaveBeenCalledWith('removeMap', undefined);
     });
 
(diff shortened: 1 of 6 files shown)


HarelM:
Jest tests are using a message bus implementation that simply doesn't use a worker but uses the same "thread".
Also there's a template for the PR which helps making sure everything is added like changelog (which is missing, mapbox packporting) etc.
There are some tests that check all kind of dispatched messages etc I would recommend looking at those.
Bottom line, this needs tests.

src/ui/map.test.ts
```diff
@@ -1311,6 +1311,19 @@ describe('Map', () => {
         canvas.dispatchEvent(new window.Event('webglcontextlost'));
     });
 
+    test('#remove broadcasts removeMap to worker', () => {
+        const map = createMap();
+        let dispatchType;
```
HarelM, line 1316: I think you can simply query the spy parameters instead of defining an extra variable...

HarelM:
Besides the above nit picking, this looks good, thanks!
Once changelog conflicts are resolved, I'll merge this.

HarelM (approved):

PR #3748, opened 2024-02-26
add codecov to ignore, downgrade to version 3

HarelM:
## Launch Checklist

See title.
I hope this will stop dependabot from trying to upgrade codecov action again...


louwers (approved):
Oh, I also need this.

Issue #1559, opened 2022-08-28
Image in raster layer shows only partially on certain zoom levels when 3D terrain is enabled.


danielfaust:
![CEA884BC-D1C4-49DB-9611-8BBC39B1C392](https://user-images.githubusercontent.com/320098/187073663-e3bf2755-52ed-4056-a651-3352a907d728.GIF)

**maplibre-gl-js version**: v2.4.0 (already present in v2.2.0-pre.2)

**browser**: Google Chrome Version 104.0.5112.102 (Offizieller Build) (64-Bit)

No errors or warnings appear in the console

### Steps to Trigger Behavior

 1. Zoom in and out into the image to see the problem, then zoom in to prepare to toggle elevation layer
 2. Toggle elevation layer a couple of times to see the problem.
 
### Link to Demonstration

~https://jsbin.com/qeverim/edit?html,output~

https://jsbin.com/yowixah/1/edit?html,output

### Expected Behavior

The image should always fully render into the layer

### Actual Behavior

The image is cropped at high zooming levels.

danielfaust:
There seems to be a threshold at zoom level 9 and at level 10.

If the zoom level is equal or above 10, then latitudinal and longitudinal clipping occurs.

If the zoom level is equal or above 9 but below 10, then only longitudinal clipping occurs.

If the zoom level is below 9, then no clipping occurs.

So I assume that the threshold equal or above 9 applies the longitudinal clipping and equal or above 10 the latitudinal.

I have no idea where these numbers come from, but they could be a hint.

prozessor13:
Thx for bugreport, i looks like that only the center-coordinate of the image is rendered onto terrain. When looking into *draw_raster* for ImageSource a `boundsBuffer` object is transfered to the shader. But because terrain-tile rendering stops on tile-borders this looks not sufficient. In this case i think a list of terrain-tiles, calculated by the `source.coordinates`, must be rendernd.
Currently i not have the time to look further onto this, anyone else?

HarelM:
I'll see if this is something I can try and tackle, sounds fairly straight forward from your deception.

HarelM:
Unfortunately I'm not making any progress :-(
I've created the following debug page in order to be able to better debug this issue:
Clicking zoom one time shows the issue and allows changing the code using `npm run start-debug` to see the changes in the code take effect on the page.
[...]

PR #3783, opened 2024-03-04
Globe - basic infrastructure, raster layer adaptation for globe

kubapelc:
This is the first PR of the implementation of vector globe view for MapLibre. It is a selection of changes from my main [vector globe branch](https://github.com/windycom/maplibre-gl-js/tree/kubapelc/globe-vector) as well as interface and changes from @pheonor's globe implementation. More PRs will follow with more of the globe rendering implementation. I plan to split PRs into smaller ones for easier reviewing, as requested.

Issue & discussion of globe:

## Additions and changes

- base classes and interface for globe, a unification of both my and @pheonor's forks, so that we can both do PRs independently in the future
- changes to `map.ts` and Map constructor to include projection
- system for projection-dependent vertex shader prelude code
	- such code is injected at the beginning of every vertex shader
	- see `src/shaders/_projection_mercator.vertex.glsl` for an example
	- example: function `projectTile()` projects a point in local tile coordinates (0..EXTENT) to screen
		- for mercator, this is mostly equivalent to a simple matrix multiplication (usually replaces `u_matrix * vec4(a_pos, 0.0, 1.0)` in shaders)
		- for globe, this does a projection from mercator to a 3D point on the surface of a sphere, which is then transformed with a projection matrix
		- additionally, for globe projection, this function automatically handles animated transition from globe to mercator (for high zoom levels)
	- this system should (in theory) allow for simpler implementation of different projections in the future
- `ProjectionBase` class, and its implementations for Mercator and Globe projections
	- note that globe projection uses an instance of mercator projection internally, since globe transitions to mercator at high zoom levels
	- the currently used projection class is exposed in `map.ts` to allow custom layers to implement globe view using shader projection preludes
	- the projection class will likely expand in the future once proper controls and other non rendering-related features are implemented for globe
- adaptation of the raster layer for globe (other layers are still mercator)
 	- raster layer is a relatively simple example of how the new projection classes and projection shader preludes are used
- an example webpage with globe view (`test/examples/globe.html`)
- some minor changes
	- additional comments
	- typo fixes
	- simplify `tilesAfterTime` implementation in `terrain_source_cache.ts`

## Example

A screenshot of a globe with a satellite map

See `test/examples/globe.html` for a web page that displays a satellite map projected onto a globe. Note that only the raster layer is currently adapted for globe projection, so lines and symbols will still display as if on a mercator map.

The projection will automatically transition to mercator at high zoom levels. The transition is animated and happens at a point when globe is already almost indistinguishable from mercator projection, so it is nearly unnoticeable.

## Changes compared to Pheonor's repo

@pheonor please also review this PR, especially the following changes and adaptations from your branch:

- I've excluded the `project` and `unproject` methods from the projection classes for now, as they are not used by anything (yet).
- I've moved the place where projection is stored from `transform.ts` into `map.ts`, as transform might exist in multiple instances, but map should (as far as I know) only ever exist in one single instance.
- The globe example webpage is from your branch, just slightly modified (added a title and description).

## Problem: where to place granularity settings?

Granularity determines how much subdivision is applied to tile geometry. For example, a subdivision granularity of 4 will "split" all geometry in a tile as if it was cut by the edges of a 4x4 grid. Different zoom levels have different granularities. Granularity for a given zoom level is determined by a simple function.

It is relatively clear how to set granularity levels for globe projection, and thus, if MapLibre only ever used globe view, those values could be static. But what if the user wants a map that will always use mercator? Then there is no need for subdivision to slow down tile parsing and rendering. Or what if a different projection in the future needs different granularities?

Subdivision functions are used in `*_bucket.ts` (fill, fill-extrusion and line buckets), right before generating vertex buffers. The bucket functions run in a separate web worker which has no access to the map or transform instance, as far as I know. Furthermore, the granularity doesn't depend *only* on zoom level, but also on what type of geometry is getting subdivided - for example line layers use higher granularity than fill layers.

Currently, granularity settings are static, both in the sense that they cannot change eg. with projection and that they use the `static` keyword to be accessible from anywhere.

I have no idea how to solve this. The data about which tiles need to be loaded (data which ultimately ends up in the bucket) passes through complex machinery and it is hard for me to even navigate through it.

In this PR, subdivision is only used for stencil masks, but following PRs will utilize it for fill, fill extrusion and line geometry as well.

## Failing tests & other issues

As far as I know, about 300 render tests fail in my branch. After consulting with @wipfli I decided to publish the PR anyway, but I will continue debugging the tests and I'll try to find a fix. I don't know about the state of unit tests, I haven't managed to even run those yet. But using this branch of MapLibre for displaying a map in a web browser should work without any issues.

~~On Chrome, you may notice a warning about a GPU buffer guarded by a FenceSync object being read before the fence is signaled. I know about this issue and will attempt to fix it. It is related to globe projection (see class `ProjectionErrorMeasurement` in `src/geo/projection/globe.ts`).~~ This is now fixed.

## Checklist

- [x] Link to related issues.
	- Completed, as long as doc comments are enough.
- [ ] Post benchmark scores.


HarelM:
Thanks a lot for opening this PR!!
It will take me some time to review it as it has a lot of changes.
I'll start with some basic stuff and then later on move to the more in depth, so please bare with me.
Also note that you can use this PR to see the results of the render tests.
If your code is "opt in" then the render tests should not fail, in theory, but everything is possible.
If you have any issue running the tests or understanding what they do, let me know and I'll do my best to help.
It seems that the unit tests passed, so this is a good sign.

src/geo/projection/projection_base.ts
```diff
@@ -0,0 +1,125 @@
...
+import {UnwrappedTileID} from '../../source/tile_id';
+import {Transform} from '../transform';
+import Point from '@mapbox/point-geometry';
+import {ProjectionData} from '../../render/program/projection_program';
+import {PreparedShader} from '../../shaders/shaders';
+
+/**
+ * An abstract class the specializations of which are used internally by MapLibre to handle different projections.
+ */
+export abstract class ProjectionBase {
```
HarelM, line 13: Can this be converted to an interface? I prefer composition over inheritance and also this class seems to have very little implementation.

kubapelc: Yes, this can absolutely be an interface, I don't think ProjectionBase should ever contain any nontrivial logic. Converted to interface in the latest commit.

kubapelc:
I fixed the issue with Chrome warning related to FenceSync and pixel readback.

HarelM:
@kubapelc do you think you can create a PR only for the `Mesh` class addition and changes? (should be like 4-5 files I think).
I want to push only this and try and use this `Mesh` class in the sky maybe?
I can try and do it if you find it complicated...

HarelM:
For now I copied the `mesh.ts` file to the sky2 branch, I hope it won't cause a merge issue once sky is merge...

src/ui/map.ts
```diff
@@ -3324,4 +3362,24 @@ export class Map extends Camera {
...
+
+    /**
+     * Returns the active projection name.
+     * @returns The projection name
+     * @example
+     * ```ts
+     * let projectionName = map.getProjectionName();
+     * ```
+     */
+    getProjectionName(): string { return this.projection.name; }
```
HarelM, line 3384: I'm not sure this is an interesting API since we have `getProjection` and one can get the name from it.

src/ui/map.ts
```diff
@@ -58,9 +58,32 @@ import type {
...
 
+/**
+ * Name of MapLibre's map projection. Can be:
+ *
+ * - `mercator` - A classic Web Mercator 2D map
+ * - 'globe' - A 3D spherical view of the planet when zoomed out, transitioning seamlessly to Web Mercator at high zoom levels.
+ */
+type ProjectionName = 'mercator' | 'globe';
+
+function getProjectionFromName(name: ProjectionName, map: Map): ProjectionBase {
```
HarelM, line 75: To avoid all kind of circular dependencies, I would recommend placing this method in a file of its own, it will probably grow in the future too.

src/ui/map.ts
```diff
@@ -3324,4 +3362,24 @@ export class Map extends Camera {
...
+
+    /**
+     * Returns the active `ProjectionBase` object.
+     * @returns The projection object.
+     * @example
+     * ```ts
+     * let projection = map.getProjection();
+     * ```
+     */
+    getProjection(): ProjectionBase { return this.projection; }
```
HarelM, line 3374: What about changing the projection in terms of API?

src/source/source_cache.ts
```diff
@@ -950,8 +950,23 @@ export class SourceCache extends Evented {
         return tileResults;
     }
 
-    getVisibleCoordinates(symbolLayer?: boolean): Array<OverscaledTileID> {
-        const coords = this.getRenderableIds(symbolLayer).map((id) => this._tiles[id].tileID);
+    getVisibleCoordinates(symbolLayer?: boolean, deduplicateWrapped?: boolean): Array<OverscaledTileID> {
```
HarelM, line 953: It is not clear what `deduplicateWrapped` flag is for, can you elaborate?

HarelM: Probably worth adding a unit test to this case as well as the method is not trivial.

src/source/image_source.ts
```diff
@@ -104,6 +105,7 @@ export class ImageSource extends Evented implements Source {
     _boundsArray: RasterBoundsArray;
     boundsBuffer: VertexBuffer;
     boundsSegments: SegmentVector;
+    tileCoords: Array<Point>;
```
HarelM, line 108: Can you elaborate on this change?
There's this issue that is worth considering if imagesource is touched I guess:
https://github.com/maplibre/maplibre-gl-js/issues/1559

kubapelc:
> For now I copied the `mesh.ts` file to the sky2 branch, I hope it won't cause a merge issue once sky is merge...

I think just copying the class is ok, it should not cause any difficult merge issues. But I can still create the separate PR if you want.

src/source/image_source.ts
kubapelc: Images used to be rendered with a quad that used a vertex buffer with custom positions (covering the rectagle needed by a given ImageSource) and texcoords that were always the corners of the 0..1 range. Regular raster tiles were drawn in the same way, just the vertex positions were always 0..EXTENT range corners.

For globe I needed to draw both raster tiles and custom images with subdivided geometry. I didn't want to generate a new subdivided grid mesh for each custom image, si I did the following:

Every image/raster tile uses a mesh with vertex positions in range 0..EXTENT. I can cache and reuse those meshes for a given level of subdivision, so only about a dozen of those will ever exist. Images may need to have different corner positions than 0,0 or EXTENT,EXTENT but I already need to pass the actual extents of the tile to the shader for globe (see code around [this line](https://github.com/windycom/maplibre-gl-js/blob/3cc71ca392f71d16b5414d3cf82fd7f27679a124/src/render/draw_raster.ts#L104) ), so I can infer the correct corner coordinates from that in the shader.

The line you highlighed exposes the corner coordinates for image sources, so that I can access them during rendering and pass them to the shader.

Now that I know that the mesh will always be a grid in 0..EXTENT, I no longer need the texture coordinates in the vertex buffer, because I can compute them from vertex positions. So I removed the texture coordinates.

I've also noticed just now that the `boundsBuffer` and other vertex-buffer related members of ImageSource are now unused, so I will also remove them.

kubapelc: Also I think this sadly has no relation to the issue you mentioned, that one seems to be specific to terrain rendering.

src/ui/map.ts
kubapelc: I'm not sure whether to allow projection to be changed at runtime. See the section on granularity settings in the PR description. Tl;dr globe needs to subdivide almost all geometry during the loading stage, which slows down loading of new tiles and maybe also slows down rendering a bit due to drawing more polygons.

Right now, this subdivision happens regardless of what projection is used, but I think that it would be a really good idea in the future to allow some projections (mercator mostly) to disable subdivision altogether to boost performance.

Problem is that if you then wanted to switch from mercator to globe, you would need to reload all tiles to apply proper subdivision to them.

If there is a mechanism to force the reload of all tiles (or at least the regeneration of their vertex buffers) in MapLibre already, then runtime projection changing can be implemented with relative ease.

HarelM: I'm not sure there is a mechanism to invalidate all the tiles, but I know this is needed, I've seen plugins that needed it and are using internal APIs related to source cache, which is a great way to break stuff (maplibre-inspect for example).
So generally speaking, I think we need an API to clear tiles...

src/source/source_cache.ts
kubapelc: I've added a comment that should clarify it:

```ts
getVisibleCoordinates(symbolLayer?: boolean, deduplicateWrapped?: boolean): Array<OverscaledTileID> {
        let coords = this.getRenderableIds(symbolLayer).map((id) => this._tiles[id].tileID);

        // When rendering a mercator map, the screen may be larger than the world,
        // causing multiple "copies" of the world to be visible and
        // the same tile might be drawn in two different places (see wrap in OverscaledTileID).
        // For globe, we only ever want to draw a single copy of the world, so no (canonical) tile should be present
        // multiple times in the resulting list of tile IDs.
        if (deduplicateWrapped) {
```

I'm not 100% confident about what *exactly* the method is doing and why `symbolLayer` flag exists to write a proper doc comment for it.

src/ui/map.ts
kubapelc: True, I removed it.

src/ui/map.ts
kubapelc: I've moved it to a separate file.

src/ui/map.ts
Pheonor: When I add this API, I want a symetric API to set projection with just the name and get the projection name we use just before. But I also think that getting the full projection object could be usefull.

src/ui/map.ts
Pheonor: Effectively, it could be useful to change projection at runtime. We just need to reset all cache and render all tiles.

kubapelc: I'd like to go ahead without runtime projection switching at the moment, and add that (plus proper handling of subdivision) in a later PR. My goal is to have this PR merged rather sooner than later, so that I can polish and create PRs of other globe features. The reviewing process for PRs will necessarily have a lot of latency, so I can work on projection switching in the meantime.

src/ui/map.ts
kubapelc: I think that setting a projection at runtime should not be done with a setter, but rather with a function, since it will probably be a quite heavy operation. Something like `setProjectionByName(...)`.

HarelM: Sure, I agree.

HarelM: P.S. here's a meduim article I wrote:
https://medium.com/@harel.mazor/why-i-never-use-get-and-set-in-any-code-language-7a0f7de739ce

src/ui/map.ts
HarelM: Sounds like a good approach.

src/ui/map.ts
Pheonor: Sure, this is exactly I propose into my branch.
```
    setProjection(projectionName: string): Map {
        this.transform.projection = projectionName;

        this.updateRenderToTexture(this.getTerrain());

        this.fire(new Event('projection', {projection: projectionName}));
        return this._update();
    }
```
We could add it later, in another PR.

test/examples/globe.html
```diff
@@ -0,0 +1,28 @@
...
+        body { margin: 0; padding: 0; }
+        html, body, #map { height: 100%; }
+    </style>
+</head>
+<body>
+<div id="map"></div>
+<script>
+    const map = new maplibregl.Map({
+        container: 'map',
+        projection: 'globe',
```
HarelM, line 20: This should be part of the style itself, or if the spec is still not up-to-date, set using a map API call like "setProjection".

src/ui/map.ts
```diff
@@ -318,6 +320,13 @@ export type MapOptions = {
      * You shouldn't set this above WebGl `MAX_TEXTURE_SIZE`. Defaults to [4096, 4096].
      */
     maxCanvasSize?: [number, number];
+    /**
+     * Map projection to use. Options are:
+     * - 'mercator' - The default, a classical flat Web Mercator map.
+     * - 'globe' - A 3D spherical view of the planet when zoomed out, transitioning seamlessly to Web Mercator at high zoom levels.
+     * @defaultValue 'mercator'
+     */
+    projection?: ProjectionName;
```
HarelM, line 329: See my other comments about where this definition should be.

src/source/video_source.ts
```diff
@@ -161,14 +159,6 @@ export class VideoSource extends ImageSource {
         const context = this.map.painter.context;
         const gl = context.gl;
 
-        if (!this.boundsBuffer) {
```
HarelM, line 164: Can you elaborate on why this was removed?

src/source/source_cache.ts
```diff
@@ -950,8 +950,28 @@ export class SourceCache extends Evented {
...
+        let coords = this.getRenderableIds(symbolLayer).map((id) => this._tiles[id].tileID);
+
+        // When rendering a mercator map, the screen may be larger than the world,
+        // causing multiple "copies" of the world to be visible and
+        // the same tile might be drawn in two different places (see wrap in OverscaledTileID).
+        // For globe, we only ever want to draw a single copy of the world, so no (canonical) tile should be present
+        // multiple times in the resulting list of tile IDs.
```
HarelM, line 960: There is the renderWorldCopies flag, sounds like it does similar behavior, can we use it instead of deduplicating things?

src/source/source_cache.ts
HarelM: If this is the case, changing it sounds risky... I would consider trying to better understand what it does before changing it, once you are sure about what it does, writing a comment should be an easy task.

src/source/image_source.ts
HarelM: OK, thanks for the info.

src/shaders/_projection_mercator.vertex.glsl
```diff
@@ -0,0 +1,24 @@
+float projectLineThickness(float tileY) {
```
HarelM, line 1: Why does this file name start with "_"?

src/render/subdivision.ts
```diff
@@ -0,0 +1,52 @@
...
+
+    constructor(fill: SubdivisionGranularityExpression, line: SubdivisionGranularityExpression, stencil: SubdivisionGranularityExpression) {
+        this.granularityFill = fill;
+        this.granularityLine = line;
+        this.granularityStencil = stencil;
+    }
+}
+
+export const granularitySettings: SubdivisionGranularitySetting = new SubdivisionGranularitySetting(
+    new SubdivisionGranularityExpression(7, 1), // Fill
```
HarelM, line 47: These are hard coded, I'm guessing they were chosen empirically, consider allowing these to be changed somehow if this wasn't already done.

src/render/program.ts
```diff
@@ -75,12 +74,15 @@ export class Program<Us extends UniformBindings> {
         if (showOverdrawInspector) {
             defines.push('#define OVERDRAW_INSPECTOR;');
         }
-        if (terrain) {
+        if (hasTerrain) {
             defines.push('#define TERRAIN3D;');
         }
+        if (programDefines) {
```
HarelM, line 80: Isn't this a bit too general? I think being explicit here would be better.
I also saw some `#ifndef GLOBE` that I guess should be added here...?

src/render/painter.ts
```diff
@@ -566,20 +617,42 @@ export class Painter {
...
             this.cache[key] = new Program(
                 this.context,
                 shaders[name],
                 programConfiguration,
                 programUniforms[name],
                 this._showOverdrawInspector,
-                this.style.map.terrain
+                useTerrain,
+                projectionPrelude,
+                concatenatedDefines
```
HarelM, line 655: See my previous comment.

src/render/painter.ts
```diff
@@ -566,20 +617,42 @@ export class Painter {
...
+        const projectionKey = `/${forceSimpleProjection ? MercatorShaderVariantKey : projection.shaderVariantName}`;
+        const concatenatedDefines = defines ? [projectionDefine].concat(defines) : [projectionDefine];
+
         const key = name +
             (programConfiguration ? programConfiguration.cacheKey : '') +
+            projectionKey +
             (this._showOverdrawInspector ? '/overdraw' : '') +
-            (this.style.map.terrain ? '/terrain' : '');
+            (useTerrain ? '/terrain' : '') +
+            (concatenatedDefines ? (`/defines:${concatenatedDefines.join('//')}`) : '');
```
HarelM, line 645: This seems to be getting pretty complicated, can this be simplified?

src/render/painter.ts
```diff
@@ -507,36 +577,17 @@ export class Painter {
...
-        const translation = [
-            inViewportPixelUnitsUnits ? translate[0] : pixelsToTileUnits(tile, translate[0], this.transform.zoom),
-            inViewportPixelUnitsUnits ? translate[1] : pixelsToTileUnits(tile, translate[1], this.transform.zoom),
-            0
-        ] as vec3;
-
-        const translatedMatrix = new Float32Array(16);
-        mat4.translate(translatedMatrix, matrix, translation);
-        return translatedMatrix;
+    // Temporary function - translate & translate-anchor handling will be moved to projection classes,
```
HarelM, line 580: Please use `/**` type of comments above methods, it's better for the docs.
If this is to be moved to a different place you can ignore this comment assuming this will be part of this PR.

src/render/painter.ts
```diff
@@ -217,9 +232,13 @@ export class Painter {
         mat4.ortho(matrix, 0, this.width, this.height, 0, 0, 1);
         mat4.scale(matrix, matrix, [gl.drawingBufferWidth, gl.drawingBufferHeight, 0]);
 
-        this.useProgram('clippingMask').draw(context, gl.TRIANGLES,
+        const projectionData = this.style.map.projection.getProjectionData(null, null);
```
HarelM, line 235: Why are we sending `null, null` to the method? should these be optional parameters?

src/render/painter.ts
```diff
@@ -244,16 +263,27 @@ export class Painter {
...
+        const projection = this.style.map.projection;
+
+        // tiles are usually supplied in ascending order of z, then y, then x
         for (const tileID of tileIDs) {
             const id = this._tileClippingMaskIDs[tileID.key] = this.nextStencilID++;
             const terrainData = this.style.map.terrain && this.style.map.terrain.getTerrainData(tileID);
 
+            let mesh = this.tileExtentMesh;
+
+            if (projection instanceof GlobeProjection) {
```
HarelM, line 275: I usually don't like to use `instanceof` since this will require painter to be familiar with the implementation of globe project and may cause circular dependencies, consider using some interface proerty like `name` or `type`.

src/render/painter.ts
```diff
@@ -356,15 +421,17 @@ export class Painter {
         const coordsDescending: {[_: string]: Array<OverscaledTileID>} = {};
         const coordsDescendingSymbol: {[_: string]: Array<OverscaledTileID>} = {};
 
+        const deduplicateWrapped = !style.map.projection.drawWrappedTiles;
```
HarelM, line 424: As I said in previous comment, consider using renderWorldCopies.

src/render/painter.ts
```diff
@@ -408,6 +475,9 @@ export class Painter {
             this.renderLayer(this, sourceCaches[layer.source], layer, coords);
         }
 
+        // Execute offscreen GPU tasks of the projection manager
+        this.style.map.projection.updateGPUdependent(this);
```
HarelM, line 479: Passing `this` as a parameter is usually a code smell and might need a redesign or rethink of what part is truely needed to be passed down.

src/render/draw_raster.ts
```diff
@@ -22,58 +32,97 @@ export function drawRaster(painter: Painter, sourceCache: SourceCache, layer: Ra
     const source = sourceCache.getSource();
     const program = painter.useProgram('raster');
 
+    const projection = painter.style.map.projection;
+    const globe = (projection instanceof GlobeProjection && projection.useGlobeRendering);
```
HarelM, line 36: Same as other comment about instance of.

src/render/draw_raster.ts
```diff
@@ -22,58 +32,97 @@ export function drawRaster(painter: Painter, sourceCache: SourceCache, layer: Ra
...
+    // To combat this, tile meshes for globe have a slight border region.
+    // However tiles borders will overlap, and a part of a tile often
+    // gets hidden by its neighbour's border, which displays an ugly stretched texture.
+    // To both hide the border stretch and avoid tiny gaps, tiles are first drawn without borders (with gaps),
+    // and then any missing pixels (gaps, not marked in stencil) get overdrawn with tile borders.
+    // This approach also avoids pixel shader overdraw, as any pixel is drawn at most once.
 
-    const minTileZ = coords[coords.length - 1].overscaledZ;
+    // Stencil and two-pass is not used for ImageSource sources.
+    const passCount = (globe && !(source instanceof ImageSource)) ? 2 : 1;
```
HarelM, line 51: There are too many if related to globe and image source, I would consider splitting this logic to different methods and call each one according to what is needed. It's almost impossible to understand what this method is doing...

src/geo/projection/projection_base.ts
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * An abstract class the specializations of which are used internally by MapLibre to handle different projections.
+ */
+export interface ProjectionBase {
+    /**
+     * @internal
+     * A short, descriptive name of this projection, such as 'mercator' or 'globe'.
+     */
+    get name(): string;
```
HarelM, line 18: I generally don't like `get`ters and `set`ters. I prefer a method to get the relevant information.
https://medium.com/@harel.mazor/why-i-never-use-get-and-set-in-any-code-language-7a0f7de739ce

src/geo/projection/globe.ts
```diff
@@ -0,0 +1,767 @@
...
+import {ProjectionBase} from './projection_base';
+import {PreparedShader, shaders} from '../../shaders/shaders';
+
+/**
+ * The size of border region for stencil masks, in internal tile coordinates.
+ * Used for globe rendering.
+ */
+const EXTENT_STENCIL_BORDER = EXTENT / 128;
+
+function clamp(a: number, min: number, max: number): number {
```
HarelM, line 36: Consider moving this to utils file.

src/geo/projection/globe.ts
```diff
@@ -0,0 +1,767 @@
...
+    return t * t * (3.0 - 2.0 * t);
+}
+
+const globeTransitionTimeSeconds = 0.5;
+const zoomTransitionTimeSeconds = 0.5;
+const maxGlobeZoom = 12.0;
+const errorTransitionTimeSeconds = 0.5;
+
+export class GlobeProjection implements ProjectionBase {
+    private _map: Map | undefined;
```
HarelM, line 56: While I'm all for `private` and `public` stuff, this codebase doesn't use it at all, I don't know if we should introduce it. Your call...

src/geo/projection/globe.ts
```diff
@@ -0,0 +1,767 @@
...
+import {Color} from '@maplibre/maplibre-gl-style-spec';
+import {DepthMode} from '../../gl/depth_mode';
+import {CullFaceMode} from '../../gl/cull_face_mode';
+import {projectionErrorMeasurementUniformValues} from '../../render/program/projection_error_measurement_program';
+import {warnOnce} from '../../util/util';
+import {mercatorYfromLat} from '../mercator_coordinate';
+import {granularitySettings} from '../../render/subdivision';
+import Point from '@mapbox/point-geometry';
+import {ProjectionData} from '../../render/program/projection_program';
+import * as Mercator from './mercator';
```
HarelM, line 26: Why not import only the part we are interested from this file?

src/geo/projection/globe.ts
```diff
@@ -0,0 +1,767 @@
...
+    private _errorMeasurement: ProjectionErrorMeasurement;
+    private _errorQueryLatitudeDegrees: number;
+    private _errorCorrectionUsable: number = 0.0;
+    private _errorMeasurementLastValue: number = 0.0;
+    private _errorCorrectionPreviousValue: number = 0.0;
+    private _errorMeasurementLastChangeTime: number = -1000.0;
+
+    private _globeProjectionOverride = true;
+
+    private _globeProjMatrix: mat4 = [
```
HarelM, line 80: Isn't this a simple identity matrix? I think gl-matrix has methods for this...?

src/geo/projection/globe.ts
```diff
@@ -0,0 +1,767 @@
...
+
+    private _tileMeshCache: {[_: string]: Mesh} = {};
+    private _cachedClippingPlane: [number, number, number, number] = [1, 0, 0, 0];
+
+    // Transition handling
+    private _lastGlobeStateEnabled: boolean = true;
+    private _lastGlobeChangeTime: number = -1000.0;
+    private _lastLargeZoomStateChange: number = -1000.0;
+    private _lastLargeZoomState: boolean = false;
+    private _globeness: number = 1.0;
```
HarelM, line 67: I'm not sure this is a great name, but I think a comment to better explain what it does can help.

src/geo/projection/globe.ts
```diff
@@ -0,0 +1,767 @@
...
+
+        const cameraPos: vec4 = [0, 0, -1, 1];
+        vec4.transformMat4(cameraPos, cameraPos, invProj);
+        this._globeCameraPosition = [
+            cameraPos[0] / cameraPos[3],
+            cameraPos[1] / cameraPos[3],
+            cameraPos[2] / cameraPos[3]
+        ];
+
+        // We want to compute a plane equation that, when applied to the unit sphere generated
```
HarelM, line 242: This comment probably indicates that this should be moved to a method, preferably with the below input variables.

src/geo/projection/globe.ts
```diff
@@ -0,0 +1,767 @@
...
+        gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_S, gl.CLAMP_TO_EDGE);
+        gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_WRAP_T, gl.CLAMP_TO_EDGE);
+        gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MIN_FILTER, gl.NEAREST);
+        gl.texParameteri(gl.TEXTURE_2D, gl.TEXTURE_MAG_FILTER, gl.NEAREST);
+        gl.texImage2D(gl.TEXTURE_2D, 0, this._texFormat, this._texWidth, this._texHeight, 0, this._texFormat, this._texType, null);
+
+        this._fbo = context.createFramebuffer(this._texWidth, this._texHeight, false, false);
+        this._fbo.colorAttachment.set(texture);
+
+        if (this._allowWebGL2 && gl instanceof WebGL2RenderingContext) {
```
HarelM, line 635: Is this error correction only relevant to webgl2? We are writing code to facilitate for both webgl1 and 2 in maplibre.

src/geo/projection/globe.ts
```diff
@@ -0,0 +1,767 @@
...
+        }
+
+        // If we made it here, _resultBuffer contains the new measurement
+        this._readbackQueue = null;
+        this._measuredError = parseRGBA8float(this._resultBuffer);
+        this._lastReadbackFrame = this._updateCount;
+    }
+}
+
+function parseRGBA8float(buffer: Uint8Array): number {
```
HarelM, line 758: Same as above, consider moving this method to utils.

HarelM:
I have reviewed all the files, not sure my review is good enough though as my knowledge of GPU and shaders is very limited.
Would be nice if @Pheonor and maybe @JannikGM would review this as well.

Since this is targeting the globe branch, we can decide if we want to merge this sooner to facilitate more work on the globe branch.
Any feedback is welcome.

src/source/video_source.ts
kubapelc: I changed the implementation of rendering raster data (apart from raster layers this also includes video, images and canvas), now it no longer needs custom vertex buffers. `boundsBuffer` was thus removed from the base class `ImageSource`. See my comment at https://github.com/maplibre/maplibre-gl-js/pull/3783#discussion_r1518571524

HarelM: OK, thanks!

src/source/source_cache.ts
kubapelc: I've taken a deeper look at how visible tiles are determined, and also at what `renderWorldCopies` flag does, as you suggested in another thread. You are right that the flag does something similar to what I am trying to do with `deduplicateWrapped`, but it also has some side effects (constrains zoom and map center to only show one world copy at any time).

It might be better to implement the tile deduplication in `transform.ts`, in the `coveringTiles` method. But, additionally, proper determination of what tiles are visible on a globe map is actually part of the same problem. I haven't even started work on that, but @Pheonor has already implemented it in his branch, see changes around https://github.com/Pheonor/maplibre-gl-js/blob/300f8303c4b4e2df03b9d1773e42e4b2f8e4baea/src/geo/transform.ts#L498

Right now I see two options: I can temporarily move `deduplicateWrapped` functionality to `transform.ts`, or I can just remove it altogether for now. The reason is that the entire `coveringTiles` will need to be changed in the near future to accomodate different projections, and probably moved from `transform.ts` to the projection classes. So I see no point in coming up with a nice solution at this moment.

src/source/source_cache.ts
kubapelc: This is related to the conversation about `deduplicateWrapped`, please see https://github.com/maplibre/maplibre-gl-js/pull/3783#discussion_r1523424897

src/shaders/_projection_mercator.vertex.glsl
kubapelc: It starts with "_" to stay consistent with the naming of `_prelude.vertex.glsl` and `_prelude.fragment.glsl`, since this file also contains a "prelude" that is placed at the beginning of every shader. Do you think removing the "_" or moving it to a subdirectory would be better?

src/render/subdivision.ts
kubapelc: Yes, these values are for globe projection, a sweetspot between producing a sufficiently tesselated/round globe while not generating too much extra geometry. I want to keep them readonly, since they cannot change once a map is loaded. See conversation around https://github.com/maplibre/maplibre-gl-js/pull/3783#discussion_r1518572684 - once runtime projection switching is implemented, every projection would have its own set of granularity settings. But they should still be constant for a given projection.

src/render/subdivision.ts
```diff
@@ -0,0 +1,52 @@
...
+     * granularity used for stencil mask tiles.
+     */
+    public readonly granularityStencil;
+
+    /**
+     * granularity used for the line layer.
+     */
+    public readonly granularityLine;
+
+    constructor(fill: SubdivisionGranularityExpression, line: SubdivisionGranularityExpression, stencil: SubdivisionGranularityExpression) {
```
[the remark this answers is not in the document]
kubapelc: Good point, I changed the constructor argument to an object.

src/render/painter.ts
kubapelc: I've refactored it a bit. Not necessarily to make it simpler, but at least easier to read I hope.

src/render/program.ts
kubapelc: I've refactored this constructor a bit.

src/render/painter.ts
kubapelc: I've changed it to `/**`.

src/render/painter.ts
kubapelc: I've noticed that most places where I need globe-specific codepaths are not actually globe-specific, but rather specific to any projection which uses subdivision, so I refactored the interface to avoid `instanceof GlobeProjection` checks.

src/render/painter.ts
kubapelc: This is due to `ProjectionErrorMeasurement`. It mostly just needs the GL context, but it also needs to call `useProgram` on a Painter instance. I could just pass the function alongside the GL context, but that feels too specific for a generic projection interface.

src/render/draw_raster.ts
kubapelc: Yeah this is a mess 🤦 I've refactored it into something that is hopefully much more readable.

src/geo/projection/globe.ts
kubapelc: I'd prefer to use `private` and `public`, and it seems both keywords are already used in some places (`worker.ts`, `render_pool.ts`).

src/geo/projection/globe.ts
kubapelc: Done, I've also removed `smoothStep` and used `easeCubicInOut` instead.

src/geo/projection/globe.ts
kubapelc: Done.

src/geo/projection/globe.ts
kubapelc: I've replaced the initialization with `mat4.create();`

src/geo/projection/globe.ts
kubapelc: I've added a comment to `_globeness`.

src/geo/projection/globe.ts
kubapelc: Yes this makes sense, I've moved it to a separate method.

src/geo/projection/globe.ts
kubapelc: It has a fallback for WebGL1. The `_allowWebGL2` variable is something I used to test this fallback, it is not needed anymore, so I'll remove it.

src/geo/projection/globe.ts
kubapelc: This method is specific to how ProjectionErrorMeasurement encodes the value, so I'd prefer to leave it where it is. It is not a general utility function.

src/render/painter.ts
HarelM: It is better to be more specific.
Let's start with passing only what's needed, in the future if more parameters are needed we can pass those, this way you create less coupling.

src/render/draw_raster.ts
```diff
@@ -47,82 +40,86 @@ export function drawRaster(painter: Painter, sourceCache: SourceCache, layer: Ra
...
 
-    // Stencil and two-pass is not used for ImageSource sources.
-    const passCount = (globe && !(source instanceof ImageSource)) ? 2 : 1;
-
-    let stencilModesLow, stencilModesHigh, coords: Array<OverscaledTileID>;
-
-    if (passCount > 1) {
-        [stencilModesHigh, stencilModesLow, coords] = painter.stencilConfigForOverlapTwoPass(tileIDs);
+    // Stencil mask and two-pass is not used for ImageSource sources regardless of projection.
+    if (source instanceof ImageSource) {
```
HarelM, line 44: This looks a lot better! but I would still consider moving all the logic related to ImageSource to a separate method (that might call other small methods in this class) as there are still a few places in the draw tiles that checks if this is an image source or not.

src/geo/projection/globe_projection_error_measurement.ts
```diff
@@ -0,0 +1,239 @@
...
+        }
+
+        // If we made it here, _resultBuffer contains the new measurement
+        this._readbackQueue = null;
+        this._measuredError = parseRGBA8float(this._resultBuffer);
+        this._lastReadbackFrame = this._updateCount;
+    }
+}
+
+function parseRGBA8float(buffer: Uint8Array): number {
```
HarelM, line 230: Why not move it to be a private method of the above class? You can make it static if you want.

src/geo/projection/globe_projection_error_measurement.ts
```diff
@@ -0,0 +1,239 @@
...
+                frameNumberIssued: this._updateCount,
+                sync: null,
+            };
+        }
+    }
+
+    private _tryReadback(context: Context): void {
+        const gl = context.gl;
+
+        if (this._pbo && this._readbackQueue && gl instanceof WebGL2RenderingContext) {
```
HarelM, line 199: I think we have `isWebGl2Supported` method somewhere, maybe we should use it?

src/shaders/_projection_mercator.vertex.glsl
HarelM: If this server the same purpose of _prelude in terms of being placed at the beginning of each shader why not add it to the prelude file?
While I don't like the idea of having a large file with a lot of methods, it might server the purpose.
Or we can call it `_prelude_for_projections`?

src/source/source_cache.ts
HarelM: If it doesn't break anything and we are about to replace it, let's remove it.

HarelM:
I think I resolved all the current implemented review comments, I left those that are still relevant I believe.
I've opened a design proposal for the style spec:
https://github.com/maplibre/maplibre-style-spec/issues/568
We might be able to ignore it for now like we do for the zoom and pitch that can be set for both the map configuration and the style, so it's not a blocker I guess, make sure to check my assumption.

test/examples/globe.html
HarelM: Might not be relevant, see my comments:
https://github.com/maplibre/maplibre-gl-js/pull/3783#issuecomment-1996770425

src/geo/projection/projection_base.ts
kubapelc: I think a getter makes sense in this case, since its implementations should always just return a constant with no side effects.

src/render/draw_raster.ts
kubapelc: I've refactored it a bit more to remove `instanceof ImageSource` from `drawTiles`.

src/geo/projection/globe_projection_error_measurement.ts
kubapelc: Yes that makes sense, done.

src/geo/projection/globe_projection_error_measurement.ts
kubapelc: I've refactored it to use `isWebGL2`.

src/render/painter.ts
kubapelc: This is a bit of a special case, the method just draws a fullscreen quad, so I force a simple (mercator) shader that only uses a matrix multiplication for projection, and pass that matrix as an input instead of full `projectionData`. But `getProjectionData` should not have its arguments marked as optional I think, since it never makes sense to not use them - other than in this place.

I've refactored it so that instead of calling `getProjectionData(null, null)`, the mostly empty `ProjectionData` object is just created directly instead.

src/source/source_cache.ts
kubapelc: Done, removed.

src/shaders/_projection_mercator.vertex.glsl
kubapelc: Either `_projection_mercator.vertex.glsl` or `_projection_globe.vertex.glsl` gets inserted at the start of all shaders based on what projection is used, each file has a projection-dependet implementation of the shader functions. I used to have both in `_prelude.vertex.glsl` and used `#ifdef GLOBE` to switch, but I figured placing them into different files is a nicer approach, especially since there might be more than two projections in the future.

test/integration/render/run_render_tests.ts
```diff
@@ -56,6 +57,8 @@ type TestData = {
     actual: string;
     diff: string;
     expected: string;
+
+    projection?: ProjectionName;
```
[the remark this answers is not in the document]
kubapelc: I've looked at `StyleSpecification`, but it seems to be an auto-generated file. I'm not sure about what I need to do to add projection into style spec.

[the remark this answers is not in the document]
HarelM: There's a need to update the style spec package.
See my comment here: https://github.com/maplibre/maplibre-gl-js/pull/3783#issuecomment-1996770425
Let's just make sure this is similar to the zoom/center part of the map configuration.

src/shaders/_projection_mercator.vertex.glsl
HarelM: I see. Your call, the "-" prefix is OK.

HarelM:
I think most code review comments were resolved.
Please update the build test to have the updated bundle size so that all tests pass.
I'm wondering if more tests are needed or should we add them after more code changes have been made, as currently there are unit tests missing for the new classes, but this can wait for later PRs.

src/render/painter.ts
kubapelc: Fixed.

src/render/painter.ts
```diff
@@ -408,7 +474,17 @@ export class Painter {
             this.renderLayer(this, sourceCaches[layer.source], layer, coords);
         }
 
+        // Execute offscreen GPU tasks of the projection manager
+        const painterInstance = this;
```
HarelM, line 478: `this` is captured when you create a lambda `() => {}`, so this assignment is not needed.

src/geo/projection/globe.ts
```diff
@@ -0,0 +1,514 @@
...
+
+    /**
+     * When true, globe view fill function as normal. When false, mercator will be used at all zoom levels instead.
+     * Transitioning between states will be animated.
+     * Map should be updated after changing this value.
+     * True by default.
+     */
+    get globeView(): boolean { return this._globeProjectionOverride; }
+    set globeView(value: boolean) {
+        if (value !== this._globeProjectionOverride) {
```
HarelM, line 136: What's the point of this if?
Generally speking, I would simply make `_globeProjectionOverride` public, these getter and setter do not add anything I believe.

src/ui/map.ts
```diff
@@ -58,6 +58,8 @@ import type {
 import type {MapGeoJSONFeature} from '../util/vectortile_to_geojson';
 import type {ControlPosition, IControl} from './control/control';
 import type {QueryRenderedFeaturesOptions, QuerySourceFeatureOptions} from '../source/query_features';
+import {ProjectionBase} from '../geo/projection/projection_base';
```
HarelM, line 61: Consider calling the projection interface simply "projection" and avoid the base part of the word.

kubapelc: I wanted to avoid confusion with `symbol/projection.ts`, but as that file doesn't contain a class I think it's ok to go ahead with your suggestion.

src/geo/projection/globe.ts
kubapelc: Setting this used to have a side effect of updating the map (yeah, this was the bad kind of setter). I've now changed both setter and getter to be actual functions and refactored things a bit.

src/render/painter.ts
kubapelc: Fixed.

src/geo/projection/globe.ts
```diff
@@ -0,0 +1,514 @@
...
+
+    get useSubdivision(): boolean {
+        return this.useGlobeRendering;
+    }
+
+    get useSpecialProjectionForSymbols(): boolean {
+        return this.useGlobeRendering;
+    }
+
+    get isRenderingDirty(): boolean {
```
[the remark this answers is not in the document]
kubapelc: I've changed this to a function.

kubapelc:
@HarelM I've added some render tests in a previous commit, and now I've added unit tests for globe and mercator as well, although they don't test too much functionality. I'll think about what else I can cover with unit tests.

I've fixed the bundle size in build test.

HarelM (approved):

HarelM:
This is now merged to globe branch, I'll do my best to keep this brach up-to-date with main.
Please continue pushing improvements to it.

PR #3793, opened 2024-03-05
Fix bad curve on easeTo/flyTo in constrained space

sbachinin:
 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

I know there's a lot to review now so don't haste with this )

This PR involves several methods that animate the map's scale and center:
- `flyTo` (and therefore `fitBounds` and `fitScreenCoordinates`)
- `easeTo `(and therefore `panTo `and `panBy`)

These methods don't animate well when the destination point is near the bounds: those set by maxBounds, by vertical edges of the world, by antimeridian on single-globe map.

What is apparently incorrect in current easeTo and flyTo algorithms is that they don't apply these constraints when deciding on the destination. They just take the destination lngLat as it was provided by a user.
The result is that this "optimistic" animation hits the bounds somewhere in the process, and a smooth curve breaks into something more linear.

This PR fixes this problem. It runs the code from `transform._constrain` in advance, before the animation starts.
For that I had to extract most of `transform._constrain` code into a new `getConstrained` method that doesn't set anything on Transform instance but only calculates zoom and center for a given point. There are a lot of changes in this method but it's mostly just renaming the magic variables. 

The main meaningful change there is that I deal not with actual `transform.worldSize` but with a "theoretical" `worldSize `that corresponds to a requested zoom level.
Also, the requested zoom level is now clamped to [`minZoom`, `maxZoom`] _inside_ `getConstrained`. (Instead of doing that in `easeTo`/`flyTo`). I think it just makes sense because zoom bounds are closely related to coord bounds.

This change in itself is non-destructive and it surely changes the situation for the better.
The implications are:
* easeTo is always smooth near the bounds:
https://github.com/maplibre/maplibre-gl-js/assets/11789697/b5347134-a529-4575-993c-e7dd1883cf95

* flyTo is smooth near the bounds only in _some_ cases - when there is no significant zooming out (on low zoom levels or when a user sets `curve` option to something small).
In other cases flyTo trajectory gets only slightly better but is still broken.

The problem of flyTo is not only the final destination but intermidate frames too because of zooming out. Ideal solution to this will be to calculate the animation path with respect of constraints. But this will be a lot of engineering with little impact.

I see a couple of (supposedly) simple changes here. They may look a bit controversial but I think they will provide a better UX. They are _not included_ in this PR.

1) **Disable constraints** for the time of flyingTo.
This will result in:
- proper animation path
- "restricted" space (possibly white bg) will be visible for a little while. (I believe this is not essential).
- panning / zooming / whatever during the unfinished "flight" outside the bounds will cause an abrupt return into the bounds.
The last problem is of course not that small.	Though still we are in a better position: bad behaviour will happen only on interaction and in a situation where animation was ugly anyway.
In terms of code, I tried it and it looks easy.

2) It must be possible to **ignore user actions** when there is an unfinished flight and it's happening outside the bounds.
Result:
- animation always looks good.
- user can't pan/zoom/... during this "overflow". IDK if it's a big problem.
I can't say yet anything about code complexity of this step.

As for tests, I didn't add any. The only thing that changes here in terms of UX is the trajectory of animation. (Final point doesn't change).
Now it's more possible to test `getConstrained` in isolation but IDK if it's desirable.

### DIFF
--- a/src/geo/transform.ts
+++ b/src/geo/transform.ts
@@ -711,8 +711,11 @@ export class Transform {
         return this.mercatorMatrix.slice() as any;
     }
 
-    _constrain() {
-        if (!this.center || !this.width || !this.height || this._constraining) return;
+    getConstrained(center: LngLat, zoom: number): {center: LngLat; zoom: number} {
+        const result = {
+            center: new LngLat(center.lng, center.lat),
+            zoom
+        };
 
         let lngRange = this.lngRange;
 
@@ -721,15 +724,12 @@ export class Transform {
             lngRange = [-almost180, almost180];
         }
 
-        this._constraining = true;
-
         let minY = -90;
         let maxY = 90;
         let minX = -180;
         let maxX = 180;
         let sy, sx, x2, y2;
-        const size = this.size,
-            unmodified = this._unmodified;
+        const size = this.size;
 
         if (this.latRange) {
             const latRange = this.latRange;
@@ -755,19 +755,17 @@ export class Transform {
             sx = maxX - minX < size.x ? size.x / (maxX - minX) : 0;
         }
 
-        const point = this.point;
+        const point = this.project(center);
 
         // how much the map should scale to fit the screen into given latitude/longitude ranges
         const s = Math.max(sx || 0, sy || 0);
 
         if (s) {
-            this.center = this.unproject(new Point(
+            result.center = this.unproject(new Point(
                 sx ? (maxX + minX) / 2 : point.x,
                 sy ? (maxY + minY) / 2 : point.y));
-            this.zoom += this.scaleZoom(s);
-            this._unmodified = unmodified;
-            this._constraining = false;
-            return;
+            result.zoom += this.scaleZoom(s);
+            return result;
         }
 
         if (this.latRange) {
@@ -792,11 +790,21 @@ export class Transform {
 
         // pan the map if the screen goes off the range
         if (x2 !== undefined || y2 !== undefined) {
-            this.center = this.unproject(new Point(
+            result.center = this.unproject(new Point(
                 x2 !== undefined ? x2 : point.x,
                 y2 !== undefined ? y2 : point.y)).wrap();
         }
 
+        return result;
+    }
+
+    _constrain() {
+        if (!this.center || !this.width || !this.height || this._constraining) return;
+        this._constraining = true;
+        const unmodified = this._unmodified;
+        const {center, zoom} = this.getConstrained(this.center, this.zoom);
+        this.center = center;
+        this.zoom = zoom;
         this._unmodified = unmodified;
         this._constraining = false;
     }
--- a/src/geo/transform.ts
+++ b/src/geo/transform.ts
@@ -729,10 +729,10 @@ export class Transform {
             lngRange = [-almost180, almost180];
         }
 
-        let minY = -90;
-        let maxY = 90;
-        let minX = -180;
-        let maxX = 180;
+        let minY = 0;
+        let maxY = this.worldSize;
+        let minX = 0;
+        let maxX = this.worldSize;
         let scaleY = 0;
         let scaleX = 0;
         const {x: screenWidth, y: screenHeight} = this.size;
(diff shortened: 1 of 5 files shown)


src/geo/transform.ts
```diff
@@ -711,8 +711,16 @@ export class Transform {
         return this.mercatorMatrix.slice() as any;
     }
 
-    _constrain() {
-        if (!this.center || !this.width || !this.height || this._constraining) return;
+    // Get center lngLat and zoom to ensure that
```
HarelM, line 714: Can you use `/** ...` type of comment, this way IDEs can pick it up.

src/geo/transform.ts
```diff
@@ -721,82 +729,90 @@ export class Transform {
...
-
-        let minY = -90;
-        let maxY = 90;
-        let minX = -180;
-        let maxX = 180;
-        let sy, sx, x2, y2;
-        const size = this.size,
-            unmodified = this._unmodified;
+        const worldSize = this.tileSize * this.zoomScale(result.zoom); // A world size for the requested zoom level, not the current world size
+        let minY = 0;
```
HarelM, line 733: Can you elaborate on this change of min/max/y/x from -180..180 -90..90 to 0..worldsize?

src/ui/camera.ts
```diff
@@ -979,15 +979,18 @@ export abstract class Camera extends Evented {
...
             padding = 'padding' in options ? options.padding : tr.padding;
 
         const offsetAsPoint = Point.convert(options.offset);
         let pointAtOffset = tr.centerPoint.add(offsetAsPoint);
         const locationAtOffset = tr.pointLocation(pointAtOffset);
-        const center = LngLat.convert(options.center || locationAtOffset);
+
+        const {center, zoom} = tr.getConstrained(
+            LngLat.convert(options.center || locationAtOffset),
+            options.zoom ?? startZoom
```
HarelM, line 992: is zoom = 0 a valid input?

HarelM:
I would feel better if there was a test specifically to cover this bug fix.
Might be a stupid question, but did you make sure fly to and ease to work as expected if renderWorldCopies is ture?

If you can add some videos of before and after for both cases it would be amazing.
Thanks for all the work on this!!

src/geo/transform.ts
sbachinin: Done

src/geo/transform.ts
sbachinin: These variables were apparently meant to store world coords, not lngLat.
It follows from how these variables are used later in the code.
LngLat-like initial values were a mistake. (Though it didn't create any visible problems)

src/ui/camera.ts
sbachinin: I guess 0 is a valid zoom level in general, and in this case too. If you `map.easeTo({zoom: 0})`, it will ease to the lowest possible zoom.

sbachinin:
[Extended before/after](https://github.com/maplibre/maplibre-gl-js/assets/11789697/d09adfa6-aef2-40ca-89ee-81e0eba2deb1)

sbachinin:
>I would feel better if there was a test specifically to cover this bug fix.

As I said, I don't think it's testable.
The bug was only in the trajectory of easeTo/flyTo, not in the final destination.
So the fix shows up only in the intermediate frames.
Perhaps it's technically possible to check intermediate values but such test will surely be very unstable.

src/ui/camera.ts
HarelM: So doing ?? on 0 might create the wrong behavior in this case, wouldn't it?

sbachinin: If options.zoom is 0, the whole `options.zoom ?? startZoom` expression will return 0.
Therefore 0 will be passed to getConstrained.
I think that's what we need.
Previously it was written as
`'zoom' in options ? +options.zoom : startZoom`
and I've changed it to a shorter equivalent.

HarelM: OK, thanks! was confused with `||` operator which is used in the code widely and can cause the issue I thought is the problem here.

HarelM (approved):

PR #3796, opened 2024-03-05
Proof of bug: missing mercatorMatrix3d

olsen232:
When writing a CustomLayerInterface in a 3D context, the developer has need of a 3D equivalent to the mercatorMatrix, but none is provided by the interface. This showcases the issue caused by the lack of one, and has a basic example of how to fix.

Currently only intended to showcase the issue - will rewrite once I've opened an issue to discuss the proper fix.

Issue filed as https://github.com/maplibre/maplibre-gl-js/issues/3797


HarelM:
@olsen232 can we close this PR?

PR #3836, opened 2024-03-14
Fix for #2094 - unwanted zoom changes

### LINKED ISSUE #2094: Quick vertical dragging of pitched 3d map with mouse results in zoom action [3.0.0-pre.3]
**maplibre-gl-js version**: 3.0.0-pre.3

**browser**: Google Chrome Version 109.0.5414.75 (Offizieller Build) (64-Bit)

### Steps to Trigger Behavior

 1. Quickly and repeatedly click mouse, push mouse up and release mouse
 2. Quickly and repeatedly click mouse, push mouse down and release mouse

891E321D-8C19-4150-A649-BAEA13C713F6

### Link to Demonstration

~~https://output.jsbin.com/decicab/1 -- maplibre-gl@v3.0.0-pre.3 -- broken behavior~~
~~https://output.jsbin.com/netasuw/1 -- maplibre-gl@2.4.0 -- correct behavior~~
(Does no longer work because of new MapTiler restrictions)

https://output.jsbin.com/decicab/11 -- maplibre-gl@v3.0.0-pre.3 -- broken behavior
https://output.jsbin.com/netasuw/11 -- maplibre-gl@2.4.0 -- correct behavior
(Terrain limited to Innsbruck area)

### Expected Behavior

The altitude of the camera should remain at the same height, making a mouse up/down drag result only in panning of the map, like in version 2.4.0.

### Actual Behavior

Panning has zooming added to it. So multiple quick drags backing off of the visible location result in not only panning backward, but also in zooming out. Multiple quick drags to fly over the visible location not only result in panning forward, but also in zooming in.

olsen232:
As described in https://github.com/maplibre/maplibre-gl-js/issues/2094
panning the map can result in a sudden jump in zoom-level as the
drag is released. This jump tends to be worse when more zoomed out,
and repeated short pans cause the zoom jumps to happen repeatedly
in the same direction, which will grow exponentially if each
one is zooming out - each zoom out causes the bug to get worse.

The problem lies in recalculateZoom, which tries to calculate a new
zoom level such that the previous camera altitude + terrain elevation
is the same as the current camera altitude + terrain elevation.
Camera altitude here refers specifically to the camera's height
above the terrain elevation, and terrain elevation is the terrain
height at the centre of the screen. (This doesn't really leave any
good nouns for describing the sum of these two variables, but
this sum should remain constant and is the camera's total height
above sea level).

A specific zoom maps - more-or-less - to a specific camera altitude.
Since the terrain elevation can be different (eg 100m higher) at the
end of a terrain drag, we need the camera altitude to be different
too (eg 100m lower to compensate). So, we need to recalculate the
zoom to one that maps to the required camera altitude. And so,
the function recalculateZoom exists to fulfil this purpose.

Unfortunately, the function miscalculates the required zoom, as
can be verified by comparing camera altitude + terrain elevation
before recalculateZoom is called vs after - very slight changes
in terrain elevation (10s of metres) lead to large and unrelated
changes in camera altitude (1000s of metres) when the camera is
very zoomed out.

Currently, the logic in recalculateZoom uses different math to
the math used to calculate the camera's altitude from the
current zoom. This different math may be close to correct, but,
it's impossible for me to debug. In this commit I've simply
replaced it with the inverse of the math used to calcualte the
camera's altitude from its zoom - ie:
if a zoom Z gives a camera altitude A = f(Z)

Then, to get the camera to required altitude A,
we need to set the zoom to Z = inverse-of-f(A)

Fixes #2094

## Launch Checklist

 - [x] Link to related issues.
 - [x] Document any changes to public APIs. (Not applicable)
 - [ ] Post benchmark scores.


HarelM:
Thanks!! I was hoping the issue was with incorrect calculation.
Can you add a unit test for zoom recalculation somehow - to show that the math is inverted basically, much like doing a convert and convert back test to see that you get the same object.

Also a changelog item is important.

And if you can add a video of after the fix and make sure this works well on touch devices doing pinch zoom.

olsen232:
Before fix:
https://github.com/maplibre/maplibre-gl-js/assets/7425442/a2234e54-24f2-449b-ae44-3633a5eceb12

After fix:
https://github.com/maplibre/maplibre-gl-js/assets/7425442/6a86dac4-b730-493c-9fa9-8fc94a354b30

I'm travelling over the next few days - I will eventually work through the checklist, I may finish it while travelling, or I may get to it next week once I'm back at a desk.

olsen232:
This is now mostly done. 
Not done:
- Post benchmark scores (is this relevant?)
- Test on a touch device.

HarelM (approved):

HarelM (approved):

prozessor13:
@olsen232 thank you very much for this fix!! I always knew that my code “just kind of works,” but it’s certainly not correct. Really good that this is fixed now!

prozessor13:
Beside that fix i notice still some small jumps at mouse-moveend. Now, that recalcuteZoom is fixed, i am pretty sure that this issue is located [here](https://github.com/maplibre/maplibre-gl-js/blob/7308da584248c76e17f20f1f17df0b63acd34ade/src/render/terrain.ts#L329). This issue can either be fixed with:
* calculate correct position with ray/triangle intersection from camera to terrain-mesh
* or the pragmatic, still not perfect way: grab more than only pixel from the coords-framebuffer and make some sort of interpolation to get a more accurate position.

olsen232:
Wrote up the "small jumps at mouse-moveend" issue mentioned by prozessor13 here: 
https://github.com/maplibre/maplibre-gl-js/issues/3872

Issue #3313, opened 2023-11-04
Mutipolygon buildings do not work well with 3d terrain, show floating or underground.


acalcutt:
When using a planetiler 0.7.0 generated pmtiles file in maplibre-js with 3d terrain enabled, I observed floating buildings and buildings going under the ground. With a file generated with planetiler 0.6.0, this issue did not occur. After discussion on slack in https://osmus.slack.com/archives/C03TFH5NE83/p1698949699781319 , I think this relates to a change in planetiler 0.7.0 which merges buildings with the same tags into a single multipolygon to save space. In 0.6.0 each building was it's own polygon and each was tagged individually.

0.6.0
![image](https://github.com/maplibre/maplibre-gl-js/assets/3792408/df0c2333-00a6-4452-ab50-b6daf5db651a)

0.7.0
![image](https://github.com/maplibre/maplibre-gl-js/assets/3792408/cb510fdb-15cf-4014-8425-329235e3c9a9)

While this could be considered a planetiler issue, I am also not sure that maplibre-js is handling the multipolygon properly with 3d terrain enabled.

For example, I pulled this building geojson out of a planetiler 0.7.0 pbf file with [vt2geojson](https://github.com/mapbox/vt2geojson)
`vt2geojson --layer building https://tiles.wifidb.net/data/planetiler-0-7-0_pmtiles/14/3094/6569.pbf > 14-3094-6569.geojson`
this makes a multipolygon like
![multipolygon](https://github.com/maplibre/maplibre-gl-js/assets/3792408/dbaa5152-a7b6-4686-b006-683a2c675ce9)
[14-3094-6569.geojson.txt](https://github.com/maplibre/maplibre-gl-js/files/13255723/14-3094-6569.geojson.txt)

If I put that into the maplibre building example I get this stackblitz example page
https://stackblitz.com/edit/web-platform-zaxezv?file=index.html
In this example you can see these buildings floating in the zoom 14 range
![floating_buildings](https://github.com/maplibre/maplibre-gl-js/assets/3792408/4f5dec57-caf6-4efd-9b4d-88f3f5345e01)
it looks like they get placed at a height in the center of the multipolygon above, which is much higher than the terrain where the building sits, causing them to float.

I also made this example using a live planetiler 0.7.0 pmtiles file, you can see it looks much worse than the geojson example, with this issue showing past zoom 14
https://stackblitz.com/edit/web-platform-qmqmrs?file=index.html
[...]

PR #3841, opened 2024-03-14
Elevation per polygon for fill extrusions

antonmarsden:
This PR is a basic solution to [issue #3313](https://github.com/maplibre/maplibre-gl-js/issues/3313), and renders multipolygons at their own centroid's elevation.

It is potentially a breaking change for some edge cases though, and it may be worth introducing a new **fill-extrusion** style property (e.g., ""fill-extrusion-elevation-per-polygon") to optionally activate this behaviour. Suggestions welcomed :)

 - [X] Link to related issues.
 - [ ] Post benchmark scores.

**Before:**
before

**After:**
after


HarelM:
Thanks for taking the time to look into this!

This is only relevant for terrain as it will be kept at 0 for non terrain maps, right?
If my above assumption is correct I don't think there's a need to change the style spec for that.
Can you add a render test that demonstrates this issue/solution?
Also, in the linked bug, there's a jump in the elevation between two zoom levels (the buildings starts floating), did you happen to try and solve that as well?

HarelM:
Also worth making sure is that buildings with multiple parts are presented correctly after this fix.
We had an issue in the past with terrain and second level building porch etc.

acalcutt:
I just want to mention I was talking to someone in slack a while back that wanted to keep the heights of separate polygons the same when using terrain ( https://osmus.slack.com/archives/C01G3D28DAB/p1705407931974739 ) . I had suggested they could possibly use the Multipolygon grouping to kind of do what they wanted, but the fix we are trying to use here would break that option for them. If you did add an option like the 'fill-extrusion-elevation-per-polygon' it could possibly still allow that kind of use case.

HarelM:
I think that the default should be that every polygon would be placed in the right place.
If someone needs something "special" then a design proposal in the spec and a PR with implementation can go a long way :-)

acalcutt:
Makes sense to me and it would fix my original issue https://github.com/maplibre/maplibre-gl-js/issues/3313 , just wanted to bring up there may be some cases where this isn't wanted, like we talked about around https://github.com/maplibre/maplibre-gl-js/issues/3313#issuecomment-1803598318

I think it makes sense that the change in this PR is the default.

HarelM:
@antonmarsden Can you please add a render test so I can merge this?

antonmarsden:
@HarelM I have added a basic test to verify that centroids are created on a per-polygon basis. Have not had much spare time recently, am hoping this is sufficient to move things forward.

HarelM:
Thanks! I was hoping for a render test, something similar to the screenshot you added, but hopefully simpler.
Unit test is great but less powerful in this case I think.

antonmarsden:
@HarelM I assume you mean adding to `integration/render/tests`?

HarelM:
Yes, create a style with a few building in a multipolygon and use a terrain from other tests to show that the buildings don't have the same elevation.

package.json
```diff
@@ -99,6 +99,7 @@
     "is-builtin-module": "^3.2.1",
     "jest": "^29.7.0",
     "jest-environment-jsdom": "^29.7.0",
+    "jest-mock-extended": "^3.0.5",
```
HarelM, line 102: Is this really needed? I believe jest has a built in capabilities of mocking classes, doesn't it?

HarelM:
CC: @louwers note for parity. Should be excluded as it used terrain.

louwers:
Can you move the render tests under `terrain`?

louwers (dismissed):
Can you move the render tests under `terrain` since it is related to (and requires) this feature? Then it won't show up as a disparity.

CHANGELOG.md
```diff
@@ -2,12 +2,16 @@
 
 ### ✨ Features and improvements
 
+- Improve animation curve when easeTo and flyTo with constraints ([#3793](https://github.com/maplibre/maplibre-gl-js/pull/3793))
+- For filled extrusions, calculate the elevation per polygon (see issue [#3313](https://github.com/maplibre/maplibre-gl-js/issues/3313))
+
```
HarelM, line 7:
```suggestion
```

HarelM (approved):

louwers (approved):

louwers:
Thanks!

petrsloup:
@HarelM any plans for a bugfix release soon? We would like to get this fix to production as soon as possible...

HarelM:
Sure, I'll push a release in the next few days.

PR #3854, opened 2024-03-18
Return above-sea-level elevation from queryTerrainElevation + Pass lowered mercator matrix to custom layer

### LINKED ISSUE #3736: map.queryTerrainElevation not returning expected values
`map.queryTerrainElevation` returns unexpected values. I've created a demo page that loads elevation tiles and centers on Everest. It then displays the value of `map.queryTerrainElevation` whenever the map `idle` event is triggered.

**maplibre-gl-js version**: 4.0.2

**browser**: Chrome Version 121.0.6167.85

### Steps to Trigger Behavior

 1. Load https://design-deploy.eu/maptest.html
 2. Move around the map and observe the value in the top left corner

### Link to Demonstration

https://design-deploy.eu/maptest.html

### Expected Behavior

The page runs:
```
  map.on('idle', () => {
    document.getElementById("elevation").innerHTML = map.queryTerrainElevation([86.922623, 27.986065]); 
  });
```
I'd expect the values written to be close to the height of Everest.
### Actual Behavior
Values fluctuate quite widely, also including minus values.

### LINKED ISSUE #3797: Custom Layer Interface is missing a mercatorMatrix3d
The [Custom Layer Interface documentation](https://maplibre.org/maplibre-gl-js/docs/API/interfaces/CustomLayerInterface/) gives an example of how to make a custom 2D layer.
It documents that you can choose whether the layer is 2D or 3D - but the effects of that choice are quite limited, possibly only enabling or disabling the depth buffer.
Different transformational matrices are used for 2D and 3D rendering, but the Custom Layer Interface is always provided with a matrix called `mercatorMatrix`. This matrix is appropriate, I believe, for 2D rendering, but not for 3D rendering, since it is missing one of the translate steps that the matrices used for 3D rendering have. Specifically, it doesn't have this line of code, or any equivalent:
`mat4.translate(m, m, [0, 0, -this.elevation]); // elevate camera over terrain`

Because it is missing this translation - which is not always the same amount, and depends in some way on the terrain height at the center of the camera - using this matrix to render a point in a 3D context causes the point to appear too high, by a small amount when the camera-center-terrain-height is near sea level, and by a large amount when the camera-center-terrain-height is high altitude.

**maplibre-gl-js version**:
4.1.0 or main @ 33457131

**browser**:
Brave, but this logic is not browser dependent.

### Steps to Trigger Behavior

Use a CustomLayerInterface to render points / objects in a 3D view, with terrain enabled, in a place with hilly terrain (effects of the bug will be minimal if the terrain is close to sea level).

### Link to Demonstration

https://github.com/maplibre/maplibre-gl-js/pull/3796 - which is still a draft - shows the issue (the red point is too high), and shows a basic fix (the green point is in the correct location). To try it out, checkout these changes, run build-dev, and open `test/examples/custom-layer-interface-bug.html`
Moving the view around will cause the red dot's height to jump to some other value whenever `recalculateZoom` is called, which happens at the end of each panning movement.

Here is a screenshot of the issue as shown by the #3796
Screenshot 2024-03-06 at 10 47 25 AM

Note that the green point has the right position, this is possible since I've modified `src/geo/transform.ts` to expose another matrix called `mercatorMatrix3D`.

### Expected Behavior
- Red point should be shown where the green point is.
- A matrix should be provided to make it possible / straight-forward to show the red point in the right place.
- Documentation at https://maplibre.org/maplibre-gl-js/docs/API/interfaces/CustomLayerInterface/ should explain how to render in 2D and in 3D - or doing either one should work without modifying the given example code.

### Actual Behavior
- Red point is in the wrong place.
- No matrix provided by CustomLayerInterface to get the red point in the right place.
- Documentation doesn't explain how to get the red point in the right place when rendering in 3D.

sbachinin:
This fixes #3736 without introducing breaking changes.
Fixes #3797.
I think it would also be "the proper way" in terms of positioning custom layer objects. 

Actually it creates changes to public API.
But I'll argue that they are non-breaking in a sense that all existing user-written code will work as before.

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

Solution is to
1) return above-sea-level elevation from `queryTerrainElevation` (instead of returning this strange offset between the elevation of a given point and elevation of centerPoint).
2) pass a "lowered" mercator matrix (instead of just mercator matrix) to `render()` method of a custom layer. "Lowered" means that mercator matrix is translated by the same Z distance as proj matrix. And this allows to use intuitive above-sea-level elevation to position custom layer objects.

These 2 changes together ensure that:
- user who expects queryTerrainElevation to return the actual above-sea-level elevation will get what he needs.
- user who uses queryTerrainElevation to translate his custom layer objects will get the same results without changing the existing code.

Technically this PR changes the public API (queryTerrainElevation returns a different number + different matrix passed to `render`) but I believe that these 2 changes compensate each other, with all existing code working as before.

That is, there is no change in how these values can be used.

So far the only sensible way to use them was to translate a given matrix by "x" (whatever is returned from queryTerrainElevation). From the user's POV, this "x" in itself was a meaningless number (some cryptic offset). And the Z component of the matrix (based on the elevation of the centerPoint) is meaningless too.

So it seems very unlikely that anyone uses them for anything other than translation of Z by x.
And this translation doesn't change, only now these values make more sense, "x" being just above-sea-level elevation. And we no longer need to explain this weird offset .

If this approach looks ok, I'll try add some tests about custom layer rendering.


HarelM:
I would prefer to merge this when we work on 5.x just to be on the safe side.
I agree that the public API didn't change in terms of breaking the code, but the behavior changed and we can't know for sure how people are using this API call...

olsen232:
FYI this fixes https://github.com/maplibre/maplibre-gl-js/issues/3797
Or to put this another way #3736 and #3797 are duplicates or at least are two sides of the same coin

HarelM:
This is a breaking change, thus this PR has the 5.0 milestone.

acalcutt:
I am just curious, with this returning above-sea-level elevation, what happens when the elevation is below sea level? does it return a negative?

sbachinin:
@acalcutt , it will return an elevation from original dem data, multiplied by exxageration

birkskyum:
It might be better to not multiply the exaggeration then for the Above Sea Level query, and let exxageration be a way to tweak the visuals without effectively manipulating the DEM data.

HarelM:
I've created the following PR to do as this PR is doing:
https://github.com/maplibre/maplibre-gl-js/pull/4876
@sbachinin let me know if you prefer to update your branch or use the PR I created.
@kubapelc @sbachinin @NathanMOlson @ibesora I would greatly appreciate your input here now with all the globe changes...

sbachinin:
I don't mind if it will be done in another PR.
Surprised to see that this PR is still relevant after all that was done since v4 )

HarelM:
Better to have it here, I'll close mine once this is updated. Most of the comments here are relevant, not sure why I opened a different one.. :-/

sbachinin:
At the moment this branch contains a working solution for mercator map.
After merging of the globe, this solution (lowering of the matrix for custom layer) looks different from the original one. Now it's just a removal of a couple of lines from MercatorTransform where matrices were translated, and they no longer need to be translated.

In globe, as far as I understand, there is no support for terrain and so the models don't work too. At least they don't show up when I switch add-3d-model-with-terrain example to globe.

I've experimented with `GlobeTransform.getProjectionDataForCustomLayer`, replicating my logic from MercatorTransform and models began to appear (in globe projection and with globeness false). But I haven't enough understanding of the globe to prepare correct matrices for all cases - for true/false globeness, with/without terrain etc. Anyway I suppose it's too early to mess with this problem.

HarelM:
Looks good to me, can you fix the lint warning, and I'll merge this?
`Warning:   8:17  warning  'LngLatLike' is defined but never used  @typescript-eslint/no-unused-vars`

HarelM (approved):

HarelM:
Thanks, can you add a change log item with the breaking change icon please?

HarelM (approved):

PR #3865, opened 2024-03-19
Remove popup when marker is behind terrain

sbachinin:
Currently popups do not change opacity when hidden by terrain.
It looks especially bad when a popup is attached to a marker. Marker becomes semi-transparent, and popup doesn't.

By this PR I suggest to hide a "dependent" popup when its marker is behind terrain.
It's not the only possible solution but **1)** it's simple  and **2)** it actually seems more adequate than making a popup semi-transparent because such semi-transparent popup will create a lot of visual noise.

So here I make sure that
1) when marker gets hidden, it will close its popup
2) when marker is hidden, click will not open its popup

https://github.com/maplibre/maplibre-gl-js/assets/11789697/e48cc6f8-9abf-4551-be44-a8e977fc454c

This solution does not affect independent popups. They will still be fully opaque behind terrain.
It seems that visibility of dependent and independent popups are 2 different problems.
Visibility of a dependent popup should be determined by the marker.
Visibility of an independent popup should be determined by popup's own position relative to terrain.
It will require to somehow copy or reuse the bulky opacity-related code from marker.ts.

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/ui/marker.test.ts
+++ b/src/ui/marker.test.ts
@@ -1066,6 +1066,30 @@ describe('marker', () => {
         map.remove();
     });
 
+    test('Does not open a popup when behind 3d terrain', async () => {
+        const map = createMap();
+        const marker = new Marker()
+            .setLngLat([0, 0])
+            .addTo(map)
+            .setPopup(new Popup());
+
+        map.transform.lngLatToCameraDepth = () => .95;
+
+        map.terrain = {
+            getElevationForLngLatZoom: () => 0,
+            depthAtPoint: () => .92
+        } as any as Terrain;
+        map.fire('terrain');
+
+        await sleep(500);
+
+        marker.togglePopup();
+
+        expect(marker._popup.isOpen()).toBeFalsy();
+
+        map.remove();
+    });
+
     test('Marker\'s lng is wrapped when slightly crossing 180 with {renderWorldCopies: false}', () => {
         const map = createMap({width: 1024, renderWorldCopies: false});
         const marker = new Marker()
--- a/src/ui/marker.test.ts
+++ b/src/ui/marker.test.ts
@@ -1040,6 +1040,32 @@ describe('marker', () => {
         map.remove();
     });
 
+    test('Removes an open popup when going behind 3d terrain', async () => {
+        const map = createMap();
+        const marker = new Marker()
+            .setLngLat([0, 0])
+            .addTo(map)
+            .setPopup(new Popup());
+
+        await sleep(500);
+        marker.togglePopup();
+
+        expect(marker._popup.isOpen()).toBeTruthy();
+
+        map.transform.lngLatToCameraDepth = () => .95; // Mocking distance to marker
+
+        map.terrain = {
+            getElevationForLngLatZoom: () => 0,
+            depthAtPoint: () => .92
+        } as any as Terrain;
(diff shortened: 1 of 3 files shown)


src/ui/marker.ts
```diff
@@ -140,7 +140,7 @@ export class Marker extends Evented {
     _pitchAlignment: Alignment;
     _rotationAlignment: Alignment;
     _originalTabIndex: string; // original tabindex of _element
-    _opacity: string;
+    _fullOpacity: string;
```
HarelM, line 143: What's the reasoning behind this change?

HarelM:
Looks like a reasonable UX, the code is basically a one line change, so I don't have real comments.

src/ui/marker.test.ts
```diff
@@ -157,13 +157,15 @@ describe('marker', () => {
         expect(!marker.getPopup()).toBeTruthy();
     });
 
-    test('Marker#togglePopup opens a popup that was closed', () => {
+    test('Marker#togglePopup opens a popup that was closed', async () => {
```
HarelM, line 160: Why is the sleep needed in the tests? Has the change requires a sleep now?

src/ui/marker.ts
sbachinin: Only readability. This time I found it hard to understand in places like `this._element.style.opacity !== this._opacity`.

`_fullOpacity` gives more clue what's going on.

src/ui/marker.test.ts
sbachinin: Yes, now it's necessary. When togglePopup is called right away, it doesn't open a popup because its marker hasn't attained full opacity. (Opacity is set via frameAsync)

src/ui/marker.ts
HarelM: I'm not sure I find it more readable to be honest, if there's a full opacity I would expect partial or half opacity, but there isn't something like that.
Also does full means fully opac?
I tend to find it more confusing than helping... Sorry...

HarelM:
Also changelog is needed.

src/ui/marker.ts
sbachinin: Ok, done.
My reasoning was that this time when reading `this._opacity`, my first guess was that it's the "actual opacity of a marker", whereas it's only "one of possible opacities".

src/ui/marker.ts
```diff
@@ -515,6 +515,8 @@ export class Marker extends Evented {
     togglePopup(): this {
         const popup = this._popup;
 
+        if (this._element.style.opacity !== this._opacity) return this;
```
HarelM, line 518: Can you clarify this condition? I'm not sure I understand why when the opacity is different you ignore the togglePopup request...

HarelM (approved):

src/ui/marker.ts
HarelM: Can we use the following condition instead:
`if (this._element.style.opacity === this._opacityWhenCovered)`?

sbachinin: RIght, that will be better

HarelM:
Can you verify that the added scenario is covered in tests? According to the code coverage report it seems that it's not well covered...?

HarelM:
THANKS!!

Issue #3872, opened 2024-03-21
Unwanted pan-jump as requested pan action finishes (3D terrain)


olsen232:
This is not about inertia. Any panning that happens due to inertia is outside the scope of this bug -
so, this is not a duplicate of https://github.com/maplibre/maplibre-gl-js/issues/1303

I believe this is only happens when 3D terrain is enabled.

**maplibre-gl-js version**: v4.1.1

**browser**: Brave

### Steps to Trigger Behavior

Here is one way I've found of reliably triggering an unwanted pan.
(There may be others - this is a bug I've come across by using ML, and it seems to be better in some circumstances, worse in others, hard to say if this way of demonstrating it showcases the only relevant bug at its most extreme, or if there are other ways to trigger similar bugs or worse instances of this bug).

 1. Open `test/examples/3d-terrain.html`
 2. Zoom in the maximum amount (18).
 3. Drag the map around and let go.
 4. Observe that, on occasion, once the map has finished panning, it jumps back in the opposite direction, part way or all of the way back to where it started.

### Link to Demonstration

https://github.com/maplibre/maplibre-gl-js/assets/7425442/47a95225-1b5c-4815-b9e9-93a0049e2c09

### Expected Behavior
Pan should continue in the same direction until it drag stops / inertia runs out, then stop.

### Actual Behavior
Pan jumps back the way it came in a single-frame unwanted discontinuity.

olsen232:
On another PR, @prozessor13 who I believe wrote some of this code, wrote the following: https://github.com/maplibre/maplibre-gl-js/pull/3836#issuecomment-1999838674

> Beside that fix i notice still some small jumps at mouse-moveend. Now, that recalcuteZoom is fixed, i am pretty sure that this issue is located [here](https://github.com/maplibre/maplibre-gl-js/blob/7308da584248c76e17f20f1f17df0b63acd34ade/src/render/terrain.ts#L329). This issue can either be fixed with:
>
> - calculate correct position with ray/triangle intersection from camera to terrain-mesh
> - or the pragmatic, still not perfect way: grab more than only pixel from the coords-framebuffer and make some sort of interpolation to get a more accurate position.

I've done some digging and I mostly agree with Max - this is an issue caused by `Terrain.pointCoordinate` returning bad values. However, I don't think it's due to poor accuracy that could be improved by a larger sample - I think the data in `coords` framebuffer is simply stale.

The following line of code
[...]

PR #3873, opened 2024-03-21
Fix for #3872 - unwanted pan-jumps at end of pan

olsen232:
Always recompute coords framebuffer if matrix has changed -
any time that it is not mat4.exactEquals compared to how it was previously.
Don't wait for it to be not mat4.equals compared to how it was previously.

Fixes https://github.com/maplibre/maplibre-gl-js/issues/3872 (or at least improves it)

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
Can you clarify the fix? What's the difference between the two methods? Reference equal vs values equal?

HarelM:
Nevermind, I read the linked issue now and understand the fix. Can this cause perfirmance degradation since we are updating the coords buffer more frequently now?

olsen232:
Sure - this also occurred to me as a possible concern. I'll have a try at only adding the extra render at `moveend`, leave performance unchanged the rest of the time.
Do you have any pointers on how to do this? The cause of the bug is when the coords buffer is used in by the call to `pointLocation` in `recalculateZoom`. A potential fix would be, immediately before that call to `pointLocation`, call `drawCoords` in `draw_terrain.ts` and make sure the coords buffer is up to date: but I don't know if that's possible.

Firstly I don't know if `transform.ts` holds the right references so that it can do that - and if it did hold those references, would that be poor separation of concerns.
Secondly I don't yet know how the render loop works. I know in lots of frameworks, you can't simply call "render" - you have to call "request-re-render-at-earliest-opportunity" and wait for the framework to call it for you.

HarelM:
Not sure if the following is related, but it's worth to see if there's any affect:
- #3878

olsen232:
Pushed my next attempt at solving this without affecting performance.
What do you think?

prozessor13:
Hi, i think this is not sufficient. I think in every terrain.pointCoordinte call it should be checked, if the coords-framebuffer is in sync with the help of exactEquals. So something like this (pseudocode):

```
terrain.pointCoordinate(p: Point): MercatorCoordinate {
    this.painter.checkDepthAndCoords();
    ...
}

painter.checkDepthAndCoords() {
    if (!mat4.exactEquals(this.terrainFacilitator.matrix, this.transform.projMatrix))
        this.drawDepthAndCoords();
}

```

olsen232:
Makes sense. Done, please take another look.

src/render/painter.ts
```diff
@@ -468,6 +454,35 @@ export class Painter {
...
+     * Update the depth and coords framebuffers, if the contents of those frame buffers is out of date.
+     * If requireExact is false, then the contents of those frame buffers is not updated if it is close
+     * to accurate (that is, the camera has not moved much since it was updated last).
+     */
+    maybeDrawDepthAndCoords(requireExact: boolean) {
+        if (this.renderToTexture) {
+            this.renderToTexture.prepareForRender(this.style, this.transform.zoom);
+            // this is disabled, because render-to-texture is rendering all layers from bottom to top.
+            this.opaquePassCutoff = 0;
+
```
prozessor13, line 467: the rendertotexture stuff has complete nothing to do with depth or coords buffer, so i think you should that part leave where it was.

olsen232: ... what is it doing? why is the depth or coords buffer only rendered `if (this.renderToTexture)` if it's not related?

prozessor13: the rendertotexture stuff tells the webgl draw methods (for line, fill, background, raster layers) not to draw onto the screen framebuffer, instead they draw to a texture. All texutres are then drawn to the terrain-mesh later in the render method. The rendertotexture object is created with the terrain object. So you can say a 
    if (this.renderToTexture)
is equivalent to
    if (this.map.terrain)
both of that variables contains either an object-instance or are null.
Although both logic, rendertotexture and drawDepth/Coords, are in one If statement, they are two different things. Sorry if that contributed to the confusion

olsen232: Ok, done

src/render/painter.ts
```diff
@@ -468,6 +460,32 @@ export class Painter {
...
+    maybeDrawDepthAndCoords(requireExact: boolean) {
+        if (!this.style.map.terrain) {
+            return;
+        }
+        const prevMatrix = this.terrainFacilitator.matrix;
+        const currMatrix = this.transform.projMatrix;
+
+        // Update coords/depth-framebuffer on camera movement, or tile reloading
+        let doUpdate = this.terrainFacilitator.dirty;
+        doUpdate ||= requireExact ? !mat4.exactEquals(prevMatrix, currMatrix) : !mat4.equals(prevMatrix, currMatrix);
```
prozessor13, line 477: In my tests the exactEquals always fails because prevMatrix is of type Float32Array, and currMatrix is of type Float64Array.

src/render/painter.ts
```diff
@@ -468,6 +460,32 @@ export class Painter {
...
+        doUpdate ||= this.style.map.terrain.sourceCache.tilesAfterTime(this.terrainFacilitator.renderTime).length > 0;
+
+        if (doUpdate) {
+            mat4.copy(prevMatrix, currMatrix);
```
prozessor13, line 481: You have to explicit create a Float64Array when copy the matrix. I do not know the internal details but a code like the following works:
```
this.terrainFacilitator.matrix = mat4.identity(new Float64Array(16) as any);
mat4.copy(this.terrainFacilitator.matrix, currMatrix);
```

olsen232: Done. Note that the matrix is only created once and maintains its type - you don't need to recreate it each time you copy it.

src/render/painter.ts
```diff
@@ -468,6 +460,32 @@ export class Painter {
...
+        }
+        const prevMatrix = this.terrainFacilitator.matrix;
+        const currMatrix = this.transform.projMatrix;
+
+        // Update coords/depth-framebuffer on camera movement, or tile reloading
+        let doUpdate = this.terrainFacilitator.dirty;
+        doUpdate ||= requireExact ? !mat4.exactEquals(prevMatrix, currMatrix) : !mat4.equals(prevMatrix, currMatrix);
+        doUpdate ||= this.style.map.terrain.sourceCache.tilesAfterTime(this.terrainFacilitator.renderTime).length > 0;
+
+        if (doUpdate) {
```
HarelM, line 480: Condider returning here to avoid code indentation and reduce complexity.

src/render/terrain.ts
```diff
@@ -327,6 +327,9 @@ export class Terrain {
      * @returns mercator coordinate for a screen pixel
      */
     pointCoordinate(p: Point): MercatorCoordinate {
+        // First, ensure the coords framebuffer is up to date.
```
HarelM, line 330: Isn't this called on every movement of the map in the handler manager code? I would consider to first call the only in the recalculate zoom method or make sure this is not called "too many times". I might be looking at this wrong though...

prozessor13: Hi Harel,

i also thought about this, but i think it is ok. Because pointCoordinate is called on very different positions, i think it is needed to check here, at the root, if the coords-framebuffer is uptodate. Elsewise it think it comes to race-conditions where the coords-framebuffer is not in sync.

And i think the doUpdate check is very cheap, and most of the time it should return false.
The `maybeDrawDepthAndCoords` Method may do a litte bit too much, for example updating the depth-framebuffer or checking for new tiles, which is not strictly needed in this scenario, but i think it should be ok.

olsen232: It is called quite often, but to defend this approach:
- it's not called continuously while the map is moving - it's called a few times at the start and end of map pans but then not during the middle of the pan.
- this code doesn't do anything or take up any CPU if it's called repeatedly while the map is not moving
- if you call pointCoordinate *without* ensuring the coords framebuffer is up to date, it noticeably gives you the wrong answer. 

However, ultimately this is for you the reviewers to decide - #3872 is fixed as long as this method is called immediately before the recalculateZoom code, so only calling it then would be one option... or we could work harder at making sure this function is cheap to call in the event that nothing has changed, but I think it's already pretty cheap - just comparing two arrays of length 16 for exact equality.

HarelM: Fair enough. Let's leave it as is.

HarelM (approved):

prozessor13 (approved):

HarelM (approved):

PR #1986, opened 2022-12-24
Build process cleanup

rotu:
- Moved code generation from postinstall to prepare hook
- Remove unused file (that typescript didn't like anyway)
- Unify typescript config
- Remove top-level rollup directory

PR #3879, opened 2024-03-21
Use const enum more often to improve readability and reduce generated script size as byproduct

zhangyiatmicrosoft:
## Launch Checklist

I noticed that MapLibre in general prefers plain strings for message/status/state. This can result in repetitive strings everywhere and eventually the generated JS is more bloated than necessary.
Readibility is certainly a concern, but we can solve both problems with const enum, which is descriptive in TS and at compile time replaced with abbreviations in favor of script size.

In fact, plain strings are less friendly for readability and code searching -- for example the common word 'loaded' is being used by both tile state and RTLPluginState. If one were to search 'loaded', they both show up. This can be addressed by const enum as well

This draft PR is a staring point --- I only defined the const enum as an illustration. Will finish when everyone is ok with this direction.

Rough calculation: current MessageType has 21 keys, average length is 13. These keys are being repeated 46 times in the code base, which means 598 bytes.
With const enum of 2 or 3 chars, we can reduce script size by 500 bytes with just MessageType. 

If this practice is widely adopted throughout the code base, noticeable size reduction can be archived. Thanks.


HarelM:
I found enum in type typescript and javascript a total miss in terms of ease of use and ergonomics in general.
While it might prove to reduce bundle size I'm not sure it will be meaningful.
I would also like to avoid breaking the buplic API for this if possible.
The package is gzipped and this it might prove even smaller.
The globe for example adds like 20k IIRC, so I'm not 100% it's work the time, but, your call...

zhangyiatmicrosoft:
> I found enum in type typescript and javascript a total miss in terms of ease of use and ergonomics in general. While it might prove to reduce bundle size I'm not sure it will be meaningful. I would also like to avoid breaking the buplic API for this if possible. The package is gzipped and this it might prove even smaller. The globe for example adds like 20k IIRC, so I'm not 100% it's work the time, but, your call...

Agree enum are very bad.
"const enum" are different and they are great.
I like the most is that <code>const enum</code> is a pure TS thing, and completely replaced with plain strings in compilation and still maintain readability. For example, for TS code:
<code>
/** Parameters for a vertex */
const enum VertexParameter
{
    /** Each vertex position data has 4 numbers: xyzw, */
    PositionElements = 4,
</code>
compiled JS looks like
<code>
for (var j = 0; j < 4 /* VertexParameter.PositionElements */; j++) {
</code>

so IMO it is the best for both worlds.

HarelM:
I don't object in general as long as we are not changing public APIs.

zhangyiatmicrosoft:
In order to take advantage of const enum, isolatedModules must be false. It is currently true and was added by this PR
https://github.com/maplibre/maplibre-gl-js/pull/1986/files
<pre>
"isolatedModules": true,
</pre>
This flag is intended for "...other transpilers only operate on a single file at a time". see more:
https://www.typescriptlang.org/tsconfig#isolatedModules

@HarelM  @rotu are we expecting to ever use other compilers?

HarelM:
I think we tried to yse swc to improve build time at some point in time, but it was too buggy.
In any case, if it's not needed right now, there's no reason to keep it, if needed, we can add it back in the future.

HarelM:
Can you check how this change affects the generated d.ts file?
How would `isolatedModules` affect consuming libraries?
This looks OK, although I'm not sure it's worth the effort given the changes that are expected to land in the coming version with the globe and projection...

zhangyiatmicrosoft:
> Can you check how this change affects the generated d.ts file? How would `isolatedModules` affect consuming libraries? This looks OK, although I'm not sure it's worth the effort given the changes that are expected to land in the coming version with the globe and projection...

- d.ts --- I am very surprised that things like <code>MessageType</code> are even in the generated d.ts, since people outside should not care and should not use it anyway. Nevertheless, I modified the type name to minimize the changes. Here is a sample PR to illustrate exactly what will be changed:
https://github.com/zhangyiatmicrosoft/my-maplibre-gl-js/pull/9/files
We can easily mark these internal APIs with <code>@internal</code> and TS will remove them from d.ts, which is the best approach IMO. Let me know.

- How would `isolatedModules` affect consuming libraries? --- it does not. According to TS link shared earlier <code>...It does not change the behavior of your code, or otherwise change the behavior of TypeScript’s checking and emitting process.</code>

- coming version with the globe and projection... --- is there a pending PR for that? Again, this PR is much more for the benefit of devs to improve readability, code searching, and debugging plain JS without sourcemap, which are a very common scenarios. Script side reduction is only a byproduct.

HarelM:
> We should mark these internal APIs with @internal and TS will exclude them while generating d.ts,

I don't disagree, but the d.ts generation and the typedoc are not easy to please when it comes to internal and public stuff...

If the main driving force is dev experience then I'm all for it.

Globe PRs: 
- #3882
- #3783

HarelM (approved):

zhangyiatmicrosoft:
> > We should mark these internal APIs with @internal and TS will exclude them while generating d.ts,
> 
> I don't disagree, but the d.ts generation and the typedoc are not easy to please when it comes to internal and public stuff...
> 
> If the main driving force is dev experience then I'm all for it.
> 
> Globe PRs:
> 
> * [Globe - fill layer #3882](https://github.com/maplibre/maplibre-gl-js/pull/3882)
> * [Globe - basic infrastructure, raster layer adaptation for globe #3783](https://github.com/maplibre/maplibre-gl-js/pull/3783)

Thanks. Separately I will probably investigate <codde>@internal</code>.

zhangyiatmicrosoft:
please merge at your conveniences. thanks

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@
 - Fix type definition on `localIdeographFontFamily` ([#3896](https://github.com/maplibre/maplibre-gl-js/pull/3896))
 - Fix unwanted panning changes at the end of a panning motion ([#3872](https://github.com/maplibre/maplibre-gl-js/issues/3872))
 - Fix `close` events being fired for popups that aren't open
+- Added const enum for actor messages to improve readability and maintainability. In tsconfig.json, `isolatedModules` flag is set to false in favor of generated JS size. ([#3879](https://github.com/maplibre/maplibre-gl-js/issues/3879))
```
HarelM, line 20: This is not the right place... Version 4.1.2 was released.

zhangyiatmicrosoft: oops. My bad

HarelM (approved):

PR #3882, opened 2024-03-22
Kubapelc/globe pr

kubapelc:
# Fill layer adaptation for globe

Issue & discussion of globe:

Additions and changes since [the previous PR](https://github.com/maplibre/maplibre-gl-js/pull/3783):

- Adapted fill layer type for globe
- Added `subdivision.ts`, a collection of functions for subdividing geometry to allow smooth projection to globe, including many unit tests. Used by the fill layer.
- Subdivision granularity settings are now part of the Projection interface (instead of being static/global), and are passed to `*_bucket.ts`, where subdivision is called
- Added render tests for fill

I've decided to leave line and fill-extrusion layers to a later PR to make this PR smaller.

## Example

I've added a globe example page with vector tiles: `test/examples/globe-vectortiles.html`. Note that only the raster and fill layers are adapted for globe as of now, so all other layers (most visibly line and symbol) will display as if they were on a mercator map.

A vector globe demo image

## Checklist

 - [x] Link to related issues.
   - added render tests for fill layer on globe
   - added unit tests for subdivision and `fillArrays()`
 - [ ] Post benchmark scores.


test/integration/render/tests/projection/globe/fill-planet-tiles/style.json
```diff
@@ -0,0 +1,40 @@
...
+      "projection": "globe"
+    }
+  },
+  "center": [
+    0.0,
+    0.0
+  ],
+  "zoom": 2,
+  "sources": {
+    "mapbox": {
```
HarelM, line 14: can we use maplibre instead of mapbox? I'm trying to minimize the number of appearances of this word in this repo.

src/util/classify_rings.ts
```diff
@@ -4,8 +4,10 @@ import {calculateSignedArea} from './util';
 
 import type Point from '@mapbox/point-geometry';
 
-// classifies an array of rings into polygons with outer rings and holes
-export function classifyRings(rings: Array<Array<Point>>, maxRings: number) {
+// Classifies an array of rings into polygons with outer rings and holes.
```
HarelM, line 7: Use `/**` type of comment.

test/bench/lib/tile_parser.ts
```diff
@@ -15,6 +15,7 @@ import type {OverscaledTileID} from '../../../src/source/tile_id';
 import type {TileJSON} from '../../../src/types/tilejson';
 import type {Map} from '../../../src/ui/map';
 import type {IActor} from '../../../src/util/actor';
+import {subdivisionGranularitySettingsNoSubdivision} from '../../../src/render/subdivision';
```
HarelM, line 18: This is a very long type name and does not adhere to casing convention I believe.

src/source/worker_tile.ts
```diff
@@ -23,6 +23,7 @@ import type {
 import type {PromoteIdSpecification} from '@maplibre/maplibre-gl-style-spec';
 import type {VectorTile} from '@mapbox/vector-tile';
 import type {GetGlyphsResponse, GetImagesResponse} from '../util/actor_messages';
+import {SubdivisionGranularitySetting} from '../render/subdivision';
```
HarelM, line 26: `import type` I think as you do not create an instance of this type here.

src/source/vector_tile_source.test.ts
```diff
@@ -17,7 +18,8 @@ function createSource(options, transformCallback?, clearTiles = () => {}) {
         _getMapId: () => 1,
         _requestManager: new RequestManager(transformCallback),
         style: {sourceCaches: {id: {clearTiles}}},
-        getPixelRatio() { return 1; }
+        getPixelRatio() { return 1; },
+        projection: new MercatorProjection()
```
HarelM, line 22: [Optional] Consider creating something more "mocky" as initializing a MercatorProjection might require all kinds of quirks in the future.

src/render/subdivision.ts
```diff
@@ -21,26 +30,1032 @@ export class SubdivisionGranularityExpression {
...
-export const granularitySettings: SubdivisionGranularitySetting = new SubdivisionGranularitySetting({
-    fill: new SubdivisionGranularityExpression(128, 1),
-    line: new SubdivisionGranularityExpression(512, 1)
+register('SubdivisionGranularityExpression', SubdivisionGranularityExpression);
+register('SubdivisionGranularitySetting', SubdivisionGranularitySetting);
+
+/**
+ * Granularity settings that disable subdivision altogether.
+ */
+export const subdivisionGranularitySettingsNoSubdivision = new SubdivisionGranularitySetting({
```
HarelM, line 78: consider naming this `noSubdivisionGranularity`

src/render/subdivision.ts
```diff
@@ -21,26 +30,1032 @@ export class SubdivisionGranularityExpression {
...
+        return index;
+    }
+
+    /**
+     * Subdivides a polygon by iterating over rows of granularity subdivision cells. See comments inside the function for more details.
+     * @param inputIndices - Indices into the internal vertex buffer of the triangulated polygon (after running `earcut`).
+     * @returns Indices into the internal vertex buffer for triangles that are a subdivision of the input geometry.
+     */
+    private _subdivideTrianglesScanline(inputIndices: Array<number>): Array<number> {
+        // A granularity cell is the square space between axes that subdivide geometry.
```
HarelM, line 153: Consider moving the comments to the method's definition comment above.

src/render/subdivision.ts
```diff
@@ -21,26 +30,1032 @@ export class SubdivisionGranularityExpression {
...
+                        const x = aX + dirX * tExit;
+                        const y = aY + dirY * tExit;
+                        if (x < leftmostX) {
+                            leftmostX = x;
+                            leftmostIndex = ring.length;
+                        }
+                        ring.push(this._getVertexIndex(x, y));
+                    }
+
+                    // When to split inter-edge boundary segments?
```
HarelM, line 322: Seems like this method is too long anyway.
Please split it to smaller method, each one with its own responsibility.
Check out [SOLID](https://en.wikipedia.org/wiki/SOLID) principles.

src/data/segment.ts
```diff
@@ -25,32 +25,74 @@ export type Segment = {
...
+            return this.createNewSegment(layoutVertexArray, indexArray, sortKey);
+        } else {
+            return lastSegment;
+        }
+    }
+
+    /**
+     * Creates a new empty segment and returns it.
+     */
+    createNewSegment(
```
HarelM, line 60: Was this only a refactoring or are there tests missing here?

src/data/bucket/fill_attributes.ts
```diff
@@ -1,8 +1,5 @@
 import {createLayout} from '../../util/struct_array';
 
-const layout = createLayout([
+export const layout = createLayout([
```
HarelM, line 3: Any reason to change this? Seems like the others are default exported, why not this one?

src/data/bucket.ts
```diff
@@ -8,6 +8,7 @@ import type {ImagePosition} from '../render/image_atlas';
 import type {CanonicalTileID} from '../source/tile_id';
 import type {VectorTileFeature, VectorTileLayer} from '@mapbox/vector-tile';
 import Point from '@mapbox/point-geometry';
+import {SubdivisionGranularitySetting} from '../render/subdivision';
```
HarelM, line 11: `import type` I believe.

HarelM:
`subdevision.ts` is a large file with long methods. I would advise to break the methods a bit and also improve coverage for this file as it seems it is 70% covered I would say that 80-90 is where we should aim.

test/integration/render/tests/projection/globe/fill-planet-tiles/style.json
kubapelc: I've renamed the source to "vector_tiles" in all new render tests.

src/render/subdivision.ts
kubapelc: I've refactored it into a static readonly field inside the `SubdivisionGranularitySetting` class and renamed it to `noSubdivision`.

src/source/worker_tile.ts
kubapelc: I've replaced `import` with `import type` whereever applicable.

src/render/subdivision.ts
kubapelc: The comment just details the internal workings of the function, and not what the function does or how it should be used, so I'd prefer to keep in inside the function instead of moving it to the doc comment.

src/render/subdivision.ts
kubapelc: I've broken the scanline subdivision function down a bit, I don't think it makes sense to split any other functions in that file.

src/data/segment.ts
kubapelc: It was both refactoring and adding a bit of new functionality. SegmentVector didn't actually have any tests. I've now added unit tests that should cover all of its functionality.

src/data/bucket/fill_attributes.ts
kubapelc: I've reverted it to its original state with default export. This was a leftover change from some experiements I later reverted, and I seem to have forgotten to revert this as well.

kubapelc:
> `subdevision.ts` is a large file with long methods. I would advise to break the methods a bit and also improve coverage for this file as it seems it is 70% covered I would say that 80-90 is where we should aim.

I've added a lot more unit tests for subdivision, I think all functions should be covered now, with the exception of the SVG generation, but that is only useful for debugging and not used as of now. Should I remove it?

src/data/bucket/fill_bucket.ts
```diff
@@ -168,58 +170,25 @@ export class FillBucket implements Bucket {
...
-            const holeIndices = [];
-
-            for (const ring of polygon) {
-                if (ring.length === 0) {
-                    continue;
+            const subdivided = subdivideFill(polygon, canonical, subdivisionGranularity.fill.getGranularityForZoomLevel(canonical.z));
+
+            const vertexArray = this.layoutVertexArray;
+
+            fillArrays(
```
HarelM, line 179: I think this method name is a bit too generic, I would consider making it a bit more specific.

src/data/segment.test.ts
```diff
@@ -0,0 +1,200 @@
...
+        expect(third).toBeTruthy();
+        expect(first === second).toBe(false);
+        expect(second === third).toBe(true);
+        expect(first.vertexLength).toBe(5);
+        expect(second.vertexLength).toBe(10);
+        expect(segmentVector.segments).toHaveLength(2);
+    });
+});
+
+class VirtualVertexBuffer {
```
HarelM, line 162: If possible, I prefer to use jest's mocking capabilities.
I'm not sure it can be applied in this case, but I would like to avoid creating classes in the test code as much as possible.

src/geo/projection/projection.ts
```diff
@@ -70,6 +71,13 @@ export interface Projection {
      */
     get vertexShaderPreludeCode(): string;
 
+    /**
+     * @internal
+     * An object describing how much subdivision should be applied to rendered geometry.
+     * The subdivision settings should be a constant for a given projection.
```
HarelM, line 77: [Optinal] Given the fact that developers would need to implement this, I would consider adding a comment about using the noSubdevision if it is not explicitly required.

src/render/fill_arrays.test.ts
```diff
@@ -0,0 +1,324 @@
...
+    test('Very large random mesh is correctly split into multiple segments.', () => {
+        SegmentVector.MAX_VERTEX_ARRAY_LENGTH = 1024;
+        const mesh = getGridMeshRandom(64, 8192, 1024);
+        const split = splitMesh(mesh);
+        expect(split.segmentsTriangles.length).toBeGreaterThan(1);
+        testMeshesEqual(mesh, split);
+    });
+});
+
+class VirtualVertexBuffer {
```
HarelM, line 49: These are duplicated I believe from another test file, right?
If this is the case, and you need this in more than one file I would consider adding them to a test util file.

src/render/fill_arrays.test.ts
```diff
@@ -0,0 +1,324 @@
...
+    const stringsExpected = getStrings(expected);
+    const stringsActual = getStrings(actual);
+    expect(stringsActual.stringsTriangles).toEqual(stringsExpected.stringsTriangles);
+    expect(stringsActual.stringsLines).toEqual(stringsExpected.stringsLines);
+}
+
+/**
+ * Returns a string representation of the geometry of the mesh.
+ */
+function getStrings(mesh: SimpleMesh) {
```
HarelM, line 139: Isn't `JSON.strigify` good enough?

src/render/fill_arrays.test.ts
```diff
@@ -0,0 +1,324 @@
...
+        t = t ^ t >>> 15;
+        t = Math.imul(t, 0x735a2d97);
+        return ((t = t ^ t >>> 15) >>> 0) / 4294967296;
+    };
+}
+
+/**
+ * Generates a mesh with the vertices of a grid, but random triangles and lines.
+ */
+function getGridMeshRandom(size: number, triangleCount: number, lineCount: number): SimpleMesh {
```
HarelM, line 269: There's a lot of logic in this test file, it might be even that this logic has its own bugs.
I think there is some mocking of webgl stuff that is being used in some tests, is it possible to using these mocks instead of using all this logic (although the webgl mocking can have bugs too, it's hopefully more widely used and we can also contribute to it and fix bugs if they exists).

src/render/fill_arrays.ts
```diff
@@ -0,0 +1,253 @@
...
+            triangleIndexArray.emplaceBack(
+                triangleIndex + triangleIndices[i],
+                triangleIndex + triangleIndices[i + 1],
+                triangleIndex + triangleIndices[i + 2]);
+        }
+
+        triangleSegment.vertexLength += numVertices;
+        triangleSegment.primitiveLength += triangleIndices.length / 3;
+
+        let lineIndicesStart;
```
HarelM, line 39: What type are those two variables?

src/render/fill_arrays.ts
```diff
@@ -0,0 +1,253 @@
...
+        for (let i = 0; i < flattened.length; i += 2) {
+            addVertex(flattened[i], flattened[i + 1]);
+        }
+
+        if (segmentsLines && lineIndexArray) {
+            for (let listIndex = 0; listIndex < lineList.length; listIndex++) {
+                const lineIndices = lineList[listIndex];
+
+                for (let i = 1; i < lineIndices.length; i += 2) {
+                    lineIndexArray.emplaceBack(
```
HarelM, line 59: Why is this different than `addVertex`? i.e. I tend to say that either both do emplaceBack or both are abstracted out, but this doesn't feel "the same" right now.

src/render/fill_arrays.ts
```diff
@@ -0,0 +1,253 @@
...
+                triangleIndex + triangleIndices[i + 2]);
+        }
+
+        triangleSegment.vertexLength += numVertices;
+        triangleSegment.primitiveLength += triangleIndices.length / 3;
+
+        let lineIndicesStart;
+        let lineSegment;
+
+        if (segmentsLines && lineIndexArray) {
```
HarelM, line 42: This condition seems to repeat a few times, I would consider creating a boolean to hold it, hopefully with more meaningful name as currently it's not intuitive what this condition checks.

src/render/fill_arrays.ts
```diff
@@ -0,0 +1,253 @@
...
+import {SegmentVector} from '../data/segment';
+import {StructArray} from '../util/struct_array';
+
+/**
+ * This function will take any "mesh" and fill in into vertex buffers, breaking it up into multiple drawcalls as needed
+ * if too many (\>65535) vertices are used.
+ * This function is mainly intended for use with subdivided geometry, since sometimes subdivision might generate
+ * more vertices than what fits into 16 bit indices.
+ */
+export function fillArrays(
```
HarelM, line 11: It seems that part of the parameters for this method can be null or can be made optional, I would condiser moving them to be the last parameters and add a "?" to indicate that they are optional.

src/render/fill_arrays.ts
```diff
@@ -0,0 +1,253 @@
+import {LineIndexArray, TriangleIndexArray} from '../data/array_types.g';
+import {SegmentVector} from '../data/segment';
+import {StructArray} from '../util/struct_array';
+
+/**
+ * This function will take any "mesh" and fill in into vertex buffers, breaking it up into multiple drawcalls as needed
```
HarelM, line 6: Consider adding a `@param` to explain what each parameter does/responsible for.

src/render/fill_arrays.ts
```diff
@@ -0,0 +1,253 @@
...
+            i1needsVertexCopy = true;
+            i2needsVertexCopy = true;
+            baseVertex = 0;
+        }
+
+        let actualIndex0 = -1;
+        let actualIndex1 = -1;
+        let actualIndex2 = -1;
+
+        if (i0needsVertexCopy) {
```
HarelM, line 139: This part seems somewhat duplicated, I'm not sure there's an easy way to remove this duplication while still keeping it readable, but I would suggest to try.

src/render/fill_arrays.ts
```diff
@@ -0,0 +1,253 @@
...
+                currentSegmentCutoff = totalVerticesCreated;
+                i0needsVertexCopy = true;
+                i1needsVertexCopy = true;
+                baseVertex = 0;
+            }
+
+            let actualIndex0 = -1;
+            let actualIndex1 = -1;
+
+            if (i0needsVertexCopy) {
```
HarelM, line 225: Same as above with the duplication, but this time it's for "2" and not for "3".
I would see if there's an elegant way to solve this for both cases and reduce duplication.

src/render/subdivision.ts
```diff
@@ -1,46 +1,1044 @@
...
+}
+
+/**
+ * Returns a SVG image (as string) that displays the supplied triangles and lines. Only vertices used by the triangles are included in the svg.
+ * @param flattened - Array of flattened vertex coordinates.
+ * @param triangles - Array of triangle indices.
+ * @param edges - List of arrays of edge indices. Every pair of indices forms a line. A single triangle would look like `[[0 1 1 2 2 0]]`.
+ * @returns SVG image as string.
+ */
+export function getDebugSvg(flattened: Array<number>, triangles?: Array<number>, edges?: Array<Array<number>>, granularity: number = 1): string {
```
HarelM, line 901: Consider moving this to some test utils file as this should not be part of a production code file.

src/render/subdivision.ts
```diff
@@ -1,46 +1,1044 @@
...
+}
+
+/**
+ * Triangulates a ring of vertex indices. Appends to the supplied array of final triangle indices.
+ * @param vertexBuffer - Flattened vertex coordinate array.
+ * @param ring - Ordered ring of vertex indices to triangulate.
+ * @param leftmostIndex - The index of the leftmost vertex in the supplied ring.
+ * @param finalIndices - Array of final triangle indices, into where the resulting triangles are appended.
+ */
+export function scanlineTriangulateVertexRing(vertexBuffer: Array<number>, ring: Array<number>, leftmostIndex: number, finalIndices: Array<number>): void {
```
HarelM, line 964: Should this method get its own file?

src/render/subdivision.ts
```diff
@@ -1,46 +1,1044 @@
...
+    // Triangulate the ring
+    // It is guaranteed to be convex and ordered
+    if (ring.length === 0) {
+        console.error('Subdivision vertex ring length 0, smells like a bug!');
```
HarelM, line 968: Shouldn't this throw?

src/render/subdivision.ts
HarelM: Shouldn't this method return the `finalIndices` instead of receiving and filling it?

src/render/subdivision.ts
```diff
@@ -1,46 +1,1044 @@
...
+        holeIndices
+    };
+}
+
+/**
+ * Returns a new array of indices where all triangles have the counter-clockwise winding order.
+ * @param flattened - Flattened vertex buffer.
+ * @param indices - Triangle indices.
+ */
+export function fixWindingOrder(flattened: Array<number>, indices: Array<number>): Array<number> {
```
HarelM, line 856: This feels like it should already exist somewhere in the code, doesn't it?

src/render/subdivision.ts
```diff
@@ -1,46 +1,1044 @@
...
 
-export const granularitySettings: SubdivisionGranularitySetting = new SubdivisionGranularitySetting({
-    fill: new SubdivisionGranularityExpression(128, 1),
-    line: new SubdivisionGranularityExpression(512, 1)
-});
+/**
+ * Takes a polygon as an array of point rings, returns a flattened array of the X,Y coordinates of these points.
+ * Also creates an array of hole indices. Both returned arrays are required for `earcut`.
+ */
+function flatten(polygon: Array<Array<Point>>) {
```
HarelM, line 826: Add return type to the method signature.
Also note that javascript has some means of flatenning, I'm not sure they can be used here, but it's worth checking.

src/render/subdivision.ts
```diff
@@ -1,46 +1,1044 @@
...
+ * //   0, 0
+ * //   4, 0
+ * //   8, 0
+ * //   4, 4
+ * //   0, 8
+ * //   0, 4
+ * //   0, 0
+ * ```
+ */
+export function subdivideVertexLine(linePoints: Array<Point>, granularity: number, isRing: boolean = false): Array<Point> {
```
HarelM, line 707: Maybe `turf` can do this?

src/render/subdivision.ts
```diff
@@ -1,46 +1,1044 @@
...
+ * @param generateOutlineLines - When true, also generates index arrays for subdivided lines that form the outline of the supplied polygon. True by default.
+ * @returns An object that contains the generated vertex array, triangle index array and, if specified, line index arrays.
+ */
+export function subdivideFill(polygon: Array<Array<Point>>, canonical: CanonicalTileID, granularity: number, generateOutlineLines: boolean = true): SubdivisionResult {
+    const subdivider = new Subdivider(granularity, canonical);
+    return subdivider.subdivideFillInternal(polygon, generateOutlineLines);
+}
+
+/**
+ * Returns an array of line indices for rendering a wireframe of the supplied triangle array. Useful for debugging.
```
HarelM, line 611: For debugging like with tile borders and show collision boxes type of debug?

src/render/subdivision.ts
```diff
@@ -1,46 +1,1044 @@
...
+        this._granularity = granularity;
+        this._granularityCellSize = EXTENT / granularity;
+        this._canonical = canonical;
     }
 
-    public getGranularityForZoomLevel(zoomLevel: number): number {
-        const divisor = 1 << zoomLevel;
-        return Math.max(Math.floor(this._baseZoomGranularity / divisor), this._minGranularity, 0);
+    private _getKey(x: number, y: number) {
+        // Assumes signed 16 bit positions.
```
HarelM, line 51: This feels like it can easily break, are we OK with this assumption?

HarelM: Can't we use strings for key creation?

src/render/subdivision.ts
```diff
@@ -1,46 +1,1044 @@
...
 
-export class SubdivisionGranularitySetting {
     /**
-     * granularity settings used for fill layer (both polygons and their anti-aliasing outlines).
+     * Returns an index into the internal vertex buffer for a vertex at the given coordinates.
+     * If the internal vertex buffer contains no such vertex, then it is added.
      */
-    public readonly fill;
+    private _getVertexIndex(x: number, y: number): number {
+        const xInt = Math.round(x) | 0;
```
HarelM, line 62: Why do we need the `| 0`?

src/render/subdivision.ts
```diff
@@ -1,46 +1,1044 @@
...
+                const x = aX + dirX * tExit;
+                const y = aY + dirY * tExit;
+                if (x < leftmostX) {
+                    leftmostX = x;
+                    leftmostIndex = ring.length;
+                }
+                ring.push(this._getVertexIndex(x, y));
+            }
+
+            // When to split inter-edge boundary segments?
```
HarelM, line 269: This method is still too long to my taste.
Try and see if you can create small methods to facilitate for the below cases described in these comments.

src/render/subdivision.ts
```diff
@@ -1,46 +1,1044 @@
...
+    /**
+     * Checks the internal vertex buffer for all vertices that might lie on the special pole coordinates and shifts them by one unit.
+     * Use for removing unintended pole vertices that might have been created during subdivision. After calling this function, actual pole vertices can be safely generated.
+     */
+    private _ensureNoPoleVertices() {
+        const flattened = this._vertexBuffer;
+
+        // Special pole vertices have Y coordinate -32768 for the north pole and 32767 for the south pole.
+        // First, find any *non-pole* vertices at those coordinates and move them slightly elsewhere.
+        const northY = NORTH_POLE_Y;
```
HarelM, line 386: why do we need these extra two variables? Can't we simply use the const definition?

src/render/subdivision.ts
```diff
@@ -1,46 +1,1044 @@
...
+        const flattened = this._vertexBuffer;
+
+        const northEdge = 0;
+        const southEdge = EXTENT;
+
+        // Generates a quad from an edge to a pole with the correct winding order.
+        // i0, i1 - indices of the edge vertices
+        // v0x, v1x - X coordinates of the edge vertices
+        // poleY - the Y coordinate of the desired pole (NORTH_POLE_Y or SOUTH_POLE_Y)
+        const generatePoleQuad = (i0, i1, v0x, v1x, poleY) => {
```
HarelM, line 422: Why is this method defined inside another method? Consider moving it to be a private or private static method.

src/render/subdivision.ts
```diff
@@ -1,46 +1,1044 @@
...
+            }
+        }
+    }
+
+    /**
+     * Adds all vertices in the supplied flattened vertex buffer into the internal vertex buffer.
+     */
+    private _initializeVertices(flattened: Array<number>) {
+        for (let i = 0; i < flattened.length; i += 2) {
+            this._getVertexIndex(flattened[i], flattened[i + 1]);
```
HarelM, line 486: how is a `_get...` method initializing something? I guess it should have a better name...

src/render/subdivision.ts
```diff
@@ -1,46 +1,1044 @@
...
+    /**
+     * Subdivides an input mesh. Imagine a regular square grid with the target granularity overlaid over the mesh - this is the subdivision's result.
+     * Assumes a mesh of tile features - vertex coordinates are integers, visible range where subdivision happens is 0..8192.
+     * @param polygon - The input polygon, specified as a list of vertex rings.
+     * @param generateOutlineLines - When true, also generates line indices for outline of the supplied polygon.
+     * @returns Vertex and index buffers with subdivision applied.
+     */
+    public subdivideFillInternal(polygon: Array<Array<Point>>, generateOutlineLines: boolean): SubdivisionResult {
+        if (this._used) {
+            console.error('Subdivider: multiple use not allowed.');
```
HarelM, line 499: I'm not sure about these `console.error`, I prefer `throw` if possible.

src/render/subdivision.ts
```diff
@@ -1,46 +1,1044 @@
...
+            const cut = this._convertIndices(flattened, earcutResult);
+            subdividedTriangles = this._subdivideTrianglesScanline(cut);
+        } catch (e) {
+            console.error(e);
+        }
+
+        // Subdivide lines
+        const subdividedLines = [];
+        if (generateOutlineLines) {
+            for (const ring of polygon) {
```
HarelM, line 522: this for should be part of the generateOutline method I believe...? If such method doesn't exist, please create one.

src/render/subdivision.ts
```diff
@@ -1,46 +1,1044 @@
...
+        if (this._canonical) {
+            if (this._canonical.y === 0) {
+                north = true;
+            }
+            if (this._canonical.y === (1 << this._canonical.z) - 1) {
+                south = true;
+            }
+        }
+        if (north || south) {
+            this._fillPoles(subdividedTriangles, north, south);
```
HarelM, line 553: Consdier moving all the logic above this method into this method so that the code will read something like:
```
this._ensureNoPoleVertices();
this._fillPolesIfNeeded(...);
```

src/render/subdivision.ts
```diff
@@ -1,46 +1,1044 @@
...
+            verticesFlattened: this._vertexBuffer,
+            indicesTriangles: subdividedTriangles,
+            indicesLineList: subdividedLines,
+        };
+    }
+
+    /**
+     * Sometimes the supplies vertex and index array has duplicate vertices - same coordinates that are referenced by multiple different indices.
+     * That is not allowed for purposes of subdivision, duplicates are removed in `this.initializeVertices`.
+     * This function checks all indices, and replaces any index that references a duplicate vertex with the an index that vertex that is actually valid in `this._finalVertices`.
```
HarelM, line 566: "the an index" - no need for the "the" I guess.

src/render/subdivision.ts
```diff
@@ -1,46 +1,1044 @@
...
+     * @param vertices - Flattened vertex array used by the old indices. This may contain duplicate vertices.
+     * @param oldIndices - Indices into the supplied vertex array.
+     * @returns Indices transformed so that they are valid indices into `this._finalVertices` (with duplicates removed).
+     */
+    private _convertIndices(vertices: Array<number>, oldIndices: Array<number>): Array<number> {
```
HarelM, line 571: `_deduplicateIndices`?

src/render/subdivision.ts
kubapelc: I'm wary of splitting the function any further, since its performance has significant impact on tile load times (it gets executed for every post-earcut triangle, potentially multiple times, so maybe around hundreds of thousands of executinons per tile loaded), and I got a significant performance boost by inlining function calls in javascript in the past. Tomorrow I'll try to measure it and see if splitting it further actually results in a performance regression or not.

src/render/subdivision.ts
```diff
@@ -1,46 +1,1044 @@
...
+ * @param polygon - An array of point rings that specify the polygon. The first ring is the polygon exterior, all subsequent rings form holes inside the first ring.
+ * @param canonical - The canonical tile ID of the tile this polygon belongs to. Needed for generating special geometry for tiles that border the poles.
+ * @param granularity - The subdivision granularity. If we assume tile EXTENT=8192, then a granularity of 2 will result in geometry being "cut" on each axis
+ * divisible by 4096 (including outside the tile range, so -8192, -4096, or 12288...), granularity of 8 on axes divisible by 1024 and so on.
+ * Granularity of 1 or lower results in *no* subdivision.
+ * @param generateOutlineLines - When true, also generates index arrays for subdivided lines that form the outline of the supplied polygon. True by default.
+ * @returns An object that contains the generated vertex array, triangle index array and, if specified, line index arrays.
+ */
+export function subdivideFill(polygon: Array<Array<Point>>, canonical: CanonicalTileID, granularity: number, generateOutlineLines: boolean = true): SubdivisionResult {
```
HarelM, line 605: I guess this would grow to be a factory?

src/render/subdivision.test.ts
```diff
@@ -0,0 +1,1113 @@
...
+            }
+            outlineEdges.add(key);
+        }
+    }
+
+    if (uncoveredEdges.size !== outlineEdges.size) {
+        throw new Error(`Polygon exposed triangle edge count ${uncoveredEdges.size} and outline line count ${outlineEdges.size} does not match.`);
+    }
+
+    const isSubsetOf = (a: Set<string>, b: Set<string>): boolean => {
```
HarelM, line 1058: Avoid internal method sas much as possible.

src/render/subdivision.test.ts
```diff
@@ -0,0 +1,1113 @@
...
+        expect(result).toEqual([0, 1, 0, 2, 1, 3, 2, 3]);
+    });
+});
+
+/**
+ * Converts an array of points into an array of simple \{x, y\} objects.
+ * Jest prints much nicer comparisons on arrays of these simple objects than on
+ * arrays of points.
+ */
+function toSimplePoints(a: Array<Point>): Array<{x: number; y: number}> {
```
HarelM, line 978: I would consider removing as much of these methods as possible in order to avoid bugs that are in the logic of the tests, instead, check for specific stuff in the tests (a certain index, a specific vertex, etc.), and make the tests simpler.

HarelM:
I went through most of the code.
Overall, this looks good.
I have added some comments for readability and maintanability mostly.

HarelM:
Looking at the coverage here I think this is in great shape:
https://app.codecov.io/gh/maplibre/maplibre-gl-js/pull/3882/blob/src/render/subdivision.ts
I still feel like there are a few edge cases that are not covered, and maybe some checks that are redundant (dead code).

src/render/subdivision.ts
kubapelc: Is `test/unit/lib` the right dir where to place such test utils?

HarelM: Yup.

src/render/fill_arrays.test.ts
kubapelc: No, this function simulates how the GPU would fetch vertices for the given mesh and segments, respecting `vertexOffset` and `primitiveOffset`, and returns a string representation of that.

The goal of the tests is to check that a mesh (in this context, a mesh is a vertex buffer, index buffer and segment vector) with more than `SegmentVector.MAX_VERTEX_ARRAY_LENGTH` vertices results in the same rendered geometry as the result of passing that mesh through `fillLargeMeshArrays`, which creates a mesh that respects the vertex count limit.

I've renamed the function to better reflect its purpose and added comments clarifying what it does and why.

src/data/segment.test.ts
kubapelc: I think mocking is not enough in this case, since I need the `length` property of the buffer to reflect its contents, and I also test that the contents of those buffers actually match some expected values. But it would probably be possible to use the actual StructArrays for this, the `FillLayoutArray`, `TriangleIndexArray` and `LineIndexArray`. I'm not sure whether to keep using these test classes or to use the real thing.

src/render/fill_arrays.test.ts
kubapelc: Those functions create data for the tests rather than mock objects. I've moved them to a separate util file. I've also added tests that test the mesh comparison function and test that the grid mesh is actually a grid.

src/render/fill_arrays.ts
kubapelc: addVertex used to also be a simple `emplaceBack` with just the two vertex coordiantes, but when I implemented fill-extrusion, I found that it needs to store several additional values for every vertex, that is why I changed it to a function. So it could be a `emplaceBack` now, but not once fill-extrusion gets added in a future PR. Indices will always be a simple `emplaceBack`, since index will always be just an integer.

So I believe that there are good reasons for the current state where one is abstracted out and the other isn't, but on the other hand it really is a bit weird. I'm not sure what to do about that.

HarelM: Fair enough, I did not have the full context you do.

src/data/segment.test.ts
HarelM: If the real classes do the work and are simple enough, use those.

zhangyiatmicrosoft:
(late in the discussion; sorry if already talked before)
Wonder if the global logic should be made into an optional plugin? Not sure if majority of apps will be using it and maybe beneficial to have barebone version.

HarelM:
I'm not sure there's an easy way to strip this out as it requires running the code for all the shaders and other internal stuff, etc.
In general, this is the highest priority of the community to add to maplibre.

src/render/subdivision.ts
kubapelc: Not that I'm aware, I only found `calculateSignedArea` which does something different.

HarelM: I see, similar to this signedArea I guess.
https://github.com/maplibre/maplibre-style-spec/pull/519/files
Ok, fair enough, had to ask.
It's also worth checking if `turf` has something like this. But maybe not...

src/data/bucket/fill_bucket.ts
kubapelc: I've renamed it to a more descriptive `fillLargeMeshArrays`.

src/data/segment.test.ts
kubapelc: I'm now using the real classes.

src/data/segment.ts
```diff
@@ -25,32 +25,74 @@ export type Segment = {
...
-                primitiveOffset: indexArray.length,
-                vertexLength: 0,
-                primitiveLength: 0
-            } as any);
-            if (sortKey !== undefined) segment.sortKey = sortKey;
-            this.segments.push(segment);
+        const lastSegment: Segment = this.segments[this.segments.length - 1];
+
+        if (numVertices > SegmentVector.MAX_VERTEX_ARRAY_LENGTH) {
+            warnOnce(`Max vertices per segment is ${SegmentVector.MAX_VERTEX_ARRAY_LENGTH}: bucket requested ${numVertices}`);
```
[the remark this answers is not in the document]
kubapelc: I've added a mention of `fillLargeMeshArrays` since it can resolve this error. This warning is actually from the original segments.ts and not my addition, but there was no workaround for it before `fillLargeMeshArrays`.

src/geo/projection/projection.ts
kubapelc: I've added the comment.

src/render/fill_arrays.ts
kubapelc: I've deduplicated it by converting the repeated condition into a function.

src/render/subdivision.ts
kubapelc: I didn't find any usable alternative in turf.

src/render/subdivision.ts
kubapelc: It is very specific to how it is used inside `_fillPoles` , and only serves to deduplicate some code, I would prefer to keep it contained inside the function.

src/render/subdivision.ts
kubapelc: I've renamed it to `_vertexToIndex`.

src/render/subdivision.ts
kubapelc: It really is converting the input indices (from being indices into one vertex array to indices into a different vertex array) rather than just deduplicating. I've improved the explanation in the comment.

src/render/subdivision.test.ts
kubapelc: With subdivision a lot of the problems and bugs might only occur when subdividing some large or weird geometry (long thin triangles produces by `earcut`), and it is not practical to check such large output by hand, that is why I use functions that check specific properties of the resulting geometry. The "Subdivide a polygon" test is the largest that I checked by hand and it also runs all the functional checks as well.

I could add tests for the helper functions though.

`toSimplePoints` in particular is used to compare two arrays of `Point` instances, because just using `expect(...).toEqual(...)` directly will fail when comparing two different instances of `Point` that actually have the same coordinates, but I need the test to pass in that case.

HarelM: In this specific case I think you can use "toDeepEqual"

src/render/subdivision.ts
kubapelc: > Should this method get its own file?

It is heavily tied to subdivision, I would prefer to keep it in `subdivision.ts`. Or maybe create a new directory and break up more of subdivision into smaller files?

> Shouldn't this method return the finalIndices instead of receiving and filling it?

Then an extra copy of the returned array would be needed to fill `finalIndices`. I observed a performance degradation of around 15% when returning a new array instead of filling the existing array.

src/render/subdivision.ts
kubapelc: > Can't we use strings for key creation?

I've measured a significant slowdown (>50%) when using strings instead of integers as keys.

> This feels like it can easily break, are we OK with this assumption?

I've added a check that throws if a vertex is outside the int16 range. All tile geometry should be well within that though, any many other parts of MapLibre rely on that being true (16bit GPU vertex coordinates mostly).

src/render/subdivision.ts
kubapelc: The idea is to force the vertex coordinates to be integers. Removing the `| 0` causes a performance degradation in Chrome of about 10%, but curiously seems to have no impact in Firefox

src/render/subdivision.test.ts
kubapelc: I can't find any reference to `toDeepEqual` or anything similar in jest, `.toEqual` is already supposed to do deep equality (as opposed to `.toBe`).

src/render/subdivision.ts
kubapelc: I'm not sure what you mean by factory in this context. `Subdivision` is a class only because it is useful to have some context shared between the subdivision functions, mostly the vertex dictionary, it is not supposed to ever exist beyond a single subdivision operation.

src/render/subdivision.ts
kubapelc: It's a bit more involved, it requires modifying the fill bucket to run this function and then pass its results to the renderer instead of the actual tile outlines, and pass an empty array of triangles. This will cause only the wireframe to be rendered.

I'm not sure whether to keep this function.

src/render/subdivision.ts
kubapelc: I've taken another look at it, I don't think there is a good way to split it further, it's all very interconnected.

HarelM: When I studied computer science, long ago, they told me that a method should be 30 lines long max.
While it's a bit strict, it does force you to split a method into several part, even if interconnected, it still a lot easier to read and understand.

src/render/subdivision.ts
HarelM: This still seems weird to me, is math.round for 0 returns a float of 0?

src/render/subdivision.ts
HarelM: Methods within methods create a strong coupling and hard to understand and read, I have yet to find a good reason to define a method within another method.

src/render/subdivision.ts
HarelM: OOP factory - generate an instance based on input parameters, meaning this method will grow to generate more types of subdivision based on what is needed in terms of context.
In that case, I purpose to move it to its own file either now or when adding more instances types.

src/render/subdivision.test.ts
HarelM: Although I still think you should be able to write tests that are small enough to check the code you wrote without creating large objects (when writing the tests, not when debugging the code in production) I'll mark this as resolved

src/data/segment.ts
```diff
@@ -25,32 +25,74 @@ export type Segment = {
...
+        layoutVertexArray: StructArray,
+        indexArray: StructArray,
+        sortKey?: number
+    ): Segment {
+        const segment = ({
+            vertexOffset: layoutVertexArray.length,
+            primitiveOffset: indexArray.length,
+            vertexLength: 0,
+            primitiveLength: 0
+        } as any);
```
HarelM, line 70: There shouldn't be a `as any` here I believe.
There's an unneeded `(` and `)`.

src/render/fill_large_mesh_arrays.ts
```diff
@@ -0,0 +1,242 @@
...
+        // Thus, if we encounter a vertex that doesn't fit into MAX_VERTEX_ARRAY_LENGTH,
+        // we can just stop appending into the old segment and start a new segment and only append to the new segment,
+        // copying vertices that are already present in the old segment into the new segment if needed,
+        // because there will not be too many of such vertices.
+
+        // Normally, (out)lines share the same vertex buffer as triangles, but since we need to somehow split it into several drawcalls,
+        // it is easier to just consider (out)lines separately and duplicate their vertices.
+
+        fillSegmentsTriangles(segmentsTriangles, vertexArray, triangleIndexArray, flattened, triangleIndices, addVertex);
+        if (segmentsLines && lineIndexArray) {
```
HarelM, line 97: Should this if use `hasLines`?

src/render/fill_large_mesh_arrays.ts
```diff
@@ -0,0 +1,242 @@
...
+) {
+    // Array, or rather a map of [vertex index in the original data] -> index of the latest copy of this vertex in the final vertex buffer.
+    const actualVertexIndices: Array<number> = [];
+    for (let i = 0; i < flattened.length / 2; i++) {
+        actualVertexIndices.push(-1);
+    }
+
+    let totalVerticesCreated = 0;
+
+    const copyOrReuseVertex = (index, needsCopy, segment) => {
```
HarelM, line 193: I can't find the difference between this method and the other `copyOrReuseVertex`, is this a copy-paste mistake or a need to remove one method?

src/render/subdivision.ts
HarelM: I would say to remove it if you are not sure it's worth keeping.

src/render/subdivision.ts
kubapelc: I know what an OOP factory is. This function won't grow into one, at least not within any of the code I plan to add, that's why I was confused. It just wraps around the creating the single use class so that the final API for subdivision is just a simple function call. Only two subdivision functions needed for globe are this one (for fill and fill-extrusion layers) and `subdivideVertexLine` (for fill, fill-extrusion and line layers).

I also want to experiment with having all the internal subdivision functions return flattened vertex coordinates instead of indices to see if it performs better. This would also remove the need for a "Subdivison" class altogether. But I'd prefer to leave that for another PR.

I now think that I didn't name the function properly - calling it "subdividePolygon" would be more descriptive.

src/render/subdivision.ts
kubapelc: I guess `Math.round()` must work with numbers beyond 32 bit int range and thus returns a standard `number`, but `| 0` is specified to always return 32 bit ints.

src/render/subdivision.test.ts
```diff
@@ -0,0 +1,1114 @@
...
+
+    if (uncoveredEdges.size !== outlineEdges.size) {
+        throw new Error(`Polygon exposed triangle edge count ${uncoveredEdges.size} and outline line count ${outlineEdges.size} does not match.`);
+    }
+
+    expect(isSubsetOf(outlineEdges, uncoveredEdges)).toBe(true);
+    expect(isSubsetOf(uncoveredEdges, outlineEdges)).toBe(true);
+}
+
+function isSubsetOf(a: Set<string>, b: Set<string>): boolean {
```
HarelM, line 1063: Doesn't this method exists for sets?
Using the browser console I was able to use `isSubsetOf`.
Consider removing this methos.

HarelM: But this is only for tests, do the tests pass if you use the MDN one?

kubapelc: I've tried, they fail when I use the built-in method.

HarelM: Ok, thanks for the info.
Consider moving this to the test utils file.

src/render/subdivision.ts
kubapelc: I've removed it.

src/render/subdivision.ts
kubapelc: I've split it down, not quite to 30 lines (definitely not with the extensive comments), but I think it is more readable now. This split does seem have a negative performance impact, but it seems to be below 5%.

src/render/subdivision.ts
kubapelc: It's moved outside now.

src/data/segment.ts
kubapelc: I've fixed it. Btw this originates from the original state of segments.ts, some tests even relied on this. But there really was no good reason to keep the `as any` though.

src/render/fill_large_mesh_arrays.ts
kubapelc: Yes, good catch, fixed.

src/render/fill_large_mesh_arrays.ts
kubapelc: I've refactored both into a single function outside the methods.

src/render/subdivision.test.ts
```diff
@@ -0,0 +1,1113 @@
...
+        }
+        set.add(key);
+    }
+    return false;
+}
+
+/**
+ * Passes if all triangles have the correct winding order, otherwise throws.
+ */
+function checkWindingOrder(flattened: Array<number>, indices: Array<number>): void {
```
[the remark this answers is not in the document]
kubapelc: I don't think there is any easier way. Right now both functions are conceptually very simple: load triangle, compute cross product of its edges, do something based on the result. I don't think it is a good idea to unify them, since that would result in something much more complex. Also, changing any of the two functions without also changing the other would cause the tests to fail.

kubapelc:
The pushed code and comments should all be up-to-date now.

src/render/subdivision.ts
kubapelc: > I also want to experiment with having all the internal subdivision functions return flattened vertex coordinates instead of indices to see if it performs better.

The experiment ended up performing about 25% worse than current subdivision.

src/render/subdivision.test.ts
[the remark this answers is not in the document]
HarelM: fair enough.

src/render/subdivision.ts
HarelM: Aare you testing a production or dev build? I think rollup knows how to inline methods that are used once to improve performance.

HarelM: Looks great BTW!

src/render/subdivision.test.ts
```diff
@@ -0,0 +1,1106 @@
...
+    for (let i = 0; i < a.length; i++) {
+        result.push({
+            x: a[i].x,
+            y: a[i].y,
+        });
+    }
+    return result;
+}
+
+function subdivideFillFromRingList(rings: Array<Array<Point>>, canonical: CanonicalTileID, granularity: number) {
```
HarelM, line 982: This function doesn't look like it adds anything besides wrapping an existing method, is it needed? Am I missing anythign?

src/render/subdivision.ts
kubapelc: > Are you testing a production or dev build? I think rollup knows how to inline methods that are used once to improve performance.

I've created a benchmark for subdivision that I'm running using `npm run start-bench` and opening http://localhost:9966/test/bench/versions/index.html#Subdivide in a browser. I haven't investigated what the benchmarking system actually does, I assumed it tests the production builds.

src/render/subdivision.test.ts
kubapelc: Yes it is useless now, I changed the API of `subdividePolygon`/`subdivideFill` at some point and didn't notice that it made this function redundant. It used to convert the arguments to what `subdividePolygon` used to accept. I will remove it.

src/data/segment.ts
```diff
@@ -25,32 +25,75 @@ export type Segment = {
 export class SegmentVector {
     static MAX_VERTEX_ARRAY_LENGTH: number;
     segments: Array<Segment>;
+    private _invalidateLast: boolean = false;
```
HarelM, line 28: Can we find a better name for this? invalidate last is not intuitive for me on what it does.
maybe something like `_createEmptySegnmentOnNextPrepare`?

src/data/segment.ts
```diff
@@ -25,32 +25,75 @@ export type Segment = {
...
-            } as any);
-            if (sortKey !== undefined) segment.sortKey = sortKey;
-            this.segments.push(segment);
+        const lastSegment: Segment = this.segments[this.segments.length - 1];
+
+        if (numVertices > SegmentVector.MAX_VERTEX_ARRAY_LENGTH) {
+            warnOnce(`Max vertices per segment is ${SegmentVector.MAX_VERTEX_ARRAY_LENGTH}: bucket requested ${numVertices}. Consider using the \`fillLargeMeshArrays\` function if you require meshes with more than ${SegmentVector.MAX_VERTEX_ARRAY_LENGTH} vertices.`);
+        }
+
+        if (!lastSegment || lastSegment.vertexLength + numVertices > SegmentVector.MAX_VERTEX_ARRAY_LENGTH || lastSegment.sortKey !== sortKey || this._invalidateLast) {
```
HarelM, line 50: Optional: Consider moving the invalidate last to be the first check for readability

src/data/segment.ts
```diff
@@ -25,32 +25,75 @@ export type Segment = {
...
+     */
+    getOrCreateLatestSegment(
+        layoutVertexArray: StructArray,
+        indexArray: StructArray,
+        sortKey?: number
+    ): Segment {
+        return this.prepareSegment(0, layoutVertexArray, indexArray, sortKey);
+    }
+
+    invalidateLast() {
```
HarelM, line 93: Should this check if it is already true and throw to avoid creating two empty segments in a row?
I might not understand what this actually does though.

test/bench/benchmarks/subdivide.ts
```diff
@@ -0,0 +1,63 @@
...
+        }
+
+        generateHole(0.25, 0.5, 0.15, 16 * vertexCountMultiplier);
+        generateHole(0.75, 0.5, 0.15, 2 * vertexCountMultiplier);
+        generateHole(0.5, 0.1, 0.05, 4 * vertexCountMultiplier);
+
+        this.polygon = polygon;
+    }
+
+    async bench() {
```
HarelM, line 43: This doesn't look `async`...?

HarelM (approved):

src/data/segment.ts
kubapelc: I've renamed it to `_forceNewSegmentOnNextPrepare`.

src/data/segment.ts
kubapelc: Done.

test/bench/benchmarks/subdivide.ts
kubapelc: Fixed.

src/data/segment.ts
```diff
@@ -25,32 +25,81 @@ export type Segment = {
...
+        sortKey?: number
+    ): Segment {
+        return this.prepareSegment(0, layoutVertexArray, indexArray, sortKey);
+    }
+
+    /**
+     * Causes the next call to {@link prepareSegment} to always return a new segment,
+     * not reusing the current segment even if the new geometry would fit it.
+     */
+    forceNewSegmentOnTextPrepare() {
```
HarelM, line 99:
```suggestion
    forceNewSegmentOnNextPrepare() {
```

src/render/fill_large_mesh_arrays.ts
```diff
@@ -0,0 +1,257 @@
...
+        // it is easier to just consider (out)lines separately and duplicate their vertices.
+
+        fillSegmentsTriangles(segmentsTriangles, vertexArray, triangleIndexArray, flattened, triangleIndices, addVertex);
+        if (hasLines) {
+            fillSegmentsLines(segmentsLines, vertexArray, lineIndexArray, flattened, lineList, addVertex);
+        }
+        // Triangles and lines share vertex buffer, but we increment vertex counts of their segments by different amounts.
+        // This can cause incorrect indices to be used if we reuse those segments, so we force the segment vector
+        // to create new segments on the next `prepareSegment` call.
+        segmentsTriangles.forceNewSegmentOnTextPrepare();
```
HarelM, line 102:
```suggestion
        segmentsTriangles.forceNewSegmentOnNextPrepare();
```

src/render/fill_large_mesh_arrays.ts
```diff
@@ -0,0 +1,257 @@
...
+        segmentsLines?.forceNewSegmentOnTextPrepare();
```
HarelM, line 103:
```suggestion
        segmentsLines?.forceNewSegmentOnNextPrepare();
```

src/data/segment.ts
kubapelc: ...and I've now also fixed the typo in the name.

src/data/segment.ts
kubapelc: I've pushed the new unit tests, `fillLargeMeshArrays` works as it should.

Issue #3897, opened 2024-03-25
Sprite url gets validated before transformRequest 


Kai-W:
I was experimenting with relative URLs in styles and transforming all relative URLs to absoulte ones with the transformRequst Method.
Unfortunatly the loading of Sprites failes with an invalid URL Error (because its relative) before the transformRequest Method gets called.
https://github.com/maplibre/maplibre-gl-js/blob/f6035db6670de1af2d294e3cb5f6702f8616b53f/src/style/load_sprite.ts#L31C54-L31C70

**maplibre-gl-js version**:
4.1.1

**browser**:
Chrome Version 123.0.6312.59

### Steps to Trigger Behavior

 1. Load a Stylesheet with realative urls
 2. add transformRequest to transform relative to absoulte URLs
 3. Error: Unable to parse URL "/test" (url before transform)

### Link to Demonstration
https://jsbin.com/ranucez/edit?html,console,output

### Expected Behavior
transformRequest should be called bevor any checks if the URL is valid, eg a relative one

### Actual Behavior
Error: Unable to parse URL "/maps/streets/sprite" and transformRequest never gets called for Sprites

HarelM:
Feel free to submit a PR with a fix.

HarelM:
Can you maybe share the transform request method for the relative urls?
I think the following issue describes a similar issue:
 - #182

HarelM:
@Kai-W is this issue still relevant?

Kai-W:
probably not

PR #3898, opened 2024-03-25
fix: order of normalizeSpriteURL  and transformRequest in load sprites

Kai-W:
This PR fixes the order of transformRequest and normalizeSpriteURL on loading Sprites.

([#3897](https://github.com/maplibre/maplibre-gl-js/issues/3897))

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


src/util/request_manager.ts
```diff
@@ -7,8 +7,7 @@ export const enum ResourceType {
     Glyphs = 'Glyphs',
     Image = 'Image',
     Source = 'Source',
-    SpriteImage = 'SpriteImage',
-    SpriteJSON = 'SpriteJSON',
+    Sprite = 'Sprite',
```
HarelM, line 10: Doesn't this break the current API of request transform, which is a public API?

Kai-W: Havent thought about that, could be. But if we call first transformRequest we dont have the inforamtion if it is an SpriteImage or SpriteJSON. And we cant first convert the urls to an Image/JSON because we need a absolute URL for normalizeSpriteURL to work

HarelM: I'm not saying this shouldn't be fixed, I just want to see if there's a way to do it without a breaking change.

HarelM: Otherwise, this will have to go into version 5 breaking change version, which can take a few month to be published.

Kai-W: simply appending doesnt work if the url doesnt work if there are query parameters.
The current parsing of Urls is done by an [regex ](https://github.com/maplibre/maplibre-gl-js/blob/c91d7fb60adc05ac3a357b03cbab2ef63d4cda23/src/util/request_manager.ts#L56) in request manager and the format and extension is appended to the path part.
Personaly i dont like custom URL parsing with a regex. Is there a reason why the browser build in URL() wasn't used?

HarelM: This was probably written before url existed.
I think the right solution would be to replace this regexp with the url class.

Kai-W: i was wrong about the URL API: "If url is a relative URL, base is required" new URL(url, base)
new URL("/test") VM57:1 Uncaught TypeError: Failed to construct 'URL': Invalid URL
So we are back to the relative url Problem. And quick testing with the base param only uses the origin of the base url.

Kai-W: quick and dirty solution: split the Sprite URL at "?" insert the format/extension and concatinate it back together.

HarelM: Not sure I fully follow.

Kai-W: Using the URL api to parse the spriteURl does work for absoulte URLs
For relative URLs we have to provide an additional base parameter to resolve against. The progres gets then resolved (base+relative)
This would be an replacement of the current implementation of absolute URLs but doesn't solve the Problem that relative URLs are parsed and validated before the transformRequest call.

The quick and dirty way:
To use the current system with differnt SpriteImage and SpriteJSON calls we need to extend the provided SpriteUrl with the information of ${format}${extension} before we call Transform.
We could try to modify the URL without parsing just based on String Operations.
This extension is always appended at the path of the URL.
If there are query Parameters thats directly before the first ? or without params at the end of the String
E.g. @2x.png should be inserted:
http://www.foo.com/bar?fresh=true -> http://www.foo.com/bar@2x.png?fresh=true
http://www.foo.com/bar -> http://www.foo.com/bar@2x.png
/bar -> /bar@2x.png
This doesn't need validation before the transformRequest. 
This will work for absolute and relative urls.
It will break if its an absolute url and there is no path component 
http://www.foo.com-> http://www.foo.com@2x.png
or sombody has arbitrary strings that get converted to urls in the transformRequest method
foobar -> foobar@2x.png

Kai-W: i tried implementing it in this branch. It works but with the drawbacks mentioned above
https://github.com/maplibre/maplibre-gl-js/compare/main...Kai-W:maplibre-gl-js:test
I still prefer the original solution with no modification of the SpriteURL before the call of transformRequest. It feels like a cleaner solution, but it has a small Public api change

HarelM: We can do public API changes as part of version 5, which we probably will target late this year or early next year.
If the normalize URL is only used in this code path then it makes sense to move it next to it.
Can you better clarify how the current method and the previous method are different? What won't work in the code you wrote (changing the normalizeUrl method)?

Kai-W: Current Method:
Parse URL with an regex in protocol, authority, path and params. If path is empty set to "/". Append ${format}${extension} to path and then reconstruct the url from the parts.
Afterwards call transformRequest twice with the modifyed urls (.josn + .png)
Works only with correct absolute URLs, otherwise parsing with the regex failes.
Works as well for absoulte URLs with no Path element e.g. http://www.foo.com -> http://www.foo.com/@2x.png/ 

New normalizeUrl:
Splits the String at the first "?" Assuming the Sting is an URL and a ? would seperate path and params of an URL
Then Appennd ${format}${extension} to the first part (path) and concatinate it back together. Then calling the transformRequest with both modifyed URLs
This works with absolute and relative URLs. But it wont work with arbitrary strings or absolute URLs with no Path. E.g
http://www.foo.com gets http://www.foo.com@2x.png/ is missing a "/" between .com and @2x
But i think thats an edgcase that would not happen in production. It would mean that there is no filename used for the sprites
Additionally with my current implemetation there is no validation of the URL returned by transformRequest.

New Api breaking Change:
First call transformRequest once withoput appending ${format}${extension} (.png/.json) to the Path.
Assume the result of transformRequest must be a valid absolute URL and then use normalizeSpriteURL to generate the .png/ .json urls based on the result of transfomrRequest.
Works with any sprite String.
Api change: transformRequest gets called once without  .png/.json in the URL and one generic Sprite ResourceType
Additionaly we could change the normalizeSpriteURL to use the Browser URL API to remove the regex parsing.

An Option could be to merge the new normalize URL now to fix the bug and then merge the api breaking Change for Version 5.

HarelM: I think I'm OK with the last suggestion about merging this fix now and adding a breaking change feature to v5.

HarelM:
@Kai-W can you resolve conflict and prepare this PR to be merged? I want to move forward with the breaking changes of version 5.

Kai-W:
I rebased the branch, but i dont know why the windows rendert test 0 is failing. @HarelM can you look at it?

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
+- Fix order of normalizeSpriteURL and transformRequest in loadSprite ([#3897](https://github.com/maplibre/maplibre-gl-js/issues/3897))
```
HarelM, line 8:
```suggestion
- ⚠️ Fix order of normalizeSpriteURL and transformRequest in loadSprite ([#3897](https://github.com/maplibre/maplibre-gl-js/issues/3897))
```

HarelM (approved):
I hope this will work well with all kind of tokens etc...

PR #3901, opened 2024-03-26
Fix close events being fired for popups that aren't open

tzetter:
This moves the line that fires close events to be inside the map section. This means that close events won't be fired for popups that weren't open when the popup was removed. The popup close events are currently fired for example when removing a marker that is linked to a popup from the map. 

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### 🐞 Bug fixes
 - Fix type definition on `localIdeographFontFamily` ([#3896](https://github.com/maplibre/maplibre-gl-js/pull/3896))
 - Fix unwanted panning changes at the end of a panning motion ([#3872](https://github.com/maplibre/maplibre-gl-js/issues/3872))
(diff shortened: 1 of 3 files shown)


HarelM:
Can you please add a test and a chengelig entry?

tzetter:
There is already a test that verifies that the close event is fired when a popup is closed. I could add a test that verifies that it doesn't fire inappropriately if you like, but it doesn't seem super useful to me. 

I'll add a changelog entry.

HarelM:
This is what this fix is all about, adding a test to cover it makes sense to me.

tzetter:
> This is what this fix is all about, adding a test to cover it makes sense to me.

You're probably right 🙂 I added a test that verifies that it doesn't fire

HarelM (approved):

HarelM:
THANKS!

PR #3909, opened 2024-03-26
Add layers slots

sbachinin:
This PR facilitates toggling of layers, especially groups of layers.
Changes to API:
- user can specify a layer of `{type: 'slot')` (Therefore it requires a corresponding change in style-spec)
- user can fill a slot with an array of layers (and remove all layers from a slot) via `map.populateSlot` method

----
 - [x] Link to related issues.
 - [ ] Post benchmark scores.

## Rationale.
This is motivated by a recent [discussion](https://github.com/maplibre/maplibre-gl-js/discussions/3787) and my own hard experience with toggling.
User currently has no straightforward solutions for toggling. It's overall possible but in a rather awkward way.
It gets especially difficult when you need to toggle _groups_ of layers, and especially groups of layers that come _after each other_ (because `beforeId` in such case becomes unreliable).
Current API compels a user to add layers to the style object from the start and later toggle their visibility. This is 1) really inconvenient and 2) it seems like wasting computational time for stuff that might be never shown.

Benefits of this change are that:
- it gives a more expressive way to toggle, without the need to know beforeIds
- makes it easier to add/remove layers in batches.

## Comparison to mapbox slots
The concept and implementation are quite different.
Both my slots and mapbox slots are "named places where you can insert extra layers".
But in mapbox:
- slots are not made for toggling. They allow to put layers into certain named positions but they do not allow to remove/replace all layers from a slot.
- positions of slots are not user-defined but hardcoded into mapbox basemap.
- layers are added to a slot one by one. It's done by providing a "slot" property on a layer object.
- there are no special methods to populate a slot

My implementation allows a user to define his own points and to populate them in batches.
I assume that a user has a logically coherent group of layers, such as "all about roads" or "all about contours" and wants to be able to switch on/off the entire group.
This case is demonstrated in the example page that is included in this PR.

## Style-spec
Style spec has to permit layers of type `slot` that have no other properties except `id`.

## Some possible problems with this code
Implementation is quite simple but it creates a lot of cases where things can go wrong.
I tried to summarize the possible quirks in a test file that's included.
Other possible improvements:
- Currently I insert a batch of layers just by calling style.addLayer in a loop - not sure if it's ok from performance pov, perhaps there is better way to do it.
- There is no way to define a _filled_ slot as part of style object; slot can be filled only later via populateSlot.
- In order to populate a layer from the start, you have to wait for style.load

## Video of example page

https://github.com/maplibre/maplibre-gl-js/assets/11789697/b3a692e5-bdd7-4769-89ea-a927658eba96

### DIFF
--- a/src/style/style.ts
+++ b/src/style/style.ts
@@ -186,6 +186,7 @@ export class Style extends Evented {
     _serializedLayers: {[_: string]: LayerSpecification};
     _order: Array<string>;
     _orderWithSlots: Array<string>;
+    _slots: {[_: string]: { layerIds: string[]}};
     sourceCaches: {[_: string]: SourceCache};
     zoomHistory: ZoomHistory;
     _loaded: boolean;
@@ -228,6 +229,8 @@ export class Style extends Evented {
         this._layers = {};
 
         this._order = [];
+        this._orderWithSlots = [];
+        this._slots = {};
         this.sourceCaches = {};
         this.zoomHistory = new ZoomHistory();
         this._loaded = false;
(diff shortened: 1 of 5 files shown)


HarelM:
I don't object to the idea as it makes sense to me, especially when talking about switching between maps without loading the entire style (overlay type of layers), but I think this discussion should start in the style spec repo to define how this would look in the style.json and after that define the map API that should facilitate for what's in here.
A good direction, thanks for pushing this forward!

PR #3944, opened 2024-04-04
Remove finalize elevation when updating elevation on the fly

### LINKED ISSUE #3878: When simulating a GPS bearing the map zooms in/out with terrain
**maplibre-gl-js version**: 4.1.1
**browser**: chrome

### Steps to Trigger Behavior
See below.
1. Set terrain
2. Move the bearing slowly back and forth
 
### Link to Demonstration
https://jsbin.com/citaminibe/2/edit?html,js,output

### Expected Behavior
Like in 2D - no movement

### Actual Behavior
Depend on the terrain, this will cause a slow zoom in or zoom out.

olsen232:
Don't _finalizeElevation if freezeElevation has not been set.

This means that elevation logic has two distinct operating modes.

Mode one - freezeElevation is false - default:
- updateElevation is called continuously during the movement
- finalizeElevation is *not* called at the end.

Mode two - freezeElevation is true:
- updateElevation is *not* called continuoulsy during the movement
- finalizeElevation is called at the end.

Previously, both methods were called in mode one, which lead to issue #3878 if freezeElevation was not set.

Fixes #3878

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
This fix makes sense to me regardless of what I wrote in the issue.

HarelM:
Can you add a test and a changelog item? 
Since elevation is constantly updated there's no need to finalize it, right? 
I think this should get it.

prozessor13 (approved):
I also think this change makes absolutely sense!

olsen232:
Added a test, updated the changelog.

src/ui/camera.test.ts
```diff
@@ -1728,7 +1728,40 @@ describe('#flyTo', () => {
...
+
+        const terrainCallbacks = {prepare: 0, update: 0, finalize: 0} as any;
+        camera.terrain = {} as Terrain;
+        camera._prepareElevation = () => { terrainCallbacks.prepare++; };
+        camera._updateElevation = () => { terrainCallbacks.update++; };
+        camera._finalizeElevation = () => { terrainCallbacks.finalize++; };
+
+        camera.setCenter([-10, 0]);
+
+        camera.on('moveend', () => {
```
HarelM, line 1743: Use `once` and avoid the usage of `done` for easier failure report.

olsen232: Can you elaborate? I can't see any guidance in the documentation about avoiding `done` and the word done occurs 130 times in this file, so it seems to be more or less standard practise. There do seem to be a variety of ways to use it, perhaps I'm using it wrong. But anyway, let me know what you want me to do here

HarelM: Use async await instead of using done.
It's not written anywhere, I'm trying to reduce as much as possible the usage of done in tests.

olsen232: Done

src/ui/camera.test.ts
```diff
@@ -1728,7 +1728,7 @@ describe('#flyTo', () => {
         camera.flyTo({center: [100, 0], bearing: 90, animate: true});
     });
 
-    test('check freezeElevation events', done => {
+    test('check elevation events freezeElevation=false', async() => {
```
HarelM, line 1731:
```suggestion
    test('check elevation events freezeElevation=false', async () => {
```

src/ui/camera.test.ts
```diff
@@ -1740,25 +1740,44 @@ describe('#flyTo', () => {
 
         camera.setCenter([-10, 0]);
 
-        camera.on('moveend', () => {
+        await camera.once('moveend', () => {
```
HarelM, line 1743: I'm not sure this will work - it's either await or give a callback method to the `once` call.
you'd need to so something like:
```
const promise = camera.once('moveend');
// continue logic
await promise;
expect(...)
```

src/ui/camera.test.ts
```diff
@@ -1740,25 +1740,44 @@ describe('#flyTo', () => {
...
-        }, 0);
+        const terrainCallbacks = {prepare: 0, update: 0, finalize: 0} as any;
+        camera.terrain = {} as Terrain;
+        camera._prepareElevation = () => { terrainCallbacks.prepare++; };
+        camera._updateElevation = () => { terrainCallbacks.update++; };
+        camera._finalizeElevation = () => { terrainCallbacks.finalize++; };
+
+        camera.setCenter([-10, 0]);
+
+        await camera.once('moveend', () => {
```
HarelM, line 1769: Same as above.

src/ui/camera.test.ts
olsen232: Done

src/ui/camera.test.ts
olsen232: Done

Issue #3887, opened 2024-03-23
Globe - Improve tile management


Pheonor:
## Rationale

Tile loading mechanism should be updated to work with a Globe projection.

## Impact

Load every tile needed to cover the whole Globe with the right resolution.

HarelM:
Bounty direction:
https://github.com/maplibre/maplibre/issues/190

vlarrieu:
@HarelM Could you assign it to me please ?

kubapelc:
@Pheonor Hi, I'm currently doing a lot of changes in my branch that affect how this would be implemented. Since your other PR is nearly finished, I thought I should tell you about this so we can avoid hard to solve merge conflicts :)

I've moved most logic from the Projection interface to the Transform class. Transform is now an abstract class, with implementations for mercator and globe projections. I decided to do this, because Transform is the class already used by MapLibre to do everything related to projecting and unprojecting pixels and coordinates. Transform also already stores the map's center, pitch, bearing, etc., which is all needed for projections.

Transform is also cloneable in MapLibre, and this is used to store the map's state from a given frame, and use it for a longer period of time to do projections. This is used in symbols, and the old Projection interface didn't account for this at all, so the globe symbol implementation was slightly wrong. Moving most projection stuff to Transform fixes this.

I think the abstract transform mightmake implementing better tile management for different projections easier.

You can preview the changes here: https://github.com/windycom/maplibre-gl-js/blob/kubapelc/globe-vector/src/geo/projection/globe_transform.ts

I hope to have a PR ready within about two weeks (hopefully).

Pheonor:
> @Pheonor Hi, I'm currently doing a lot of changes in my branch that affect how this would be implemented. Since your other PR is nearly finished, I thought I should tell you about this so we can avoid hard to solve merge conflicts :)
> 
> I've moved most logic from the Projection interface to the Transform class. Transform is now an abstract class, with implementations for mercator and globe projections. I decided to do this, because Transform is the class already used by MapLibre to do everything related to projecting and unprojecting pixels and coordinates. Transform also already stores the map's center, pitch, bearing, etc., which is all needed for projections.
> 
> Transform is also cloneable in MapLibre, and this is used to store the map's state from a given frame, and use it for a longer period of time to do projections. This is used in symbols, and the old Projection interface didn't account for this at all, so the globe symbol implementation was slightly wrong. Moving most projection stuff to Transform fixes this.
> 
[...]

Issue #3834, opened 2024-03-13
Breaking changes MapLibre GL JS v5


sbachinin:
This issue is to track breaking changes for MapLibre GL JS v5:

- [x]  queryTerrainElevation should return meters above sea level. (Currently it returns a special offset used for positioning of custom 3d objects). - #3854
- [x] getElevationOffset method should be added. It should return the offset that is currently retrieved from queryTerrainElevation - there is elevation definition now as part of the spec.
- [x] https://github.com/maplibre/maplibre-style-spec/pull/519
- [x] #3898
- [x] https://github.com/maplibre/maplibre-gl-js/pull/3136#issuecomment-2101231496
- [x] Fix sky for globe - #4853
- [x] Fix performance issues for globe related to `projectAndGetPerspectiveRatio` - #4778
- [x] Add projection outside the style for initialization much like zoom/breaking etc. - #4909
- [x] Add more parameters to custom layer to better support globe
- [x] ~Geojson close to the poles with globe - https://github.com/maplibre/maplibre-gl-js/discussions/1598#discussioncomment-10538001, https://github.com/maplibre/maplibre-gl-js/pull/3963#issuecomment-2344778823~ - this is a limitation of mercator coordinates system.
- [x] Return value for `on` method to allow unsubsribing - #5080
- [x] Change how bearing and roll UX on desktop when dragging - https://github.com/maplibre/maplibre-gl-js/pull/4780#discussion_r1803869111, #5074
- [x] Globe showstpeer issues (terrain + globe is not a showstopper as one can disable terrain for lower zoom levels when globe is on): https://github.com/maplibre/maplibre-gl-js/labels/showstopper
- [x] Incorporate globe, vertical-perspective and expression for projection - #5139, #5114 , #5056
- [x] Move webgl initialization configuration into an object to future proof it - #5135, #5198

sbachinin:
This is likely to produce breaking changes: https://github.com/maplibre/maplibre-gl-js/issues/3736:
- `queryTerrainElevation` should return meters above sea level. (Currently it returns a special offset used for positioning of custom 3d objects).
- `getElevationOffset` method should be added. It should return the offset that is currently retrieved from queryTerrainElevation

HarelM:
The following issue will also create breaking change to the output of `geometry-type`:
- https://github.com/maplibre/maplibre-gl-js/issues/3516

This issue has a part in the style-spec repo as well.
[...]

Issue #1596, opened 2022-08-31
Implement ray/triange intersection in combination of coordsframebuffer


prozessor13:
currently screen-pixel to geo-coordinate transormation is done via a hidden framebuffer, which contains RGB-encoded geo-coordinates. But aspecially in heavy terrain this has rounding-errors, so i propose an improvement:

* instead of geo-coordinates store mesh-triange indexes in this framebuffer
* when knowing the mesh triange it should be easy to cast a ray to this triange and find the exact geo-coordinate

https://www.npmjs.com/package/ray-triangle-intersection

HarelM:
Sounds interesting!
This package is fairly old and might not be up to date with security and other issues.
Since this is basically a single method of around 20 lines I would consider simply writing the method in here at maplibre:
https://github.com/substack/ray-triangle-intersection/blob/master/index.js
I hope we have the relevant math function that are referenced there in our code already.

HarelM:
Can this problem be the source of the fact that the marker and the point on the terrain looks slightly moved aside:
![image](https://user-images.githubusercontent.com/3269297/191355715-9b17e75b-dd6a-494e-8fa9-866fbb514933.png)
This is how it look in 2D:
![image](https://user-images.githubusercontent.com/3269297/191355889-00b69bfd-397f-4d87-b389-f0e0aa83868d.png)

prozessor13:
what is the difference of the 2 markers? symbol-layer & maplibregl.Marker?

HarelM:
Yes, one is an html object and the other is part of the canvas.

HarelM:
Assigned L bounty.
Link to parent Bounty: https://github.com/maplibre/maplibre/issues/189

rotu:
Can you check whether some of the supposed rounding errors are actually fixed by #2772? There was a minor mistake in the elevation logic.

PR #3963, opened 2024-04-10
Globe final PR

### LINKED ISSUE #307: Globe view - Earth is not flat
## Motivation
Hello, I think it would be interesting when zooming out instead of duplicating the merchant projection it would be nice to show the view of the entire globe, similar to google earth and the new Apple Maps

## Design Alternatives

[](https://postimg.cc/3dsqFmhC)
[](https://postimg.cc/K3gN1p1X)

HarelM:
## Launch Checklist

- #307

Resolves #307

Keeping this as draft right now, this PR will allow merging of the entire globe branch into main.

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
This is still in draft, but I wanted coverage support, which doesn't work well for some reason...

HarelM:
@kubapelc I've rigged codecov to work on this PR, feel free to see which classes are covered well and which need more work.
I'll keep this PR open so that we can get up-to-date coverage information on every PR merge.

HarelM:
@kubapelc, I've merged main into this branch.
I've left some "HM TODO" in places where I replaced some logic that was needed.
Please check this thoroughly.

HarelM:
Bahhh.... I messed this one up.
Let me know if you need me to do anything to resolve this, or if you are taking care of it...

HarelM:
@kubapelc what still missing for this branch besides the atmosphere and the style spec definition for globe?

HarelM:
Note that there were examples that were added, which will appear in the docs site.
For those to appear as expected there's a need to generate images for them using the generate image script.
The following are the warning from the build generation code:
```
WARNING -  Doc file 'examples/index.md' contains a link '../assets/examples/globe-fill-extrusion.png', but the target 'assets/examples/globe-fill-extrusion.png' is not found among documentation files.
WARNING -  Doc file 'examples/index.md' contains a link '../assets/examples/globe-vectortiles.png', but the target 'assets/examples/globe-vectortiles.png' is not found among documentation files.
WARNING -  Doc file 'examples/index.md' contains a link '../assets/examples/globe.png', but the target 'assets/examples/globe.png' is not found among documentation files.
```

See here:
https://github.com/maplibre/maplibre-gl-js/blob/main/docs/README.md#writing-examples

kubapelc:
>what still missing for this branch besides the atmosphere and the style spec definition for globe?

Everything rendering related on my part is basically done and merged, what remains is adapting the rest of MapLibre to work with globe projection. Most notably controls need to be adapted for globe, then animations such as `flyTo`, placing custom HTML at a location, custom layer API for globe and for placing 3D objects onto the globe (as mentioned in the globe discussion), and in general everything that projects screen pixels to coordinates or vice versa. I'm working on it, it's a long list, but I think (hope) it should be much less complex than the rendering part...

Also https://github.com/maplibre/maplibre-gl-js/issues/3887 should be implemented for the globe to be usable.

kubapelc:
I'll fix the globe examples.

HarelM:
Thanks for the info!
If you happen to resolve the following enhancement it will be great:
- #1596

HarelM:
Please also take a look at the coverage report.
I think some unit tests will be helpful here to cover more of the code scenarios as render test are good but are not suitable for edge cases.
I would recommend aiming to 80-90% for each file so that the overall coverage would stay around 80-90%.

AZProductions:
It seems like that the coordinates for placed markers seem to not work correctly. When you rotate the globe they move out of screen instead of rotating with the globe.

kubapelc:
Hello, in this branch only the rendering of layers is adapted for globe, other things still behave as if they were mercator. I'm working on fixing that right now, and proper marker behaviour is definitely on my TODO-list, as well as better controls for globe :)

HarelM:
@kubapelc there's a small difference in a render test after merging a fix from main branch related to elevation of polygon.
See this PR:
 - #3841

It's a very small fix to move the centroid calculation into the for loop instead of outside.
The diff is very small, but noticeable in the render test.
My gut feeling is that we should update the expected render test image, but let me know what you think.

kubapelc:
Hello, this line:

```ts
const oldVertexCount = this.layoutVertexArray.length;
```

at `fill_extrusion_bucket.ts` line 171 should be moved into the for loop, so that the resulting code looks roughly like this:

```ts
addFeature(feature: BucketFeature, geometry: Array<Array<Point>>, index: number, canonical: CanonicalTileID, imagePositions: {[_: string]: ImagePosition}, subdivisionGranularity: SubdivisionGranularitySetting) {
        for (const polygon of classifyRings(geometry, EARCUT_MAX_RINGS)) {
            // Compute polygon centroid to calculate elevation in GPU
            const oldVertexCount = this.layoutVertexArray.length; // <---- moved here
            const centroid: CentroidAccumulator = {x: 0, y: 0, sampleCount: 0};
```

The for-loop that later iterates over `addedVertices` needs to fill the centroid center for every vertex added in `processPolygon`, and it uses `oldVertexCount` to compute how many new vertices were added. If `oldVertexCount` was initialized outside the `classifyRings` for-loop, it would instead track how many vertices were added for each ring/polygon so far in total.

As for the render test, this `oldVertexCount` change does not affect it. I think the test is different because the globe code takes into account that polygons sometimes do (and sometimes don't) have their first/last vertex duplicated. If it is duplicated, globe code ignores it in centroid computation. I think the main branch does not do this, and incorrectly counts the first/last vertex twice, thus it computes a slightly different centroid.

So updating the render test is correct.

HarelM:
Thanks for the tip, I have moved the above line, I'll see if the render test still fails and update the image if needed.

kubapelc:
> Hi, how can I define the projection in the options? And is there a way to change the projection after initialization? (`setProjection` and the `projection` attribute don't do anything)

Hi, I think something like this should work:

```js
map.on("load", () => {
    map.style.setProjection({type: "globe"});
});
```

But this will only set the projection once the map is loaded, but globe needs tiles to be reloaded if that happens and afaik this globe branch doesn't yet have a mechanism to trigger that, so you will get artifacts. This is already fixed in my dev branch.

HarelM:
Seems like a broke some of the terrain tests with the latest merge from main... :-/ I'll look at it later on...

test/build/min.test.ts
```diff
@@ -36,9 +36,9 @@ describe('test min build', () => {
         const decreaseQuota = 4096;
 
         // feel free to update this value after you've checked that it has changed on purpose :-)
-        const expectedBytes = 800300;
+        const expectedBytes = 866666;
```
```suggestion
        const expectedBytes = 861111;
```

HarelM:
@kubapelc don't forget to resolve conflicts here first as I see there are conflict with transform class and recent changes/fixes.

HarelM:
@kubapelc I've updated this branch with the fixes from main.
The main conflict was around the collision index `projectAndGetPerspectiveRatio` which is now unoptimized again since there was a lot of code changes there related to globe and mercator projection, not sure I know how to solve this, as copying the code now is a lot more complicated...

willemarcel:
As I commented in https://github.com/maplibre/maplibre-gl-js/discussions/1598#discussioncomment-10538001, the globe view doesn't render the geojson lines correctly when close to the poles. This is a simplified version of the data that can be used to reproduce this issue: https://gist.github.com/willemarcel/4995190eaf24e408b188a4d6fefca560

HarelM:
@kubapelc what's still missing in order to merge this?
I thought about merging this and creating a version 5 pre-release, do you think it's possible?

HarelM:
I see, that's very good news!
I'll take this branch for a spin and see that it works as expected. 
If everything is in order, I'll merge it.
Thanks again for all the hard work on this! Truly, truly appreciated!

HarelM:
@Pessimistress, in case you want to check things out before this is merged (in continue to the conversation we had yesterday).
I'm planning on merging this and create the first pre-release of version 5 of maplibre.

neodescis:
> @Pessimistress, in case you want to check things out before this is merged (in continue to the conversation we had yesterday). I'm planning on merging this and create the first pre-release of version 5 of maplibre.

I'd be happy to test out the prerelease version! However, I'm thinking we should do a final 4.x release with the latest fixes, just before merging this.

HarelM:
Sure, currently there's only a little fix left that was not included in 4.7.
I'll go over other PRs, maybe we could include some of those before this merge...

laem:
Happy to test the prerelease version too on https://cartes.app as soon as it lands :)

birkskyum:
I tried out the cartes app, and it's really cool!

After trying it out, I think we might have been making a bit of a design mistake here, by adding the `projection` setting inside of the style...
- https://github.com/maplibre/maplibre-style-spec/pull/687

This would mean that in order to use the globe, we'd need to make duplicates of all styles, such as https://demotiles.maplibre.org/style.json, or download them, manually modify the object, and set that json as the style.

What we could do instead, is to have the projection setting as a sibling to the style, and other viewport settings (center, pitch etc.), instead, and let people toggle the globe with the styles they already use. What do you think of this idea?

HarelM:
We might consider adding this to the map initialization object as well, but it should definitely be part of the style.
It doesn't have to be either-or.

birkskyum:
Another observation is that the projection API isn't flexible enough to accommodate this:
- #168

Right now we have a `projection { type: 'globe' | 'mercator' }`, but what is said between the lines is:

- **mercator**: 2d, pseudo-mercator
- **globe**: 3d, adaptive glove, with 2d, pseudo-mercator on high zoom.

So, we make an assumption here that the desired 2d projection for the globe always is mercator, and this assumption breaks down the second we introduce another 2d projection.

What we need (but can ignore for now since we only have 'mercator') is a setting for the 2d projection which can take values like:
- `[ESPG:3857](https://epsg.io/3857)` (pseudo-mercator),
- `[ESPG:7416](https://epsg.io/7416)` (Denmark),
- `[EPSG:1078](https://epsg.io/1078-method)` (Equal Earth),

And a toggle weather or not to enable the adaptive globe as you zoom out.

HarelM:
The more we wait the more hard conflict we will need to solve.
I'll probably merge this to main in the next few days...

kubapelc:
I think `globe-mercator` is a good name. As for setting globe outside the style object, I think it would be nice, since it works mostly seamlessly with mercator maps. I'm not sure how exactly the API shoul look...

High zoom globe uses mercator in particular because globe and mercator projections converge at high zooms. In this context "globe projection" is a classical perspective projection camera looking at a perfect sphere, and moving closer to the surface with increasing zoom. I'm not sure projections other than mercator have this property or that something would be gained by using them instead of mercator. But I can definitely see a use case for accepting tiles that are not web mercator, but that is outside the scope of globe implementation.

birkskyum:
> High zoom globe uses mercator in particular because globe and mercator projections converge at high zooms. In this context "globe projection" is a classical perspective projection camera looking at a perfect sphere, and moving closer to the surface with increasing zoom. I'm not sure projections other than mercator have this property or that something would be gained by using them instead of mercator. But I can definitely see a use case for accepting tiles that are not web mercator, but that is outside the scope of globe implementation.

A lot of governmental GIS services opt for non-mercator projections that have lower distortion/higher accuracy in the local region they cover. Take an example like this https://miljoegis.mim.dk/spatialmap?profile=lbst which is built on OpenLayers, it's the Environmental Ministry of Denmark - if the cookie bar is removed, in the lower right it says 'ETRS89 EAST / NORTH', which is the this [projection](https://epsg.io/25832).

Samarth1696:
> @Samarth1696 @cutephoton I've merged main branch into here, Please let me know if the fixes you have made are still OK in this branch, my main concern in the heatmap fix to be honest, I hope the tests added in the PR are covering this feature well to indicate if my merge was correct or not.

The things you changed looks correct to me!

But I have some concerns regarding terrain with the globe. I don't know maybe my configuration might be incorrect. Feel free to correct me if I am wrong. I am getting this on poles when I have set Terrain off:
{183CA944-EC13-4146-88B0-85A6BE6BE9E7}
When I have set the Terrain on:
{8979C2FE-58FD-4936-BBA5-4E930284A7D0}
Maybe there is some issue in the projection with raster layers?

I used this configuration to test:
```
<!DOCTYPE html>
<html lang="en">
<head>
    <title>Display a globe with a vector map</title>
    <meta property="og:description" content="Display a globe with a vector map." />
    <meta charset='utf-8'>
    <style>
    </style>
</head>
<body>
<div id="map"></div>
<script>
        container: 'map',
        zoom: 14.66,
        center: [-112.0392, 33.54002],
        pitch: 66,
        hash: true,
        style: {
            version: 8,
            sources: {
                osm: {
                    type: 'raster',
                    tiles: ['https://a.tile.openstreetmap.org/{z}/{x}/{y}.png'],
                    tileSize: 256,
                    attribution: '&copy; OpenStreetMap Contributors',
                    maxzoom: 19
                },
                // Use a different source for terrain and hillshade layers, to improve render quality
                terrainSource: {
                    type: 'raster-dem',
                    url: 'https://api.maptiler.com/tiles/terrain-rgb/tiles.json?key=API_KEY',
                    tileSize: 256
                },
                hillshadeSource: {
                    type: 'raster-dem',
                    url: 'https://api.maptiler.com/tiles/terrain-rgb/tiles.json?key=API_KEY',
                    tileSize: 256
                }
            },
            layers: [
                {
                    id: 'osm',
                    type: 'raster',
                    source: 'osm'
                },
                {
                    id: 'hills',
                    type: 'hillshade',
                    source: 'hillshadeSource',
                    layout: {visibility: 'visible'},
                    paint: {'hillshade-shadow-color': '#473B24'}
                }
            ],
            terrain: {
                source: 'terrainSource',
                exaggeration: 1
            },
            sky: {}
        },
        maxZoom: 18,
        maxPitch: 85
    });

    map.on('style.load', () => {
        map.setProjection({
            type: 'globe', // Set projection to globe
        });
    });

    map.addControl(
        new maplibregl.NavigationControl({
            visualizePitch: true,
            showZoom: true,
            showCompass: true
        })
    );

map.addControl(
        new maplibregl.TerrainControl({
            source: 'terrainSource',
            options: {
        exaggeration: 2,
        elevationOffset: 0
      }
     })
    );
</script>
</body>
</html>
```

cutephoton (commented):
@HarelM it looks like it should work. I verified the shaders and confirm they look correct. I did the eyeball check using multiple geometries example. It was ok. The render tests are taking an eternity. I am seeing a couple failing tests sporadically. I'll post an update when that eventually finishes.

cutephoton (approved):
Tested with `--open-browser`. No issues with circles. Failed tests do not seem related to the changes here, fail on main and are only ever so mildly outside of expectations on the few tests that failed.

```
 1/7: failed tests/text-local-ideographs/cjk 0.0021619791666666666 
 2/7: failed tests/raster-masking/overlapping-zoom 0.003936767578125 
 3/7: failed tests/projection/globe/raster-warped 0.054424285888671875 
 4/7: failed tests/projection/globe/raster-pole 0.000293731689453125 
 5/7: failed tests/debug/tile-raster 0.0036478042602539062 
 6/7: failed tests/debug/tile 0.00693511962890625 
 7/7: failed tests/debug/raster 0.0161285400390625 
```

Pessimistress:
Should `CustomRenderMethodInput` include `globeness` for third-party integration?

https://github.com/maplibre/maplibre-gl-js/blob/d0981a07962dfcf8dee077621bd12ca0fec0623f/src/render/draw_custom.ts#L17C28-L17C51

Thinking of future functionalities, this transition factor may be needed by other projections as well. Maybe "mercatorness" would be a more generic name?

HarelM:
That's a good question, I don't have a good answer, but I do want to merge this.
So I'll move all the relevant stuff that are still not solved to the following issue and release a pre-release of version 5:
- #3834

birkskyum (approved):

HarelM:
Here we go...

kubapelc:
Hi @Pessimistress the `globeness` parameter is exposed in `CustomRenderMethodInput`, it is in `defaultProjectionData.projectionTransition`. I'll think about how to document this better.

I think that the `globeness` name is only used in globe-related classes, all generic code (just shaders, really) uses `projectionTransition`.

kubapelc:
I'm working on one more globe-related PR. It has some optimizations, mostly around symbol placement.

HarelM:
Thanks for the update, feel free to take a look at the list I wrote here
- #3834 
At the beginning of the issue I wrote a check list of things that should be added/solved as part of version 5 release. feel free to continue the discussion there.

laem:
It's online on https://cartes.app :) Thanks for the great work.

russss:
I also just put it live on [OpenInfraMap](https://openinframap.org). It all looks good except for #4781. Thanks again for all the hard work!

eikes:
I have been playing with this and it is such a delight! Thank you very much! It would be nice to have an example where this is used.

src/render/painter.ts
```diff
@@ -590,20 +655,41 @@ export class Painter {
         return !imagePosA || !imagePosB;
     }
 
-    useProgram(name: string, programConfiguration?: ProgramConfiguration | null): Program<any> {
+    /**
+     * Finds the required shader and its variant (base/terrain/globe, etc.) and binds it, compiling a new shader if required.
+     * @param name - Name of the desired shader.
+     * @param programConfiguration - Configuration of shader's inputs.
+     * @param defines - Additional macros to be injected at the beginning of the shader. Expected format is `['#define XYZ']`, etc.
```
NathanMOlson, line 662: @HarelM @kubapelc It looks like `defines` was added to the comments here but not to the actual code here. Do you have any insight on what happened here? Was adding `defines` decided to be a bad idea?

I'd like to pass a #define to a program to set the length of a uniform array. For multidirectional hillshade, I would use this for the illumination source directions and colors. For hypsometric tint, I would use this for the colormap. Without being able to set a custom #define, the two alternatives I'm aware of are 1) Set the uniform arrays to a maximum length, hopefully longer than we ever need, or 2) pass the data to the shader using a texture.

Is there any reason not to hook up the functionality described in the comment?

kubapelc: Hmm this is super weird, looking at the file history it seems that the `defines` code never even made it to upstream or a PR, and I can't even find it in my own git history. So I guess I planned to implement it and wrote a doccomment for it, but then ended up not needing it and forgot to remove the comment. But I'm not sure.

I think it makes sense to support it. (But it should also be used somewhat sparingly - it encourages switching shaders often, and switching shaders is one of the most expensive GL calls. But your use case is a good fit.)

PR #3961, opened 2024-04-09
Globe - line layer

kubapelc:
# Line layer adaptation for globe

Issue & discussion of globe:

https://github.com/maplibre/maplibre/issues/190
https://github.com/maplibre/maplibre/discussions/161

Additions and changes since [the previous PR](https://github.com/maplibre/maplibre-gl-js/pull/3882):

- Adapted line layer type for globe
- Added render tests for line layer on globe

[Main vector globe branch with all changes](https://github.com/windycom/maplibre-gl-js/tree/kubapelc/globe-vector)

A vector globe demo image

## Checklist

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.
 - [x] Link to related issues.
 - [x] Include before/after visuals or gifs if this PR includes visual changes.
 - [x] Write tests for all new functionality.
 - [ ] Document any changes to public APIs.
 - [ ] Post benchmark scores.
 - [ ] Add an entry to `CHANGELOG.md` under the `## main` section.

Issue #3187, opened 2023-10-07
fill-extrusion has glitching and missing faces


sebastianoscarlopez:
The fill-extrusion layer misses some faces, also if we set the opacity to < 1 it'll display glitches (maybe z-fighting)

https://github.com/maplibre/maplibre-gl-js/assets/10349981/61625301-b62a-4318-86fe-c3ad0db46e1e

> note: there are other scenarios where the z-fighting is plausible for example when two extrusions have one or more faces in the same position. This issue is with only one geometry.

**maplibre-gl-js version**: 3.3.1

**browser**: Chrome on Mac misses the faces and displays glitches.
Edge on Mac, Chrome on Windows, and Linux miss the faces but not glitches.

### Steps to Trigger Behavior

 1. Extrude a polygon
 2. Move the camera
 3. If we use Chrome on Mac and put opacity < 1 we also see the glitching 

### Link to Demonstration
Here's a reproducible example showing how a rotating extruded square with < 4 walls loses a face and something that looks like a z-fighting

https://jsbin.com/xohogomawi/1/edit?html,output

### Expected Behavior
The square extrusion must have 4 solid walls with no glitches.

### Actual Behavior
The square extrusion has < 4 walls.

tpneumat:
Can confirm same issue, especially noticeable in iOS retina display.  setting opacity to 1 until this can be fixed.

kubapelc:
I believe I've fixed the missing face issue in https://github.com/maplibre/maplibre-gl-js/pull/3968 . Note that your polygon does not have the first and last points repeated, which seems to be invalid according to geojson spec.

However most of MapLibre parses polygons without repeated first/last points just fine, and I think that is reasonable behaviour it makes sense for fill-extrusion to be no different.

Replace the link to maplibre-gl.js with `https://kvaleya.gitlab.io/maplibre/aligndemo4/maplibre-gl-dev.js` in the jsbin for a demonstration.

No idea about the Z-buffer issues though.

ibesora:
With globe, z-fighting is even more visible

The video below is the `3d-extrusion-foorplan` but using the `globe` projection
https://github.com/user-attachments/assets/4b910a56-d81a-49a7-99a7-5b4ddc40ebb4

Issue #2560, opened 2023-05-20
`fill-extrusion` elements distorted by terrain exaggeration


zstadler:
**maplibre-gl-js version**:
2.4.0

**browser**:
Firefox, Chrome

### Link to Demonstration
https://codepen.io/zstadler/pen/bGmQJxJ?editors=0010

### Expected Behavior
![image](https://github.com/maplibre/maplibre-gl-js/assets/1304610/5959442a-b284-42e8-a088-63235729c8ed)

### Actual Behavior
![image](https://github.com/maplibre/maplibre-gl-js/assets/1304610/69e6354b-dbb5-4274-8b1f-bbc3c2674a36)
Increasing the exaggeration increases the size of the "teeth"

prozessor13:
I think this problem is because of the geojson source. I think the geojson-vt library may splits the square into multiple geometries, and then, when rendering, for each sub-geometry a different centroid is calculated. And because of terrain each part is raised differentely. So, my first thing to do would be to debug the geojson-vt tile-geometries.

zstadler:
Isn't it the same with the geometries stored mbtiles? I though all geometries are split between tiles.

Also, [rotating the map and increasing the pitch](https://codepen.io/zstadler/pen/OJBdRbz?editors=0010) shows this:
![image](https://github.com/maplibre/maplibre-gl-js/assets/1304610/d6a967f7-5e83-4d5c-b5e8-07e68da0c6af)

It looks like the original rectangle was was not split and it was properly extruded into a cube. However, there is an additional flat triangle above it.

When [a different bearing is used with the same pitch](https://codepen.io/zstadler/pen/QWZYKrz?editors=0010), there seem to be two such triangles:
![image](https://github.com/maplibre/maplibre-gl-js/assets/1304610/e55f69f5-631f-4ac4-8b36-3d0970fb96d2)

Here is a closer look at the top-right corner:
<image width="10%" src="https://github.com/maplibre/maplibre-gl-js/assets/1304610/780c2c2f-3336-4424-b42c-0558f6f7b6a5"/>

If you play a bit with zoom, pitch and bearing various other artifacts are seen, such as
![image](https://github.com/maplibre/maplibre-gl-js/assets/1304610/6f751d11-7f5a-417f-bd01-470e6f55f980)

prozessor13:
[...]

PR #3968, opened 2024-04-10
Globe - fill extrusion layer

kubapelc:
# Fill extrusion layer adaptation for globe

Issue & discussion of globe:

Additions and changes since [the previous PR](https://github.com/maplibre/maplibre-gl-js/pull/3961):

- Adapted fill extrusion layer type for globe
- Added an example with fill extrusion on globe
- Fixed some issues with fill-extrusion (see below)

A demo image of fill-extrusion on globe

This PR also fixes the missing face issue (but NOT the zbuffer issue) from https://github.com/maplibre/maplibre-gl-js/issues/3187 . This was caused by the code that was generating the side faces - it was expecting the first and last point in a vertex ring to be duplicate, but the source geojson here didn't do that.

However fill-extrusion has several other issues, mostly related to terrain. The problem is that fill-extrusion is placed at the height of the terrain under its centroid, and centroid computation is inconsistent.

I've fixed some of those inconsistencies related to discarding edges/polygons outside tile bounds, which fixes *some* bugs - hence the terrain render test update - you can actually see the roof of the blue rectangle levitating above its walls in the old expected image! But other terrain + fill-extrusion issues stem from the fact that buildings might get split (during tile generation, not runtime) along tile borders into two polygons with different centroids, so two halves of a building might sample different positions in the terrain textures and get different heights. I cannot think of a simple solution to this, and it is outside the scope of a globe implementation.

Related issues:

- https://github.com/maplibre/maplibre-gl-js/issues/3313
- https://github.com/maplibre/maplibre-gl-js/issues/2560 (I believe this one is fixed by this PR)


src/render/draw_fill_extrusion.ts
```diff
@@ -61,6 +62,9 @@ function drawExtrusionTiles(
     const crossfade = layer.getCrossfadeParameters();
     const opacity = layer.paint.get('fill-extrusion-opacity');
     const constantPattern = patternProperty.constantOr(null);
+    const projection = painter.style.map.projection;
+    const globeCameraPosition: [number, number, number] = (projection instanceof GlobeProjection) ? projection.globeCameraPosition : [0, 0, 0];
```
HarelM, line 66: Can this be a method in the interface instead of using `instanceof`?

kubapelc: I've moved it to the interface, but I really don't think any other projection might use these - this is very 3D/globe specific.

src/geo/projection/mercator.ts
```diff
@@ -29,6 +30,10 @@ export class MercatorProjection implements Projection {
         return false;
     }
 
+    get cameraPosition(): vec3 {
+        return [this._cameraPosition[0], this._cameraPosition[1], this._cameraPosition[2]];
```
HarelM, line 34: I think you can use `[...this._cameraPosition]`, can't you?

src/geo/projection/mercator.ts
```diff
@@ -147,6 +158,10 @@ export class MercatorProjection implements Projection {
         this._cachedMesh = new Mesh(tileExtentBuffer, quadTriangleIndexBuffer, tileExtentSegments);
         return this._cachedMesh;
     }
+
+    transformLightDirection(_: Transform, dir: vec3): vec3 {
+        return [dir[0], dir[1], dir[2]];
```
HarelM, line 163: Is there a good reason we clone the array?

src/render/draw_fill_extrusion.ts
HarelM: True, but that's how inheritance works.

src/geo/projection/mercator.ts
kubapelc: The function doc comment promises to return a new array. Also I can use `...dir` here.

Also vectors should IMO be treated as a value type, type/javascript doesn't give us tools to do that (afaik), so the next best thing is to never mutate a vector that is an argument of a function. Code that uses this function might very well depent on the original vector not being mutated.

src/geo/projection/mercator.ts
kubapelc: I can't just use `[...this._cameraPosition]` because a `vec3` must be an array of exactly 3 numbers, but I can use `[...this._cameraPosition] as vec3`, I think that is better than the original state, because it is less error-prone.

kubapelc: I see there is `vec3.clone()` which is even better.

HarelM (approved):

PR #3969, opened 2024-04-10
Upgrage to covecov 4 and add token

HarelM:
See title


HarelM:
Coverage now shows comments and non code parts as covered which is what causing the difference in resutls.

sbachinin (approved):

PR #3979, opened 2024-04-11
Kubapelc/globe pr hillshade

kubapelc:
# Hillshade and background layers adaptation for globe

Both layers are fairly simple, so I merged them into a single PR.

Issue & discussion of globe:

Additions and changes since [the previous PR](https://github.com/maplibre/maplibre-gl-js/pull/3968):

- Adapted background layer type for globe
- Adapted hillshade layer type for globe
- Fixed seams / tiny holes seen usually in oceans when looking south under high pitch (with globe enabled)

A demo of hillshade on a globe

The seams that got fixed:

Screenshot of the seams/holes bug that got fixed

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


kubapelc:
Huh, the failing render test is a difference of only a few pixels at the very edges of the globe, otherwise it looks okay. Should I add the failing image to the test as another expected image or increase the allowed difference threshold?

HarelM:
If you can replace the existing image it would be the best approach, otherwise, add another image. I don't want to change the threshold. Thanks!

kubapelc:
I've added the images from the CI pipeline - they are also correct.

src/render/draw_hillshade.ts
```diff
@@ -18,49 +18,72 @@ export function drawHillshade(painter: Painter, sourceCache: SourceCache, layer:
...
-
-    for (const coord of coords) {
-        const tile = sourceCache.getTile(coord);
-        if (typeof tile.needsHillshadePrepare !== 'undefined' && tile.needsHillshadePrepare && painter.renderPass === 'offscreen') {
-            prepareHillshade(painter, tile, layer, depthMode, StencilMode.disabled, colorMode);
-        } else if (painter.renderPass === 'translucent') {
-            renderHillshade(painter, coord, tile, layer, depthMode, stencilModes[coord.overscaledZ], colorMode);
+    if (painter.renderPass === 'offscreen') {
+        // Prepare tiles
+        for (const coord of tileIDs) {
```
HarelM, line 29: Since you moved the coords "for" into renderHillshade I would advise to move this as well into the method, this way the methods are a bit more similar.

HarelM:
I think it would be great to add a test for the holes you fixed as part of this PR.
I'm not sure we have global data for hillshade, but I would be interested to see how the hillshare behaves on the globe.
Note that in terrain we had issue with hillshare due to render to texture when rotating the map.
If my understanding is correct, there's no render to texture as part of the globe, so I hope this issue doesn't exist here as well, but, it's worth making sure.

HarelM:
Otherwise, looks great, thanks!

kubapelc:
There is currently no global DEM tile available for tests, I would also prefer to use it though. I don't know where to acquire tiles that can be copied into MapLibre repo. I also need a vector tile of the ocean (just a simple polygon) for the seams fix render test, but that shouldn't be too hard to create.

src/render/draw_hillshade.ts
kubapelc: Done!

kubapelc:
I've managed to create a render test for the single-pixel seams fix.

HarelM (approved):

HarelM:
Just to be on the safe side, I would consider making sure the render test of the seams doesn't pass when the fix is not in place (you remove the code that fixes the seams and run the test to see that it fails).
In any case, I'll merge this and you can test the above locally.

kubapelc:
I checked, it makes really noticeable lines when running the old version.

wipfli:
> I also need a vector tile of the ocean (just a simple polygon) for the seams fix render test, but that shouldn't be too hard to create.

I would make a geojson, and convert it to an mbtiles with tippecanoe in this case. From there you can extract the tiles.

kubapelc:
I've already made the tiles using vtzero, but it's good to know that there are tools to convert geojson to tiles, thanks for the tip!

Issue #3623, opened 2024-01-25
Unit test "Map#isSourceLoaded (equivalent to event.isSourceLoaded)" may have stability issue


zhangyiatmicrosoft:
### Steps to Trigger Behavior

 1. npm run test-unit

### Expected Behavior
100% stable and all pass

### Actual Behavior
Roughly 10% chance it would fail on one test "Map#isSourceLoaded (equivalent to event.isSourceLoaded)"

I narrowed it down by commenting out all others in that file and leave only 2:
https://github.com/maplibre/maplibre-gl-js/blob/4ed9cf07b8ee6cb2ad5ed2258287d988252bfa8d/src/ui/map.test.ts#L557C1-L593C12
and running just these two, almost 50% time it would fail.

Interesting enough, running them one by one will always success. I am not sure if the problem is from test cases or production code itself.

This test was added by:
https://github.com/maplibre/maplibre-gl-js/commit/ded439d314293afb20751ba9652483821228c0dd#diff-1369f2165e0778abbb2cdd9a17b1527434e25ed2cd4f9dd49dfe31396d51c426

@ambientlight would you mind to take a look? Anyone else is available to help? Greatly appreciated !!
Meanwhile I will try to find some time to keep digging.

HarelM:
I've seen it too, seems like some sort of race condition with dirty state, but it was too complicated to understand and needed more time.
Feel free to pursue this, I generally think there's an actual bug here...

PR #3992, opened 2024-04-13
Fix race condition in SourceCache

zhangyiatmicrosoft:
## Launch Checklist

Fix for https://github.com/maplibre/maplibre-gl-js/issues/3623

SourceCache.update can be called from several places, the two entry points related to this bug is 
1) Style._updateSources
2) 'data' event handler in SourceCache ctor

- Happy path for 90% of UT runs, and most if not all real map scenarios:
(2) is called first, since <code>this.transform</code> is undefined, SourceCache.update is *not* called, which is correct.
when (1) is called, <code>this._didEmitContent</code> is already set to true, noPendingDataEmissions is true from this line:
https://github.com/maplibre/maplibre-gl-js/blob/da9ac17b09e0e793a4bcf7dc0f0672d30d8f9a83/src/source/source_cache.ts#L553
and idle event is fired.

- Error condition:
(1) is called first, when <code>this._sourceLoaded</code> is falsy, it is rejected by this line but <code>this.transform</code> is still set (bug)!
https://github.com/maplibre/maplibre-gl-js/blob/da9ac17b09e0e793a4bcf7dc0f0672d30d8f9a83/src/source/source_cache.ts#L503C9-L503C61
When (2) enters, because of the bug above, SourceCache.update is entered, <code>this._didEmitContent</code> is false,  idle event cannot be fired but <code>this.updated</code> is still set to true (derived bug)!
All future calls of (1) can never fire idle event because of the derived bug.

- Fix: when <code>update</code> method return because <code>this._sourceLoaded</code> is falsy, do not set <code>this.transform</code> and <code>this.terrain</code>. 
- Enhancements:
-- style.ts was firing a redundant visibility event when no layer is added because of incorrect comparison. Also added comments for readability. This change is well covered by test cases in src/ui/control/attribution_control.test.ts. No new ones are needed.
-- SourceCache: moved data handler to be its own function instead of nesting in ctor. Improve readability.

 - [x] Link to related issues.


src/ui/control/attribution_control.test.ts
```diff
@@ -2,6 +2,8 @@ import {AttributionControl, defaultAtributionControlOptions} from './attribution
 import {createMap as globalCreateMap, beforeMapTest, sleep} from '../../util/test/util';
 import simulate from '../../../test/unit/lib/simulate_interaction';
 import {fakeServer} from 'nise';
+import {Map} from '../../ui/map';
+import {MapDataEvent, MapSourceDataEvent} from '../events';
```
HarelM, line 6:
```suggestion
import {MapSourceDataEvent} from '../events';
```

src/ui/map_loaded.test.ts
```diff
@@ -0,0 +1,81 @@
+import {createMap, beforeMapTest, createStyle, createStyleSource} from '../util/test/util';
```
HarelM, line 1: I'm not sure these two tests are worth the extra file.
I understand why this can be helpful to reproducing the bug, but the same can be achieved with jest's `-t` flag and generally, we try and keep the test file name equal the source code file name.

src/source/source_cache.ts
```diff
@@ -498,9 +490,11 @@ export class SourceCache extends Evented {
      * are inside the viewport.
      */
     update(transform: Transform, terrain?: Terrain) {
+        if (!this._sourceLoaded || this._paused) {
```
HarelM, line 493: This is basically the fix, right?
Other changes seems like a movement of code from one place to the other...?

HarelM:
I've looked at the version 1.x code base and I see this assignment of transform in the update method before the check for loaded etc.
I don't know the history of that, but it seems that it was intentional, I just don't know how to understand why was this intentional...
Any thoughts are welcome here.

Generally speaking, it sounds like it will solve the test issue, I just don't know if it will create other bugs...

src/source/source_cache.ts
zhangyiatmicrosoft: Yes. this is the fix.
Another thing worth doing is the changes in Style.ts where a redundant and useless "visibility" event is now eliminated.
Othters are simple redibilty changes.

zhangyiatmicrosoft:
> I've looked at the version 1.x code base and I see this assignment of transform in the update method before the check for loaded etc. I don't know the history of that, but it seems that it was intentional, I just don't know how to understand why was this intentional... Any thoughts are welcome here.
> 
> Generally speaking, it sounds like it will solve the test issue, I just don't know if it will create other bugs...

I puzzled about that too. At the end the most plausible explanation is that this part of the code was never a problem until https://github.com/maplibre/maplibre-gl-js/pull/2543, which introduced additional logic surrounding <code>this._didEmitContent = true;"</code>. 
I do think this is not just a testing problem though. It is very likely an edge case in real usage as well --- just very difficult to isolate and people usually refresh and it went away.

src/ui/map_loaded.test.ts
zhangyiatmicrosoft: Map.test.ts is way too big already with ~200 test cases. Yes -t flag works but that is just another thing people have to type and just annoying. :)

By any means it should be split up -- I was going to do it in this PR but figured that should not do two different things in same PR, so took a baby step instead.

HarelM:
@ambientlight let me know if there's a chance you can review this as well as you did the original change in #2543.

I'm too under the impression there's a real bug here, not just a test issue.

src/ui/map_loaded.test.ts
HarelM: Let's have this in a different PR then, and keep this PR a bit cleaner.

zhangyiatmicrosoft:
> @ambientlight let me know if there's a chance you can review this as well as you did the original change in #2543.
> 
> I'm too under the impression there's a real bug here, not just a test issue.

Additional info --- Code path 1 (Style._updateSources) is driven from requestAnimation frame, so it is related to machine speed or CPU load at the time. 
These UTs fail 100% on my newest dev machine with 24 core CPU and never happen on an older machine with only 8 cores, which explains why they always pass in CI, which are all slower VMs.

src/ui/map_loaded.test.ts
zhangyiatmicrosoft: done

HarelM (approved):

HarelM:
Seems like a new test has started to fall now (at least it did now on main branch):
https://github.com/maplibre/maplibre-gl-js/actions/runs/8695135980/job/23845526815
So we might not be out of the woods yet...

Issue #3373, opened 2023-11-16
Paint property changed via setPaintProperty is well reflected right after via getPaintProperty but not via getStyle().layers.


DanielForniessoria-TomTom:
**maplibre-gl-js version**: 3.6.1

**browser**: Chrome 119.0.6045.123 MacOS

### Steps to Trigger Behavior

 1. Draw a line with a blue color using a GeoJSON source.
 2. Change e.g. a line-color property of a line layer rendering a GeoJSON LineString via setPaintProperty(YOUR_LAYER, "line-color", "red")
 3. Verify the change visually (line appears red)
 4. Verify the new color property via map.getPaintProperty(YOUR_LAYER, "line-color")
 5. [FAILS] Then verify that line-color property in map.getStyle().layers.filter(layer => layer.id == YOUR_LAYER) -> The line-color inside the paint object still shows "blue".

### Link to Demonstration

The screenshot below showcases the property inconsistency:
<img width="1728" alt="Screenshot 2023-11-16 at 11 41 14" src="https://github.com/maplibre/maplibre-gl-js/assets/73886099/731a4671-7ad5-45e6-9690-f0943d3714d2">

### Expected Behavior

getPaintProperty should be consistent with map.getStyle().layers for the same layer for a paint property changed via setPaintProperty.
The same should apply to getLayoutProperty, and setLayoutProperty (but didn't test that yet)

### Actual Behavior

When setting a paint property via setPaintProperty, it's reflected back via getPaintProperty but not in map.getStyle().layers for that layer.

HarelM:
Style class is more of an internal class, using its properties is not advised. Map object is the main API and it sounds consistent according to the report here...

DanielForniessoria-TomTom:
https://maplibre.org/maplibre-gl-js/docs/API/classes/maplibregl.Map/#getstyle

Looks like a public API to me.

HarelM:
`getStyle` is a public method, but the return value is a class that not all of its properties are public, like `layers`, I think...

DanielForniessoria-TomTom:
What makes you think that property is not public?
"layers" is defined in typescript's **StyleSpecification** which is the type returned in getStyle()
**StyleSpecification** is also mentioned as the return type in the public getStyle() method.

`export type StyleSpecification = {
	"version": 8;
	"name"?: string;
	"metadata"?: unknown;
	"center"?: Array<number>;
	"zoom"?: number;
	"bearing"?: number;
[...]

PR #3993, opened 2024-04-14
Reset serializedLayers for lazy serialization in setPaintProperty

### LINKED ISSUE #2651: Style serialization will miss changes to paint properties that don't require re-layout
When @zhangyiatmicrosoft added lazy serialization in #2306, `_updateLayer` was modified to clear the in-memory serialization whenever a layer changed:

https://github.com/maplibre/maplibre-gl-js/blob/d9f276f9fa4960a91ebb502da7559f147f7a58dd/src/style/style.ts#L1195-L1197

The problem is that in `setPaintProperty`, we only call `_updateLayer` if the layer requires re-layout:

https://github.com/maplibre/maplibre-gl-js/blob/d9f276f9fa4960a91ebb502da7559f147f7a58dd/src/style/style.ts#L1071-L1074

I ran into this as a problem trying to change a layer's `raster-resampling` property back and forth between `nearest` and `linear` by updating the style JSON (ie using `setStyle`, not `setPaintProperty`). With the first change, from `nearest` to `linear`, the style diff picked up the change and modified the paint property, but it skipped clearing `this._serializedLayers`. Then, trying to change back to `nearest`, it diffed against the originally serialized value, didn't see any change, and didn't do anything.

I think the simple fix here is just to promote `this._serializedLayers = null` to be called either way in `setPaintProperty`, but it'd be nice to have @zhangyiatmicrosoft 's eyes on this to look for other cases where something like this might be necessary (I'm not very familiar with the original lazy-serialization PR and what kind of audit we did to figure out where the state had to be reset).

cc @ibesora

kumiko-haraguchi:
## Launch Checklist

Fixes #2651

Related to:
- #3373
- #2651

The style's `_serializedLayers` doesn't reset when `setPaintProperty` is called in some cases. It retains the old value when the style is serialized by `map.getStyle()`. To resolve this issue, reset `_serializedLayers` in the `setPaintProperty` method.

[Before]
[before.webm](https://github.com/maplibre/maplibre-gl-js/assets/28984604/0b0c7e6c-53d6-4604-84e4-8651e4268d2e)

[After]
[after.webm](https://github.com/maplibre/maplibre-gl-js/assets/28984604/36aeff0d-8534-4144-958d-c0bf86e7d326)

----
I also noticed that the documentation includes private methods [_serializebyids](https://maplibre.org/maplibre-gl-js/docs/API/classes/Style/#_serializebyids), [_serializedalllayers](https://maplibre.org/maplibre-gl-js/docs/API/classes/Style/#_serializedalllayers) for the Style class. I removed the doc comments for these methods, as they may not have been intended to be documented.

----
 - [x] Link to related issues.


src/ui/map.test.ts
```diff
@@ -1818,6 +1818,33 @@ describe('Map', () => {
             });
         });
 
+        test('#3373 paint property should be synchronized with an update', done => {
```
HarelM, line 1821: Can we avoid using `done` in tests? It doesn't give a good error message when it fails and using `once` and `async/await` is more readable.

src/style/style.ts
```diff
@@ -470,11 +470,6 @@ export class Style extends Evented {
         return true;
     }
 
-    /**
```
HarelM, line 473: Why were these comments removed?

kumiko-haraguchi: I noticed that the documentation includes private methods [_serializebyids](https://maplibre.org/maplibre-gl-js/docs/API/classes/Style/#_serializebyids), [_serializedalllayers](https://maplibre.org/maplibre-gl-js/docs/API/classes/Style/#_serializedalllayers) for the Style class. I removed these doc comments for these methods, as they may not have been intended to be documented. I'll put them back if these are intended!

Style Doc

HarelM: Ahh, in that case I suggest to use `@hidden` instead, these comments can be meaningful when trying to understand the code.

kumiko-haraguchi: @HarelM Sure, thanks for the suggestion!

HarelM (approved):

PR #3994, opened 2024-04-15
Experimental "fix" for tile LOD when terrain is on

olsen232:
Experimenting with changing the behaviour of raster LOD when terrain is enabled - making it more similar to when terrain is disabled.

### DIFF
--- a/src/geo/transform.ts
+++ b/src/geo/transform.ts
@@ -401,7 +401,7 @@ export class Transform {
                 fullyVisible = intersectResult === 2;
             }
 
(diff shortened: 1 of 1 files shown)


birkskyum:
@olsen232 , this is likely resolved on the `main` branch - can you check?

olsen232:
Yes, fixed or at least much improved
With a bug like "terrain LOD is too low for certain tiles at certain angles / zoom / terrain" it's always going to be subject to interpretation if it's a complete fix or just an improvement

birkskyum:
you're absolutely right - Thanks for checking!

PR #4003, opened 2024-04-15
Fix MapOptions#style is optional in the Map constructor

vicb:
See `if (options.style) this.setStyle(...);` in the constructor

My use case is to use a base layer switcher so I do not want to set a style explicitly when instantiating the map.

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -298,7 +298,7 @@ export type MapOptions = {
      * the schema described in the [MapLibre Style Specification](https://maplibre.org/maplibre-style-spec/),
      * or a URL to such JSON.
      */
-    style: StyleSpecification | string;
+    style?: StyleSpecification | string;
     /**
      * If `false`, the map's pitch (tilt) control with "drag to rotate" interaction will be disabled.
      * @defaultValue true
@@ -387,7 +387,8 @@ const defaultOptions = {
     crossSourceCollisions: true,
     validateStyle: true,
     /**Because GL MAX_TEXTURE_SIZE is usually at least 4096px. */
-    maxCanvasSize: [4096, 4096]
+    maxCanvasSize: [4096, 4096],
+    style: '',
 } as CompleteMapOptions;
 
 /**
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -342,7 +342,7 @@ const defaultMaxPitch = 60;
 // use this variable to check maxPitch for validity
 const maxPitchThreshold = 85;
 
-const defaultOptions = {
+const defaultOptions: Readonly<CompleteMapOptions> = {
     center: [0, 0],
     zoom: 0,
     bearing: 0,
@@ -389,7 +389,16 @@ const defaultOptions = {
     /**Because GL MAX_TEXTURE_SIZE is usually at least 4096px. */
     maxCanvasSize: [4096, 4096],
     style: '',
-} as CompleteMapOptions;
+    container: '',
+    logoPosition: 'bottom-left',
+    antialias: false,
+    maxBounds: undefined,
+    locale: null,
+    collectResourceTiming: false,
+    bounds: undefined,
+    fitBoundsOptions: undefined,
+    pixelRatio: undefined
+};
 
 /**
  * The `Map` object represents the map on your page. It exposes methods
@@ -471,7 +480,7 @@ export class Map extends Camera {
     _locale: typeof defaultLocale;
     _removed: boolean;
     _clickTolerance: number;
-    _overridePixelRatio: number | null;
+    _overridePixelRatio?: number;
     _maxCanvasSize: [number, number];
     _terrainDataCallback: (e: MapStyleDataEvent | MapSourceDataEvent) => void;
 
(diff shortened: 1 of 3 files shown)


src/ui/map.ts
```diff
@@ -387,7 +387,8 @@ const defaultOptions = {
     crossSourceCollisions: true,
     validateStyle: true,
     /**Because GL MAX_TEXTURE_SIZE is usually at least 4096px. */
-    maxCanvasSize: [4096, 4096]
+    maxCanvasSize: [4096, 4096],
+    style: '',
```
HarelM, line 391: If style can be undefined why setting it to an empty sting?

vicb: > If style can be undefined 

This is an ambiguous statement. `options.style` can not be set to undefined but the `style` key can be omitted from `options`.

Then the reason is https://github.com/maplibre/maplibre-gl-js/blob/d76c0447f49ddf9ea238ebe458b2488b03c0b361/src/ui/map.ts#L332-L333

i.e. the value after the default are applied should satisfy `StyleSpecification | string`

HarelM: I'm not sure I'm following...

vicb: CompleteMapOptions#style is a required key of type string | StyleSpecification.

If the style prop is not set in the options passed to the ctor, it has to be added with a tour satisfying string | StyleSpecification.

Does this make sense?

neodescis: Why are you adding `style: ''` though if you're making it optional above?

vicb: @neodescis that's a good question and @HarelM already asked it before.

Please check my answer above first.

The confusion comes from the fact that there are 2 distincts types: the options passed to the ctor (`MapOptions`) and the resolved options with default applied (`CompleteMapOptions`).

Pseudo code would be `const compleOptions: CompleteMapOptions = {...ctorOptions, ...optionDefault}`

I am new to this lib but I can feel that the types are an afterthought and were sometimes added only to stop TS from complaining - for example when using type casts `... as CompleteMapOptions`.

I am going to add a commit to this PR to try and improve this.

HarelM: The code was converted from flow and there was a lot of work to make it typescript compatible, some places are worse than others. We try and improve it with every commit we make. 

Regarding the issue here, I suggest to use the emptyStyle defined in this project, or use "Omit" for the defalut options (I would also consider removing the emoty contaner string here since container is mandatory). Setting an empty sting for a style send wrong to me...

vicb: using `emptyStyle` is a good idea, I just amended the last commit with that and a couple of other minor cleanup.

But yes, it's a can of worms.

For example the default value for `transformRequest` is `null` while it's defined as `transformRequest?: RequestTransformFunction;`

I can help tackle this if there is some interest, my suggestion would be to add https://github.com/piotrwitek/utility-types (MIT licence) to the project and start cleaning the `MapOptions` and the dependencies.

WDYT?

vicb: Turned out using `emptyStyle` broke the tests, I had to revert 🪱

HarelM: :-(

I'm not very keen on introducing a new dependency, we already have too many, so if this can be done without adding a new dependency, I think it will be better.
Generally speaking, any help improving the types would be very much appreciated.

Is it possible, as part of this PR, to only change the types to work correctly without introducing logical changes (such as setting style to '')?

Thanks for all your work on this BTW!

vicb: > I'm not very keen on introducing a new dependency

Makes sense.
Moreover this lib doesn't need much type sorcery and it's probably better not to make it too easy as well :)

> Generally speaking, any help improving the types would be very much appreciated.

I'll get a stab at this.

> Is it possible, as part of this PR, to only change the types to work correctly without introducing logical changes (such as setting style to '')?

This lib is supposed to guarantee that style is `StyleSpecification | string` and I think `''` is the best way to achieve this. If we keep it undefined then we might break client code relying on that. I don't think it's worth potentially breaking client code for the sake of clean code - `''` is an acceptable compromise to me.

HarelM: The current behavior is that the default `style` value is undefined, setting it to '' will change the current behavior and has a higher probability of breaking exiting client code than keeping it as before.
Am I missing something?

vicb: > setting it to '' will change the current behavior and has a higher probability of breaking exiting client code than keeping it as before.

It all depends what you consider the source of truth. I am definitely a TS developer and consider types the source of truth and I like to use TS strict mode. I you think the average user look at the source code rather than than the TS types, you might be right.

HarelM: It's not that I think developers look at the code, but if you have something that is working right now, this might break it, changing only the types will make it more correct and might cause a break in the build, but not in run time. 
It's a slim chance, but if the purpose of this change is to improve the public facing API types, let's focus on changing the types only at this point and avoid unexpected changes to the "logic".

Not sure I'm successful in passing my point here...

vicb: Can we discuss on slack (victor at suumit.com). I'm busy tonight but should be available tomorrow. Thx.

HarelM: Are you in the maplibre slack? I can't find you using a simple "victor" search... 😕

HarelM:
I believe this was addressed by #4151, right?
- #4151
Closing, if this is not the case, please let me know and I'll reopen this PR.

PR #4005, opened 2024-04-15
Split up map.test.ts

zhangyiatmicrosoft:
## Launch Checklist

This is a pure engineering effort.
- Map.test.ts has ~200 tests and it is 3k lines, which makes it difficult to navigate. Running the test generates 5 pages of result to scroll through. Splitting up is desirable.
- Roughly 80% of tests do not need fake server, so skipping it before/after speeds up the testing process.
- Moving all new tests to a "map_tests" folder to avoid clutters.
- Moving all existing tests in "src/ui/map" to a "src/ui/map_tests" and rename them according to others as well.


zhangyiatmicrosoft:
@HarelM I saw your comments regarding additional test instabilities. It never fails for me but will certain keep an eye on it and fix when have better repro. Thanks

HarelM:
I think the problem is not with the test file, but with the original class - the map.ts file, which is super long and complicated.
I think some/most of the tests in map.test.ts file are not really unit tests, but more of integration tests as they test a lot of code that is not part of the map.ts file.
What do you think about moving these tests to integration tests folder under map folder maybe and have all these tests there?

I think that breaking the convention of having a single test file for every single code file it not a great idea as people will have a hard time understanding where to place new tests, won't they?

zhangyiatmicrosoft:
> I think the problem is not with the test file, but with the original class - the map.ts file, which is super long and complicated. I think some/most of the tests in map.test.ts file are not really unit tests, but more of integration tests as they test a lot of code that is not part of the map.ts file. What do you think about moving these tests to integration tests folder under map folder maybe and have all these tests there?
> 
> I think that breaking the convention of having a single test file for every single code file it not a great idea as people will have a hard time understanding where to place new tests, won't they?

Agree ... map.ts has 3.3k lines and just daunting to work with. I doubt anyone has time (or even want to) split that anytime soon. :)
Just prefer not to scroll 5 pages of test result while running map.test.ts...

Sure I can easily move all these ~20 files over to integrations and leave only one "map.test.ts", is that what you are proposing?

HarelM:
Yes, unless you have a better idea.
Maybe try to keep a few actual unit tests?

zhangyiatmicrosoft:
> Yes, unless you have a better idea. Maybe try to keep a few actual unit tests?

FYI though, "map_events.test.ts" already exists and one test per file convention is already broken for map.ts, which maybe qualify for an exception anyway. :)
I am impartial either way, so if you still want to move them, I will do so tomorrow. Thanks

HarelM:
When I run jest all I care is if a test is failing or not, I'm not scrolling through anything, but I might see things differently.
I just looked again in the repo and the map folder under the UI has only tests, so I guess this is already implemented and both of us didn't know about it...
I suggest to move all the tests files under map_tests folder, including the existing tests and remove map.test.ts completely to avoid people adding tests there by mistake. 

So basically what you did, with the addition of 4 more files.

zhangyiatmicrosoft:
> When I run jest all I care is if a test is failing or not, I'm not scrolling through anything, but I might see things differently. I just looked again in the repo and the map folder under the UI has only tests, so I guess this is already implemented and both of us didn't know about it... I suggest to move all the tests files under map_tests folder, including the existing tests and remove map.test.ts completely to avoid people adding tests there by mistake.
> 
> So basically what you did, with the addition of 4 more files.

Good call. 
Tried to move them but UT and IT do not have the same jest config (one is jsdom and the other node) so existing UTs cannot run.
https://github.com/maplibre/maplibre-gl-js/blob/0502606ae5eb5cc0df5bac242ef2a4104bd425d1/jest.config.ts#L37
In the new iteration I just move all to "map_test" folder, is that better?

PR #1800, opened 2022-10-31
Add contentMatch to sprites

DavidBuerer:
This is to provide a way to work around shields with small text getting too small (left) without creating multiple layers with multiple copies of the shields.

image

The content rectangle alone is not sufficient to keep shields looking nice when using "icon-text-fit" as it allows the shields to get too narrow if the text is small.  To solve this, a limit is needed on how much distortion can happen to a sprite with content.

A few considerations:
1. Since the size of the text is not known to the containing sprite, a minimum width and height would not work, but an aspect ratio does.
2. We could let the designer specify an aspect ratio, but this would generally require an icon designer to calculate the aspect ratio from their "content" rectangle. Rather than making them do this, I use the "content" rectangle's aspect ratio.
3. In addition to an aspect ratio, we need to know whether we are locking the text size horizontally or vertically to the aspect ratio. A common case will be horizontal for text that is written horizontally, however, if using vertical Chinese characters, for example, vertical should be supported.
4. Since this value could vary per sprite, it is best to make it available on sprite rather than on the layer.

So what I did is add a contentMatch enum to sprites that may be 'none', 'height', and 'width'.  The content rectangle will stretch to match the specified dimension and the other dimension will scale with the content, but not shrink below the aspect ratio of the content rectangle.  

This image illustrates the effect that the different values have on sprites.

image

## Launch Checklist

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.
 - [x] Link to related issues.
 - [x] Include before/after visuals or gifs if this PR includes visual changes.
 - [x] Write tests for all new functionality.
 - [x] Document any changes to public APIs.
 - [ ] Post benchmark scores.
 - [x] Manually test the debug page.
 - [x] Add an entry to `CHANGELOG.md` under the `## main` section.

PR #4019, opened 2024-04-20
Add textFitWidth and textFitHeight to sprites

DavidBuerer:
This PR is to re-open https://github.com/maplibre/maplibre-gl-js/pull/1800 with the changes recommended by @zekefarwell and picked by vote.  To quote the original PR: This is to provide a way to work around shields with small text getting too small (left) without creating multiple layers with multiple copies of the shields.

image

To do this, a textFitWidth and textFitHeight property are added to sprites.  The anticipated use (besides setting both to stretchOrShrink which is the same as not adding them) is to set textFitWidth=stretchOnly and textFitHeight=proportional

The following is the table of results using narrow content and wide content for the property.
image
Refreshing my memory on the proposal I came to a different understanding of the implementation, so this table is different than the one I had come up with manually before.  Note that if neither axis is `proportional`, there is basically no effect.  Let me know if I am interpreting this incorrectly and if `stretchOnly` makes sense without `proportional`.  Note also that setting both to `proportional` requires a default axis.  I have it set to height, which leads to some odd results in longer strings as per the table.


HarelM:
Can you please open an issue in maplibre-style-spec repo with the description here and the documentation updates of the sprite json? 
This will allow us to both have all the spec changes in one place and also allow native library to follow up with the implementation. 

Also please check why the tests fail.

DavidBuerer:
> Thanks for taking the time to open this PR! Can you please open an issue in maplibre-style-spec repo with the description here and the documentation updates of the sprite json? This will allow us to both have all the spec changes in one place and also allow native library to follow up with the implementation.
> 
> Also please check why the tests fail.

Will do.  

Not sure why the ubuntu rendering is different.  Will have to get a VM spun up on Monday to look at it.  Is there a way to see the results.html that the server build generated?

Also not clear on why the package is that much bigger.  Any thoughts there?

HarelM:
The build summary has the html file with the images generated from the linux worker.
The images are base64 encoded, so you can extract them and commit them.
But I'm not sure I understand why existing tests would fail...

src/style/style_image.ts
```diff
@@ -31,6 +31,27 @@ export type StyleImageData = {
     spriteData?: SpriteOnDemandStyleImage;
 };
 
+/**
+ * Enumeration of possible values for StyleImageMetadata.textFitWidth and textFitHeight.
+ */
+export enum TextFit {
```
HarelM, line 37: Should be const enum, to avoid enum quirkiness.

src/symbol/shaping.ts
```diff
@@ -802,6 +803,49 @@ function shapeIcon(
     return {image, top: y1, bottom: y2, left: x1, right: x2};
 }
 
+/**
+ * Called after a PositionedIcon has already been run through fitIconToText,
+ * but needs further adjustment to apply textFitWidth and textFitHeight.
+ * @param shapedIcon - The icon that will be adjusted.
+ * @returns Extents of the shapedIcon with text fit adjustments if necessary.
+ */
+function applyTextFit(shapedIcon) {
```
HarelM, line 812: Please add unit tests for this new method.

test/examples/textFit.html
```diff
@@ -0,0 +1,150 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <title>MapLibre GL JS debug page</title>
```
HarelM, line 5: This will be added to the examples, if this is intended, please add a better title and og:description, otherwise delete this page.

test/integration/render/tests/icon-text-fit/textFit-anchors-long/README.md
```diff
@@ -0,0 +1,2 @@
+This an adaptation of stretch-nine-part-content but with a long
```
HarelM, line 1: I believe you can use the metadata in the style.json file for this kind of comments.

test/examples/textFit.html
DavidBuerer: Removed

HarelM:
I would consider keeping the existing tests as is and create new tests. This would make sure this is a non breaking change. 
If this is a breaking change, then it can only be part of a breaking change version which might happen at September the earliest, as we have released a breaking change version at the beginning of this year...

DavidBuerer:
> I would consider keeping the existing tests as is and create new tests. This would make sure this is a non breaking change. If this is a breaking change, then it can only be part of a breaking change version which might happen at September the earliest, as we have released a breaking change version at the beginning of this year...

That was my bad.  Sometimes unit tests are alongside their targets and sometimes in tests.  Didn't realize this was the later.  Moved tests over.

src/render/image_atlas.ts
```diff
@@ -4,6 +4,7 @@ import {register} from '../util/web_worker_transfer';
 import potpack from 'potpack';
 
 import type {StyleImage} from '../style/style_image';
+import {TextFit} from '../style/style_image';
```
HarelM, line 7: `impoer type ...`?

test/integration/symbol-shaping/shaping.test.ts
```diff
@@ -389,3 +389,77 @@ describe('fitIconToText', () => {
     });
 
 });
+
+describe('applyTextFit', () => {
```
HarelM, line 393: I think you can move this test to be next to `shaping.ts` i.e. `shaping.test.ts` as you are unit testing the method.
Some other tests in this file (not sure if all of them) are using the file system to compare the results of the test and in some cases to update the test results.
But the tests here are unit tests and doesn't use the file system at all as far as I understand.

HarelM: Feel free to move other tests that are not using any json files to that file.

test/integration/render/tests/icon-text-fit/textFit-grid-long/style.json
```diff
@@ -0,0 +1,182 @@
+{
+    "version": 8,
+    "metadata": {
+      "test": {
+        "width": 360,
+        "height": 180,
+        "description": "This draws the matrix of all possible combinations of textFitWidth (columns) and textFitHeight (rows) where they proceed from the top-left in the order stretchOrShrink, stretchOnly, and proportional.  This version shows the effect of a long string with a wider aspect ratio than the source image."
```
HarelM, line 7: Can you add a render test where the text is from top to bottom to see the effect when the text stretches the icon downwards?

src/symbol/shaping.ts
```diff
@@ -802,6 +803,56 @@ function shapeIcon(
...
+    y2: number;
+}
+
+/**
+ * Called after a PositionedIcon has already been run through fitIconToText,
+ * but needs further adjustment to apply textFitWidth and textFitHeight.
+ * @param shapedIcon - The icon that will be adjusted.
+ * @returns Extents of the shapedIcon with text fit adjustments if necessary.
+ */
+function applyTextFit(shapedIcon: PositionedIcon): Box {
```
HarelM, line 819: I'd like to see the coverage on this method once the unit test are moved, as I don't have coverage for the integration tests, only on render and unit tests.

src/symbol/quads.ts
```diff
@@ -65,8 +65,12 @@ export function getIconQuads(
     const imageWidth = image.paddedRect.w - 2 * border;
     const imageHeight = image.paddedRect.h - 2 * border;
 
-    const iconWidth = shapedIcon.right - shapedIcon.left;
-    const iconHeight = shapedIcon.bottom - shapedIcon.top;
+    let icon = {
```
HarelM, line 68:
```suggestion
    let icon: Box = {
```

src/symbol/shaping.ts
```diff
@@ -802,6 +803,56 @@ function shapeIcon(
     return {image, top: y1, bottom: y2, left: x1, right: x2};
 }
 
+export interface Box {
```
HarelM, line 806: Isn't `PositionedIcon` good enough? Why do we need this extra type?

src/symbol/quads.ts
```diff
@@ -88,28 +92,39 @@ export function getIconQuads(
 
     if (image.content && hasIconTextFit) {
         const content = image.content;
+        const contentWidth = content[2] - content[0];
```
HarelM, line 95: Is this where the stretching happens? Maybe we can have the two methods of stretch and fit in the same file/place?

src/symbol/shaping.ts
DavidBuerer: applyText fit returns only the box (x1,y1,x2,y2).  If it returned a PositionedIcon, then it returns an image and collisionPadding also.  If space was a concern, I could see it, but it is a type that goes away once transpiled.

test/integration/symbol-shaping/shaping.test.ts
DavidBuerer: You already asked me to move it out, but I can move it back.

src/render/image_atlas.ts
DavidBuerer: It looks like enums cannot be treated as types.

src/symbol/quads.ts
DavidBuerer: Done

src/symbol/quads.ts
DavidBuerer: It could.  I attempted to lay it out so that if you don't set textFitWidth or textFitHeight the original logic was preserved; to reduce the chance that this change affects mainstream scenarios.

DavidBuerer:
The images for textFit-grid-long generate differently on my machine than on the render server.  I can "fix" this by using the images from the server, but then they may not run locally successfully.  I am continuing to investigate why this is.
image

HarelM:
There are differences between machines, what we usually do is add more than one image to a folder so that it passes locally and on the CI server.

test/integration/symbol-shaping/shaping.test.ts
HarelM: I don't think I did, but the important part from my point of view is to keep integration tests in integration folder and unit tests next to the file, it's sometimes not a clear cut obviously, and I might have said something that got lost in translation.
Thanks for taking the time to do so!

src/symbol/quads.ts
HarelM: That's the right approach, no argue here, I was thinking more about refactoring of moving similar logic to the same file.

test/integration/render/tests/icon-text-fit/textFit-grid-long/style.json
DavidBuerer: Added textFit-grid-long-vertical

src/symbol/shaping.ts
DavidBuerer: They have been moved.  Is there action you want me to take?

test/integration/render/tests/icon-text-fit/textFit-grid-long/style.json
DavidBuerer: Also added textFit-grid-short-vertical.  The results are somewhat unexpected to me... but could still be "correct".

HarelM:
Based on the coverage report (although the result may change when all the render test pass) there are still some lines missing. 
Can you add a test to cover the missing lines (I'm not talking about partial if yellow lines, but rather missing red lines in the codecov website or the jest report).

src/symbol/shaping.ts
HarelM: Improve coverage as stated below.

HarelM:
A few last comments: 
1. Please update the build size to pass the build test
2. Please see if there are more places that needs tests, as I see there are some coverage gaps in newly added lines
3. Please open a PR in the maplibre-style-spec repo to update the documentation about sprites, please make sure to add a sdk support table to the new parameters alongside some of the explanations here. I just finished refactoring these docs and it should be fairly easy to update them (look for sprite in the docs folder there).

Thanks!!

HarelM (approved):

DavidBuerer:
> A few last comments:
> 
> 1. Please update the build size to pass the build test
> 2. Please see if there are more places that needs tests, as I see there are some coverage gaps in newly added lines
> 3. Please open a PR in the maplibre-style-spec repo to update the documentation about sprites, please make sure to add a sdk support table to the new parameters alongside some of the explanations here. I just finished refactoring these docs and it should be fairly easy to update them (look for sprite in the docs folder there).
> 
> Thanks!!

Updated build size and addressed missing coverage lines.  Will start in on the docs.

DavidBuerer:
Is there anything else that I have to do, or am I waiting for someone with write access to complete?

DavidBuerer:
> The style spec repo changes I mentioned above.

Working on that, but that has to be separate PR.  You want that in place before completing this?

HarelM:
Yes, to make sure that when this is merged and released the style spec is updated as well. 
Sorry for linking the two, but I think it's for the best.

DavidBuerer:
https://github.com/maplibre/maplibre-style-spec/pull/637

Issue #3888, opened 2024-03-23
Globe - Add an atmosphere layer


Pheonor:
## User Story

 As a User I can add a realistic atmosphere layer to the Globe, so that the Globe look more realistic when view from a far distance.

## Rationale

A realistic atmosphere layer that appear and disappear fonction of the zoom level could add a touch of realism to Globe display.

## Impact

A new layer with its associated style and configuration should be available.

HarelM:
See my comment about the sky spec here:
- https://github.com/maplibre/maplibre-gl-js/issues/3889

HarelM:
Bounty direction:
https://github.com/maplibre/maplibre/issues/190

vlarrieu:
@HarelM Could you assign it to me please ?

kubapelc:
I think it would also make sense to add an option for physically based sky, fog and atmosphere simulation to the sky implementation. Not right now, but sometime in the future. I think it could result in some very nice visuals, and provide an easy option for adding realistic sky and fog to terrain, and perhaps integrate seamlessly with atmosphere for globe.

@Pheonor your globe's atmosphere implementation already looks very realistic. Do you actually simulate light scattering in the atmosphere?

Pheonor:
Yes, I simulate light scattering with the Rayleigh and Mie phases. It allows to have a reallistic rendering of the atmosphere.
In my previous PoC, I remove the atmosphere when I zoom close to the Earth to avoid adding blue effect on all rendering.
I want to allow more configurability to let atmosphere render even when close to the ground.

HarelM:
If this is the case, I would propose to open an issue in the spec repo and discuss the needed changes to the spec, either use the globe issue, the sky issue or a different one if this should be a different field, although I'm not sure it doesn't fall under the sky specifications.
- https://github.com/maplibre/maplibre-style-spec/issues/568
- https://github.com/maplibre/maplibre-style-spec/issues/163

CommanderStorm:
I think this issue is resolved, as the following was merged, right?
- https://github.com/maplibre/maplibre-gl-js/pull/4020#issuecomment-2181690420

HarelM:
Yes, thanks, this is available in latest pre-release.

onuryb:
Maplibre sky is not displayed everywhere is covered with maps

CommanderStorm:
@onuryb could you please open a proper issue. This is likely very unrelated given the other issues you also posted the same comment in.

[...]

PR #4020, opened 2024-04-20
Add an Atmosphere layer for Globe (#3888)

Pheonor:
## Launch Checklist

This PR add a realistic atmosphere layer to the Earth when using a globe projection.
Some options allows to change the zoom level to fade in and out the atmosphere.
By default, the sun position is computed based on current time.
An optionnal date and time allows to set the sun to any position.

Capture d’écran 2024-04-20 à 09 48 45

 - [X] Link to related issues.


HarelM:
Can you please add some render tests (and maybe unit test) and resolve conflict before I review this thoroughly?

Pheonor:
I resolve conflict. I will add some test.

src/geo/projection/globe.ts
```diff
@@ -72,11 +72,19 @@ export class GlobeProjection implements Projection {
...
+    private _globeMatrixNoCorrection: mat4 = mat4.create();
     private _globeProjMatrix: mat4 = mat4.create();
     private _globeProjMatrixNoCorrection: mat4 = mat4.create();
 
     private _cameraPosition: vec3 = [0, 0, 0];
 
+    private _globePosition: vec3 = [0, 0, 0];
+    private _globeRadiusPixels: number = 0.0;
+
+    private _projMatrix: mat4 = mat4.create();
```
kubapelc, line 85: I think the atmosphere math can be done with matrices that were already present.

The `_globeProjMatrix` matrix assumes the globe is a unit sphere located at (0,0,0), unrotated (so the null island is always at the unit sphere's +Z (0,0,1), see `_angularCoordinatesToVector` ) and projects it to screen. The already present `cameraPosition` property is the camera position relative to the unit-sphere globe.

src/render/program/projection_program.ts
```diff
@@ -1,21 +1,27 @@
...
 
 export type ProjectionPreludeUniformsType = {
     'u_projection_matrix': UniformMatrix4f;
     'u_projection_tile_mercator_coords': Uniform4f;
     'u_projection_clipping_plane': Uniform4f;
     'u_projection_transition': Uniform1f;
     'u_projection_fallback_matrix': UniformMatrix4f;
+    'u_globe_position': Uniform3f;
+    'u_globe_radius': Uniform1f;
+    'u_inv_proj_matrix': UniformMatrix4f;
```
kubapelc, line 13: I think these new uniforms would be better suited for `atmosphereUniformsType`. ProjectionData and `ProjectionPreludeUniformsType` is passed to every single shader, but the uniforms added here are only used by atmosphere.

src/geo/projection/globe.ts
Pheonor: The atmosphere math are performed into camera frame. I need the real globe position into the camera frame and to project each pixel as a ray into this frame also.
This is why I extract the projection matrix and compute the globe position.
The previously computed `cameraPosition` do not give me the expected value. I will try to reuse some other value to avoid a new computation.

src/render/program/projection_program.ts
Pheonor: Ok, I will do like that.

Pheonor: Done.
To be clean, it could be better to centralize all the projection code from `transform.ts`into `mercator.ts`.
But I propose to not do that in the frame of this PR and do this into a refactoring PR.

src/geo/projection/globe.ts
Pheonor: I try to view if I could use `cameraPosition`directly, but this position include the projection effect. I need the globe position in the camera frame without the projection effect.

Pheonor:
I propose to wait the dicussion on https://github.com/maplibre/maplibre-style-spec/issues/163 before adding some render test with the proposed style.

HarelM:
Sure, that's a good idea.
I hope this discussion won't take too long...

src/render/program/projection_program.ts
kubapelc: I would also like to do that, I already started looking at adapting interactions (map panning, etc.) and animations (easeTo, etc.) for globe, changing the transform and refactoring this will definitely be required. As for controls, I would like to have a "globe" and "flat" control scheme (as opposed to "mercator"), since the flat controls would likely get used by all projections except globe. As suggested at the end of this comment: https://github.com/maplibre/maplibre-style-spec/issues/568#issuecomment-1998113647

src/geo/projection/globe.ts
kubapelc: Maybe this shader could be helpful: https://github.com/maplibre/maplibre-gl-js/blob/globe/src/shaders/fill_extrusion.fragment.glsl

It computes ray-globe intersection for large fill-extrusions, which is very close to what atmopshere needs. The only part the linked shader doesn't do is getting the ray from a pixel, but your approach with the inverse matrix should work fine.

(The varying `v_sphere_pos` contains the current pixel's position on the globe, relative to the unit sphere, so a point on the surface would have distance 1 from 0,0,0, a point higher up would have greater distance. See the matching vertex shader for more details.)

kubapelc:
I finally had the time to play with the example. It looks amazing!

I'm worried about performance though, I also tried to run it on my phone (Snapdragon 636, 1080x2280 pixel screen) and got framerate in the single digits, and the sunlight was missing for some reason.

Pheonor:
> I finally had the time to play with the example. It looks amazing!
> 
> I'm worried about performance though, I also tried to run it on my phone (Snapdragon 636, 1080x2280 pixel screen) and got framerate in the single digits, and the sunlight was missing for some reason.
> 
> Edit: also runs at low framerate on Intel HD 630.

Thank you :)
Effectively, I need to check if I could tune step parameters to improve performance.

HarelM:
Can this PR be updated?
I think we have finalized (hopefully) the sky definitions, so I think this can be pushed forward, right?
I'll work on merging the style spec changes for both projection and sky so that the style spec package will be up to date with latest definitions.

Pheonor:
Based on the updated atmosphere style sepc, sure !

Pheonor:
@HarelM I update with the new style definition and I use the sky/atmosphere-blend parameter to decide if the atmosphere should be visible or not. The default value of 0.8 add an atmosphere to any project even if not required. Do you think a default value of 0.0 could be used to ensure no atmosphere is displayed until required ?

HarelM:
Yes, the default should be 0 to keep rendering backward compatible. 
We can change that in the next breaking change version, but I prefer to keep things backward compatible right now.

HarelM:
Unless you are talking about the default when globe is "on", which is not the default. 
In that case I have no good opinion, and a way to present which are the options would be the best approach, IMHO.

Pheonor:
> Yes, the default should be 0 to keep rendering backward compatible. We can change that in the next breaking change version, but I prefer to keep things backward compatible right now.

The default value of 0.8 is define directly into the style spec. Could you made the change please ?

Pheonor:
> Unless you are talking about the default when globe is "on", which is not the default. In that case I have no good opinion, and a way to present which are the options would be the best approach, IMHO.

No, I render the atmosphere only when a globe projection is selected.

HarelM:
I see, if we only render the atmosphere when globe projection is selected then there's no problem, and I think atmosphere should be on by default so that it will look good when someone uses a globe (using globe is not the default).
I'm not sure what should be the default, but let's play with it a bit and better understand what the default should be, and then set this default.
Sounds like a good plan?

Pheonor:
> I see, if we only render the atmosphere when globe projection is selected then there's no problem, and I think atmosphere should be on by default so that it will look good when someone uses a globe (using globe is not the default). I'm not sure what should be the default, but let's play with it a bit and better understand what the default should be, and then set this default. Sounds like a good plan?

Ok, sounds like a good plan.
I will update existing test with globe to hide the atmosphere and ensure results are good.

Pheonor:
@HarelM I update this PR with new style spec definition for sky and use the light position definition. I also add some tests. I think you could review it now.

HarelM:
@Pheonor did you address the comments made by @kubapelc?
Also, it seems that there are more tests to be added as the coverage seems a bit low...

test/examples/globe-atmosphere.html
```diff
@@ -0,0 +1,67 @@
...
+                    7, 0
+                ]
+            },
+            'light': {
+                'anchor': 'map',
+                'position': [1.5, 90, 80]
+            }
+        },
+        atmosphere: true,
+        atmosphereOptions: {
```
HarelM, line 59: Is this still relevant?

test/examples/globe-atmosphere.html
```diff
@@ -0,0 +1,67 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>Display a globe with a satellite map</title>
+    <meta property="og:description" content="Display a globe with a satellite raster baselayer." />
```
HarelM, line 5: Maybe add a word about the atmosphere here?

src/ui/map.ts
```diff
@@ -443,6 +445,7 @@ export class Map extends Camera {
     painter: Painter;
     handlers: HandlerManager;
     projection: Projection;
+    atmosphere: Atmosphere;
```
HarelM, line 448: Shouldn't this be part of the sky object too?

HarelM:
I know this is a bit out of scope, but can we incorporate the sky implementation as part of this PR?
I see a lot of similarity, which is not a surprise.
I can push the sky PR to main and then merge it to the globe, but I'm not sure it's worth the complexity, what do you think?
- #3645

src/geo/projection/projection.ts
```diff
@@ -180,6 +195,20 @@ export interface Projection {
      */
     getMeshFromTileID(context: Context, tileID: CanonicalTileID, hasBorder: boolean, allowPoles: boolean): Mesh;
 
+    /**
+     * @internal
+     * Transform of `LngLat` coordinate with its `elevation` with the projection.
+     * @param lngLat - The longitude and latitude coordinate to transform.
+     * @param elev - The elevation coordinate to transform.
+     */
+    transformPosition(lngLat: LngLat, elev: number): vec3;
```
HarelM, line 204: use `elevation` instead of  `elev`.

src/geo/projection/projection.ts
```diff
@@ -92,6 +92,21 @@ export interface Projection {
      */
     get vertexShaderPreludeCode(): string;
 
+    /**
+     * Globe position in camera frame.
+     */
+    get globePosition(): vec3;
```
HarelM, line 98: These are not great as this is an interface for all the projections, some of them are not globe.
Can you think of a better way to solve this?

src/render/atmosphere.ts
```diff
@@ -0,0 +1,158 @@
+import {TriangleIndexArray, AtmosphereBoundsArray} from '../data/array_types.g';
+import {atmosphereAttributes} from '../data/atmosphere_attributes';
+import {SegmentVector} from '../data/segment';
+import {Painter} from './painter';
+import {Mesh} from './mesh';
+import {mat3, mat4, vec3} from 'gl-matrix';
+import {Projection} from '../geo/projection/projection';
+import {LngLat} from '../geo/lng_lat';
+
+export class Atmosphere {
```
HarelM, line 10: This class needs more tests to cover its logic.

src/ui/map.ts
Pheonor: For the moment, Sky class contains just the style definition, not render elements as mesh that need a Context from the painter.
The Atmosphere class contains the render stuffs.
What is your proposition ? merge the two classes ? instantiate the atmosphere class into sky class ?

test/examples/globe-atmosphere.html
Pheonor: Effectively, I forget to remove this elements.

Pheonor:
> I know this is a bit out of scope, but can we incorporate the sky implementation as part of this PR? I see a lot of similarity, which is not a surprise. I can push the sky PR to main and then merge it to the globe, but I'm not sure it's worth the complexity, what do you think?
> 
>     * [Basic Sky Implementation working with actual codebase #3645](https://github.com/maplibre/maplibre-gl-js/pull/3645)

I look at the Sky PR during the implementation of this PR. It could propably be merged after this one also.

src/ui/map.ts
HarelM: Check out the sky PR, I don't see any addition to the map object besides the sky specification API.
I'm not sure what is the correct solution here though.

Pheonor: Ok, I will use the same approach.

src/geo/projection/projection.ts
Pheonor: Ok, I propose to use `worldCenterPosition` and `worldSize` as more general properties. So, now, the information is available for any projection.

src/render/atmosphere.ts
Pheonor: I simplify this class to remove the Sun position computation from the current time as with the style default value, this code could not be used. I also remove this class to split the code between Sky ans Light Style classes and draw_atmosphere method.

Pheonor:
> @Pheonor did you address the comments made by @kubapelc? Also, it seems that there are more tests to be added as the coverage seems a bit low...

I tune a little the atmosphere parameters. Render seems style Ok. I hope performance are better.

kubapelc:
I've also tried how well atmosphere runs with the reduced sample counts back when the PR was first opened, and it doesn't help all that much with performance on integrated gpus. I'm not sure such a performance-heavy feature should be on by default. Sorry for mentioning this after you already did the style spec.

src/geo/projection/projection.ts
kubapelc: See my previous comment, no new projection properties should be needed for atmosphere, except maybe inverse projection matrix. Lighting can be done in world space instead of camera space, it is common practice in games. In world space, the globe is always a unit sphere placed at 0,0,0. I can suggest the code changes for this if you want.

https://github.com/maplibre/maplibre-gl-js/pull/4020#discussion_r1573919136

Pheonor: I change the parameter to be more generic.
I do not understand what it is so important to do the render in world space instead of camera space.
What improvement could be expected ?

kubapelc: Good point, the benefits aren't very significant. It would make the code a bit smaller and maybe help with maintainability, since computing lighting is usually done in world space (or at least I have the impression that is the case).

Pheonor: Ok. In fact, the shader computes atmospheric transmittance. Many implementations do it in camera space. If it is for you, I propose to let the shader like that.

Pheonor:
@HarelM  I add some Unit Tests to improve the coverage. I hope its better now.

I have some problem to test the `-transition` option on the `atmosphere-blend`. As far I understand, the `-transition` option should be check into the Style Spec. Do you confirm ?

src/geo/projection/projection.ts
```diff
@@ -180,6 +195,12 @@ export interface Projection {
      */
     getMeshFromTileID(context: Context, tileID: CanonicalTileID, hasBorder: boolean, allowPoles: boolean): Mesh;
 
+    /**
+     * @internal
+     * Return true if the projection correspond to a Globe.
+     */
+    isGlobe(): boolean;
```
HarelM, line 202: Is there a way to avoid this method?

src/style/light.ts
```diff
@@ -132,4 +134,22 @@ export class Light extends Evented {
             styleSpec
         }));
     }
+
+    getSunPos(transform: Transform): vec3 {
```
HarelM, line 138: Is this the right place for this math? I think that most of the math is done in the draw calls and not here, am I right?

src/render/painter.ts
```diff
@@ -536,6 +537,11 @@ export class Painter {
             this.renderLayer(this, sourceCache, layer, coords);
         }
 
+        // Render atmosphere, only for Globe projection
+        if (this.style.map.projection.isGlobe()) {
```
HarelM, line 541: Maybe you can use the new style projection property instead of the projection instance?
I'll see if I can add it to the globe branch so that you can use it.

HarelM:
> I have some problem to test the -transition option on the atmosphere-blend. As far I understand, the -transition option should be check into the Style Spec. Do you confirm ?

Not sure I fully understand the question, can you clarify what you mean?

Pheonor:
> > I have some problem to test the -transition option on the atmosphere-blend. As far I understand, the -transition option should be check into the Style Spec. Do you confirm ?
> 
> Not sure I fully understand the question, can you clarify what you mean?

I try to test a 'transition' on the new `atmosphere-blend` parameter but the Style Validation does not accept the `atmosphere-blend-transition` parameter. I check quickly the code into Maplibre Style Spec and effectively, the `-transition` parameters are not managed.
I am not sure If I need to test this `-transition` part or if an update of the MapLibre Style Spec chould be made.

src/render/painter.ts
Pheonor: Ok, I am waiting the PR to use it.

src/style/light.ts
Pheonor: Ok, I will move it.

src/geo/projection/projection.ts
Pheonor: I could check directly to the projection name.

src/render/painter.ts
Pheonor: Updated.

src/style/light.ts
Pheonor: Move into draw_atmopshere.

src/geo/projection/projection.ts
Pheonor: Done

src/geo/projection/projection.ts
```diff
@@ -1,13 +1,13 @@
-import type {mat4, vec3} from 'gl-matrix';
-import type {Tile} from '../../source/tile';
-import type {CanonicalTileID, UnwrappedTileID} from '../../source/tile_id';
-import type {ProjectionData} from '../../render/program/projection_program';
-import type {PreparedShader} from '../../shaders/shaders';
-import type {Context} from '../../gl/context';
-import type {Mesh} from '../../render/mesh';
-import type {Program} from '../../render/program';
+import {mat4, vec3} from 'gl-matrix';
```
HarelM, line 1: This looks like a bad merge...

src/style/light.ts
```diff
@@ -18,6 +18,8 @@ import type {
     PropertyValue,
     TransitionParameters
 } from './properties';
+import {mat4, vec3} from 'gl-matrix';
+import {Transform} from '../geo/transform';
```
HarelM, line 22: This file should not have any changes...

src/style/style.ts
```diff
@@ -1287,6 +1300,7 @@ export class Style extends Evented {
             name: myStyleSheet.name,
             metadata: myStyleSheet.metadata,
             light: myStyleSheet.light,
+            sky: myStyleSheet.sky,
```
HarelM, line 1303: I don't think this will serialized currently as mystylesheet is not updated as part of set sky.

src/ui/map.ts
```diff
@@ -2664,6 +2665,31 @@ export class Map extends Camera {
...
     }
 
+    /**
+     * Sets the value of style's sky properties.
+     *
+     * @param sky - Sky properties to set. Must conform to the [MapLibre Style Specification](https://maplibre.org/maplibre-style-spec/sky).
+     *
+     * @example
+     * ```ts
+     *
```
HarelM, line 2675: Please add an example

src/geo/projection/projection.ts
Pheonor: Ok, I fix the import section.

src/style/light.ts
Pheonor: Ok, I revert the file.

src/style/style.ts
Pheonor: Ok, I update the setSky code, fix the stylesheet update and add a dedicated unit test.

src/ui/map.ts
Pheonor: Done

Pheonor:
@HarelM I did not understand why a test is in time out. Could you help me please ?

kubapelc:
If it is a render test, then a timeout is usually an error in the json, like a missing comma, wrong property somewhere, etc... No idea how to debug them efficiently,I usually just remove parts of the json until it starts working to find where the error roughly is :D

HarelM:
Render test are ran in the browser, you can open the browser and see what went wrong usually by changing the headless to false.

src/style/style.ts
```diff
@@ -629,6 +640,7 @@ export class Style extends Evented {
         }
 
         this.light.recalculate(parameters);
+        if (this.sky) this.sky.recalculate(parameters);
```
HarelM, line 643: It seems that sky is always set, is this if condition needed?

HarelM:
There are still two open comments above which I don't know if they are resolved or not.
@kubapelc can you check them as you have started the conversation around them?

Other than that, I believe this is in a good shape.
@Pheonor what about incorporating the sky PR into this PR? have you thought about it? Do you want me to try and push the sky PR into main and then try and merge this into the globe branch? Let me know how you want to tackle this.
- https://github.com/maplibre/maplibre-gl-js/pull/3645

kubapelc:
I might have found a bug, there is a weird black edge around the atmosphere. It is visible when enabling globe on a maptiler map (with no other style changes). This might just be weird default values in the style, or maybe a bug with alpha blending?

Globe with atmosphere on a white background with a weird black edge

Also atmosphere doesn't respect large fill extrusions. But honestly I don't think this needs addressing, at least not right now. I doubt that it would be a frequent use-case and fixing it would probably be hard or cost a lot of performance.

Fill extrusion with atmosphere

kubapelc:
I'm not sure what two comments @HarelM means, but what I would like to see addressed are default style values and performance. I'll post about performance shortly, but I think atmosphere can be enabled by default from a perfromance point of view, as current state is much better!

# Style defaults

As seen in the screenshots in the previous post, current default atmosphere seems to obscure the map by darkening it. I think having a default that obscures vision is not good. I also think that changing the default sun position to make the visible planet be lit would not help, since the map would just be tinted blue instead of gray.

I suggest adding a new feature to atmosphere, some kind of blending approach, where the globe center would have atmosphere blended out entirely, thus preserving visibility, but atmosphere would still be present on the horizons, where most of the pretty colors happen.

Alternatively the default atmosphere can just be made more subtle in general, which would also solve this problem and probably be much easier.

HarelM:
The two open review comments:
https://github.com/maplibre/maplibre-gl-js/pull/4020#discussion_r1573919136
https://github.com/maplibre/maplibre-gl-js/pull/4020#discussion_r1573915231

I don't mind changing the default, making atmosphere on is probably mainly for raster to mimic reality better, which is not the main usage for maplibre I believe, I might be wrong, I don't really have the relevant statistics.

There is a way to change the blending as you zoom in that can be defined in the style using expressions, it's just not the default.

kubapelc:
I think you can mark both of the linked comments as resolved.

HarelM:
Done, thanks.

src/style/style.ts
Pheonor: I chery pick this change from the Sky PR.
The sky is not always set, at least during loading.

Pheonor:
> There are still two open comments above which I don't know if they are resolved or not. @kubapelc can you check them as you have started the conversation around them?
> 
> Other than that, I believe this is in a good shape. @Pheonor what about incorporating the sky PR into this PR? have you thought about it? Do you want me to try and push the sky PR into main and then try and merge this into the globe branch? Let me know how you want to tackle this.
> 
>     * [Basic Sky Implementation working with actual codebase #3645](https://github.com/maplibre/maplibre-gl-js/pull/3645)

I could try to merge the Sky PR. I hope it will be Ok.

Pheonor:
> I might have found a bug, there is a weird black edge around the atmosphere. It is visible when enabling globe on a maptiler map (with no other style changes). This might just be weird default values in the style, or maybe a bug with alpha blending?
> 
> Globe with atmosphere on a white background with a weird black edge
> 
> Also atmosphere doesn't respect large fill extrusions. But honestly I don't think this needs addressing, at least not right now. I doubt that it would be a frequent use-case and fixing it would probably be hard or cost a lot of performance.
> 
> Fill extrusion with atmosphere

Effectively, the render is strange.
The atmosphere expects a black background to be real with real satellite raster data.
Perhaps a fake glow effect could be used for other aproach.
This could be the subject of another ticket :)

HarelM:
Maybe we should do the sky in a different PR.
I'll do a final review in the next few days. 
@kubapelc let me know if you think there are other stuff that should be fixed before we merge this. 
The defaults are defined in the style spec package and should be changed in a different PR anyway...

src/geo/projection/globe.ts
```diff
@@ -390,8 +424,8 @@ export class GlobeProjection implements Projection {
         return dotResult < 0.0;
     }
 
-    public transformLightDirection(transform: { center: LngLat }, dir: vec3): vec3 {
-        const sphereX = transform.center.lng * Math.PI / 180.0;
+    public transformLightDirection(transform: { center: LngLat; pitch: number; angle: number }, dir: vec3): vec3 {
```
HarelM, line 427: Where are `pitch` and `angle` used in this method?

src/geo/projection/mercator.ts
```diff
@@ -158,7 +184,7 @@ export class MercatorProjection implements Projection {
         return this._cachedMesh;
     }
 
-    public transformLightDirection(_transform: { center: LngLat }, dir: vec3): vec3 {
+    public transformLightDirection(_: any, dir: vec3): vec3 {
```
HarelM, line 187: Better to keep the original interface definition here.

src/render/draw_atmosphere.ts
```diff
@@ -0,0 +1,80 @@
...
+    const depthMode = new DepthMode(gl.LEQUAL, DepthMode.ReadOnly, [0, 1]);
+
+    const projection = painter.style.projection;
+    const projectionData = projection.getProjectionData(null, null);
+
+    const sunPos = getSunPos(light, painter.transform);
+
+    const atmosphereBlend = sky.getAtmosphereBlend();
+    if (atmosphereBlend === 0) {
+        // Dont draw anythink if atmosphere is fully transparent
```
HarelM, line 48:
```suggestion
        // Don't draw anything if atmosphere is fully transparent
```

src/style/style.ts
```diff
@@ -586,6 +595,8 @@ export class Style extends Evented {
 
             this.light.updateTransitions(parameters);
 
+            if (this.sky) this.sky.updateTransitions(parameters);
```
HarelM, line 598: Same here, I'm not sure this if is needed, sky and light object are created in the same place basically...

HarelM:
I've added some last nit picking... overall I believe this is ready to be merged.

src/geo/projection/globe.ts
Pheonor: Removed. I used in a previous version of the PR ;)

src/geo/projection/mercator.ts
Pheonor: Done

src/render/draw_atmosphere.ts
Pheonor: Done

src/style/style.ts
Pheonor: Done

kubapelc:
As for me, this can be merged, the black edge bug can be resolved with another PR.

I also did some more performance eyeballing, so I might as well post the results:

- 1080p(120Hz) display with Nvidia GTX 1060 MaxQ 6gb: around 120 fps - ok! (also probably CPU bottlenecked anyway)
- 4K(60Hz) display with Nvidia GTX 1060 MaxQ 6gb: solid 60 fps - ok!
- 1080p(120Hz) display with Intel HD 630: around 80 fps, ok, but has a noticeable lag for some reason. I don't think we can do anything about that though.
- 4K(60Hz) display with Intel HD 630: around 22 fps, unpleasant to look at, map interactions have very noticeable lag.
- 1080x2408(60Hz) phone with Adreno 619: around 33 fps, suboptimal but usable.

Context: I'm looking at an atmosphere'd globe so that every screen pixel is covered by it and executed the atmosphere pixel shader, so that I get the worst case results. A globe without atmosphere runs at 60 fps (120 on the faster monitor) on all these GPUs. Right now the most powerful available integrated GPUs are roughly comparable to the GTX 1060 in performance (eg. AMD Radeon 780M, and excluding Apple chips).

My takeaway: disregard my previous comment about reducing sample counts not working all that well :) At least on current hardware, current status of performance is mostly fine. Performance on a phone looks somewhat suboptimal, but also usable. But atmosphere is not usable for people who pair a big screen with an old iGPU - but I'm not sure how large that group is.

But I think it is still worth it to pursue optimization for atmosphere before globe release - both to make it more pleasant on old/slow hardware, and to improve power consumption when using the map on any battery powered device.

I also have some gamedev-inspired ideas on how to improve the performance of atmosphere. I can make a PR about this later, or post the ideas here if you want.

HarelM:
I've merged this.
I'm working on reviving the sky implementation here:
- #3645

I'm looking into updating the code about the fog, horizon and sky colors and blending.
I'm in the final stages, so I might merge this into main soon and then merge it to the globe branch.
I'm not sure how this will work together, but feel free to review what I already did...

onuryb:
Maplibre sky is not displayed everywhere is covered with maps

PR #4051, opened 2024-04-29
New map option to decide whether to cancel previous pending tiles while zooming in

DanielForniessoria-TomTom:
This PR creates an opt-out behaviour for the changes which were introduced here since version 3.0:
https://github.com/maplibre/maplibre-gl-js/pull/2377

Now there is a new map option (which can also be changed during runtime) that determines whether pending (still loading) tiles from previous (farther) zoom levels are going to be canceled as we zoom in.
The change is backwards compatible, since the default behavior is just as before.
However if we disable the pending tile cancelation, we get the original (pre v3) behavior, which helps to progressively unveil map details when zooming into areas where tiles are not yet cached. For modern devices this could be the desired behavior.

Behavior with tile canceling (now the default, and active since v3):
https://github.com/maplibre/maplibre-gl-js/assets/73886099/d5ce23c4-ad5d-4548-96d0-8bb4f8ff49ae

Behavior without tile canceling (was default before v3):
https://github.com/maplibre/maplibre-gl-js/assets/73886099/7f42dc32-f526-4da5-bd86-03915d4e531d

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 - Update Popup methods `addClass` & `removeClass` to return instance of Popup ([#3975](https://github.com/maplibre/maplibre-gl-js/pull/3975))
 
 ### ✨ Features and improvements
+- _...Add new stuff here..._
+- New map option to decide whether to cancel previous pending tiles while zooming in ([#4051](https://github.com/maplibre/maplibre-gl-js/pull/4051))
 - Sprites include optional textFitHeight and textFitWidth values ([#4019](https://github.com/maplibre/maplibre-gl-js/pull/4019))
 
 ### 🐞 Bug fixes
--- a/src/source/source_cache.test.ts
+++ b/src/source/source_cache.test.ts
@@ -1,4 +1,5 @@
 import {SourceCache} from './source_cache';
+import {Map} from '../ui/map';
 import {Source, addSourceType} from './source';
 import {Tile} from './tile';
 import {OverscaledTileID} from './tile_id';
@@ -1184,8 +1185,40 @@ describe('SourceCache#_updateRetainedTiles', () => {
 
     });
 
-    test('Only retain loaded parent tile when zooming in', () => {
+    test('Retain, then cancel loading tiles when zooming in', () => {
         const sourceCache = createSourceCache();
+        // Disabling pending tile canceling (thus retaining) in Map mock:
+        const map = {cancelPendingTileRequestsWhileZooming: false} as Map;
+        sourceCache.onAdd(map);
+        sourceCache._source.loadTile = async (tile) => {
+            tile.state = 'loading';
+        };
+
+        let idealTiles = [new OverscaledTileID(9, 0, 9, 0, 0), new OverscaledTileID(9, 0, 9, 1, 0)];
+        sourceCache._updateRetainedTiles(idealTiles, 9);
+        idealTiles = [new OverscaledTileID(10, 0, 10, 0, 0), new OverscaledTileID(10, 0, 10, 1, 0)];
+        let retained = sourceCache._updateRetainedTiles(idealTiles, 10);
+        expect(Object.keys(retained).sort()).toEqual([
+            new OverscaledTileID(9, 0, 9, 0, 0).key,    // retained
+            new OverscaledTileID(10, 0, 10, 0, 0).key,
+            new OverscaledTileID(10, 0, 10, 1, 0).key
+        ]);
+
+        // Canceling pending tiles now via runtime map property:
+        map.cancelPendingTileRequestsWhileZooming = true;
+        retained = sourceCache._updateRetainedTiles(idealTiles, 10);
+        // Parent loading tiles from z=9 not retained:
+        expect(Object.keys(retained).sort()).toEqual([
+            new OverscaledTileID(10, 0, 10, 0, 0).key,
+            new OverscaledTileID(10, 0, 10, 1, 0).key
+        ]);
+    });
+
+    test('Cancel, then retain, then cancel loading tiles when zooming in', () => {
+        const sourceCache = createSourceCache();
+        // Applying tile canceling default behavior in Map mock:
+        const map = {cancelPendingTileRequestsWhileZooming: true} as Map;
+        sourceCache.onAdd(map);
         sourceCache._source.loadTile = async (tile) => {
             tile.state = 'loading';
         };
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 - Update Popup methods `addClass` & `removeClass` to return instance of Popup ([#3975](https://github.com/maplibre/maplibre-gl-js/pull/3975))
 
 ### ✨ Features and improvements
+- New map option to decide whether to cancel previous pending tiles while zooming in.
```
HarelM, line 5: Can you please add a link to this PR?

src/ui/map.ts
```diff
@@ -318,6 +318,13 @@ export type MapOptions = {
      * You shouldn't set this above WebGl `MAX_TEXTURE_SIZE`. Defaults to [4096, 4096].
      */
     maxCanvasSize?: [number, number];
+    /**
+     * Determines whether to cancel, or retain, tiles from the current viewport which are still loading but which belong to a farther (smaller) zoom level than the current one.
+     * * If true, when zooming in, the previous zoom level(s) tiles will progressively appear, giving a smoother map details experience. However, more tiles will be rendered in a short period of time.
+     * * If false, when zooming in, tiles which didn't manage to load for previous zoom levels will become cancelled. This might save some computing resources for slower devices, but the map details might appear more abruptly at the end of the zoom.
+     * @defaultValue true
```
HarelM, line 325: Isn't the current default false?

HarelM: Oh, the description for true/false should be switched, that what got me confused...

src/source/source_cache.ts
```diff
@@ -734,7 +734,7 @@ export class SourceCache extends Evented {
                 }
                 if (tile) {
                     const hasData = tile.hasData();
-                    if (parentWasRequested || hasData) {
+                    if (hasData || !this.map?.cancelPendingTileRequestsWhileZooming || parentWasRequested) {
```
HarelM, line 737: Is this the only place? Seems strange... I would expect to see places that talk about the cancellation of tiles.
So if the tile is "retained" there won't be cancellation of the request?

src/source/source_cache.test.ts
```diff
@@ -1204,6 +1208,21 @@ describe('SourceCache#_updateRetainedTiles', () => {
             idealTiles.map((tile) => tile.key).sort()
         );
 
+        // Stopping tile canceling via runtime map property:
```
HarelM, line 1211: Consider splitting this into 2-3 tests. You can use beforeEach to setup the initial state.

src/ui/map.ts
DanielForniessoria-TomTom: Oh you're right! Will fix it

src/source/source_cache.ts
DanielForniessoria-TomTom: I naively focused on where the previous change was done.
Retaining should imply non-cancelling indeed.

CHANGELOG.md
DanielForniessoria-TomTom: Done

src/ui/map.ts
DanielForniessoria-TomTom: Fixed

src/source/source_cache.test.ts
DanielForniessoria-TomTom: Added one more test which starts with the config in the opposite way

src/source/source_cache.test.ts
```diff
@@ -1184,8 +1185,40 @@ describe('SourceCache#_updateRetainedTiles', () => {
...
+        map.cancelPendingTileRequestsWhileZooming = true;
+        retained = sourceCache._updateRetainedTiles(idealTiles, 10);
+        // Parent loading tiles from z=9 not retained:
+        expect(Object.keys(retained).sort()).toEqual([
+            new OverscaledTileID(10, 0, 10, 0, 0).key,
+            new OverscaledTileID(10, 0, 10, 1, 0).key
+        ]);
+    });
+
+    test('Cancel/retain loading tiles when zooming in', () => {
```
HarelM, line 1217: This is a strange name for a test, is it canceling or retaining?

src/source/source_cache.test.ts
```diff
@@ -1184,8 +1185,40 @@ describe('SourceCache#_updateRetainedTiles', () => {
 
     });
 
-    test('Only retain loaded parent tile when zooming in', () => {
+    test('Retain, then cancel loading tiles when zooming in', () => {
         const sourceCache = createSourceCache();
+        // Disabling pending tile canceling (thus retaining) in Map mock:
+        const map = {cancelPendingTileRequestsWhileZooming: false} as Map;
```
HarelM, line 1191: I think you can do with by calling `sourceCache.map.cancelPendingTileRequestsWhileZooming = false`, can't you? It's done in other tests are removes the need to import map in these tests, which I find better.

src/source/source_cache.test.ts
```diff
@@ -1184,8 +1185,40 @@ describe('SourceCache#_updateRetainedTiles', () => {
...
+        map.cancelPendingTileRequestsWhileZooming = true;
+        retained = sourceCache._updateRetainedTiles(idealTiles, 10);
+        // Parent loading tiles from z=9 not retained:
+        expect(Object.keys(retained).sort()).toEqual([
+            new OverscaledTileID(10, 0, 10, 0, 0).key,
+            new OverscaledTileID(10, 0, 10, 1, 0).key
+        ]);
+    });
+
+    test('Cancel or retain loading tiles when zooming in depending on current configuration', () => {
```
HarelM, line 1217:
```suggestion
    test('Cancel, then retain loading tiles when zooming in', () => {
```
So that it will be aligned with the previous test...?

HarelM (approved):

src/source/source_cache.test.ts
DanielForniessoria-TomTom: Well it'd be more like "Cancel, then retain, then cancel again..."  :) Maybe too much

src/source/source_cache.test.ts
DanielForniessoria-TomTom: I can't do that before sourceCache.map is defined. That's why the first step is to define it, and set it to the source (I use onAdd to try to reuse regular mechanism, but could also set it directly, not sure).

src/source/source_cache.test.ts
HarelM: Updated my suggestion - this only tells us that the tests is testing "too much" :-)

src/source/source_cache.test.ts
DanielForniessoria-TomTom: I see many "onAdd" calls in that test file too, I guess either way is reasonable.

HarelM:
Thanks!!

PR #4015, opened 2024-04-18
Globe - circle and heatmap layers

kubapelc:
# Circle and heatmap layers adaptation for globe

Issue & discussion of globe:

https://github.com/maplibre/maplibre/issues/190
https://github.com/maplibre/maplibre/discussions/161

Additions and changes since [the previous PR](https://github.com/maplibre/maplibre-gl-js/pull/3968):

- Adapted circle layer type for globe
- Adapted heatmap layer type for globe

The size of (`circle-pitch-alignment: 'map'`) circles and heatmap kernels of a certain radius is the same across all latitudes under globe projection, same as with mercator circles. Additionally, the size of circles/kernels is consistent when changing projection from globe to mercator or back. Everything should visually behave the way the user intuitively expects it to.

I've updated the public globe demo, select the "Heatmap" or "Circles" map type to experiment with this behavior. Note that the demo is a build from my main branch, which additionally has symbol adaptation.

https://kvaleya.gitlab.io/maplibre/globe/globedemo.html

The demo also doesn't auto-enlarge the globe when map center is near the poles, as it used to, now globe size is consistent across map positions. This is just an early hack, and moving the map center near the poles is bugged and will result in the map jumping around, but having consistent globe size is needed to properly demonstrate circle behavior.

[Main vector globe branch with all changes](https://github.com/windycom/maplibre-gl-js/tree/kubapelc/globe-vector)

Screenshot of circles on a globe

Screenshot of a heatmap on a globe

## Launch Checklist

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.
 - [x] Link to related issues.
 - [x] Include before/after visuals or gifs if this PR includes visual changes.
 - [ ] Write tests for all new functionality.
 - [ ] Document any changes to public APIs.
 - [ ] Post benchmark scores.
 - [ ] Add an entry to `CHANGELOG.md` under the `## main` section.

Issue #3456, opened 2023-12-06
Weird interaction between text-rotation-alignment and text-translate-anchor


kubapelc:
Hi,

while studying the behaviour of `text-translate`, `text-translate-anchor` and `text-rotation-alignment` I noticed some weird interactions between those style tags. Below is a summary of their interactions. Term "map space" refers to text pitch aligned with map (flat text as if printed onto the map), "screen space" is text with pitch aligned with viewport, so always flat on the screen. The described behaviour applies to both `symbol-placement` set to `point` and `line`.

| text-rotation-alignment | translate-anchor | current behaviour                                                     | note       | suggested behaviour                             |
| ----------------------- | ---------------- | --------------------------------------------------------------------- | ---------- | ----------------------------------------------- |
| viewport                | map              | viewport-rotated translation done in map space, screen space text     | reasonable | unmodified |
| viewport                | viewport         | viewport-aligned screen space translation, screen space text           | reasonable | unmodified |
| map                     | map              | map space translation rotated in *opposite* direction relative to viewport rotation, text in map space | broken?    | correct map-aligned rotation, text in map space |
| map                     | viewport         | map-aligned rotation of text translation, text translated in map space, map space text                  | weird?     | text translated in map space in direction that is *viewport*-rotated, text rendered in map space |

See these demos below, try to rotate & pitch the camera. The first one is rotating in the wrong direction, the second one should rotate, but doesn't (and also it exhibits some weird stretching).

~~https://kvaleya.gitlab.io/maplibre/aligndemo/rotation-align-map-translate-anchor-map.html~~

~~https://kvaleya.gitlab.io/maplibre/aligndemo/rotation-align-map-translate-anchor-viewport.html~~

Better demo: https://kvaleya.gitlab.io/maplibre/aligndemo/combineddemo.html

I'm currently modifying the symbol rendering code for globe and I am kinda familiar with it, so I can probably make a PR with a fix fairly quickly, but I'm not sure what the intended behaviour is.

Is this unintended behaviour? Should I make a PR with these suggested changes?

kubapelc:
I've played around with this for a bit and made a better demo:

https://kvaleya.gitlab.io/maplibre/aligndemo/combineddemo.html

Now all 4 variants seem inconsistent with how `-translate` and `-translate-anchor` works for any other layer.

HarelM:
I think the red and black dot in the demo are doing a fine job at explaining the problem.
I tend to agree about your conclusions.
[...]

Issue #210, opened 2021-07-14
Using text-translate in symbol paint doesn't update collision boxes


timsluis:
When adding a `text-translate` property to a layer the collision boxes won't move together with the labels. Also click and hovers are registered on the original place of the label before the translate.

<img width="216" alt="Screenshot 2021-07-14 at 11 03 00" src="https://user-images.githubusercontent.com/235880/125595298-d255d130-5dc2-4224-9ec1-c4f14223c62d.png">

**maplibre-gl-js version**: 1.15.2 

**browser**: Chrome 91

### Steps to Trigger Behavior

 1. Add a symbol layer
 2. Provide a `text-translate` on the paint property of a layer

### Link to Demonstration

See: https://codepen.io/timsluis/pen/qBmRYQM

### Expected Behavior

Collision boxes are updated.

### Actual Behavior

Collision boxes are not updated and labels flow over each other.

wipfli:
Wasn't there something related in the terrain3d pull request? @timsluis would you be interested in fixing this?

mrflix:
The bug is still very much there: I'm using both `text-translate` and `icon-translate` to offset symbols so that the pin points to the exact location. I would like to use `icon-offset` and `text-offset` instead, but I can't because I use `text-variable-anchor` and `text-radial-offset` for dynamic label placement and that overwrites `text-offset`.

This is the result of using `text-translate` and `icon-translate`:
![Screenshot 2023-12-21 at 13 47 58](https://github.com/maplibre/maplibre-gl-js/assets/60777/8be799ce-a385-4143-a5ed-5c3c12a444c0)
Outlined in 🟦 is the original position, 🟥 the translated position. The mouseover and click area is 🟦.
So only the bottom right of the visual area can be clicked 😟

HarelM:
Reopening.
I think it's somewhat related to:
- https://github.com/maplibre/maplibre-gl-js/issues/3456

The collision box logic is not great in general...

mrflix:
@kubapelc thanks for looking into this and sharing your work. I tested it with http://kvaleya.gitlab.io/maplibre/aligndemo2/maplibre-gl-dev.js. Unfortunately the labels all move into the top left corner of the canvas:

![Screenshot 2024-03-12 at 11 20 24](https://github.com/maplibre/maplibre-gl-js/assets/60777/97416509-beb7-4d3a-b40d-a05f1817e23d)

[...]

PR #4067, opened 2024-05-02
Globe - symbols & symbol bugfixes

kubapelc:
# Symbol layer adaptation + symbol bugfixes

Issue & discussion of globe:

Additions and changes since [the previous PR](https://github.com/maplibre/maplibre-gl-js/pull/4015):

- Adapted symbol layer type for globe.
- Fixed another source of tiny seams in the fill layer when globe is enabled (the `draw_fill` and `painter` changes are related to this).
- Symbol bugfixes:
  - Fixed symbol collision debug view (`showCollisionBoxes`) not showing the actual bounding boxes used for collision and click areas. The displayed boxes now match actual collision boxes exactly. Displayed collision boxes will react to map motion with some latency. (Just as the actual collision boxes do.)
  - Fixed symbol collisions using inaccurate and sometimes entirely wrong collision boxes when the map is pitched or rotated. ([#210](https://github.com/maplibre/maplibre-gl-js/issues/210))
  - Fixed symbol collision boxes not being accurate for variable-anchor symbols.
  - Fixed icon collision boxes using `text-translate` property for translation instead of the correct `icon-translate`.
  - Fixed `text-translate` and `icon-translate` behaving weirdly and inconsistently with other `-translate` properties. ([#3456](https://github.com/maplibre/maplibre-gl-js/issues/3456))
  - Updated some old render tests to reflect new symbol & collision box behaviour, added several new render tests (in addition to render tests for symbols on a globe).

I've updated the public globe demo with MapLibre build from this PR:

https://kvaleya.gitlab.io/maplibre/globe/globedemo.html

Screenshot of a globe with symbols and debug collisions enabled.

Screenshot of a globe with symbols and debug collisions, with many test texts.

## Launch Checklist

 - [x] Link to related issues.
   - Most changes and new functionality is covered by render tests. 
 - [ ] Post benchmark scores.


kubapelc:
Globe render tests for symbols are duplicated - each test has two variants, one with collision box debug disabled, and one with enabled. In base MapLibre, collision tests are separate, so I separated them here as well. But I'm wondering if I could keep just the variant with collisions for globe render tests, since it also inherently tests regular symbol rendering.

kubapelc:
Also, in the second screenshot on the left side you can notice the pitch-alignment: map texts having their bounding box slightly off. I have no idea why that happens. I've decided to go ahead with the PR anyway, since it is a minor offset and only happens on pitch-alignment: map texts that are "planetary scale", which is probably an edge case in real maps.

HarelM:
I'll review this in the coming days, thanks for this!
I thought I already asked, but I prefer to ask again, just to be on the safe side:
Is it possible to fix the collision boxes in the main branch and not only here?
It would be beneficial to people using the production version until this branch is merged.

kubapelc:
> Is it possible to fix the collision boxes in the main branch and not only here?

I know I've previously said that the changes are too deeply rooted into globe, but after taking another look at the actual diffs of the files, I think it might not be too much work to port it over to main. The easiest approach is to create a greatly reduced version of the `Projection` interface from globe and use that in symbols. Is that an ok approach?

HarelM:
I believe so, as long as we keep the changes minimal I think it will be a good way forward.

HarelM:
Let's focus on the following PR first:
- #4071

Once merged to main and to globe branch and to globe-pr-symbol I'll review this PR.
Thanks for your patience.

HarelM:
There's a need to resolve conflict I suppose.
Let me know if there's anything I can do to help.

src/render/draw_symbol.ts
```diff
@@ -219,7 +219,7 @@ function updateVariableAnchorsForBucket(
             symbolProjection.hideGlyphs(symbol.numGlyphs, dynamicTextLayoutVertexArray);
         } else  {
             const tileAnchor = new Point(symbol.anchorX, symbol.anchorY);
-            const projectionContext = {
+            const projectionArgs = {
```
HarelM, line 222: Why have you renamed this back to args instead of context?

src/render/draw_symbol.ts
```diff
@@ -156,7 +156,7 @@ function updateVariableAnchors(coords: Array<OverscaledTileID>,
             const getElevation = painter.style.map.terrain ? (x: number, y: number) => painter.style.map.terrain.getElevation(coord, x, y) : null;
             const translation = projection.translatePosition(transform, tile, translate, translateAnchor);
             updateVariableAnchorsForBucket(bucket, rotateWithMap, pitchWithMap, variableOffsets,
-                transform, labelPlaneMatrix, coord.posMatrix, tileScale, size, updateTextFitIcon, projection, translation, coord.toUnwrapped(), getElevation);
+                transform, labelPlaneMatrix, coord.posMatrix, tileScale, size, updateTextFitIcon, painter.style.map.projection, translation, coord.toUnwrapped(), getElevation);
```
HarelM, line 159: use the `projection` variable defined above...?

src/render/draw_symbol.ts
kubapelc: I must have overlooked this when merging, I'll rename it back.

src/geo/projection/mercator.ts
```diff
@@ -134,11 +125,15 @@ export class MercatorProjection implements Projection {
         return 1.0;
     }
 
-    public translatePosition(transform: { angle: number; zoom: number }, tile: Tile, translate: [number, number], translateAnchor: 'map' | 'viewport'): [number, number] {
+    public getPitchedTextCorrection(_transformCenter: any, _textAnchor: any, _tileID: any): number {
```
HarelM, line 128: Better not to use "any" if possible.

src/geo/projection/projection.ts
```diff
@@ -186,10 +186,13 @@ export interface Projection {
...
-    // HM TODO: this needs to be fixed.
-    getPitchedTextCorrection(_transform: any, _anchor: any, _tile: any): number;
-
-    // HM TODO: fix this!
-    projectTileCoordinates(_x, _y, _t, _ele);
+    /**
+     * @internal
+     * Projects a point in tile coordinates. Used in symbol rendering.
+     */
+    projectTileCoordinates(x: number, y: number, unwrappedTileID: UnwrappedTileID, getElevation: (x: number, y: number) => number): {
```
HarelM, line 193: Please create a type for the return value and add this type to the implementations.

src/geo/projection/projection.ts
```diff
@@ -143,25 +143,25 @@ export interface Projection {
...
-    getPixelScale(transform: { center: LngLat }): number;
+    getCircleRadiusCorrection(transform: { center: LngLat }): number;
 
     /**
      * @internal
-     * Allows the projection to adjust the radius of `circle-pitch-alignment: 'map'` circles and heatmap kernels based on the transform's zoom level and latitude.
-     * Circle and kernel radius is multiplied by this value.
+     * Allows the projection to adjust the scale of `text-pitch-alignment: 'map'` symbols's collision boxes based on the map's center and the text anchor.
+     * Only affects the collision boxes (and click areas), scaling of the rendered text is mostly handled in shaders.
+     * @param transformCenter - The map's longitude and latitude.
```
HarelM, line 160:
```suggestion
     * @param transform.center - The map's longitude and latitude.
```

src/symbol/collision_index.ts
```diff
@@ -323,7 +323,51 @@ export class CollisionIndex {
...
     projectPathToScreenSpace(projectedPath: Array<Point>, projectionContext: SymbolProjectionContext, labelToScreenMatrix: mat4) {
-        return projectedPath.map(p => projection.project(p, labelToScreenMatrix, projectionContext.getElevation));
+        let screenSpacePath: Array<{
+            point: Point;
+            signedDistanceFromCamera: number;
+            isOccluded?: boolean;
+        }>;
+        if (projectionContext.projection.useSpecialProjectionForSymbols) {
+            // Globe (or other special projection) is enabled in this branch.
+            const inverseLabelPlaneMatrix = mat4.create();
```
HarelM, line 333: Consider moving the content of this if to a different method, I would even consider moving it to a different class if possible (maybe even the projection class, IDK...)

HarelM: Looking at this code more, I think this if is basically saying "globe or not globe", so maybe `projectPathToScreenSpace` should be part of the projection interface?

src/geo/projection/mercator.ts
```diff
@@ -166,13 +161,11 @@ export class MercatorProjection implements Projection {
...
     }
 
-    // HM TODO: fix this!
-    getPitchedTextCorrection(_transform: any, _anchor: any, _tile: any) {
-        return 1;
-    }
-
-    // HM TODO: fix this!
-    projectTileCoordinates(_x, _y, _t, _ele) {
+    public projectTileCoordinates(_x: number, _y: number, _unwrappedTileID: UnwrappedTileID, _getElevation: (x: number, y: number) => number): {
```
HarelM, line 164: The fact that this method throws an error of "not implemented" is a "smell".
If there's a location in the code that checks for `useSpecialProjectionForSymbols` and then does not call this method, it might be better to move some of the logic into this call, and maybe even remove this `useSpecialProjectionForSymbols` all together so that we don't get tempted to check its value and simply implement the right method in the right class.

It's a rough thought, I think the separation of these classes is very good in general and you did a great job here.

kubapelc: Yes, I've been thinking about something similar, looking at the code now, I think I really can remove `useSpecialProjectionForSymbols` altogether.

kubapelc: I've played around with removing `useSpecialProjectionForSymbols` and I think it would be best left for the transform refactor & rewrite. Since symbol projection happens over several frames, it has its own copy of the transform class which doesn't change and which effectively serves as a "mercator projection" handler. I currently have no such mechanism for the projection class, and would prefer to solve this along with decoupling the transform class from mercator projection.

src/symbol/collision_index.ts
kubapelc: I've refactored it somewhat in the latest commit. I would prefer to keep it as it is now for this PR, since merging the globe+not globe paths (or moving them elsewhere) is related to removing the `useSpecialProjectionForSymbols` flag.

src/geo/projection/projection.ts
```diff
@@ -9,6 +9,7 @@ import type {Program} from '../../render/program';
 import type {SubdivisionGranularitySetting} from '../../render/subdivision_granularity_settings';
 import Point from '@mapbox/point-geometry';
 import type {LngLat} from '../lng_lat';
+import {PointProjection} from '../../symbol/projection';
```
HarelM, line 12: I'm guessing since this is an interface class, that all the imports here should be "import type", right?

src/symbol/collision_index.ts
```diff
@@ -322,52 +322,17 @@ export class CollisionIndex {
...
-                    backProjected.point.x,
-                    backProjected.point.y,
-                    projectionContext.unwrappedTileID,
-                    projectionContext.getElevation
-                );
-                projected.point.x = (projected.point.x * 0.5 + 0.5) * projectionContext.width;
-                projected.point.y = (-projected.point.y * 0.5 + 0.5) * projectionContext.height;
-                return projected;
-            });
+            const screenSpacePath = projection.projectPathSpecialProjection(projectedPath, projectionContext);
```
HarelM, line 328: this looks as if there's a projection class that is being used, which got me confused.
Can you please remove the `import * from`... and replace it with importing the relevant methods only?
Otherwise this looks good.

src/geo/projection/mercator.ts
HarelM: OK, fair enough, but let's make sure we remember to do it later on. Or at least decide not to do it actively.

src/geo/projection/projection.ts
kubapelc: Yes, fixed.

src/symbol/collision_index.ts
kubapelc: Fixed.

src/symbol/collision_index.ts
```diff
@@ -17,7 +15,8 @@ import type {
 } from '../data/array_types.g';
 import type {OverlapMode} from '../style/style_layer/overlap_mode';
 import {UnwrappedTileID} from '../source/tile_id';
-import {SymbolProjectionContext} from '../symbol/projection';
+import {SymbolProjectionContext, pathSlicedToLongestUnoccluded, placeFirstAndLastGlyph, project, projectPathSpecialProjection} from '../symbol/projection';
+import type {PointProjection} from '../symbol/projection';
```
HarelM, line 19: You can add this to the previous line using `import { ... type PointProjection} from...`

HarelM (approved):

src/symbol/collision_index.ts
kubapelc: Thanks! I didn't know about that!

kubapelc:
I've managed to remove `useSpecialProjectionForSymbols` altogether in my dev branch, as discussed in this thread: https://github.com/maplibre/maplibre-gl-js/pull/4067#discussion_r1602805394

HarelM:
GREAT! Thanks for the update!

PR #4071, opened 2024-05-03
Symbols: fix translation, variable-anchors and collision boxes

kubapelc:
This is a port of the symbol bugfixes from: https://github.com/maplibre/maplibre-gl-js/pull/4067 The porting ended up being relatively straightforward :)

Fixes https://github.com/maplibre/maplibre-gl-js/issues/210 (as far as I can tell)

Fixes https://github.com/maplibre/maplibre-gl-js/issues/3456

## Changes (same as the bugfixes in the globe PR)

- Fixed symbol collision debug view (`showCollisionBoxes`) not showing the actual bounding boxes used for collision and click areas. The displayed boxes now match actual collision boxes exactly. Displayed collision boxes will react to map motion with some latency. (Just as the actual collision boxes do.)
- Fixed symbol collisions using inaccurate and sometimes entirely wrong collision boxes when the map is pitched or rotated. ([#210](https://github.com/maplibre/maplibre-gl-js/issues/210))
- Fixed symbol collision boxes not being accurate for variable-anchor symbols.
- Fixed icon collision boxes using `text-translate` property for translation instead of the correct `icon-translate`.
- Fixed `text-translate` and `icon-translate` behaving weirdly and inconsistently with other `-translate` properties. ([#3456](https://github.com/maplibre/maplibre-gl-js/issues/3456))
- Updated some old render tests to reflect new symbol & collision box behaviour, added several new render tests

A screenshot of many, many collision boxes.

## Launch Checklist

 - [x] Link to related issues.
 - [x] Post benchmark scores.


kubapelc:
A render test `tests\icon-text-fit\textFit-grid-long` is failing on Windows, but it seems to a an issue in the unmodified main as well. A build test seems to fail, I'm not sure why.

I've also noticed that `pitch-alignment: map` texts also have their collision boxes slightly off, even in mercator projection. Since it is not a globe-only edge case after all, I would like to try to debug it and fix it, please wait with merging until that gets fixed.

src/symbol/collision_index.ts
```diff
@@ -277,6 +319,54 @@ export class CollisionIndex {
...
     }
 
+    projectPathToScreenSpace(projectedPath: Array<Point>, projectionArgs: ProjectionArgs, labelToScreenMatrix: mat4) {
+        let screenSpacePath: Array<{
+            point: Point;
+            signedDistanceFromCamera: number;
+            isOccluded?: boolean;
+        }>;
+        if (projectionArgs.projection.useSpecialProjectionForSymbols) {
+            // Globe (or other special projection) is enabled in this branch.
```
HarelM, line 329: Is it?

src/symbol/collision_index.ts
```diff
@@ -358,28 +448,37 @@ export class CollisionIndex {
...
+            projected = projection.project(new Point(x, y), posMatrix, getElevation);
         }
-        const a = new Point(
-            (((p[0] / p[3] + 1) / 2) * this.transform.width) + viewportPadding,
-            (((-p[1] / p[3] + 1) / 2) * this.transform.height) + viewportPadding
-        );
         return {
-            point: a,
+            point: new Point(
+                (((projected.point.x + 1) / 2) * this.transform.width) + viewportPadding,
```
HarelM, line 460: You said something about padding, I'm just throwing it here in case it is relevant, if not, ignore this comment.

src/symbol/collision_index.ts
```diff
@@ -398,4 +497,141 @@ export class CollisionIndex {
...
+        }
+
+        return {
+            box: getAABB(points),
+            allPointsOccluded: !anyPointVisible
+        };
+    }
+}
+
+function getAABB(points: Array<Point>): [number, number, number, number] {
```
HarelM, line 623: I just wrote the same method to get the bounding box of points in maplibre-style-spec, let me know if you want to me to export it, otherwise, move this one to some geometry utils class.

src/symbol/placement.ts
```diff
@@ -615,6 +650,25 @@ export class Placement {
                             }
                         }
 
+                        if (placedBox.box.length === 0) {
```
HarelM, line 653: `!placedBox.placable`?

src/symbol/placement.ts
```diff
@@ -615,6 +650,25 @@ export class Placement {
...
+                            // No box was successfully placed
+                            // Generate bounds for a fake centered box, so that we can at least display something for collision debug.
```
HarelM, line 655: This code runs only when debug is on?

src/symbol/projection.ts
```diff
@@ -572,6 +708,16 @@ function placeGlyphAlongLine(
         lineStartIndex + anchorSegment :
         lineStartIndex + anchorSegment + 1;
 
+    // Project anchor point to proper label plane and cache it
+    let anchorPoint: Point;
+
+    if (projectionArgs.projectionCache.cachedAnchorPoint) {
```
HarelM, line 714: Are you sure about keeping the cache as part of the projection arguments? I think it might be better to pass it along the project args? or give projectionArg a better name.

src/geo/projection/projection.ts
```diff
@@ -0,0 +1,76 @@
+import type Point from '@mapbox/point-geometry';
+import type {Tile} from '../../source/tile';
+import {pixelsToTileUnits} from '../../source/pixels_to_tile_units';
+
+/**
+ * A greatly reduced version of the `Projection` interface from the globe branch,
```
HarelM, line 6: Will this create issues if I merge this to main and then merge main to globe? I would like it to be as painless as possible, I know it's a big ask..

HarelM:
This looks amazing! Thanks! I still need to wrap my head around what you did, so any illustration of how the code used to work vs now would be helpful for me to understand.
Consider using mermain for that as github supports is natively now.

src/symbol/collision_index.ts
kubapelc: It was referring to the `if` condition branch :D but the comment seems pointless, I'll remove it.

src/symbol/placement.ts
kubapelc: The condition tests something a bit different, whether any `attemptAnchorPlacement` returned a valid box. I've rewritten it so that `placedBox` is null if that doesn't happen, which should make the condition a bit clearer.

src/symbol/placement.ts
kubapelc: Yes, it should only be relevant with `showCollisionBoxes`, I've added it to the condition.

src/symbol/projection.ts
kubapelc: `projectionArgs` is grouping together parameters that were passed to nearly all the projection functions before, I think projection cache isn't special in that regard. The cache is created and used everywhere where ProjectionArgs are created, so I guess it makes sense for the cache to be included.

I'm not sure what would be a better name. `SymbolProjectionContext`?

src/symbol/collision_index.ts
kubapelc: This should work ok, the projection is never applied to a box multiple times. The bug I mentioned in the first comment is still present even if I set `viewportPadding` to 0.

kubapelc:
At this point I'm not sure myself about all how it works now, this is an accumulation of small changes and tweaks since about January. Most of the bugged behaviour was about some part of symbols not respecting translation or pitch alignment or rotation alignment properly, I think.

src/symbol/projection.ts
HarelM: Yes, context would be a better name for something that passes through an algorithm/flow.

src/geo/projection/projection.ts
HarelM: Yes, I'm concerned about how I would mess up things when I merge it to main and then the globe branch.
I hope I won't break things up.
In any case, I'll focus on this before I merge the symbols to the globe branch, so you'll be able to handle the merge of the symbols code.

HarelM:
@ChrisLoer any chance you might be able to look at this PR? It would help a lot I believe.
I know you introduced some changes which touch this very code in a previous PR: 
- #2260

So your help would be greatly appreciated here.

src/render/draw_collision_debug.ts
```diff
@@ -38,10 +37,7 @@ export function drawCollisionDebug(painter: Painter, sourceCache: SourceCache, l
         const tile = sourceCache.getTile(coord);
         const bucket: SymbolBucket = (tile.getBucket(layer) as any);
         if (!bucket) continue;
-        let posMatrix = coord.posMatrix;
-        if (translate[0] !== 0 || translate[1] !== 0) {
-            posMatrix = painter.translatePosMatrix(coord.posMatrix, tile, translate, translateAnchor);
-        }
+        const posMatrix = coord.posMatrix;
```
HarelM, line 40: Are `translate` and `translateAnchor` still being used?

src/symbol/projection.ts
```diff
@@ -440,26 +546,53 @@ type ProjectionArgs = {
...
+    point: Point;
+    signedDistanceFromCamera: number;
+    isOccluded: boolean;
+} {
+    const translatedX = x + projectionArgs.translation[0];
+    const translatedY = y + projectionArgs.translation[1];
+    let projection;
+    if (!projectionArgs.pitchWithMap && projectionArgs.projection.useSpecialProjectionForSymbols) {
+        projection = projectionArgs.projection.projectTileCoordinates(translatedX, translatedY, projectionArgs.unwrappedTileID, projectionArgs.getElevation);
+        projection.point.x = (projection.point.x * 0.5 + 0.5) * projectionArgs.width;
```
HarelM, line 589: I'm definitely sure I don't fully understand the code, but why do we add more logic outside the `projectTileCoordinates`? Is it possible to somehow do it inside the above method?

src/symbol/projection.ts
```diff
@@ -360,20 +397,64 @@ function placeGlyphsAlongLine(symbol, fontSize, flip, keepUpright, posMatrix, la
...
+/**
+ * Do not use directly, use {@link projectTruncatedLineSegmentToLabelPlane} or {@link projectTruncatedLineSegmentToViewport} instead,
+ * depending on the target space.
+ *
+ * Projects a "virtual" vertex along a line segment.
+ * If `projectionMatrix` is not undefined, does a simple projection using this matrix.
+ * Otherwise, either projects to label plane using the `labelPlaneMatrix`
+ * or projects to viewport using the special map projection (mostly for globe) by calling {@link projectTileCoordinatesToViewport}.
+ */
+function projectTruncatedLineSegment(previousTilePoint: Point, currentTilePoint: Point, previousProjectedPoint: Point, minimumLength: number, projectionMatrix: mat4 | undefined, projectionArgs: ProjectionArgs) {
```
HarelM, line 438: consider adding a "_" prefix to function name to better indicate is it "very" internal?

src/symbol/placement.ts
```diff
@@ -771,9 +840,51 @@ export class Placement {
                         textFeatureIndex,
                         collisionGroup.ID);
                 }
+            }
+
+            if (showCollisionBoxes) {
+                const id = bucket.bucketInstanceId;
```
HarelM, line 846: Consider moving all this code to a method, from what I can see it looks like some caching...

HarelM: Also looking at the code that uses the member in this "if", it is possible that this is not caching the results in case the `showCollisionBoxes` is not set to true?

src/symbol/placement.ts
```diff
@@ -1109,24 +1229,40 @@ export class Placement {
                             }
                         }
 
-                        if (collisionArrays.textBox) {
-                            updateCollisionVertices(bucket.textCollisionBox.collisionVertexArray, opacityState.text.placed, !used || horizontalHidden, shift.x, shift.y);
-                        }
-                        if (collisionArrays.verticalTextBox) {
-                            updateCollisionVertices(bucket.textCollisionBox.collisionVertexArray, opacityState.text.placed, !used || verticalHidden, shift.x, shift.y);
+                        if (collisionArrays.textBox || collisionArrays.verticalTextBox) {
+                            let anchorTileX, anchorTileY;
```
HarelM, line 1233: What type are those variables?

src/symbol/placement.ts
```diff
@@ -1109,24 +1229,40 @@ export class Placement {
...
-
-                    if (collisionArrays.iconBox) {
-                        updateCollisionVertices(bucket.iconCollisionBox.collisionVertexArray, opacityState.icon.placed, verticalIconUsed,
-                            hasIconTextFit ? shift.x : 0,
-                            hasIconTextFit ? shift.y : 0);
-                    }
-
-                    if (collisionArrays.verticalIconBox) {
-                        updateCollisionVertices(bucket.iconCollisionBox.collisionVertexArray, opacityState.icon.placed, !verticalIconUsed,
+                    if (collisionArrays.iconBox || collisionArrays.verticalIconBox) {
```
HarelM, line 1250: This code looks very similar to the above code with the text vs icon, is it possible to reduce code replication some how?

src/symbol/collision_index.ts
```diff
@@ -398,4 +497,125 @@ export class CollisionIndex {
...
+            ).point;
+            const toEast = projectedEast.sub(projectedPoint.point).unit();
+            const angle = Math.atan(toEast.y / toEast.x) + (toEast.x < 0 ? Math.PI : 0);
+            const sin = Math.sin(angle);
+            const cos = Math.cos(angle);
+            vecEast = new Point(cos, sin);
+            vecSouth = new Point(-sin, cos);
+        }
+
+        if (!rotateWithMap && pitchWithMap) {
```
HarelM, line 545: `else if`?

src/symbol/collision_index.ts
```diff
@@ -398,4 +497,125 @@ export class CollisionIndex {
...
+        const offsetYmax = collisionBox.y2 * distanceMultiplier;
+        const offsetYhalf = (offsetYmin + offsetYmax) / 2;
+
+        // 0--1--2
+        // |     |
+        // 7     3
+        // |     |
+        // 6--5--4
+        const points = [
+            basePoint.add(vecEast.mult(offsetXmin)).add(vecSouth.mult(offsetYmin)),
```
HarelM, line 584: use the `_add`, `_mult` etc to avoid cloning the points while doing the calculations.

src/symbol/collision_index.ts
```diff
@@ -398,4 +497,125 @@ export class CollisionIndex {
...
+            basePoint.add(vecEast.mult(offsetXmin)).add(vecSouth.mult(offsetYhalf)),
+        ];
+
+        // Is any point of the collision shape visible on the globe (on beyond horizon)?
+        let anyPointVisible = false;
+
+        if (pitchWithMap) {
+            for (let i = 0; i < points.length; i++) {
+                const oldPoint = points[i];
+                const newPoint = this.projectAndGetPerspectiveRatio(
```
HarelM, line 600: use `points.map(...)` instead of this for loop as you are "converting objects" and `map` is known to do that, readability wise.

src/symbol/collision_index.ts
```diff
@@ -277,6 +320,53 @@ export class CollisionIndex {
         };
     }
 
+    projectPathToScreenSpace(projectedPath: Array<Point>, projectionArgs: ProjectionArgs, labelToScreenMatrix: mat4) {
```
HarelM, line 323: Is this method really needed as part of this PR? is `useSpecialProjectionForSymbols` true for the main branch?

src/shaders/symbol_icon.vertex.glsl
```diff
@@ -21,10 +19,18 @@ uniform mat4 u_coord_matrix;
...
 uniform vec2 u_texsize;
+uniform bool u_is_along_line;
+uniform bool u_is_variable_anchor;
+uniform vec2 u_translation;
+uniform float u_pitched_scale;
 
 out vec2 v_tex;
 out float v_fade_opacity;
 
+vec4 projectTileWithElevation(vec2 posInTile, float elevation) {
```
HarelM, line 30: This is for easier adoption of the tile projection in the globe branch, right?
(it's not really doing something different in this code)?

ChrisLoer:
> @ChrisLoer any chance you might be able to look at this PR? It would help a lot I believe.

I just spent about 20 minutes going through it, which unfortunately is pretty superficial for a big PR like this, but my initial reaction is that this seems great!

I'm very happy to see the text/icon-translate collision inconsistencies addressed. I think they've stuck around for so long because it's such a big job to fix them, but the approach here looks like what I'd expect. Between the existing render tests and the new ones added, I feel pretty good about this not breaking too much existing behavior. It's especially heartening that the render tests using collision debug boxes are relatively stable.

My biggest concern would be performance. Again on a superficial read, it looks like almost everything here should be some (hopefully small) constant factor increase on the cost of render-time reprojection (and the projection that happens during collision detection). This is where it would be really nice to see benchmark results (ideally a benchmark that exercises text-translate, although I don't know that it'll make a huge difference).

I'm sorry I don't have time to get into poking it enough to find errors or suggest improvements. I'll try to come back to it if I get time, but I wanted to get something written down now in case I can't come back to it. One thing I'd definitely like to do is sync the branch and play around with it locally in a debug environment, just to see how it feels.

Btw, this was my first time playing with the globe demo, and it's looking pretty good!

src/symbol/projection.ts
kubapelc: I've renamed all occurences to "SymbolProjectionContext" or similar.

src/render/draw_collision_debug.ts
kubapelc: Not anymore. I think the translate usage in the old code at this spot was wrong. Collision circles already incorporated translate when they were placed, so it should not be applied a second time.

src/symbol/projection.ts
kubapelc: In other places `projection.projectTileCoordinates` is used to get projected coordinates in clip space, as opposed to the screenspace result in pixels returned by this function. Plus adding the translation to the projected point is not always desirable when calling `projection.projectTileCoordinates`.

src/symbol/placement.ts
kubapelc: I've refactored it into a function. The insides aren't really "caching", rather it eventually passes collision boxes to the bucket, and this is only needed when `showCollisionBoxes` is true.

src/symbol/placement.ts
kubapelc: I've noticed that a lot of that code is unused I removed it. As a result the parts that were are now much smaller and mostly do different things.

src/symbol/placement.ts
kubapelc: I've added explicit types.

src/symbol/collision_index.ts
kubapelc: I can't really use either here, I need to avoid mutating both `basePoint` and `vecEast`/`vecSouth`. I've rewritten in to a form that just does the math directly with numbers, thus avoiding copying points around.

src/symbol/collision_index.ts
```diff
@@ -398,4 +497,125 @@ export class CollisionIndex {
...
+            basePoint.add(vecEast.mult(offsetXmax)).add(vecSouth.mult(offsetYmin)),
+            basePoint.add(vecEast.mult(offsetXmax)).add(vecSouth.mult(offsetYhalf)),
+            basePoint.add(vecEast.mult(offsetXmax)).add(vecSouth.mult(offsetYmax)),
+            basePoint.add(vecEast.mult(offsetXhalf)).add(vecSouth.mult(offsetYmax)),
+            basePoint.add(vecEast.mult(offsetXmin)).add(vecSouth.mult(offsetYmax)),
+            basePoint.add(vecEast.mult(offsetXmin)).add(vecSouth.mult(offsetYhalf)),
+        ];
+
+        // Is any point of the collision shape visible on the globe (on beyond horizon)?
+        let anyPointVisible = false;
```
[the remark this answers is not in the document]
kubapelc: Using a separate call is impractical, I need to access the `isOccluded` property which is only present on the intermediate result.

src/symbol/collision_index.ts
kubapelc: If I use `map`, I end up needing three loops (two maps and one for loop) instead of just one, but I would guess that performance impact would be negligible, since it is iterating over arrays of just 8 elements...

src/symbol/collision_index.ts
kubapelc: It is not needed, I've removed the globe-only branch.

kubapelc:
I've managed to fix the bug mentioned in the first comment, where `pitch-alignment: map` texts had a slightly wrong collision box in the distance.

src/symbol/collision_index.ts
```diff
@@ -580,36 +551,46 @@ export class CollisionIndex {
...
-                    oldPoint.y,
-                    unwrappedTileID,
-                    getElevation
-                );
-                points[i] = newPoint.point;
-                if (!newPoint.isOccluded) {
+            const projected = points.map(p => this.projectAndGetPerspectiveRatio(posMatrix, p.x, p.y, unwrappedTileID, getElevation));
+
+            // Is at least one of the projected points NOT behind the horizon?
+            for (let i = 0; i < projected.length; i++) {
```
HarelM, line 583: Since `i` is only used to get the current item in the for loop, consider using `for (let projectedPoint of projected)`

HarelM: I think you can you something shorter here:
`anyPointVisible = projected.some(p => !p.isOcculated)` or something similar.

src/symbol/collision_index.ts
```diff
@@ -398,4 +457,146 @@ export class CollisionIndex {
...
+        const offsetYmin = collisionBox.y1 * distanceMultiplier;
+        const offsetYmax = collisionBox.y2 * distanceMultiplier;
+        const offsetYhalf = (offsetYmin + offsetYmax) / 2;
+
+        // 0--1--2
+        // |     |
+        // 7     3
+        // |     |
+        // 6--5--4
+        const offsetsArray = [
```
HarelM, line 554: Instead of creating an array and iterate on it every second item please create an array of `{offsetX: ..., offsetY: ...}`.
This way the for loop is easier to read.
You can also do `for (let {offsetX, offsetY} of offsetsArray)` and make the for easier to read.

HarelM:
Seems like some render tests are failing, can you please check why?
I left a few last comments.

src/symbol/projection.ts
```diff
@@ -360,20 +397,64 @@ function placeGlyphsAlongLine(symbol, fontSize, flip, keepUpright, posMatrix, la
...
+ * or projects to viewport using the special map projection (mostly for globe) by calling {@link projectTileCoordinatesToViewport}.
+ */
+function _projectTruncatedLineSegment(previousTilePoint: Point, currentTilePoint: Point, previousProjectedPoint: Point, minimumLength: number, projectionMatrix: mat4 | undefined, projectionContext: SymbolProjectionContext) {
     // We are assuming "previousTilePoint" won't project to a point within one unit of the camera plane
     // If it did, that would mean our label extended all the way out from within the viewport to a (very distant)
     // point near the plane of the camera. We wouldn't be able to render the label anyway once it crossed the
     // plane of the camera.
-    const projectedUnitVertex = project(previousTilePoint.add(previousTilePoint.sub(currentTilePoint)._unit()), projectionMatrix, getElevation).point;
+    const unitVertexToBeProjected = previousTilePoint.add(previousTilePoint.sub(currentTilePoint)._unit());
+    let projectedUnitVertex;
```
HarelM, line 444: I think you can use "?" here:
```
projectedUnitVertex = projectionMatrix !== undefined 
  ? ...
  : ...
```
This will better explain that this is an assignment call.
If you prefer not to, at least add a type to `projectedUnitVertex` when defining it.

src/symbol/projection.ts
```diff
@@ -413,53 +506,93 @@ type ProjectionArgs = {
...
+}
+
+function projectTileCoordinatesToViewport(x: number, y: number, projectionContext: SymbolProjectionContext): {
+    point: Point;
+    signedDistanceFromCamera: number;
+    isOccluded: boolean;
+} {
+    const translatedX = x + projectionContext.translation[0];
+    const translatedY = y + projectionContext.translation[1];
+    let projection;
```
HarelM, line 586: Add the type. I see that this is also the return value of the method, so it will be better to define this type somewhere in this file I guess...?

HarelM:
Is there a way to benchmark this to make sure we didn't create any performance issues?

src/symbol/collision_index.ts
HarelM: There was a "break" here that was removed, was it intentional? I think you added it in order to fix some bug?
I would advise to try and create a test around this code to make sure it does what is expected.

kubapelc: The break was just an optimization (a rather pointless one for an array of size 8), there was no need to keep iterating the point array after we found first unoccluded point.

kubapelc: And thanks for the tip with `some`, I was looking for a similar function, but under the name "any"!

src/symbol/projection.ts
```diff
@@ -413,53 +521,89 @@ type ProjectionArgs = {
...
-    return projectTruncatedLineSegment(previousTilePoint, currentVertex, previousVertex, absOffsetX - distanceFromAnchor + 1, labelPlaneMatrix, getElevation);
+    const minimumLength = syntheticVertexArgs.absOffsetX - syntheticVertexArgs.distanceFromAnchor + 1;
+    return projectTruncatedLineSegmentToViewport(previousTilePoint, currentVertex, syntheticVertexArgs.previousVertex, minimumLength, projectionContext);
+}
+
+function projectTileCoordinatesToViewport(x: number, y: number, projectionContext: SymbolProjectionContext): PointProjection {
+    const translatedX = x + projectionContext.translation[0];
+    const translatedY = y + projectionContext.translation[1];
+    let projection;
+    if (!projectionContext.pitchWithMap && projectionContext.projection.useSpecialProjectionForSymbols) {
```
HarelM, line 598: Looking at the code coverage report, it seems that the content of this "if" is not covered by tests, it's worth adding a test for it I believe as this was added as part of this PR, right?

src/symbol/collision_index.ts
```diff
@@ -358,28 +407,38 @@ export class CollisionIndex {
...
-    projectAndGetPerspectiveRatio(posMatrix: mat4, x: number, y: number, getElevation?: (x: number, y: number) => number) {
-        let p;
-        if (getElevation) { // slow because of handle z-index
-            p = [x, y, getElevation(x, y), 1] as vec4;
-            vec4.transformMat4(p, p, posMatrix);
-        } else { // fast because of ignore z-index
-            p = [x, y, 0, 1] as vec4;
-            projection.xyTransformMat4(p, p, posMatrix);
+    projectAndGetPerspectiveRatio(posMatrix: mat4, x: number, y: number, unwrappedTileID: UnwrappedTileID, getElevation?: (x: number, y: number) => number) {
+        let projected;
```
HarelM, line 411: Use `... = condition ? value1 : value2`
Also this new `projectTileCoordinates` doesn't seemed to be covered as well, not sure if it's worth a test or not. your call.

src/symbol/collision_index.ts
```diff
@@ -358,28 +407,38 @@ export class CollisionIndex {
...
-            perspectiveRatio: 0.5 + 0.5 * (this.transform.cameraToCenterDistance / p[3])
+            perspectiveRatio: 0.5 + 0.5 * (this.transform.cameraToCenterDistance / projected.signedDistanceFromCamera),
+            isOccluded: (projected.isOccluded !== undefined) ? projected.isOccluded : false,
+            signedDistanceFromCamera: projected.signedDistanceFromCamera
         };
     }
 
+    getPerspectiveRatio(posMatrix: mat4, x: number, y: number, unwrappedTileID: UnwrappedTileID, getElevation?: (x: number, y: number) => number): number {
+        // We don't care about the actual projected point, just its W component.
+        let projected;
```
HarelM, line 433: Same as above.

src/geo/projection/projection.ts
```diff
@@ -0,0 +1,72 @@
...
+export function createProjection(): Projection {
+    return {
+        isOccluded(_x: any, _y: any, _t: any) {
+            return false;
+        },
+        getPitchedTextCorrection(_transform: any, _anchor: any, _tile: any) {
+            return 1.0;
+        },
+        get useSpecialProjectionForSymbols() { return false; },
+        projectTileCoordinates(_x, _y, _t, _ele) {
```
HarelM, line 28: Is this method begin used?
Maybe it's possible to reduce some of the function here?

src/geo/projection/projection.ts
```diff
@@ -0,0 +1,72 @@
...
+    isOccluded(_x, _y, _t): boolean;
+    projectTileCoordinates(_x, _y, _t, _ele): PointProjection;
+    getPitchedTextCorrection(_transform, _anchor, _tile): number;
+    translatePosition(transform: { angle: number; zoom: number }, tile: Tile, translate: [number, number], translateAnchor: 'map' | 'viewport'): [number, number];
+    getCircleRadiusCorrection(tr: any): number;
+};
+
+export function createProjection(): Projection {
+    return {
+        isOccluded(_x: any, _y: any, _t: any) {
```
HarelM, line 21: According to the code coverage, this method isn't used, is it possible?

HarelM:
Added a few comments after looking at the coverage report, which looks great BTW, good job on that!

src/symbol/projection.ts
kubapelc: The first branch of the if is new, but it is implicitly covered by render tests. The new branch mostly calls projection's `projectTileCoordinates`, which should be covered by unit tests for a given projection (in the mercator/globe sense).

src/symbol/collision_index.ts
kubapelc: Fixed.

src/geo/projection/projection.ts
kubapelc: Both this and `isOccluded` are unused, because there is no globe projection present and `useSpecialProjectionForSymbols` is always false. All the `if`s testing for `useSpecialProjectionForSymbols` could be removed, but merging globe would be somewhat harder.

src/symbol/projection.ts
HarelM: I see. Thanks.

kubapelc:
I've added a benchmark that measures the performance of adding symbols to the collision grid and compared this branch to main.

When testing 20K symbols and adding all non-colliding symbols to the grid (as normal symbol placement would), I got ~170 ms on this branch compared to 90-100 ms on main.

When testing 20K symbols and only testing the collision box projection (replacing the grid test with a empty function), I got ~42 ms on this branch, compared to ~3 ms on main.

Only the second test is now present, the first is in my branch's commit history.

## Analysis

I'm not sure how much merit the first test has, since the old collision box projection was pretty broken, so a lot of the performance difference might be caused by the collision grids being filled with different contents.

The second test shows that the accurate projection of collision boxes has a lot of overhead compared to the old approach. But the old approach basically only did 4 additions and 4 multiplications per box.

Both tests use 20K symbols, which is pretty extreme, I'd expect a typical map to have roughly 100x less symbols. Extra cost due to accurate projection would then be 0.4 ms. This is probably not worth optimizing further, especially with the assumption that symbol placement runs asynchronously in a separate thread, so it would not slow down rendering.

HarelM (approved):
THANKS!

ChrisLoer:
> This is probably not worth optimizing further, especially with the assumption that symbol placement runs asynchronously in a separate thread, so it would not slow down rendering.

Is that asynchronous assumption based on work in progress somewhere else? I just did a quick check and it looks to me like placement still follows the logic I remember: synchronous in the main/rendering thread, but with logic in pauseable_placement to pause itself if it goes over 2ms and continue the work in the next frame. That is, there's still a safety valve to try to keep slow placement from causing dropped frames, but it's not asynchronous.

.4ms is definitely enough to matter, but as you said that's with a pretty large set of symbols -- and fundamentally this functionality is going to require some extra computation, so it doesn't seem terrible to me.

kubapelc:
> Is that asynchronous assumption based on work in progress somewhere else?

Hi, no, I'm dumb and only remembered that collisions somehow don't slow down main rendering or happen alongside it, and assumed that it is asynchronous. You are of courses right.

The 0.4 ms figure is an estimate of how long would 200 symbols take. Actually running the benchmark with 200 symbols takes 0.6 ms.

PR #4072, opened 2024-05-03
Support cross-fading between old & new tiles of the same zoom level

### LINKED ISSUE #3595: Raster zoom or layer replace flicker when raster layer takes up most of the browser viewport
**maplibre-gl-js version**: 3.5.0

**browser**: Chrome 120.0.6099.224

### Steps to Trigger Behavior

 1. Load raster layer
 2. Try stepping back & forth with raster layer taking up ~2/3 or less of the browser viewport 
 3. Decrease size of browser so that the raster layer takes up all or nearly all of the browser viewport and step back & forth between the same 2 zoom levels. Observe flicker

### Link to Demonstration

I'm not sure how clear this demo is since these videos get auto-resized to fit, but please understand that both of these demonstrations are moving between 2 zoom levels that request that same tile zoom/z of 5. The only difference between them is the size of the viewport. Also, apologies for video jankiness in general, but I think you can still see the load hesitation and flicker on the 2nd one, even with the jank factor.

## What works fine

[Screencast from 2024-01-18 20-57-54.webm](https://github.com/maplibre/maplibre-gl-js/assets/133472204/0669612b-c242-4669-978e-0fa16c9f7e8f)

## What doesn't work as well, same zoom levels, smaller viewport

[Screencast from 2024-01-18 20-58-51.webm](https://github.com/maplibre/maplibre-gl-js/assets/133472204/5ae0101b-48b9-4314-b345-97df2ffc6c2e)

### Why I care

I'd like to understand if this is avoidable behavior. I'm currently working on cache-driven playback capability and this is the simplest repeatable example I can provide to show the issue that affects my ability to "play" layers back over time. Basically, I have a setup that supports swapping out raster layers quickly, and it works really well when the layer doesn't take up most of the viewport, but there's a clear flicker and queue backlog when the raster layer takes up most of the viewport.

Happy to tackle a fix (if fixable) provided someone can help me understand the problem!

wagewarbler:
These changes resolve #3595. 

`maplibre-gl-js` tracks both parent and children tiles so that when a user zooms in/out, tiles of the most recent previous zoom level are retained and kept on the map until the current layer of tiles populates. This creates a more seamless experience when navigating the layer.

Maintaining just parents and children was sufficient to cover zoom changes, but with the introduction of the `setTiles` method in #3208, it also then became desirable to maintain the most recent batch of tiles at the same zoom level. This PR adds a "sibling" tile cache.

Demo of map performance before the change (note the labels you can see in the underlying basemap during transitions):

3May24-a

Demo of map performance after the change (no labels seen during transitions, older tiles kept until new ones appear):

3May24-b


HarelM:
Is this logic relevant only to raster source or to vector too?

Cc: @msbarry as you are looking into this area as well. 
The more I read the more I think source cache is doing "too much" and should be split to different types/classes/strategies acording to the source it manages...

wagewarbler:
> Is this logic relevant only to raster source or to vector too?

Very possibly since I've seen similar behavior in vector tiles. It's on my list to better understand the parallels there since we're working on animation capability for both raster imagery and vector tiles.

> Cc: @msbarry as you are looking into this area as well. The more I read the more I think source cache is doing "too much" and should be split to different types/classes/strategies acording to the source it manages...

I'm sure there are people who have a more complete view of the library as a whole, but my 2 cents here is that this file, and much of the associated logic is ripe for a refactor. Most of it doesn't appear to have been modified since well before Maplibre-gl split off. Selfishly I'd like to see this change incorporated prior to such an effort, but I'm happy to work with whatever makes the most sense here.

msbarry:
Got it! As long as there are some comments in source_cache explaining what sibling tiles are and why they are needed we should be able to include in future refactors.

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 - Fix ScaleControl options should be optional. ([#4002](https://github.com/maplibre/maplibre-gl-js/pull/4002))
 - Fix race condition in `SourceCache` that makes unit tests unstable. Eliminate a redundant 'visibility' event fired from Style class. ([#3992](https://github.com/maplibre/maplibre-gl-js/issues/3992))
 - Fix paint property not being updated by setPaintProperty ([#2651](https://github.com/maplibre/maplibre-gl-js/issues/2651))
+- Allow cross-fading between raster tile source updates at the same zoom level ([#4072](https://github.com/maplibre/maplibre-gl-js/pull/4072))
```
HarelM, line 25: Can you push this up? This is currently under 4.1.3, it should be under "main".

src/source/source_cache.test.ts
```diff
@@ -1800,6 +1800,101 @@ describe('SourceCache#findLoadedParent', () => {
...
+        tr.width = 512;
+        tr.height = 512;
+        sourceCache.updateCacheSize(tr);
+
+        const tile = new Tile(new OverscaledTileID(1, 0, 1, 0, 0), 512);
+        sourceCache._cache.add(tile.tileID, tile);
+
+        expect(sourceCache.findLoadedSibling(new OverscaledTileID(1, 0, 1, 1, 0))).toBeNull();
+        expect(sourceCache.findLoadedSibling(new OverscaledTileID(1, 0, 1, 0, 0))).toBe(tile);
+        expect(sourceCache._cache.order).toHaveLength(1);
```
HarelM, line 1837: Is there a better way to write this expect without looking into "internal" fields?

src/source/source_cache.ts
```diff
@@ -411,6 +412,14 @@ export class SourceCache extends Evented {
         }
     }
 
+    /**
+     * Find a loaded parent of the given tile (up to minCoveringZoom)
```
HarelM, line 416: "parent"? is this intentional?

HarelM:
Is it somehow possible to take a look at this issue while we are changing the logic here:
- #4074

Seems like it touches the same code and since you changed it I would hope you fully understand it and it should be easy to fix the above linked issue...?

HarelM:
I added some minor comments.

src/source/source_cache.test.ts
```diff
@@ -1800,6 +1800,101 @@ describe('SourceCache#findLoadedParent', () => {
...
+
+        // Arbitrary tiles should not in the cache
+        const notLoadedTiles = [
+            new OverscaledTileID(2, 1, 2, 0, 0),
+            new OverscaledTileID(2, 0, 2, 3, 0),
+            new OverscaledTileID(2, 0, 2, 3, 3),
+            new OverscaledTileID(3, 0, 3, 2, 1)
+        ];
+
+        expect(notLoadedTiles[0].key in sourceCache._loadedSiblingTiles).toBe(false);
```
HarelM, line 1891: Is it possible to use here more "public" APIs like `findLoadedSibling` or have the cache be mocked by a spy and check the content of that spy?

src/source/source_cache.ts
```diff
@@ -559,7 +568,7 @@ export class SourceCache extends Evented {
         const retain = this._updateRetainedTiles(idealTileIDs, zoom);
 
         if (isRasterType(this._source.type)) {
-            const parentsForFading: {[_: string]: OverscaledTileID} = {};
+            const tilesForFading: {[_: string]: OverscaledTileID} = {};
```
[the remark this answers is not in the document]
wagewarbler: Agreed it was a hefty chunk of logic. Moved to a new method

src/source/source_cache.test.ts
wagewarbler: Removed all internal field refs in the new logic. Added methods `SourceCache` > `getSource` and `getCache` to help here.

wagewarbler:
> Got it! As long as there are some comments in source_cache explaining what sibling tiles are and why they are needed we should be able to include in future refactors.

Matching the trend in `source_cache` of throwing more comments on the methods, I did the same in [this commit](https://github.com/maplibre/maplibre-gl-js/pull/4072/commits/897355176defae37b5286f9daaa99e2e2e2e0492). Additionally, please feel free to tag me on any refactors in the future. Happy to help however I can.

HarelM:
@wagewarbler any inputs about the issue by any chance? #4074

CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 ## main
+- Update Popup methods `addClass` & `removeClass` to return instance of Popup ([#3975](https://github.com/maplibre/maplibre-gl-js/pull/3975))
```
HarelM, line 2: This was incorrectly merged.

Also the changelog item should be under "✨ Features and improvements"

src/source/source_cache.test.ts
```diff
@@ -1800,6 +1800,101 @@ describe('SourceCache#findLoadedParent', () => {
...
+
+        // Arbitrary tiles should not in the cache
+        const notLoadedTiles = [
+            new OverscaledTileID(2, 1, 2, 0, 0),
+            new OverscaledTileID(2, 0, 2, 3, 0),
+            new OverscaledTileID(2, 0, 2, 3, 3),
+            new OverscaledTileID(3, 0, 3, 2, 1)
+        ];
+
+        expect(sourceCache.findLoadedSibling(notLoadedTiles[0])).toBe(null);
```
HarelM, line 1891: there are some lint warnings, can you please fix them?

src/source/source_cache.test.ts
```diff
@@ -1800,6 +1800,101 @@ describe('SourceCache#findLoadedParent', () => {
...
+
+        // Loaded tiles should be in the cache
+        expect(sourceCache.findLoadedSibling(tiles[0]).tileID).toBe(tiles[0]);
+        expect(sourceCache.findLoadedSibling(tiles[1]).tileID).toBe(tiles[1]);
+        expect(sourceCache.findLoadedSibling(tiles[2]).tileID).toBe(tiles[2]);
+        expect(sourceCache.findLoadedSibling(tiles[3]).tileID).toBe(tiles[3]);
+        expect(sourceCache.findLoadedSibling(tiles[4]).tileID).toBe(tiles[4]);
+        expect(sourceCache.findLoadedSibling(tiles[5]).tileID).toBe(tiles[5]);
+
+        expect(tiles[0].key in sourceCache._loadedSiblingTiles).toBe(true);
```
HarelM, line 1876: This still doesn't look great...

src/source/source_cache.ts
```diff
@@ -149,6 +150,14 @@ export class SourceCache extends Evented {
         return this._source;
     }
 
+    getCache(): TileCache {
```
HarelM, line 153: Is there a better way to test these internal properties without exposing them?

src/source/source_cache.ts
```diff
@@ -485,6 +502,90 @@ export class SourceCache extends Evented {
         }
     }
 
+    _handleFadingTiles(
+        retain: { [_: string]: OverscaledTileID },
+        minCoveringZoom: number, maxCoveringZoom: number, zoom: number, idealTileIDs: OverscaledTileID[], terrain?: Terrain) {
```
HarelM, line 507: Can you split this to more lines for readability?

src/source/source_cache.ts
```diff
@@ -485,6 +502,90 @@ export class SourceCache extends Evented {
...
+            fadingTiles[id] = tileID;
+        }
+
+        // for tiles that are still fading in, also find children to cross-fade with
+        this._retainLoadedChildren(fadingTiles, zoom, maxCoveringZoom, retain);
+
+        for (const id in tilesForFading) {
+            if (!retain[id]) {
+                // If a tile is only needed for fading, mark it as covered so that it isn't rendered on it's own.
+                this._coveredTiles[id] = true;
```
HarelM, line 542: Is this the only member that this method updates? If so, the name of the method might be:
`_updateCovereredTile` or `_updateCoveredAndRetainedTiles`.

wagewarbler:
> @wagewarbler any inputs about the issue by any chance? #4074

So far only that I've also observed this behavior and I have some high-level guesses that require a little more digging. Will comment on the issue next! Thanks for the latest round of feedback, had one other thing come up but will circle back soon.

src/source/source_cache.test.ts
wagewarbler: Actually removed these. They're redundant with the previous testing block

src/source/source_cache.ts
wagewarbler: I've [removed these methods](https://github.com/maplibre/maplibre-gl-js/pull/4072/commits/350a2c1bbf71f0b202c86c297dfbcb547a9c1c11) and replaced with them with ones [on the mock](https://github.com/maplibre/maplibre-gl-js/pull/4072/commits/8f7980a0f2958a7633492010b23918b7db97b153) instead

CHANGELOG.md
HarelM: Please fix this... This changelog already exist in a previous version...

wagewarbler: Yikes. Bad merge. Fixed!

HarelM:
Are the following coverage missing lines related to this PR or are they considered missing because you moved the logic into a method?
image

HarelM (approved):

HarelM:
I found the answer, it was moved.
The following are not covered though:
image

Seems like an edge case, so it would be nice to add a test that cover this flow, but not mandatory.
Let me know what you decide to do.

wagewarbler:
> Seems like an edge case, so it would be nice to add a test that cover this flow, but not mandatory. Let me know what you decide to do.

Not a problem. I've added a test to cover it.

HarelM:
Are you sure? I don't think the coverage report have changed...
Am I missing anything?

wagewarbler:
> Are you sure? I don't think the coverage report have changed... Am I missing anything?

Strange, maybe I'm missing something. I've added (this test)[https://github.com/maplibre/maplibre-gl-js/pull/4072/commits/5c6d8f5721bafe5399753d22ecbffb0a16c7525b] which I've run locally and executes & passes with the expected results:

image

I could certainly make the test more rigorous, but seems like overkill. In its current form it should be sufficiently executing code within the `if` statement, otherwise we'd end up with the original length 4 array.

wagewarbler:
Actually, I think I see what's happening. The tile cache [doesn't permit duplicate tiles](https://github.com/maplibre/maplibre-gl-js/blob/0ecc76e3aa65aabf4807165f844461f1d2999c36/src/source/source_cache.ts#L794), so this logic is unnecessary. I can go ahead and remove the test and the `if` statement`.

PR #4075, opened 2024-05-04
Upgrade docs plugin to 4.0

HarelM:
## Launch Checklist

Upgrade the docs to latest version of the markdown plugin, which allows to remove the "Class" prefix from all the doc files of the API.
It created other issues that needed fixing.
I have removed all the duplicate return `this` from the docs.

I also removed most of the use of `function` in the code and docs.
I also reduced the time in the marker tests, which was too long.

Before:
image
After:
image

Before:
image
After:
image


sbachinin (approved):
As for these big timeouts in marker tests, I remember I had unstable results on my machine with 100 timeout. It's not the fastest machine but reasonable

HarelM:
Let's see if we experience failures in CI add address those.
It took like extra 10 second for this class only, which is a bit too much (jest even complain).
We might need a better trigger than "sleep".

PR #4082, opened 2024-05-06
Implement the `getData` method

### LINKED ISSUE #4080: Add `getData` method for GeoJSON sources.
## User Story

I suggest adding a `getData` method to `GeoJSON` sources as a unified way of accessing all the source's features regardless of their visibility.

## Rationale

There are currently 4 (?) known ways to that kind of thing:

1. `queryRenderedFeatures` - it allows to query the features of a source, but only those in the viewport.
2. `querySourceFeatures`  - as it turned out, the name is a bit misleading. If the target feature's not in the viewport, it won't locate it (just like with `queryRenderedFeatures`). It also can provide duplicates of real data, which leads to mandatory client-side filtering thus decreasing the performance.
3. Use `map.getSource(sourceId)._data`. This `_data` property seems to be holding all the features of a given source. The issue's that it's not a part of the public API, so it's a bad thing to rely on it. Secondly, it won't reflect the changes to the source made with the `updateData` method.
4. Use a separate features pool. Basically, write functions like `addFeatures`, `removeFeatures` and so on where it'd be required not only to call the `source.updateData` , but also to write into some kind of a local pool of features. That way, all the saved features will then be available to search through them.

It's been advised to use option number 4 as our current way-to-go. However, I consider it to be not a very nice one for the following reasons:

1. Retreiving data is a pretty common operation. Making all the users have their own omplementation of such a common thing is not a good idea, as it's something that could be done on the level of the library.
2. Holding such a features pool increases the complexity of data manipulation: it's like re-inventing the `updateData` method but the one which writes to a different location.
3. All the source data exists all the time and the map knows about it. There should be no need to store a separate copy of that data.

Another reason I think why it might be a good idea to have such an API is that there are already methods like `setData` and `updateData`, which logically leads to also having a read-only counterpart `getData`. And at the moment it's not there.

## Impact

If we don't provide this method, nothing special would happen. It's just a metter of convenience.
On the other hand, having such a functionality might help to decrease some unnecessary performance overhead for the users of MapLibre.

smellyshovel:
## Launch Checklist

Closes #4080.
 - [x] Link to related issues.
 - [ ] Post benchmark scores.

This one is "in progress" untill confirmed that the implementation is OK. After that I can start adding wirting tests and will also check the docs.


HarelM:
The direction is good, but I think there's alot of cleaning needed in the geojson source that runs on the main thread, as this proposal will add data in the worker (which might be more than one worker BTW) and there's a need to clean this data from the main thread. 
Sorry for misleading in terms of the time it will take, I thought the geojson source was in a better shape 😔

smellyshovel:
> and there's a need to clean this data from the main thread.

Do you mean that I need to remove the `_data` from `geojson_source`? The one that's used in `setData` method e.g.?

Or if not, then I'm not sure what "cleaning" it's about. Moreover, cleaning (in general) seems to me like a separate effort anyway, but if you were talking about `_data` then maybe it's a good idea to remove it as a part of this PR, I don't know.

HarelM:
I'm mainly talking about `_data` in geojson_source.
But it looks like it is being used, so there is probably a need to do some other changes in this class in order to make sure the data is only stored in the worker.
I hope this is clearer...

smellyshovel:
> I'm mainly talking about `_data` in geojson_source. But it looks like it is being used, so there is probably a need to do some other changes in this class in order to make sure the data is only stored in the worker. I hope this is clearer...

Seems like there's no way to get rid of *storing the data* in the main worker.

I could get rid of the `_data` property as I first thought that it's only used to pass data from `setData` to `_updateWorkerData` (so I wanted to pass it as a parameter instead), but I've just noticed that the `serialize` method also depends on it.

On the other hand, there is also `this._options.data` (which looks to me holding the same thing as `_data` does), but even if I get rid of `_data` itself, it still won't mean getting rid of storing the copy of the data itself on the main worker.

So, apparently, the situation is the following:

- There are two copies of the GeoJSON data
  - Main worker (and it's stored in two places, however, I'm not sure whether it's the same data or a copy)
    - `this._data` - *we can get rid of this one*
    - `this._options.data` - *but there'd still remain this one*
  - Thread worker - *requires async access. So making a call to get the data from the worker each time instead of accessing some local property is not an option, as it would require to change the `serialize` method's signature (at least)*

smellyshovel:
Also, if you think about it. Passing GeoJSON data from a caller to the worker directly (bypassing the main thread) is impossible since the data is initially passed from the main thread.

When it's passed from the caller to the `geojson_source`, it's the same data. So there's no duplication.

When it's passed from the main thread to the worker, it's going to be copied anyway. 

In other words, the data which is needed to be passed from the main thread to a worker thread can not not exist on the main thread. The data will have to be duplicated between main and worker whether we like it or not. What can be achieved instead is *hiding* the data on the main thread from `geojson_source`, but is it really necessary?

source

HarelM:
Uh, I see you point.
I think that data in the main thread can be a url and not only geojson, and I wouldn't want to make the serialization async.
So let's just make sure that when the data is requested, we get the geojson data that will include diff if they were applied.

smellyshovel:
@HarelM
I changed the implementation a bit. Now we're explicitly saving not only the data itself in the worker, but also the promise which resolves with that data. This is necessary for the cases when GeoJSON is e.g. loaded via a URL, and the network request takes some time to complete, but the `getData` method was invoked without waiting (i.e. right after setting/updating the data with `setData` or `updateData`).

I tested it manually with live GeoJSON, static GeoJSON, setting data with `setData` and after updating it with `updateData`.

I also added tests for the aforementioned cases on the worker level, and added a test for the main worker (but it's a bit dumb since the worker and communication with if is mocked there).

Could you please take a look and tell me whether anything else's needed?

HarelM:
I'm so happy I introduced promises...

src/source/geojson_source.test.ts
```diff
@@ -403,6 +403,28 @@ describe('GeoJSONSource#update', () => {
     });
 });
 
+describe('GeoJSONSource#getData', () => {
+    const mapStub = {
+        _requestManager: {
+            transformRequest: (url) => { return {url}; }
+        }
+    } as any;
+    test('sends a message with a correct type to the worker and forwards the data provided returned byit', async () => {
```
HarelM, line 412:
```suggestion
    test('sends a message with a correct type to the worker and forwards the data provided returned by it', async () => {
```

src/source/geojson_source.ts
```diff
@@ -14,6 +14,7 @@ import type {GeoJSONSourceSpecification, PromoteIdSpecification} from '@maplibre
 import type {GeoJSONSourceDiff} from './geojson_source_diff';
 import type {GeoJSONWorkerOptions, LoadGeoJSONParameters} from './geojson_worker_source';
 import {MessageType} from '../util/actor_messages';
+import {GeoJSON} from 'geojson';
```
HarelM, line 17: Use GeoJSON. Instead of importing as this is only a type.

src/source/geojson_source.ts
```diff
@@ -232,6 +233,21 @@ export class GeoJSONSource extends Evented implements Source {
...
+     *
+     * @returns a promise which resolves to the source's actual GeoJSON data
+     */
+    async getData(): Promise<GeoJSON> {
+        const options: LoadGeoJSONParameters = extend({type: this.type}, this.workerOptions);
+
+        try {
+            return await this.actor.sendAsync({type: MessageType.getData, data: options});
+        } catch (err) {
+            this.fire(new ErrorEvent(err));
```
HarelM, line 247: I would remove the try catch fire here. If someone is interested in catching let them...

src/source/geojson_worker_source.ts
```diff
@@ -18,6 +18,7 @@ import type {LoadVectorTileResult} from './vector_tile_worker_source';
 import type {RequestParameters} from '../util/ajax';
 import {isUpdateableGeoJSON, type GeoJSONSourceDiff, applySourceDiff, toUpdateable, GeoJSONFeatureId} from './geojson_source_diff';
 import type {ClusterIDAndSource, GeoJSONWorkerSourceLoadDataResult, RemoveSourceParams} from '../util/actor_messages';
+import {GeoJSON} from 'geojson';
```
HarelM, line 21: Remove, same as the other comment

src/source/geojson_worker_source.ts
```diff
@@ -59,6 +60,17 @@ type GeoJSONIndex = ReturnType<typeof geojsonvt> | Supercluster;
...
+     * The actual GeoJSON takes some time to load (as there may be a need to parse a diff, or to apply filters, or the
+     * data may even need to be loaded via a URL). This promise, when resolved, indicates that the actual data in the
+     * `_data` field is ready to be consumed.
+     */
+    _pendingData: Promise<GeoJSON>;
+    /**
+     * This `_data` property holds the source's actual GeoJSON. It's updated each time the data is updated in the
+     * `loadData` method.
+     */
+    _data: GeoJSON;
```
HarelM, line 73: I don't think this data is needed now, take the value from the promise itself.

smellyshovel: The issue is that (from my understanding) the awaited `_pendingData` won't have the modifications applied further in the `loadData` if we do it that way.

Namely, `this._data` includes modifications from lines 133-141, and the resolved promise won't have them applied.

smellyshovel: Screenshot_20240507_165814

smellyshovel: I also thought that relying on `_pendingData` but returning `_data` may be a bad idea as `_data` might not yet be fully handled, but that won't be a case (tested) because this `loadData` method will have been finihed already (and thus the `_data` updated) by the time when `getData` is invoked.

src/source/geojson_source.ts
smellyshovel: Sorry, don't get this one. You mean that I should use `import type {GeoJSON} ...` here and in the next commented place?

src/source/geojson_source.ts
smellyshovel: Done

src/source/geojson_worker_source.ts
HarelM: You can move the entire handing of data to an async method calls say `processGeojsonData` and have the pending promise use the return value of that method.
This way you know that what you want is the output of this method.
This is what you want to basically wait for - all the processing, otherwise, you are complicating the state even further with two variables that has dependency on one another.

src/source/geojson_source.ts
HarelM: Remove the import and use `GeoJSON.GeoJSON` instead.

smellyshovel: Done

src/source/geojson_worker_source.ts
smellyshovel: Done

src/source/geojson_worker_source.ts
smellyshovel: Totally agree with all you say. I also didn't like that code, but was afraid to touch anything.

I took a slightly different approach. Instead of moving all the processing logic into a separate call, I put it all inside `loadGeoJSON`. At least because its documentation says

Screenshot_20240507_173810

smellyshovel: So, the flow is:
`setData(geoJSON)` -> main worker `_updateWorkerData()` -> thread worker `loadData(params)` -> `loadGeoJSON(params)` -> write to `this._pendingData`.
When `getData` is called, the `await this._pendingData` is read.

src/util/actor_messages.ts
```diff
@@ -7,6 +7,7 @@ import type {PluginState} from '../source/rtl_text_plugin_status';
 import type {LayerSpecification} from '@maplibre/maplibre-gl-style-spec';
 import type {OverscaledTileID} from '../source/tile_id';
 import type {GetResourceResponse, RequestParameters} from './ajax';
+import {GeoJSON} from 'geojson';
```
HarelM, line 10: Remove import from here as well.

test/examples/live-geojson.html
```diff
@@ -68,4 +68,4 @@
     });
 </script>
 </body>
-</html>
\ No newline at end of file
+</html>
```
HarelM, line 71: Undo change here?

HarelM:
I think there's a test failing that is related to geojson style, can you check what went wrong?

test/examples/live-geojson.html
smellyshovel: I can't! :smile: 
It won't remove it even with via the GitHub editor.
I tried to rollback, but it's still there. Not sure what's the matter.
I can try opening a new PR later without touching this file at all, if that's crucial.

smellyshovel:
> I think there's a test failing that is related to geojson style, can you check what went wrong?

Trying to reproduce it locally. But no luck so far. Maybe a flacky test?

HarelM:
I'm not familiar with this test as being flacky, and I also did a lot of work to reduce flacky tests.
It also looks like a genuine diff:
image

src/source/geojson_worker_source.ts
```diff
@@ -111,25 +117,11 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
...
-
-                const features = (data as any).features.filter(feature => compiled.value.evaluate({zoom: 0}, feature));
-                data = {type: 'FeatureCollection', features};
-            }
+            this._pendingData = this.loadGeoJSON(params, this._pendingRequest);
 
             this._geoJSONIndex = params.cluster ?
-                new Supercluster(getSuperclusterOptions(params)).load((data as any).features) :
-                geojsonvt(data, params.geojsonVtOptions);
+                new Supercluster(getSuperclusterOptions(params)).load((await this._pendingData as any).features) :
```
HarelM, line 123: I'm not sure if this is the cause, but super cluster is being built before the promise resolves.
Try moving the `await` to be before this part and not inside the load.
That's the only thing I can think of.
Or maybe the params object is somehow changed when creating the expression...
IDK...

smellyshovel: Thanks for the tip. I can't reproduce this locally. Will try your suggested fixes soon.

HarelM: You can try codespaces too, it's linux based and is using the same CI infrastructure.
But the CI can also do it.

smellyshovel: This particular render-test (the one that fails) relies on dev-build. And I was updating the prod-build. This is why I didn't see it fail.
Might be worth mentioning in the contributing guide.
Looking into it now.

smellyshovel:
@HarelM now tests pass. Still don't understand what was the issue. I though about early-returns in `loadGeoJSON`, but niether dealing with them helped. Anyway, the issue's gone now.

HarelM:
This is so strange!
This is the diff between the commits:
https://github.com/maplibre/maplibre-gl-js/compare/98364e6..e0ee2d2
I took the code and moved from one version to the other with simply moving code from one place to another and I couldn't find the actual differences.
It might be how the code is translated to javascript, I can't think of another reason...
In any case, it passes the tests...

Can you please add more tests to the worker source to improve coverage?
Coverage right now is not great...

HarelM:
Please add changelog item too.
Thanks!

smellyshovel:
> This is so strange! This is the diff between the commits: https://github.com/maplibre/maplibre-gl-js/compare/98364e6..e0ee2d2 I took the code and moved from one version to the other with simply moving code from one place to another and I couldn't find the actual differences. It might be how the code is translated to javascript, I can't think of another reason... In any case, it passes the tests...
> 
> Can you please add more tests to the worker source to improve coverage? Coverage right now is not great...

Yeah, that's really weird.

I'll take a look at the coverage and see what I can do.

HarelM:
I don't think you added code that is not covered but rather moved code that is not covered. 
It can be out of scope for this PR, I was just hoping to improve the situation since you are already familiar with the code and tests.
Your call. Whatever you choose is fine.

smellyshovel:
> moved code that is not covered

exactly. It was not covered, I moved it, and it's still not covered. So, realistically, nothing's changed.

> familiar with the code and tests

quite a bold statement :smile: 

> Your call. Whatever you choose is fine.

I'd choose not to touch them now really. I'm eager for a release which would include the new `getData` method, and I don't relly want to postpone it anymore with tests, if possible. Thanks for your understanding :+1: Would love to contribute more to the project, but I'm really low on spare time now.

HarelM (approved):

PR #4092, opened 2024-05-08
use `with` for json modules instead of `assert`

birkskyum:
The JSON modules / import attribute syntax has been changed from `assert` to `with`, which it appear to finally stabilize on.

https://github.com/tc39/proposal-json-modules
https://github.com/tc39/proposal-import-attributes/

### DIFF
--- a/build/banner.ts
+++ b/build/banner.ts
@@ -1,4 +1,4 @@
-import packageJSON from '../package.json' assert {type: 'json'};
+import packageJSON from '../package.json' with {type: 'json'};
 
 export default
 `/**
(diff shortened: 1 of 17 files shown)


HarelM:
Does it require a newer node version? Does it create any rollup issue? 
I'm surprised nothing is failing...

HarelM (approved):

birkskyum:
I believe it works because the `with` syntax has been added in minor versions, like here for 18.20 (https://github.com/nodejs/node/pull/52165), and probably the same for node 20.

PR #4122, opened 2024-05-16
Fix node-canvas source compile for node >= v21

springmeyer:
The `maplibre-gl-js` install currently fails on arm64 macs when running Node >= 21.

The error is:

```
../../nan/nan.h:2546:8: error: no matching member function for call to 'SetAccessor'
tpl->SetAccessor(
```

The context is that `maplibre-gl-js` depends on `node-canvas@2.x` which has a dependency on the `nan` package and `nan` 2.17 had a bug that surfaced with node v21: https://github.com/nodejs/nan/pull/966.

Additionally `node-canvas` does not currently provide pre-compiled binaries for Arm64, so a source compile is needed.

So, the solution here is to upgrade the `nan` version to the latest release which fixes support with recent node.

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3942,7 +3942,7 @@
       "hasInstallScript": true,
       "dependencies": {
         "@mapbox/node-pre-gyp": "^1.0.0",
-        "nan": "^2.17.0",
+        "nan": "^2.19.0",
         "simple-get": "^3.0.3"
       },
       "engines": {
@@ -9446,7 +9446,7 @@
       "license": "MIT"
     },
     "node_modules/nan": {
-      "version": "2.17.0",
+      "version": "2.19.0",
       "dev": true,
       "license": "MIT"
     },


HarelM:
Wouldn't a newer version of node-canvas will solve this in the future?
Node 20 is what the CI and everything is supporting and we usually only use LTS to be on the safe side...
What's the motivation behind using node 21 and working on this library?

springmeyer:
Thanks for your thoughtful questions @HarelM 

> Wouldn't a newer version of node-canvas will solve this in the future?

Yes. Upgrading to a newer node-canvas would also solve this, if/when a newer version is released (because the `package-lock.json` would get updated in the same way this PR updates it). Currently maplibre-gl-js uses the latest release from last year: [v2.11.2](https://github.com/Automattic/node-canvas/releases/tag/v2.11.2) and I don't see a newer release available. I did notice that the `main` branch of node-canvas declares a 3.x series. So in anticipation that 3.x might bring breaking changes and pose difficulty upgrading, it seemed like working around this problem in maplibre-gl-js was reasonable. Or we could advocate for a new node-canvas v.2.x release.

> Node 20 is what the CI and everything is supporting and we usually only use LTS to be on the safe side...

Good to know. 

> What's the motivation behind using node 21 and working on this library?

Node v22 will enter LTS in 5 months (https://nodejs.org/en/blog/announcements/v22-release-announce). I happened to be running it when I built latest maplibre today, so I was motivated to help others in the future avoid this potential install problem.

HarelM (approved):

HarelM:
Fair enough, thanks!

PR #4124, opened 2024-05-17
Symbols: Allow text line breaks on lines

### LINKED ISSUE #3814: Line breaks with \n or \r not working properly for symbol layers with line or line-center placements
**maplibre-gl-js version**: 4+

**browser**: chrome (probably any)

### Steps to Trigger Behavior

 1. Create a GeoJSON source and symbol layer
 2. Show a piece of text with "text-field" containing a line break in it \n
 3. Start with "symbol-placement": "point", give it some coordinates
 4. The text should appear correctly in 2 lines
 5. Now change the source to a line string, draw a basic line
 6. Change the symbol placement to "line" or "line-center"
 7. BUG: the text now appears in one line
 8. GLITCH: if you zoom all the way out, wait a moment, and zoom back in, now the text might appear in 2 lines

### Expected Behavior

Symbol layers with "line" or "line-center" placement containing "text-field" values with line breaks should display such texts in multiple lines.

### Actual Behavior

Symbol layers with "line" or "line-center" placement containing "text-field" values with line breaks are displaying the text on a single line.

LostDragonist:
Resolves #3814 

The change is pretty simple. The text shaper was explicitly disallowing any text breaks if the 'symbol-placement' was not set to 'point'.  This removes that logic, allowing text breaks to be processed for any 'symbol-placement' value.  The other bits are just removing the now unused variable from calling functions.

No new unit test as the text shaper unit test is sufficient now that it's not specifically avoiding the problem.  

The only pain point in my mind is that someone may be relying on this behavior for some reason. There doesn't appear to be any property on symbol layers or any expressions that could be used to quickly change back to the previous behavior. Though I'd argue if you don't want explicit line breaks in your data, don't add explicit line breaks in your data?

For non-explicit line breaks, setting 'text-max-width' to a large number should get back to the previous behavior.

Before (string is "A\nBB\nCCC"): 

After (string is "A\nBB\nCCC"):

## Launch Checklist

 - [x] Link to related issues.
 - [x] Write tests for all new functionality. (none, see above)
 - [ ] Fix / update all the affected render tests
 - [x] Document any changes to public APIs. (none)
 - [x] Post benchmark scores.


HarelM:
This seems intentional, any chance you can try and find the historical reason for this decision? 
Also please add a changelog item and fix the tests.

HarelM:
Thanks for looking into this! 
Style-spec docs are residing in the maplibre-style-spec repo. Please open a PR there. 
A render test with line break would be great here. 
Also coverage of changed lines isn't great, some unit tests to cover the changes world be helpful.

wipfli:
Thanks for opening this pull request. I found the behavior around newlines strange in the past and here you describe well what the situation is.

This pull request will make existing maps look different which is something we should avoid. Is it possible to introduce an new style spec property to take account of the line breaks?

HarelM:
@wipfli I think we should also avoid complicating the style due to old bugs or missing implementation.
If people don't want to have new lines in the map they should make sure they don't have new lines in their data and not relay on something that is not implemented.
We can introduce this, assuming we think it's a breaking change, in a breaking change version (like we plan to for geometry-type).

HarelM:
What will happen if the line is not straight? Like when adding a text to a river? What will happen in this scenario when adding a new line?

HarelM (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
+
+- Fix ignoring embedded line breaks when `symbol-placement` is `line` or `line-center` ([#4124](https://github.com/maplibre/maplibre-gl-js/pull/4124))
```
HarelM, line 10:
```suggestion
- ⚠️ Fix ignoring embedded line breaks when `symbol-placement` is `line` or `line-center` ([#4124](https://github.com/maplibre/maplibre-gl-js/pull/4124))
```

wipfli:
Since this is a breaking change I suggest we go into pre-release mode of MapLibre GL JS v5.

wipfli:
@HarelM can you open a js-parity issue in MapLibre Native for this change?

HarelM:
@louwers, I'm delegation opening the feature parity issue to you.

PR #4140, opened 2024-05-20
Globe - example images

kubapelc:
This PR adds generated images for the globe examples. I've also removed one example, since it was not really showing anything the other examples don't show already, and it was generating a blank image for whatever reason.


HarelM (approved):

HarelM:
THANKS!

PR #4145, opened 2024-05-21
Improve the handling of map options

vicb:
Content of this PR:

- Fix comments in `type MapOptions`
- Order default values using the type order,
- Update Map props:
  - use inline init where applicable
  - fix types where applicable
  - fix init to match type where applicable
- A few other minor updates

I'd like to get some feedback and I'll create an issue and fill out the launch checklist after that.

Thanks! 

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


src/ui/map.ts
```diff
@@ -553,80 +552,76 @@ export class Map extends Camera {
...
+        this._refreshExpiredTiles = resolvedOptions.refreshExpiredTiles === true;
+        this._fadeDuration = resolvedOptions.fadeDuration;
+        this._crossSourceCollisions = resolvedOptions.crossSourceCollisions === true;
+        this._collectResourceTiming = resolvedOptions.collectResourceTiming === true;
+        this._locale = {...defaultLocale, ...resolvedOptions.locale};
+        this._clickTolerance = resolvedOptions.clickTolerance;
+        this._overridePixelRatio = resolvedOptions.pixelRatio;
+        this._maxCanvasSize = resolvedOptions.maxCanvasSize;
+        this.transformCameraUpdate = resolvedOptions.transformCameraUpdate;
+        this.cancelPendingTileRequestsWhileZooming = resolvedOptions.cancelPendingTileRequestsWhileZooming === true;
```
HarelM, line 593: Why this `=== true` is needed?

HarelM: I see it in a few more lines.

vicb: You're right that's is not needed here.

But for i.e. `failIfMajorPerformanceCaveat?: boolean;`, `_failIfMajorPerformanceCaveat: boolean;` is not expected to be undefined so `=== true` would transofm `undefined` to `false`.

I'll update to remove the place where it's not required.

src/types/util.ts
```diff
@@ -0,0 +1,25 @@
...
+ *  bar: number | undefined;
+ *  baz: number | undefined;
+ * }
+ *
+ * ```
+ *
+ * See https://medium.com/terria/typescript-transforming-optional-properties-to-required-properties-that-may-be-undefined-7482cb4e1585
+ */
+
+export type Complete<T> = {
```
HarelM, line 22: I was actually thinking of removing this folder entirely, and move the types to either the utils folder or the files that are using those.
I also don't see this is used in another place, so I'm not sure there's an actual need to move it...

HarelM:
I don't have any objections to this PR in general.
Can we close the followin PR:
- #4003?

src/types/util.ts
vicb: I prefer not to have too much noise in `map.ts` but no strong opinion, I can do whatever you think is best.

One other unrelated question, do you know why `const defaultMinZoom = -2;`?
I would love to add a comment there because it doesn't look intuitive

vicb:
> I don't have any objections to this PR in general. Can we close the following PR:
> 
> * [Fix MapOptions#style is optional in the Map constructor #4003](https://github.com/maplibre/maplibre-gl-js/pull/4003)?

For now `style` is not addressed in this PR but I think it should be.
Let's close the other PR once this has been updated.
The first commit is only about cleaning the existing stuff. I'll add more commits to improve the type after we agree on the first one.

src/types/util.ts
vicb: -2 is to render maps < 512px (see https://github.com/mapbox/mapbox-gl-js/pull/9028)

vicb:
It looks like some tests are failing...
I'll try to reproduce locally and fix

HarelM:
I'm not sure the failing test is your fault. I see it on a PR for a package update.
Wait a bit with the investigation...

src/ui/map.ts
```diff
@@ -1727,28 +1721,30 @@ export class Map extends Camera {
...
-            this.style.once('style.load', () => this._updateStyle(style, options));
-            return;
-        }
-
-        const previousStyle = this.style && options.transformStyle ? this.style.serialize() : undefined;
+    _updateStyle(style: StyleSpecification | string | null | undefined, options?: StyleSwapOptions & StyleOptions) {
+        let previousStyle: StyleSpecification | undefined;
         if (this.style) {
-            this.style.setEventedParent(null);
+            // transformStyle relies on having previous style serialized.
```
HarelM, line 1727: Is this a bug fix? If so, I think it will be better to have this in a different PR...?

src/ui/map.ts
```diff
@@ -2181,8 +2172,8 @@ export class Map extends Camera {
      * let coffeeShopIcon = map.getImage("coffee_cup");
      * ```
      */
-    getImage(id: string): StyleImage {
-        return this.style.getImage(id);
+    getImage(id: string): StyleImage | undefined {
+        return this.style?.getImage(id);
```
HarelM, line 2176: Isn't this a behavior change? As far as I understand, before this change if style was not defined, this method would have thrown and exception, wouldn't it?

HarelM:
I believe this PR got too big... :-(
Let's try and keep the PRs small, with small increments so these increments can be merged.
From previous commit to latest commit there are a lot of changes which I think should be discussed before merging.
The previous commit was great in terms of adding correctness to types without changing any implementation (maybe besides moving the initialization out of the constructor).
Let's start with that and then move on to fixing things/changing logic.

src/ui/map.ts
```diff
@@ -298,14 +301,15 @@ export type MapOptions = {
      * the schema described in the [MapLibre Style Specification](https://maplibre.org/maplibre-style-spec/),
      * or a URL to such JSON.
      */
-    style: StyleSpecification | string;
+    style?: StyleSpecification | string;
```
HarelM, line 304: I think that other improvements are great, I would avoid changing this as part of this PR, keep it as is for now and open a different PR just for this change (with its impact obviously).

src/ui/map.ts
vicb: This is only code factorization:
- `this.style` is tested on l 1732, 1737, and 1738 on the left
- `options.transformStyle` is tested l 1732 and 1737 on the left

src/ui/map.ts
vicb: Discussed offline with @HarelM 

Summary of the discussion: What is the expected behavior of this function:

- throw is not documented
- the type of `this.style: Style` does not let users think it might throw
- is it valuable to throw here?

I think the behavior should be consistent across methods.

Some might throw:

```ts
    getImage(id: string): StyleImage {
        return this.style.getImage(id);
    }
```

Some are noops when `style` is undefined

```ts
    redraw(): this {
        if (this.style) {
            // cancel the scheduled update
            if (this._frameRequest) {
                this._frameRequest.abort();
                this._frameRequest = null;
            }
            this._render(0);
        }
        return this;
    }
```

Some warn when `style` is undefined

```ts
   _updateDiff(style: StyleSpecification, options?: StyleSwapOptions & StyleOptions) {
        try {
            if (this.style.setState(style, options)) {
                this._update(true);
            }
        } catch (e) {
            warnOnce(
                `Unable to perform style diff: ${e.message || e.error || e}.  Rebuilding the style from scratch.`
            );
            this._updateStyle(style, options);
        }
    }
```

vicb (commented):
@HarelM 

Maybe the best way to move forward is to remove the last commit from this PR.

I can then submit another PR making `MapOptions.style` optional and dropping all the other changes in that last commit as I don't think you find them useful?

HarelM:
Yes, let's drop the last commit and open an issue about the inconsistency of the methods.
Once we agree on the way forward in the issue we can continue to open a PR with the agreed upon changes.
It's not that I don't think they are valuable, it's just that they need a bit more discussion as some of them are breaking changes to public API methods in terms of behavior and I would like to make sure we discuss them properly.
Generally speaking, I agree that the current API is not consistent, and we should dive into it and see what improvements are considered low hanging fruits and what will need a breaking change as part of a breaking change version (see #3834) , but I believe it is out of scope for this PR.
In general, I would prefer to better document the current behavior in types and method docs as an initial non-breaking-change manner, and then later on, if needed, change the behavior as part of a breaking change version.

HarelM (approved):
THANKS!

vicb:
Thanks for your help Harel!

PR #4146, opened 2024-05-21
Globe - clipping fix

kubapelc:
I've tested globe on several phones and discovered a bug on one of them (Huawei/HiSilicon), where backfacing side of the globe was being drawn, even though it should be occluded. This makes globe unusable.

Normally globe uses the clipping step of the GL pipeline to cull the backfacing side of the planet by setting a custom Z value to all vertices, which works, because MapLibre then clamps Z values to a specific value using `glDepthRange`, so the Z values can be anything I want for clipping. I used this approach instead of regular backface culling, because MapLibre's winding order was very inconsistent. But I eventually fixed that while polishing the code for PRs.

I think the phone's GPU (or driver) is bugged and does the depth clamping *first*, and only then it does the clipping, resulting in all geometry passing the clipping step unmodified, since the Z values were already overwritten.

The workaround is to enable backface culling for every layer where it is possible (since now the winding order is consistent across most layers), and implement software clipping for the rest.

This PR also modified terrain, since terrain used to use a mirrored projection matrix, thus flipping the winding order of all triangles. This is fixed by using an unflipped projection matrix, thus ensuring the winding order is consistent for both regular rendering and terrain's render-to-texture.

I also fixed a bug related to images being placed on a globe, where they would sometimes extend all the way to the poles.

Demo of the bug

The screenshot is not from the phone itself, but this is what the bug looks like.


src/gl/cull_face_mode.ts
```diff
@@ -21,7 +22,13 @@ export class CullFaceMode {
...
+
+    /**
+     * Opposite of {@link backCCW}. Culls front-facing triangles when counterclockwise vertex order is used.
+     */
+    static frontCCW: Readonly<CullFaceMode>;
 }
 
 CullFaceMode.disabled = new CullFaceMode(false, BACK, CCW);
 CullFaceMode.backCCW = new CullFaceMode(true, BACK, CCW);
+CullFaceMode.frontCCW = new CullFaceMode(true, FRONT, CCW);
```
HarelM, line 34: There were some places that we changed `disabled` to `backCCW`. Do we still need disabled after this PR?

src/source/image_source.ts
```diff
@@ -98,6 +98,7 @@ export class ImageSource extends Evented implements Source {
     image: HTMLImageElement | ImageBitmap;
     tileID: CanonicalTileID;
     tileCoords: Array<Point>;
+    flippedWindingOrder: boolean = false;
```
HarelM, line 101: Where is this being used? I don't think I saw a usage for this...?

HarelM: I now see, can we fix the coords instead of adding this flag?

kubapelc: This is set in image_source.ts line 224:

```ts
// Transform the corner coordinates into the coordinate space of our
// tile.
this.tileCoords = cornerCoords.map((coord) => this.tileID.getTilePoint(coord)._round());
this.flippedWindingOrder = hasWrongWindingOrder(this.tileCoords);
```

...and then used during rendering in draw_raster.ts line 46:

```ts
if (source instanceof ImageSource) {
    // Image source - no stencil is used
    drawTiles(painter, sourceCache, layer, tileIDs, null, false, false, source.tileCoords, source.flippedWindingOrder);
} else (...)
```

src/gl/cull_face_mode.ts
kubapelc: Yes, `disabled` is still used in many places.

src/shaders/circle.vertex.glsl
```diff
@@ -74,6 +74,15 @@ void main(void) {
     } else {
         gl_Position = projectTileWithElevation(circle_center, ele);
 
+        if (gl_Position.z / gl_Position.w > 1.0) {
+            // Same as in fill_outline.fragment.glsl and line.fragment.glsl, we need to account for some hardware
+            // doing glFragDepth and clipping in the wrong order by doing clipping manually in the shader.
+            // For screenspace (not u_pitch_with_map) circles, it is enough to detect whether the anchor
+            // point should be clipped here in the vertex shader, and clip it by moving in beyond the
+            // renderable range in X and Y.
+            gl_Position.xy = vec2(10000.0);
```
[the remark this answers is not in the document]
kubapelc: I've updated the comment above. Since `gl_Position` is later modified by adding vertex offsets to it, this just ensures that even after the addition (which is likely to be in the range -1..1) the final `gl_Position` will still be beyond the renderable range (also -1..1) and thus all geometry gets clipped.

src/source/image_source.ts
kubapelc: > I now see, can we fix the coords instead of adding this flag?

I've tried that first and it didn't work, but maybe I switched the wrong coords...

kubapelc: I don't it is possible to fix this by swizzling the coords, changing the winding order with the coordinates necessarily means flipping the image in exactly one axis.

kubapelc: The reason is that texture coordinates are hardcoded to the given vertex indices. For examplet he first vertex will always display the image's top left regardless of what we use as its spacial coordinate. This means that we must keep the current order of coordinates for the image to be displayed properly.

Fixing the winding order either means setting face culling to FRONT, or using a different index buffer, or rewriting how texture coordinates are handled in the raster layer, which would either mean having a custom vertex buffer for every tile, or passing even more uniforms to the shader, both of which is more complex than just setting the cull face mode with a flag.

HarelM: But you are passing the coordinates with frontCCW basically, isn't it possible to "rotate" the coordinates and change drawing to backCCW.
Is this the only place that uses frontCCW?

HarelM:
Can you add a render/integration test for the image source? maybe I'll be able to better understand it this way...

kubapelc:
> Can you add a render/integration test for the image source? maybe I'll be able to better understand it this way...

`regressions/mapbox-gl-js#7271` tests that images render correctly when they traverse the antimeridian, which is exactly the case where the winding order gets broken. This test will also catch any image that gets mirrored.

src/source/image_source.ts
HarelM: So if I specify the coordinates in a wrong order I'll get the image flipped? weird.
What I'm trying to achieve is hopefully to remove the need for the frontCCW.
If that's too complicated, then ignore my comment.

kubapelc: Yes, I think it is too much work for very little benefit.

PR #4151, opened 2024-05-23
Fix MapOptions#style is optional in the Map constructor

vicb:
The use case is to delegate the style setting to a layer switcher control.

Note that is is already possible to set the style to null using `Map#setStyle(null)`. This PR updates the MapOptions interface to also allow that when a map instance is created.

Supersedes #4003

## Launch Checklist

 - [n/a] Link to related issues.
 - [n/a] Include before/after visuals or gifs if this PR includes visual changes.
 - [n/a] Write tests for all new functionality.
 - [n/a] Document any changes to public APIs - documentation is the type
 - [n/a] Post benchmark scores.
 - [n/a] Add an entry to `CHANGELOG.md` under the `## main` section.

### DIFF
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -299,8 +299,9 @@ export type MapOptions = {
      * The map's MapLibre style. This must be a JSON object conforming to
      * the schema described in the [MapLibre Style Specification](https://maplibre.org/maplibre-style-spec/),
      * or a URL to such JSON.
+     * When the style is not specified, calling {@link Map#setStyle} is required to render the map.
      */
-    style: StyleSpecification | string;
+    style?: StyleSpecification | string;
     /**
      * If `false`, the map's pitch (tilt) control with "drag to rotate" interaction will be disabled.
      * @defaultValue true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ### ✨ Features and improvements
 
 - Add events to `GeolocateControl` to allow a more granular interaction ([#3847](https://github.com/maplibre/maplibre-gl-js/pull/3847))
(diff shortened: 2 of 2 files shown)


HarelM:
Can you add a changelog item for that? Up until now, you needed to supply the style in typescript, but now it's not needed anymore, which is great as it aligns with the code.

src/ui/map.ts
```diff
@@ -300,7 +300,7 @@ export type MapOptions = {
      * the schema described in the [MapLibre Style Specification](https://maplibre.org/maplibre-style-spec/),
      * or a URL to such JSON.
      */
-    style: StyleSpecification | string;
+    style?: StyleSpecification | string;
```
HarelM, line 303: Please add to the tsdoc comment what will happen in case this is not provided.

vicb:
@HarelM I have addressed your feedback, let me know if this looks good. Thanks!

HarelM (approved):
Thanks!

src/ui/map.ts
```diff
@@ -299,8 +299,9 @@ export type MapOptions = {
      * The map's MapLibre style. This must be a JSON object conforming to
      * the schema described in the [MapLibre Style Specification](https://maplibre.org/maplibre-style-spec/),
      * or a URL to such JSON.
+     * When the style is not specified, calling to `Map.setStyle` is required to render the map.
```
HarelM, line 302: Do you want to use a tsdoc type of comment? I.e. `{link Map#setStyle}`

src/ui/map.ts
vicb: Great idea, done!

src/ui/map.ts
vicb: Great idea, done!

vicb:
Thanks!

Issue #4198, opened 2024-05-29
Feature Properties Transform


wipfli:
The things that a user can do with MapLibre style expressions based on feature properties is limited to the implemented styling language. While the style language has support for basic math, logic, and rudimentary string operations, many things cannot be done easily.

For example, it is hard to format a number to the locale India with 3 significant digits, e.g.: turn 1434389.043 to 14,30,000. 

Another example: it is not possible to [extract a value from a JSON](https://github.com/maplibre/maplibre-gl-js/discussions/4004) encoded dictionary.

Both of these tasks are trivial to achieve in JavaScript or any other programming language and we usually motivate users to do such transformation at tile generation time.

In the past, we rejected a [proposal to reference user-defined javascript functions](https://github.com/maplibre/maplibre-gl-js/issues/1295) from the style.json.

What I would like to introduce instead is a user-defined Feature Properties Transform, a callback which gets executed during vector tile parsing.

The Feature Properties Transform gets executed on every feature and has access to the following data:

- source name
- source layer name
- tileID z/x/y
- feature properties

With this information, one can easily filter based on source and source-layer, like people do usually in style.json documents and since the transform has access to all feature properties, it can do arbitrary logic and execute javascript on the properties dictionary. This is very powerful and opens the way to an alternative way of styling maps, namely more styling in JavaScript and less in the styling language.

Proof-of-concept implementation: #4199

HarelM:
I think this is a very good idea for a hook.
We need to remember that allowing things to be only in javascript will cause issues with parity between native and web.
This is true for add protocol as well, but addProtocol forces you to put something "Wrong" in the style as opposed to this feature which you might just get an empty label and this can "fail" siliently.

In any case, these hooks are important and people using them should know that they create issues with parity.

One thing to keep in mind looking at the PR is that the transform is done in the worker thread while the map API is available in the main thread, and passing a method to the worker is very hard (you can probably pass very basic method, without the context of the main thread).
So there needs to be a way to pass a method to the worker, which is only using importScript, which needs a javascript file or a string converted to dataURI.

In any case, I think adding that hook, since it is in the worker can be done differently maybe, using some "noop" method you override in the worker tile class - i.e something like

```js
class WorkerTile {
...
    transfromFeature(feature) = return feature // this is a new noop method

}
```

and call this code in the worker context:
```js
WorkerTile.prototype.transfromFeature = (f) => {//do something else with f and return it }
```

[...]

PR #4199, opened 2024-05-29
Add Feature Properties Transform

wipfli:
Proof-of-concept implementation for #4198. The implantation is inspired the `addProtocol` implementation.

Demo: 

https://github.com/wipfli/maplibre-feature-properties-transform-example

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


wipfli:
This is ready for review. Let me know what needs to be changed

src/source/protocol_crud.ts
```diff
@@ -46,3 +46,11 @@ export function addProtocol(customProtocol: string, loadFn: AddProtocolAction) {
 export function removeProtocol(customProtocol: string) {
     delete config.REGISTERED_PROTOCOLS[customProtocol];
 }
+
+export function getFeaturePropertiesTransform() {
```
HarelM, line 50: I would consider moving this to a different file or change the file name.
Also, l don't think there's a need to use the config object as this is not configurable from the main API, so simply use an object defined in a new file. there are some examples to this pattern in the source code.

src/source/worker.ts
```diff
@@ -76,6 +76,9 @@ export default class Worker {
         this.self.addProtocol = addProtocol;
         this.self.removeProtocol = removeProtocol;
 
+        this.self.getFeaturePropertiesTransform = getFeaturePropertiesTransform;
```
HarelM, line 79: Where is this method used in terms of "public" API? I'm not sure get... is needed to be registered as part of the worker code.

src/source/worker_tile.ts
```diff
@@ -81,6 +83,7 @@ export class WorkerTile {
         };
 
         const layerFamilies = layerIndex.familiesBySource[this.source];
+        const featurePropertiesTransform = getFeaturePropertiesTransform();
```
HarelM, line 86: Have you checked what I suggested about abstract method of the worer_tile class? Is it less readable?

src/util/config.ts
```diff
@@ -7,6 +7,8 @@ import type {RequestParameters, GetResourceResponse} from './ajax';
  */
 export type AddProtocolAction = (requestParameters: RequestParameters, abortController: AbortController) => Promise<GetResourceResponse<any>>
 
+export type FeaturePropertiesTransform = (source: string, sourceLayer: string, tileID: string, geometryType: string, featureID: number, properties: { [_: string]: string | number | boolean }) => Promise<void>
```
HarelM, line 10: This is not an extendable API - please define an object like `FeaturePropertiesTransformInput` and define all the fields in it, and have the method receive it as the only argument.
Please also add TSDoc comments to better explain how to use this transform, this is probably the only place a user will be able to read about it.

src/util/config.ts
```diff
@@ -17,6 +19,7 @@ type Config = {
     MAX_PARALLEL_IMAGE_REQUESTS_PER_FRAME: number;
     MAX_TILE_CACHE_ZOOM_LEVELS: number;
     REGISTERED_PROTOCOLS: {[x: string]: AddProtocolAction };
+    FEATURE_PROPERTIES_TRANSFORM: FeaturePropertiesTransform;
```
HarelM, line 22: Same as above...

HarelM:
Thanks again for the PR!
Can you please resolve conflicts and also create some tests and an example page of how to use this feature?

src/source/worker_tile.ts
wipfli: I am not sure how one would do the pattern you suggested here. Can you elaborate?

src/source/worker.ts
wipfli: The `getFeaturePropertiesTransform` function is needed from a script that is loaded with `maplibregl.importScriptInWorker`. In that scope, `self` refers to the worker global scope and therefore the user-provided script gets access to set the transform. If you know a better way of doing this let me know.

src/source/protocol_crud.ts
wipfli: Moved

src/util/config.ts
wipfli: Created a `FeaturePropertiesTransformOption` type

wipfli:
For the tests I will need some help. Do we already have tests for `maplibregl.importScriptInWorker()`?

HarelM:
Import script requires integration tests, so I wouldn't start with that. 
Try unit testing the worker and the tile parser by mocking and spying a lot if the stuff.
Also, don't forget the changelog entry.

src/source/worker_tile.ts
HarelM: You should be able to do it using JavaScript's "prototype" reserved keyword, I think...
I'll try to see if I can find the time to experiment...

HarelM: Check out the following:
https://github.com/wipfli/maplibre-gl-js/compare/vector-tile-transform...HarelM:maplibre-gl-js-transform-props:vector-tile-transform

I'm not sure it's better in terms of API/readability... IDK...

src/source/worker.ts
HarelM: The public API has a `setFeaturePropertiesTransform`, in order to set the transfrom.
What would a user do with this `getFeaturePropertiesTransform` method? `hasFeaturePropertiesTransform` makes more sense to be, but also seems a bit redundant.
I would like to keep the public API changes to minimum to reduce maintenance efforts and backward compatibility requirements...

src/source/worker_tile.ts
wipfli: Thanks for sharing. From the user perspective it looks the same I think. Happy to merge if you find this more maintainable. Up to you...

HarelM: I don't think I do, practically the same, but feels more hackish to me, what do you think?

wipfli: I prefer the API which is more similar to the already existing self.addProtocol version, i.e., the version proposed here in this PR.

src/source/worker.ts
wipfli: The getter indeed seems redundant. Maybe we could store the featurePropertiesTransform directly on the worker global scope instance assuming that there is just one. At the moment the variable is stored in the feature_properties_transform.ts file.

HarelM: Hmm, maybe we can keep the public API, and add a static method to workertile?

HarelM:
I've opened the following PR:
https://github.com/wipfli/maplibre-gl-js/pull/341
Hope it helps...

HarelM:
The image is empty, I think it's a bug in chrome or headless chrome... not sure.
I think the following arguments may help:
https://github.com/maplibre/maplibre-gl-js/pull/4299/files

Or you can toggle headless for image creation.

src/util/web_worker.ts
```diff
@@ -8,6 +9,8 @@ export interface WorkerGlobalScopeInterface {
     registerRTLTextPlugin: (_: any) => void;
     addProtocol: (customProtocol: string, loadFn: AddProtocolAction) => void;
     removeProtocol: (customProtocol: string) => void;
+    getFeaturePropertiesTransform: () => FeaturePropertiesTransform;
```
HarelM, line 12: I think get is not needed anymore, right?

test/unit/lib/web_worker_mock.ts
```diff
@@ -11,6 +11,8 @@ export class MessageBus implements WorkerGlobalScopeInterface, ActorTarget {
     registerRTLTextPlugin: any;
     addProtocol: any;
     removeProtocol: any;
+    getFeaturePropertiesTransform: any;
```
HarelM, line 14: Same here

HarelM:
Now you have a failing test that reproduces what I said :-)

louwers (commented):
I didn't know that you could send some script to be executed to a worker like this! I am intrigued.

Here it says that after using `createObjectURL`, the URL needs to be cleaned up again with `revokeObjectURL ` to avoid a memory leak: https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL_static

Since the script is only executed once, I think MapLibre GL JS should take care of this cleanup after the script is executed.

Maybe `importScriptInWorkers` can be updated to also take a function with `(self: ServiceWorkerGlobalScope) => void` as prototype. This has the advantage that instead of writing the script as a string, you get type and syntax checking. Also, you don't need to repeat yourself to prepare the data URL and the cleanup can be handled automatically.

If a function is passed to `importScriptInWorkers`, you would call `toString()`, strip the resulting string to get the function body, and then use `decodeURIComponent`, `Blob` and `URL.createObjectURL` internally.

Modify the `MessageType.importScript` to pass along an optional boolean to indicate that `revokeObjectURL` needs to be called on the worker side once the script has been executed.

HarelM:
The example is a way to inline everything into a single file for demo purposes, I would not advise to use `createObjectURL` in production, but instead use a javascript file stored next to the code or even on a CDN. This file can be created using the regular typescript tooling to support type check and everithing.
I wouldn't like to complicate the import script interface for this edge case. 
This is also outside the scope of this PR I believe.

louwers:
Ah OK, I thought that was the recommended approach. Makes sense. 👍

wipfli:
The documentation for queryRenderedFeatures says that the result "contains the properties of its source feature":

https://github.com/maplibre/maplibre-gl-js/blob/a9420817af8507a9884bc4bfbb4318cc70f3761e/src/ui/map.ts#L1519

Based on this I would suggest to add a note to the documentation of the setFeaturePropertiesTransform function that changes will not be reflected in queryRenderedFeatures. For tests I would suggest to write some render tests. What do you think?

HarelM:
I think that you'll have a hard time creating render tests for this feature as render tests are mainly style driven and this feature "breaks" the style. 
Browser integration tests are enough. 

Can you elaborate on the limitation of query render features? If this hook is creating a discrepancy between the rendered results and the returned value of the query it might confusing to users, wouldn't it?

wipfli:
The functionality proposed in here seems to have a lot of overlap with feature-state. It is just global on all features and not on individual features. I will think a bit if there is a better api than what is proposed here...

HarelM:
@wipfli is this still something you want to push forward?
Are any breaking changes needed while we are at version 5 pre-release for this feature?

HarelM:
Yes, I believe the issue was around `queryRenderedFeatures`.
Thanks for the update.

neodescis:
Closing in favor of https://github.com/maplibre/maplibre-gl-js/pull/5370

PR #4205, opened 2024-05-30
Fix a few minor accessibility issues with examples

coliff:
- Link labels to inputs
- Add missing title and viewport to GL Stats page

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/test/bench/gl-stats.html
+++ b/test/bench/gl-stats.html
@@ -2,7 +2,9 @@
 <html lang="en">
 <head>
 <meta charset="utf-8">
+<meta name="viewport" content="width=device-width, initial-scale=1">
 <link rel="icon" href="about:blank">
+<title>GL Stats</title>
 <script>
 // stub performance.now for deterministic rendering per-frame;
 // we'll later increment the value by 16 on every frame (16ms per frame for 60fps)
--- a/test/examples/filter-within-layer.html
+++ b/test/examples/filter-within-layer.html
@@ -95,7 +95,8 @@
 
             <fieldset>
                 <legend>🌊 Tsunami (0 or 1)</legend>
-                <input id="tsunami" type="checkbox" /> <label>Apply <code>tsunami</code> filter</label>
+                <input id="tsunami" type="checkbox" />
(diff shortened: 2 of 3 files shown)


HarelM (approved):
Gl stats is internal for testing, but I don't mind having it improved.

coliff:
I was going to suggest adding a HTMLHint test to make sure there are no bugs with all the HTML pages in the repo. For now you can run:

`npx htmlhint **/*.html`

for a `.htmlhintrc` config in the root with the following config:

```json
{
  "alt-require": true,
  "attr-lowercase": true,
  "attr-no-duplication": true,
  "attr-unsafe-chars": true,
  "attr-value-double-quotes": false,
  "attr-value-not-empty": false,
  "doctype-first": false,
  "doctype-html5": true,
  "head-script-disabled": false,
  "href-abs-or-rel": false,
  "html-lang-require": true,
  "id-class-ad-disabled": true,
  "id-class-value": false,
  "id-unique": true,
  "inline-script-disabled": false,
  "inline-style-disabled": false,
  "spec-char-escape": false,
  "src-not-empty": true,
  "style-disabled": false,
  "tag-pair": false,
  "tag-self-close": false,
  "tagname-lowercase": true,
  "title-require": true
}
```

HarelM:
But we have eslint with the html plugin already running... Are there rules there we can add?

coliff:
> But we have eslint with the html plugin already running... Are there rules there we can add?

I'll take a look!

coliff:
@HarelM - re: eslint with the html plugin already running

https://www.npmjs.com/package/eslint-plugin-html
> This plugin focuses on applying ESLint rules on inline scripts contained in HTML. It does not provide any rule related to HTML. 

So maybe need to add this too: https://html-eslint.org/

HarelM:
Feel free to add this. although, all these plugins are a hazard when eslint is releasing a breaking change version like they did with 9.0, as these tend to not be maintained, as start to fail.
But we can always remove them when they get stale I guess...

PR #4210, opened 2024-05-31
Fix accept header overwrite

CalRobert:
## Launch Checklist

Currently, when a user specifies an Accept header in a custom RequestTransformer, that header is overwritten with "application/json" when getJson is called in ajax.ts. This fixes that so the user's custom-set header is preserved. This is particularly helpful when using [Postgrest](https://postgrest.org/) which will automatically respond with Geojson when "application/geo+json" is the Accept header

 - [N/A] Link to related issues.
 - [N/A] Include before/after visuals or gifs if this PR includes visual changes.
 - [N/A] Document any changes to public APIs.
 - [?] Post benchmark scores.


HarelM:
Thanks! Looks good! Can you please add a changelog item and fix lint and tests?

CalRobert:
> Thanks! Looks good! Can you please add a changelog item and fix lint and tests?

Thanks, updated the logic so we also don't overwrite the header in makeXMLHttpRequest and added changelog item, and linted

src/util/ajax.ts
```diff
@@ -187,7 +188,10 @@ function makeXMLHttpRequest(requestParameters: RequestParameters, abortControlle
         }
         if (requestParameters.type === 'json') {
             xhr.responseType = 'text';
-            xhr.setRequestHeader('Accept', 'application/json');
+            // Do not overwrite the user-provided Accept header
+            if (!requestParameters.headers?.Accept) {
```
CalRobert, line 192: As before, I would expect that we don't want to overwrite the user's specified Accept header, but I would be curious if the reviewer has thoughts on this. I changed this because in testing this results in an Accept header that takes  for application/json and application/geo+json. I suppose that _could_ be desired, though.

(Without this line, we get:
Expected: "application/geo+json"
Received: "application/geo+json, application/json"
)

from the test case

test/build/min.test.ts
```diff
@@ -36,7 +36,7 @@ describe('test min build', () => {
         const decreaseQuota = 4096;
 
         // feel free to update this value after you've checked that it has changed on purpose :-)
-        const expectedBytes = 791799;
+        const expectedBytes = 792823;
```
CalRobert, line 39: I believe this is a result of other changes in `main` as well but someone who commits more regularly may have thoughts here

HarelM: Probably.

HarelM:
Are you sure about the tests' coverage? It reports that the coverage of the changes is not 100%...?

CalRobert:
> Are you sure about the tests' coverage? It reports that the coverage of the changes is not 100%...?

Thanks - I'm not completely sure but I think the post from codecov was before my later commits (code coverage was why I added the `file:///` tests actually). Going here it looks like coverage is up by 0.10% now with this PR, if I am reading it right.

https://app.codecov.io/gh/maplibre/maplibre-gl-js/commit/2eb4382583ce5d4eb2bfff35efd3ba2dfee1d5ca

HarelM (approved):
Thanks!

CalRobert:
To be honest I'm not sure how to see which lines are offending w.r.t. coverage - https://app.codecov.io/gh/maplibre/maplibre-gl-js/commit/2eb4382583ce5d4eb2bfff35efd3ba2dfee1d5ca/blob/src/util/ajax.ts says it can't read the file. Can I run this locally?

CalRobert:
NM looks like you merged it, thanks!

HarelM:
You should be able to run locally with npm run test-unit -- --coverage
I think.

Issue #4024, opened 2024-04-23
on load method： Occasionally the load method does not trigger a callback


qlanto224:
**maplibre-gl-js version**:4.1.3 and 2.4.0

**browser**:

### Steps to Trigger Behavior
1. open https://codepen.io/oberhamsi/pen/ZENLQEN
2. Look in the console for the loaded text - no loaded text meaning no load event and map is no loaded. 

### Link to Demonstration

https://codepen.io/oberhamsi/pen/ZENLQEN

### Expected Behavior
Loaded event is fired or the map is loaded

### Actual Behavior
Niether

qlanto224:
when I click map ， the load method trigger

HarelM:
Please provide a jsbin with a reproduction.

qlanto224:
> Please provide a jsbin with a reproduction.

sorry， I cant provide this style.  
I found that when the requested layer data is too large, the request time is more than 10 seconds, it will not be called.
I'm not sure if there is a timeout inside 'load'.

HarelM:
I'm not aware of timeout definition. 
Also, without a reproduction I can't help. 
You don't have to provide your style, but rather something that reproduces the issue...

qlanto224:
```js
import { Map } from 'maplibre-gl';
import { useEffect, useRef } from 'react';
export default () => {
  const mapRef = useRef<any>(null);
  useEffect(() => {
    const style ;// this style have some slow-requested data sources
    let map = new Map({
      container: mapRef.current,
      antialias: true,
      maxPitch: style.maxPitch ? style.mapPitch : 72,
      minZoom: style.minZoom,
      hash: false,
      attributionControl: false,
      maplibreLogo: true,
      logoPosition: 'bottom-right',
      style: style,
    });
    document
      .getElementsByClassName('maplibregl-ctrl-logo')[0]
      .removeAttribute('href');
    map.setRenderWorldCopies(false);
    map.on('load', () => {
      console.log('load finish');
    });
  }, []);
  return <div style={{ width: '100%', height: '50vh' }} ref={mapRef}></div>;
};
```
example：
![image](https://github.com/maplibre/maplibre-gl-js/assets/58463221/b23f2e74-7cf3-4d25-a6b1-5dc98dd43bb3)

qlanto224:
> I'm not aware of timeout definition. Also, without a reproduction I can't help. You don't have to provide your style, but rather something that reproduces the issue...

[...]

PR #3615, opened 2024-01-24
Change the coverage for jest to be aligned with render tests

HarelM:
## Launch Checklist
Related to:
- #3598

Related to this discussion https://github.com/cenfun/monocart-coverage-reports/issues/4

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.
 - [x] Link to related issues.
 - [x] Include before/after visuals or gifs if this PR includes visual changes.

PR #3598, opened 2024-01-21
Replace v8 with monocart for coverage

HarelM:
## Launch Checklist

In order to solve the coverage issues introduced with v8 and the render test a different coverage tool is needed apparently.
See here:
https://github.com/cenfun/monocart-coverage-reports/issues/4
And here:
https://github.com/istanbuljs/v8-to-istanbul/issues/238

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.
 - [x] Link to related issues.
 - [ ] Include before/after visuals or gifs if this PR includes visual changes.

PR #4211, opened 2024-05-31
ajaxerror not serialized properly

oberhamsi:
fixes https://github.com/maplibre/maplibre-gl-js/issues/4024

the original cause for `load` not firing was that `source_cache` did not see the 404 returned by a failed tile request here https://github.com/maplibre/maplibre-gl-js/blob/main/src/source/source_cache.ts#L172

this was caused by AjaxError no longer being serialized since commit https://github.com/maplibre/maplibre-gl-js/commit/d68b919b04864d1c11491ddbc75d3f16dc2838e0#diff-0c7952cb97b3abe689d267fa5663ac8d387fbd6a637905cce9581a5cc2f1ff1c AjaxError is registered for transfer serialization but it's also an instance of Error and was thus just returned

i think the worker_transfer tests are better if they use `structuredClone`. this works for the other tests too but not for the once check against `Blob` - seems to be a bug with jsdoms custom `Blob` in combination with nodes native `structuredClone`. i'm hoping this will be fixed with https://github.com/jsdom/jsdom/issues/3363

if you don't like the rather big change to the tests i can revert this and not do any tests for this issue.

## Launch Checklist

 - [x] Link to related issues.
 - [x ] Write tests for all new functionality.


oberhamsi:
i had so much trouble with jsdom+structuredclone that i wrote a quick demo how it really behaves in browsers, might be helpful to understand what i did https://codepen.io/oberhamsi/pen/VwOpYqz

src/util/web_worker_transfer.ts
```diff
@@ -114,8 +114,7 @@ export function serialize(input: unknown, transferables?: Array<Transferable> |
         input instanceof String ||
         input instanceof Date ||
         input instanceof RegExp ||
-        input instanceof Blob ||
-        input instanceof Error) {
+        input instanceof Blob) {
```
HarelM, line 117: This was added to allow sending errors instead of stringify those from the RTL plugin.
I hope there's a better way to solve this issue...
I believe the failing test caught it.

jest.config.ts
```diff
@@ -35,7 +35,7 @@ const config: Config = {
     projects: [
         {
             displayName: 'unit',
-            testEnvironment: 'jsdom',
+            testEnvironment: './jest.jsdom.environment.ts',
```
HarelM, line 38: Can you use a setupfile instead, or add it to the `web_worker_mock.ts` setup file?

HarelM:
Thanks for taking the time to look into this!

oberhamsi:
thanks for the quick feedback! i'll keep working on it

src/util/web_worker_transfer.ts
oberhamsi: yes, i noticed other errors not going throw when testing it in the browsers.

i'm thinking keeping Error here but also checking whether`klass` is in the registry before bailing out so soon. might be a good solution.

jest.config.ts
oberhamsi: i would like to but i don't see how i can access the node global `structuredClone` in the setup function. i found the environment workaround here https://github.com/jsdom/jsdom/issues/3363

another solution that would work in the setupfile would be to require a structuredClone polyfill

oberhamsi:
i'll look into the tests. it's not due to the environment :/

src/util/web_worker_transfer.ts
HarelM: I believe it is a better direction.

jest.config.ts
HarelM: I think you can simply use global.sturctureClone = something, can't you?

oberhamsi: i'm not sure i understand - in the jest setupfile i can't map the existing node global `structuredClone` to be the missing jsdom environment global `structuredClone`. afaict this kind of mapping of node functions into the jsdom env is only possible when creating the environment.

src/util/web_worker_transfer.ts
```diff
@@ -91,6 +91,20 @@ function isArrayBuffer(value: any): value is ArrayBuffer {
...
 
+function isRegistered(input: unknown) {
+    if (input === null || typeof input !== 'object') {
+        return false;
+    }
+    const klass = (input.constructor as any);
+    if (klass._classRegistryKey && klass._classRegistryKey !== 'Object') {
+        return true;
+    }
+    if ((<SerializedObject>input).$name) {
```
HarelM, line 102:
```suggestion
    if ((input as SerializedObject).$name) {
```

HarelM, line 102:
```suggestion
    if ((input as SerializedObject).$name) {
```

src/util/web_worker_transfer.test.ts
```diff
@@ -1,5 +1,11 @@
+import {AJAXError} from './ajax';
 import {register, serialize, deserialize} from './web_worker_transfer';
 
+const mockTransfer = (input, transferables?) => {
+    const serialized = serialize(input, transferables);
+    return deserialize(structuredClone(serialized, {transfer: transferables}));
```
HarelM, line 6: Is this the only place we need the sturctureClone in jest?

src/util/web_worker_transfer.ts
```diff
@@ -91,6 +91,20 @@ function isArrayBuffer(value: any): value is ArrayBuffer {
            (value instanceof ArrayBuffer || (value.constructor && value.constructor.name === 'ArrayBuffer'));
 }
 
+function isRegistered(input: unknown) {
```
HarelM, line 94: Can we use this method elsewhere? Feels like this code is somewhat duplicated from another place...

src/util/web_worker_transfer.ts
```diff
@@ -104,7 +118,8 @@ function isArrayBuffer(value: any): value is ArrayBuffer {
  * this should happen in the client code, before using serialize().)
  */
 export function serialize(input: unknown, transferables?: Array<Transferable> | null): Serialized {
-    if (input === null ||
+    if (!isRegistered(input) && (
```
HarelM, line 121: This condition is duplicated between the serialize and deserialize, isn't it?

jest.config.ts
HarelM: I'm not sure I fully understand the need to add this complexity to jest test setup, can you clarify what is the issue with structure clone?

src/util/web_worker_transfer.ts
oberhamsi: i see, yes. i will provide a more general check "(de)serialize as built-in?" function and use it in both places.

src/util/web_worker_transfer.ts
HarelM: as in here - seems similar...
https://github.com/maplibre/maplibre-gl-js/blob/d6c5c4f3f5ce6c6bf0e3ec43cce79d69bef6acfb/src/util/web_worker_transfer.ts#L159

jest.config.ts
oberhamsi: `structuredClone` is not yet available in jsdom (see jest ticket linked above). the recommended workaround is to use node's structuredClone in the jsdom environment.

using structuredClone is a good way to demonstrate the issue with Error since cloning an object with `structuredClone` is equivalent to what happens to objects passed through the messagebus e.g. from a webworker to the main thread. 

structuredClone also supports transferables, which is why i changed the tests to actually test whether the objects are transfered.

i can remove structuredClone and the env changes altogether and remove the tests but i would say the worker_transfer tests are better when using structuredClone.

HarelM: What about using a polifill? The solution mentioned about ungap core-js. Is it possible and simpler? Or it's just an issue with jsdom and a polifill will not help?

oberhamsi: yes! i can try with a polyfill and use that in the setupFile

it can still be changed to standard structuredClone once jsdom supports it. all browsers already do

HarelM: Is it possible to do something like:
window.structureClone = global.structureClone? Wouldn't that solve this issue too?

oberhamsi: i will check again but i couldn't find a way to access the node globals in the setupfile

oberhamsi: the polyfill works for all but one test in web_transfer. the ArrayBuffer transfer can't be polyfilled by core-js:

`DataCloneError: Transferring of ArrayBuffer cannot be properly polyfilled in this engine`

using the core-js polyfill would mean:

- install core-js as dev depenceny
- `import 'core-js/stable/structured-clone'` in web_worker_transfer_test.ts
- undo custom jsdom environment

i can also undo my changes to the other tests in that file so they no longer use `structuredClone`.

i could also not use `structuredClone` in my new AjaxError test and instead check if serialize() adds the `$name` property.

may i ask what you don't like about the current solution? is it the new jsdom environment file? 

overall i think using `structuredClone` for the worker_transfer tests is a good solution. but maybe it's better to wait until jsdom supports it which would mean no custom-jsdom-env. https://github.com/jsdom/jsdom/issues/3363

HarelM: > may I ask what you don't like about the current solution? is it the new jsdom environment file?

Yes, I don't like the fact that we are customizing the environment, as this doesn't look like an out of the box solution and these things tend to break when you upgrade a version.
I'm concern that this will create maintenance overhead.

Up until now, when we had issues with jest we moved the relevant test to the integration tests where we run these tests in a browser (the failed test you encountered with RTL was an integration test).

So if we can leverage the strength of the integration tests instead of complicating the jest setup, I think it will be better.

The reason I'm saying this, is because when I started to take over this project there were tons of unmaintained packages and I don't want to get to this situation again. So a dependency is a last resort from my point of view, and also complicated test configurations.

I hope this makes my comments clearer and explain my point of view better.

oberhamsi: understood! thanks for elaborating. i will find a solution which does not modify the jest-setup. i'll look into how the integration tests work and if it fits in there.

oberhamsi: i have changed the test and removed structureClone and the custom env

src/util/web_worker_transfer.ts
oberhamsi: i have reworked both serialize & deserialize. i changed more then i wanted to but the tests run thought and imo it's now clearer to read.

HarelM:
The coverage of these methods wasn't great before you changed the code :-(
So the only way to know the logic wasn't changed is by reviewing the code manually, which isn't a great idea...
What are the odds you might be willing to fully cover this class with tests?
Sorry to nag, I'm just not sure from reading the code that it has the same logic...

oberhamsi:
i've added a few more tests and will keep looking into it.

thanks again for the feedback.

oberhamsi:
test coverage is now good but if you're not happy with the sweeping changes, what do you think about reverting the refactor? https://github.com/maplibre/maplibre-gl-js/pull/4211/commits/7fd8cab80b911c75533c557c8f00412d0ae1d0e7

you didn't like the duplication but they were (mostly) present before and i don't see how to make this significantly better without moving a few things, as i did. this is how it was before the refactor: https://github.com/orfon/maplibre-gl-js/blob/1374644017b439854a4bd1ab5246fe878fe27266/src/util/web_worker_transfer.ts

maybe I can do the refactor once structureClone arrives in jsdom. this would make me feel a lot better about the unit tests.

HarelM:
The coverage of the file is at about 78%, I would feel better if is was closer to 85-90%.
I think some error flows are needed to be checked.
Overall the changes look good and I think you did a good job of refactoring.
I reviewed the code and it seemed to me mostly the same, but still confusing to compare, especially the $name, constructor etc...
So I think a few more tests should be good and making sure they pass also before the changes in the file to make sure the logic is the same.

HarelM:
It would also be great if we could solve the following comment:
```
// (Temporary workaround) allow a class to provide static
        // `serialize()` and `deserialize()` methods to bypass the generic
        // approach.
        // This temporary workaround lets us use the generic serialization
        // approach for objects whose members include instances of dynamic
        // StructArray types. Once we refactor StructArray to be static,
        // we can remove this complexity.
```
I'm not sure how easy it would be to solve this though. StructArray are generated classes, so it might be easy to change the generation code? I don't know, might be out of scope for this PR.
I just want to pick your brain since you dug deep into this class.

oberhamsi:
alright! coverage is now high 90s. i think the bot didn't update the coverage properly 

Screenshot from 2024-06-18 09-49-00

oberhamsi:
regarding serialize. yeah, one thing i don't like about that logic is how implicit it is. imo `register()` should have an additional parameter `hasSerialize boolean`. as it is right now, someone can remove `serialize()` from the StructArray implementation and it would fail but not be obvious that the failure is related to web_transfer

the reverse is also true: the comment tells us only StructArray uses `serialize` - but is that true? (it is, but it requires manually checking all registered classes)

otherwise i think it's fine to have custom serialize/deserialize if an object needs special handling to get revived after a transfer. i think moving the StructArray.serialize/deserialize into web_transfer would not be an improvement. and it does need special handling.

HarelM:
Can you update the branch from main? I think the conflicts prevent it from running the CI, maybe...

src/util/web_worker_transfer.ts
```diff
@@ -91,6 +92,43 @@ function isArrayBuffer(value: any): value is ArrayBuffer {
            (value instanceof ArrayBuffer || (value.constructor && value.constructor.name === 'ArrayBuffer'));
 }
 
+function getClassRegistryKey(input: Object|SerializedObject) {
```
HarelM, line 95: What is the return type of this method? string?

HarelM:
Looks good in general.
Can you check if the tests pass (except for the ajax) also for the version of the code of the web worker in main? i.e. see if you can copy paste the web_worker content and see if the tests pass. Since this is a refactoring all the tests (except ajax) should pass in the old version it self.

HarelM (approved):

oberhamsi:
tests work in main :partying_face: 

i also added an explicit return type in two more places

Screenshot from 2024-06-19 14-13-16

oberhamsi:
i also checked the 4 lines in main not covered by the new tests.

two of them are still not covered because i don't think they can be triggered

one is covered once refactored (okay, semi-great). and one line is now removed because of the new Error handling.

HarelM:
Are you sure about the coverage? 
I see the following in the CI log for the unit tests:
```
web_worker_transfer.ts                     |   84.25 |     78.4 |     100 |   84.25 | 147-150,152-156,158-161,174-175,178-179,183-189,202-203,205-208,211-212,230-231,234-235,238-239,242-243
```

oberhamsi:
yeah, that's odd. i'll investigate. some of the lines CI says are not covered, i'm certain they should be hit by the test.

oberhamsi:
okay this is a fun one.

node 20.10 has a coverage bug https://github.com/jestjs/jest/issues/14766 https://github.com/nodejs/node/issues/51251

if i downgrade to node 20.9 the test coverage for web_worker_transfer is correct. the test coverage for a lot of other files is also different with 20.9

my screenshots where from a run with `npx jest --coverage src/util/web_worker_transfer.test.ts --reporters=default` which I believe uses `codeCoverage=babel`. if i change codeCoverage to babel in jest.config.js the coverage output by `npm run test-unit-ci` looks different (correct for web_worker_transfer).

i would suggest a different MR and switch to codeCoverage=babel. the only benefit of v8 would be speed.

HarelM:
We are using a special coverage report tool due to the differences between browser test coverage (render tests) and unit test.
The response in this package is usually fast: monocart-coverage-reports
It might be interesting to poke around there to see if we can get the right results.
We can move the build to use 20.9 if there's no solution in later version.
IDK... I don't understand why coverage is reinvented in node and jest... so annoying...

oberhamsi:
the reporting is fine but the coverage reported by node is just wrong in 20.10

i would suggest switching to babel coverageReporter instead of v8, which is a simple change and the default used by jest anyways. i don't see a reason to use v8 expect for speed.

HarelM:
These are the relevant PR related to this topic, I honestly don't understand all the details, I just want it to work...
- https://github.com/maplibre/maplibre-gl-js/pull/3598
- https://github.com/maplibre/maplibre-gl-js/pull/3615

Feel free to create a different PR to do the switch.

oberhamsi:
I will look into it but probably mid july

HarelM:
What do you suggest to do? I think general fixes to coverage can wait for a different PR and there's no need to hold this PR hostage...

oberhamsi:
honestly i haven't worked with jest coverage before, so i don't know how to fix it just for this test.

looking at the jest-docs yesterday, and trying different things, i feel changing [this line](https://github.com/maplibre/maplibre-gl-js/blob/d13e3db244229eb03570970bb8c4a1a0ad1ca89e/jest.config.ts#L17) to `babel` (the default) should fix the coverage reported. but it will change the coverage for a bunch of files (interestingly not all of them). so i would want to dig deeper before really doing this change. and mabye there's a way to change the coverageProvider only for for the "unit" project in jest.

btw: thanks for all the guidance & hints! i'm really interested in getting this MR done - our current workaround involves `setTimeout` to check for the loaded state :) but i'll be on holidays now - so for me it's okay if the PR wait a little longer.

HarelM:
@CenFun any insight regarding the above issue with the coverage? Do you have any tips?

cenfun:
I don't think it will work if we merge babel coverage (unit test ) with v8 coverage (e2e test). This is why we use Jest v8 coverage providor.
However, there is indeed a bug caused by the upgrade of nodejs 20.10, which may cause Jest's v8 coverage to not work.
I have created a PR to fix the Jest v8 issue, but unfortunately it does not get merged
https://github.com/SimenB/collect-v8-coverage/pull/235

It's been a month already, It seems like it hasn't been fixed, whether it's Nodejs or Jest. 
I don't know how to push forward it.

HarelM:
Thanks for the quick reply @cenfun!

I'll merge this and hope the fix will merged soon...

PR #4243, opened 2024-06-06
Don't reuse old requestedCameraState.

### LINKED ISSUE #4233: transformCameraUpdate leads to unexpected camera jumps when targetting terrain

**browser**: Firefox 126, Chrome 125

### Steps to Trigger Behavior

 1. Use the 3D-terrain example
 2. Add a transformCameraUpdate function in the options that returns the input or an empty object.
 3. Drag pan the map a little bit (inertia must be triggered). After end of drag, the jump happens.
 4. After the jump, it works fine until the terrain is targetted again with the camera.

### Link to Demonstration

https://jsbin.com/pobidixeqa/1/edit?html,output

### Expected Behavior

The empty `((tr)=>({})` or identity `(tr)=>tr` transformCameraUpdate option should never have an effect on the camera movement.

### Actual Behavior

The camera jumps to a different location when drag panning with triggered inertia and terrain targeted.

chrneumann:
Fixes unexpected camera jumps when transformCameraUpdate is specified and user is panning on maps with terrain (Fixes #4233).

Camera._getTransformForUpdate clones the transform on first call. After that, it is reused on later calls. I am not sure why, maybe only performance reasons? If the transform is cloned on every call (as changed by the PR), the jumps are gone. Could not find any problems yet and the tests run fine. Maybe I overlooked something. 

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/ui/camera.ts
+++ b/src/ui/camera.ts
@@ -308,10 +308,6 @@ export abstract class Camera extends Evented {
         this._zooming = false;
         this.transform = transform;
         this._bearingSnap = options.bearingSnap;
-
-        this.on('moveend', () => {
-            delete this._requestedCameraState;
-        });
     }
 
     /**
@@ -1125,6 +1121,7 @@ export abstract class Camera extends Evented {
         if (bearing !== undefined) nextTransform.bearing = bearing;
         if (elevation !== undefined) nextTransform.elevation = elevation;
         this.transform.apply(nextTransform);
+        delete this._requestedCameraState;
     }
 
     _fireMoveEvents(eventData?: any) {
--- a/src/ui/camera.ts
+++ b/src/ui/camera.ts
@@ -1098,10 +1098,7 @@ export abstract class Camera extends Evented {
      */
     _getTransformForUpdate(): Transform {
         if (!this.transformCameraUpdate) return this.transform;
-
-        if (!this._requestedCameraState) {
-            this._requestedCameraState = this.transform.clone();
-        }
+        this._requestedCameraState = this.transform.clone();
         return this._requestedCameraState;
     }
 


HarelM:
All tests are passing, meaning who ever added this code didn't add a test to make sure this behavior is kept. 
I would advise to dig a bit into the history of this to see if your can understand why this code was added...

chrneumann:
@Pessimistress Seems to be your code, do you remember why you added this and if it's safe to remove?

chrneumann:
Okay, I'm still not sure why requestedCameraState is kept, but Camera.easeTo seems to ignore it at some points and instead uses the original transform (that is not yet updated). Needs some more tests.

chrneumann:
As far as I understand it correctly, the only reason for `requestedCameraState` is it's use in the event handlers. Normally it would be enough to call `camera._getTransformForUpdate`, change the returned transform, and after that call `camera._applyUpdatedTransform` with the changed transform. But as the returned transform (from `_getTransformForUpdate`) can't be passed directly to the event handlers (they accessed the transform directly through the map before the change), the `TransformProvider` solution was used together with `requestedCameraState`. Still, I don't see a need to keep the `requestedCameraState` after the `_applyUpdatedTransform`. There should be no access to the `requestedCameraState` outside a `_getTransformForUpdate` / `_applyUpdatedTransform` pair. After the `_applyUpdatedTransform`, the `requstedCameraState` might differ from the transform, which I think leads to these subtle bugs.

Still not 100 percent sure, but this is as far I can get without feedback or some real world usage examples (could not find any at all). It's not even used in `react-map-gl` yet I think.

Added a commit to delete the `requestedCameraState` after `_applyUpdatedTransform`, not after the `moveend` event.

HarelM:
@Pessimistress @birkskyum can you help here?
We are struggling to understand the original reason for writing the code the way it was written...

HarelM:
@chrneumann can you add a test to simulate the bugs you experience? This way at least we'd know if we change the solution that the original issue is still fixed.

Pessimistress:
Your change will break the `transformCameraUpdate` feature for the [React use case](https://github.com/maplibre/maplibre-gl-js/issues/1545).

In short, only `_requestedCameraState` gets updated with the user input, and `transform` will be updated with an external state store (updates may be batched and asynchronous). If the terrain interaction is experiencing issues, it's likely because some changes are applied to `transform` directly, bypassing `_getTransformForUpdate`.

chrneumann:
@Pessimistress Do you have an example/test of your use case or how this breaks it?

Pessimistress:
> Do you have an example/test of your use case

https://codepen.io/Pessimistress/pen/xxNPvVr

chrneumann:
> https://codepen.io/Pessimistress/pen/xxNPvVr

Thanks, that makes it clearer. But the example seems to have an issue: it's not possible to drag pan the map while the map is still moving because of inertia of a last drag pan. Can't reproduce the problem without the `transformCameraUpdate`.

HarelM:
@chrneumann can you please provide a reproduction of an issue that this PR solves?

We need two tests here basically:
1. A test that simulates how the react library is using this API or something similar
2. A test that simulates the camera jump as described here: #4233

This requires a test from each one of you so that when both tests are fixed we'd know we can merge the change.

If there's a need for me to create a branch in this repo to facilitate for this work please let me know.

chrneumann:
>     2. A test that simulates the camera jump as described here: [transformCameraUpdate leads to unexpected camera jumps when targetting terrain #4233](https://github.com/maplibre/maplibre-gl-js/issues/4233)

See #4299.

chrneumann:
Found an other way to fix the issue in #4299. Still, a use case / test for the React use case is missing.

PR #4267, opened 2024-06-13
Move projection to style class

HarelM:
## Launch Checklist

Move the projection to the style, added an initial attempt to make the projection dynamic...?

### DIFF
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -53,13 +53,12 @@ import type {
     StyleSpecification,
     LightSpecification,
     SourceSpecification,
-    TerrainSpecification
+    TerrainSpecification,
+    ProjectionSpecification
 } from '@maplibre/maplibre-gl-style-spec';
 import type {MapGeoJSONFeature} from '../util/vectortile_to_geojson';
 import type {ControlPosition, IControl} from './control/control';
 import type {QueryRenderedFeaturesOptions, QuerySourceFeatureOptions} from '../source/query_features';
-import {Projection} from '../geo/projection/projection';
-import {ProjectionName, createProjectionFromName} from '../geo/projection/projection_factory';
 
 const version = packageJSON.version;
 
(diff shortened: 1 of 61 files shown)


HarelM:
@kubapelc @Pheonor Feel free to review this. I still want to add a test or two, but otherwise this is OK.
I even think the projection change works to some extent, maybe a forced update is needed somewhere, IDK.

src/style/style.test.ts
Pheonor, line 1: Perhaps It could be interesting to check if the default value of `projection` is well `mercator`

Pheonor (approved):
No more comment on this PR :)

src/style/style.test.ts
HarelM: Basically all the non globe render tests check that, but I don't mind adding a small test.

kubapelc:
Looks good!

kubapelc (approved):

src/ui/map.ts
```diff
@@ -3105,7 +3091,7 @@ export class Map extends Camera {
         // Even though `_styleDirty` and `_sourcesDirty` are reset in this
         // method, synchronous events fired during Style#update or
         // Style#_updateSources could have caused them to be set again.
-        const somethingDirty = this._sourcesDirty || this._styleDirty || this._placementDirty || this.projection.isRenderingDirty();
+        const somethingDirty = this._sourcesDirty || this._styleDirty || this._placementDirty || this.style.projection.isRenderingDirty();
```
HarelM, line 3094: @kubapelc should this check be moved to be part of `style.hasTransitions()`?

HarelM:
I'll merge this so that @Pheonor can continue his work, @kubapelc please let me know what you think about what I wrote with the dirty stuff...

src/ui/map.ts
kubapelc: Sorry that I forgot to respond to this earlier. This is really ambiguous, both keeping it here and moving it to `hasTransitions` makes sense. But in my dev branch I've already moved `isRenderingDirty` into the transform class, so maybe it makes sense to keep the check where it is for now, since it will no longer be part of the projection interface in the future.

HarelM: No problem, if this is not part of the projection class and will be removed in the future, there's no need to do anything now...

kubapelc: I'm now in the process of merging up-to-date globe branch into my dev branch, and I noticed that it probably makes most sense to keep an `isRenderingDirty` function in both Projection and Transform, since both can be dirty for different reasons. I think it makes sense to keep both checks on this line here in map.ts, since neither the transform's nor the projection's dirtiness is necessarily a visual transition (so I wouldn't move it to `style.hasTransitions()`).

HarelM: Fair enough, thanks for the update.

PR #4276, opened 2024-06-15
[WIP] Use moduleResolution: "NodeNext"

### LINKED ISSUE #4275: Use a modern moduleResolution strategy
Currently, the tsconfig is set to use moduleResolution: "Node", which is deprecated:

Screenshot 2024-06-15 at 11 59 21

Going to a modern module resolution such as NodeNext (or Bundler) would likely allow us to update dependencies like:
- #4263 (update, fixed by : https://github.com/maplibre/maplibre-gl-js/pull/4292)

And be a step towards
- #1595 

In practice "NodeNext" is more strict than what we do now, and would demand us to add file extensions (.ts/.js) on relative imports.

birkskyum:
Fixes #4275
- #4275 

NodeNext is more strict, so there is a bit of work still to make everything work - but this is a starting point fixing all the simple stuff like adding file extensions.


HarelM:
As far as I can see bundler doesn't require a file extension change: 
https://www.typescriptlang.org/tsconfig/#moduleResolution
Maybe it will be a simpler solution?

birkskyum:
Might be easier. The main headache I've seen this far is actually that jest doesn't seem to pick up the 'module' set in tsconfig, so for both nodenext and bundler it'll complain about invalid `module` settings.

PR #4284, opened 2024-06-17
Fix terrain listener memory leak when removing Marker

rczobor:
A memory leak is caused when adding and removing markers.  `.on('terrain', this._update)` event listener is added when Marker's `addTo` is called, but `off` call is missing from the `remove` method.

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
+- Fix `terrain` listener memory leak when adding and removing Marker ([#4284](https://github.com/maplibre/maplibre-gl-js/pull/4284))
 - _...Add new stuff here..._
 
 ## 4.4.0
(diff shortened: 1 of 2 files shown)


TimMcCauley:
Thank you @rczobor this is a really nice find which will help us in our project, too.

HarelM (approved):
THANKS!

HarelM:
Did you happen to check if there are other event that are causing a leak in this class?
I think there are in the geolocation control, but this control doesn't get added and removed often...

rczobor:
Thanks!

We didn't notice any other events so far, only this one. We do use supercluster to render thousands of dynamic markers with click/drag event handlers on the map via react-nap-gl, and this one popped out. It can crash a tab in minutes when abused.
Screenshot 2024-06-14 at 13 20 11

PR #4287, opened 2024-06-18
Prevent forks from attempting to publish the npm package (resulting in errors)

coliff:
This changes the GitHub Action to prevent forks from attempting to publish the npm package (resulting in errors) which could be frustrating for contributors.

e.g.
image

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,6 +9,7 @@ jobs:
   release-check:
     name: Check if version changed
     runs-on: ubuntu-latest
+    if: github.repository == 'maplibre/maplibre-gl-js'
     defaults:
       run:
         shell: bash


HarelM:
This will run only if you change the version number in the package.json.
What is the scenario that will trigger that?

coliff:
> This will run only if you change the version number in the package.json. What is the scenario that will trigger that?

When you sync after a new release

HarelM:
Ahh, right... Is there something less "hard-coded" to solve this?

coliff:
I think the repo name could be avoided with this change instead:

``yml
    # Only run on Pull Requests within the same repository, and not from forks.
    if: github.event.pull_request.head.repo.full_name == github.repository
```

HarelM:
But we want to run PRs that their origin is from forks.
I would recommend checking if npm token variable is not empty when trying to publish and tag, and issue a warning in case it is missing.

coliff:
> But we want to run PRs that their origin is from forks. I would recommend checking if npm token variable is not empty when trying to publish and tag, and issue a warning in case it is missing.

hmm, well the idea of the PR was to not show any warnings or errors for contributors who have the repo forked. I don't think we'd want the GitHub Action to run at all in such cases.

But it's not a really big deal either way, if you're not keen on this change you can close the PR and I will just have to manually deactivate the GitHub Action on my fork :-)

HarelM:
I'll close this one then.
I appreciate your contribution!
If I see that more people are complaining about this, then I'll think of an elegant way to solve this.
The reason I'm not keen about using the repo name is that I'm using a similar approach in other repos and I'm trying to keep this code similar between them for easier maintenance. Not ideal and there are probably better methods, but it is what it is... 🙂

PR #4299, opened 2024-06-19
Fix camera jumps with terrain and transformCameraUpdate.

### LINKED ISSUE #4233: transformCameraUpdate leads to unexpected camera jumps when targetting terrain

chrneumann:
Fixes camera jumps at end of movement when terrain and transformCameraUpdate is enabled (fixes #4233).

Had to enable WebGL for headless Chrome, as I could not reproduce the bug without it in headless mode.
The problem was caused by directly accessing the map's transform, instead of using _getTransformForUpdate and _applyUpdatedTransform.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


test/integration/browser/browser.test.ts
```diff
@@ -400,4 +406,76 @@ describe('Browser tests', () => {
...
+                        }
+                    ],
+                    terrain: {
+                        source: 'terrainSource',
+                        exaggeration: 1
+                    }
+                });
+
+            return new Promise<any>((resolve) => {
+                map.once('idle', () => {
```
HarelM, line 458: You can use `await map.once('idle');`.

test/integration/browser/browser.test.ts
```diff
@@ -400,4 +406,76 @@ describe('Browser tests', () => {
         await expect(rtlPromise).rejects.toThrow(regex);
 
     }, 2000);
+
+    test('Movement with transformCameraUpdate and terrain', async () => {
+        await page.evaluate(() => {
+            map.setPitch(52)
+                .setZoom(15)
+                .setCenter([11.40, 47.30])
+                .setStyle({
```
HarelM, line 415: Can we use a simpler style for the test?
Can we use a different type of test maybe?
This is more of a e2e test, which takes long to run, can we use a unit-test instead?

test/integration/browser/browser.test.ts
```diff
@@ -26,7 +26,13 @@ describe('Browser tests', () => {
         );
         await new Promise<void>((resolve) => server.listen(resolve));
 
-        browser = await puppeteer.launch({headless: true});
+        browser = await puppeteer.launch({
+            headless: true,
+            args: [
```
HarelM, line 31: Do the test fail if these parameters aren't used?

HarelM:
Thanks for taking the time to fix this issue!
This looks like the right solution to the problem.
See my small comments above.

test/integration/browser/browser.test.ts
chrneumann: The test will *never* fail if not used.

test/integration/browser/browser.test.ts
chrneumann: Tried it, but could not reproduce it with a unit test.

chrneumann: No, not with simply adding the transformCameraUpdate option. As said, the browser test does not even fail with WebGL disabled. I don't know why, I can't say that I understand every bit of the changes involved (as said elsewhere, it's hard to figure out without use case examples for the transformCameraUpdate stuff).

UberMouse: Here is our `transformCameraUpdate` usage as an example

```
map.transformCameraUpdate = (next) => {
	return {
    	...next,
        bearing: displaySettings.rotationLock ?? next.bearing,
        pitch: displaySettings.tiltLock ?? next.pitch,
    };
};
```

This is to enable us to lock the camera to a specific bearing and/or pitch based on settings. Turning off interactions isn't quite foolproof enough, but this locks it perfectly.

HarelM:
I think the following unit test can reproduce this issue to some extent - as the recalculate zoom will be done on the original transform and won't take the zoom changes that are done as part of the transfrom camera update:
```ts
    test('recalculate zoom is done on the camera update transform', () => {
        const map = createMap({
            interactive: true, 
            clickTolerance: 4,
            transformCameraUpdate: ({zoom}) => ({ zoom: zoom + 0.1})
        });
        map.terrain = {
            pointCoordinate: () => null,
            getElevationForLngLatZoom: () => 1000
        } as any;
        const canvas = map.getCanvas();
        simulate.dragWithMove(canvas, {x: 100, y: 100}, {x: 100, y: 150});
        map._renderTaskQueue.run();

        expect(map.getZoom()).toBe(0.20008255564976615);
    });
```
What do you think?

chrneumann:
Seems fine, still I think an integration test that involves transformCameraUpdate wouldn't be bad. But fine for me :)

HarelM:
I'm good with both tests, just please simplify the style in the integration test...

chrneumann:
> I'm good with both tests, just please simplify the style in the integration test...

Ah, I did that on purpose. Thought about simplifying it, which would be fine for the automatic testing, but if you want to observe the bug, the more complete style is really helpful. But yeah, I can simplify it if you don't think thats important.

HarelM:
Yeah, the test should be as simple as possible, for debugging you can change the examples in order to reproduce an issue.
Please add the test I sent and simplify the integration test and I'll merge this PR.
Thanks!!

chrneumann:
> I think the following unit test can reproduce this issue to some extent - as the recalculate zoom will be done on the original transform and won't take the zoom changes that are done as part of the transfrom camera update:
> What do you think?

Ah, you did not actually test it? At least, it never fails for me.

HarelM:
I've test it with and without the fix and it gave different results.

chrneumann:
Maybe I'm missing something. I pushed your unit test, you can check it with the commit "04278f129 | Fix terrain jumps with transformCameraUpdate (#4233)" reversed, which lets the integration test fail for me but not the unit test.

chrneumann:
Ah okay, it fails when the fix is applied.

chrneumann:
Ah, it fails when the fix is applied? Did you calculate the zoom by hand? Or is it fine to just fix the expected zoom value with the fix applied?

HarelM:
Must have written the test when I removed the fix, sorry...

HarelM (approved):

PR #4300, opened 2024-06-19
Keep camera outside terrain

### LINKED ISSUE #1542: Camera can move inside 3D terrain
This is a follow-on ticket from #1241. It's possible to move the camera to _within_ terrain when it's extruded from the 2D map surface.

@zdila said:

> May be a different issue, but in my case not only closest tiles but all. Open https://labs.maptiler.com/samples/maplibre/terrain/#style=satellite&lat=48.57344071&lng=20.46078220&zoom=17.16&bearing=-127.95&pitch=76.00&3d=true then zoom a little bit and tiles starts to disappear. If you refresh the page, nothing is shown until you zoom out. MapLibre 2.3.0.

@prozessor13 replied:

> Walking with the camera into terrain is still an open issue :/

### Expected behaviour

When the camera "bumps up" against 3D terrain, it should not be possible to move the camera to within the 3D surface.

### Actual behaviour

When you move the camera towards 3D terrain, nothing stops you when you "hit" the surface. You can keep moving through the surface.

chrneumann:
Fixes #1542. Currently proof of concept, comments appreciated.

It uses the `transformCameraUpdate` functionality to validate the next camera position, possibly adjusting it. By using `transformCameraUpdate`, the changes can be kept small. The assigned function checks if the next camera position will be inside terrain. If that would be the case, the camera is moved up (keeping its center), so that it is above the terrain. The pitch is adjusted to keep the map's center. IMHO this leads to a good user experience. Tried other options (stopping the camera or just adjusting height), but this seems best.

Proof of concept: Needs some more testing. `transformCameraUpdate` type signature is changed to pass the whole transform. Might not be needed, did not look at this yet. If needed, an internal wrapper for `transformCameraUpdate` might be the solution. 

Includes a temporary workaround for #4233. 

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


src/ui/map.ts
```diff
@@ -589,7 +589,28 @@ export class Map extends Camera {
         this._clickTolerance = resolvedOptions.clickTolerance;
         this._overridePixelRatio = resolvedOptions.pixelRatio;
         this._maxCanvasSize = resolvedOptions.maxCanvasSize;
-        this.transformCameraUpdate = resolvedOptions.transformCameraUpdate;
+        // this.transformCameraUpdate = resolvedOptions.transformCameraUpdate;
+        this.transformCameraUpdate = (tr: Transform) => {
+            try {
+                const buffer = Math.min(500, 20 * (25 - tr.zoom));
```
HarelM, line 595: Not sure it's clear what this buffer holds, a better name is probably a good idea here.

src/ui/map.ts
```diff
@@ -589,7 +589,28 @@ export class Map extends Camera {
         this._clickTolerance = resolvedOptions.clickTolerance;
         this._overridePixelRatio = resolvedOptions.pixelRatio;
         this._maxCanvasSize = resolvedOptions.maxCanvasSize;
-        this.transformCameraUpdate = resolvedOptions.transformCameraUpdate;
+        // this.transformCameraUpdate = resolvedOptions.transformCameraUpdate;
+        this.transformCameraUpdate = (tr: Transform) => {
```
HarelM, line 593: I'm guessing this entire method should be part of terrain class? Not sure, but it shouldn't be here I think...

src/ui/map.ts
```diff
@@ -589,7 +589,28 @@ export class Map extends Camera {
...
+            try {
+                const buffer = Math.min(500, 20 * (25 - tr.zoom));
+                const camera = tr.getCameraPosition();
+                const minAllowedAltitude = this.terrain.getElevationForLngLatZoom(camera.lngLat, tr.zoom) + buffer;
```
HarelM, line 597: Not sure I understand why this is the minimal altitude, can you add some diagram with some explanation on how you know if the camera is inside the terrain or not?

HarelM:
I think the approach presented here is a very elegant one. Thanks! This still needs some love and care, but otherwise a very good idea to solve this issue.
Assuming we solve the questions raised in #4243

chrneumann:
Waiting for #4299 (fixes camera jumps) to be approved before polishing this PR.

HarelM:
Generally, the linked PR looks like the right solution, so you can assume it will be merged, the only comments I have there are related to tests...

src/ui/map.ts
chrneumann: - `camera.lngLat` is the position of the camera (above the ground)
- `tr.zoom` is the current zoom
- `this.terrain.getElevationForLngLatZoom(camera.lngLat, tr.zoom)` gives the elevation of the terrain at the given camera's position at the given current zoom. The camera must be above this elevation / altitude.
- The buffer adds some extra altitude, so that the camera doesn't sit directly on the terrain surface (to get a better view).

src/ui/map.ts
chrneumann: I will move it to a private method of the map. Not sure about that. It's not exclusively about the terrain nor the camera. Maybe the Map's class is the right spot, as it handles the interaction between the two. Maybe an extra class that's used by Map, but that might be overkill?

chrneumann: Ok I think the Camera class is better than the Map. It already handles terrain stuff, so the fix will fit there IMHO.

chrneumann:
Found more problems (fixed in https://github.com/maplibre/maplibre-gl-js/pull/4300/commits/f6bdf15fe07a039135755f38a231d54c40ced9c7) with the direct manipulation of the transform, which leads to jumps and other problems. One should keep in mind that there might be other places where the transform is manipulated directly instead of using the _getTransformForUpdate / _applyUpdatedTransform combo.

Moved the logic to the camera.

src/ui/camera.ts
```diff
@@ -1091,43 +1091,84 @@ export abstract class Camera extends Evented {
...
         }
         return this._requestedCameraState;
     }
 
+    /**
+     * @internal
+     * Checks the given transform for the camera being below terrain surface and
+     * returns new pitch and zoom to fix that.
+     *
+     * With the new pith and zoom, the camera will be at the same ground
```
HarelM, line 1114:
```suggestion
     * With the new pitch and zoom, the camera will be at the same ground
```

src/ui/camera.ts
```diff
@@ -1091,43 +1091,84 @@ export abstract class Camera extends Evented {
...
-        } = this.transformCameraUpdate(nextTransform);
-        if (center) nextTransform.center = center;
-        if (zoom !== undefined) nextTransform.zoom = zoom;
-        if (pitch !== undefined) nextTransform.pitch = pitch;
-        if (bearing !== undefined) nextTransform.bearing = bearing;
-        if (elevation !== undefined) nextTransform.elevation = elevation;
-        this.transform.apply(nextTransform);
+        const modifiers = [];
+        if (this.terrain) {
+            modifiers.push(this._elevateCameraIfInsideTerrain.bind(this));
```
HarelM, line 1146: can you change from "bind(this)" to `() => ...`? I'm trying to remove `bind(this)` from the code as much as possible.

src/ui/camera.ts
```diff
@@ -1091,43 +1091,84 @@ export abstract class Camera extends Evented {
...
-     * If `transformCameraUpdate` is specified, a copy of the current transform is created to track the accumulated changes.
+     * If `transformCameraUpdate` is specified or terrain is enabled, a copy of
+     * the current transform is created to track the accumulated changes.
      * This underlying transform represents the "desired state" proposed by input handlers / animations / UI controls.
      * It may differ from the state used for rendering (`this.transform`).
      * @returns Transform to apply changes to
      */
     _getTransformForUpdate(): Transform {
-        if (!this.transformCameraUpdate) return this.transform;
+        if (!this.transformCameraUpdate && !this.terrain) return this.transform;
```
HarelM, line 1101: ahh, this is interesting, so in case of terrain you use the same infrastructure as for `transformCameraUpdate`. nice!

HarelM:
This looks good! There are failing render tests though, please see if there's a need to update the expected image of the fog tests due to this fix.

src/ui/camera.ts
chrneumann: Done :)

chrneumann:
> This looks good! There are failing render tests though, please see if there's a need to update the expected image of the fog tests due to this fix.

Yes, the pitch of the camera changed because the camera sits on the terrain surface instead of having the minimum distance to the surface. Changed the render images.

chrneumann:
Build size is bigger, I think that's fine?

HarelM:
Yes, don't worry about it.

src/ui/map.ts
```diff
@@ -591,6 +591,7 @@ export class Map extends Camera {
         this._overridePixelRatio = resolvedOptions.pixelRatio;
         this._maxCanvasSize = resolvedOptions.maxCanvasSize;
         this.transformCameraUpdate = resolvedOptions.transformCameraUpdate;
+
```
HarelM, line 594:
```suggestion
```

src/ui/camera.ts
```diff
@@ -1091,43 +1091,84 @@ export abstract class Camera extends Evented {
...
+            modifiers.push(tr => this._elevateCameraIfInsideTerrain(tr));
+        }
+        if (this.transformCameraUpdate) {
+            modifiers.push(tr => this.transformCameraUpdate(tr));
+        }
+        if (!modifiers.length) {
+            return;
+        }
+        const finalTransform = tr.clone();
+        modifiers.forEach((modifier) => {
```
HarelM, line 1155: Can you change to regular "for...of" please?

HarelM:
> the pitch of the camera changed because the camera sits on the terrain surface instead of having the minimum distance to the surface. Changed the render images.

Really? This seems strange... is the pitch this high? Are you sure the threshold is not too tight or something?

chrneumann:
> Really? This seems strange... is the pitch this high? Are you sure the threshold is not too tight or something?

I removed the padding altogether. Makes code simpler and I don't think its needed at all for UX.

HarelM:

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@
 - Fix GeolocateControl may be added twice when calling addControl/removeControl/addControl rapidly ([#4454](https://github.com/maplibre/maplibre-gl-js/pull/4454))
 - Fix `style.loadURL` abort error being logged when removing style ([#4425](https://github.com/maplibre/maplibre-gl-js/pull/4425))
 - Fix vector tiles not loading when html is opened via "resource://android" (i.e., the assets folder) in GeckoView on Android ([#4451](https://github.com/maplibre/maplibre-gl-js/pull/4451))
+- Fix camera being able to move into 3D terrain ([#1542](https://github.com/maplibre/maplibre-gl-js/issues/1542))
```
HarelM, line 25: This should be under "main" bug fixes.

chrneumann: Not sure what you mean with "main"?

HarelM: Move this up, currently it's under 4.5.1 which was released.

chrneumann: Oops, yeah, fixed that.

HarelM (approved):

HarelM:
Thanks!!

PR #4314, opened 2024-06-22
Terrain fog blending improvements

kubapelc:
I've tried to implement exponential fog, and it didn't look all that good, so I left the `pow(a, 2.0)` fog in place.

I implemented my other comments from the original [PR](https://github.com/maplibre/maplibre-gl-js/pull/3645). I've also modified all color blending to be gamma-correct (note the absence of a dark bar in the updated render test) and added a `smoothstep` function to make blending from fog color to horizon color smoother.


kubapelc:
I have no idea what is going on with that failing render test, on my own machine it fails in a different way (produces empty image instead of timing out), and also does so on the original sky2 branch.

HarelM:
It's a flickering test, I'll see if I can better understand why it fails

HarelM:
I was thinking about smoothstep as well, thanks for that!

PR #4341, opened 2024-06-28
Globe - transform+projection changes

kubapelc:
# Globe - transform+projection changes

**I've removed all changes to camera controls and animations for easier reviewing (they will come later), but I'll keep the changelog as it was. The demo still includes controls changes, but if you compile from source, you won't have them.**

Issue & discussion of globe:

## Changes

**Controls for globe** - I believe these should now feel natural, but here is a quick summary of how they work and what edge cases are handled:

- camera can now look at (or nearly at) the poles at any zoom level
- panning works in a way that is very similar to mercator, and converges to mercator controls at high zooms
	- this means that the pan speed of longitude is increased as the camera moves nearer to the poles
	- at low zooms however, longitude pan speed is reduced when near the poles to prevent the globe from spinning wildly
- zooming will now correctly keep the location under the cursor in the same place, as it does in mercator
	- this behaviour is not used near the planet edges, as it would move the camera too quickly and eventually cause glitches - an approximation is seamlessly used instead when cursor is near the horizon
	- zooming now works even if cursor is not on the planet surface at all - in this case the map zooms to a location near the horizon that is closest to the cursor
	- additionaly when zooming in/out with the cursor being placed too far away from the planet, the map is prevented from panning too quickly

Camera animations and functions implemented for globe projection:

- `fitBounds` - correctly computes zoom & center to fit a given box, respects to planet curvature & padding
- `flyTo`
- `easeTo`
- `jumpTo`

Other changes:

- **Transform is now an abstract class**, with specialized implementations for every projection
	- moved most functions from the Projection interface to the Transform class
- Projection interface is now much simpler, and mostly used for things with are independent of the map's state
- implemented all relevant projection and unprojection functions for globe
	- unprojecting a pixel that does not lie on the planet surface will instead return the location on the planet horizon that is nearest to the pixel in screenspace
- removed `posMatrix` - a per-tile projection matrix that was specific to Mercator projection and was explicitly used and passed to shaders, stored in the  `Tile` class, etc. Mercator now supplies this matrix through the same mechanism as globe (`getProjectionData` function).
- further symbol refactor - simplified projection code, removed all mercator-only code paths - the Projection interface (or rather the now general Transform class) is used instead, handling both projections as needed
- changing map projection will now reload all tiles

## Zoom & apparentZoom

Under globe, the planet's size is automatically adjusted when latituded is changed so that the map's center always exactly matches a mercator map. For example, zoom level 0 at equator will result in a small planet, but near Greenland it would be much bigger (think of how oversized Greenland is under mercator projection). Changing projection for a given map center & zoom will however display the exact same location with the exact same size thanks to this behaviour (experiment with the globe toggle in the demo!) Under globe, panning and other map controls automatically counteract this effect to keep the planet size constant, unless the user actually wants to zoom in.

This weird relation of planet size to zoom level can have unexpected effects when programmatic animations are used. Imagine you want to fly from Greenland at zoom 4 to the equator at zoom 4 with globe projection. You would expect the planet size to remain constant, since the zoom levels are the same, but it would actually decrease a lot at the equator!

For this reason most animations have a new optional parameter in options: `apparentZoom`. This exists alongside the already present `zoom` paremeter and either can be used. It specifies target "planet size" in relation to the starting planet size. So if want to fly from Greenland to equator, starting map zoom is 4 and target *apparentZoom* is 4, the final planet size will be the same as the starting size! If we would instead specify `apparentZoom: 5`, the planet at the end of the animation would be twice as large as at the start, if `apparentZoom: 3` it would be half size, etc. If globe is disabled, `apparentZoom` works exactly like the old `zoom` option.

## What is left

Globe should now be mostly usable. Here are the things that *don't* yet work:

- custom layer API for globe
- 3D model API for globe
- markers are not hidden/faded when behind planet horizon
- `queryRenderedFeatures`
- globe transform cannot be constrained at the moment - it is unclear how constraining should be implemented, since snapping the camera so that out of bounds regions aren't at all visible would probably feel very unnatural
- terrain - I'm not sure I'm going to be implementing that though

Other bugs at the moment:

- Markers do not respect terrain depth, even on mercator projection. Hence the failing unit test. I have no idea what caused this...
- `options.offset` semantics for camera animations (`flyTo`, `easeTo`) differ across globe and mercator, and I think the mercator implementation is actually bugged? (At least it does something different that the docs claim it should.)

## Demo link

[Play around with new globe controls here.](https://kvaleya.gitlab.io/maplibre/globe/globedemo.html)

## Launch Checklist

 - [x] Link to related issues.
 - ~~[ ] Include before/after visuals or gifs if this PR includes visual changes.~~
 - [x] Write tests for all new functionality. - mostly includes tests for new globe projection/unprojection functions and for camera animations
 - ~~[ ] Post benchmark scores.~~
 - ~~[ ] Add an entry to `CHANGELOG.md` under the `## main` section.~~


sbachinin:
I also observed a lot of quirks with `options.offset`. This feature seems to be very buggy.
Also there are some unit tests that treat these bugs as the desired behavior. At least so it appears to me.

HarelM:
This is a huge PR to review, is there a way to split it somehow to smaller features/ increments?

kubapelc:
I've removed all the modifications to map controls, which should make this somewhat smaller. Now the "only" change in this PR is the projection+transform rewrite, which includes mostly small changes, but across many many files (and a huge change to transform.ts).

src/util/util.ts
```diff
@@ -4,6 +4,85 @@ import {isOffscreenCanvasDistorted} from './offscreen_canvas_distorted';
 import type {Size} from './image';
 import type {WorkerGlobalScopeInterface} from './web_worker';
 
+/**
+ * Solves a quadratic equation in the form ax^2 + bx + c = 0 and returns its roots in no particular order.
+ * Returns null if the equation has no roots.
+ */
+export function solveQuadratic(a: number, b: number, c: number): {
```
HarelM, line 11: Methods in this file should be covered with unit tests.

src/util/smart_wrap.test.ts
```diff
@@ -1,12 +1,12 @@
...
-import {Transform} from '../geo/transform';
 import {smartWrap} from './smart_wrap';
+import {MercatorTransform} from '../geo/projection/mercator_transform';
 
-const transform = new Transform();
-transform.width = 100;
-transform.height = 100;
-transform.getHorizon = () => 0; // map center
+const transform = new MercatorTransform();
+transform['_width'] = 100;
```
HarelM, line 7: Is it possible to initialize the transform with these values (withs, height) instead of using this "unsafe" process?

src/ui/map_tests/map_style.test.ts
```diff
@@ -113,8 +113,8 @@ describe('#setStyle', () => {
 
     test('style transform overrides unmodified map transform', done => {
         const map = new Map({container: window.document.createElement('div')} as any as MapOptions);
-        map.transform.lngRange = [-120, 140];
-        map.transform.latRange = [-60, 80];
+        map.transform['_lngRange'] = [-120, 140];
```
HarelM, line 116: same here for `_lngRange`. I think jest has better ways, maybe even using spys?

src/symbol/placement.ts
```diff
@@ -274,42 +258,20 @@ export class Placement {
...
 
-        const posMatrix = this.transform.calculatePosMatrix(unwrappedTileID);
-
-        const pitchWithMap = layout.get('text-pitch-alignment') === 'map';
         const rotateWithMap = layout.get('text-rotation-alignment') === 'map';
         const pixelsToTiles = pixelsToTileUnits(tile, 1, this.transform.zoom);
 
-        const translationText = this.collisionIndex.mapProjection.translatePosition(
-            this.transform,
+        const translationText = this.collisionIndex.transform.translatePosition(
```
HarelM, line 264: Can you elaborate why `pitchWithMap = layout.get('text-pitch-alignment') === 'map'` was removed?

src/symbol/collision_index.test.ts
```diff
@@ -1,19 +1,17 @@
...
-import {Transform} from '../geo/transform';
-import {MercatorProjection} from '../geo/projection/mercator';
-
 describe('CollisionIndex', () => {
-
     test('floating point precision', () => {
-        const posMatrix = mat4.create();
         const x = 100000.123456, y = 0;
-        const transform = new Transform(0, 22, 0, 60, true);
+        const transform = new MercatorTransform(0, 22, 0, 60, true);
```
HarelM, line 9: Does this mean that there are missing tests for `GlobeTransform` for collision index?

src/symbol/collision_index.test.ts
```diff
@@ -1,19 +1,17 @@
...
         transform.resize(200, 200);
+        const tile = new UnwrappedTileID(0, new CanonicalTileID(0, 0, 0));
+        transform['_posMatrixCache'][tile.key] = mat4.create();
```
HarelM, line 12: same here, consider spys or a different solution... preferably by not using private members.

src/source/source_cache.ts
```diff
@@ -1011,8 +1011,8 @@ export class SourceCache extends Evented {
 
     getVisibleCoordinates(symbolLayer?: boolean): Array<OverscaledTileID> {
         const coords = this.getRenderableIds(symbolLayer).map((id) => this._tiles[id].tileID);
-        for (const coord of coords) {
-            coord.posMatrix = this.transform.calculatePosMatrix(coord.toUnwrapped());
+        if (this.transform) {
```
HarelM, line 1014: Can you explain this change, I'm not sure I understand it...

src/render/program/symbol_program.ts
```diff
@@ -228,9 +219,9 @@ const symbolSDFUniformValues = (
     const transform = painter.transform;
 
     return extend(symbolIconUniformValues(functionType, size,
-        rotateInShader, pitchWithMap, isAlongLine, isVariableAnchor, painter, matrix, labelPlaneMatrix,
+        rotateInShader, pitchWithMap, isAlongLine, isVariableAnchor, painter, labelPlaneMatrix,
         glCoordMatrix, translation, isText, texSize, pitchedScale), {
-        'u_gamma_scale': (pitchWithMap ? Math.cos(transform._pitch) * transform.cameraToCenterDistance : 1),
+        'u_gamma_scale': (pitchWithMap ? Math.cos(transform.pitch * Math.PI / 180.0) * transform.cameraToCenterDistance : 1),
```
HarelM, line 224: Why was this changed? The unit are now different?

src/render/draw_symbol.test.ts
```diff
@@ -224,3 +225,25 @@ describe('drawSymbol', () => {
     });
 
 });
+
+function createMockTransform() {
```
HarelM, line 229: Can we move this to the beginning of the file? This way it's easier to see what is imported from other files and used in the test code.
Also consider using Jest utilities for this.

src/render/draw_fill.test.ts
```diff
@@ -36,6 +36,7 @@ describe('drawFill', () => {
 
         const sourceCacheMock = new SourceCache(null as any, null as any, null as any);
         (sourceCacheMock.getTile as jest.Mock).mockReturnValue(mockTile);
+        //sourceCacheMock.getTile = (_: any) => mockTile;
```
HarelM, line 39: Why keeping this comment?

src/data/bucket/symbol_bucket.test.ts
```diff
@@ -18,19 +17,19 @@ import {IndexedFeature, PopulateParameters} from '../bucket';
...
 const feature = vt.layers.place_label.feature(10);
 
 /*eslint new-cap: 0*/
 const collisionBoxArray = new CollisionBoxArray();
-const transform = new Transform();
-transform.width = 100;
-transform.height = 100;
-transform.cameraToCenterDistance = 100;
+const transform = new MercatorTransform();
+transform['_width'] = 100;
```
HarelM, line 30: See my other commnets.

src/symbol/placement.ts
kubapelc: I think that previously pitching was handled in the `labelPlaneMatrix`, but now it is handled by calling some sort of `project*` function on the transform (since you can't represent pitch on a globe with just a matrix). So it is no longer needed in `getBucketParts`.

src/source/source_cache.ts
kubapelc: Since `calculatePosMatrix` is only used by mercator projection, the loop which computed `posMatrix` for all visible tiles was moved to transform under the name `precacheTiles` and is now an abstract method that only does something on mercator transform.

src/render/program/symbol_program.ts
kubapelc: Yes, old transform used to have both `_pitch` (in radians) and `pitch` (in degrees). I think that `_pitch` was supposed to be private (so it should not have been used here in the first place), but only marked as such with the underscore. Now it is also made `private` by the typescript keyword, and I modified all outside usages to use the public `pitch` in degrees instead.

src/geo/projection/globe_transform.test.ts
```diff
@@ -0,0 +1,388 @@
...
+            expect(unprojected.lng).toBeCloseTo(coords.lng, precisionDigits);
+            expect(unprojected.lat).toBeCloseTo(coords.lat, precisionDigits);
+            expect(projected.x).toBeCloseTo(point.x, precisionDigits);
+            expect(projected.y).toBeCloseTo(point.y, precisionDigits);
+            expect(globeTransform.center.lat).toBeCloseTo(20.659450722109348, precisionDigits);
+        });
+    });
+});
+
+function testPlaneAgainstLngLat(lngDegrees: number, latDegrees: number, plane: Array<number>) {
```
HarelM, line 367: Can we move these methods up?

src/render/draw_fill.test.ts
kubapelc: Fixed (it was for debugging and should not have made it into a commit).

src/util/smart_wrap.test.ts
kubapelc: Fixed, now uses `transform.resize()`

src/geo/projection/globe_transform.ts
```diff
@@ -0,0 +1,1013 @@
+import {mat4, vec3, vec4} from 'gl-matrix';
```
HarelM, line 1: I'm not sure I have a good way to review this, I don't know what was passed from the previous file to this one. Any thoughts on how to solve this?

src/geo/projection/mercator_transform.ts
```diff
@@ -0,0 +1,894 @@
...
+import {PointProjection, xyTransformMat4} from '../../symbol/projection';
+import {LngLatBounds} from '../lng_lat_bounds';
+
+/**
+ * Convert from LngLat to world coordinates (Mercator coordinates scaled by 512).
+ * @param transform - The reference transform instance - only the `worldSize` property is used, which itself depends on zoom.
+ * @param lnglat - The location to convert.
+ * @returns Point
+ */
+export function projectToWorldCoordinates(transform: {worldSize}, lnglat: LngLat): Point {
```
HarelM, line 23: Why are there exported method that are not part of the class?

src/ui/map_tests/map_style.test.ts
kubapelc: Fixed, now uses `transform.setMaxBounds()`

src/geo/transform.ts
```diff
@@ -1,83 +1,76 @@
...
  */
-export class Transform {
-    tileSize: number;
-    tileZoom: number;
-    lngRange: [number, number];
-    latRange: [number, number];
-    scale: number;
-    width: number;
-    height: number;
+export abstract class Transform {
```
HarelM, line 29: Is it possible to use composition instead of inheritance?
Abstract base classes usually tend to hold after a while methods that are specific to an implementation and are hard to maintain correctly...

src/geo/transform.ts
```diff
@@ -217,12 +269,12 @@ export class Transform {
 
     get zoom(): number { return this._zoom; }
     set zoom(zoom: number) {
-        const constrainedZoom = Math.min(Math.max(zoom, this.minZoom), this.maxZoom);
+        const constrainedZoom = this.getConstrained(this._center, zoom).zoom;
```
HarelM, line 272: I hete this setter, and in general setters that have logic inside them such as `_calcMatrices` in this case.
If we refactor this class I would like to change the signature of all the setters that are not "straight forward" to be `setZoom` etc.

src/symbol/projection.ts
```diff
@@ -218,45 +208,45 @@ function updateLineLabels(bucket: SymbolBucket,
...
+            continue;
+        }
+
+        const cameraToAnchorDistance = anchorPos.signedDistanceFromCamera;
+        const perspectiveRatio = getPerspectiveRatio(painter.transform.cameraToCenterDistance, cameraToAnchorDistance);
+
+        const fontSize = symbolSize.evaluateSizeForFeature(sizeData, partiallyEvaluatedSize, symbol);
+        const pitchScaledFontSize = pitchWithMap ? fontSize / perspectiveRatio : fontSize * perspectiveRatio;
+
+        const placeUnflipped: any = placeGlyphsAlongLine(projectionContext, pitchedLabelPlaneMatrixInverse, symbol, pitchScaledFontSize, false /*unflipped*/, keepUpright,
```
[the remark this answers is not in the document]
HarelM: Also, If you need to inline a comment about the variable "name" an input object is probably a better solution, as this method has too many argument.

src/geo/transform.ts
kubapelc: I think abstract class is a better fit in this case, since there is a lot of logic and boilerplate that can be implemented in the base class and make specialized classes easier to write and smaller.

src/source/source_cache.ts
HarelM: Can `transform` be undefined or null?

src/geo/transform.ts
HarelM: Composition can do the same in most cases.
Reading an implementation of an abstract class is usually a very hard task, as you don't know which part is implemented by the base class and which is not.
An interface for transform and a transform helper will be easier to read and understand I believe.
It's also easier to maintain, since changing a method in the helper is a lot less risky than changing a method in the base class, and a base class usually gets methods that should not be part of it after a while of maintenance.
I can't stress this enough.
Here's an article about it, but my main reasoning is from my personal experience. I also think that camera and map for example are poor choices of inheritance in this project, and while you can start of with a good design it usually don't end up well.
https://sheldonrcohen.medium.com/favoring-composition-over-inheritance-ff2ece6b7b4e

HarelM:
The failing unit tests are related to marker's opacity changes when behind the terrain.
It is probably related to some of the method name changes, and the fact that mocking them now isn't working as before.
It should probably be an easy fix, I would look at the following method changes:
`map.transform.lngLatToCameraDepth` and `map.terrain.depthAtPoint`

src/geo/transform.ts
kubapelc: My main goal with choosing inheritance was to avoid forcing every transform implementation to implement all the getters and setters, since they alone are a few hundred lines of code. But if we would remove the logic in the setters, this argument would no longer apply.

You make very good points. I'll think about it some more, but now I tend to agree that an interface is the better choice.

kubapelc: Refactoring transform to an interface is taking me a lot of time... there are still bugs and failing tests. Preview here:

https://github.com/windycom/maplibre-gl-js/blob/kubapelc/globe-vector-transform-interface/src/geo/transform_interface.ts

https://github.com/windycom/maplibre-gl-js/blob/kubapelc/globe-vector-transform-interface/src/geo/transform_helper.ts

https://github.com/windycom/maplibre-gl-js/blob/kubapelc/globe-vector-transform-interface/src/geo/projection/mercator_transform.ts

Turning the transform class into an interface is the easy part, it already was mostly abstract methods with no real inheritance anyway. Handling the complex setters is much harder, I use a new TransfromHelper class for that. I'm not sure this achives the goal of better maintainability.

src/geo/transform.ts
kubapelc: I've converted all setters (for consistency) to functions.

src/geo/transform.ts
kubapelc: I've pushed the inheritance->interface refactor.

src/geo/projection/mercator_transform.ts
kubapelc: They are specific to mercator projection and only used (iirc) in camera.ts for animations and such. Rather that adding them to the generic transform interface only for them to be implemented only in mercator transform, I made them standalone methods instead.

src/data/bucket/symbol_bucket.test.ts
kubapelc: Fixed (it stopped working anyway after refactoring transform to an interface).

src/symbol/collision_index.test.ts
kubapelc: The only transform method called in this test is `projectTileCoordinates`, I think that should be covered in transform-specific tests instead.

src/symbol/collision_index.test.ts
kubapelc: Fixed.

src/source/source_cache.ts
kubapelc: I think it cannot, I removed the check.

src/render/draw_symbol.test.ts
kubapelc: Moved to the beginning of the file.

src/geo/projection/globe_transform.test.ts
kubapelc: Moved to the beginning of the file.

src/geo/projection/globe_transform.ts
kubapelc: Some of the code should be from globe.ts originally, but there are many new projection/unprojection functions (at least I think I didn't include them in previous PRs). Significant changes are (I think):

- newFrameUpdate (only small changes)
- isOccluded (only small changes)
- calculateFogMatrix (this and everything below should be entirely new)
- lngLatToCameraDepth
- precacheTiles
- getBounds
- getConstrained
- setLocationAtPoint
- locationPoint
- pointCoordinate
- pointLocation
- isPointOnMapSurface
- getRayDirectionFromPixel
- isSurfacePointVisible
- rayPlanetIntersection
- unprojectScreenPoint

src/geo/transform.ts
kubapelc: The change is now pushed.

src/symbol/projection.ts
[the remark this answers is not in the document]
kubapelc: I think I already found it inlined like that. I've fixed the typing and refactored `placeGlyphsAlongLine` args into an object.

HarelM:
I've merged the following PR to main and afterward to the globe branch: 
- #3136

I think this is what created the last conflict.
We'll need to think how to move forward as changing the transform object might create harder merges in the near future, so we should aim at finalizing the globe effort soon to avoid grief.

In any case, I see that some tests are failing, let me know if you need help fixing those.

src/source/source_cache.test.ts
```diff
@@ -154,9 +154,9 @@ describe('SourceCache#addTile', () => {
         };
         sourceCache.on('dataloading', () => { add++; });
 
-        const tr = new Transform();
-        tr.width = 512;
-        tr.height = 512;
+        const tr = new MercatorTransform();
+        tr['_width'] = 512;
```
HarelM, line 158: `resize(...)`

src/geo/projection/mercator_transform.ts
HarelM: I see, I would consider removing the transform parameter and use "regular" input arguments.
Maybe moving them to the a separate file - mercator_utils.ts? I saw similar files with this concept such as "mesh_utils", query_utils", "handler_utils". This is just a thought.

src/geo/projection/mercator_transform.ts
```diff
@@ -652,91 +567,65 @@ export class Transform {
...
      * @returns screen point
      */
-    coordinatePoint(coord: MercatorCoordinate, elevation: number = 0, pixelMatrix = this.pixelMatrix): Point {
+    coordinatePoint(coord: MercatorCoordinate, elevation: number = 0, pixelMatrix = this._pixelMatrix): Point {
         const p = [coord.x * this.worldSize, coord.y * this.worldSize, elevation, 1] as vec4;
         vec4.transformMat4(p, p, pixelMatrix);
         return new Point(p[0] / p[3], p[1] / p[3]);
     }
 
-    /**
```
HarelM, line 661: This comment was removed, is it not helpful?

src/geo/projection/mercator_transform.ts
```diff
@@ -851,156 +740,146 @@ export class Transform {
...
         const topHalfMinDistance = Math.min(topHalfSurfaceDistance, topHalfSurfaceDistanceHorizon);
-        const farZ = (Math.cos(Math.PI / 2 - this._pitch) * topHalfMinDistance + lowestPlane) * 1.01;
+        const farZ = (Math.cos(Math.PI / 2 - this._helper._pitch) * topHalfMinDistance + lowestPlane) * 1.01;
 
         // The larger the value of nearZ is
         // - the more depth precision is available for features (good)
         // - clipping starts appearing sooner when the camera is close to 3d features (bad)
         //
-        // Other values work for mapbox-gl-js but deck.gl was encountering precision issues
+        // Other values work for mapbox-gl-js but deckgl was encountering precision issues
```
HarelM, line 789: I think deck.gl is the right way to write it, isn't it?

src/geo/projection/mercator_transform.ts
```diff
@@ -1066,7 +910,163 @@ export class Transform {
...
+        }
+    }
+}
+
+/**
+ * Transform a matrix to incorporate the *-translate and *-translate-anchor properties into it.
+ * @param inViewportPixelUnitsUnits - True when the units accepted by the matrix are in viewport pixels instead of tile units.
+ * @returns matrix
+ */
+export function translatePosMatrix(
```
HarelM, line 992: Some exported methods are at the top and some are at the bottom, as I said before, moving them to a separate file would be the cleanest solution I think...

src/geo/projection/mercator_transform.ts
```diff
@@ -1066,7 +910,163 @@ export class Transform {
...
+    let mainMatrix: mat4;
+    if (overscaledTileID && overscaledTileID.terrainRttPosMatrix && !ignoreTerrainMatrix) {
+        mainMatrix = overscaledTileID.terrainRttPosMatrix;
+    } else if (tilePosMatrix) {
+        mainMatrix = tilePosMatrix;
+    } else {
+        mainMatrix = mat4.create();
+    }
+
+    const data: ProjectionData = {
```
HarelM, line 1063: This looks like gl related stuff, It should be elsewhere, like sky, light, etc...

src/source/source_cache.ts
kubapelc: Turns out it can be null sometimes, but I'm not sure why. It seems that whenever transform is null, the `coords` array is also empty, so it wasn't an issue before this PR. I'll put the null check back in.

src/geo/projection/mercator_transform.ts
kubapelc: I've moved all such functions into mercator_utils.ts and also did the same for globe, moved stuff from there to globe_utils.ts.

src/source/source_cache.test.ts
kubapelc: Fixed.

src/geo/projection/mercator_transform.ts
kubapelc: The comment is still present in transform_interface.ts. I think most programmers will interact with the method through the interface and this see the comment, I'm not sure if it should be repeated here in the mercator specific implementation.

src/geo/projection/mercator_transform.ts
kubapelc: Fixed. I have no idea why that changed in the first place 😅

src/geo/projection/mercator_transform.ts
kubapelc: Yes, this makes an object that describes uniform values for shaders, but the contents are projection-specific, this is a helper function that is used by both mercator transform and (indirectly) globe transform, which then also further modifies the uniform object's contents.

src/geo/projection/mercator_transform.ts
HarelM: I think you can use inheritdoc for these kind of situations.

src/source/source_cache.ts
HarelM: Would be interesting to understand why it is bull though...

src/geo/projection/globe_transform.test.ts
```diff
@@ -0,0 +1,389 @@
...
+            projected = globeTransform.locationPoint(globeTransform.center);
+            expect(projected.x).toBeCloseTo(screenCenter.x, precisionDigits);
+            expect(projected.y).toBeCloseTo(screenCenter.y, precisionDigits);
+
+            globeTransform.setCenter(new LngLat(0, 84));
+            projected = globeTransform.locationPoint(globeTransform.center);
+            expect(projected.x).toBeCloseTo(screenCenter.x, precisionDigits);
+            expect(projected.y).toBeCloseTo(screenCenter.y, precisionDigits);
+
+            // Try projecting a location that is slightly above and below map's center point
```
HarelM, line 185: Use test description instead of comments.
I would also recommend splitting the tests more, it seems like you have multiple steps in the tests, which might indicate you can split them into several smaller tests with more specific title.

src/geo/projection/globe_transform.ts
```diff
@@ -0,0 +1,1120 @@
...
+function tileCoordinatesToMercatorCoordinates(inTileX: number, inTileY: number, tileID: UnwrappedTileID): [number, number] {
+    const scale = 1.0 / (1 << tileID.canonical.z);
+    return [
+        inTileX / EXTENT * scale + tileID.canonical.x * scale,
+        inTileY / EXTENT * scale + tileID.canonical.y * scale
+    ];
+}
+
+/**
+ * Returns the angle in radians between two 2D vectors.
```
HarelM, line 44: Please use tsdoc type of comments with `@param` and `@returns`.

src/geo/transform_helper.ts
```diff
@@ -0,0 +1,494 @@
...
+        this._unmodified = unmodified;
+        this._constraining = false;
+    }
+
+    /**
+     * This function is called every time one of the transform's defining properties (center, pitch, etc.) changes.
+     * This function should update the transform's internal data, such as matrices.
+     * Any derived `_calcMatrices` function should also call the base function first. The base function only depends on the `_width` and `_height` fields.
+     */
+    protected _calcMatrices(): void {
```
HarelM, line 477: Is `protected` still needed here?

HarelM:
According to latest coverage report (which might have bugs unfortunately, so you might need to verify this locally) the following files are missing some coverage:
the utils classes should be 100% covered or something very close to it, regarding globe transform, it would be great if we could bring it up to 90% or so.
image

Other than that, I don't have other inputs, it's hard to review the transform changes due to how it moved and the ability to compare to what was there before, so I'm relying mostly on code coverage here...

src/geo/projection/mercator_transform.ts
kubapelc: It seems that Inheritdoc doesn't inherit the body of the doccomment, which is where most of the useful information is.

obrazek

I think it is still better to remove all the duplicate comments, since it would avoid errors when only one of the duplicate comments gets modified in the future (instead of all its duplicates), and at least in VS code the doccomment from the interface still shows up.

obrazek

HarelM: Does this part get to the mkdocs documentation? How does it show there? I think I used inherit doc somewhere in this project, didn't I?

src/symbol/collision_index.test.ts
kubapelc: `projectTileCoordinates` is now covered by tests for both transform variants.

src/geo/projection/globe_transform.ts
kubapelc: Fixed.

src/geo/transform_helper.ts
kubapelc: No, I forgot to remove it. Fixed.

src/geo/projection/mercator_transform.ts
kubapelc: It is used in attribution control, it seems to work fine with the generated docs: https://maplibre.org/maplibre-gl-js/docs/API/classes/AttributionControl/ (see onAdd - entire doccomment in inherited)

src/geo/projection/globe_transform.test.ts
kubapelc: I've split some of them up and added more `describe` blocks. Naming most of the tests is hard, since they just test a given function for some arbitrary nontrivial inputs that aren't special. Most of the special ones should have a named test now.

kubapelc:
I've added many more tests now.

src/geo/projection/globe_transform.ts
HarelM: Is it? Did you forget to commit the change?

HarelM:
Thanks, there are a few last comments that I need to check on my side, I'll report back once I do.
Overall, I think this is mostly ready, I'll do a last review in the coming days.

src/geo/projection/globe_transform.ts
kubapelc: I think I originally added doccomment to a different function than what you had in mind. I improved the comment of `tileCoordinatesToMercatorCoordinates` as well now.

src/geo/projection/globe_transform.ts
```diff
@@ -0,0 +1,1155 @@
...
+        if (oldGlobeness !== this._globeness) {
+            this.setCenter(new LngLat(
+                this._mercatorTransform.center.lng + differenceOfAnglesDegrees(this._mercatorTransform.center.lng, this.center.lng) * this._globeness,
+                lerp(this._mercatorTransform.center.lat, this.center.lat, this._globeness)
+            ));
+            this.setZoom(lerp(this._mercatorTransform.zoom, this.zoom, this._globeness));
+        }
+    }
+
+    isRenderingDirty(): boolean {
```
HarelM, line 437: This seems strange to be part of the transform, isn't it?
Unless I'm misunderstanding what transform should be responsible for.

src/geo/projection/globe_transform.ts
```diff
@@ -0,0 +1,1155 @@
...
+        const now = browser.now();
+        // Globe transition
+        return (now - this._lastGlobeChangeTime) / 1000.0 < (Math.max(globeConstants.globeTransitionTimeSeconds, globeConstants.zoomTransitionTimeSeconds) + 0.2);
+    }
+
+    getProjectionData(overscaledTileID: OverscaledTileID, aligned?: boolean, ignoreTerrainMatrix?: boolean): ProjectionData {
```
HarelM, line 443: This looks like gl related data, it is usually stored and fetched in a different way, usually where the gl interface is defined.
I would consider moving this method to a different file and have it accept a transform in order to calculate the projection data.

src/geo/projection/globe_transform.ts
```diff
@@ -0,0 +1,1155 @@
...
+        }
+
+        const newLng = validLng / Math.PI * 180;
+        const newLat = validLat / Math.PI * 180;
+        const oldLat = this.center.lat;
+        this.setCenter(new LngLat(newLng, clamp(newLat, -90, 90)));
+        this.setZoom(this.zoom + getZoomAdjustment(oldLat, this.center.lat));
+    }
+
+    locationPoint(lnglat: LngLat, terrain?: Terrain): Point {
```
HarelM, line 934: Is this the "opposite" of `setLocationAtPoint`? If so I would consider renaming to `getLocationAtPoint`.

src/geo/projection/globe_transform.ts
```diff
@@ -0,0 +1,1155 @@
...
+    }
+
+    /**
+     * Returns the two intersection points of the ray and the planet's sphere, or null if no intersection occurs.
+     * The intersections are encoded in the direction along the ray, with `tMin` being the first intersection and `tMax` being the second.
+     * @param origin - The ray origin.
+     * @param direction - The normalized ray direction.
+     */
+    private rayPlanetIntersection(origin: vec3, direction: vec3): {
+        tMin: number;
```
HarelM, line 1057: what does the "t" stands for? Is it possible to return a longer more meaningful name?

src/geo/projection/globe_transform.ts
```diff
@@ -0,0 +1,1155 @@
...
+            const intersectionPoint = createVec3();
+            vec3.add(intersectionPoint, rayOrigin, [
+                rayDirection[0] * intersection.tMin,
+                rayDirection[1] * intersection.tMin,
+                rayDirection[2] * intersection.tMin
+            ]);
+            const sphereSurface = createVec3();
+            vec3.normalize(sphereSurface, intersectionPoint);
+            return sphereSurfacePointToCoordinates(sphereSurface);
+        } else {
```
HarelM, line 1117: Since there's a return just above this else, this else is not needed.
I would also suggest to reduce some indentation (by removing some "else"s) to improve redability.

src/geo/projection/globe_transform.ts
```diff
@@ -0,0 +1,1155 @@
...
+            // Ray does not intersect the sphere -> find the closest point on the horizon to the ray.
+            // Intersect the ray with the clipping plane, since we know that the intersection of the clipping plane and the sphere is the horizon.
+            const originDotPlaneXyz = this._cachedClippingPlane[0] * rayOrigin[0] + this._cachedClippingPlane[1] * rayOrigin[1] + this._cachedClippingPlane[2] * rayOrigin[2];
+            const directionDotPlaneXyz = this._cachedClippingPlane[0] * rayDirection[0] + this._cachedClippingPlane[1] * rayDirection[1] + this._cachedClippingPlane[2] * rayDirection[2];
+            const tPlane = -(originDotPlaneXyz + this._cachedClippingPlane[3]) / directionDotPlaneXyz;
+
+            const maxRayLength = 2.0; // One globe diameter
+            const planeIntersection = createVec3();
+
+            if (tPlane > 0) {
```
HarelM, line 1127: What is `tPlane`? Can we use a longer name?

src/geo/projection/globe_transform.ts
```diff
@@ -0,0 +1,1155 @@
...
+            } else {
+                // When the ray takes too long to hit the plane (>maxRayLength), or if the plane intersection is behind the camera, handle things differently.
+                // Take a point along the ray at distance maxRayLength, project it to clipping plane, then continue as normal to find the horizon point.
+                const distantPoint = createVec3();
+                vec3.add(distantPoint, rayOrigin, [
+                    rayDirection[0] * maxRayLength,
+                    rayDirection[1] * maxRayLength,
+                    rayDirection[2] * maxRayLength
+                ]);
+                const distanceFromPlane = distantPoint[0] * this._cachedClippingPlane[0] + distantPoint[1] * this._cachedClippingPlane[1] + distantPoint[2] * this._cachedClippingPlane[2] + this._cachedClippingPlane[3];
```
HarelM, line 1142: Don't we have a utility method for distance from plane?

src/geo/projection/globe_transform.ts
```diff
@@ -0,0 +1,1155 @@
...
+        // Here we instead use the approach suggested in the book Ray Tracing Gems, chapter 7.
+        // https://www.realtimerendering.com/raytracinggems/rtg/index.html
+        const inner = createVec3();
+        const scaledDir = createVec3();
+        vec3.scale(scaledDir, direction, originDotDirection);
+        vec3.sub(inner, origin, scaledDir);
+        const discriminant = planetRadiusSquared - vec3.dot(inner, inner);
+
+        if (discriminant >= 0) {
+            const c = vec3.dot(origin, origin) - planetRadiusSquared;
```
HarelM, line 1074: Better variable names would make it more readable.

src/geo/projection/globe_transform.ts
```diff
@@ -0,0 +1,1155 @@
...
+}
+
+/**
+ * Returns mercator coordinates in range 0..1 for given coordinates inside a specified tile.
+ * @param inTileX - X coordinate in tile units - range [0..EXTENT].
+ * @param inTileY - Y coordinate in tile units - range [0..EXTENT].
+ * @param canonicalTileID - Tile canonical ID - mercator X, Y and zoom.
+ * @returns Mercator coordinates of the specified point in range [0..1].
+ */
+function tileCoordinatesToMercatorCoordinates(inTileX: number, inTileY: number, canonicalTileID: CanonicalTileID): [number, number] {
```
HarelM, line 39: I would expect to see this method in mercator_utils file...?

src/geo/projection/globe_transform.ts
```diff
@@ -0,0 +1,1155 @@
...
+ * Returns the angle in radians between two 2D vectors.
+ * The angle is signed and describes how much the first vector would need to be be rotated clockwise
+ * (assuming X is right and Y is down) so that it points in the same direction as the second vector.
+ * @param vec1x - The X component of the first vector.
+ * @param vec1y - The Y component of the first vector.
+ * @param vec2x - The X component of the second vector.
+ * @param vec2y - The Y component of the second vector.
+ * @returns The signed angle between the two vectors, in range -PI..PI.
+ */
+function angleToRotateBetweenVectors2D(vec1x: number, vec1y: number, vec2x: number, vec2y: number): number {
```
HarelM, line 57: Where did we place other trigonometry utils methods? I would advise to move it there...

src/geo/projection/globe_transform.ts
```diff
@@ -0,0 +1,1155 @@
...
+
+    private _lastGlobeChangeTime: number = -1000.0;
+    private _globeProjectionEnabled = true;
+
+    /**
+     * Note: projection instance should only be accessed in the {@link newFrameUpdate} function
+     * to ensure the transform's state isn't unintentionally changed.
+     */
+    private _projectionInstance: GlobeProjection;
+    private _globeRendering: boolean = true;
```
HarelM, line 268: Can this be computed from globeness instead of adding this as a property?

src/geo/projection/globe_transform.ts
```diff
@@ -0,0 +1,1155 @@
...
+    private _cameraPosition: vec3 = createVec3();
+
+    private _lastGlobeChangeTime: number = -1000.0;
+    private _globeProjectionEnabled = true;
+
+    /**
+     * Note: projection instance should only be accessed in the {@link newFrameUpdate} function
+     * to ensure the transform's state isn't unintentionally changed.
+     */
+    private _projectionInstance: GlobeProjection;
```
HarelM, line 267: How is it possible that mercator transfrom doesn't use a mercator projection instance? Seems strange to me this asymmetry.
If this is something that is only used in one method, it might be that this is not the place for it. I don't know if that's true, it's just a rough thought...

src/geo/projection/globe_transform.ts
```diff
@@ -0,0 +1,1155 @@
...
+    private _lastGlobeRenderingState: boolean = true;
+    private _globeLatitudeErrorCorrectionRadians: number = 0;
+
+    /**
+     * Globe projection can smoothly interpolate between globe view and mercator. This variable controls this interpolation.
+     * Value 0 is mercator, value 1 is globe, anything between is an interpolation between the two projections.
+     */
+    private _globeness: number = 1.0;
+    private _mercatorTransform: MercatorTransform;
+    private _initialized: boolean = false;
```
HarelM, line 278: Is this really needed? Javascript is single threaded and I don't see any async operations in the constructor...

src/geo/projection/globe_transform.ts
```diff
@@ -0,0 +1,1155 @@
...
+
+    /**
+     * Sets whether globe view is allowed. When allowed, globe fill function as normal, displaying a 3D planet,
+     * but transitioning to mercator at high zoom levels.
+     * Otherwise, mercator will be used at all zoom levels instead.
+     * @param allow - Sets whether glove view is allowed.
+     * @param animateTransition - Controls whether the transition between globe view and mercator (if triggered by this call) should be animated. True by default.
+     */
+    public setGlobeViewAllowed(allow: boolean, animateTransition: boolean = true) {
+        if (allow !== this._globeProjectionEnabled) {
```
HarelM, line 351: Consider returning and inverting this if to improve readbility.

src/geo/projection/globe_transform.ts
```diff
@@ -0,0 +1,1155 @@
...
+            if (!animateTransition) {
+                this._skipNextAnimation = true;
+            }
+            this._globeProjectionEnabled = allow;
+            this._lastGlobeChangeTime = browser.now();
+        }
+    }
+
+    translatePosition(tile: Tile, translate: [number, number], translateAnchor: 'map' | 'viewport'): [number, number] {
```
HarelM, line 360: Do we really need this as part of the transform interface if there's a method that receives a transform?
Consider removing this method from the transform class and keep it "external".

src/geo/projection/globe_transform.ts
```diff
@@ -0,0 +1,1155 @@
...
+        // In the future, some better translation for globe and other weird projections should be implemented here,
+        // especially for the translateAnchor==='viewport' case.
+        return translatePosition(this, tile, translate, translateAnchor);
+    }
+
+    /**
+     * Should be called at the beginning of every frame to synchronize the transform with the underlying projection.
+     * May change the transform's state - do not call on cloned transforms that should behave immutably!
```
HarelM, line 368: Is there a way to solve this in the code - this immutability transform state/object?
While this comment is important, it is hard to really know when a transform is immutable and when it isn't just from reading the code...

src/geo/projection/globe_transform.ts
```diff
@@ -0,0 +1,1155 @@
...
+        if (this._lastGlobeRenderingState !== this._globeRendering) {
+            forcePlacementUpdate = true;
+        }
+        this._lastGlobeRenderingState = this._globeRendering;
+        return {
+            forcePlacementUpdate
+        };
+    }
+
+    private _updateAnimation() {
```
HarelM, line 396: This looks to me like the sky/light/atmosphere code that is responsible for rendering transition. I think it should not be part of the transfrom class, but rather a rendering specific globe related class.
It might use the transform like recalculateZoom does from outside, but animation and transitioning should not be here, I think, not sure, what do you think?

src/geo/projection/globe_transform.ts
HarelM: I reviewed most of the class now, and added some comments. Thanks for the info.

src/geo/projection/globe_transform.ts
kubapelc: This method has different implementations in both transform variants, moving it elsewhere would mean it would have to have a type check on what type of transform was passed to it.

I think transform is also the right place for this particular method. Even though it does return uniform values for a shader, which is usually done elsewhere, it only returns everything needed to project a vertex to screen, which is what transform is ultimately for. Until now it was a simple matrix, but globe needs a bit more data.

src/geo/projection/globe_transform.ts
kubapelc: It is not, it is instead the opposite of `pointLocation`. `setLocationAtPoint` moves the map so that a given lnglat is at a given pixel, `locationPoint` and `pointLocation` just compute lnglat or pixel coordinates.

I personally don't really like the  `locationPoint` and `pointLocation` naming convention used in transform, I find it pretty unclear and can't easily tell which method does what conversion (is locationPoint lnglat to pixel or pixel to lnglat?). Do you think this PR is a good opportunity to rename them?

HarelM: Yes, this PR is mainly about transform changes, it is a good place to rename methods.
I would recommend the `toSreenPoint` to `toLngLat` type of convention - I find it the most intuitive.

src/geo/projection/globe_transform.ts
kubapelc: "t" is the parameter for parametric ray equation `origin + direction * t` and "t" is the standard way to refer to it. I'm not sure what a better name would be, something like "intersectionMin/Max" comes to mind, but that is inaccurate, since "intersection" usually refers to the coordinates of the intersection point. But I'll definitely improve the doccomment to better explain what it is.

src/geo/projection/globe_transform.ts
HarelM: Is it possible somehow to move it to a place where we know we need to draw the globe? like we do with the sky and atmosphere?

src/geo/projection/globe_transform.ts
HarelM: A return type with proper docs might be able to solve this as you said.

src/geo/projection/globe_transform.ts
kubapelc: No, the globe uniforms it returns are used by every layer.

src/geo/projection/globe_transform.ts
HarelM: Also I think there is a way to use `Transitionable` and `Transitioning` for this type of stuff, I'm not sure how they work under the hood, but sky and atmosphere are using them...

kubapelc: Transform decides when to transition from globe to mercator, so it needs to be able to signal that the animation is in progress. This animation state does not need to be cloneable with the transform (only whether mercator is or isn't used must be passed to clones), so it could be moved to the projection instance (globe.ts), but that would introduce a cyclical dependency. Right now globe.ts does not import globe_transform.ts, but globe_transform.ts does import globe.ts.

I'm not sure how Transitionable would help, it seems to be intended for style properties.

src/geo/projection/globe_transform.ts
kubapelc: I've renamed them. (I'll push the changes later all at once, as always.)

src/geo/projection/globe_transform.ts
kubapelc: I've added a return type and improved the comments.

src/geo/projection/globe_transform.ts
kubapelc: Done.  (I slightly prefer if-else even if both branches return to emphasize that the code paths are exclusive, but the else branch is a bit too long in this case.)

src/geo/projection/globe_transform.ts
kubapelc: I've renamed the variables to make it a bit clearer what the computation does:

```
const originToPlaneDistance = (originDotPlaneXyz + this._cachedClippingPlane[3]);
const distanceToIntersection = -originToPlaneDistance / directionDotPlaneXyz;
```

src/geo/projection/globe_transform.ts
kubapelc: It follows the equation from the linked book exactly. I'm not sure what "c" it represents, other than it apparently being a difference of squares. What the "t" variables are is now explained in the doccomment.

src/geo/projection/globe_transform.ts
kubapelc: Good catch, I've moved it there.

src/geo/projection/globe_transform.ts
kubapelc: I've moved it to util.ts

src/geo/projection/globe_transform.ts
kubapelc: Yes, I've changed it to a getter and also refactored how globeness is computed.

src/geo/projection/globe_transform.ts
kubapelc: Globe needs to periodically measure the error of GPU's `atan` implementation at current latitude, and this is handled in the projection instance, since it needs a few GPU resources and a shader which shouldn't exist multiple times, and projection instance is mostly a singleton (whereas transform often gets cloned often).

Globe transform then needs to get the results of the error measurement, so it needs a reference to projection instance, and it also needs to be typed to GlobeProjection. The easiest way to achieve this without typecasting was to pass the projection instance to GlobeTransform in the constructor in projection_factory.ts.

src/geo/projection/globe_transform.ts
kubapelc: It is no longer needed, I've removed it. It was needed when transform was an abstract class, since the base constructor sometimes called derived class' `_calcMatrices` which crashed if the derived class wasn't properly initialized yet.

src/geo/projection/globe_transform.ts
kubapelc: I think it might be useful for future projections if this was part of transform, but that is speculative, right now it is simpler to have it as a standalone utility function. I've now refactored it to be one.

src/geo/projection/globe_transform.ts
kubapelc: I've added a `IReadonlyTransform` interface that can be used in code that should not modify the transform. I've also added a comment to it warning that even if I have a `IReadonlyTransform` instance, some other code may have the same object with `ITransform` and can still modify it.

src/geo/projection/globe_transform.ts
kubapelc: Done.

HarelM:
I think my review is mostly done.
Thanks @kubapelc for you patience and changes, and for putting up with my annoying review comments.
There are still a few open review comments in this PR, but I don't know if they are worth following through.
I'll merge this into globe and you can decide if you want to follow them or not, if there's anything to do with what I wrote.

kubapelc:
Thank you @HarelM for the in-depth code review! I'll probably leave the last two commented snippets (projection instance reference and animation) as they now, I'm not sure what better to do.

Issue #4362, opened 2024-07-04
Lagging when zooming in/out of the map (especially with free rolling mouse wheel)


Codebreaker101:
**maplibre-gl-js version**: v4.5.0

**browser**: Chromium (Version 126.0.6478.126 (Official Build) (64-bit)) and their derivatives

### Steps to Trigger Behavior

 1. Go to https://maplibre.org/maplibre-gl-js/docs/ in a Chromium (or derivative) browser
 2. Zoom in/out fast and you can see that it lags (especially noticable using a mouse with free rolling wheel (like mx master 3) )

### Link to Demonstration

Laggy Chromium: 

https://github.com/maplibre/maplibre-gl-js/assets/16204269/62e4ec0a-08e7-44ee-bec0-91e137971e0f

Smooth Firefox: 

https://github.com/maplibre/maplibre-gl-js/assets/16204269/3848aa28-bce4-414f-a7dd-70caeed950e0

### Expected Behavior

For smooth zooming to be the same as in Firefox browser

### Actual Behavior

Lagging when zooming in Chromium

Codebreaker101:
This is what I have so far:
- the lag does not occur when DevTools are open in Chromium (interesting!)
- after adding a throttle function of 20ms on the HandlerManager.handleEvent function call the lag went away and the zooming was smooth (my screen is 60 FPS so throttle of anything higher than 17ms removes the lag)

Next I will investigate if there are more renders happening during a single frame received from `requestAnimationFrame`

Codebreaker101:
The cause of the lag is an issue in the easing functionality in the scroll_zoom.ts. Will investigate further.

Codebreaker101:
Figured it out!

https://github.com/maplibre/maplibre-gl-js/blob/0327486058079000767bac9e30111c84909c225a/src/ui/handler/scroll_zoom.ts#L293-L302

When the mouse scroll wheel is rotating fast the value of `t` is 0 which causes the easing function to return the same result, hence why the zooming stops. Eventually the value of `t` becomes >0 which causes the zoom level to change suddenly but then again it stops when  `t` becomes 0 and that is the cause of the lag. 

The fix that I found is the following:

``` ts
let t = Math.min((browser.now() - this._lastWheelEventTime) / 200, 1);
if (startZoom !== targetZoom && t === 0) {
    t = 0.1;
}
```

Number `0.1` seems to be a sweet spot for smooth zooming. 

[...]

PR #4366, opened 2024-07-05
fix lag when zooming in/out fast #4362

Codebreaker101:
This fixes lag caused by zooming in/out fast. 

Tests could not simulate the lag.

## Launch Checklist

 - [x] Link to related issues.


Codebreaker101:
After I had some trouble writing consistent test I altered my original solution. Instead of using an augmented value for `t`, when time diff is zero, just end the easing function all together since it does not have any use easing on that small time difference. It is still smooth when fast zooming but is now faster to zoom and more consistent.

HarelM:
What about inertia? Is it still supported with this fix?

Codebreaker101:
My mouse scroll gets recognized as a trackpad more often than not when using stepped scroll (mx master 3, magnetic stoppers) so testing is a bit tricky (and I don't have a mouse with regular scroll wheel). From what I did manage to test, it retained the same UX as before. Time difference in regular scroll is always > 1 so it utilizes the easing function.

When fast scrolling there is no need for easing since the scroll wheel itself has a physical inertia.

If you have a regular mouse wheel could you pull the changes and see if it does not break regular wheel zooming?

HarelM:
Sure, I'll check later this week. Thanks for the info!

HarelM:
Looks like it is working as expected using a regular mouse.
Thanks!

HarelM (approved):

PR #4376, opened 2024-07-08
Add Allmaps as available plugin

mclaeysb:
## Launch Checklist

This PR add the [Allmaps Maplibre](https://github.com/allmaps/allmaps/tree/main/packages/maplibre) plugin to the plugin list.

(also removes an extra 'enter' near the Terra Draw plugin)
(example might be coming in future PR)

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -27,7 +27,6 @@ Adds support for drawing and editing features on maps.
 <br/><small>[View on GitHub](https://github.com/mapbox/mapbox-gl-draw)</small>
 
 #### terra-draw
-
 Provides a MapLibre GL JS adapter to allow creation, selection and editing of geometries.
 <br/><small>[View on GitHub](https://github.com/JamesLMilner/terra-draw)</small>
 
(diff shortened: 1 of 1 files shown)


HarelM:
I would consider changing the layer's name to sometime different than wrapped maplibre, but that might be just me...

docs/plugins.md
```diff
@@ -27,7 +27,6 @@ Adds support for drawing and editing features on maps.
 <br/><small>[View on GitHub](https://github.com/mapbox/mapbox-gl-draw)</small>
 
 #### terra-draw
-
```
HarelM, line 30: Why was this space removed?

mclaeysb:
Wrt the comment on "wrapped maplibre": I don't see what you are pointing to, but possibly to the "WarpedMapLayer" name we use when loading the plugin from a CDN. There are actually tho ways of loading the plugin:

- [HTML example using ESM and Skypack](https://raw.githubusercontent.com/allmaps/allmaps/develop/packages/maplibre/examples/skypack.html) where we import the WarpedMapLayer class from the plugin
- [HTML example using UMD and jsDelivr](https://raw.githubusercontent.com/allmaps/allmaps/develop/packages/maplibre/examples/jsdelivr.html) where the plugin is loaded under the 'Allmaps' global, on which we call the WarpedMapLayer class.

The name 'WarpedMapLayer' is an internal class name in the plugin - a heritage from the first Allmaps plugin which was written for OpenLayers. We'll probably keep it like that.

HarelM (approved):

Issue #3983, opened 2024-04-12
Raster tiles are much lower detail when terrain is enabled


olsen232:
(This may not be a bug - if that's the case, I am just looking for info about how to control this behaviour).

Adding terrain to a map viewer causes it to load raster tiles at lower zoom, that is, load the lower detail rasters with a lower "z" value. Also, the zoom drops away much quicker when terrain is enabled, so that one end of the screen is noticeably lower detail than the other even for a reasonably small camera pitch, and worse for extreme pitches. Maps without terrain, by contrast, do a good job of maintaining the illusion that all tiles change zoom at once.

**maplibre-gl-js version**: 4.1.2

**browser**: Brave

### Steps to Trigger Behavior

 1. Set up a map with raster tiles. Turn on map.showTileBoundaries
 2. Compare it with the same map but also with 3d terrain enabled.
 3. Observe different tile zoom levels in each case.

### Link to Demonstration

Map with terrain disabled:
https://jsbin.com/mesowukici/1/edit?html,output
<img width="878" alt="Screenshot 2024-04-12 at 2 19 50 PM" src="https://github.com/maplibre/maplibre-gl-js/assets/7425442/8ee9f804-cc34-4984-9bbc-b84c27d880ab">

Identical map with terrain enabled:
https://jsbin.com/rigisawuze/1/edit?html,output
<img width="877" alt="Screenshot 2024-04-12 at 2 20 01 PM" src="https://github.com/maplibre/maplibre-gl-js/assets/7425442/f8425b99-d363-41a8-8d6c-79db29cc0f98">

### Expected Behavior
Adding terrain shouldn't affect raster tile zoom - maybe?
This may be by design for performance reasons. If so, do you know if it can be tweaked?

### Actual Behavior
Adding terrain has a large and obvious impact on raster tile zoom.

HarelM:
Yes, this was by design, I don't remember exactly why, probably related to render to texture performance. 
I'm not familiar with a way to tweak it. 
Cc: @prozessor13

olsen232:
FYI I found the most visible difference between the two is in [this line](https://github.com/maplibre/maplibre-gl-js/blob/da9ac17b09e0e793a4bcf7dc0f0672d30d8f9a83/src/geo/transform.ts#L404) in transform.ts:
`const refPoint = options.terrain ? cameraPoint : centerPoint;`

[...]

PR #4408, opened 2024-07-15
Globe - camera controls

kubapelc:
Issue & discussion of globe:

## Changes

Most of this changelog is copypasted from the previous PR (as it included these changes at first, but they were later removed for easier reviewing), but with this additional change:

- markers should now work for globe, including fading when hidden behind the planet

**Controls for globe** - I believe these should now feel natural, but here is a quick summary of how they work and what edge cases are handled:

- camera can now look at (or nearly at) the poles at any zoom level
- panning works in a way that is very similar to mercator, and converges to mercator controls at high zooms
	- this means that the pan speed of longitude is increased as the camera moves nearer to the poles
	- at low zooms however, longitude pan speed is reduced when near the poles to prevent the globe from spinning wildly
- zooming will now correctly keep the location under the cursor in the same place, as it does in mercator
	- this behaviour is not used near the planet edges, as it would move the camera too quickly and eventually cause glitches - an approximation is seamlessly used instead when cursor is near the horizon
	- zooming now works even if cursor is not on the planet surface at all - in this case the map zooms to a location near the horizon that is closest to the cursor
	- additionaly when zooming in/out with the cursor being placed too far away from the planet, the map is prevented from panning too quickly

Camera animations and functions implemented for globe projection:

- `fitBounds` - correctly computes zoom & center to fit a given box, respects to planet curvature & padding
- `flyTo`
- `easeTo`
- `jumpTo`

## Launch Checklist

 - [x] Link to related issues.
 - [ ] ~~Include before/after visuals or gifs if this PR includes visual changes.~~
 - [ ] ~~Post benchmark scores.~~
 - [ ] ~~Add an entry to `CHANGELOG.md` under the `## main` section.~~


kubapelc:
Most of the added camera tests are the old flyTo/jumpTo/easeTo test suites, but copypasted and adapted to use globe projection and with minor adjustments. Eg. instead of testing that flyTo doesn't cross the antimeridian when renderWorldCopies is disabled, globe flyTo tests that it *does* cross the antimeridian when flying from -179 to 179 longitude.

src/geo/projection/globe_transform.ts
```diff
@@ -332,34 +337,42 @@ export class GlobeTransform implements ITransform {
...
+            result.fireProjectionEvent = true;
+        }
+        return result;
+    }
+
+    private _updateErrorCorrectionValue(): void {
+        if (!this._projectionInstance) {
+            return;
+        }
+        // Note: the _globeRendering field is only updated inside this function.
```
HarelM, line 370: Consider moving this comment to be the method's comment instead of inside the method.

src/geo/projection/globe_transform.ts
```diff
@@ -221,6 +221,7 @@ export class GlobeTransform implements ITransform {
 
     private _lastLargeZoomStateChange: number = -1000.0;
     private _lastLargeZoomState: boolean = false;
+    private _lastUpdateTime = -1e9;
```
HarelM, line 224: Why not 0 or `browser.now()`?

src/geo/projection/globe_transform.ts
```diff
@@ -332,34 +337,42 @@ export class GlobeTransform implements ITransform {
...
-            forcePlacementUpdate = true;
-        }
-        this._lastGlobeRenderingState = this._globeRendering;
-        return {
-            forcePlacementUpdate
+        const result: TransformUpdateResult = {
+            forcePlacementUpdate: false,
+            fireProjectionEvent: false,
         };
+        if (oldGlobeRendering !== this._globeRendering) {
```
HarelM, line 359: consider returning one or the other in this if/else instead of initializing the result and then updating it - When first reading it I tried to understand if there's a situation where one if false and the other if true, which I don't think is possible, but writing the code this way forced me to think about it, which is not the intention, as far as I understand.

src/geo/projection/globe_transform.ts
HarelM: What's the difference between `_lastGlobeChangeTime` and `_lastUpdateTime`?
If both are needed I would consider placing them next to each other, and add comments on why they are different.
Also consider moving all these "_last..." variables into an state like object that will hold all these together, maybe, I'm not sure I understand it fully...

src/geo/transform_interface.ts
```diff
@@ -412,7 +427,14 @@ export interface IReadonlyTransform extends ITransformGetters {
      * @param y - Tile space coordinate in range 0..EXTENT.
      * @param unwrappedTileID - TileID of the tile the supplied coordinates belong to.
      */
-    isOccluded(x: number, y: number, unwrappedTileID: UnwrappedTileID): boolean;
+    isTilePositionOccluded(x: number, y: number, unwrappedTileID: UnwrappedTileID): boolean;
+
+    /**
+     * @internal
+     * Returns whether the supplied location is occluded in this projection.
```
HarelM, line 434: This has the same description as `isTilePositionOccluded`, is the only difference is the input parameters?

src/geo/transform_interface.ts
```diff
@@ -9,7 +9,41 @@ import {Terrain} from '../render/terrain';
...
+    maxzoom?: number;
+    /**
+     * `true` if tiles should be sent back to the worker for each overzoomed zoom level, `false` if not.
+     * Fill this option when computing covering tiles for a source.
+     * When true, any tile at `maxzoom` level that should be overscaled to a greater zoom will have
+     * its zoom set to the overscaled greater zoom. When false, such tiles will have zoom set to `maxzoom`.
+     */
+    reparseOverscaled?: boolean;
+    renderWorldCopies?: boolean;
+    terrain?: Terrain;
```
HarelM, line 40: Looking at this options type, I find the terrain argument weird, I'm not sure I have a good idea how to solve this, but it doesn't feel like the other "simple" options passed here...

src/ui/camera.test.ts
```diff
@@ -2296,3 +2314,1153 @@ describe('#transformCameraUpdate', () => {
...
+    });
+
+    test('changing center with apparentZoom larger should adjusts zoom and increase it', () => {
+        const camera = createCameraGlobe({zoom: 1});
+        camera.jumpTo({center: [0, 40], apparentZoom: 2});
+        expect(camera.getCenter()).toEqual({lng: 0, lat: 40});
+        expect(camera.getZoom()).toBe(1.6154999996223638);
+    });
+
+    // jumpTo tests that mimic mercator tests
```
HarelM, line 2348: Use a new `describe` statement instead of this comment, also move the `onst camera = createCameraGlobe({zoom: 1});` to beforeEach if needed.

src/ui/camera.test.ts
```diff
@@ -2296,3 +2314,1153 @@ describe('#transformCameraUpdate', () => {
         expect(fixedNum(camera.getZoom())).toBe(3);
     });
 });
+
```
HarelM, line 2317: Kodos on writing all these tests!!! Truly great work!

src/ui/camera.ts
```diff
@@ -7,15 +7,74 @@ import Point from '@mapbox/point-geometry';
...
+/**
+ * Returns `newValue` if it is:
+ *
+ * - not null AND
+ * - not negative AND
+ * - smaller than `newValue`,
+ *
+ * ...otherwise returns `oldValue`.
+ */
+function getLesserNonNegativeNonNull(oldValue: number, newValue: number): number {
```
HarelM, line 70: Should be in util I guess...?

src/ui/camera.ts
```diff
@@ -7,15 +7,74 @@ import Point from '@mapbox/point-geometry';
...
+ * @param vector - Position of the queried location on the surface of the unit sphere globe.
+ * @param toCenter - Position of current transform center on the surface of the unit sphere globe.
+ * This is needed because zooming the globe not only changes its scale,
+ * but also moves the camera closer or further away along this vector (pitch is disregarded).
+ * @param projection - The globe projection matrix.
+ * @param targetDimension - The dimension in which the scaled vector must match the target value in clip space.
+ * @param targetValue - The target clip space value in the specified dimension to which the queried vector must project.
+ * @returns How much to scale the globe.
+ */
+function solveVectorScale(vector: vec3, toCenter: vec3, projection: mat4, targetDimension: 'x' | 'y', targetValue: number): number | null {
```
HarelM, line 33: Not sure where this method should be, but probably not here...?

src/ui/camera.ts
```diff
@@ -79,14 +138,23 @@ export type CenterZoomBearing = {
...
+     * The desired mercator zoom level.
      */
     zoom?: number;
     /**
      * The desired bearing in degrees. The bearing is the compass direction that
      * is "up". For example, `bearing: 90` orients the map so that east is up.
      */
     bearing?: number;
+    /**
+     * The desired apparent zoom level, relative to the animation's starting zoom level.
```
HarelM, line 150: This might be very late to ask that, but I'll ask anyway:
Can we solve this in a different way so that when nearing the poles the zoom will look like in other places?

src/ui/camera.ts
```diff
@@ -740,16 +825,83 @@ export abstract class Camera extends Evented {
...
 
-        return {
+        const result = {
             center,
             zoom,
             bearing
         };
+
+        // If globe is enabled, we use the parameters computed for mercator, and just update the zoom to fit the bounds.
+        if (this.transform.useGlobeControls) {
```
HarelM, line 841: I would advise to move the entire content of this "if" to a method with the appropriate name.
Might be part of globe_utils maybe, not sure...

HarelM: I must have missed that in the previous review, but I think there shouldn't be a `useGlobeControls` as part of the transfrom interface if we want to keep it "globe-agnostic", right?
Maybe some of these methods that contain this if can be moved into the transfrom? or to its utils files? or helper?

src/ui/camera.ts
```diff
@@ -857,13 +1014,40 @@ export abstract class Camera extends Evented {
...
             pitchChanged = false;
 
-        if ('zoom' in options && tr.zoom !== +options.zoom) {
-            zoomChanged = true;
-            tr.setZoom(+options.zoom);
-        }
+        const optionsZoom = typeof options.zoom === 'number';
+        const optionsApparentZoom = typeof options.apparentZoom === 'number';
+
+        if (this.transform.useGlobeControls) {
```
HarelM, line 1020: Same as above.

src/ui/camera.ts
```diff
@@ -1007,55 +1183,170 @@ export abstract class Camera extends Evented {
...
+                    const speedup = Math.pow(base, 1 - k);
+                    const newCenter = unprojectFromWorldCoordinates(tr.worldSize, from.add(delta.mult(k * speedup)).mult(scale));
+                    tr.setLocationAtPoint(tr.renderWorldCopies ? newCenter.wrap() : newCenter, pointAtOffset);
+                }
+
+                this._applyUpdatedTransform(tr);
+
+                this._fireMoveEvents(eventData);
+
+            }, (interruptingEaseId?: string) => {
```
HarelM, line 1345: This is too long of a method, please see how you can split it up.

src/ui/camera.ts
```diff
@@ -1215,12 +1506,21 @@ export abstract class Camera extends Evented {
...
+     *   apparentZoom: map.getZoom()
+     * });
      * ```
      * @see [Fly to a location](https://maplibre.org/maplibre-gl-js/docs/examples/flyto/)
      * @see [Slowly fly to a location](https://maplibre.org/maplibre-gl-js/docs/examples/flyto-options/)
      * @see [Fly to a location based on scroll position](https://maplibre.org/maplibre-gl-js/docs/examples/scroll-fly-to/)
      */
     flyTo(options: FlyToOptions, eventData?: any): this {
+        if (options.around) {
+            warnOnce('flyTo does not support the "around" option.');
```
HarelM, line 1521: Is it part of the flyTo options? If so, isn't it better to remove it from the options type?

src/ui/marker.ts
```diff
@@ -526,7 +528,9 @@ export class Marker extends Evented {
     _updateOpacity(force: boolean = false) {
         const terrain = this._map?.terrain;
         if (!terrain) {
-            if (this._element.style.opacity !== this._opacity) { this._element.style.opacity = this._opacity; }
+            const occluded = this._map.transform.isLocationOccluded(this._lngLat);
```
HarelM, line 531: What about popup?

src/ui/handler_manager.ts
```diff
@@ -28,6 +32,8 @@ class RenderFrameEvent extends Event {
     timeStamp: number;
 }
 
+function createVec3(): vec3 { return new Float64Array(3) as any; }
```
HarelM, line 35: I think this method exists somewhere else as well, I would recommend moving it to utils or vec_utils, or somewhere similar.

src/ui/handler_manager.ts
```diff
@@ -490,40 +496,149 @@ export class HandlerManager {
...
+            const oldZoom = tr.zoom;
+            if (zoomDelta) tr.setZoom(tr.zoom + zoomDelta);
+            const actualZoomDelta = tr.zoom - oldZoom;
+
+            if (actualZoomDelta !== 0) {
+                // Problem: `setLocationAtPoint` for globe works when it is called a single time, but is a little glitchy in practice when used repeatedly for zooming.
+                // - `setLocationAtPoint` repeatedly called at a location behind a pole will eventually glitch out
+                // - `setLocationAtPoint` at location the longitude of which is more than 90° different from current center will eventually glitch out
+                // But otherwise works fine at higher zooms, or when the target is somewhat near the current map center.
+                // Solution: use a heuristic zooming in the problematic cases and interpolate to `setLocationAtPoint` when possible.
```
HarelM, line 530: Is there a different solution we can use? This seems a bit too hackish... :-/

src/ui/handler/scroll_zoom.ts
```diff
@@ -46,8 +45,9 @@ export class ScrollZoomHandler implements Handler {
     _lastWheelEvent: any;
     _lastWheelEventTime: number;
 
+    _lastExpectedZoom: number = null;
     _startZoom: number;
-    _targetZoom: number;
+    _targetZoom: number | null = null;
```
HarelM, line 50: Why was this changed?

src/geo/projection/globe_transform.ts
kubapelc: Fixed. Also there were a few more problems here: the comment was outdated, _globeRendering is no longer even set there, and also the function was actually called on a cloned transform, which is not refactored so that the relevant private field is set on clones directly instead.

src/geo/projection/globe_transform.ts
kubapelc: I've added comments to both and set initial values to ones derived from `browser.now()`.

src/geo/transform_interface.ts
kubapelc: Yes, shame than typescript doesn't allow method overloading (afaik). Do you think I should use a single function with the signature `isTilePositionOccluded(x: number |LngLat, y?: number, unwrappedTileID?: UnwrappedTileID): boolean;`? Though I think using two different functions is clearer in this case.

src/geo/transform_interface.ts
kubapelc: Terrain seems to be used to compute tile AABBs in 3D by querying terrain's min and max elevations for a given tile.

This type and `CoveringZoomOptions` are my attempts at parsing what the `coveringTiles` parameters mean, since I'll probably need to do at least a simple implementation of it for globe soon, even though it is assigned to Vivian. I need it mostly to make globe more presentable for the school project.

kubapelc: I've added doccomments to `terrain` and `renderWorldCopies`, since these were the only undocumented parameters in that type.

HarelM: Yeah, I don't think covering tile should be part of the transfrom - I think it should be a separate class, it might need to include the logic of which tiles to keep as well, instead of having that logic in sourceCache.
Feel free to refactor that part, probably better to have it in a separate PR though.

src/geo/transform_interface.ts
kubapelc: I can also create a new util function which converts x+y+tileID to LngLat, and only have the LngLat version of the function in the transform's interface. I think that might be the best solution?

src/geo/transform_interface.ts
HarelM: See a partial discussion about this here:
- https://github.com/maplibre/maplibre-gl-js/issues/3983

src/geo/transform_interface.ts
HarelM: Sounds good!

src/ui/camera.test.ts
kubapelc: Thanks! But most of them are really just copypastes of mercator tests :D

src/ui/camera.test.ts
kubapelc: Done (also for flyTo, easeTo, etc.)

src/ui/camera.ts
kubapelc: It is very specific to how it is used in camera.ts. I personally wouldn't be looking for such a function in utils.

src/ui/camera.ts
kubapelc: This one is also very specific to how it is used in camera.ts, so I don't think it should be accessible from elsewhere.

src/ui/camera.ts
kubapelc: I'm not entirely sure what you mean, I'm parsing it as "for a given zoom level globe/planet has the same size independent of map center latitude" (which is not the case now).

I made the choice early on that for a given x,y,z, both globe and mercator maps should look roughly the same, meaning that whatever place is in the center should have the same size for both projections. This allows for seamless transitions between them, without magically changing the map's zoom. (But panning the globe does "magically" change the zoom level.)

So now, if you were to `flyTo` some x,y,z coordinates, you will end up in the same place for both globe and mercator.

What I'm trying to solve with `apparentZoom` is when you want to move to a given x,y coordinates, but keep the same zoom level. I think users would query the zoom level before initiating the animation, and pass that as the `zoom` parameter. This would cause map to actually zoom in/out under globe projection, as long as the latitude also changes.
If the user would pass the queried zoom to `apparentZoom` instead, the map would keep the same zoom/size for both projections.

The alternative would be to expose some sort of function for recomputing zoom for globe (or recomputing it for the active projection, whatever that might be). Its signature would need to be something like `recomputeZoom(initialZoom: number, initialLocation: LngLat, zoomDelta: number, targetLocation: LngLat)`. I think using the `apparentZoom` parameter is an easier solution.

src/ui/camera.ts
kubapelc: > I would advise to move the entire content of this "if" to a method with the appropriate name.

I'll try to refactor it.

> there shouldn't be a useGlobeControls as part of the transfrom interface if we want to keep it "globe-agnostic", right?

I think it makes sense to make an exception in this one case. Globe transform needs to decide its value (moving it to projection instance is problematic, as discussed earlier).

But it is true that only globe will use "globe controls", I think other projections tend to produce flat maps and mercator controls (but using projection-specific transform functions such as `setLocationAtPoint`) would probably work fine for them.

I think transform should be focused on storing the map state and providing projection/unprojection and similar functions. Camera controls and animations are very different things.

But it is definitely not very elegant to check `useGlobeControls` in like 8 different places, but I'm not sure how to fix that... Maybe we can have a separete helper object for camera controls, which would be created in projection factory and thus can have a typed reference to GlobeTransform or GlobeProjection and thus can access fields that are not covered by the generic interfaces?

src/ui/camera.ts
HarelM: I think that as a standard, functions should either be inside the class in the file, or in a different file.
"dangling" function like those create the wrong feeling when you read the file as you are not sure there are other function "outside" the class or not.
This is mainly for keeping the files clean and descriptive - i.e. if the file has a name camera.ts it should only have the camera class and maybe some other types that are related to API.

src/ui/camera.ts
HarelM: Why not move it to be a static (or just private) method inside the camera class then?

src/ui/camera.ts
HarelM: > "for a given zoom level the globe/planet has the same size independent of map center latitude" (which is not the case now)

This is what I mean, yes.
I tend to think that people won't know that they need to use `apparentZoom` and simply open bugs on this, similar to what we have with `freezeElevation` for terrain - although the defaults there are a bit better, maybe.
I'm not sure I know how to properly solve this honestly.
I agree that the transition to mercator should be smooth, but maybe we can do it only when they are already close enough in terms of transition?

In the following example of cesium you can switch between globe and mercator (they have a nice animation, but otherwise bad UX), and globe the globe doesn't get bigger when you look at the poles.
https://sandcastle.cesium.com/

In any case, as I said, maybe it's too late to change that, but we'll need to think of a more intuitive solution...

src/ui/camera.ts
HarelM: That might be a good elegant solution, it would be great to keep the code that is only related to globe controls in a different place and avoid these if as much as possible, or if we still keep them, have a one liner inside them and that this code is not part of the camera class, but elsewhere, `GlobeCamera` or `GlobeCameraControls` class or file.

src/geo/transform_interface.ts
kubapelc: Done! (I'll push all the changes later.)

src/ui/camera.ts
kubapelc: This makes sense, I've changed it to a private static function.

src/ui/camera.ts
kubapelc: Done, same as with the other function.

src/ui/camera.ts
kubapelc: I'd like to keep the "globe changes z along with latitude" behaviour, even though it makes some animation config unintuitive. And the only case that is unintuitive is when you want to move to a given x,y *and* change zoom level by some delta. Everything else behaves intuitively, including not passing any zoom at all to the options - then the globe just pans, without changing its size, and mercator pans without changing z. The resulting z of the animation is different depending on projection, but the map makes the expected motion for both.

On the other hand, if globe used the same z and same planet size for any latitude, then an animation from, say, xyz of 0,0,0 to 0,80,4 would result in wildy different maps depending on whether the projection was globe or mercator. Also copypasting x,y,z from any other web map to maplibre globe would also result in a different zoom than expected.

I'm now not sure whether to even keep `apparentZoom`, since its usage is very limited, and it is hard to explain what it even does. Maybe just explaining how zoom works on globe somewhere and linking that from any docs that use zoom would be better? The function to compute zoom at a different latitude that keeps the same globe size isn't even that complex, its just `newZoom = oldZoom + log2(cos(newLatitude) / cos(oldLatitude))`.

src/ui/marker.ts
kubapelc: They don't seem to change opacity when behind terrain like markers do. Should popups react to being on the backfacing side of the planet?

kubapelc: As for me, the answer is... maybe? If a popup is displayed only when clicking something, then they probably don't need to have extra logic, if someone places a lot of them on a map, then they should fade when occluded, I guess.

src/ui/map.ts
```diff
@@ -3109,6 +3109,10 @@ export class Map extends Camera {
         const transformUpdateResult = this.transform.newFrameUpdate();
         this._placementDirty = this.style && this.style._updatePlacement(this.transform, this.showCollisionBoxes, fadeDuration, this._crossSourceCollisions, transformUpdateResult.forcePlacementUpdate);
 
+        if (transformUpdateResult.fireProjectionEvent) {
+            this.fire(new Event('projectiontransition'));
```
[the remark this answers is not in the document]
kubapelc: I'm not sure. I've added the event, because I needed to somehow trigger an update on all markers when globe changes its internal projection to mercator, otherwise markers would remain in wrong positions until the map moved. Adding the event seemed to be the easiest way to do that, but I'm really not sure it is the best way. Do you have an idea on how else to trigger markers to update? Changing the projection could generate a sequence of movestart, move, moveend events?

src/ui/camera.ts
HarelM: Maybe the globe example should have this code so that people that look for a globe example will find it there? Documenting it sounds like a better solution from my point of view.

src/ui/marker.ts
HarelM: Let's keep it as it is now, if someone is interested in improving the UX, then they are welcome to open a PR.

src/ui/map.ts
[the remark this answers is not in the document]
HarelM: I think raising an event, like done for terrain, makes sense. My question was mainly about other parameters that should be sent as part of the event, like current protection, previous projection, and other stuff that might be interesting to the person who registers to this event.

src/ui/camera.ts
kubapelc: My progress implementing a projection-dependent "CameraHelper" is here: https://github.com/windycom/maplibre-gl-js/tree/kubapelc/globe-pr-controls-dev

It is taking a lot of time and I'm not sure it is as elegant as it sounded at first. A lot of the `useGlobeControls `branches reuse a lot of things that also get computed for mercator, and it is difficult to separate it into two CameraHelper implemetations without repeating a lot of code. I'm not sure if the changes are all positive so far.

HarelM: Can you maybe open a PR against your branch so I can have a better understand of the changes? I'm not sure I'm looking at the right code...

kubapelc: https://github.com/windycom/maplibre-gl-js/pull/1

I'm a bit less doubtful now, but I'm still curious what you think. The adaptation of handler_manager.ts to the CameraHelper classes is the most convoluted. I think I can work around that one in particular by refactoring how mercator panning and zooming work.

HarelM: Sorry for the delay in the response.
I think this looks good in general, it split stuff according to globe vs Mercator, I like it.

kubapelc: Thank you for checking it out! I'll continue working on it then. No worries about the delay, I had to focus on other things in the meantime.

src/ui/handler_manager.ts
kubapelc: I think there is no solution that wouldn't be similarly "hacky". This solution uses accurate zooming to cursor as much as possible, while avoiding the problems it would bring if it were used all the time.

If you were looking at the whole planet and tried zooming somewhere at the horizon of the planet, it would spin wildly to keep the same location under the cursor. If you tried to zoom in with the cursor not on the planet surface at all, then however you choose to handle this case, you will have to switch to accurate zooming eventually, because the planet will keep getting bigger as you zoom in and will eventually cover the cursor.

I tried to make something that works and feels okay without abrupt motions and stuff. I have no idea how to do this properly, I just went the "best effort" way. If someone who knows what they are doing wants to modify this in the future, it should be doable without touching anything else in the globe implementation.

HarelM: Fair enough, thanks for looking into it.

src/geo/transform_interface.ts
kubapelc: I don't want to change this logic any further, at least not in this PR.

I think transform is not the worst place for coveringTiles right now, since the function is projection dependent, so having it in transform allows us to easily have projection-specific implementations.

src/ui/camera.ts
kubapelc: I've finished the refactor, I'll push the changes shortly.

src/geo/transform_interface.ts
HarelM: Covering tiles is should be different for different sources.
But I agree it should not be part of this PR, just something to take into consideration.

src/ui/camera.ts
kubapelc: I think this method was shortened by the CameraHelper refactor.

src/ui/map.ts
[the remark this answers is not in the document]
kubapelc: I've added `newProjection` field to the event, I've also added unit tests to test that the event is fired properly.

src/ui/handler_manager.ts
kubapelc: I've moved the functions to util.ts and renamed them to `createVec3f64`, etc., to better specify their purpose.

src/ui/handler/scroll_zoom.ts
kubapelc: I thought using `null` as a special value indicating that no zoomscroll is active was better than `delete`ing the field, as was done previously. But I guess the `| null` in the type is superfluous?

src/ui/camera.ts
kubapelc: The comment is a little ambiguous, do you mean getting rid of `apparentZoom` and documenting `newZoom = oldZoom + log2(cos(newLatitude) / cos(oldLatitude))` somewhere?

HarelM: Yes, present this code only in one of the globe example.

src/ui/map_tests/map_events.test.ts
```diff
@@ -780,4 +781,64 @@ describe('map events', () => {
...
+    describe('projectiontransition event', () => {
+        test('projectiontransition events is fired when setProjection is called', done => {
+            const map = createMap();
+            map.on('load', () => {
+                let stage = 0;
+                const expected = [
+                    'globe',
+                    'mercator',
+                ];
+                map.on('projectiontransition', event => {
```
HarelM, line 794: Can you use a spy here instead of this function and test all the expected values on the spy?

src/ui/map_tests/map_events.test.ts
```diff
@@ -780,4 +781,64 @@ describe('map events', () => {
...
+                });
+                map.setProjection({
+                    type: 'mercator',
+                });
+            });
+        });
+        test('projectiontransition is fired when globe transitions to mercator', done => {
+            const map = createMap();
+            jest.spyOn(GlobeProjection.prototype, 'updateGPUdependent').mockImplementation(() => {});
+            map.on('load', async () => {
```
HarelM, line 812: use `await map.once('load')` instead of this.
As a general rule, I would like to avoid using `done` in jest tests as much as possible due to debuggability issues.

src/ui/handler/scroll_zoom.ts
HarelM: Other fields are also deleted, so I think keeping it as it was before is better than introducing a new way.

src/ui/camera.ts
```diff
@@ -310,6 +329,7 @@ export abstract class Camera extends Evented {
         this._zooming = false;
         this.transform = transform;
         this._bearingSnap = options.bearingSnap;
+        this.cameraHelper = new MercatorCameraHelper();
```
HarelM, line 332: Should this be part of the constructor like transform?

src/ui/camera.ts
```diff
@@ -165,6 +179,10 @@ export type FlyToOptions = AnimationOptions & CameraOptions & {
 export type EaseToOptions = AnimationOptions & CameraOptions & {
     delayEndEvents?: number;
     padding?: number | RequireAtLeastOne<PaddingOptions>;
+    /**
+     * If `zoom` is specified, `around` determines the point around which the zoom is centered.
+     */
+    around?: LngLatLike;
```
HarelM, line 185: Is this relevant only to ease to? what about flyTo? jumpTo?

src/ui/camera.ts
```diff
@@ -853,19 +823,21 @@ export abstract class Camera extends Evented {
...
-            zoomChanged = true;
-            tr.setZoom(+options.zoom);
-        }
+        const oldZoom = tr.zoom;
 
-        if (options.center !== undefined) {
-            tr.setCenter(LngLat.convert(options.center));
+        // The jumpTo function is sometimes called before the projection and cameraHelper is initialized.
+        // Fall back to mercator behavior for such calls.
+        if (this.cameraHelper) {
```
HarelM, line 833: How is this possible? cameraHelper is initialized in the constructor, right?
If it's only a problem with projection, then please change the comment.

src/ui/camera.test.ts
```diff
@@ -2296,3 +2319,1165 @@ describe('#transformCameraUpdate', () => {
...
+                bearing: 180,
+                pitch: 60
+            });
+            expect(camera.getCenter()).toEqual({lng: 10, lat: 20});
+            expect(camera.getZoom()).toBe(10);
+            expect(camera.getBearing()).toBe(180);
+            expect(camera.getPitch()).toBe(60);
+        });
+
+        test('emits move events, preserving eventData', done => {
```
HarelM, line 2424: Avoid using `done` if possible, also this is called at the end of the method, so I'm not sure it's needed.

src/ui/camera.ts
kubapelc: I've added a new example `globe-flyto-zoom-delta.html` to demonstrate how to adjust zoom on globe, and removed `apparentZoom` from all camera animations.

src/ui/handler/scroll_zoom.ts
kubapelc: I've reverted this to use `delete`.

src/ui/map_tests/map_events.test.ts
kubapelc: Done for both projection event tests.

src/ui/map_tests/map_events.test.ts
kubapelc: Done for both projection event tests.

src/ui/camera.ts
kubapelc: Yes, that makes more sense, fixed.

src/ui/camera.ts
kubapelc: Originally `around` was in the options for all animations, but only easeTo actually used it, so I moved it here.

src/ui/camera.test.ts
kubapelc: I've removed all such usages of `done()` that I found in camera.test.ts

src/ui/camera.test.ts
```diff
@@ -2296,3 +2319,1165 @@ describe('#transformCameraUpdate', () => {
...
+                });
+
+            camera.easeTo({center: [0, 20], duration: 0});
+        });
+
+        test('does not emit zoom events if not changing latitude and not zooming', done => {
+            const camera = createCameraGlobe();
+
+            camera
+                .on('zoomstart', () => { done('zoomstart failed'); })
```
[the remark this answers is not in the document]
kubapelc: Fixed.

HarelM:
Try changing the headless parameter in the image-generation script, I think it works in some OSs in one way and doesn't in other OSs.
If this doesn't work let me know.

src/ui/map_tests/map_events.test.ts
```diff
@@ -780,4 +781,52 @@ describe('map events', () => {
...
+                });
+                expect(spy).toHaveBeenCalledTimes(2);
+                expect(spy).toHaveBeenNthCalledWith(1, {newProjection: 'globe'});
+                expect(spy).toHaveBeenNthCalledWith(2, {newProjection: 'mercator'});
+            });
+        });
+        test('projectiontransition is fired when globe transitions to mercator', async () => {
+            const map = createMap();
+            jest.spyOn(GlobeProjection.prototype, 'updateGPUdependent').mockImplementation(() => {});
+            await map.once('load', async () => {
```
HarelM, line 805: There's no need to add a callback when using `await`.
you can 
```ts
await map.once('load');
// the code after this line is executed after the map was loaded
// which reduces indentation and makes things easier to read
```

src/ui/map_tests/map_events.test.ts
```diff
@@ -780,4 +781,52 @@ describe('map events', () => {
...
+                const spy = jest.fn();
+                map.on('projectiontransition', spy);
+
+                map.setProjection({
+                    type: 'globe',
+                });
+                map.setZoom(18);
+                map.redraw();
+                await sleep(1500);
```
HarelM, line 814: This is a sleep of 1.5 sec - it's a pretty high number for unit test, making this test run more than 3 seconds, which is a bit high.
is 100 not enough in this case?

test/examples/globe-flyto-zoom-delta.html
```diff
@@ -0,0 +1,82 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>Camera animations with zoom delta on a globe</title>
```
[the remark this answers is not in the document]
kubapelc: I've made the title and description more descriptive (I hope) and also renamed the example.

src/ui/map_tests/map_events.test.ts
kubapelc: Fixed. I mostly understood the previous `await map.once` suggestion as a way to keep the test from exiting before all assertions had a chance to run.

src/ui/map_tests/map_events.test.ts
kubapelc: The transition animation for globe takes 500 ms. I used 1500 ms just to be safe, but 550 ms also seems to work fine, so I changed it to that.

kubapelc:
> Try changing the headless parameter in the image-generation script

Without headless mode the browser opens and loads the page, but it also uses the unpkg release of MapLibre instead of the dev version, so I had to work around that. It also doesn't seem to ever take the screenshot. So I just created the image manually.

I also ran most of the globe example images through compresspng, with the exception of the atmosphere example image, since that has a lot of gradients that get degraded with an 8 bit palette.

HarelM (approved):

PR #4415, opened 2024-07-17
docker based build

### LINKED ISSUE #57: Suggest docker solution to help on setting up build/test/run environment
## Motivation

As a developer, I'd like to be able to quickly manage environment without changing my own or spending time on it.

## Design

Dockerfile and docker-compose.yml setup could help a lot to build with simple commands as:

Build environment: ```docker-compose build base```

Debug: ```docker-compose run debug```

Make packages (single js, minify, and so on): ```docker-compose run package```

clement-igonet:
* Setup to buils maplibre-gl-js with `docker compose build` command.
* For development purpose: `docker compose up`
* related issue: /link https://github.com/maplibre/maplibre-gl-js/issues/57

Fixes #57


louwers:
Can this run the browser tests? I had some trouble with that on an M1 Mac.

docker-compose.yml
```diff
@@ -0,0 +1,15 @@
+services:
```
louwers, line 1: Does a docker-compose really make sense with only one service?

Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM --platform=$BUILDPLATFORM node:lts-bookworm-slim
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+RUN echo "hello world"
+
+RUN apt-get update && apt-get install -y \
```
louwers, line 7: The `.devcontainer` should use the Dockerfile, otherwise this is duplicated.

HarelM: I don't think this docker container is published, so I'm not sure .devcontainer can use this docker.
I would reccoment trying to find a container that already has most of these installed and simply use it instead of installing everything.

Dockerfile
```diff
@@ -0,0 +1,33 @@
+FROM --platform=$BUILDPLATFORM node:lts-bookworm-slim
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+RUN echo "hello world"
```
HarelM, line 5: This shouldn't be here

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+  libpixman-1-dev \
+  libcairo2-dev \
+  libpango1.0-dev \
+  libgif-dev \
+  && rm -rf /var/lib/apt/lists/*
+
+COPY ./ /maplibre-gl-js/
+
+WORKDIR /maplibre-gl-js
+RUN npm install
```
HarelM, line 30: I guess this should be `npm ci`.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+  build-essential \
+  git \
+  libglew-dev \
+  libxi-dev \
+  default-jre \
+  default-jdk \
+  xvfb \
+  && rm -rf /var/lib/apt/lists/*
+
+RUN apt-get update && apt-get install -y \
```
HarelM, line 17: Why is this split into two lines? Why not installing everything at one?

docker-compose.yml
HarelM: I would reccomnd to add an entrypoint to the docker file by using `npm run start` and export `9966` port.
Also add some explanation to the `CONTRIBUTING.md` file on how to use this image.

clement-igonet:
> Can this run the browser tests? I had some trouble with that on an M1 Mac.

Good question.
What is the npm command to launch for that ?
I'm ready to have a try on my side, and to look for a fix if needed.

HarelM:
Checkout the CI code, it has most of what the CI is running, which is stuff in most cases you should run locally to make sure the PR checks pass.
npm run test-unit, npm run test-integration, npm run test-render

docker-compose.yml
clement-igonet: > Does a docker-compose really make sense with only one service?

OK. My original idea is to make run the dev version in watch mode, behind a web server (caddy_server by default).

clement-igonet: > Does a docker-compose really make sense with only one service?

I already imagine at least 1 or 2 other services (so I start by a first):
* a headless web browser for testing purposes
* a web server to mak run a release with some examples

Dockerfile
clement-igonet: > Why is this split into two lines? Why not installing everything at one?

Well, I'm just following documentation from CONTRIBUTING.md : https://github.com/maplibre/maplibre-gl-js/blob/main/CONTRIBUTING.md#linux-and-by-extension-github-codespaces

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+  libcairo2-dev \
+  libpango1.0-dev \
+  libgif-dev \
+  && rm -rf /var/lib/apt/lists/*
+
+COPY ./ /maplibre-gl-js/
+
+WORKDIR /maplibre-gl-js
+RUN npm install
+RUN npm run build-css
```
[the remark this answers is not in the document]
clement-igonet: > You can run `npm run dist`

I don't know this command.

In my case, it just returns:
```
npm error Missing script: "dist"
```

Dockerfile
clement-igonet: > I would reccoment trying to find a container that already has most of these installed and simply use it instead of installing everything.

I'm sure to get this specific point. Are you talking about `.devcontainer` stuff ?

clement-igonet:
> npm run test-unit, npm run test-integration, npm run test-render

Only `npm run test-unit` runs well for the moment.
I suggest to focus on `npm run test-integration` and `npm run test-render` in another step

Dockerfile
louwers: You can use a Dockerfile https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/adding-a-dev-container-configuration/introduction-to-dev-containers#dockerfile

docker-compose.yml
HarelM: I still don't see a huge value here...
You can simply mount the entire cloned repo instead of a selective number of folders.
You can switch the entry point command to run different scripts from the package.json file. 
Bottom line, this can be dramatically simplified.

Dockerfile
HarelM: Probably a mistake there too, or done with readability in mind.
We should aim at reducing docker layers.

Dockerfile
clement-igonet: OK, good point.

I did not consider to integrate docker facilities so far away, I just though about not having to change my environment (installing node and dependencies).

So, I do not aim to manage .devcontainer in that specific P.R.

Is it something mandatory here ?

Dockerfile
clement-igonet: OK, it will be done today.

HarelM:
Not sure what you mean by another step, but if the docker image is not good at the moment, I suggest to fix it before merging it, otherwise it misses the point a bit...
Also the part about devcontainer is worth looking into as part of this PR.

clement-igonet:
> Not sure what you mean by another step, but if the docker image is not good at the moment, I suggest to fix it before merging it, otherwise it misses the point a bit... Also the part about devcontainer is worth looking into as part of this PR.

I"ll do my best. github CI is not a part I know well.

HarelM:
It might be interesting to publish this dev container to ghcr for every push to main so that people won't need to build it locally...

clement-igonet:
> It might be interesting to publish this dev container to ghcr for every push to main so that people won't need to build it locally...

OK. In that case, do we already have a github docker repository for maplibre ?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+
+ARG USERNAME=node
+ARG NPM_GLOBAL=/usr/local/share/npm-global
+
+# Add NPM global to PATH.
+ENV PATH=${NPM_GLOBAL}/bin:${PATH}
+
+RUN \
+    # Configure global npm install location, use group to adapt to UID/GID changes
+    if ! cat /etc/group | grep -e "^npm:" > /dev/null 2>&1; then groupadd -r npm; fi \
```
HarelM, line 15: Why not use a node docker image as a base image instead of all this complexity?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+    && usermod -a -G npm ${USERNAME} \
+    && umask 0002 \
+    && mkdir -p ${NPM_GLOBAL} \
+    && touch /usr/local/etc/npmrc \
+    && chown ${USERNAME}:npm ${NPM_GLOBAL} /usr/local/etc/npmrc \
+    && chmod g+s ${NPM_GLOBAL} \
+    && npm config -g set prefix ${NPM_GLOBAL} \
+    && su ${USERNAME} -c "npm config -g set prefix ${NPM_GLOBAL}" \
+    # Install eslint
```
HarelM, line 24: Not sure why you would need to install eslint globally, it is installed locally as part of npm ci command.

.devcontainer/devcontainer.json
```diff
@@ -1,4 +1,5 @@
 {
     "name": "MapLibre container using node.js",
+    "build": { "dockerfile": "Dockerfile" },
```
HarelM, line 3: I would assume that using a docker image would avoid the need for all the below installations. 
Also, if we publish an image as part of the commits to main I would expect to be able to use this image for codespaces...?

.github/workflows/release.yml
```diff
@@ -6,6 +6,26 @@ on:
   workflow_dispatch:
 
 jobs:
+  push-store-image:
```
HarelM, line 9:
```suggestion
  push-dev-image:
```

.github/workflows/release.yml
```diff
@@ -6,6 +6,26 @@ on:
...
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: './devcontainer'
+    steps:
+      - name: 'Checkout GitHub Action'
+        uses: actions/checkout@main
```
HarelM, line 16: Better to use a version, like in other places in the CI.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+    python3 \
+    python-is-python3 \
+    pkg-config \
+    libpixman-1-dev \
+    libcairo2-dev \
+    libpango1.0-dev \
+    libgif-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+CMD npm ci && npm run start
```
HarelM, line 47: Add npm run build-dist here.
npm run start should be the entry point to allow replacing it with other commands.
This will allow for better ergonomics and avoid the need for docker compose.

clement-igonet:
A clarification point: what is expected for CI ?
What released docker image should do ?
* Just contains environment to run any node comment for maplibre ?
* Contains also source code ?
* Contain only distribution folder content ?

.devcontainer/Dockerfile
clement-igonet: OK. fixed.

.devcontainer/Dockerfile
clement-igonet: OK, fixed.

.github/workflows/release.yml
clement-igonet: OK, applied.

.github/workflows/release.yml
clement-igonet: > Better to use a version, like in other places in the CI.

OK. So I used this:
```
ghcr.io/maplibre/maplibre-gl-js:${{ steps.package-version.outputs.current-version }}
```

.devcontainer/Dockerfile
clement-igonet: OK, done

.github/workflows/release.yml
HarelM: I meant v4 or so for the checkout action...

HarelM:
CI should stay the same, besides pushing an image for every commit to main. 
Docker image should contain everything to allow running any maplibre script written in package.json (besides docker within a docker I guess).
It should contain the src and dist, and should allow a developer to replace them with their code.

clement-igonet:
> CI should stay the same, besides pushing an image for every commit to main. Docker image should contain everything to allow running any maplibre script written in package.json (besides docker within a docker I guess). It should contain the src and dist, and should allow a developer to replace them with their code.

I'll work on that way.
However, in term of use case, here is what I have in mind:
* I'd like to rely on an environment that is easy to reproduce (in dev environment or production environment)
  * So a Dockerfile without sources is welcome, for dev and CI purposes
* As a developer, I'd like to make run maplibre from my own code
  * So a Dockerfile without sources is enough, as docker or docker-compose can bind the current local code inside the docker container
* As a user, I'd like to have a try with any version of maplibre-gl-js library
  * In that way, I do not need docker as I can manage maplibre-gl-js version in html source code.
  
=> So I would suggest not to include source code inside provoded docker image
  * It does not really cover a use clearly defined case
  * It would provide a big docker image
  
What do you think ?

* I'd like github CI to be able to take benefit

HarelM:
I don't think it will change the size dramatically, so the argument is weak. 
Since you can override the code I don't think it will matter much.
In order to run a debug version (npm run start) you'll need the code, otherwise it won't work. 
And I think that for completeness and simplicity it's better to include it. 

We can always remove it later on, I suggest to include it, see how it feels and then see if we truly need to optimize it.

clement-igonet:
F.Y.I.: For testing purpose, I realize that google-chrome must be installed.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,38 @@
+ARG VARIANT=20-bookworm-slim
```
HarelM, line 1: Since we are using puppeteer for e2e tests I would recommend a base image that already has chrome and node like:
https://pptr.dev/guides/docker
And avoid all these complexity of installing chrome etc.

HarelM:
I would consider starting from the following docker image and see if there are missing stuff that needs to be added instead of bootstrapping an image:
[cypress/browsers](https://hub.docker.com/r/cypress/browsers/tags).

HarelM:
This looks stale.
Closing, let me know if you would like to continue working on it and I'll reopen.

PR #4425, opened 2024-07-18
Ignore `style.loadURL` abort error when style is removed

### LINKED ISSUE #4400: `style.loadURL` prints an error when the request is aborted

**browser**: Chrome

### Steps to Trigger Behavior

 1. Create a map with a style URL
 2. Use `map.setStyle` immediately after (to make sure the first style request is aborted)
 3. An error log appears in the browser console
 
> `map.setStyle` is one way to trigger this error, but other methods can too, as long as it calls `style._remove`.

### Link to Demonstration

https://jsbin.com/garasukubu/edit?html,output

### Expected Behavior

No error should be printed in the console.

### Actual Behavior

A `DOMException: signal is aborted without reason` error is printed in the console.

madoci:
Fixes #4400 

Ignore the abort error raised by `style.loadURL` when the style is removed.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
Looks like a good and clean solution, thanks!

src/style/style.ts
```diff
@@ -290,12 +290,13 @@ export class Style extends Evented {
 
         const request = this.map._requestManager.transformRequest(url, ResourceType.Style);
         this._loadStyleRequest = new AbortController();
+        const abortController = this._loadStyleRequest;
```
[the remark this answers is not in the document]
HarelM: I thought that might be the case, fair enough then...

madoci:
Thank you @HarelM for the review. Should I open a new PR to add a test of the error firing in case it isn't an abort error, to ensure converage of the `if` condition ?

HarelM:
Here would be better. THANKS!

src/style/style.test.ts
```diff
@@ -184,6 +184,20 @@ describe('Style#loadURL', () => {
 
         expect(spy).toHaveBeenCalledTimes(0);
     });
+
+    test('fires an error if the request fails', done => {
```
HarelM, line 188: Can you change it to use `once`, `async` and promise?
`done` tests are horrible to debug when they fail...

src/style/style.test.ts
```diff
@@ -172,6 +172,32 @@ describe('Style#loadURL', () => {
         style._remove();
         expect((server.lastRequest as any).aborted).toBe(true);
     });
+
+    test('does not fire an error if removed', async () => {
+        const style = new Style(getStubMap());
+        const spy = jest.fn();
+
+        style.on('error', spy);
```
HarelM, line 180: you can use `once` and expect that the promise will not be rejected, instead of `sleep(0)`

HarelM:
Besides the test nit-picking this looks good.

HarelM:
Besides the test nit-picking this looks good.

src/style/style.test.ts
madoci: I'm not sure how to do that. Since we expect the `style.once('error')` promise to neither resolves nor rejects (because `this.fire` should never be called), we can't await it. And since the abort happens asynchronously, we can't test the spy function immediately after removing the style. Ideally, I would wait for the abort error to be thrown, but I couldn't find a way to do that with the nise fake requests.

Any advice would be welcomed, as I'm not very familiar with these types of tests.

HarelM: Ahh, I see, I misread the code, can you change it to `not.toHaveBeenCalled` instead of times 0?

HarelM: Also, I think the following example raises an abort signal: 
https://github.com/maplibre/maplibre-gl-js/blob/6a996a11958a85b6a2192b4f9957b5b27348b495/src/source/vector_tile_worker_source.test.ts#L35

I'm not 100% sure as I'm not sure this is actually documented...

madoci: Of course, I'll change it to `not.toHaveBeenCalled`.

I don't fully understand this test, but I think one key difference is that `source.loadTile` ans `source.abortTile` both return promises that can be awaited, while `style.loadURL` and `style._remove` don't. So I'm not sure the same method can be applied here.

HarelM: I think calling this abort url will cause an abort signal, which you wrote you needed, not that important though.

madoci: After doing some test on `vector_tile_worker_source.test.ts`, it seems replacing the URL by any string doesn't change the test result. So I think it's just an arbitrary fake URL.

I managed to make it work with this however : 
```js
style.on('error', spy);
style.loadURL('style.json');
const promise = new Promise((_, reject) => {
    (server.lastRequest as any).addEventListener('abort', reject);
});
style._remove();
await promise.catch(() => {});

```
It might not be the prettiest code, but just awaiting the promise doesn't work, a `then` or `catch` is needed. It can also be done like that : 
```js
const promise = new Promise((resolve) => {
    (server.lastRequest as any).addEventListener('abort', resolve);
});
...
await promise.then(() => {});
```
IMO, I find the `reject/catch` to be cleaner, because it mirrors more accurately the behaviour of the request, and adding a catch also serves the purpose of ignoring the error.

I will commit with these changes as I won't be able to work on it for a few days. I'll come back to it later if it needs other changes.

HarelM (approved):
Nice!

src/style/style.test.ts
```diff
@@ -172,6 +172,35 @@ describe('Style#loadURL', () => {
...
+        style.loadURL('style.json');
+        const promise = new Promise((_, reject) => {
```
HarelM, line 182: I see other places are using a simpler method:
`expect((server.lastRequest as any).aborted).toBe(true);`
I think it can be used here as well.
I'll open a PR to update the types so we can remove these "as any".

madoci: `expect((server.lastRequest as any).aborted).toBe(true);` is used to test if the request has been aborted, which the test above does (`'cancels pending requests if removed'`). In our case, we want to test that, when the request is aborted, no error are fired.

Removing these "as any" would be nice, thanks.

HarelM: Using `expect((server.lastRequest as any).aborted).toBe(true);` would remove the need to use this promise, and wait for it rejection. I find it more elegant. You might still need a `sleep(0)` though.

HarelM: I've updated the nise types, now you can use it without the "as any":
https://github.com/maplibre/maplibre-gl-js/pull/4446

madoci: > I've updated the nise types, now you can use it without the "as any"

Thanks !

> Using expect((server.lastRequest as any).aborted).toBe(true); would remove the need to use this promise, and wait for it rejection. I find it more elegant. You might still need a sleep(0) though.

We would still need a `sleep(0)` since the rejection is asynchronous. Without it, the test succeed even if the Style fires the abort error. The promise is just a way to wait for the rejection we want to test rather than just wait for a bit with the `sleep(0)`, but I agree it's not very elegant and it might be too bound to implementation. Should I revert to `sleep(0)` ?

HarelM: Yes please.

HarelM (approved):

PR #4431, opened 2024-07-19
[#4340] Fixes possible access to undefined object

### LINKED ISSUE #4340: TypeError: Cannot read properties of undefined (reading 'sky')
The error comes from the line https://github.com/maplibre/maplibre-gl-js/blob/main/src/render/painter.ts#L412

The stylesheet object doesn't exist when we call `map.redraw()` before the style is loaded. I guess the line should be 
```
if (this.style.sky) drawSky(this, this.style.sky);
```

**browser**: all

### Steps to Trigger Behavior

 1. Create a map 
 2. Call map.redraw()
 3.

### Link to Demonstration
https://codepen.io/oterral/pen/PovvbJx

### Expected Behavior

No error about sky property

### Actual Behavior

A TypeError is triggered in the console

birdofpreyru:
Fixes regression in the latest [v4.5.0](https://github.com/maplibre/maplibre-gl-js/releases/tag/v4.5.0) release &mdash;
Fixes #4340.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
Any chance you could create a unit test to cover this scenario?

birdofpreyru:
@HarelM I have added the test.

I also got a better idea, why the problem happened &mdash; when a new `Style` instance is set on the map, the painter's `render()` method might be called before the style's asynchronous `loadURL()` operation completes, and the `stylesheet` value is attached to the style only when `loadURL()` completes, inside the `_load()` method it calls.

It makes me think, perhaps there is a larger problem, and in the perfect world, something should be synchronized better, so that the style is not set on the map before the style is loaded, or render() is not called before the map style is loaded... but regardless, my PR hotfixes the problem with library crashing.

HarelM (approved):
THANKS

PR #3710, opened 2024-02-16
Subpixel positioning for popup

ericmagnuson:
Hey, everyone!

This PR implements the idea [in this Discussion](https://github.com/maplibre/maplibre-gl-js/discussions/3614), allowing subpixel positioning of a `Popup`.  This addition allows a `Popup` to be animated more smoothly. The behavior is optional and can be set at instantiation or via class methods.  Please see the discussion for screen recordings of the behavior before and after.

### Details
A `subpixelPositioning` option (a boolean) is added to the `Popup`.  If the boolean is true, the offset rounding that happens when setting the CSS `translate` property of the `Popup` is bypassed.  The option can also be configured using a getter method, `Popup#setSubpixelPositioning`.

### Before
https://github.com/maplibre/maplibre-gl-js/assets/479605/369b3def-ceb4-4983-b493-2b2c94f847a4

### After
https://github.com/maplibre/maplibre-gl-js/assets/479605/cee14f85-3062-4b37-938b-c558360dc81f

## Launch Checklist

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.
 - [x] Link to related issues.
 - [x] Include before/after visuals or gifs if this PR includes visual changes.
 - [x] Write tests for all new functionality.
 - [x] Document any changes to public APIs.
 - [ ] Post benchmark scores.
 - [x] Add an entry to `CHANGELOG.md` under the `## main` section.

PR #4458, opened 2024-07-26
Fix Markers jumping in 1 pixel due to sub pixel percision

### LINKED ISSUE #1014: [Bug] Markers jump in 1 pixel after map zooming
Greetings 👋

## The problem

I found weird markers behaviour. After map zooming they jump in 1px. 

<details>

<summary>gif</summary>

</details>

## Reason

It happens because screen coordinates are rounded after map movement ([see](https://github.com/maplibre/maplibre-gl-js/blob/964bbef052cf787b7e00f6fa4eeca3e14f340828/src/ui/marker.ts#L467)).

Reproducible on:
- Macbook Pro 2015 15
- macOS, Google Chrome 98
- mapbox-gl v1 / maplibre-gl-js v1 / maplibre-gl-js v2

But it should be noticeable everywhere due to subpixel rendering.

## Solution

Round screen coordinates during movement taking into account [Window.devicePixelRatio](https://developer.mozilla.org/ru/docs/Web/API/Window/devicePixelRatio). Theoretically it should solve [stuttering during zooming](https://github.com/maplibre/maplibre-gl-js/blob/964bbef052cf787b7e00f6fa4eeca3e14f340828/src/ui/marker.ts#L464) (which happens on HiDPI screens only?).

```ts
const pixelRatio = (typeof window !== 'undefined' && window.devicePixelRatio) || 1;

export function roundPixel(pixel: number): number {
  return Math.round(pixel * pixelRatio) / pixelRatio;
}
```

```ts
 DOM.setTransform(this._element, `${anchorTranslate[this._anchor]} translate(${roundPixel(this._pos.x)}px, ${roundPixel(this._pos.y)}px) ${pitch} ${rotation}`);
```

JeroendeJong:
This pull request aims to address the issue raised in #1014. The implementation is aimed to be inline with https://github.com/maplibre/maplibre-gl-js/pull/3710 by adding a `setSubpixelPositioning()` API method to the marker. When setting this to `true` the position will not rounded. Closes #1014 

The suggestions in the issue don't fully resolve the issue, because it sometimes also happens when `window.devicePixelRatio = 1`.

#### Before
https://github.com/user-attachments/assets/14ecd417-1a65-47b7-82d3-52a27ce9c331

#### After (with `.setSubpixelPositioning(true)`)
https://github.com/user-attachments/assets/b518cad0-5058-4e90-81ba-1f99ec676bfd

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ### ✨ Features and improvements
 
 - Expose projection matrix parameters ([#3136](https://github.com/maplibre/maplibre-gl-js/pull/3136))
+- Add option to position markers at subpixel coordinates to prevent markers jumping on `moveend`.
```
HarelM, line 6: Please add a link to this PR.

src/ui/marker.test.ts
```diff
@@ -1109,4 +1109,44 @@ describe('marker', () => {
...
+
+    test('should round the marker transform position to sub pixels when subpixel positioning is enabled', () => {
+        const map = createMap();
+        const marker = new Marker()
+            .setLngLat([0, 0])
+            .setSubpixelPositioning(true)
+            .addTo(map);
+
+        const transform = marker.getElement().style.transform;
+        expect(transform).toBe(`
```
HarelM, line 1141: This is order specific and in general a bit fragile, is there a better way to test this which is a bit more robust?

src/ui/marker.ts
```diff
@@ -84,6 +84,12 @@ type MarkerOptions = {
      * @defaultValue 0.2
      */
     opacityWhenCovered?: string;
+    /**
+      * If `true`, rounding is disabled for placement of the popup, allowing for
```
HarelM, line 88:
```suggestion
      * If `true`, rounding is disabled for placement of the marker, allowing for
```

src/ui/marker.ts
```diff
@@ -459,6 +467,22 @@ export class Marker extends Evented {
         return this;
     }
 
+    /**
+      * Set the option to allow subpixel positioning of the popup by passing a boolean
```
HarelM, line 471:
```suggestion
      * Set the option to allow subpixel positioning of the marker by passing a boolean
```

src/ui/marker.ts
```diff
@@ -459,6 +467,22 @@ export class Marker extends Evented {
         return this;
     }
 
+    /**
+      * Set the option to allow subpixel positioning of the marker by passing a boolean
+      *
+      * @param value - When boolean is true, subpixel positioning is enabled for the popup.
```
HarelM, line 473:
```suggestion
      * @param value - when set to `true`, subpixel positioning is enabled for the marker.
```

src/ui/marker.ts
```diff
@@ -600,7 +624,7 @@ export class Marker extends Evented {
         // because rounding the coordinates at every `move` event causes stuttered zooming
         // we only round them when _update is called with `moveend` or when its called with
         // no arguments (when the Marker is initialized or Marker#setLngLat is invoked).
-        if (!e || e.type === 'moveend') {
+        if (!this._subpixelPositioning && (!e || e.type === 'moveend')) {
```
HarelM, line 627: Can this condition be simplified using the "?" operator?

src/ui/marker.test.ts
JeroendeJong: I updated the test to use substring matching. Initially I tested on the internal `marker._pos` value, but felt it was better to test the css transform itself. I hope this changes resolves your concerns!

JeroendeJong:
Thanks for your comments. Sorry for the simple mistakes, I tried to get this out end of Friday and it shows haha.

src/ui/marker.ts
```diff
@@ -600,7 +624,8 @@ export class Marker extends Evented {
         // because rounding the coordinates at every `move` event causes stuttered zooming
         // we only round them when _update is called with `moveend` or when its called with
         // no arguments (when the Marker is initialized or Marker#setLngLat is invoked).
-        if (!e || e.type === 'moveend') {
+        const shouldRoundPosition = !e || e.type === 'moveend';
```
HarelM, line 627:
```suggestion
        const shouldRoundPosition = e?.type === 'moveend';
```

HarelM: Using this should allow to add it inside the below if and still keep the if readable, I think.

JeroendeJong: The rounding previous happens in two cases: 
  1) in the absence of any event (e.g. when update is called after setting a new LatLng) and 
  2) when it is triggered by specifically a `moveend` event. 

Your suggested change would mean the absence of any event now returns false, which would represent a regression because `e?.type === 'moveend'` yields `false` for `undefined`, when `true` should be given.

HarelM: Ahh, right, my bad...

HarelM, line 627: Sorry for my previous mistake, I would consider either reverting the change so that the "if" has all the conditions or change the definition of `shouldRoundPosition` to have the condition of `_subpixelPositioning`. Otherwise, this variable is a bit misleading.
Sorry again....

JeroendeJong: No worries! I opted to revert that part and include them in the same if as before.

HarelM (approved):
Thanks!

PR #4465, opened 2024-07-30
Allow pinch trackpad gesture while cooperative gestures is enabled

### LINKED ISSUE #4463: Allow pinch-to-zoom on trackpads with no extra keypress with `cooperativeGestures` enabled.
## User Story

As a user of a map embed, I want to be able to zoom an embedded map by pinching on my trackpad without having to press Cmd or Ctrl, so that the experience is more similar to using other embedded maps on the web like Google Maps and Mapbox.

## Rationale

**Why?** To make the experience more like Google Maps and Mapbox, which are common embedded map experiences.

I am not aware of any workaround for this. Perhaps it might be possible to detect the pinch ourselves on a outer component with `capture: true` and disable co-operative gestures quickly, but I've not tried that.

## Impact

Using an embedded map isn't as intuitive as other platforms, and users will have a feeling that MapLibre maps behave "worse" than other maps.

tomhicks:
## What this does

Enables pinching on a trackpad to zoom the map while `cooperativeGestures` is enabled.

Fixes #4463

## Implementation

A pinch on a trackpad is actually a `wheel` event but with `ctrlKey: true`. This change adds a new method to the `CooperativeGesturesHandler` called `shouldPreventWheelEvent` which is used internally for deciding whether to show the cooperative gestures message, and from the `ScrollZoomHandler::wheel` method, to determine whether or not to handle the gesture in the map.

## Launch Checklist

 - [X] Link to related issues.

### DIFF
--- a/src/ui/handler/cooperative_gestures.test.ts
+++ b/src/ui/handler/cooperative_gestures.test.ts
@@ -1,8 +1,8 @@
-import { browser } from '../../util/browser';
+import {browser} from '../../util/browser';
 import {Map} from '../map';
-import { DOM } from '../../util/dom';
+import {DOM} from '../../util/dom';
 import simulate from '../../../test/unit/lib/simulate_interaction';
-import { beforeMapTest, sleep } from '../../util/test/util';
+import {beforeMapTest, sleep} from '../../util/test/util';
 
 function createMap(cooperativeGestures) {
     return new Map({
--- a/src/ui/handler/cooperative_gestures.test.ts
+++ b/src/ui/handler/cooperative_gestures.test.ts
@@ -1,8 +1,8 @@
-import {browser} from '../../util/browser';
+import { browser } from '../../util/browser';
 import {Map} from '../map';
-import {DOM} from '../../util/dom';
+import { DOM } from '../../util/dom';
 import simulate from '../../../test/unit/lib/simulate_interaction';
-import {beforeMapTest, sleep} from '../../util/test/util';
+import { beforeMapTest, sleep } from '../../util/test/util';
 
 function createMap(cooperativeGestures) {
     return new Map({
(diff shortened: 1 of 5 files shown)


tomhicks:
err, code coverage is being weird:
image

How can lines 110 and 111 be not covered yet line 113 is? That's literally impossible.

HarelM:
There's a bug with latest node regarding code coverage unfortunately...

src/ui/handler/cooperative_gestures.ts
```diff
@@ -97,7 +97,20 @@ export class CooperativeGesturesHandler implements Handler {
         if (!this._map.scrollZoom.isEnabled()) {
             return;
         }
-        this._onCooperativeGesture(!e[this._bypassKey]);
+
+        const isPrevented = this.shouldPreventWheelEvent(e);
```
HarelM, line 101: I think the above `isEnabled` is not required now that it is Incorportated into this shouldPrevent method.

src/ui/handler/cooperative_gestures.test.ts
```diff
@@ -97,6 +97,39 @@ describe('CoopGesturesHandler', () => {
...
+    test('Zooms on trackpad pinch when metaKey is the bypass key', () => {
+        // NOTE: This should pass regardless of whether cooperativeGestures is enabled or not
+        const browserNow = jest.spyOn(browser, 'now');
+        let now = 1555555555555;
+        browserNow.mockReturnValue(now);
+
+        const map = createMap(true);
+
+        // pretend we're on a Mac, where the ctrlKey isn't the bypassKey
+        map.cooperativeGestures._bypassKey = 'metaKey';
```
HarelM, line 109: I would consider spying on `navigator.userAgent.indexOf('Mac')` instead of using internal fields.

HarelM:
Did you check the pinch on a mac as well to see that it behaves the same?

src/ui/handler/cooperative_gestures.test.ts
tomhicks: On that I was wondering why the codebase doesn't use private properties in TypeScript.

HarelM: Legacy code that we don't want to change due to breaking plugins that relay on it not being private...

src/ui/handler/cooperative_gestures.ts
tomhicks: One is checking `scrollZoom.isEnabled()` and one is checking `this.isEnabled()`.

If you look at my subsequent PR on top of this one, you'll see I've tried to clear up this confusion of those types of checks being done in different directions by making them all one way.

HarelM: Right, my bad!

src/ui/handler/cooperative_gestures.test.ts
tomhicks: makes sense.

tomhicks:
> Did you check the pinch on a mac as well to see that it behaves the same?

I have relied on the tests for this PR. Is there a way to easily test an embed or something while developing? Or do I need to make another codebase to test this one using npm link or something?

tomhicks:
@HarelM I've tested this on Chrome and Firefox on my MacBook Pro. I can't test Safari because the server started via `npm run start` doesn't open on Safari for some reason. Wonder if you've seen that before?

HarelM:
I never tried opening on safari to be honest...

HarelM (approved):

PR #4470, opened 2024-07-30
Emit event when cooperative gestures blocks event

### LINKED ISSUE #4464: Fire events on cooperative gesture triggers
## User Story

As a developer embedding a MapLibre map with cooperative gestures, I want to receive events when the cooperative gestures are detected and triggered, so that I can put up my own UI instead of hacking the internal one.

## Rationale

We want a custom cooperative gesture UI, which I have managed to get pretty much working by overriding the styles globally, but it's starting to get a bit fiddly. For instance, I need to separate the scrim from the message to be able to layer the scrim behind some UI but have the message above it. I can work around this by putting mutation observers on the DOM elements that are modified as the gestures are triggered, but that's a bit flaky.

If we were able to register for an event like "cooperativegestureblocked" or something we would be able to more easily add our own custom UI.

## Impact

Makes it harder to implement custom UIs for cooperative gestures.

## PR

I'm very happy to make a PR for this, if someone could just point me to the right way to get the event emitter instance into the handler (or to import a singleton).

tomhicks:
## Launch Checklist

Fixes #4464

Adds the ability to subscribe to the case when the `CooperativeGestures` handler prevents a gesture from being handled by the map. This allows users to add custom logic in this case to put up custom UI, for instance.

The events emitted are of type `cooperativegestureprevented` and have a `gestureType: "wheel_zoom" | "touch_pan"` and `originalEvent: Event` properties, so the developer has as much detail as possible. Arguably, `gestureType` can be deduced from the `originalEvent` but it feels nice to explain the gesture type in MapLibre-ish terms.

This also refactors the way in which coop gesture preventions are detected and notified. It wasn't obvious to me when I was working on this _how_ the coop gesture class was preventing the gestures from being handled/bubbled/whatever. It took a while to realise that the `touchmove` and `wheel` events in this are run in "parallel" to the _other_ gesture handlers.

The refactor makes the coop gesture handler have two distinct responsibilities:
- answering questions about whether a gesture should be prevented
- notifying the user of a prevention, when requested

This avoids duplicate logic and makes it easy to find how this module is used elsewhere as it has no intrinsic abilities: it performs logic and actions when requested by other modules.

 - [X] Link to related issues.


src/ui/handler/cooperative_gestures.ts
```diff
@@ -82,26 +83,40 @@ export class CooperativeGesturesHandler implements Handler {
...
+        this._destroyUI();
     }
 
     isEnabled() {
         return this._enabled;
     }
 
-    touchmove(e: TouchEvent) {
-        this._onCooperativeGesture(e.touches.length === 1);
+    shouldPreventTouchPan(numberOfTouches: number) {
```
HarelM, line 93: I'm not sure cooperative gestures should be aware of touch, pan, wheel as this creates weird coupling, I think...
I think every handler should have the relevant method instead, shouldn't it?

HarelM:
I still don't understand how raising these events will allow replacing the current UI.
I don't mind adding these event, but I'm lacking a proper motivation.

src/ui/handler/cooperative_gestures.ts
tomhicks: Cooperative gestures was already aware of touch and wheel though... it's just that before it used to check things to determine whether or not to show the UI in here, but other code would do its own logic to determine whether or not to prevent the gesture.

If anything this has made the coop gestures class less aware of touches, wheels, etc.

Another way to structure it would be to move all the logic out to the calling methods, and then the coop gestures class isn't so much a handler as a thing that tracks enabled/disabled state (as a central place for other handlers to check its enabledness) and gives the ability for other handlers to trigger the UI.

If you really want it as a handler tat handles gestures, there would need to be some way for handlers to stop "propagation" through other handlers in the list (not sure if you have that ability or not) but then it would still need to have the logic for trapping say `touchmove` events in the same way as say the touch_pan handler.

My advice would be to make this not a handler at all and just have `enable(), disable(), isEnabled(), notifyGesturePrevented()`.

tomhicks:
> I still don't understand how raising these events will allow replacing the current UI.

```css
.maplibre-cooperative-gestures-container {
  display: none !important;
}
```

```typescript
map.on("cooperativegestureprevented", () => showToast("Press Ctrl"));
map.on("cooperativegestureprevented", () => someObservableState.gesturePrevented = Date.now());
map.on("cooperativegestureprevented", () => analytics.sendEvent("user_tried_to_scroll_map"));
```

I can do whatever I want with this event, including putting my own UI on top of MapLibre, in my host app, sending things to analytics, etc.

src/ui/handler/cooperative_gestures.ts
HarelM: It wasn't a handler before, it was worse IMHO, that's why I changed it to be a handler.
There is a handler manager that helps with preventing events of different handlers, but I don't find its logic readable, so I'm not sure it's a great idea 
Feel free to try out stuff and see which one works best, I don't have strong feelings anyway...

tomhicks: I think this way is better than before even if it's not the best it could be. The main advantage being that there is no logic duplication.

I think it would be also OK to move the logic out and have each of the other handlers just check if it's enabled, then ask it to perform the side effects. The disadvantage here is you lose the ability to go look at the cooperative gestures module and understand what cooperative gestures means.

Thinking as I write, maybe that makes more sense, because if you were to change a gesture now, you'd need to go and change two places: the handler itself and the cooperative gestures `shouldXyz` method... think I'm going to put that back, actually!

tomhicks:
OK, documented that and put the logic back to the individual handlers so things like touch counts, etc are all in the concerned handler.

HarelM:
Hi, please note that the following was not updated with this logic:
https://github.com/tomhicks/maplibre-gl-js/blob/619773155b3126a453cfded634b35ae8d506fc8d/src/ui/handler/two_fingers_touch.ts#L289

I'm not sure what should be the solution there, but please make sure to test all scenarios to make sure nothing is broken.
I have a feeling not all gesture preventions is covered by tests considering the above "miss".

I have updated some small parts of the code, just small refactoring...

tomhicks:
> Hi, please note that the following was not updated with this logic

@HarelM I'm not sure what change you are expecting there. The logic has not and should not change.

HarelM:
There's no event in that case, and since you remove the `move` event from cooperative gesture class you broke part of this logic. 
Did you test this code with touch devices (dev tool mobile mode)?

tomhicks:
> There's no event in that case, and since you remove the `move` event from cooperative gesture class you broke part of this logic.

I'm not sure what part of the logic is broken. What I moved is the code that shows the message if the number of touches is exactly equal to 1, which is now in touch_pan, because it relates to touch panning. There was never any message shown (and now there should be no event emitted) when using 2 fingers in an upwards motion (which in non-coop mode means pitch, but in coop mode means pan) because although it's changing the nature of the gesture, it's not _blocking_ it per se.

> Did you test this code with touch devices (dev tool mobile mode)?

Yes, and it all seems to work exactly as on main, and using a real mobile device and multiple fingers, too.

The behaviour on touch devices is:

normal mode:
- 1 finger drag: pan
- 2 fingers spin: rotate
- 2 fingers vertical drag: pitch
- 2 fingers non-vertical pan: pan

coop mode:
- 1 finger drag: blocked
- 2 fingers spin: rotate
- 2 fingers vertical drag: pan
- 2 fingers non-vertical drag: pan
- 3 fingers vertical drag: pitch

That's how it is on main and that's how it is on this branch.

tomhicks:
One thing we could do would be to emit a `cooperativegesturechanged` and describe the change in that case, i.e. to convey "you might be expecting this two-finger vertical drag to result in a pitch gesture, but you actually need to use 3"

HarelM:
Thanks for the clarification, I probably got it wrong, sorry.

HarelM (approved):

PR #4478, opened 2024-07-31
Fix GeckoView tile loading android assets folder

### LINKED ISSUE #4451: Failing to process map tiles in Android Webview when run from the "asset" folder
In Actor.receive() there is a check for the origin/location.

However, if running on Android using some WebView implementations (such as Mozilla's GeckoView) and running from the Android asset folder (which is really just "file://"), the data.origin is actually set to "resource://android".

That is, the following line:
  if (data.origin !== 'file://' && location.origin !== 'file://' && data.origin !== location.origin) {

Should be extended to be:
  if (data.origin !== 'file://' && location.origin !== 'file://' **&& data.origin !== 'resource://android' && location.origin !== "resource://android"** && data.origin !== location.origin) {

As, otherwise, no map tiles are loaded and displayed.

Many thanks.

jwoodwardtfx:
Fixes #4451

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Fix remove hash string when map is removed ([#4427](https://github.com/maplibre/maplibre-gl-js/pull/4427))
 - Fix GeolocateControl may be added twice when calling addControl/removeControl/addControl rapidly ([#4454](https://github.com/maplibre/maplibre-gl-js/pull/4454))
 - Fix `style.loadURL` abort error being logged when removing style ([#4425](https://github.com/maplibre/maplibre-gl-js/pull/4425))
+- Fix vector tiles not loading when html is opened via "resource://android" (i.e., the assets folder) in GeckoView on Android ([#4451](https://github.com/maplibre/maplibre-gl-js/pull/4451))
 - _...Add new stuff here..._
 
 ## 4.5.0
--- a/src/util/actor.test.ts
+++ b/src/util/actor.test.ts
@@ -201,4 +201,18 @@ describe('Actor', () => {
 
         expect(spy).toHaveBeenCalled();
     });
+
+    test('should process a message when origin is "resource://android"', async () => {
+        const worker = workerFactory() as any as WorkerGlobalScopeInterface & ActorTarget;
+        const actor = new Actor(worker, '1');
+
+        const spy = jest.fn().mockReturnValue(Promise.resolve({}));
+        worker.worker.actor.registerMessageHandler(MessageType.getClusterExpansionZoom, spy);
+
+        actor.target.postMessage({type: MessageType.getClusterExpansionZoom, data: {} as any, origin: 'resource://android'});
+
+        await sleep(0);
+
+        expect(spy).toHaveBeenCalled();
+    });
 });
--- a/src/util/actor.ts
+++ b/src/util/actor.ts
@@ -132,7 +132,7 @@ export class Actor implements IActor {
     receive(message: {data: MessageData}) {
         const data = message.data;
         const id = data.id;
-        if (data.origin !== 'file://' && location.origin !== 'file://' && data.origin !== location.origin) {
+        if (data.origin !== 'file://' && location.origin !== 'file://' && data.origin !== 'resource://android' && location.origin !== 'resource://android' && data.origin !== location.origin) {
             return;
         }
         if (data.targetMapId && this.mapId !== data.targetMapId) {


jwoodwardtfx:
No other changes are required. "file://" isn't referenced anywhere else in the code base, and, it's only this check that is causing an issue.

HarelM:
See here:
https://github.com/maplibre/maplibre-gl-js/blob/04e39c5c55f84b5eab311244b0cc2cdf5ea2dcf6/src/util/ajax.ts#L140

I'll be surprised if this wouldn't need a patch as well...

jwoodwardtfx:
> See here:
> 
> https://github.com/maplibre/maplibre-gl-js/blob/04e39c5c55f84b5eab311244b0cc2cdf5ea2dcf6/src/util/ajax.ts#L140
> 
> I'll be surprised if this wouldn't need a patch as well...

You are right, I was just looking at that line. However, if I just patch the one line in the commit, it all springs into life. I'll do a bit more digging. Thanks.

jwoodwardtfx:
> > See here:
> > https://github.com/maplibre/maplibre-gl-js/blob/04e39c5c55f84b5eab311244b0cc2cdf5ea2dcf6/src/util/ajax.ts#L140
> > 
> > I'll be surprised if this wouldn't need a patch as well...
> 
> You are right, I was just looking at that line. However, if I just patch the one line in the commit, it all springs into life. I'll do a bit more digging. Thanks.

It looks like the code is trying to avoid using the Fetch API when using "file://" and instead drop back to the legacy XMLHttpRequest.

It's not clear to me why, as I can't find anything in Mozilla Fetch API documentation (https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) or Chrome documentation (https://web.dev/articles/introduction-to-fetch) that states that the Fetch API can't work with a "file://" request (or "resource://android" in my case).

I've done some tests an using GeckoView on Android and both the Fetch API and XMLHttpRequest successfully load the fonts and glyphs.

I've had a quick look at Chromium on Windows too using "file://" and this is the same, both Fetch API and XMLHttpRequest seem to work fine.

So, I think this check is redundant, though, I'm not keen to back it out, even if not needed, as I assume it was added for an older browser than may not support "file://" access via the Fetch API? Removing it may cause a breakage for someone else, but, with a modern browser, it doesn't appear to be needed anymore.

HarelM:
I can find old articles about fetch not supporting file protocol: 
https://github.com/JakeChampion/fetch/pull/92#issuecomment-140665932
https://stackoverflow.com/questions/55133723/does-will-the-fetch-api-support-other-protocols-than-http

It might be that this is no longer relevant, IDK...

jwoodwardtfx:
If you agree, maybe the best option for now is to apply this fix for GeckoView, and the use of the Fetch API vs XMLHttpRequest can be looked at at a later point? A change to ajax.ts isn't needed to resolve the specific issue for GeckoView :)

HarelM:
@jwoodwardtfx it seems that this PR introduced some unwanted console logging in the tests.
Any chance you can take a look and maybe help solve the issues?
For example running:
`npm run test-unit src/ui/map_tests/map_disable_handlers.test.ts`
Gives the following console errors:
```
stderr | Timeout._onTimeout (/Users/harelmazor/dev/maplibre-gl-js/test/unit/lib/web_worker_mock.ts:41:25)
ReferenceError: location is not defined
    at Actor.receive (/Users/harelmazor/dev/maplibre-gl-js/src/util/actor.ts:145:42)
    at /Users/harelmazor/dev/maplibre-gl-js/src/util/actor.ts:82:81
    at Timeout._onTimeout (/Users/harelmazor/dev/maplibre-gl-js/test/unit/lib/web_worker_mock.ts:38:21)
    at listOnTimeout (node:internal/timers:573:17)
    at processTimers (node:internal/timers:514:7)
```

jwoodwardtfx:
Are you sure it's this commit that is causing the issue?

The location object was already being referenced prior to it (i.e., this commit didn't reference anything that wasn't already being referenced - namely "location.origin" and "data.origin").

I think the root question is, what in the test infrastructure would cause "location" to end up as being undefined? Or, has there been a change to the test infrastructure such that "location" is no longer available?

It looks like the use of location was introduced back in 2023 to protect against XSS attacks (https://github.com/maplibre/maplibre-gl-js/pull/3329/files). Every change since then has just been minor tweaks on this.

HarelM:
You are right, it might not have been this PR, I only took a look recently after a long time without looking at this...
It would be great if you could help solve this issue even if it wasn't introduced as part of this PR.

PR #4488, opened 2024-08-01
map.getStyle() modification issue

### LINKED ISSUE #4481: Modifying the results of map.getStyle() modifies the style
**maplibre-gl-js version**: 3.0.0 and higher.

- More specifically, this issue was introduced in https://github.com/maplibre/maplibre-gl-js/pull/2306

**browser**: all

### Steps to Trigger Behavior

 1. Call `map.getStyle()` and assign it to a variable.
 2. Modify that variable in some way.
     - Expected: modifying that variable should not change style. Only calling `map.setStyle()` should do that.
     - Actual: modifying that variable **does** change style.

### Link to Demonstration

1. Open https://jsbin.com/zeyusuf/4/edit?html,output , which is using `maplibre-gl-js@2.4.0`. Note that even though I'm modifying the variable like this: 
    ```js
      const newStyle = map.getStyle();
      
      newStyle.layers[0].paint = {
        "fill-color": '#AAAAAA',
      };
    ```
   ... calling `map.getStyle()` after I make that change still returns the original value. This is the **correct** behavior because I have not called `map.setStyle()`.

2. Open https://jsbin.com/zeyusuf/6/edit?html,output , which is using `maplibre-gl-js@3.0.0` but otherwise the exact some code as above.
    - Expected: calling `map.getStyle()` after the variable is changed still returns the original value.
    - Actual: calling `map.getStyle()` after the variable is changed **does not return the original value** ❌.

## Note

- I think this is the core issue causing the problems in #4472.
- I think this issue was introduced in https://github.com/maplibre/maplibre-gl-js/pull/2306

gavinr-maps:
This change returns a clone of the object so that when calling `.getStyle()` the user cannot inadvertently modify the internally stored style. 

- Fixes #4481
- Unit tests: see fe7957bc16b09061b2428cc0e6d18e68d2975880
- Document any changes to public APIs: no changes

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


src/style/style.ts
```diff
@@ -483,7 +483,7 @@ export class Style extends Evented {
      */
     private _serializeByIds(ids?: Array<string>): Array<LayerSpecification> {
 
-        const serializedLayersDictionary = this._serializedAllLayers();
+        const serializedLayersDictionary = clone(this._serializedAllLayers());
```
HarelM, line 486: I would consider doing the clone only later, inside the "for" to only clone the needed layers.

HarelM: Also, since this is also used to send the data to the workers, which should not be cloned I would consider adding a flag to this method to clone only if this flag is "on".
One last comment, can we use structuredClone instead of lodash clone?

src/ui/map_tests/map_style.test.ts
```diff
@@ -359,6 +359,32 @@ describe('#getStyle', () => {
         });
     });
 
+    test('returns the previous style even if modified', done => {
```
HarelM, line 362: use `await once('load')` instead of using `done` as we are trying to reduce the number of "done" in the tests.

HarelM:

src/style/style.ts
tomhicks: Agree with the flag suggestion. This is potentially very performance sensitive - we use getStyle on mousemove and have recently _removed_ a lodash cloning of our MSS style (in favour of a shallow clone because of how we were using it) and it saved us ~7ms per mousemove on some maps.

Another alternative is deep freezing, which would prevent mutation and put the cloning requirement onto users.

src/ui/map_tests/map_style.test.ts
gavinr-maps: thanks! changed it to use async/await instead.

src/style/style.ts
gavinr-maps: It looks like `structuredClone` was added in Node v17.0 - where this library is currently supporting https://github.com/maplibre/maplibre-gl-js/blob/5332a1c8524dd08ddb0a40590e32312d467b85bd/package.json#L190
... so I'm not allowed to use `structuredClone`, am I? I do not see it used elsewhere in the codebase.

HarelM: I see, you are right, it is not used elsewhere, we can decide to replace all the `clone` with this later on, let's keep it outside the scope of this PR.

gavinr-maps:
@HarelM thank you for the feedback and your time maintaining this project. I have addressed your feedback and is ready for your re-review when you have some time. Thanks.

HarelM (approved):
Looks good THANKS!

anneb:
Is this a breaking change?
A way to get the list of all layers is:
```
map.getStyle().layers
```
When handling layers, it is legal to modify the layer metadata. Modified layer metadata should be kept with the layer, not with a clone.

My workaround:
```
const layers = myMap.getStyle().layers;
const unclonedLayers = layers.map(layer => myMap.getLayer(layer.id));
```

From this issue, I can not infer why it was required to modify getStyle() to return a clone, not the original.
Layer metadata is a good way to store application information with a layer, so shouldn't it be updatable?

neodescis:
> Is this a breaking change? A way to get the list of all layers is:
> 
> ```
> map.getStyle().layers
> ```
> 
> When handling layers, it is legal to modify the layer metadata. Modified layer metadata should be kept with the layer, not with a clone.
> 
> My workaround:
> 
> ```
> const layers = myMap.getStyle().layers;
> const unclonedLayers = layers.map(layer => myMap.getLayer(layer.id));
> ```
> 
> From this issue, I can not infer why it was required to modify getStyle() to return a clone, not the original. Layer metadata is a good way to store application information with a layer, so shouldn't it be updatable?

Actually, getStyle().layers is a list of layer specifications, not the layer implementations themselves. The call to getLayer() in your example is retrieving the implementation, not the specification. The two are not interchangeable. 

Generally, updating an existing layer is done through getLayer(). Updating the original style or layer specification instances directly after they have been incorporated into the current map style isn't really going to work anyway, as nothing is watching for those changes.

PR #4525, opened 2024-08-08
Anisotropic filtering only when pitch is greater than 20 degrees

e-k-m:
Enable anisotropic filtering only when the pitch is greater than 20 degrees to preserve image sharpness on flat or slightly tilted maps.

## Launch Checklist

Enable anisotropic filtering on tiles beyond 20 degrees pitch to prevent it from compromising image crispness on flat or low-tilted maps.
        
 - [x] Link to related issues.
 
Non known.
 
Before the fix paint: { "raster-resampling": "nearest"} has no effect on Windows and looks like:

image

with the fix it is correctly displayed:

image

         No public API changes.
         
 - [ ] Post benchmark scores.
 

HarelM:
Is there an open issue related to this PR by any chance?

e-k-m:
> Is there an open issue related to this PR by any chance?

No I have not opened a related issue to this PR.

HarelM:
Are we OK with the license of the raster images added here?

HarelM:
What's the license of the added raster images?
Can you explain why you moved the code from one place to the other?
CC: @louwers for render parity
Generally, approved.

HarelM:
Don't forget to add a changelog entry.

e-k-m:
> What's the license of the added raster images?

The raster images are from ©swisstopo for details on the usage terms see https://www.swisstopo.admin.ch/en/terms-of-use-free-geodata-and-geoservices. In summary: "The free geodata and geoservices of swisstopo may be used, distributed and made accessible. Furthermore, they may be enriched and processed and also used commercially."

According to the usage terms a reference to the source is mandatory. Where should this reference be written down? 

> Can you explain why you moved the code from one place to the other?

I moved it from the function `loadTile` to `drawRaster`, since it seems more suiting / expecting that anisotropic filtering happens when drawing and not while loading.

HarelM:
Since these are using only in tests I wouldn't want to add them to license.txt file.
So I think adding a small note at the bottom of one of the readme files in the tests folder makes sense to me. 
But it might also be an overkill, and stating it in this PR might be enough, IDK...

HarelM (approved):

e-k-m:
Concerning the copy right statement, I am also not sure what is the best solution. But I think we are fine with only having it in PR.

PR #795, opened 2022-01-19
Attribution default open for osm

acalcutt:
This is an attempt to Show OSM attribution by default, like mentioned in ( https://github.com/maplibre/maplibre-gl-js/issues/205 )

This change makes the compact mode default to being open to comply with the [OpenSteetMap Attribution Guidelines](https://wiki.osmfoundation.org/wiki/Licence/Attribution_Guidelines#Interactive_maps) that expect attribution is [visible for at least some time](https://lists.openstreetmap.org/pipermail/osmf-talk/2021-June/007914.html) without user interaction to fall within the safe harbour rules. 

- code has been added so the button is default open when going into compact mode.
- code has been added so attribution will close automatically when there is a drag action
- code has been added to not show the compact button if there are no attributes to prevent a open attribute button with no attributes. (I actually noticed the compact button existing when there were no attributes bug on the maplibre site on <= 640 screen width, but it is more noticeable when it is open by default. )

 - [X] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [X] Briefly describe the changes in this PR.
 - [ ] Include before/after visuals or gifs if this PR includes visual changes.
 - [X] Write tests for all new functionality.
 - [ ] Document any changes to public APIs.
 - [ ] Post benchmark scores.
 - [ ] Manually test the debug page.
 - [ ] Suggest a changelog category: bug/feature/docs/etc. or "skip changelog".
 - [ ] Add an entry inside this element for inclusion in the `maplibre-gl-js` changelog: `<changelog></changelog>`.

PR #4526, opened 2024-08-08
Add option to initially hide attributions on small screens

birkskyum:
It's been requested ( https://github.com/plotly/plotly.js/issues/7076#issuecomment-2276170127 )  to have a setting that'll allow initially collapsing the attribution text when it's in toggle-mode (compact or screen smaller than 640px).

Currently, attribution will collapse on drag, but for plotting libs like plotly.js show below, there's typically many small figures (attribution is repeated and takes up relatively more space), with the same viewport (dragging puts them out of sync).

Screenshot 2024-08-08 at 18 35 20

### DIFF
--- a/src/ui/control/attribution_control.ts
+++ b/src/ui/control/attribution_control.ts
@@ -14,6 +14,12 @@ export type AttributionControlOptions = {
      * **Attribution should not be collapsed if it can comfortably fit on the map. `compact` should only be used to modify default attribution when map size makes it impossible to fit default attribution and when the automatic compact resizing for default settings are not sufficient.**
      */
     compact?: boolean;
+
+    /**
+     * Collapse attributions on small sceens or in compact mode by default. Default: false
+     */
+    collapsedByDefault?: boolean;
+
     /**
      * Attributions to show in addition to any other attributions.
      */
@@ -40,6 +46,7 @@ export class AttributionControl implements IControl {
     options: AttributionControlOptions;
     _map: Map;
     _compact: boolean | undefined;
+    _collapsedByDefault: boolean | undefined;
     _container: HTMLElement;
     _innerContainer: HTMLElement;
     _compactButton: HTMLElement;
@@ -63,6 +70,7 @@ export class AttributionControl implements IControl {
     onAdd(map: Map) {
         this._map = map;
         this._compact = this.options.compact;
+        this._collapsedByDefault = this.options.collapsedByDefault;
         this._container = DOM.create('details', 'maplibregl-ctrl maplibregl-ctrl-attrib');
         this._compactButton = DOM.create('summary', 'maplibregl-ctrl-attrib-button', this._container);
         this._compactButton.addEventListener('click', this._toggleAttribution);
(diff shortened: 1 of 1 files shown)


src/ui/control/attribution_control.ts
```diff
@@ -14,6 +14,12 @@ export type AttributionControlOptions = {
      * **Attribution should not be collapsed if it can comfortably fit on the map. `compact` should only be used to modify default attribution when map size makes it impossible to fit default attribution and when the automatic compact resizing for default settings are not sufficient.**
      */
     compact?: boolean;
+
+    /**
+     * Collapse attributions on small sceens or in compact mode by default. Default: false
```
archmoj, line 19:
```suggestion
     * Collapse attributions on small screens or in compact mode by default. Default: false
```

HarelM:
If you touch the map the attribution will collapse, is this really a must? 
I'm guessing (not sure though) this can also be done simply with CSS so no API changes areb required...?
If there's no other alternative, please add a test.

acalcutt:
The point of starting the attributions expanded in https://github.com/maplibre/maplibre-gl-js/pull/795 was to comply with [OpenSteetMap Attribution Guidelines](https://wiki.osmfoundation.org/wiki/Licence/Attribution_Guidelines#Interactive_maps). Based on the screenshots above it looks like OSM is being used there, so adding this option goes against the spirit of the license (in my opinion)

birkskyum:
@acalcutt , this is technically Interactive Maps, which means the only hard requirement of the linked license I can see is:

> If the attribution has been collapsed, the user must still be able to find the licence information if they look for it, for example from an ‘(i)’ button in the corner of the map or an ‘About’ option in a menu.

I am not a lawyer, and this ticket isn't specifically about osm which was just an example, but I don't think it technically breaches the license of OSM to have it hidden by default give the above license, if a user of maplibre want to do that. I don't have much opinion in the matter myself, and I can sympathize with the "spirit of the license" you mention which is also why this is an opt-in. At the end of the day I don't really want to tell anyone how to comply with various licenses, but rather just provide the flexibility necessary to do so and have good defaults. We can in the docs of this flag write that users should be wary of complying with the attribution license of the style / data-provider in question, in case there are styles out there with a legal requirement to be present initially or even at all times. 

@HarelM , I did try a bit with the css, but the problem I hit was that the -show class is added to the dom element initially, and I can't remove it from the css. This would do so that the -show isn't added initially if this flag is set.

@archmoj , let me know if this is still a blocking issue for the next RC of plotly.js.

acalcutt:
I think you missed the bit right above it
```
You may use a mechanism to fade/collapse the attribution under certain conditions:

immediately with a dismiss interaction, for example clicking an ‘x’ in the corner of a dialog
automatically on map interaction such as panning, clicking, or zooming
automatically after five seconds. This also applies to splash screens or pop-ups.
```

Maplibre I don't think is obligated to comply with the osm license, but it should be able to. before it wasn't possible to to comply with any of those recommended options, but right now we comply with the second one. With your option at least it would allow the user to change the default if they did not need to comply with those rules (like if they used another source with a different license).  

We have had other conversation about this recently in the maplibre-native side in https://github.com/maplibre/maplibre-native/issues/2600

birkskyum:
@acalcutt , you're right, I read that hard requirement wrong, so it's only if it's a non osm data effectively that this setting comply.

acalcutt:
ya, we can also comply with the first option right now, as "toggle can be clicked to dismiss the attribution", but it must be shown first (which is wat my change did)

acalcutt:
I think right now if you really needed this, it might be able to set your own blank attribution on your sources. If there are no attributions I had made it so the toggle button doesn't even show.... idk i would have to test.

birkskyum:
apple maps only has a tiny "Legal" link in the lower right... @acalcutt is that within the rules, these way you interpret then?

 imageimage

acalcutt:
Personally I don't think the apple implantation does meet the guidelines, but they have better lawyers than me :-) I will also note the change to require showing it isn't very old (i was changed in the last few years ago), so it likely was enough before that change.

I'd say what apple has done is similar to maplibre-native implementation talked about in https://github.com/maplibre/maplibre-native/issues/2600

acalcutt:
I don't think adding and option would be a bad thing though. It would be up to the person deploying the map, but at least we have safe defaults.

HarelM:
I'm fine either way, I don't have strong feelings.
Please add the relevant unit tests if you would like to push this forward.

PR #4531, opened 2024-08-09
WIP - Allow importing .ts files instead of sloppy imports

birkskyum:
This TS 5.0 [feature](https://www.typescriptlang.org/tsconfig/#allowImportingTsExtensions) allow us to incrementally move beyond sloppy imports and use .ts extensions like

`import { namedImport } from "./file.ts"`

I've update the ./src/util/ folder in a separate commit here to validate it's effect

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "forceConsistentCasingInFileNames": true,
     "allowJs": true,
+    "allowImportingTsExtensions": true,
     "allowSyntheticDefaultImports": true,
     "checkJs": true,
     "isolatedModules": false,


HarelM:
What's the added value of this change?

birkskyum:
The idea is that it's faster for the LSP and the bundler to reach the right file in the first hit, than to search through .js/.jsx/.ts/.tsx for each import.

HarelM:
How much build time does this actually reduces in practice for this project?
Are there guidelines to start using this type of import somewhere?
My main concern is for tools running on the code that will have a hard time with this change, not sure if this is a valid concern, but it is still concerning me, especially since I haven't this this used widely...
I don't have hard objections either way, but I would still like to wait after the globe PR is merged to do this to avoid merge conflicts.

birkskyum:
It's fine with me to reassess this later.

Issue #1595, opened 2022-08-31
ECMAScript


birkskyum:
ESM is the new standard with import/export etc. We don't use it now because there are some old mapbox dependencies that only export .cjs, but nothing more than we can handle. Also, firefox has for 5 years struggled to support this in web workers, and now it's about time to move on:

### Firefox
Lack ESM support for web workers: https://caniuse.com/mdn-api_worker_worker_ecmascript_modules
5+ years of ongoing effort: https://bugzilla.mozilla.org/show_bug.cgi?id=1247687

### Benchmarks
I once attempted to move the benchmark to ESM #964 , but found the change wasn't backward compatible with the already uploaded benchmarks of prev versions, so it was rolled back #969 . 

This can be a tracking issue where we first map the dependencies we have that aren't compatible, and then keep this updated when they are updated

HarelM:
Do you mean that we currently export maplibre as commonjs and we should export it as esm?
I think the problem of switching out of commonjs besides the FF support was the ability to have the worker URL pushed while building this.
There are some complicated build steps to facilitate for this which I think are very hard to split.
I've looked at this when I tried to understand if we could leverage tree shaking here: #977
I generally think this is a good idea, even better if we can introduce it as part of 3.0 version as a breaking change.
But I'm not sure how to solve all the build complexity with the workers url and also the shared code between the main thread and the worker thread...

birkskyum:
Sure, I think we might be able to take small steps towards it. A major blocker previously was i.e. also all this node assert that would break the dev build if exported as esm. The first step is likely to have the build as esm.

backspaces:
The published dist/ dir already has multiple versions: https://unpkg.com/browse/maplibre-gl@2.4.0/dist/
And indeed, the individual src files appear use import/export, although "bare".

So shouldn't it be easy to just add a new one: maplibre-gl-esm.js?
And add a "module": "dist/maplibre-gl-esm.js" to package.json?
That seems to be the common approach for folks wanting to support both forms.

geraldo:
> import maplibreGl from 'https://cdn.skypack.dev/maplibre-gl'

@backspaces That doesn't work for me:

`var map = new maplibreGl.Map()`

gives me an `[ERR_UNSUPPORTED_ESM_URL_SCHEME]` error.

backspaces:
Try it using this:

```
<!DOCTYPE html>
<html>
<head>
    <title>Imports</title>
</head>
<body>
[...]

PR #4533, opened 2024-08-10
Use moduleResolution: Bundler

### LINKED ISSUE #4275: Use a modern moduleResolution strategy

birkskyum:
Closes #4275 
- #4275 

Related to
- #1595 

Released with TS5 - [Release notes](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#moduleresolution-bundler)


HarelM:
Does this change create any changes in the output files in the dist folder?
Can you check that an external library that uses the code after this PR does not need to change anything for this to work?

birkskyum:
I've tried swapping the dist folder in an app i have, and it doesn't appear to make a difference. It also seems the production bundle (and the other bundles) is the same size down to the byte.

Screenshot 2024-08-11 at 13 47 22

birkskyum:
Let me know if other information is needed.

HarelM (approved):

Issue #4534, opened 2024-08-10
spector.js no longer able to capture a frame with maplibre-gl-js


xabbu42:
Since commit 859b1a7 capturing a single frame with spector.js no longer works. I assume it is the same problem as https://github.com/BabylonJS/Spector.js/issues/115.  This issue is about different ways to handle that as spector.js is quite useful.

Can we go back to doing the rendering directly in the requestAnimationFrame callback or are there technical issues against that approach? I'll add a pull request with my experiments in this direction.

If not, can we find a way to still capture a single frame with spector.js and document that?

HarelM:
The main purpose of the linked commit was to remove callback hell from the code.
It might have been an overkill to do that for requestAnimationFrame as it is not used widely, but generally speaking, I would like to keep it as is in terms of APIs.
I would like to receive a better definition of the problem as the linked issue is not specific enough to maplibre, and I would like to get a better reproduction of the issue in terms relevant to this repo and library if possible, i.e. link to jsbin and better description of the issue.

xabbu42:
This is a basic map using v4.5.2 and enabling the spector ui: https://output.jsbin.com/tikepev/1
if you click on the red circle to start a frame capture the map remains responsive and after a while you get the error "No frames with gl commands detected. Try moving the camera.".

The same example with v3.6.2: https://output.jsbin.com/tikepev/2
here a click on the red circle blocks the map and after a while you get the capture of a single frame.

xabbu42:
This was fixed by the merge of #4535.

PR #4535, opened 2024-08-10
Add back synced frame method and use it for triggerRepaint.

xabbu42:
## Launch Checklist

An experimental fix for #4534. With this I can again a capture a single frame with spector.js, as all the rendering is directly done inside the requestAnimationFrame callback.

Unfortunatly tests dont run through but fail after some random point. Probably a race I introduced?

 - [x] Link to related issues.


HarelM:
requestAnimationFrame is not a method that runs right away as far as I know, it might have higher priority than a promise, but I'm not sure I saw it anywhere, so I'm not sure I know how this fix will actually solve anything, but let me know if it does.
I think the code introduced here is fine in terms of how it looks.
A test is needed to cover the relevant issue (spector.js if I understood correctly).

In case this is not ready for review, please change it to a draft and ping me when you change it back to regular PR.

THANKS!

xabbu42:
this experiment is not about the time between the call to requestAnimationFrame and when the passed callback is called, but about the time between the call to the passed callback and the actual map._render call. Currently the passed callback only schedules the _render and then returns whereas before the callback directly called _render.

Come to think of it, it should be possible to keep with the promise based interface for frameAsync for the time between the call to frameAsync and when the callback of requestAnimationFrame actually gets triggered while at the same time calling _render directly in the callback. I'll try to implement that approach.

HarelM:
Can we pass down the callback to the request animation frame and then call the resolve at the end? Would that work? Or is this the scratched out part in your comment?

HarelM:
@xabbu42 what's the status of this PR?
Are you planning on marking it as ready for review or close it?

xabbu42:
I still plan to finish this. There are uncaught exceptions during the unit tests which I still have to look into.

Sidenote: I was lead quite astray because the exceptions from the unit tests also aborted the render tests when running `npm run test` so all I see in the console are a lot of render test failures.

HarelM:
`npm run test` is running all the tests in the repo, I always run just a single part of the test using: `npm run test-unit`, `npm run test-build`, `npm run test-render` etc.
Also the CI runs them separately so you can see which is failing and why.

HarelM:
Looks like something is failing when the test harness is completing or something.
Try and use `test.only` to find the offeding test.
The test are failing because style.projection is not defined, so it should be mocked as part of the test.
I hope this helps, if not feel free to reach out in slack and I'll see how else I can help out.

xabbu42:
I found a fix for the failing tests. Problem was that the style was not yet loaded and with these changes the tests triggered a render immediately instead of just scheduling it.

Last remaining question for me is if it would be worth it to try to eliminate the other uses of frameAsync. I don't have a strong opinion. My only 2c are that it may be bad to offer two ways of doing the same thing without clear guidance on when to use which. What do you think @HarelM?

HarelM:
I generally tried to avoid the callback hell and this was one place that contributed to the complexity of reading the code. 
I understand that this causes an issue using spectator.js, is it possible to fix the issue there as I don't think there's an actual problem with the code here, is there? 
Bottom line, I like the fact that the code is using promises and I would be glad to be able to continue using promises. 
Do you think it's achievable?

xabbu42:
I don't see a way how this could be fixed generically on the spector.js side. They need some way to know when a single frame starts and ends.

But also, from reading https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame, it seems wrong to me to just schedule the actual repaint in the callback and not execute it directly. Isn't the main advantage of using requestAnimationFrame to tightly couple the map paints with the screen refresh rate and don't we loose that coupling again by just scheduling a repaint and then immediately returning?

As far as I know there is no way to force synchronized execution of a promise. So there is no way keep the aesthetics of a promise based interface but executing the code synchronized under the hood.

So I currently don't think that your request is achievable. And there might be actual problems with the code as is although very hard to notice ones.

src/util/browser.ts
```diff
@@ -16,13 +16,17 @@ export const browser = {
      */
     now,
 
+    frame(abortController: AbortController, fn: (paintStartTimestamp: number) => void, reject): void {
```
HarelM, line 19: Please add a type to reject variable.

HarelM:
I would consider removing other frameAsync calls as part of different PRs so that we can discuss case by case. 
Other than that, this looks good, thanks!

src/util/browser.ts
xabbu42: Is the type I added correct? Or should the argument to the reject function be optional?

HarelM: Looks good, thanks. We can change it in the future if there will be a problem.

HarelM:
Can you please add a changelog item and I'll merge this.
Thanks.

HarelM (approved):

PR #4540, opened 2024-08-11
Move changelog entry for Arabic Extended-B fix

1ec5:
Fixed a bad conflict resolution on my part in #4536 that put the changelog entry under the wrong heading.


HarelM (approved):
Thanks!

PR #4542, opened 2024-08-12
Remove postcss-inline-svg overrides because of upsteam fix

birkskyum:
In Jan 2023, I noticed we had a security bug, in a dependency `postcss-inline-svg`, because of a dependency it had.

There was already a ticket in the repo to update said deps:
- https://github.com/TrySound/postcss-inline-svg/pull/103

But until it got merged/released, we could fix it locally by using overrides.

This has now been fixed upstream since v6 of postcss-inline-svg, which we already use. This version use the exact same versions of nested deps as we set in the overrides (can be verified in the [package.json](https://github.com/TrySound/postcss-inline-svg/blob/master/package.json), and by the fact our lockfile doesn't change after this), so it's now ready to be cleaned up.

### DIFF
--- a/package.json
+++ b/package.json
@@ -138,14 +138,6 @@
     "typedoc-plugin-missing-exports": "^3.0.0",
     "typescript": "^5.5.4"
   },
-  "overrides": {
-    "postcss-inline-svg": {
-      "css-select": "^5.1.0",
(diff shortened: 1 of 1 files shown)


HarelM (approved):

HarelM:
Thank!

PR #4546, opened 2024-08-12
Globe: bugfixes: raster layer & projection change

kubapelc:
This PR fixes several bugs:

- Raster layers on globe were sometimes warped or deformed, and by different amounts depending on zoom level, making the warping especially visible when transitioning between eg. zoom level 8 and 9. This behaviour can still be observed in the [public demo](https://kvaleya.gitlab.io/maplibre/globe/globedemo.html) at this time. This was caused by the textured quad mesh used for raster layers not being subdivided enough. Added separate subdivision settings for raster layers (fixed the already present bug ignored setting) and stencil masks.
- Fixed that the globe would not load additional tiles after the automatic globe-mercator transition inside globe projection.
- Fixed some globe examples that were using the old API to set globe projection.

obrazek


test/examples/globe-fill-extrusion.html
```diff
@@ -15,87 +15,93 @@
...
 <div id="map"></div>
 <script>
-    const map = new maplibregl.Map({
-        container: 'map',
-        projection: 'globe',
-        style: 'https://api.maptiler.com/maps/basic/style.json?key=get_your_own_OpIi9ZULNHzrESv6T2vL',
-        zoom: 0,
-        center: [30.0, 40.0],
-    });
+    fetch('https://api.maptiler.com/maps/basic/style.json?key=get_your_own_OpIi9ZULNHzrESv6T2vL').then(response => response.json()).then(fetchedStyle => {
```
HarelM, line 18: I think it would be simpler to use the public API: `map.setProjection('globe')` instead of this complicated fetch.

test/examples/globe-vector-tiles.html
```diff
@@ -15,13 +15,22 @@
...
 <div id="map"></div>
 <script>
-    const map = new maplibregl.Map({
-        container: 'map',
-        projection: 'globe',
-        zoom: 0,
-        center: [137.9150899566626, 36.25956997955441],
-        style:
-            'https://api.maptiler.com/maps/streets/style.json?key=get_your_own_OpIi9ZULNHzrESv6T2vL',
+    fetch('https://api.maptiler.com/maps/streets/style.json?key=get_your_own_OpIi9ZULNHzrESv6T2vL').then(response => response.json()).then(fetchedStyle => {
```
HarelM, line 18: Same as above

test/integration/render/tests/projection/globe/raster-warped/style.json
```diff
@@ -0,0 +1,47 @@
...
+  ],
+  "zoom": 8,
+  "projection": {
+    "type": "globe"
+  },
+  "sources": {
+    "source": {
+      "type": "raster",
+      "tiles": [
+        "local://tiles/checkerboard-raster/{z}-{x}-{y}.png"
```
HarelM, line 23: I don't think you need to use `z-x-y` here but instead use a non-parameterized tiles URL, this way you don't need to duplicate the tile.
I also created a pink "X" tile you can use.

test/examples/globe-vector-tiles.html
kubapelc: Doing that results in `Error: Style is not done loading.`. If called after map load, then first you have to wait for mercator map to load, then `setProjection` gets called, causing the map to reload, so you have to wait a bit more until globe displays. The original API that specified projection in map constructor was handy in that this behaviour could be avoided by loading the map with the correct projection from the start.

HarelM: Maybe you can use `style.load` event for faster UX. but generally, I prefer the simplicity in the examples even if it isn't the best UX.

src/geo/projection/projection.ts
```diff
@@ -108,5 +116,5 @@ export interface Projection {
      * @param hasBorder - When true, the mesh will also include a small border beyond the 0..EXTENT range.
      * @param allowPoles - When true, the mesh will also include geometry to cover the north (south) pole, if the given tileID borders the mercator range's top (bottom) edge.
      */
-    getMeshFromTileID(context: Context, tileID: CanonicalTileID, hasBorder: boolean, allowPoles: boolean): Mesh;
+    getMeshFromTileID(context: Context, tileID: CanonicalTileID, hasBorder: boolean, allowPoles: boolean, usage: TileMeshUsage): Mesh;
```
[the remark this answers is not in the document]
kubapelc: Fixed.

HarelM (approved):

HarelM:
The globe branch has a failing test due to recent changes.
I've checked the images and they are almost the same, so they can be updated.
Any chance you can handle this? Sorry to throw it on you...

test/examples/globe-vector-tiles.html
kubapelc: This works perfectly, thank you!

Should I listen to the `style.load` event on the map object or on the style object? Style seems to work, I'm not sure if map isn't the indented one though.

test/integration/render/tests/projection/globe/raster-warped/style.json
kubapelc: Fixed. I've kept the checkerboard for now, since it might reveal more slight deviations of the images.

kubapelc:
No problem, I added the new image.

The raster-warped test (a different test that what failed now) is also very inconsistent, it seems to render differently by half a pixel every time, which is bad, since it is also trying to measure a distortion that is also only a few pixels different. Maybe carefully changing the difference threshold is a good idea here? It also renders with what seems like different zoom level on Linux, and I have no idea why...

test/examples/globe-vector-tiles.html
HarelM: I think you should listen to the map object as the style object is internal.
`style.load` is also somewhat undocumented, but it is used widely I believe...

test/examples/globe-fill-extrusion.html
```diff
@@ -17,12 +17,19 @@
...
-        zoom: 0,
         center: [30.0, 40.0],
     });
 
+    map.style.on('style.load', () => {
+        map.setProjection({
+            type: 'globe', // Set projection to globe
+        });
+        map.setSky({
+            'atmosphere-blend': 0.0, // Disable sky
```
HarelM, line 29: We should probably fix that, it should be `setSky(undefined)` or something similar, but I think it's not working.
Also in general, if we haven't set the atmosphere in the style there shouldn't be any, shouldn't it?

HarelM: This can be solved outside the scope of this PR.

HarelM:
Let me know how you want to handle the above setSky and I'll merge or wait accordingly.
Thanks for all the hard work!

test/examples/globe-fill-extrusion.html
kubapelc: Yes, `setSky(undefined)` seems to be broken, and the default is that sky/atmosphere is active.

kubapelc:
You can go ahead with the merge, I looked at the code that gets called in `map.setSky` and I don't know what to change to make passing in undefined work.

test/examples/globe-fill-extrusion.html
HarelM: This is also a problem in main, not just this branch.
I'm not sure what's causing this, as it looks like setting sky to undefined should work.
Maybe there's a need to "clear" the sky mesh in case the sky is undefined here:
https://github.com/maplibre/maplibre-gl-js/blob/5cd09f0260681f019542656b760af6bd7c25e671/src/render/painter.ts#L412
Not sure though...

HarelM:
I've fixed the set sky to undefined in #4587.
Let's make sure we remember to remove the atmosphere blend after we merge the fix.

Issue #2307, opened 2023-03-27
Labels omit any character outside of the Basic Multilingual Plane


1ec5:
Symbol layers can only render Unicode characters in the Basic Multilingual Plane (BMP, up to U+FFFF), but they do not render any character in one of the other non-BMP [planes](https://en.wikipedia.org/wiki/Plane_(Unicode)) (“astral planes”). GL JS simply skips the character without even leaving a space.

## Example

According to OpenStreetMap, [this tourist attraction in Taiwan](https://www.openstreetmap.org/node/4322299689) is named “卡司𥰆拉樂園” in Traditional Chinese. But [this stripped-down GL JS rendering of that label](https://jsbin.com/foroneqayo/1/edit?js,output) appears as “卡司拉樂園”, skipping over the “𥰆” (U+25C06). This is the case whether ideographic glyphs are being rendered on the server side or locally on the client side.

## Impact

As of 2017, this issue affects a variety of Chinese (particularly Cantonese) and Japanese names in OpenStreetMap: https://github.com/mapbox/mapbox-gl-js/issues/4001#issuecomment-312393003. In Chinese, skipping a character could easily confuse or offend the user. OSM also contains names in some minority languages that require astral-plane characters, such as Hmong and Osage. Some GL JS–based maps, such as OpenStreetMap Americana, allow the user to switch to a minority language globally, even where it lacks official status.

Apart from OSM, developers may want to label geographic features using historic languages and writing systems. For example, OpenHistoricalMap, which currently uses Mapbox GL JS, contains historic place names in [Egyptian hieroglyphics](https://taginfo.openhistoricalmap.org/keys/name%3Aegy) and [Sumerian cuneiform](https://taginfo.openhistoricalmap.org/keys/name%3Asux). These places are invisible, making the map much less useful in some time periods. Less seriously but perhaps more practically, a developer could choose to spice up their map with hard-coded emoji in symbol layers. Many emoji are located outside the BMP, although there are other obstacles to emoji support as well: https://github.com/mapbox/mapbox-gl-native/issues/15666#issuecomment-532969076.

## Diagnosis

The most naïve fix for this issue would entail removing some places where glyph IDs are capped at 65,535 and increasing the data type to something wider than `UInt16`:

[...]

Issue #4541, opened 2024-08-12
Replace some manual script detection with regular expressions


1ec5:
For several years, I’ve been manually maintaining a relevant subset of the Unicode Character Database that GL JS uses to detect scripts for vertical layout, ideographic line breaking, complex text layout, and right-to-left layout. This solution works reasonably efficiently, but it’s difficult to keep it up-to-date as Unicode evolves, because relatively few contributors are familiar with these parts of the Unicode standard or the associated text layout conventions.

I think the custom script detection logic should be replaced by Unicode-aware regular expressions where possible. I’ve identified two such opportunities; there may be others.

It should be possible to replace the following function:

https://github.com/maplibre/maplibre-gl-js/blob/007ad7b6e58dc13e12366f2cdf4f648a741bc0f8/src/util/script_detection.ts#L268-L275

with the following code:

```js
export function charInComplexShapingScript(char: number) {
    return /\p{Sc=Arab}/u.test(String.fromCodePoint(char));
}
```

and replace the following function:

https://github.com/maplibre/maplibre-gl-js/blob/007ad7b6e58dc13e12366f2cdf4f648a741bc0f8/src/util/script_detection.ts#L37-L63

with the following code:

```js
export function charAllowsIdeographicBreaking(char: number) {
    return /\p{Ideo}/u.test(String.fromCodePoint(char));
}
```

This will make the functions more future-proof, so that we don’t need to notice that a new CJK-related block has been added to Unicode etc. The downside is that this requires a relatively recent version of ECMAScript. This codebase has [never used](https://github.com/search?q=repo%3Amaplibre%2Fmaplibre-gl-js+%2F%5C%2F%5Cw*u%5Cb%7CRegExp%5C%28%5B%27%22%5D%5B%5E%27%22%5D%2B%3F%5B%27%22%5D%2C%5Cs*%5B%27%22%5D%5Cw*u%2F&type=code) a Unicode regular expression so far, let alone a [Unicode character escape](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Unicode_character_class_escape). Does GL JS still maintain capability with Chrome 63, Firefox 77, or Safari 11.0 or earlier versions of these browsers?

[...]

PR #4550, opened 2024-08-13
Render non-BMP CJKV characters locally

1ec5:
Generalized glyph management, shaping code to operate on UTF-16 characters instead of code units, no longer splitting surrogate pairs. Enabled local glyph rendering for additional ideographic code blocks in the Supplementary Ideographic Plane (SIP) and Tertiary Ideographic Plane (TIP) for improved Chinese, Japanese, Korean, and Vietnamese (CJKV) character coverage.

Enabled local glyph rendering in some additional CJKV code blocks in the SIP and TIP, as well as in Tangut, which is thrash-prone like CJKV. Disabled letter spacing for more cursive scripts.

This is a step towards #2307, which tracks support for characters outside the Basic Multilingual Plane more generally, not just for CJKV.

## Demonstration

Before | After
:--:|:--:
[那市镇 (Na Hang)](https://americanamap.org/#map=12/22.35126/105.38323&language=zh)<br>[那🌫️市镇](https://www.openstreetmap.org/node/369498042) | 那𧯄市镇 (Na Hang)<br>[那𧯄市镇](https://www.openstreetmap.org/node/369498042)
[丐市镇](https://americanamap.org/#map=12/10.33544/106.0343&language=zh)<br>[丐🌫️市镇](https://www.openstreetmap.org/node/2169297915) | 丐𦨭市镇<br>[丐𦨭市镇](https://www.openstreetmap.org/node/2169297915)
[拜社](https://americanamap.org/#map=13/10.41175/104.03079&language=zh)<br>[拜🌫️社](https://www.openstreetmap.org/node/8261100606) | 拜𩡋社<br>[拜𩡋社](https://www.openstreetmap.org/node/8261100606)

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


src/symbol/shaping.ts
```diff
@@ -130,7 +130,7 @@ class TaggedString {
     }
 
     length(): number {
-        return this.text.length;
+        return [...this.text].length;
```
HarelM, line 133: Why is this needed?

src/render/glyph_manager.ts
```diff
@@ -119,13 +125,57 @@ export class GlyphManager {
...
+             unicodeBlockLookup['Kana Extended-A'](id) ||
+             unicodeBlockLookup['Kana Extended-B'](id) ||
+             unicodeBlockLookup['Kana Supplement'](id) ||
+             unicodeBlockLookup['Kangxi Radicals'](id) ||
+             unicodeBlockLookup['Katakana'](id) ||
+             unicodeBlockLookup['Katakana Phonetic Extensions'](id) ||
+             unicodeBlockLookup['Small Kana Extension'](id) ||
+             unicodeBlockLookup['Tangut'](id) ||
+             unicodeBlockLookup['Tangut Components'](id) ||
+             unicodeBlockLookup['Tangut Supplement'](id));
```
1ec5, line 178: Should we go ahead and throw this open to any character above U+FFFF, unconditionally rendering them locally? It’s not like anyone has ever published glyph PBFs for non-BMP ranges. We’d probably need to rename the `localIdeographFontFamily` option to be less CJKV-centric. If the developer wants to provide a third-party font that isn’t installed on the system, the page’s stylesheet could define an [`@font-face`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face) rule, and it could even be segmented by [`unicode-range`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range). Not every font allows Web font distribution, but of the fonts that support non-BMP characters, most are open-source anyways.

src/util/script_detection.ts
```diff
@@ -40,21 +53,40 @@ export function charAllowsIdeographicBreaking(char: number) {
 
     if (isChar['Bopomofo Extended'](char)) return true;
     if (isChar['Bopomofo'](char)) return true;
+    if (isChar['CJK Compatibility'](char)) return true;
```
HarelM, line 56: Is there a better way to manage this? Seems like we are changing similar code in different places and are creating a error prone process.
can the method in `is_char_in_unicode_block.ts` return something more complicated to avoid the need to add these multiple times?

test/integration/render/tests/text-local-ideographs/cjk/style.json
```diff
@@ -21,16 +21,18 @@
                 "properties": {
                     "name_en": "abcde",
                     "name_ja": "あいうえお",
-                    "name_zh": "阿衣乌唉哦",
-                    "name_ko": "아이우"
+                    "name_zh": "阿衣乌唉哦𪚥",
```
HarelM, line 24: Is this the only render test that cover these changes? Do we need other tests?

src/util/script_detection.ts
1ec5: Yes, #4541 will turn this function into a one-liner. I left it out of this PR to keep it from scope-creeping, but it I can tackle it next.

src/util/verticalize_punctuation.ts
```diff
@@ -89,20 +89,28 @@ export const verticalizedCharacterMap = {
...
-        if (canReplacePunctuation && verticalizedCharacterMap[input[i]]) {
-            output += verticalizedCharacterMap[input[i]];
+        if (canReplacePunctuation && verticalizedCharacterMap[char.value]) {
+            output += verticalizedCharacterMap[char.value];
         } else {
-            output += input[i];
+            output += char.value;
         }
+
+        prevChar = {value: char.value, premature: false};
```
HarelM, line 111: I'm not sure I understand the `permutate` flag here. 
This seems like a change of logic from the original method, so I would consider adding some unit tests here to make sure we get the right results.

HarelM:
You can download the html report from the CI and commit the image that you download from the CI so that the CI will pass, assuming the image is correct.

test/integration/render/tests/text-local-ideographs/cjk/expected-mac.png
1ec5, line 1: This render test is failing on Linux and Windows because I haven’t updated the expected screenshot for those platforms. Is it possible to access the render test dashboard on the CI machine, or would I need to set up virtual machines to run the tests myself on these platforms?

src/util/script_detection.ts
HarelM: I prefer to have it in this PR if possible and see that the changes are in one (or maybe two) place only.
If it's not possible, then I would prefer a PR to solve the "redundancy" problem first and only afterwards get this PR in.

test/integration/render/tests/text-local-ideographs/cjk/expected-mac.png
HarelM: You can download the report from the CI summary and copy the images from the HTML (right click, copy image).
Let me know if the instructions are not clear and you need more help.
Bottom line, you don't need a virtual machine for this.

src/symbol/shaping.ts
HarelM: I see, thanks for the info!

src/render/glyph_manager.ts
HarelM: Not sure I understand the question, but feel free to simplify this if condition.
I wouldn't want to change a public API at this point as part of this PR (`localIdeographFontFamily`).
We can consider changing this for as a version 5 breaking change.
See #3834

src/util/verticalize_punctuation.ts
HarelM: I think unit tests can cover the logic in this method, and a render test or two can help here as well to keep parity with native and make sure this is working as expected.

HarelM:
CC: @louwers there are expected changes in parity here maybe, just FYI.

src/render/glyph_manager.ts
```diff
@@ -92,8 +92,14 @@ export class GlyphManager {
...
         const range = Math.floor(id / 256);
-        if (range * 256 > 65535) {
-            throw new Error('glyphs > 65535 not supported');
+        const isInBMP = range * 256 <= 0xFFFF;
+        if (!isInBMP) {
+            if (this._doesCharSupportLocalGlyph(+id)) {
+                entry.ranges[range] = true;
+                return {stack, id, glyph: null};
+            } else {
+                throw new Error('glyphs > 65535 not supported');
```
wipfli, line 101: Since you put everything in place for support of characters outside of the basic multilingual plane I would suggest to support server-side glyph ranges above 2**16 instead of throwing the error, or am I missing something?

src/render/glyph_manager.ts
1ec5: It might be possible to avoid a breaking change. For example, we could maintain both `localIdeographFontFamily` and `localFontFamily`, giving the developer an opportunity to specify one font family for local rendering in general and another for CJKV specifically. This could be convenient because CJKV tends to be monospace whereas other scripts tend not to be.

src/symbol/shaping.ts
```diff
@@ -75,7 +75,7 @@ export type TextJustify = 'left' | 'center' | 'right';
 const PUAbegin = 0xE000;
 const PUAend = 0xF8FF;
 
-class SectionOptions {
+export class SectionOptions {
```
HarelM, line 78: Do we really need to export those classes? Can we test the content of this file without exporting those?

src/render/glyph_manager.ts
wipfli: > Should we go ahead and throw this open to any character above U+FFFF, unconditionally rendering them locally?

No, we should keep the same logic as we have for codepoints in the BMP.

HarelM:
BTW, is this a breaking change in terms of behavior of the map/text/fonts?

src/render/glyph_manager.ts
1ec5: This error isn’t strictly necessary anymore. However, I’m unsure if anyone would be surprised by GL JS suddenly starting to request glyph PBFs for non-BMP ranges above 65,535. It’s very unlikely that anyone has ever published such glyph PBFs, so we would need to decide whether to encourage that or draw a line in the sand about local font rendering being the future for scripts that haven’t historically been supported on the server side.

wipfli:
> BTW, is this a breaking change in terms of behavior of the map/text/fonts?

No

src/render/glyph_manager.ts
wipfli: TinySDF is a good solution for CJK to save bandwidth. However, the visual quality is worse than server-side generated glyphs especially on low-res devices. There are also CJK readers who do not use local glyph rendering, cc @Kanahiro. So I would say let's follow the same approach like with have in the BMP...

wipfli: Great idea for the openhistorical map!

We could also let the user decide how MapLibre should use TinySDF with an additional map option similar to `localIdeographFontFamily : string | false`...

src/util/is_char_in_unicode_block.ts
```diff
@@ -312,34 +312,34 @@ export const unicodeBlockLookup: UnicodeBlockLookup = {
...
+    'CJK Unified Ideographs Extension I': (char) => char >= 0x2EBF0 && char <= 0x2EE5F,
+    'CJK Unified Ideographs Extension G': (char) => char >= 0x30000 && char <= 0x3134F,
+    'CJK Unified Ideographs Extension H': (char) => char >= 0x31350 && char <= 0x323AF,
+    'CJK Compatibility Ideographs Supplement': (char) => char >= 0x2F800 && char <= 0x2FA1F,
     // 'Tags': (char) => char >= 0xE0000 && char <= 0xE007F,
     // 'Variation Selectors Supplement': (char) => char >= 0xE0100 && char <= 0xE01EF,
-    // 'Supplementary Private Use Area-A': (char) => char >= 0xF0000 && char <= 0xFFFFF,
-    // 'Supplementary Private Use Area-B': (char) => char >= 0x100000 && char <= 0x10FFFF,
+    'Supplementary Private Use Area-A': (char) => char >= 0xF0000 && char <= 0xFFFFF,
+    'Supplementary Private Use Area-B': (char) => char >= 0x100000 && char <= 0x10FFFF
```
1ec5, line 344: @wipfli, if I understand [this article](https://oliverwipfli.ch/devanagari-in-the-protomaps-basemap-with-a-positioned-glyph-font-for-maplibre-2024-06-30/) correctly, your solution for Devanagari text shaping uses Private Use Area codepoints but has to steer clear of image placeholders that are also assigned to this block. With the changes in this PR, you could probably use the much more spacious Supplementary Private Use Areas instead. Most CJKV fonts informally use these blocks for characters that haven’t been officially encoded yet, but I wouldn’t expect Web maps to take advantage of these characters due to interoperability issues. You’d still be constrained by `SymbolBucket.MAX_GLYPHS`, but at least you could pick a different plane without worrying about overlapping with unrelated usage before you reach that limit.

wipfli: That's correct, your pull request would be very useful for positioned glyph fonts!

src/render/glyph_manager.ts
1ec5: I simplified this code to only enable local glyph rendering for codepoints that Unicode considers to be `Ideographic` (CJKV ideographs, strictly speaking) and anything else that we’ve enabled ideographic line breaking for. Line breaking isn’t strictly relevant to whether a given script randomly accesses a huge swath of codepoints, but the ideographic breaking test happens to align with the broader set of scripts that prefer local rendering for consistency (especially in Japanese).

src/util/script_detection.ts
1ec5: This PR is now based on #4560, which unifies the codepoint-testing functions as much as possible. `charAllowsIdeographicBreaking` didn’t turn out to be a one-liner, as I had hoped, but it’s a lot simpler than it would’ve been had I continued to list out every affected Unicode block explicitly.

src/render/glyph_manager.ts
1ec5: I removed the error, so that non-CJKV characters beyond U+FFFF are fetched from the server. On balance, I think it’ll ends up being simpler for a developer to generate the missing glyph PBFs than figure out how to wire up Web fonts, provided that the glyph PBF generators are updated to be UTF-16–aware. This approach also allows for different typefaces for the same scripts, whereas an approach based on TinySDF would force the use of a single typeface regardless of the fontstack (through no fault of TinySDF’s).

src/symbol/shaping.ts
1ec5: The new tests need to refer to these classes. Would you suggest a helper function that encapsulates the creation of a `TaggedString`, but only for testing purposes?

https://github.com/maplibre/maplibre-gl-js/blob/249b360f1a603236378985e88a9df960c9a097e5/src/symbol/shaping.test.ts#L17-L19

1ec5:
The Ubuntu build is failing on what seems to be a test that the bundle hasn’t increased in size. I’m unsure what I’ve touched that would cause the bundle to be more than three times the expected size.

```
Error: expect(received).toBeLessThan(expected)

Expected: < 1024
Received:   3176

    at Object.<anonymous> (test/build/min.test.ts:41:45)
```

HarelM:
I've fixed it in the global branch, it's not that the size had tripled but only that you added 2k, which is fine, just increase the expected size so that the test will pass...

src/render/glyph_manager.ts
```diff
@@ -92,8 +92,10 @@ export class GlyphManager {
         }
 
         const range = Math.floor(id / 256);
-        if (range * 256 > 65535) {
-            throw new Error('glyphs > 65535 not supported');
+        const isInBMP = range * 256 <= 0xFFFF;
+        if (!isInBMP && this._doesCharSupportLocalGlyph(+id)) {
+            entry.ranges[range] = true;
+            return {stack, id, glyph: null};
         }
```
wipfli, line 99: I can't understand this part. Can you explain in words what it does?

wipfli:
Just to double-check, the MVT spec does support UTF-16?

src/render/glyph_manager.ts
1ec5: Previously, the code would throw an error if the codepoint belonged to a 255-character PBF range beyond the BMP (above U+FFFF). I thought it would be necessary at this point to special-case characters that can be rendered locally, but it turns out that we would’ve returned just above here in that case, making this early return entirely redundant. I’ve removed it.

1ec5:
> Just to double-check, the MVT spec does support UTF-16?

As of MVT v2.1, a feature property value [is a `string`](https://github.com/mapbox/vector-tile-spec/blob/5330dfc6ba2d5f8c8278c2c4f56fff2c7dee1dbd/2.1/vector_tile.proto#L19), which the protocol buffer format defines as [UTF-8–encoded text](https://protobuf.dev/programming-guides/proto3/#scalar). Both UTF-8 and UTF-16 are variable-width encodings. The Chinese character 𪚥 (U+2A6A5) is encoded in UTF-8 as a sequence of four code units: `F0 AA 9A A5`. JavaScript decodes this sequence as a surrogate pair of two UTF-16 code units: `D869 DEA5`. Either way, it’s still a single character. Previously, we were incorrectly splitting the character in half, effectively assuming the obsolete UCS-2 encoding. Now we group the code units together by Unicode character, as fonts and users would expect.

Although this is an improvement, it isn’t quite perfect: some characters and emoji require variation selectors, so that what the user perceives as a grapheme consists of several UTF-16 characters. This PR doesn’t address that issue, because GL JS still pervasively assumes that one codepoint equals one glyph.

src/render/glyph_manager.ts
```diff
@@ -93,13 +93,9 @@ export class GlyphManager {
 
         const range = Math.floor(id / 256);
         const isInBMP = range * 256 <= 0xFFFF;
-        if (!isInBMP) {
-            if (this._doesCharSupportLocalGlyph(+id)) {
-                entry.ranges[range] = true;
-                return {stack, id, glyph: null};
-            } else {
-                throw new Error('glyphs > 65535 not supported');
```
1ec5: I removed this error at @wipfli’s request in https://github.com/maplibre/maplibre-gl-js/pull/4550#discussion_r1714995266, but unless we solve #4563, this will be a breaking change and a difficult one for developers to cope with.

1ec5: #4564 provides a fallback to local glyph rendering, so it won’t be a breaking change after all. The behavior is essentially what I had in this PR originally, but server-side glyphs take precedence over client-side glyphs for non-CJKV codepoints.

src/util/script_detection.ts
```diff
@@ -79,13 +79,22 @@ export function charAllowsIdeographicBreaking(char: number) {
     // Return early for characters outside all ideographic ranges.
     if (char < 0x2E80) return false;
 
-    if (isChar['CJK Compatibility Forms'](char)) return true;
     if (isChar['CJK Compatibility'](char)) return true;
+    if (isChar['CJK Compatibility Forms'](char)) return true;
```
HarelM, line 83: This still looks like we are repeating the process of uncommenting a line in `unicodeBlockLookup` and then adding it as another "if" row here.
Is it possible to change this in one place instead of two?

src/symbol/shaping.ts
HarelM: No, as far as I understand (which is not a lot), these classes were used inside this file by other exported classes or methods.
If this is the case, I think it might be better to test the exported classes/methods in the unit test in a flow that reaches these non exported classes.
If this is not possible, I would then prefer to have these classes moved to their own file and have unit test to these new files separately. In general, I like it better when I have a single file for a single class, but that might just be me...

src/util/script_detection.ts
1ec5: Do you mean that we should inline the lookup functions wherever we use them instead of keeping them in a separate lookup table? We might as well eliminate that file in favor of regular expressions like `[\uFE30-\uFE4F]`, along with inline comments giving each character class a name.

I think the separate lookup table was created originally out of concern that these character classes would be less readable or more error-prone. I don’t have a strong preference, but then again, I have a higher tolerance for regular expressions than most developers.

src/symbol/shaping.ts
1ec5: I think we’ll have to split the classes into separate files. The other exported functions in this file do so much that it would be difficult to isolate the most important behavior that these tests are intended to exercise.

src/util/script_detection.ts
HarelM: It's not about regular expression (I don't like them much, they are not readable when things get complicated), but more about using the same hard coded string in two places.
I don't mind keeping the original lookup table, but then why do we pass the same hardcoded string here again? Wht not use something like `Object.keys(unicodeBlockLookup)...`?

1ec5: None of the functions that use `unicodeBlockLookup` should ever iterate over all the keys. Each function only selectively accesses the lookup functions relevant to the task at hand. As the table in https://github.com/maplibre/maplibre-gl-js/pull/4560#discussion_r1716668107 illustrates, it’s a big Venn diagram; every function cares about a slightly different set of Unicode blocks. I tried to unify the functions somewhat, but inherently we need to list specific blocks in each function in order to have the correct behavior.

If the problem is that these strings add to the bundle size, we could CamelCase them (as in maplibre-native) and use the property syntax so that maybe the property names will get minified at compile time.

HarelM: No, I don't care about bundle size in this respect. 
What I would like to achieve is the need to change in multiple places.
i.e. add all the needed information to the lookup table so that uncommenting a line and setting the right properties would be the only change need to support another script.

1ec5: `charAllowsIdeographicBreaking` is the only function that simply checks blocks as a whole, whereas `charHasUprightVerticalOrientation` and `charHasNeutralVerticalOrientation` just use the blocks as shortcuts before going deeper and checking individual codepoint ranges. So we’d be rewriting `unicodeBlockLookup` to look something like this:

```js
export const unicodeBlockLookup: [String: UnicodeBlock] = {
  'CJK Compatibility Forms': {
    lower: 0xFE30,
    upper: 0xFE4F,
    charSupportsLocalGlyphs: (char) => true,
    charAllowsIdeographicBreaking: (char) => true,
    charHasUprightVerticalOrientation: (char) =>
      !((char >= 0xFE49 /* dashed overline */ && char <= 0xFE4F) /* wavy low line */),
    charHasNeutralVerticalOrientation: (char) => true,
  }

  'Latin-1 Supplement': {
    lower: 0x0080,
    upper: 0x00FF,
    charSupportsLocalGlyphs: (char) => false,
    charAllowsIdeographicBreaking: (char) => false,
    charHasUprightVerticalOrientation: (char) => true,
    charHasNeutralVerticalOrientation: (char) => (char === 0x00A7 /* section sign */ ||
      char === 0x00A9 /* copyright sign */ ||
      char === 0x00AE /* registered sign */ ||
      char === 0x00B1 /* plus-minus sign */ ||
      char === 0x00BC /* vulgar fraction one quarter */ ||
      char === 0x00BD /* vulgar fraction one half */ ||
      char === 0x00BE /* vulgar fraction three quarters */ ||
      char === 0x00D7 /* multiplication sign */ ||
      char === 0x00F7 /* division sign */),
  },

  // …
};
```

This is a vastly distilled simplification of the relevant Unicode character database files (linked in the comments). Technically, none of these functions should be based on character blocks at all, and we should be using gargantuan regular expressions instead. But it would be more modular I suppose.

1ec5: Just to be clear, the only reason we’d ever modify this file is that the Unicode standard itself has changed (to add another script). If this PR works as designed, we shouldn’t have to manually add new scripts in general, though that is sort of what’s going on in `charInSupportedScript`, which I haven’t touched.

1ec5:
If this PR lands before maplibre/maplibre-style-spec#778 is fixed, there will be regressions in some styles even for some features that only involve BMP characters.

HarelM:
Style spec has a version of its own, so fixing it there first will be the right approach, after that we can merge this PR only with the upgrade of style-spec to the required version. Bottom line, there's no problem syncing it properly, we just need to make sure it is solved and applied correctly.

src/util/script_detection.ts
HarelM: I think the above suggestion is better in terms that in encapsulates all the relevant properties to a single place and later on you won't need to worry about them, hopefully, or you could create another method to have a different "query" on this object.
You can also define the type of every element so that it adheres to the relevant model.
I find it more maintainable, but it might just be me, IDK...

1ec5: I haven’t abandoned this PR. 🙂 In 1ec5/maplibre-gl-js#1, I’m prototyping, among other things, code generation for anything we’re doing based on the Unicode Character Database that isn’t directly supported by ECMAScript regular expressions. It has the potential to reduce maintenance overhead for script detection, though I suspect we’ll still need to roll a little bit of our own code for some remaining typographical functionality that isn’t expressed in the Unicode standard.

1ec5: The bespoke build script I wrote in that draft PR is simple and effective for this purpose, but I wonder if you’d prefer to pull in [`@unicode/unicode-15.1.0`](https://www.npmjs.com/package/@unicode/unicode-15.1.0) as a devDependency instead. It’s a big package to pull in at install time, but unlike my script, it won’t require fetching anything from the network. Either way, it’ll have negligible impact on the build, just a big regular expression that we fortunately wouldn’t have to maintain by hand.

HarelM: Dev dependency is better I think. 
Make sure to add the generated file to git ignore and add a .g.ts file extension to the generated file to indicate that it's generated.

HarelM:
@1ec5 version 20.3.1 of the style spec was merged to this repo. 
Does it require any changes in this repo or this is more of a bug-fix or small improvement? 
Is it related to this PR?
I would like to release a new version today-tomorrow but I don't know if I should revert the spec package update or not. 
Please let me know.

1ec5:
The risk I mentioned before was only if we landed this PR before maplibre/maplibre-style-spec#778, but the other way around should be safe. I don’t think anything will break if you release GL JS with just the upgrade to the style-spec dependency, so there’s no need to revert. This PR needs some additional work that I’ll be staging shortly in separate PRs.

HarelM:
Thanks for the info! I've published a new version - 4.7.

HarelM:
- How does this PR relates to: #4564?

I'm trying to reduce the number of open PRs as I see that I miss some review due to PR noise.
Is this not solved as part of #4564? Can this PR be closed or not?

1ec5:
This is a separate task that I’ll be picking back up after #4564 lands. The difference is that #4564 is only concerned with how to render a character, whereas this PR is concerned with correctly splitting the `text-field` into characters.

HarelM:
This PR is out of sync.
Closing it for now.
Feel free to let me know if you want to reopen it or open a different one.

src/util/script_detection.ts
1ec5: #6639

1ec5:
I reprised this PR in #6640. I’ve only rebased it atop #4564, but this does mean that it fully fixes #2307 instead of only getting halfway there. Some of the review feedback in this PR went into #4564 and #6639. There’s still some outstanding feedback here about tests that I haven’t gotten to yet.

src/symbol/shaping.ts
1ec5: Resolved in https://github.com/maplibre/maplibre-gl-js/pull/6640#discussion_r2505089726.

src/util/verticalize_punctuation.ts
1ec5: Added unit and integration tests in https://github.com/maplibre/maplibre-gl-js/pull/6640#discussion_r2505089726.

PR #4551, opened 2024-08-13
Fix/remaining issues camera inside terrain

chrneumann:
## Launch Checklist

Add buffer so that camera doesn't sit directly on terrain. Prevent crashes when zoom is wrongly calculated.

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


src/geo/transform.ts
```diff
@@ -538,12 +538,15 @@ export class Transform {
...
-        this._center = center;
-        this.zoom = zoom;
+        try {
+            const zoom = this.scaleZoom(requiredScale);
+            // update matrices
+            this._elevation = elevation;
+            this._center = center;
+            this.zoom = zoom;
+        } catch (_e) {
+            console.error(`Could not recalculate zoom. requiredScale: ${requiredScale}`);
```
HarelM, line 548: Can we add a check to `scalezoom` method instead of a general try-catch?

HarelM:
@chrneumann where are we with this?

chrneumann:
> @chrneumann where are we with this?

Waiting for your reply, see the open code review.

HarelM:
Hu?
Can you link to it?

chrneumann:
> Hu? Can you link to it?

https://github.com/maplibre/maplibre-gl-js/pull/4551/files/fe6191c166ac876ad07af08b277ec3e3d4538a9b

HarelM:
Are you sure you have submitted your response?
This is what I see on my side:
image

chrneumann:
> Are you sure you have submitted your response?

Ok, sorry, I did *not* submit the review.

src/geo/transform.ts
HarelM: My question is if we can replace the try-catch with if (scaleZoomPosdible) {...

I would prefer if the check would be specific instead of a try catch, if possible.

chrneumann: Okay, I was also misguided by the `scaleZoom`. This does not throw, but the `zoom` setter in the last line of the `try {}`. The details of the zoom calculation are rather obscure to me and thus I don't know when and why the zoom calculation fails. Maybe there is someone which more expertise in the calculation? Still, I think it wold be best to don't set elevation and center if `requiredScale` is invalid. I could wrap only the `this.zoom = zoom` call in the try {} and reset elevation and zoom in the `catch`.

HarelM: Right... the zoom is a setter under the hood... We will be fixing this in the globe branch...

HarelM: Let's do that then - reduce the try catch scope.
Maybe @kubapelc can help understand when calcmatrices can fail.

chrneumann: Ok, reworked that. I had to do the recalculation in a clone, as it would not be possible to easily reset the transform if recalculation fails.

HarelM:
Ok, looks good, can you check why the test is failing?

chrneumann:
> Ok, looks good, can you check why the test is failing?

Fixed that, a render test failed because of the readded buffer between terrain surface and camera.

HarelM:
Should this PR still be in draft mode?

HarelM:

chrneumann:
Done. New transform stuff also merged.

src/geo/transform.ts
```diff
@@ -0,0 +1,1084 @@
+import {LngLat} from './lng_lat';
```
HarelM, line 1: This file was removed, please merge the relevant code to the right place.

chrneumann: Oh, yeah, that slipped through. Fixed.

HarelM:
Looking at the fact that you added code without a failing test can indicate that the added code in the transform class is not covered by tests?

src/geo/transform.ts
kubapelc: Hi, sorry for not responding earlier, I must have somehow overlooked this...

I've looked at the current code in a debugger and it seems that the root issue is that in the `recalculateZoom` function, the variable `requiredScale` was `-338620.5476139354` in my case, which results in a NaN when plugged into `scaleZoom`, which is just `log2(scale)` internally, and logarithm of a negative value returns NaN. The NaNs then propage in calcMatrices, eventually throwing.

I think the issue is that `requiredAltitude` can sometimes be negative. I don't understand how altitude and elevation work in relation to terrain and zoom and how they are computed.

Maybe some check that `requiredAltitude` is negative can be used instead of try-catch?

HarelM:
Did you push the commit? I still see that the code coverage is not getting to this catch block...

HarelM:
Yes, regardless of if the solution is an "if" or a try-catch, there's a need to add a test the reproduces this issue (I think a unit test is the right kind of test here, but I don't mind either way).
Once you have a scenario that can be reproduced we can better talk about the solution, I believe.

chrneumann:
I can't reproduce the crash any more in the current main branch. Still kept the added surface padding (camera stays not directly on the ground), as it prevents some graphic glitches. 

But there seems to be a regression with the `transformCameraUpdate`. Possibly because the camera transform is updated in the new changes without considering this option. Needs a closer look.

chrneumann:
> But there seems to be a regression with the `transformCameraUpdate`. Possibly because the camera transform is updated in the new changes without considering this option. Needs a closer look.

Started with 5.0.0-pre.4

HarelM:
@NathanMOlson is this change problematic for what you are working on?
I think this is a good addition to the camera elevation logic, but it might complicate things for your use case.
@chrneumann is this ready to me merged?

HarelM (dismissed):

chrneumann:
> @chrneumann is this ready to me merged?

Yes, it's ready.

src/ui/camera.ts
```diff
@@ -1238,13 +1240,14 @@ export abstract class Camera extends Evented {
      *
      * @param tr - The transform to check.
      */
-    _elevateCameraIfInsideTerrain(tr: ITransform) : { pitch?: number; zoom?: number } {
+    _elevateCameraIfInsideTerrain(tr: ITransform): { pitch?: number; zoom?: number } {
         if (!this.terrain && tr.elevation >= 0 && tr.pitch <= 90) {
             return {};
         }
+        const surfacePadding = Math.min(500, 20 * (25 - tr.zoom));
```
NathanMOlson, line 1247: Why do we need this surface padding?

At a typical zoom of 15, this sets the minimum altitude at 200m. This amount of surface is problematic for ground-level use cases (like an aircraft on the runway, or touring a city at eye-level.)

NathanMOlson, line 1247: The linear scaling with zoom level doesn't make sense to me. I think it would be more appropriate to have something like `500 / (2**tr.zoom)`.

test/integration/render/tests/terrain/fog/expected.png
NathanMOlson, line 1: The "expected" image shows this bug: https://github.com/maplibre/maplibre-gl-js/issues/4859

test/integration/render/tests/high-pitch/terrain-pitch95/expected.png
NathanMOlson, line 1: I don't think any of the "expected" images for these render tests should change.  The fact that they are changing means that the pitch angle is being adjusted, and I don't think it should be in any of these cases.

HarelM:
Thanks @NathanMOlson! These are interesting questions and observations.

src/ui/camera.ts
chrneumann: It can help prevent graphic glitches when the camera is directly moving on the surface. Do these use cases have terrain enabled? If so, one could make the padding optional.

chrneumann: I don't have strong arguments for the current calculation, but why do you think non-linear is better?

test/integration/render/tests/high-pitch/terrain-pitch95/expected.png
chrneumann: It's shifted because the padding is added which results in a changed angle (the center stays the same).

src/ui/camera.ts
NathanMOlson: Because the distance from camera to center point (in meters) is inversely proportional to `2**tr.zoom`

If you increase zoom by 1, you are cutting the distance from camera to center point in half. So it seems like you should also cut the padding in half.

NathanMOlson: The "aircraft on runway" (https://github.com/maplibre/maplibre-gl-js/issues/4717) use case (https://github.com/maplibre/maplibre-gl-js/issues/4717) will definitely have terrain enabled.

Samarth1696:
Tbh I would like to disable this feature for the use case for aircraft. Currently motion control works with `jumpTo()` in every frame. From what I have seen from my work of motion control that the camera goes under the terrain regardless of `elevateCameraIfInsideTerrain` is being called, this is because of `jumpTo()` I think. This feature would be correct for the use case for normal users but things might go wrong for aircraft motion control use case.

We can have a on and off switch for this feature where the user can choose between going under the terrain and not going under it. I know for the default UX no one wants to go under the Terrain but we have to find a solution for the aircraft use case.

birkskyum:
@Samarth1696 , if you'd like to disable it, one approach you might consider at least temporarily (granted that i don't know the details of your use case) is to use the `globe` projection type. It does have some terrain support now, but there are no checks in place related to the camera hitting them.

clement-igonet:
I measured this on current main: three engines from the same commit, four camera journeys over Innsbruck terrain, worst penetration below the surface.


| journey | main | this PR alone | with the lookup fix |
|---|---|---|---|
| zoom into slope | 143 m under · guard blind 63% | 96 m clear · blind 64% | 18 m clear · blind 0% |
| walk into mountain | 338 m under · blind 100% | 338 m under · blind 100% | 18 m clear · blind 0% |
| pitch into terrain | 266 m under · blind 100% | 266 m under · blind 100% | 17 m clear · blind 0% |
| pan (real drag) | ~1300–1700 m under · blind ~80% | identical to main, to the decimal | 25–60 m under · blind 0% |


Two journeys are unchanged to the metre, because the lookup underneath returns 0: _getOverscaledTileIDFromLngLatZoom builds tile IDs from the fractional live zoom, the keys never match the cache, and the padding is added to zero.

Flooring the zoom there makes all four come out clear, and 50m of padding is then enough where 500 m lifts the camera visibly.

Branch with both, merged onto current main, CHANGELOG-only conflict and tsc clean: https://github.com/clement-igonet/maplibre-gl-js/tree/continue/camera-inside-terrain-1542

Demo, the same three engines, journeys at a quarter speed: https://clement-igonet.github.io/maplibre-gl-js/terrain-1542/

HarelM:
I can't really trust the report above. @clement-igonet I suggest to tone down the usage of AI and start spending time looking into issue yourself, this is adding noise and wastes my time over and over again.
When I run it the fist time I saw the look up go under the terrain but reported as "clear 7m", running it a second time, the harness seems to agree with me that it's not without flaws:
image

PR #3592, opened 2024-01-16
add examples using deck.gl layers

ilci66:
## Launch Checklist

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.

This change adds two basic examples displaying how can deck.gl layers can be added to an existing maplibre base.

PR #4552, opened 2024-08-13
Add an example to draw a radius with Turf.js

clementgayvallet:
This PR adds an example to show how to draw a radius with Turf.js. This is a typical use-case when a user wants to display an approximation of a location such as a real estate property.

The access token used in this example is dedicated and will work from `localhost` and `maplibre.org` origins.

## Launch Checklist


test/examples/draw-a-radius.html
```diff
@@ -0,0 +1,69 @@
...
+        body { margin: 0; padding: 0; }
+        html, body, #map { height: 100%; }
+    </style>
+</head>
+<body>
+<script src="https://cdn.jsdelivr.net/npm/@turf/turf@7/turf.min.js"></script>
+<div id="map"></div>
+<script>
+    // Don't forget to replace with your own access token
+    const accessToken = '0SMf4Ib7tchV9KXSTr9hHKZhtsPaf4XP2FW4gs4UO7E97cPae1CEuTf7owi0NLdn';
```
HarelM, line 20: Please use `demotiles` instead.

test/examples/draw-a-radius.html
```diff
@@ -0,0 +1,69 @@
...
+            steps: 64,
+            units: 'kilometers'
+        };
+        const circle = turf.circle(radiusCenter, radius, options);
+
+        // Add a fill layer with some transparency
+        map.addLayer({
+            id: 'location-radius',
+            type: 'fill',
+            source: {
```
HarelM, line 43: I would consider adding a source and use that source in both layers, I believe it is more readable.

HarelM:
This is a good example.

test/examples/draw-a-radius.html
clementgayvallet: I first thought about using `demotiles`, but the style doesn't have street-level details, so the example would make less sense with a country-wide radius. Also, I was inspired by this PR (#3592) where a style with an API key is used.

HarelM: Yes, I missed that, it's less than ideal, can't you simply use OSM raster baselayer?

HarelM (approved):
THANKS!

PR #4560, opened 2024-08-14
Simplify script detection using regular expressions

### LINKED ISSUE #4541: Replace some manual script detection with regular expressions

1ec5:
Simplified some internationalization-related code to match a given character against a regular expression that tests the Unicode script instead of manually checking whether it falls inside the ranges of a series of Unicode blocks. This is more future-proof because it offloads some decisionmaking onto the browser. Unfortunately, the most verbose methods related to vertical orientation must still be maintained by hand, because the ECMAScript standard doesn’t allow browsers to support the other relevant Unicode character properties. But at least the diff in #4550 will end up a bit shorter as a result.

Fixes #4541.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


src/render/glyph_manager.ts
```diff
@@ -120,13 +119,14 @@ export class GlyphManager {
...
+        // wasteful bandwidth consumption. For visual consistency within CJKV
+        // text, also include any other CJKV or siniform ideograph or hangul,
+        // hiragana, or katakana character.
         return !!this.localIdeographFontFamily &&
-            (unicodeBlockLookup['CJK Unified Ideographs'](id) ||
-            unicodeBlockLookup['Hangul Syllables'](id) ||
-            unicodeBlockLookup['Hiragana'](id) ||
-            unicodeBlockLookup['Katakana'](id));
-        /* eslint-enable new-cap */
+            /\p{Ideo}|\p{sc=Hang}|\p{sc=Hira}|\p{sc=Kana}/u.test(String.fromCodePoint(id));
```
1ec5, line 129: Originally, local glyph rendering was designed to be a solution to the random access problem, in which a very typical run of CJK text would require wastefully downloading a whole distinct glyph PBF for each character in the text. For this reason, it was limited to two huge blocks, CJK Unified Ideographs (20,992 codepoints spread across 82 different PBFs) and Hangul Syllables (11,172 codepoints across 44 PBFs).

While that was a technically sound design, it produced messy output in Japanese text, which typically switches freely between kanji (from CJK Unified Ideographs) and katakana. mapbox/mapbox-gl-js#8365 fixed this inconsistency, adding the tiny blocks Hiragana (93 codepoints) and Katakana (96 codepoints) even though neither block requires more than a single PBF.

This change acknowledges that reality, because the alternative would be to hard-code a much longer list of Unicode blocks, as seen in https://github.com/maplibre/maplibre-gl-js/pull/4550#pullrequestreview-2234965653.

src/ui/map.ts
```diff
@@ -289,8 +289,8 @@ export type MapOptions = {
     fitBoundsOptions?: FitBoundsOptions;
     /**
      * Defines a CSS
-     * font-family for locally overriding generation of glyphs in the 'CJK Unified Ideographs', 'Hiragana', 'Katakana' and 'Hangul Syllables' ranges.
-     * In these ranges, font settings from the map's style will be ignored, except for font-weight keywords (light/regular/medium/bold).
+     * font-family for locally overriding generation of Chinese, Japanese, and Korean characters.
```
1ec5, line 292: This documentation was too specific. I don’t think generalizing it to talk about CJK would risk confusing developers. It shouldn’t be a breaking change; after all, when mapbox/mapbox-gl-js#8365 introduced Hiragana and Katakana to this list, it went out in v1.2.0 without even being mentioned in the release notes.

src/util/script_detection.ts
```diff
@@ -23,43 +23,71 @@ export function allowsLetterSpacing(chars: string) {
...
+/**
+ * Returns a regular expression matching the given script codes, excluding any
+ * code that the execution environment lacks support for in regular expressions.
+ */
+function sanitizedRegExpFromScriptCodes(scriptCodes: Array<string>): RegExp {
+    const supportedPropertyEscapes = scriptCodes.map(code => {
+        try {
+            return new RegExp(`\\p{sc=${code}}`, 'u').source;
+        } catch (e) {
+            return null;
```
HarelM, line 35: When does this fail that this requires try-catch?

1ec5: Unicode keeps adding more scripts every year, and different browsers seem to be updating their Unicode databases at different paces. For example, Firefox 128 recognizes Chorasmian (`Chrs`), but Node v20.11 does not. An unrecognized property escape results in an exception:

> Uncaught SyntaxError: invalid property name in regular expression

We have to swallow this exception without affecting the rest of the regular expression. This check runs only once before we ever test a character against the regular expression.

src/util/script_detection.ts
```diff
@@ -23,43 +23,71 @@ export function allowsLetterSpacing(chars: string) {
...
     if (isChar['Ideographic Description Characters'](char)) return true;
-    if (isChar['Kangxi Radicals'](char)) return true;
-    if (isChar['Katakana Phonetic Extensions'](char)) return true;
-    if (isChar['Katakana'](char)) return true;
     if (isChar['Vertical Forms'](char)) return true;
-    if (isChar['Yi Radicals'](char)) return true;
-    if (isChar['Yi Syllables'](char)) return true;
-
-    return false;
+    return ideographicBreakingRegExp.test(String.fromCodePoint(char));
```
HarelM, line 90: Is there a way to remove the other "if"s as well? Or use a similar mechanism?

1ec5: Unfortunately, I haven’t found a way to simplify these methods any further. [ECMAScript only allows](https://tc39.es/ecma262/multipage/text-processing.html#sec-runtime-semantics-unicodematchproperty-p) testing for a limited set of non-binary properties (`General_Category`, `Script`, and `Script_Extensions`). Most characters are well-served by this script-based regular expression, but there are [a number of anomalies](https://www.unicode.org/reports/tr24/#Script_Anomalies) in which miscellaneous characters are assigned to the “common” script. The remaining conditionals approximate these anomalies.

I also wasn’t able to unify this method with the others very much, because they form a Venn diagram. For example:

Script | Ideographic breaking | Vertical orientation | Local glyph rendering
----|----|----|----
Canadian Aboriginal | ✖️ | ✔️ | ✖️
Egyptian Hieroglyphs (#4550) | ✖️ | ✖️ | ✔️
Latin | ✖️ | ✖️ | ✖️
Han | ✔️ | ✔️ | ✔️
Hangul | ✖️ | ✔️ | ✔️

1ec5: On the native platforms, it may be possible to further simplify the corresponding methods. At least iOS/macOS has access to ICU, which comes with [its own regular expression syntax](https://unicode-org.github.io/icu/userguide/strings/regexp.html) that supports the full complement of Unicode character properties, as well as [direct access to the character property table](https://unicode-org.github.io/icu/userguide/strings/properties.html).

HarelM:
I'm having a hard time understanding the difference between the two PR related to these files.
Can we close one of them and move all the changes to one PR?
Currently I'm jumping between one to the other and trying to find the difference, which is less than ideal.
If one could be a PR against another branch or if you want me to merge one of the PR to a branch so that I can see the addition between the two PRs.

1ec5:
#4560 is now based on this PR, so you can ignore the first five commits as part of that PR. Unfortunately I can’t tell GitHub to set the target of that branch to this branch, because I created both PRs from a fork.

I recommend reviewing the changes in this PR first. If you agree, merge this PR to main; then #4560 will automatically rebase atop main to become much simpler without these five commits.

1ec5:
Sure, have at it!

HarelM (approved):

Issue #50, opened 2020-12-29
Unicode font do not display properly in map.


aerotypebharat:
This is the name in openstreetmap.org 
![image](https://user-images.githubusercontent.com/10284102/103306225-b004da80-49d2-11eb-9dfc-cf48f34d2de8.png)
This is what displayed in map. 
![image](https://user-images.githubusercontent.com/10284102/103306313-e3476980-49d2-11eb-8c42-5865024b0137.png)
Test code available in code pen here
https://codepen.io/snickell/pen/dypOWzj

aerotypebharat:
I opened this issue in Mapbox gl js. Can anyone give me direction how this can be solved to support Devanagari letters

HarelM:
I'm not sure I fully understand the problem. It would be best if you elaborate.
Regardless, you'll probably need to download a font that support these letters and use it instead of the current font.
I had the same problem with hebrew letters I think...

aerotypebharat:
![image](https://user-images.githubusercontent.com/10284102/126075804-f0d7758d-a92c-4fb3-864b-57c610179a96.png)
This vowel character is supposed to go in front of the consonant. 
![image](https://user-images.githubusercontent.com/10284102/126075868-fbb44a12-7b7c-43aa-a985-e356d6caa05f.png)

HarelM:
Ahh, I have the same problem with "nikud" - the hebrew extra characters to indicate how to read, they appear after the letter instead of within the letter.
I have no clue how to solve this, you'll need to research how text in webgl is created and how to add these letters in the right place.
Sorry I don't have better knowledge on the subject. But if you find out please serve a PR :-)

ambientlight:
@aerotypebharat: What fontface is used in your screenshot? Do you have a public fontface that renders Devanagari correctly at https://fonts.google.com/ ? You can then use https://github.com/orangemug/font-glyphs to generate prerendered fonts (Signed Distance Field glyphs) to consume from your style

aerotypebharat:
I am using Khula font. It is public in https://fonts.google.com/. It renders properly in text. I have tried other fonts that support Devanagari but maplibre-gl.js does not have code to handle vowels and special characters of Devanagari.

anarchodin:
The approach chosen by Mapbox to render labels is unfortunately fundamentally incompatible with scripts that require more sophisticated handling than a mechanical typewriter could provide. This is not a problem with Unicode support as such; there is no problem with the encoding.

[...]

PR #4564, opened 2024-08-15
Fall back to local glyph rendering if glyph PBF is unavailable

### LINKED ISSUE #4563: Glyph PBF request failure prevents entire map from loading
If any character in a `text-field` causes GL JS to request a glyph PBF that’s unavailable for some reason, the entire map fails to load.

## Example

In [this example](https://jsbin.com/fedasil/9/edit?html,output), the variable `char` is set to `\uFFFC` (￼). The specified fontstack lacks a glyph PBF at `https://demotiles.maplibre.org/font/Open%20Sans%20Semibold/65280-65535.pbf`. GL JS receives an HTTP 404 error and refuses to load the map at all, even unrelated features in other tiles.

A | ￼
----|----
A map with a gray background and a label that reads “ABC A ABC” | A blank map

At the very least, GL JS should skip over the character but render the rest of the map. Ideally, though, it would fall back to local glyph rendering, since even a crude rendering of the glyph would be less confusing than no rendering.

## Impact

One can imagine a denial of service scenario in which someone litters OpenStreetMap with hidden control characters that don’t render in a standard text renderer but cause GL JS to refuse to render any part of the map.

In https://github.com/maplibre/maplibre-gl-js/pull/4550#discussion_r1714995266, I was asked to remove the artificial limitation that prevents GL JS from even trying to render non-CJK characters above U+FFFF. But this imposes a new requirement on the developer that the fontstack must provide new glyph PBFs for every 255-character range from U+FFFF to U+10FFFF. This is significant time commitment, given that some of the most popular glyph PBF generators don’t support codepoints beyond U+FFFF and there are fewer fonts to choose from. It’s also a lot of files to host, probably for little justification. If the developer doesn’t supply these new glyph PBFs, then upgrading to a release of GL JS that contains #4550 would expose their map to a nontrivial number of places and POIs in East Asia whose names would essentially crash the map.

A fallback would eliminate this burden, allowing the browser to render the text as a best effort. This is consistent with standard font fallback behavior in CSS-compliant browsers.

Previously reported as mapbox/mapbox-gl-js#9528 mapbox/mapbox-gl-js#9546.

1ec5:
If a glyph PBF is unavailable, try to fall back to local glyph rendering, and keep rendering the map without crashing.

A label “ABC” followed by the Object Replacement Character followed by “ABC” against a gray background.

Fixes #4563.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


src/render/glyph_manager.ts
```diff
@@ -108,7 +109,22 @@ export class GlyphManager {
             entry.requests[range] = promise;
         }
 
-        const response = await entry.requests[range];
+        let response;
```
HarelM, line 112: Can we move this block to a different method maybe? This method is already too long I think

HarelM: Also, looking at the code above this part, can we now handle the case where `url` is not defined? I know it's not optimal, but there are situations where you just want the text to be render without finding a glyph server. You'd need to define `localIdeographFontFamily` anyway, which doesn't have a default, so you'd need to know what you are doing...

HarelM:
Looks good, thanks!

src/render/glyph_manager.ts
1ec5: Ultimately, I would love to make glyph PBFs completely optional. However, the style specification currently states that [`glyphs` is required](https://maplibre.org/maplibre-style-spec/root/#glyphs) as long as any layer has a `text-field` property. Would relaxing that requirement constitute a breaking change in either the style spec or GL JS?

src/render/glyph_manager.ts
```diff
@@ -129,13 +145,13 @@ export class GlyphManager {
             /\p{Ideo}|\p{sc=Hang}|\p{sc=Hira}|\p{sc=Kana}/u.test(String.fromCodePoint(id));
     }
 
-    _tinySDF(entry: Entry, stack: string, id: number): StyleGlyph {
+    _tinySDF(entry: Entry, stack: string, id: number, force: boolean): StyleGlyph {
         const fontFamily = this.localIdeographFontFamily;
```
1ec5, line 149: I’m wary of choosing a font for non-ideographic text based on an option called `localIdeographFontFamily`. The choice should be up to the style author, in the stylesheet, and it should vary from layer to layer, just as in CSS. The iOS implementation has long attempted to [match the `text-font` property to a local or system font](https://github.com/maplibre/maplibre-native/blob/8ad4a51f2faf2e93a8799be8f16c3634e4e504f7/platform/darwin/src/local_glyph_rasterizer.mm#L106-L123). I think GL JS should attempt some matching along these lines.

Unfortunately, many non-Mapbox-hosted fontstacks are named according to the styles for which they were created, obscuring the PostScript names that they were created from. For example, OpenHistoricalMap has “[OpenHistorical Bold](https://github.com/OpenHistoricalMap/map-styles/tree/1ad7f22a04465c32ce82ffaac3aaee0df119648b/fonts/OpenHistorical%20Bold)”, and OSM Americana has “[Americana-Bold](https://github.com/osm-americana/fontstack66/blob/03a43671ee5881e6aab783ccd6cea396c67d5b1e/README.md#usage)”. These styles’ authors can’t simply append Open Sans and Noto Sans to the array, respectively, because `text-font` isn’t a fallback list. It simply joins the array into a comma-separated list and includes it in the request URL, so the request would fail and we’d be right back to where we started.

In order to typeset these layers in the designer’s choice of fonts, we would need to introduce a new layout property, such as `text-font-family` or `text-font-names`, to represent a cascading fallback list of CSS font selectors to choose from. `text-font` would remain for backwards compatibility or where a fontstack is preferred for some reason.

src/render/glyph_manager.ts
HarelM: I don't think it will be a breaking change as changing a required field to optional doesn't sound like a breaking change (unless you have invalid styles, which I'm not sure are interesting).
Relaxing it would make a far better developer experience, IMHO, as I find these glyphs requirement non-friendly...

HarelM: But this should be brought in the style spec repo first...

src/render/glyph_manager.ts
HarelM: I think fallback is better than introducing a new field which would create confusion. 
If the style and the server are not aligned, presenting something is better than not presenting anything, IMHO.

spatialillusions:
Ping. Any plans for merging this?

HarelM:
I think @1ec5 mentioned there is work that he is planning to do before this can be merged as far as I remember...

1ec5:
Yes, the to-do list looks like:

* [x] Relax the requirement for `glyphs`
* [x] Go back to only using `localIdeographFontFamily` for ideographic ranges
* [x] Try to match `text-font` to a local font (pass in the property value to TinySDF as the `fontFamily`)

These changes are straightforward but may require a proposal to change the style spec. I haven’t gotten around to that part yet.

HarelM:
@1ec5 any updates on this?
I would like to clean up some of the PR that are open for some time...

src/render/glyph_manager.ts
1ec5: maplibre/maplibre-style-spec#1045

louwers:
There should (also) be an API for this in my opinion.

I don't want people to author styles without a `glyphs` property just to take advantage of local font rendering.

HarelM:
@1ec5 where are we with regards to this PR?

1ec5:
We merged maplibre/maplibre-style-spec#1068, which unblocks this PR, so the rest is pretty straightforward. However, maplibre/maplibre-style-spec#1204 just landed too – would it be necessary to implement that feature within the scope of this PR, or can we stage it one font selection mechanism at a time?

HarelM:
I don't think it should be part of this PR.
The issue is tracked here:
https://github.com/maplibre/maplibre-gl-js/issues/50
And can (and should probably) be solved in a diff PR.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Fix white artifacts when using non-zero elevation ([#6032](https://github.com/maplibre/maplibre-gl-js/pull/6032))
 - Fix geolocate control lock loss on window resize ([#3504](https://github.com/maplibre/maplibre-gl-js/issues/3504))
 - Fix a memory leak in `GeoJSONSource` when rapidly updating data ([#6163](https://github.com/maplibre/maplibre-gl-js/pull/6163))
+- If a glyph PBF is unavailable or lacks a glyph for a character in a `text-field`, try to render it locally instead of crashing. ([#4564](https://github.com/maplibre/maplibre-gl-js/pull/4564))
```
HarelM: More of a feature than a bug fix, isn't it?

1ec5: It’s both. On the one hand, it’s a crash fix. On the other hand, the ability to render non-CJK characters locally is quite a feature. So I’ll move it up top.

src/render/glyph_manager.ts
HarelM: The original request to extract this to a method is still relevant :-)

src/render/glyph_manager.ts
HarelM, line 1: Should the error for `throw new Error('glyphsUrl is not set');` be removed?

test/integration/render/tests/text-local-glyphs/missing/style.json
```diff
@@ -0,0 +1,43 @@
...
+                    "coordinates": [0, 0]
+                },
+                "properties": {
+                    "name_en": "Georgia",
+                    "name_ka": "საქართველო"
+                }
+            }
+        }
+    },
+    "glyphs": "local://glyphs/{fontstack}/{range}.pbf",
```
HarelM, line 28: Can we add a test that doesn't have a glyphs url defined or is this not part of this PR?

src/render/glyph_manager.ts
1ec5: Yes, that’s what we’ll need to do to conform to the style spec change making `glyphs` optional.

1ec5, line 1: > There should (also) be an API for this in my opinion.

Where should this API live? I’m thinking of a `glyphs` property alongside `localIdeographFontFamily` in [StyleOptions](https://maplibre.org/maplibre-gl-js/docs/API/type-aliases/StyleOptions/), so that you can set it when calling `setStyle()`. Otherwise, an option to set `glyphs` at another point in the style’s lifecycle could have unpredictable results, depending on whether the glyph manager has attempted to render any text by the time the property changes.

HarelM: This sounds reasonable, I was thinking something along the lines of `allowLocalFontRendering` or something similar that should default to false to allow backward compatibility, what do you think?

1ec5: This option would override `glyphs` with anything, or more likely, `false` would delete it from the style. When a developer asks to override something in the style, backwards compatibility wouldn’t normally be a concern. Plus, any style lacking a `glyphs` property was previously invalid and crashy anyways. But `glyphs` isn’t a very intuitive name, so maybe `remoteGlyphsURL`?

HarelM: The current implementation, as far as I understand it, falls back to local glyph rendering when glyphs aren't available on the server. It doesn't remove the `glyph` property from the style, and I'm not sure it should either, right?
If `glyphs` is empty there's a question of what should be done, and if the default should be to render something (which I think is a better UX, but maybe can change things for app owners, IDK) or keep the current behavior.

But I think those are two different aspects of this feature, right?

1ec5: I don’t think there’s any practical reason to keep the current behavior in the event of a missing or unavailable `glyphs`. @louwers requested an API out of concern that the new behavior would incentivize style authors to create styles that don’t specify `glyphs`. I’m working on a native implementation of this feature in parallel, making that less of a concern. Still, an API could enable the use of local fonts even if the developer is working with an older style that can’t be upgraded to remove `glyphs` for some reason. I can address any API in a separate PR if you agree.

louwers: Maybe we can this explicitly opt-in in the style instead of adding an API for it.

```json
{
  "font-faces": {
    "Trickster": [
      { "local": "Trickster" }
    ]
  }
}
```

This mirrors CSS' [`@font-face`](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face).

And older clients can just continue to use `glyphs` and would ignore `font-faces`.

HarelM: I don't have an objection to this, but as stated in a previous comment, style spec changes are better discussed in the style spec repo. 

I guess the question here is if there's a requirement to not fall back to tinysdf, I'm not sure there's one and I'm not sure it's worth adding an API to facilitate something that's not needed.

I would recommend, as part of this PR to simply fall back to tinySDF in case a glyph is missing or glyphs style definition is empty. 

It's a straight forward feature/bug fix.

We can later on decide on other improvements. 
If this fallback feature is expected to be included in native an sdk table needs to be added to the glyphs style spec docs (if it's not already there, although I think it is).

1ec5: maplibre/maplibre-style-spec#1204 added `font-faces` and `local`, but as discussed above, that’s out of scope for this PR (which is essentially a bug fix about missing tiles) and will be handled separately.

test/integration/render/tests/text-local-glyphs/missing/style.json
1ec5: This depends on maplibre/maplibre-style-spec#1347.

src/render/glyph_manager.ts
1ec5: This depends on maplibre/maplibre-style-spec#1347.

src/render/glyph_manager.ts
1ec5: I factored out the warning code, which took up a good chunk of what I had added, but the rest is kind of tricky because of the async code flow.

src/render/glyph_manager.ts
```diff
@@ -22,6 +25,9 @@ type Entry = {
     tinySDF?: TinySDF;
 };
 
+/// The style specification hard-codes some last resort fonts as a default fontstack.
```
HarelM: וnit-picking: use `/**` types of comments instead as IDEs can present them on hover, which helps when reading some part of the code...

src/render/glyph_manager.ts
```diff
@@ -87,7 +93,9 @@ export class GlyphManager {
             return {stack, id, glyph};
         }
 
-        glyph = this._tinySDF(entry, stack, id);
+        if (!this.url || this._charUsesLocalIdeographFontFamily(id)) {
+            glyph = this._tinySDF(entry, stack, id);
```
HarelM: I think that as opposed to before, glyph here is always defined, right?
In that case the next if can be moved into this one, am I right?

src/render/glyph_manager.ts
```diff
@@ -102,26 +110,43 @@ export class GlyphManager {
             return {stack, id, glyph};
         }
 
-        if (!this.url) {
-            throw new Error('glyphsUrl is not set');
-        }
-
-        if (!entry.requests[range]) {
+        if (this.url && !entry.requests[range]) {
```
HarelM: Is `this.url` in this if is always true? I think it is, right?
In that case there's no need to check if it's not null.
I'm asking this since the next statement is odd as it awaits `entry.requests[range]` which might not be defined if url is not defined...

src/render/glyph_manager.ts
```diff
@@ -102,26 +110,43 @@ export class GlyphManager {
...
         }
 
-        const response = await entry.requests[range];
+        let response;
+        try {
+            response = await entry.requests[range];
+        } catch (e) {
+            glyph = this._tinySDF(entry, stack, id);
+            this._warnOnMissingGlyphRange(glyph, range, id, e);
+            if (glyph) {
```
HarelM: Is this if necessary?

src/render/glyph_manager.ts
HarelM: Reading the code again, I don't think I made the correct calculations.
But there's just too many code paths here merged into a single place.
I would advise to split the scenarios so that is clear which part of the code handles what.
For example check only if url is defined, if it's not, all the request part and async stuff is not relevant, so you can handle that (empty glyphs) and later on know the url is defined in the rest of the code.

src/render/glyph_manager.ts
1ec5: I’ve adjusted the logic so that `localIdeographFontFamily` only affects the font when we’re using TinySDF because of CJK and `text-font` isn’t overwriting the default. Otherwise, we fill in `sans-serif` as a generic font family, which stylistically matches the default default of Open Sans and Arial Unicode MS in the style specification.

src/render/glyph_manager.ts
1ec5: Oh, I thought most code editors equate `///` with  `/**` and `//*` these days (as opposed to `//`). That’s been common since the Doxygen days. I can change it for consistency with other code though.

src/render/glyph_manager.ts
```diff
@@ -102,26 +113,43 @@ export class GlyphManager {
...
+                entry.glyphs[id] = glyph;
+                return {stack, id, glyph};
+            }
+        }
         for (const id in response) {
-            if (!this._doesCharSupportLocalGlyph(+id)) {
+            if (!this.url || !this._charUsesLocalIdeographFontFamily(+id)) {
```
HarelM: This part of the for loop should be inside the try-catch I think, if the response is null this might crash, and in any case is not relevant, so no need to move it below the catch, I think...

src/render/glyph_manager.ts
HarelM: Yeah, let's keep everything similar...

src/render/glyph_manager.ts
1ec5: The next block needs to run even if this block does not run (such as if `glyphs` is specified in the stylesheet and it isn’t a CJK codepoint).

src/render/glyph_manager.ts
```diff
@@ -143,38 +171,38 @@ export class GlyphManager {
...
+        // The CJK fallback font specified by the developer takes precedence over the last resort fontstack in the style specification.
+        const usesLocalIdeographFontFamily = stack === defaultStack && this.localIdeographFontFamily !== '' && this._charUsesLocalIdeographFontFamily(id);
+        let fontFamily = usesLocalIdeographFontFamily ? this.localIdeographFontFamily : stack;
 
         // Client-generated glyphs are rendered at 2x texture scale,
         // because CJK glyphs are more detailed than others.
         const textureScale = 2;
 
-        let tinySDF = entry.tinySDF;
+        let tinySDF = usesLocalIdeographFontFamily ? entry.ideographTinySDF : entry.tinySDF;
```
HarelM: I would consider moving all this tinySDF creation and caching to a new method, something like `getTinySDFAndCacheIt` which will return it from the "cache" if it exists or create it if not.

src/render/glyph_manager.ts
1ec5: It isn’t. Looks like I managed to remove all the code paths where `_tinySDF()` could return undefined, so we no longer need to guard against that possibility.

src/render/glyph_manager.ts
HarelM: But the only way `glyph` would be defined after this block is if this "if" was hit, as there's a check just prior to this "if" block to check if `glyph` is defined, right?

In any case, see my next comment about splitting the code to different code flows.
https://github.com/maplibre/maplibre-gl-js/pull/4564/files#r2479568451

HarelM:
Sorry for the huge number of comments on such a short block of code and not so drastic changes, but it's already hard to understand it, so I want to make it a little bit better from readability aspect.
I hope you'll forgive me :-)

src/render/glyph_manager.ts
HarelM: [Optional] I would also consider renaming this method to something like generateGlyphUsingTinySDF, as tinySDF is an object/method that is being used to generate a glyph and this method is not a simple wrapper around tinySDF after all these changes and modifications, I think.
The fact that there's a `const tinySDF = something` inside the method means it's not as simple...

test/integration/render/tests/text-local-glyphs/missing/style.json
HarelM: Unresolving this comment as I think it was not addressed still, right?

1ec5:
I understand and appreciate it! It was already spaghetti code before I added at least two other variables to the mix. #4550 and 1ec5/maplibre-gl-js#1 will result in even more churn in this area, so I may not be as aggressive in refactoring this code as I normally would be.

src/render/glyph_manager.ts
```diff
@@ -3,11 +3,14 @@ import {loadGlyphRange} from '../style/load_glyph_range';
 import TinySDF from '@mapbox/tiny-sdf';
 import {unicodeBlockLookup} from '../util/is_char_in_unicode_block';
 import {AlphaImage} from '../util/image';
+import {warnOnce} from '../util/util';
 
 import type {StyleGlyph} from '../style/style_glyph';
 import type {RequestManager} from '../util/request_manager';
 import type {GetGlyphsResponse} from '../util/actor_messages';
 
+import {v8} from '@maplibre/maplibre-gl-style-spec';
```
1ec5, line 12: I think [this assertion failure](https://github.com/maplibre/maplibre-gl-js/actions/runs/18955727185/job/54131907412?pr=4564#step:6:26) started after I imported the style specification here:

> AssertionError: expected 958463 to be less than 958010

I’m going to assume that another 400&nbsp;kilobytes or so are OK.

HarelM: I think there should be similar imports elsewhere, aren't there?
Or is this bringing the entire docs and examples that are stripped out intentionally?

src/render/glyph_manager.ts
1ec5: I broke out a separate method for creating the TinySDF instance, which I expect to grow over time. I also found a more concise way to express memoizing the CJK versus non-CJK TinySDF instances in the existing method.

src/render/glyph_manager.ts
1ec5: I’ve renamed the method to simply `_drawGlyph`, since TinySDF is an implementation detail.

src/render/glyph_manager.ts
1ec5: It’s only pulling in the style spec JSON file. I can’t think of anything else I would’ve done to increase the bundle size by this amount, unless it’s something to do with the very incremental style-spec dependency upgrade. But that patch version only _deletes_ code. 🤔

test/integration/render/tests/text-local-glyphs/missing/style.json
1ec5: Oops, I must’ve accidentally wiped out the commit that I added this test in during a rebase. I’ve restored it manually.

src/render/glyph_manager.ts
1ec5: Oh, I see what you mean. The earlier code that initializes `glyph` already catches and returns early if we fetched it from the cache.

1ec5: I’ve merged the two conditionals per your earlier suggestion. I’ll see about refactoring the overall method based on your other feedback.

src/render/glyph_manager.ts
1ec5: Oh, I see that there’s an `increaseQuota` of 1024&nbsp;bytes. We were probably eating away at that allowance before my changes pushed us over the top.

src/render/glyph_manager.ts
1ec5: I split out a helper method to handle requesting and responding to server-side glyphs.

src/render/glyph_manager.ts
1ec5: I moved the rest of the method into the try-catch block. This particular conditional goes away because it’s a precondition of the method.

test/examples/style-labels-with-web-fonts.html
```diff
@@ -0,0 +1,89 @@
...
+                                },
+                                "geometry": {
+                                    "type": "Point",
+                                    "coordinates": [137.9054972, 36.3044083]
+                                }
+                            },
+                            {
+                                "type": "Feature",
+                                "properties": {
+                                    "name": "Matsumoto"
```
1ec5, line 54: Rampart One looks a lot more interesting in Japanese than in English. Unfortunately, the stylesheet provided by Google Fonts only loads Latin by default. We could force it to show some Japanese characters, but Google has subsetted Rampart One so granularly that we’d effectively have to manually specify which individual characters to load via CSS. This is more or less a symptom of broader usability issues with Web fonts for CJK, mirroring the challenges that led MapLibre to default to local font rendering for CJK. In principle, we could make CSS font loading a little more turnkey by manually loading fonts within the glyph manager, but for the mechanism that Google Fonts uses, it would amount to reverse engineering their API. I think it’s better for now to just give an example of how to load non-CJK text this way.

test/examples/style-labels-with-web-fonts.html
```diff
@@ -0,0 +1,89 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>Style labels with Web fonts</title>
+    <meta property="og:description" content="Apply Web fonts to your style’s text labels. Unlike signed distance field (SDF) glyph sets, Web fonts are available from a variety of providers, or your can make your own using popular tools. This option is suitable for fonts that are only available through a third-party content delivery network (CDN) for technical or legal reasons, as well as fonts that are incompatible with SDF, such as variable fonts. For compatibility with Android and iOS applications, specify equivalent fonts in the style’s font-faces property.">
```
1ec5, line 5: I’m handwaving about `font-faces`, because it isn’t possible to use both `font-faces` in MapLibre Native and the new local fallback in GL JS at the same time. Adding the code block in [the documentation](https://maplibre.org/maplibre-style-spec/font-faces/) causes the stylesheet to fail validation.

src/render/glyph_manager.ts
```diff
@@ -82,46 +103,67 @@ export class GlyphManager {
...
+        // Start downloading this range unless we’re currently downloading it.
         if (!entry.requests[range]) {
             const promise = GlyphManager.loadGlyphRange(stack, range, this.url, this.requestManager);
             entry.requests[range] = promise;
         }
 
-        const response = await entry.requests[range];
-        for (const id in response) {
-            if (!this._doesCharSupportLocalGlyph(+id)) {
+        let response;
```
HarelM: Can we move this `let` into the try? I don't think it is used outside, right?

test/examples/style-labels-with-web-fonts.html
HarelM: Maybe we should omit this example for now and introduce it only when `font-faces` will be implemented? I think it will create more clarity on when to use what, because while this will work, I think it's not the optimal way to do it from a cross platform consideration.
Your call, I don't have strong feelings either way...

src/render/glyph_manager.ts
```diff
@@ -73,6 +93,7 @@ export class GlyphManager {
     }
 
     async _getAndCacheGlyphsPromise(stack: string, id: number): Promise<{stack: string; id: number; glyph: StyleGlyph}> {
+        // Create an entry for this fontstack if it doesn’t already exist.
```
HarelM, line 96: This method looks a lot better now, thanks!!

HarelM:
I've added some last minor comments and this can be merged afterward. Thanks for all the work on this!

test/examples/style-labels-with-web-fonts.html
1ec5: I think it’s probably a good idea to include some sort of documentation on how to use Web fonts, since conventional HTML text typically requires a different step than `document.fonts.load`. We’ll get questions about it anyways. I’m hoping that referring readers to the `font-faces` documentation will help them learn more about the compatibility situation than we’re able to stuff inside a meta tag.

HarelM (approved):

test/examples/style-labels-with-web-fonts.html
1ec5: #6637 tracks `font-faces` support.

src/render/glyph_manager.ts
```diff
@@ -209,4 +224,63 @@ export class GlyphManager {
             }
         };
     }
+
+    _createTinySDF(stack: String | false): TinySDF {
+        // Escape and quote the font family list for use in CSS.
+        const fontFamilies = stack ? stack.split(',') : [];
+        fontFamilies.push(defaultGenericFontFamily);
+        const fontFamily = fontFamilies.map(fontName =>
+            /[-\w]+/.test(fontName) ? fontName : `'${CSS.escape(fontName)}'`
```
1ec5, line 233: This attempts to escape multiword font names in `text-font`, but the regular expression only checks whether the font name has any letter or hyphen at all. It needs to be surrounded in `^…$` to detect font names in need of escaping.

This is not necessarily a real problem in practice. Blink, Gecko, and WebKit all seem to be lenient towards `font` strings like `24px Comic Sans MS, sans-serif` without needing quotation marks, so at worst this is just dead code that can be removed for clarity.

Issue #4279, opened 2024-06-16
Allow adding multiple layers to map.on() event handler


AgentSmith0:
## User Story

It is currently not possible to directly create an event handler for multiple layers. (https://maplibre.org/maplibre-gl-js/docs/API/classes/Map/#on)

## Rationale

It would be more convient to directly pass a layer list instead of having to create the same event handler again for every layer in a list.
Mapbox GL JS has this feature already implemented: https://github.com/mapbox/mapbox-gl-js/pull/11114

## Impact

You would have to create a new event handler for every layer in your list e.g. through a for loop.

HarelM:
Feel free to pursue this. 
Please make sure not to infringe mapbox copyright by copying their code.

birkskyum:
Closed by #4570

PR #4570, opened 2024-08-16
Map#on, once and off - support for multiple layers

pstaszek:
Implements: #4279 

This PR adds support for adding listener for multiple layers using `Map#on` and `Map#once`. It also adds support for removing listeners with `Map#off` function.

To remove listener in `off` function the user has to provide array of layerIds with the same content as when creating the listener. I chose this approach because it seems the simplest, there is no need to consider complex cases (i.e. remove only part of layers from listener).

I intentionally did not check the Mapbox GL JS implementation before I wrote my code. I did it after and to some degree it is similar, because it is the most obvious implementation. Is it ok, or should we try to find some other way to implement this feature?

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
Can you check if the coverage is correct? 
Otherwise, looks good.

src/ui/map.ts
```diff
@@ -1205,15 +1205,16 @@ export class Map extends Camera {
         return this._rotating || this.handlers?.isRotating();
     }
 
-    _createDelegatedListener(type: keyof MapEventType | string, layerId: string, listener: Listener): {
-        layer: string;
+    _createDelegatedListener(type: keyof MapEventType | string, layerIds: string[], listener: Listener): {
+        layers: string[];
```
HarelM, line 1209: Can we create a return type?

src/ui/map.ts
```diff
@@ -1463,29 +1493,48 @@ export class Map extends Camera {
...
+
         const removeDelegatedListener = (delegatedListeners) => {
             const listeners = delegatedListeners[type];
             for (let i = 0; i < listeners.length; i++) {
                 const delegatedListener = listeners[i];
-                if (delegatedListener.layer === layerIdOrListener && delegatedListener.listener === listener) {
+                if (
+                    delegatedListener.listener === listener &&
+                    delegatedListener.layers.length === layerIds.length &&
+                    delegatedListener.layers.every((layerId: string) => layerIds.includes(layerId))
```
HarelM, line 1536: Is there a possibility for this not to work somehow?

pstaszek: Yes, you're correct.

I have chosen the approach that user has to provide the same layer ids to remove delegate, as provided when creating it. I believe it is the simplest approach, that does not create edge cases. If someone wants to modify layer list, he/she needs to remove listener, and register new.

Consider for example situation, when someone creates two delegates with the same callback, but different set of arrays with the same layer. If would be dificult to judge from which `off` should remove the layer.

Another edge case. If user will multiple times remove layers, he/she will need to provide exact list of layers that are left. Or - list that contains all of them, but can contain more (for example previously removed).

But then - if we will choose the approach to remove layer - by - layer, we will provide array of layerA and layerB, if there will be two listeners, one with layerA, and one with layerB, should we remove both?

What do you think?

src/ui/map.ts
```diff
@@ -1463,29 +1493,48 @@ export class Map extends Camera {
...
         }
 
+        const layerIds = typeof layerIdsOrListener === 'string' ? [layerIdsOrListener] : layerIdsOrListener as string[];
+
         const removeDelegatedListener = (delegatedListeners) => {
             const listeners = delegatedListeners[type];
             for (let i = 0; i < listeners.length; i++) {
                 const delegatedListener = listeners[i];
-                if (delegatedListener.layer === layerIdOrListener && delegatedListener.listener === listener) {
+                if (
```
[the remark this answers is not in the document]
pstaszek: Done.

src/ui/map.ts
pstaszek: Done.

src/ui/map.ts
neodescis: I've pondered this problem myself before this pull request came up. The conceptual problem is that these are not equivalent:
```
map.on('click', ['layer1', 'layer2'], () => { console.log('hello') });
```
```
const fn = () => { console.log('hello') };
map.on('click', 'layer1', fn);
map.on('click', 'layer2', fn);
```
They are not equivalent because in the first case, the event handler is only called once if there is a click on a spot where the two layers overlap, whereas the second would be called twice. So, this feature is very useful, and not just for more concise syntax!

However, I'm not really sure how removal of the handler should work. Do we limit requiring the same layers to be passed in, as you've done? Or do we allow the layers be removed individually? The more I think about it, the more I think we should do the latter, but it does seem like there are some pitfalls there. For example, what if the same layer+handler was registered twice, each time with the same layer in two different layer arrays? It seems unlikely, but it's possible.

Maybe we can iterate on the behavior? If we went with the approach you have now, and decided to later allow the handler to be removed for a subset of the layers, would that be a breaking change? It doesn't seem like it would be.

HarelM: If we go with the approach of only unregistering if the layer match exactly and add the ability to unregister only for some layers then it's not a breaking change. 
But this needs to be well documented in the "off" method.

pstaszek: I improved mentioned specifically in `off` method description, that the same array has to be provided to unregister listener.

pstaszek:
I have some issue with registering `once` event, but I'm not sure if it is related to this PR. I will investigate it further tomorrow, or during the weekend.

pstaszek:
Ok - this is something unrelated to this PR.
I notice a bug, that it is not possible to deregister the a listener added with `once` method and a layer - this is because `removeDelegatedListener` function looks for the fn reference in `this._delegatedListener`, but `once` method is not saving it there.

src/ui/map.ts
```diff
@@ -1242,24 +1246,47 @@ export class Map extends Camera {
...
                 }
             };
-            return {layer: layerId, listener, delegates: {[type]: delegate}};
+            return {layers: layerIds, listener, delegates: {[type]: delegate}};
+        }
+    }
+
+    _removeDelegatedListener(type: string, layerIds: string[], listener: Listener) {
+        if (!this._delegatedListeners || !this._delegatedListeners[type]) {
+            return this;
```
HarelM, line 1267: Not sure returning "this" is needed. 
I would simply return void.

pstaszek: Right, done.

HarelM:
Can you fix the build test please? Just update the size.

pstaszek:
Ofc, should I update this value: https://github.com/pstaszek/maplibre-gl-js/blob/1c30ee6ce8a9e37558ba0ea2c09b06fe428e7c89/test/build/min.test.ts#L39 ?

HarelM:
Yes please.

HarelM:
Thanks!!!

PR #4577, opened 2024-08-19
Globe - custom layers API and examples, globe dev guide

kubapelc:
This PR adds:

- extended API for custom layers that makes writing custom layers for globe easier
- examples of globe custom layers:
    - a [simple triangle](https://github.com/windycom/maplibre-gl-js/blob/globe-pr-custom/test/examples/globe-custom-simple.html)
    - custom [tiles](https://github.com/windycom/maplibre-gl-js/blob/globe-pr-custom/test/examples/globe-custom-tiles.html)
    - a [3D model](https://github.com/windycom/maplibre-gl-js/blob/globe-pr-custom/test/examples/globe-3d-model.html) on a globe)
- a [writeup / guide](https://github.com/windycom/maplibre-gl-js/blob/globe-pr-custom/developer-guides/globe.md) of how globe projecton works

Follow the links to read the writeup and see the globe custom layer examples. The custom layer args are now projection-dependent, are filled in by the transform class, and allow the user to write custom layers that under both projections.

globe-custom-simple

globe-custom-tiles

globe-3d-model


kubapelc:
@HarelM Please ping people who should also review the custom layer API.

test/examples/globe-custom-simple.html
```diff
@@ -0,0 +1,214 @@
...
+    <script src='../../dist/maplibre-gl-dev.js'></script>
+    <style>
+        body { margin: 0; padding: 0; }
+        html, body, #map { height: 100%; }
+    </style>
+</head>
+<body>
+<div id="map"></div>
+<script>
+    fetchedStyle['projection'] = {'type': 'globe'}; // Modify the fetched style before creating the map so that it already starts with globe projection.
```
HarelM, line 18: This is redundant, right?

test/examples/globe-custom-tiles.html
```diff
@@ -0,0 +1,390 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>Display custom layer tiles on a globe</title>
+    <meta property="og:description" content="Display custom layer tiles on a globe." />
```
HarelM, line 5: Update title and description to be more specific

src/render/draw_custom.ts
```diff
@@ -36,13 +87,7 @@ export function drawCustom(painter: Painter, sourceCache: SourceCache, layer: Cu
...
         context.setDepthMode(depthMode);
 
-        implementation.render(context.gl, painter.transform.customLayerMatrix(), {
-            farZ: painter.transform.farZ,
-            nearZ: painter.transform.nearZ,
-            fov: painter.transform.fov * Math.PI / 180, // convert to radians
-            modelViewProjectionMatrix: painter.transform.modelViewProjectionMatrix,
-            projectionMatrix: painter.transform.projectionMatrix
-        });
+        implementation.render(context.gl, customLayerMatrix, customLayerArgs);
```
HarelM, line 90: `customLayerMatrix` should be part of `customLayerArgs`.
I'm guessing that we will remove `customLayerMatrix` argument as part of version 5.
Version 5 is the globe version I believe.

src/render/draw_custom.ts
```diff
@@ -4,11 +4,62 @@ import {StencilMode} from '../gl/stencil_mode';
...
+    // Since custom layers are expected to supply mercator coordinates [0..1], we need to rescale
+    // the fallback projection matrix by EXTENT.
+    // Note that the regular projection matrices do not need to be modified, since the rescaling happens by setting
+    // the `u_projection_tile_mercator_coords` uniform correctly later.
+    const fallbackMatrixScaled = createMat4f64();
+    mat4.scale(fallbackMatrixScaled, projectionData.u_projection_fallback_matrix, [EXTENT, EXTENT, 1]);
+
+    const transformArgs = transform.getCustomLayerArgs();
+
+    const customLayerArgs: CustomLayerArgs = {
```
HarelM, line 33: Can we get a simpler object? I don't like passing around all these methods.
Custom layer has a reference to map, transfrom, etc, for method. let's try and pass only data...

This need a better design I believe as this is a public API.

test/examples/globe-custom-simple.html
kubapelc: Yes, I forgot to remove this. Will be fixed with the next push.

src/render/draw_custom.ts
kubapelc: I think Transform is not exported in MapLibre, should the examples access it anyway?

If just getting rid of the methods is enough, then maybe something like this could work:

- `getSubdivisionForZoomLevel()` - use `map.style.projection.subdivisionGranularity.tile.getGranularityForZoomLevel()` instead
- `getMatrixForModel` - this could be part of `map`, or we can export ITransform and use something like `map.transform.getMatrixForModel()`
- `getMercatorTileProjectionMatrix` - same as above

Also, the `createTileMesh` method in globe-custom-tiles.html is a direct copypaste of the same method in globe.ts, since it is not exposed by MapLibre. If I were to expose this method, where should it be placed?

src/geo/projection/mercator_transform.ts
```diff
@@ -855,4 +833,40 @@ export class MercatorTransform implements ITransform {
             this.calculatePosMatrix(coord.toUnwrapped());
         }
     }
+
+    customLayerMatrix(): mat4 {
+        return this._mercatorMatrix.slice() as any;
+    }
+
+    getCustomLayerArgs(): CustomLayerArgsTransformSpecific {
+        const current = this;
```
HarelM, line 842: This hack is usually not needed in typescript.

src/render/draw_custom.ts
HarelM: The main object that is a public API is the `Map` object, most public methods are exposed there.
Other "static" methods are exported too as part of a more general static object `maplibregl` (but mainly exported as methods which are not bound to any object or state).

src/geo/projection/globe_utils.ts
```diff
@@ -180,3 +180,15 @@ export function interpolateLngLatForGlobe(start: LngLat, deltaLng: number, delta
         );
     }
 }
+
+export function clampLngLat(val: LngLat, min: LngLat, max: LngLat): LngLat {
```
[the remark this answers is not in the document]
kubapelc: This is currently unused and should not be part of this PR, I'll remove it for now.

src/render/draw_custom.ts
HarelM: Yes, transform is not exposed, but I think the examples use it since you might need some more internal stuff, and I think it's OK to use some internal stuff when it comes to custom layer (for more complicated scenarios).

klokan:
Sounds amazing! Thanks @kubapelc

src/render/draw_custom.ts
kubapelc: Then I will move `createTileMesh` to a separate util file and export it, `getSubdivisionForZoomLevel` can be used by accessing the projection object, `getMatrixForModel `and `getMercatorTileProjectionMatrix` will be part of the transform interface.

src/render/draw_custom.ts
```diff
@@ -4,11 +4,62 @@ import {StencilMode} from '../gl/stencil_mode';
...
+        // Convert all uniforms to plain arrays
+        uniforms: {
+            'u_projection_matrix': [...projectionData.u_projection_matrix.values()],
+            // This next uniform is used to convert from [0..EXTENT] to [0..1] mercator coordinates for a given tile,
+            // but since custom layers are expected to already supply mercator coordinates, it is set to identity (offset 0,0 and scale 1,1).
+            'u_projection_tile_mercator_coords': [0, 0, 1, 1],
+            'u_projection_clipping_plane': [...projectionData.u_projection_clipping_plane.values()],
+            'u_projection_transition': projectionData.u_projection_transition,
+            'u_projection_fallback_matrix': [...fallbackMatrixScaled.values()],
+        },
```
ibesora, line 53: Wondering if a friendlier approach would be to add a new function to the API instead.
Following Maplibre `Program` class, where all the `prelude` functions and uniforms are concatenated to the actual shader source, we could define something like: 
`getProgram(yourVertSource: string, yourFragSource): Program`

This would also enable `drawCustom` to set the values of those uniforms before passing the control to the custom layer rendering code.

developer-guides/globe.md
```diff
@@ -0,0 +1,197 @@
...
+we do not want the planet to get larger or smaller when changing latitudes.
+Map movement thus compensates for the planet size change by also
+changing zoom level along with latitude changes.
+
+This behavior is completely automatic and transparent to the user.
+The only case when the user needs to be aware of this is when
+programmatically triggering animations such as `flyTo` and `easeTo`
+and using them to both change the map center's latitude and *at the same time*
+changing the map's zoom to an amount based on the map's starting zoom.
+The example `globe-zoom-planet-size-function.html` demonstrates how to
```
HarelM, line 32: `globe-zoom-planet-size-function.html` - Use a link here to documentation site, this link can be broken at fist until the globe is merged.

developer-guides/globe.md
```diff
@@ -0,0 +1,197 @@
...
+these tile-local coordinates to web mercator.
+
+The implementation of `projectTile` is automatically injected into the shader source code.
+Different implementations can be injected, depending on the currently active projection.
+Thanks to this many shaders use the exact same code for both mercator and globe,
+although there are shaders that use `#ifdef GLOBE` for globe-specific code.
+
+## Subdivision
+
+If were were to draw mercator tiles with globe shaders directly, we would end up with a deformed sphere.
```
HarelM, line 53:
```suggestion
If we were to draw mercator tiles with globe shaders directly, we would end up with a deformed sphere.
```

src/render/draw_custom.ts
kubapelc: Thank you for your feedback. I'm not sure... Adding the prelude code to the vertex shader is fairly simple from the user's point of view and probably doesn't need a helper function, see [globe-custom-tiles.html (line 227)](https://github.com/windycom/maplibre-gl-js/blob/0fbf4641dc54074377ef0379458b78438f3dd68b/test/examples/globe-custom-tiles.html#L227).

Setting the uniforms is more involved, but a `Program` class would also need to support whatever uniforms the user adds to their shader, and even the 5 uniforms that are passed now (`u_projection_tile_mercator_coords`, etc) can sometimes have user defined values, see [globe-custom-tiles.html (line 334)](https://github.com/windycom/maplibre-gl-js/blob/0fbf4641dc54074377ef0379458b78438f3dd68b/test/examples/globe-custom-tiles.html#L334).

I am honestly not sure how to design this API, where to draw the line between what should be a helper function and what should be left to the user. I think having an exposed Program class would risk that anyone wanting to use custom layers nontrivially will have to work around our abstraction anyway, or will bring their own "shader" class.

developer-guides/globe.md
```diff
@@ -0,0 +1,197 @@
...
+This is due to how polygons and lines are triangulated in MapLibre - the earcut algorithm
+creates as few triangles as possible, which can sometimes result in huge triangles, for example in the oceans.
+This behavior is desirable in mercator maps, but if we were to project the vertices of such large triangles to globe directly,
+we would not get curved horizons, lines, etc. The figure below demonstrates how globe would look without subdivision.
+Note the deformed oceans, and the USA-Canada border that is not properly curved.
+
+![](assets/no_subdivision.png)
+
+For this reason, before a tile is finished loading, its geometry (both polygons and lines) is further subdivided.
+This subdivision is handled in `subdividePolygon` and `subdivideVertexLine` functions in `src/render/subdivision.ts`
```
HarelM, line 63: Not sure the link to the code here is needed, mainly because this can move and no one will update the docs here if it does...

kubapelc: This makes sense, I'll remove all the code references.

src/render/draw_custom.ts
ibesora: My thinking was based on the way Maplibre does things on non-custom layers but I get your point: It really is difficult to draw the line between what we do vs what the user does.

developer-guides/globe.md
```diff
@@ -0,0 +1,197 @@
...
+
+To combat this, globe projection automatically switches to mercator projection around zoom level 12.
+This transition is smooth, animated and can only be noticed if you look very closely,
+because globe and mercator projections converge at high zoom levels, and around level 12
+they are already very close.
+
+The transition animation is implemented in the shader, in `interpolateProjection` function in `src/shaders/_projection_globe.vertex.glsl`.
+It is controlled by a "globeness" parameter, also stored in `globe_transform.ts`.
+
+## GPU atan() error correction
```
HarelM, line 107:
```suggestion
## GPU "atan" error correction
```

src/render/draw_custom.ts
kubapelc: I've removed the matrix completely, as it is replaced by args.defaultProjectionData

src/render/draw_custom.ts
kubapelc: I've pushed the API changes now.

src/geo/projection/mercator_transform.ts
kubapelc: Fixed by removing the function altogether.

developer-guides/globe.md
kubapelc: Fixed.

developer-guides/globe.md
kubapelc: Fixed.

kubapelc:
I've pushed API changes. The matrix that was passed to custom layers is now removed, and the object that is passed was simplified. There are no more functions inside it, instead custom layers should call MapLibre's internals directly, as is demonstrated in the examples. Shader uniforms and what used to be the main custom layer matrix are now passed in the form of ProjectionData, an object that is also used internally by MapLibre to store projection-related uniforms. I also exposed MapLibre's internal function for generating a subdivided tile mesh, as it is useful for rendering custom tiles.

I know about the failing tests, I'll fix them eventually.

src/render/draw_custom.ts
ibesora: I'd propose adding it in [src/style/style_layer/custom_style_layer.ts](https://github.com/maplibre/maplibre-gl-js/pull/4577/files#diff-1833f00564570182f4a622930fa9b67552d6ff37c93138e64a63b16ec45bae84)

src/style/style_layer/custom_style_layer.ts
```diff
@@ -33,19 +36,51 @@ type CustomRenderMethodInput = {
...
+     * A projection matrix is sufficient for simple custom layers that also only support mercator projection.
+     *
+     * When these uniforms are used, the shader's `projectTile` function projects spherical mercator
+     * coordinates to gl clip space coordinates. The spherical mercator coordinate `[0, 0]` represents the
+     * top left corner of the mercator world and `[1, 1]` represents the bottom right corner. When
+     * the `renderingMode` is `"3d"`, the z coordinate is conformal. A box with identical x, y, and z
+     * lengths in mercator units would be rendered as a cube. {@link MercatorCoordinate.fromLngLat}
+     * can be used to project a `LngLat` to a mercator coordinate.
+     */
+    defaultProjectionData: ProjectionData;
```
HarelM, line 76: This means that `ProjectionData` will become public. Is it well documented?

src/render/program/projection_program.ts
```diff
@@ -19,9 +19,35 @@ export const projectionUniforms = (context: Context, locations: UniformLocations
 });
 
 export type ProjectionData = {
+    /**
```
HarelM, line 22: Add a description of what this type is above please.

src/util/create_tile_mesh.ts
```diff
@@ -0,0 +1,176 @@
...
+}
+
+/**
+ * Creates a mesh of a quad that covers the entire tile (covering positions in range 0..EXTENT),
+ * is optionally subdivided into finer quads, optionally includes a border
+ * and optionally extends to the north and/or special pole vertices.
+ * Additionally the resulting mesh indices type can be specified using `forceIndicesSize`.
+ * @param options - Specify options for tile mesh creation such as granularity or border.
+ * @param forceIndicesSize - Specifies what indices type to use. The values '32bit' and '16bit' force their respective indices size. If undefined, the mesh may use either size, and will pick 16 bit indices if possible. If '16bit' is specified and the mesh exceeds 65536 vertices, an exception is thrown.
+ * @returns Typed arrays of the mesh vertices and indices.
```
HarelM, line 79: Add an example of how to use this method as it is a part of the public API.

src/render/program/projection_program.ts
```diff
@@ -19,9 +19,35 @@ export const projectionUniforms = (context: Context, locations: UniformLocations
...
+     * The main projection matrix. For mercator projection, it usually projects in-tile coordinates 0..EXTENT to screen,
+     * for globe projection, it projects a unit sphere planet to screen.
+     */
     'u_projection_matrix': mat4;
+    /**
+     * The extent of current tile in the mercator square.
+     * Used by globe projection.
+     * First two components are X and Y offset, last two are X and Y scale.
+     * Conversion from in-tile coordinates in range 0..EXTENT is done as follows:
+     * ```
```
HarelM, line 32: use `@example`

HarelM: Is it possible to change the variables names to be camelCase and not underscore? I'm guessing these are passed to the shaders, but I think this is the first time we have an API that is related to the shaders code...

src/style/style_layer/custom_style_layer.ts
```diff
@@ -33,19 +36,51 @@ type CustomRenderMethodInput = {
     * https://learnopengl.com/Getting-started/Coordinate-Systems
     */
     projectionMatrix: mat4;
+    /**
+     * Data required for picking and compiling a custom shader for the current projection.
+     */
+    shader: {
```
HarelM, line 42: `shaderData` maybe?
Since one will be writing shaders does this information will actually help? Or will the user need to reverse engineer the code here anyway?

src/util/create_tile_mesh.ts
```diff
@@ -0,0 +1,176 @@
...
+import {Mesh} from '../render/mesh';
+import {PosArray, TriangleIndexArray} from '../data/array_types.g';
+import {SegmentVector} from '../data/segment';
+import {NORTH_POLE_Y, SOUTH_POLE_Y} from '../render/subdivision';
+import {EXTENT} from '../data/extent';
+import posAttributes from '../data/pos_attributes';
+
+const EXTENT_STENCIL_BORDER = EXTENT / 128;
+
+export type CreateTileMeshOptions = {
```
HarelM, line 11: Same here for the description of the type itself

src/style/style_layer/custom_style_layer.ts
kubapelc: I've renamed it as you suggested and expanded the doc comments with an example of how to embed the code into a custom vertex shader (in addition to this being shown in the globe custom layer examples).

Making a simple vertex shader for a custom layer, such as in the custom tiles for globe example, should not require any reverse engineering or reading deeper into maplibre's source code. But doing anything more complex (maybe custom layer in the style of fill-extrusion) will probably require reading the shader code to see what is available and how are the functions used.

src/render/program/projection_program.ts
kubapelc: Done!

src/util/create_tile_mesh.ts
kubapelc: Done!

src/render/program/projection_program.ts
kubapelc: I've added the `@example`.

> Is it possible to change the variables names to be camelCase and not underscore?

The ProjectionData object was intended to only store uniform values (thus used the same naming convention as uniforms), but now that it is also exposed to custom layers, it makes sense to make it more standard. I've moved it into a separate file and converted names to camerCase.

src/render/draw_custom.ts
kubapelc: @ibesora I've expanded the docs in `custom_style_layer.ts` a bit in the last push. Do you think the current docs are sufficient? If not, what questions should the docs answer?

src/render/program/projection_program.ts
```diff
@@ -1,6 +1,8 @@
 import {Uniform1f, Uniform4f, UniformLocations, UniformMatrix4f} from '../uniform_binding';
 import {Context} from '../../gl/context';
-import {mat4} from 'gl-matrix';
+// This next import is needed for the \@link in the doccomment to work properly.
```
HarelM, line 3:
```suggestion
// This next import is needed for the "@link" in the documentation to work properly.
```

src/util/create_tile_mesh.ts
```diff
@@ -0,0 +1,206 @@
...
+ * @internal
+ * Creates a mesh of a quad that covers the entire tile (covering positions in range 0..EXTENT),
+ * is optionally subdivided into finer quads, optionally includes a border
+ * and optionally extends to the north and/or special pole vertices.
+ * @param context - The WebGL context wrapper.
+ * @param options - Specify options for tile mesh creation such as granularity or border.
+ * @returns The mesh vertices and indices, already allocated and uploaded into WebGL buffers.
+ */
+export function createTileMeshInternal(context: Context, options: CreateTileMeshOptions): Mesh {
+    const tileMesh = createTileMesh(options, '16bit');
```
HarelM, line 74: This is confusing, I would not expect an internal method to call the exported public method, but the other way around...
If this is a utility method to remove code residency maybe call in create16bitTileMesh?

test/examples/globe-3d-model.html
```diff
@@ -0,0 +1,150 @@
...
+            maxPitch: 80,
+            pitch: 70,
+            antialias: true, // create the gl context with MSAA antialiasing, so custom layers are antialiased
+        }));
+
+        map.on('style.load', () => {
+            map.setProjection({
+                type: 'globe', // Set projection to globe
+            });
+            map.setSky({
```
HarelM, line 62: This should be solved in main now, let me know if you want to me to update the globe branch.

kubapelc: Yes please. Is the sky disabled by default now?

HarelM: I've looked at the conflict between main and globe and the heatmap code changes are a bit complicated to solve.
Let me know if you get around to it before I do.

src/util/create_tile_mesh.ts
kubapelc: I've renamed it to `createTileMeshWithBuffers`, I think that is more descriptive as that function's main purpose is to also allocate WebGL buffers for the mesh.

test/examples/globe-3d-model.html
HarelM: I believe so, basically, I set the color to transparent in case the skys are not defined, not sure how performant that is though...

kubapelc: I believe there is a check for opacity===0 in the atmosphere code, so performance should be ok!

HarelM:
@ibesora let me know if you have other concerns before I merge this into the globe branch.

src/render/draw_custom.ts
ibesora: > what questions should the docs answer?

I think ideally people shouldn't need to go into the examples to know how to use things.

I found a little bit weird that the uniform names are in the `ProjectionData` definition in another file while the `shaderData` object is defined in place. IMO, the `shaderData` approach is better because it keeps all the documentation in a single place but we'd need to duplicate things. Maybe adding something in there saying where to look for them would be enough!

Thanks for going the extra mile!

ibesora: This looks great. Thanks!

test/examples/globe-3d-model.html
kubapelc: Both not settings the sky at all and calling `map.setSky(undefined);` results in the sky being present on globe, even with the changes from main.

HarelM: I've just opened a PR to fix another issue when setting the sky to `undefined` : 
- #4607

If setting the sky to undefined in globe doesn't help - then it means there's a bug in this branch, or a missing implementation - for example - adding the `'atmosphere-blend' = 0.0` to the sky.ts file when sky is undefined might solve this problem.

I would wait for the other fix to be merged to main and then to the globe branch before fixing this here to avoid merge conflicts.

HarelM: Can be done in a later PR.

HarelM:
Let's remember to update the changelog with the breaking change that was done as part of this PR before we merge this to main and create version 5. It is tracked as part of #3834 in general.

PR #4592, opened 2024-08-24
Fix: unable to remove listener with layer(s) registered with Map#once

pstaszek:
This PR fixes issue, that Map#off does not deregister an event registered with Map#once, when layer(s) are provided.

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


src/ui/map.ts
```diff
@@ -1496,6 +1498,18 @@ export class Map extends Camera {
 
         const delegatedListener = this._createDelegatedListener(type, layerIds, listener);
 
+        for (const key in delegatedListener.delegates) {
+            const delegate: Delegate = delegatedListener.delegates[key];
+            delegatedListener.delegates[key] = (...args: Parameters<Delegate>) => {
+                this._removeDelegatedListener(type, layerIds, listener);
```
pstaszek, line 1504: My concern is that in `this._removeDelegatedListener` the `Map#off` is called. On the other hand here we have reference to `delegate`, so it shouldn't harm. As a solution for that, we can add parameter, to skip calling `off` in `this._removeDelegatedListener`.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ### 🐞 Bug fixes
 
 - Heatmap Fix for 3D terrain ([#4571](https://github.com/maplibre/maplibre-gl-js/pull/4571))
+- Fix Map#off to not remove listener with layer(s) registered with Map#once
```
HarelM, line 10: Please add a link to this PR.

src/ui/map.ts
```diff
@@ -1496,6 +1498,18 @@ export class Map extends Camera {
...
+                delegate(...args);
+            };
+        }
+
+        this._delegatedListeners = this._delegatedListeners || {};
```
HarelM, line 1509: Can you explain why these 3 lines were added? Seems weird...

pstaszek: To be able to deregister a listener with layer(s) we need a reference to delegate - those lines saves the delegate.

How it works (in my understanding):
When the new listener is registered with layer(s), it is wrapped into another function, that filters the layers. It happens in `this._createDelegatedListener`. The reference is needed to this new function, so those delegates are saved in `this._delegatedListeners`. That was missing in `Map#once`.

CHANGELOG.md
pstaszek: Done.

src/ui/map.ts
HarelM: In this case, please more this part of the code to a private method that is used by both `on` and `once`.

src/ui/map.ts
```diff
@@ -343,9 +343,11 @@ export type CompleteMapOptions = Complete<MapOptions>;
 type DelegatedListener = {
     layers: string[];
     listener: Listener;
-    delegates: {[type in keyof MapEventType]?: (e: any) => void};
+    delegates: {[type in keyof MapEventType]?: Delegate};
```
HarelM, line 346: use `Record`?
Also maybe we can type better the `Delegate` itself so that it receives an event based on the `MapEventType` mapping?

src/ui/map.ts
pstaszek: Done.

src/ui/map.ts
pstaszek: `Record` doesn't work with `in keyof` structure. I tried to improve typings.

HarelM (approved):
Thanks!!

PR #4599, opened 2024-08-26
Symbol projection optimization

### LINKED ISSUE #4419: PR 4071 resulted ~100ms perf regression (time to first idle)
In an effort to upgrade from MapLibre 4.1.2 to 4.3.2, while measuring time to first idle we observed roughly 100ms (or 2%) perf regression at 90 percentile.
After extensive search and experiments, we narrowed it down to PR 4071: https://github.com/maplibre/maplibre-gl-js/pull/4071/
The above data were based on 3 million user hits on Bing maps over the period of 14 days.

**browser**: edge/chrome/ff/safari

Tile to first idle is affected by many factors such as the tiles, styles and individual CPU/GPU, that is probably why benchmark testing in MapLibre was not able to catch.  

We have eliminated at least two factors:
1) Shader compilation: the time it takes to compile shader did not change much (less than 10 ms). 
2) Shader run time performance measured in FPS (frames per second) is not affected.

This PR is quite sophisticated. Many changes we do not understand and need help. Specifically, while the goal is to fix bugs related to variable-anchors and collision boxes, can it maintain the same performance when the above two conditions do not apply?

@kubapelc would you please help review how many more ops were added per symbol, and give some suggestions regarding how to cut them down as much as possible?

Thanks a lot!

kubapelc:
Implements a simple fast path for the most common type of symbols into the collision index.

Should partially fix #4419 

Seems to be roughly ~10% faster in the symbol collision benchmark, where ~25% of symbols hit the fast path. No scientific measuring was done, I'm just eyeballing several runs of the benchmark.

Benchmark results

### DIFF
--- a/src/symbol/collision_index.ts
+++ b/src/symbol/collision_index.ts
@@ -48,6 +48,14 @@ export type FeatureKey = {
     overlapMode: OverlapMode;
 };
 
+type ProjectedBox = {
+    /**
+     * The AABB in the format [minX, minY, maxX, maxY].
+     */
+    box: [number, number, number, number];
+    allPointsOccluded: boolean;
+};
+
 /**
  * @internal
  * A collision index used to prevent symbols from overlapping. It keep tracks of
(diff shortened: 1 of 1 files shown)


HarelM (approved):
LGTM! Thanks!

HarelM:
CC: @zhangyiatmicrosoft

kubapelc:
I did some more changes which result in much higher perf improvements:

benchmark

But I'm not really optimizing, rather reverting things to their original state. I noticed `projectAndGetPerspectiveRatio` being significant in the profiler and reverted it to its original state, where the `project` function was manually inlined, and that seems to help performance a lot.

I hope that this `projectAndGetPerspectiveRatio` is the "other 50%" of the perf degradation seen in the linked issue, which would mean the rest of the sprawling symbol changes are mostly ok.

HarelM:
Doesn't rollup inline methods that are used once? 
In any case, if this is the cause please add a comment about this being a hot path and should be changed with caution.

kubapelc:
I've already added a comment:

```
// The code here is duplicated from "projection.ts" for performance.
// Code here is subject to change once globe is merged.
```

If rollup only inlines methods that are used once, then that doesn't apply to `project`, as it is used in many places.

Do you have any pointers on documentation of what code optimizations rollup does actually do? I'm having a lot of trouble finding anything related to function inlining, other than some issues on github.

HarelM:
I couldn't find anything when I looked besided closed issues, but you can simply look at how the maplibre-gl-js-dev.js file content is for the method you are looking for and see if it was inlined on not...

src/symbol/projection.ts
```diff
@@ -137,13 +137,13 @@ function getGlCoordMatrix(posMatrix: mat4,
     }
 }
 
-function project(point: Point, matrix: mat4, getElevation?: (x: number, y: number) => number): PointProjection {
+function project(x: number, y: number, matrix: mat4, getElevation?: (x: number, y: number) => number): PointProjection {
```
HarelM, line 140: This might expalin the performace improvements - creating a new point as done in some of the cases below might have cuased performance degredation.

HarelM:
Can you see if only changing the `project` method to receive x,y instead of point does the same performance improvements? I wonder if the creation of a point caused the performance degradation.

HarelM:
Interesting...

HarelM:
In any case, ping me when this is ready for review.

kubapelc:
@HarelM I guess I won't be changing this further. I can always make another PR later...

HarelM:
Can you add a changelog item?
I'll merge this afterwards.

kubapelc:
I've added the changelog entry. Also I'm reimplementing these optimizations for globe, since the symbol code there is now too different for a merge (would have to hand-edit the code a lot anyway).

HarelM (approved):

PR #4606, opened 2024-08-27
geojson clusterArrayType option

andrewharvey:
At high clusterMaxZoom values and low clusterRadius values with data very close together, the default supercluster Float32Array type starts causing some features to cluster when they shouldn't based on your set clusterRadius.

To resolve this I've exposed an option on cluster sources to set the supercluster type to Float64Array.

I've posted this PR as a WIP for now as I don't think we should actually expose this option like this. I'm thinking we should try to determine based on the clusterMaxZoom and clusterRadius set that (we we just assume there will be very close data present) if Float32Array would suffice or if we need Float64Array and set it transparently.

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "@mapbox/unitbezier": "^0.0.1",
     "@mapbox/vector-tile": "^1.3.1",
     "@mapbox/whoots-js": "^3.1.0",
-    "@maplibre/maplibre-gl-style-spec": "^20.3.1",
+    "@maplibre/maplibre-gl-style-spec": "andrewharvey/maplibre-gl-style-spec#cluster-array-type",
     "@types/geojson": "^7946.0.14",
     "@types/geojson-vt": "3.2.5",
     "@types/mapbox__point-geometry": "^0.1.4",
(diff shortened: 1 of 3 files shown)


HarelM:
supercluster package is currently not maintained by us.
I would suggest trying to open a PR there as a starting point to see that this is accepted and there's a new version.
Regarding how to solve this here in terms of API is a later concern from my point of view.
Maybe in the supercluster there will be a way to solve this internally instead of exposing the array type and there won't be a need to solve it here...?

andrewharvey:
I have https://github.com/mapbox/supercluster/pull/249.

I did this as a quick solution to a problem I faced, but I agree it needs more thought as to if this should be exposed like this, or if supercluster should internally decide when it would run into precision issues with Float32Array and then automatically enable Float64Array.

I feel the latter is better to avoid exposing this complexity to the user, and in that case this PR is void.

HarelM:
Let's see if there's a response in the supercluster in the next week or two.

andrewharvey:
I'll close this one, as this is likely not the best solution for the issue.

PR #4615, opened 2024-08-27
Globe - Covering tiles

kubapelc:
This PR implements proper tile visibility and LOD for globe. However I didn't do the refactor of moving `coveringTiles` into source caches. I had an early implementation of this for the school project, I developed it further after talking to @Pheonor .

CC @Pheonor, please review this PR.

The globe `covertingTiles` uses the same approach as the old mercator code, it also computes tile AABBs to test against the camera frustum. The key observation is that for any tile other than zoom levels 0 (the entire sphere) and 1 (sphere quaters) we can compute the AABB by simply taking the min and max of the 4 tile corners.


src/util/primitives.ts
```diff
@@ -136,7 +138,40 @@ class Aabb {
 
         return 1;
     }
+
+    /**
+     * Performs a halfspace-aabb intersection test. Returns 0 if there's no intersection,
+     * 1 if shapes are intersecting and 2 if the aabb if fully inside the plane's positive halfspace.
```
HarelM, line 144: 0,1,2 is not a great API, is there a good reason to use it instead of an enum or string options?

src/geo/projection/globe_transform.ts
```diff
@@ -680,9 +681,256 @@ export class GlobeTransform implements ITransform {
...
+            return fullyVisible;
+        }
+
+        return partiallyVisible;
+    }
+
     coveringTiles(options: CoveringTilesOptions): OverscaledTileID[] {
-        // Globe: TODO: implement for globe #3887
-        return this._mercatorTransform.coveringTiles(options);
+        if (!this._globeRendering) {
```
HarelM, line 819: Is there a way to split this method a bit further to make it a bit smaller?

HarelM: Also can you add some comments on what this method is doing, how and what is the return value?

HarelM:
I had a thought: in order to move the tile fetching logic a bit further but not have a dramatic change can we start by moving it to a separate file?
No big refactoring of the logic, no big changes, just place this logic in a file that later on will be used to identify the scope of the change that I'm aiming for.
We can keep the trasform call to these methods at this point.

src/geo/projection/globe_transform.ts
```diff
@@ -680,9 +681,256 @@ export class GlobeTransform implements ITransform {
         return [new UnwrappedTileID(0, tileID)];
     }
 
+    /**
+     * Returns the AABB of the specified tile. The AABB is in the coordinate space where the globe is a unit sphere.
+     * @param tileID - Tile x, y and z for zoom.
+     */
+    private getTileAABB(tileID: {x: number; y: number; z: number}): Aabb {
```
Pheonor, line 688: Are you sure that only the four corners are sufficient to define the AABB of any tile ?
During my first implementation of this for a globe, I need to take into account tile round shape and extend the AABB to ensure tiles into the globe border are wells included.

src/geo/projection/globe_transform.ts
```diff
@@ -680,9 +681,256 @@ export class GlobeTransform implements ITransform {
...
+        mat4.scale(matrix, matrix, [1, 1, -1]);
+        const cameraFrustum = Frustum.fromInvProjectionMatrix(matrix);
+
+        const cameraCoord = this.screenPointToMercatorCoordinate(this.getCameraPoint());
+        const centerCoord = MercatorCoordinate.fromLngLat(this.center);
+        const numTiles = Math.pow(2, z);
+        const cameraPoint = [numTiles * cameraCoord.x, numTiles * cameraCoord.y, 0];
+        const centerPoint = [numTiles * centerCoord.x, numTiles * centerCoord.y, 0];
+
+        const radiusOfMaxLvlLodInTiles = 3;
```
Pheonor, line 843: How do you determine the value of 3 ?

Pheonor:
I check how the tile loading work, and it seems very nice.

Sometimes, a tile take some time to load and an empty tile appear, but its not systematic.

HarelM:
I would also advise to add some unit test to the added functions related to covering tiles, as there's some math there that would benefit from unit tests.
Overall look good, thanks!

src/util/primitives.ts
kubapelc: Converted to enum. Only reason was that the original code already did this, I see no downsides of using an enum.

src/geo/projection/globe_transform.ts
kubapelc: It is also 3 in the original code (when terrain not enabled):

```
const radiusOfMaxLvlLodInTiles = options.terrain ? 2 / Math.min(this.tileSize, options.tileSize) * this.tileSize : 3;
```

I've also checked that tile LOD behaves mostly the same when looking at a heavily pitched map under both mercator and globe projections.

src/geo/projection/globe_transform.ts
kubapelc: I can get away with this simple AABB computation, because I do all math in world space, with axes aligned to poles (X is from planet center towards null island, Y is towards north pole, Z is toward 0N90E). Because of this the X,Y, and Z coordinates monotonically increase or decrease on the surface and along the edges of most tiles, never having an extreme inside the tile itself, always at the tile edge. The only tiles where this doesn't apply are at zoom level 0 and 1 and are handled separately.



Note that since every coordinate *only* increases or *only* decreases along any edge, the minimum and maximum of each must lie in one of the 4 corners.

-----

I've also added a more rigorous explanation in a comment to the code:

We can get away with only checking the 4 tile corners for AABB construction, because for any tile of zoom level 2 or higher it holds that the extremes (minimal or maximal value) of X, Y or Z coordinates must lie in one of the tile corners.

To see why this holds, consider the formula for computing X,Y and Z from angular coordinates. It goes something like this:

X = sin(lng) * cos(lat)
Y = sin(lat)
Z = cos(lng) * cos(lat)

Note that a tile always covers a continuous range of lng and lat values, and that tiles that border the mercator north/south edge are assumed to extend all the way to the poles.

We will consider each coordinate separately and show that an extreme must always lie in a tile corner for every axis, and must not lie inside the tile.

For Y, it is clear that the only way for an extreme to not lie on an edge of the lat range is for the range to contain lat=90° or lat=-90° without either being the tile edge. This cannot happen for any tile, these latitudes will always:
- either lie outside the tile entirely, thus Y will be monotonically increasing or decreasing across the entire tile, thus the extreme must lie at a corner/edge
- or be the tile edge itself, thus the extreme will lie at the tile edge

For X, considering only longitude, the tile would also have to contain lng=90° or lng=-90° (with neither being the tile edge) for the extreme to not lie on a tile edge. This can only happen at zoom levels 0 and 1, which are handled separately. But X is also scaled by cos(lat)! However, this can only cause an extreme to lie inside the tile if the tile crosses lat=0°, which cannot happen for zoom levels other than 0.

For Z, similarly to X, the extremes must lie at lng=0° or lng=180°, but for zoom levels other than 0 these cannot lie inside the tile. Scaling by cos(lat) has the same effect as with the X axis.

So checking the 4 tile corners only fails for tiles with zoom level <2, and these are handled separately with hardcoded AABBs:
- zoom level 0 tile is the entire sphere
- zoom level 1 tiles are "quarters of a sphere"

src/geo/projection/globe_transform.ts
kubapelc: I've split it up a bit and added the comment.

kubapelc:
I've moved both coveringTiles methods (mercator+globe) into separate files, and added some unit tests for the globe variant.

src/util/primitives.ts
```diff
@@ -1,10 +1,16 @@
 import {mat4, vec3, vec4} from 'gl-matrix';
 
-class Frustum {
+export enum IntersectionResult {
```
HarelM, line 3: I think `const enum` is better as it is replaced at compile time.

src/geo/transform_interface.ts
```diff
@@ -496,3 +493,15 @@ export interface IReadonlyTransform extends ITransformGetters {
  * A transform is cloneable, which is used when a given map state must be retained for multiple frames, mostly during symbol placement.
  */
 export interface ITransform extends IReadonlyTransform, ITransformMutators {}
+
+export interface ITileVisibilityProvider {
```
HarelM, line 497: Feels like `coveringTiles` should be part of this interface as well?

src/geo/projection/mercator_covering_tiles.ts
```diff
@@ -0,0 +1,136 @@
...
+            zoom: 0,
+            x: 0,
+            y: 0,
+            wrap,
+            fullyVisible: false
+        };
+    };
+
+    // Do a depth-first traversal to find visible tiles and proper levels of detail
+    const stack: Array<{
```
HarelM, line 50: Define these inline types outside the method to improve readability and make the method smaller.

src/geo/projection/mercator_covering_tiles.ts
```diff
@@ -0,0 +1,136 @@
...
+        aabb: Aabb;
+        zoom: number;
+        x: number;
+        y: number;
+        wrap: number;
+        fullyVisible: boolean;
+    }> = [];
+    const result: Array<{
```
HarelM, line 58: This type as well

src/geo/projection/globe_transform.ts
```diff
@@ -680,9 +687,124 @@ export class GlobeTransform implements ITransform {
...
+    }
+
+    /**
+     * A simple/heuristic function that returns whether the tile is visible under the current transform.
+     * @param x - Tile X.
+     * @param y - tile Y.
+     * @param z - Tile zoom.
+     * @returns 0 is not visible, 1 if partially visible, 2 if fully visible.
+     */
+    isTileVisible(x: number, y: number, z: number): IntersectionResult {
```
HarelM, line 784: Consider moving this method to the covering tiles file, and have it accept the plane, frustrum and execute the getTileAABB inside.

src/geo/projection/globe_transform.ts
```diff
@@ -680,9 +687,124 @@ export class GlobeTransform implements ITransform {
...
+    }
+
     coveringTiles(options: CoveringTilesOptions): OverscaledTileID[] {
-        // Globe: TODO: implement for globe #3887
-        return this._mercatorTransform.coveringTiles(options);
+        if (!this._globeRendering) {
+            return this._mercatorTransform.coveringTiles(options);
+        }
+
+        return globeCoveringTiles(this, options);
```
HarelM, line 807: Consider a way to past only the relevant required fields for this method instead of passing "this".

src/geo/projection/globe_transform.test.ts
```diff
@@ -598,4 +599,201 @@ describe('GlobeTransform', () => {
             expect(transform.isLocationOccluded(new LngLat(84.50, 0))).toBe(true);
         });
     });
+
+    describe('aabb', () => {
```
HarelM, line 603: Might be move to a different file if the method is moved as well...

src/geo/projection/globe_covering_tiles.ts
```diff
@@ -0,0 +1,181 @@
...
+    // z=6             |--||--||--|
+    //                       ^map center
+    // ...where "|--|" symbolizes a tile viewed sideways.
+    // This logic might be slightly different from what mercator_transform.ts does, but should result in very similar (if not the same) set of tiles being loaded.
+    const centerDist = distanceToTile(centerCoord.x, centerCoord.y, tileX, tileY, tileSize);
+    const cameraDist = distanceToTile(cameraCoord.x, cameraCoord.y, tileX, tileY, tileSize);
+    return Math.min(centerDist, cameraDist) * 2 <= radiusOfMaxLvlLodInTiles; // Multiply distance by 2, because the subdivided tiles would be half the size
+}
+
+function getWrap(centerCoord: MercatorCoordinate, tileX: number, tileSize: number): number {
```
HarelM, line 72: Please add a comment about the return value as it is not clear when reading this.

src/geo/projection/globe_transform.ts
HarelM: I've looked at the relevant methods, this doesn't look possible...

src/geo/projection/globe_covering_tiles.ts
```diff
@@ -0,0 +1,181 @@
...
+}
+
+/**
+ * Returns a list of tiles that optimally covers the screen. Adapted for globe projection.
+ * Correctly handles LOD when moving over the antimeridian.
+ * @param transform - The globe transform instance.
+ * @param options - Additional coveringTiles options.
+ * @returns A list of tile coordinates, ordered by ascending distance from camera.
+ */
+export function globeCoveringTiles(transform: IReadonlyTransform & ITileVisibilityProvider, options: CoveringTilesOptions): OverscaledTileID[] {
```
HarelM, line 93: This looks good.
General question: What will happen if the user sets terrain with globe? Will if fail? seen only when zooming in to mercator area? Do nothing?

HarelM:
I've added a few last comments, nothing big hopefully.

src/util/primitives.ts
kubapelc: Done!

src/geo/transform_interface.ts
kubapelc: I used this interface so that I can avoid a circular dependency, because GlobeTransform must use globeCoveringTiles but globeCoveringTiles must call a function on GlobeTransform. I've now found a better way of dealing with the dependency (importing GlobeTransform as a type) and will use that instead.

src/geo/projection/mercator_covering_tiles.ts
kubapelc: Fixed.

src/geo/projection/mercator_covering_tiles.ts
kubapelc: Fixed.

src/geo/projection/mercator_covering_tiles.ts
kubapelc: I've decided to define the types in both mercator_covering_tiles.ts and globe_covering_tiles.ts instead of sharing them, since they are mostly an implemetation detail of the respective functions (and the stack types are slightly different).

src/geo/projection/globe_transform.ts
kubapelc: Done!

src/geo/projection/globe_transform.ts
kubapelc: Done!

src/geo/projection/globe_transform.test.ts
kubapelc: Yes, I've created a test file for globe covering tiles.

src/geo/projection/globe_covering_tiles.ts
kubapelc: Done!

src/geo/projection/globe_covering_tiles.ts
kubapelc: It should only see the mercator area for now.

HarelM (approved):
LGTM! thanks!!

src/geo/projection/globe_covering_tiles.ts
```diff
@@ -5,6 +5,19 @@ import {OverscaledTileID} from '../../source/tile_id';
 import {MercatorCoordinate} from '../mercator_coordinate';
 import type {GlobeTransform} from './globe_transform';
 
+type CoveringTilesResult = {
+    tileID: OverscaledTileID;
+    distanceSq: number;
+    tileDistanceToCamera: number;
+};
+
+type CoveringTilesStack = {
```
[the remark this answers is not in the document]
kubapelc: Good point, fixed!

HarelM:
Feel free to merge this to the globe branch as it is now in draft state.
I don't have more code review comments.

kubapelc:
@HarelM Thank you for the review and all the feedback! (And for all the feedback on the other PRs as well!)

HarelM:
Sure, thank you for taking the feedback in a professional way.

PR #4620, opened 2024-08-28
fix incorrect bounds with fitBounds when crossing antimeridian

YoelRidgway:
If you add bounds that cross the [180th meridian (antimeridian)](https://en.wikipedia.org/wiki/180th_meridian), it causes the view to zoom out so that the northeast point of the bounds is to the left and the southeast point of the bounds is to the right. See below for two examples using map.fitBounds() for the islands of Fiji - current and expected:

Current (not correct): https://output.jsbin.com/yitecurala

Expected (my fix): https://output.jsbin.com/janaferezi

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


src/ui/camera.ts
```diff
@@ -626,11 +626,38 @@ export abstract class Camera extends Evented {
...
+     * @param bounds - The LngLatBounds to adjust
+     * @returns The adjusted LngLatBounds
+     * @example
+     * ```ts
+     * let bounds = new LngLatBounds([175.813127, -20.157768], [-178. 340903, -15.449124]);
+     * let adjustedBounds = this._adjustAntiMeridian(bounds);
+     * // adjustedBounds will be: [[175.813127, -20.157768], [181.659097, -15.449124]]
+     * ```
+     */
+    _adjustAntiMeridian(bounds: LngLatBounds): LngLatBounds {
```
HarelM, line 647: Shouldn't this be a method of latlngbounds like "wrap" exists for lnglat?

HarelM:

src/ui/camera.ts
HarelM: Yes, making a test for camera for bounds would be correct, although not super critical if you cover the new method well.
I would also advise to see what kind of "special" bounds you may have get and make sure all those are handled properly, not just the fiji island, but all kind of special cases - only one point is crossing the line, what happens if you do want to fit the bounds of a large portion of the world.
All this cases needed to be tested to make sure this solution is robust.

YoelRidgway: Ok. I will look into some better tests and move the method to LngLatBounds. I'll make the tests on the new method so as not to have any duplication slowing down the CI. That said, I suppose it would be good to have at least one test on the api itself to ensure degradation doesn't happen after any future refactoring?

HarelM: Sounds good.

YoelRidgway:
I've come across an interesting case... What should happen if the sw and ne points are flipped? In the current version, `fitBounds()` will act as if nothing happened. However, with `adjustAntiMeridian()`, it puts the ne point to the east and the sw point to the west. Is this undesirable? I feel like `LngLatBounds` is supposed to take sw and ne in order so its normal that it assumes they have been input in correctly, otherwise the bounds could just be defined as any two points, e.g `p0` and `p1`.

src/ui/camera.test.ts
```diff
@@ -2100,15 +2100,132 @@ describe('#fitBounds', () => {
...
-        const bb = [[175.813127, -20.157768], [-178.340903, -15.449124]]; // bounds for fiji (crossing antimeridian)
+        const bb = [[175.813127, -20.157768], [-178.340903, -15.449124]];
         camera.fitBounds(bb, {duration: 0});
 
-        expect(fixedLngLat(camera.getCenter(), 4)).toEqual({lng: 178.7361, lat: -17.8190});
+        expect(fixedLngLat(camera.getCenter(), 4)).toEqual({lng: 178.7361, lat: -17.819});
         expect(fixedNum(camera.getZoom(), 3)).toBe(5.944);
     });
 
+    test('kenya', () => {
```
HarelM, line 2112: I think it would be better to have these tests in the other file, unless the logic in the camera is important.

YoelRidgway: Hmm yeah. I suppose... It just felt weird testing the `adjustAntiMeridian()` function directly because all it is doing is adding 360 to the longitude. The main problem in the first place was the way `fitBounds()` was acting.

HarelM:
That's exactly what I aimed for in terms of writing tests 😃. 
You can have two points on the globe and you need to decide if you always take the "shotest" arc, or you use the arc based on the "direction" (vector) between the points.
I think the shortest is problematic in terms of cases it supports (it can't show a wide map), so the only possible solution (without adding a new flag) is to assume order and have that define the "direction".
What do you think?

src/ui/camera.test.ts
HarelM: True, but then it means the tests are basically going through the same flow and you test the same thing ac few times...

YoelRidgway:
Yes I totally agree. I think this is the most logical way of going about things. I just hope there isn't someone who has been using `fitBounds()` the wrong way all this time only to have everything messed up after the change!

src/ui/camera.test.ts
YoelRidgway: Ok. I understand. I don't really know what tests to write on `adjustAntiMeridian()` because the output isn't really logical. I suppose I'll just calculate the output of `adjustAntiMeridian()` in these cases and copy over the same tests.

YoelRidgway: In my opinion I think the logic in the camera is important whether thats in `cameraForBounds` or `fitBounds` because in case there is any refactoring and adjustAntiMeridian gets deleted somehow, the final output of the api will still be tested correctly, but you have a lot more experience than me so give the final verdict and I'll go ahead with whichever!

HarelM: Keep just 1-2 tests that describe the bug you are fixing as part the tests of the API and move the rest to be class tests.

HarelM:
It's probably important to document this as part of the method documentation, just to be on the safe side.

HarelM:
Version 5 is the next breaking change version, I believe I'll start looking at it in the next month or so, we can consider adding this change only as part of this breaking change version. 
But I don't know if this is really a behavior change and not simply a bug fix...

YoelRidgway:
Even a bug fix breaks someones workflow! ;) https://xkcd.com/1172/

HarelM:
Thanks! Can you add a changelog entry please? 
Also worth mentioning something in fit bounds docs (public API).

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ using `transformCameraUpdate` caused the `maxBounds` to stop working just for ea
 ### ✨ Features and improvements
 - Support multiple layers in `map.on`, `map.once` and `map.off` methods ([#4279](https://github.com/maplibre/maplibre-gl-js/pull/4401))
 - Ensure GeoJSON cluster sources emit a console warning if `maxzoom` is less than or equal to `clusterMaxZoom` since in this case you may see unexpected results. ([#4604](https://github.com/maplibre/maplibre-gl-js/pull/4604))
+- Fix bug where `fitBounds` and `cameraForBounds` would not display accross the 180th meridian (antimeridian)
```
HarelM, line 9: Please move this under bug fixes.

YoelRidgway: Oh whoops. Sorry about that.

HarelM (approved):

PR #4625, opened 2024-08-29
Fix missing `lngRange` when cloning Transform instances

tomhicks:
## What this does

Fixes an issue noticeable when using `transformCameraUpdate` and `maxBounds` together. Due to a bug with cloning `Transform` instances, the longitude range on the transform was lost. As `transformCameraUpdate` receives a clone of the transform object, the constraining that happens afterwards misses the longitude constraint.


HarelM (approved):
Thanks!!

Issue #4158, opened 2024-05-23
White flickering of the map when resizing the window


Kurioni:
Hi all!
Recently I encountered a problem when using a library in a project. When I changed the size of the window (respectively, if the size of the `<canvas>` changed), I noticed that the map began to flicker white. I started researching this problem, I noticed that this problem has existed for a long time, but it turns out it still exists. I opened the examples on your site and noticed that this problem was also reproduced there.

[screencast-maplibre.org-2024.05.webm](https://github.com/maplibre/maplibre-gl-js/assets/113048205/a638c863-805e-450b-90bb-a24d2ef41a76)

**maplibre-gl-js version**: 4.3.2

**browser**: Chrome 124.0.6367.210

### Steps to Trigger Behavior

 1. Open any example on the site, for example: https://maplibre.org/maplibre-gl-js/docs/examples/fullscreen/
 2. Slowly resize window

### Expected Behavior
When resizing the window, the map should not flicker.

HarelM:
Yup, this has been a pain point from the dawn of history.
In most cases, window resizes are sparse, so the flickering is minimal.
Feel free to dive into it!

PR #4429, opened 2024-07-18
Fix flicker on map resize

daylightwarbler:
Fixes #2971 and #4158. Flicker on resize was introduced in #2157, released in v3.0.0, when map resizing was switched to using resizeObserver from using window resize events.

The fix is to immediately render the map upon resizing. The lead was this [thread](https://github.com/pixijs/pixijs/issues/3395#issuecomment-328495407).

Caveat: I'm not sure what to pass for the `paintStartTimeStamp` for [`map._render()`](https://github.com/maplibre/maplibre-gl-js/blob/main/src/ui/map.ts#L3021). I set it to `Date.now()`. `0` also works fine.

Before and after gifs (half speed):

before
after

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.
 - [x] Link to related issues.
 - [x] Include before/after visuals or gifs if this PR includes visual changes.
 - [ ] Post benchmark scores.
 - [x] Add an entry to `CHANGELOG.md` under the `## main` section.

PR #4632, opened 2024-08-31
fix flickering on map resize

pstaszek:
This PR fixes issue #4158.

When screen size changes, we immediately resize the canvas. Resizing canvas causes canvas to clear. Then, next render is scheduled using `_update` method. Because it is scheduled, and it happens asynchronously, in the time between the canvas is white.

There are two ways of solving the problem. Simple - in this PR, is to trigger render every time when canvas changes.
Other option would be to resize the canvas asynchronosly as well. I didn't try this approach, as it seems more complex, and I'm not sure if it is worth.

The downside of current approach is that the render might be called multiple time, which might impact performance. Although I'm not sure if this is an issue.

In this PR I replaced `_update` with `redraw` which happens immediately. Also `_update` also updates style and sources, which doesn't seem for me like something that should happen during resizing. Can you see any reason to do that?

In this PR https://github.com/maplibre/maplibre-gl-js/pull/4429 `_render` method was added, which is not ideal, because already frame was scheduled, and also triggering `_render` directly is not clearing scheduled frames, like `redraw`.

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ using `transformCameraUpdate` caused the `maxBounds` to stop working just for ea
 - Improve types a bit for `addSource` and `getSource` ([#4616](https://github.com/maplibre/maplibre-gl-js/pull/4616))
 - Fix the color near the horizon when terrain is enabled without any sky ([#4607](https://github.com/maplibre/maplibre-gl-js/pull/4607))
 - Fix bug where `fitBounds` and `cameraForBounds` would not display accross the 180th meridian (antimeridian)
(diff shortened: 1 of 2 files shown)


HarelM:
Looks good, thanks for the info! 
Want chance you can add a test to this somehow?

pstaszek:
@HarelM
Yes! I updated unit test, and checked if `redraw` and `_render` are called after `resize`. I think this is pretty good check, as it fails if `_render` was called asynchronously (for example if I change for `_update`, spy on `_render` was not called).

I also added comment on unit test.

HarelM (approved):
THANKS!

PR #4675, opened 2024-09-07
Swap 0.0.0.0 with localhost on dev server

birkskyum:
This makes the dev server examples load properly - they tile servers are open to calls from exactly `localhost` , not 0.0.0.0  or similar alternatives.

Before:
Screenshot 2024-09-07 at 21 02 29

After:

Screenshot 2024-09-07 at 21 02 07

### DIFF
--- a/package.json
+++ b/package.json
@@ -157,7 +157,7 @@
     "watch-css": "postcss --watch -o dist/maplibre-gl.css src/css/maplibre-gl.css",
     "build-benchmarks": "npm run build-dev && rollup --configPlugin @rollup/plugin-typescript -c test/bench/rollup_config_benchmarks.ts",
     "watch-benchmarks": "rollup --configPlugin @rollup/plugin-typescript -c test/bench/rollup_config_benchmarks.ts --watch",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -157,7 +157,7 @@
     "watch-css": "postcss --watch -o dist/maplibre-gl.css src/css/maplibre-gl.css",
     "build-benchmarks": "npm run build-dev && rollup --configPlugin @rollup/plugin-typescript -c test/bench/rollup_config_benchmarks.ts",
     "watch-benchmarks": "rollup --configPlugin @rollup/plugin-typescript -c test/bench/rollup_config_benchmarks.ts --watch",
-    "start-server": "st --no-cache -H 0.0.0.0 --port 9966 .",
+    "start-server": "st --no-cache -H localhost --port 9966 .",
```
HarelM, line 160: Is this needed to be written explicitly? Isn't localhost the default?

HarelM (approved):
Approved, see my nit picking comment.

package.json
birkskyum: No, it seems `http://[::]:9966/` is the default, and it also doesn't work.

Issue #977, opened 2022-02-10
Tree shaking


birkskyum:
As I understand, tree shaking is not properly supported currently. We can make this a tracking issue and get an overview for what needs to happen in order to get it working.

chanmathew:
hi @HarelM - just wondering if there has been any updates on this or if this is still currently blocked?

birkskyum:
@chanmathew , I believe the latest update in this regards is that there not too long ago was a large refactor towards named imports, allowing unused pieces to be left out. 

We'd be able to improve our support for tree shaking by distributing an ESM bundle, but that requires all the dependencies to be updated from CJS to ESM too. There is a blocker here, related to updating the dependencies, because this one has some issues, and resolution is on hold due to a parental leave:
- https://github.com/maplibre/maplibre-gl-js/pull/4393

Another piece of the puzzle is tooling. Our test harness uses Jest, which only has [experimental ESM support](https://jestjs.io/docs/ecmascript-modules). Last time I tried running it with ESM, there were some issues for our bundle tests. I'm hoping that the upcoming [Jest 30](https://github.com/jestjs/jest/milestone/15) will improve on the situation, but should it persist when the rest of the system is in place, is there preemptively prepared a PR here than can swap Jest for the ESM-ready [Vitest](https://vitest.dev/):
- #4728

HarelM:
The main problem from my point of view is how to share code between the workers and the main thread.
This is currently relaying on how this is build using commonJs and I don't know how to do it with ESM...
So, if the purpose of ESM is tree shaking, but you get the bundle size duplicated it totally beats the purpose...

knd775:
> So, if the purpose of ESM is tree shaking, but you get the bundle size duplicated it totally beats the purpose...

It increases the package size, but the actual js bundled in an application wouldn't increase with any half-decent bundler

jasongitmail:
@birkskyum Vitest is amazing. It's much faster than Jest and is a good move on its own merit. I switched from Jest to Vitest years ago and never looked back, and it's only going to grow in usage with Evan You now creating [VoidZero](https://voidzero.dev/), which Vitest is part of.

birkskyum:
[...]

PR #4684, opened 2024-09-09
Use platform consistently instead of jest done

birkskyum:
When Jest was created some 12-13 years ago, there weren't promises nor await/async around, and therefore they added the custom done() api instead. Newer tooling [just use the platform](https://vitest.dev/guide/migration.html#done-callback), instead of having these two test modes (1. done/done('error'), 2. async/await/Promise/Error).

I've here updated our unit test suite to be consistent in using the platform mode, instead of jest.done(), and focused on keeping a minimal line diff to avoid merge conflicts (e.g. no change in indentation or order of logic anywhere). 

Related to (see https://github.com/maplibre/maplibre-gl-js/pull/4684#issuecomment-2340716107):
- #1595 
- #977 

### DIFF
--- a/src/util/throttle.test.ts
+++ b/src/util/throttle.test.ts
@@ -8,19 +8,17 @@ describe('throttle', () => {
         expect(executionCount).toBe(0);
     });
 
-    test('executes unthrottled function once per tick when period is 0', () => new Promise<void>(done => {
+    test('executes unthrottled function once per tick when period is 0', async () => {
         let executionCount = 0;
         const throttledFunction = throttle(() => { executionCount++; }, 0);
         throttledFunction();
         throttledFunction();
         expect(executionCount).toBe(1);
-        setTimeout(() => {
-            throttledFunction();
-            throttledFunction();
-            expect(executionCount).toBe(2);
-            done();
-        }, 0);
-    }));
+        await new Promise(resolve => setTimeout(resolve, 0));
+        throttledFunction();
+        throttledFunction();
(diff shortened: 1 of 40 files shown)


HarelM:
I've made a effort to remove places that use done as much as possible. 
Mainly using the `once` method. 
This is a work in progress. 
In most tests, I think done is not needed. 
It might be needed in rare cases.
Do you think you could push some changes in that direction?

birkskyum:
Yes, I managed to remove ~20% of the done's just now - there we some sync function that didn't need them, and others that could be flattened with async/await . The remaining are all triggerede from inside callbacks, so a promise appear to be needed in those cases.

src/util/throttle.test.ts
```diff
@@ -31,15 +29,13 @@ describe('throttle', () => {
...
         let executionCount = 0;
         const throttledFunction = throttle(() => { executionCount++; }, 5);
         throttledFunction();
         throttledFunction();
         throttledFunction();
-        setTimeout(() => {
-            expect(executionCount).toBe(2);
-            done();
-        }, 10);
+        await new Promise(resolve => setTimeout(resolve, 10));
```
HarelM, line 38: There's a "sleep" utility method I wrote, I believe.

HarelM (approved):
Overall, approved! Thanks, just make sure to use "sleep" where needed.

HarelM:
@birkskyum any updates on the small request above?

birkskyum:
Yes, I've used the sleep util now.

PR #4706, opened 2024-09-15
Add sync-move example

birkskyum:
Add example of sync-move plugin, for two reasons. First, it's clarifying the plugin isn't limited to only 2 maps. And second, it's called mapbox, and this makes it easy for users to validate it does indeed work for maplibre too.

Screenshot 2024-09-15 at 18 08 21


test/examples/sync-move.html
```diff
@@ -0,0 +1,55 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>Sync movement of multiple maps</title>
+    <meta property="og:description" content="Synchronize MapLibre GL JS maps with the sync-move plugin." />
+    <meta charset='utf-8'>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <link rel='stylesheet' href='../../dist/maplibre-gl.css' />
+    <script src='../../dist/maplibre-gl-dev.js'></script>
+    <script src="https://unpkg.com/@mapbox/mapbox-gl-sync-move"></script>
```
HarelM, line 10: Does this needs a version? It might break at some point, wouldn't it?

HarelM:
As the CI states, the example image is needed, which is currently missing. 
Thanks for taking the time to add this example!

test/examples/sync-move.html
birkskyum: yes

HarelM (approved):

Issue #4669, opened 2024-09-06
Future: Move from Rollup to Rolldown


birkskyum:
We use rollup for bundling.

The vast majority of rollup use today happens as part of vite projects ([stats](https://npmtrends.com/rollup-vs-vite)).

The Vite team use rollup for prod builds, because of the [larger feature set](https://vitejs.dev/guide/why#why-not-bundle-with-esbuild) and flexible plugin system (supports bundle splitting and more), but rollup is really slow compared to a lot of alternatives, so for dev builds esbuild is used instead.

To simplify this setup, the [Rolldown](https://rolldown.rs/) projects is in development - it's a rust port of rollup, with compatible api/feature set, and performance superior to esbuild.

We can swap to Rollup with minimal changes when there's a stable release. It's [quite far](https://x.com/youyuxi/status/1824023061714399292) in development already - I've tested the latest release in this codebase, and it's getting there.

HarelM:
I think we should evaluate this when it's a lot more stable and widely used. 
Last time we did a build change (I think it was swc) it ended up being rolled back due to issues we couldn't solve, and maybe only because we tried to use it prematurely.
We have a relatively complicated build (unfortunately) and we need this to work well - be battle tested. 
Just to clarify, I'm not against it, I just want to make sure we don't repeat past mistakes.

birkskyum:
I fully agree, I've been thinking a lot about our options in this space and recently seen the tangible [performance gain in bundling](https://github.com/plotly/plotly.js/pull/6909#issuecomment-2304328740) (10-20x) and work required to move to esbuild, and I just wanted to offload my thoughts on the topic, because one option looked more feasible than others.

It would be neat to get a native bundler (like swc/esbuild), but only if it's low cost to migrate (which in our rullup-setup would be rolldown), and regarding stability I also think we should only seriously evaluate this after it becomes the Vite default, which is still some 1-1.5 years out, and maybe even after it overtakes rollup usage.

[...]

PR #4728, opened 2024-09-20
Replace Jest with Vitest 2

birkskyum:
This PR replaces the test runner (unit/integration/build) from Jest to [Vitest 2](https://vitest.dev/), which has built-in support for TS/ESM and the same syntax as Jest. This PR serves to help us evaluate the option, and to move forward faster if there's a decision to proceed at some point.

A key motivation for examining this change is that I've found it quite challenging to move these tickets forward:
- #1595 
- #977 

Part of the struggle is that Jest only has [experimental ESM support](https://jestjs.io/docs/ecmascript-modules), and I've so far not managed to see an effect of the [useESM](https://kulshekhar.github.io/ts-jest/docs/next/guides/esm-support/) setting in the ts-jest plugin used in this repo.

It's a continuation of 
- #4684 

Relates to Rolldown, because vitest is downstream of vite, and thus eventually will swap internals to use rolldown:
- #4669 

Relates a bit to browser runner, because vitest is working on a browser mode:
- #4949 

## Launch Checklist


jasongitmail:
@birkskyum  I took a look at the unit test config as you requested. 

Looks good afaict. Coverage is in a separate command, `fileParallelism` is left at the default (true), etc.

If there is a particular file that consistently runs more slowly, using [`describe.concurrent()` within it could help](https://vitest.dev/api/#describe-concurrent). And I'd carefully look at how `beforeEach()` is used and consider if `beforeAll()` would be sufficient.

I don't claim to be a Vitest expert though. I switched from Jest -> Vitest -> Bun, and experienced speed wins each time. But you're using newer versions of Jest and Vitest, so things may have changed. And these tests use JSDom, which I've never needed to use, which is slower than running tests on node.

The benefit of this PR, is not only does Vitest provide ESM support, it brings some future proofing because Vitest should eventually be rewritten in native code, given VoidZero's vision for a rust based end-to-end toolchain.

birkskyum:
@jasongitmail , thank you very much for the review. It is a good idea to try to narrow down single files that might be impacting the numbers, and if beforeEach/All can help.

You can run Vitest with Bun? [I haven't had much success with that](https://github.com/oven-sh/bun/issues/4145). Unless the --bun flag is set, I believe it might be running vitest with node underneath still.

jasongitmail:
I didn't run Vitest with Bun, I ported my tests to `bun:test` by changing the imports `import { expect, test } from "bun:test";`. In my case, it mostly just worked because `bun:test` aims to be Jest compatible. https://bun.sh/docs/cli/test  

It's an option, but I'm not suggesting that in this case b/c `maplibre-gl-js` probably has many contributors and this would require them to have Bun installed to run tests, and I'd guess you'll likely end up switching back to Vitest down the road for ecosystem alignment. But it could be fun to play with to see how fast tests can be.

birkskyum:
@jasongitmail aah, that makes sense - the stats I've seen for bun:test are quite impressive, but I didn't even consider it when making this PR. As you say, it's requiring everyone to move to another runtime which would take a very convincing argument, and also until the day bun get a text-based lock-file, we'd be unable to use Dependabot, which we rely heavily on.

birkskyum:
This describes some of the differences between the two runners

https://cookbook.marmicode.io/angular/why-vitest/

HarelM:
My main concern is about coverage, I've spent a great deal of time to make the coverage work correctly for both render and unit tests. 
I still think it doesn't work well with latest jest (there's an open issue on it), but at least it's close to working correctly...

birkskyum:
@HarelM , is this how you'd expect it to look for the unit tests? it's the `npm run test-unit-ci`

Screenshot 2024-11-21 at 23 24 52

birkskyum:
This is the `npm run test-integration-ci`

Screenshot 2024-11-21 at 23 26 09

HarelM:
I'd like to solve the current problem we have we coverage, this might be an opportunity to do so while removing the need to use all kind of crazy package we have today like monocart etc.
Can you configure the CI to upload the coverage reports (json files) to codecov so that I can take a look at things I know are not reported correctly today?

birkskyum:
@HarelM , i've tried passing it the vitest json output for the Unit test suite to codecov - can you see if it's reasonable/?

HarelM:
It reports 70k lines, originally it reported arond 40k.
I'll need to see if it reports comments and non functional code...

birkskyum:
@HarelM , I made a mistake just before - it double counted the integration tests. should be resolved now

birkskyum:
note: This still uses monocart for the render tests. Vitest only swaps the Jest tests (unit/integration), so monocart is still there.

We can probably swap it by leveraging the [Vitest Browser Mode](https://vitest.dev/guide/browser/) (experimental) - But if possible, I'd like to avoid it for this PR.

birkskyum:
@HarelM the discrepancy in numbers is that vitest includes the .test.ts files. I'll try to exclude them.

birkskyum:
@HarelM  I'm seeing the same number of tracked lines now.

HarelM:
Should I be reviewing this? Did you close it by accident?

birkskyum:
It was closed by accident - I think it's ready for review

src/source/rtl_text_plugin_main_thread.test.ts
```diff
@@ -9,7 +10,7 @@ const rtlMainThreadPlugin = rtlMainThreadPluginFactory();
 
 describe('RTLMainThreadPlugin', () => {
     let server: FakeServer;
-    let broadcastSpy: jest.SpyInstance;
+    let broadcastSpy: ReturnType<typeof vi.spyOn>;
```
HarelM, line 13: Is there a better type for this?

birkskyum:
And I went with the beta release, because I saw they've made a change i the 2.2.0-beta.1 related to including the coverageMap in the json file output:

https://github.com/vitest-dev/vitest/releases

src/source/rtl_text_plugin_main_thread.test.ts
birkskyum: Yes, I've updated it to MockInstance

src/util/actor.test.ts
```diff
@@ -94,11 +95,11 @@ describe('Actor', () => {
         expect(gotAbortSignal).toBeTruthy();
     });
 
-    test('cancel a request that must be queued will not call the method at all', async () => {
+    test('cancel a request that must be queued will not call the method at all', {retry: 3}, async () => {
```
HarelM, line 98: Why was this added?

src/util/is_char_in_unicode_block.test.ts
```diff
@@ -1,14 +1,17 @@
+import {describe, test, expect} from 'vitest';
 import {unicodeBlockLookup} from './is_char_in_unicode_block';
 
 describe('unicodeBlockLookup', () => {
     test('each code block lookup function follows the same pattern', () => {
         for (const codeBlock in unicodeBlockLookup) {
             const lookup = unicodeBlockLookup[codeBlock];
-            const match = lookup.toString().match(/^\(char\) => char >= 0x([0-9A-F]{4,6}) && char <= 0x([0-9A-F]{4,6})$/);
+            const lookupString = lookup.toString();
+            console.log(`Checking code block: ${codeBlock}, function: ${lookupString}`);
```
HarelM, line 9: Remove console.log.

src/util/web_worker_transfer.ts
```diff
@@ -56,7 +56,7 @@ export function register<T extends any>(
     },
     options: RegisterOptions<T> = {}
 ) {
-    if (registry[name]) throw new Error(`${name} is already registered.`);
+    if (registry[name]) console.warn(`${name} is already registered.`);
```
HarelM, line 59: Why was this logic changed as part of this PR?

birkskyum: Screenshot 2024-11-22 at 14 10 29

HarelM:
Coverage seems like it was increased, which seems correct for a last case I know of here, which I made sure the PR covers, but the report was lacking.
I hope this will be the end of the coverage problems.
If not by any other reason, this is a good enough reason for me to switch.
https://app.codecov.io/gh/maplibre/maplibre-gl-js/pull/4728/blob/src/util/dom.ts

As a side note, for consistency I would advise to move the vitest import to be the first import in all the tests files, but that my OCD I guess...

HarelM:
Is the content of src/util/test/util.ts not needed anymore? I see that it is barely covered by tests, which it is intended for...

src/util/web_worker_transfer.ts
birkskyum: I've put in place a check to avoid re-registration:
https://github.com/maplibre/maplibre-gl-js/pull/4728/commits/18bf9fff21b307dab82ecf9728be0c1a5032ff37

birkskyum:
> Is the content of src/util/test/util.ts not needed anymore? I see that it is barely covered by tests, which it is intended for...

I see >100 uses of the StubMap, so we'd need a replacement of that

src/util/web_worker_transfer.ts
```diff
@@ -56,7 +56,10 @@ export function register<T extends any>(
     },
     options: RegisterOptions<T> = {}
 ) {
-    if (registry[name]) throw new Error(`${name} is already registered.`);
+    if (registry[name]) {
+        console.warn(`${name} is already registered.`);
+        return;
+    }
```
birkskyum, line 62: Either the register() should only be called by explicit execution (not by just loading the file), or we can do this to let it handle multiple calls more gently

HarelM: How was it able to work before this change?

src/render/draw_symbol.test.ts
```diff
@@ -17,13 +18,13 @@ import {Style} from '../style/style';
...
-jest.mock('../source/source_cache');
-jest.mock('../source/tile');
-jest.mock('../data/bucket/symbol_bucket');
-jest.mock('../symbol/projection');
-(symbolProjection.getPitchedLabelPlaneMatrix as jest.Mock).mockReturnValue(mat4.create());
+vi.mock('./painter');
+vi.mock('./program');
+vi.mock('../source/source_cache');
+vi.mock('../source/tile');
+vi.mock('../data/bucket/symbol_bucket');
```
birkskyum, line 25: @HarelM  This reads the file again, so it ends up being executed twice.

src/util/web_worker_transfer.ts
birkskyum: I can imagine it's because Vitest is esm based, and imports modules a bit different.

HarelM (approved):
Looks good to me, I'm happy to see that the problem with the coverage is not fixed with this change.

birkskyum:
>  I'm happy to see that the problem with the coverage is not fixed with this change.

Is there still a problem? This v5 that it's not reporting correctly in PRs, is that still ongoing after this?

HarelM:
@birkskyum I've updated the repo and ran the tests locally. It takes around 14 seconds to complete the run. It used to take around 5 seconds with jest. Any ideas how to get it back to 5 seconds?

birkskyum:
I do have some idea - I outlined them [here](https://osmus.slack.com/archives/C03TFH5NE83/p1732379052720919?thread_ts=1732347264.351569&cid=C03TFH5NE83).

HarelM:
Fair enough, I'll continue monitoring the behavior.

PR #4730, opened 2024-09-21
Fix hash router for urls ending with a hash

### LINKED ISSUE #4732: Hash router adds second # if url ends with a #.
This bug was surfaced by the Vitest PR here:
- #4728

If there is a url ending with an empty hash like: http://localhost/#

Then the hash router would add a secondary hash when initializing, resulting in this:

`http://localhost/##3/0/0`

### Repro:
- We don't have a hash router example yet ( #4731 ), so go to any example, like the `cluster.html` and enable hash:true
- Go to http://localhost:9966/test/examples/cluster.html#
- Result`http://localhost:9966/test/examples/cluster.html##3/40.67/-103.59`

birkskyum:
Closes #4732 
- #4732 

I've added a test for this, and a fix which is to modify the regex that recognize the hash. The section following the hashtag should be zero/one/more chars (.*), instead of one/more chars (.+).

Result before this PR:
`http://localhost:9966/test/examples/cluster.html##3/40.67/-103.59`

Result after this PR:
`http://localhost:9966/test/examples/cluster.html#3/40.67/-103.59`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 
 ### 🐞 Bug fixes
 - Fix circle won't render on mesa 24.1 with AMD GPU ([#4062](https://github.com/maplibre/maplibre-gl-js/issues/4062))
+- Fix hash router for urls ending with a hashtag ([#4730](https://github.com/maplibre/maplibre-gl-js/pull/4730))
 - _...Add new stuff here..._
 
 ## 4.7.0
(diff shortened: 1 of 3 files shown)


HarelM (approved):
It's hard to review regexp changes... 
I hope this won't break previous behavior when # was not present...

HarelM:
Please add a changelog item.

birkskyum:
> It's hard to review regexp changes... I hope this won't break previous behavior when # was not present...

I've added 5 more init tests for different url endings and a changelog

PR #4731, opened 2024-09-21
hash-router example

birkskyum:
We don't have a dedicated hash router example.

In fact, none of the 100+ examples has the hash router enabled

## Launch Checklist


HarelM:
This is presented in an iFrame so I think this example will be hard to understand from the user point of view.
Not that I have a good suggestion how to solve this though...

birkskyum:
> This is presented in an iFrame so I think this example will be hard to understand from the user point of view. Not that I have a good suggestion how to solve this though...

Gotcha, completely forgot that.
I've made a small overlay that'll show iframe hash in the top of the example.

Screenshot 2024-09-21 at 21 40 56

HarelM:
Nice solution!
Can you regenerate the example image then?

test/examples/hash-router.html
```diff
@@ -0,0 +1,50 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>Hash routing</title>
+    <meta property="og:description" content="Initialize a map in an HTML element with MapLibre GL JS." />
```
HarelM, line 5: Is the description up to date?

test/examples/hash-router.html
```diff
@@ -0,0 +1,50 @@
...
+        container: 'map', // container id
+        hash: true,
+        style: 'https://demotiles.maplibre.org/style.json', // style URL
+        center: [0, 0], // starting position [lng, lat]
+        zoom: 1, // starting zoom
+        maplibreLogo: true
+    });
+    map.on('load', async () => {
+        updateUrl();
+        map.transformCameraUpdate = () => {
```
HarelM, line 45: Is this really needed?

HarelM: Why not simply use the `move` or `moveend` event?

birkskyum:
> Nice solution! Can you regenerate the example image then?

All be ready now.

test/examples/hash-router.html
```diff
@@ -0,0 +1,55 @@
...
+        zoom: 1, // starting zoom
+        maplibreLogo: true
+    });
+    map.on('load', async () => {
+        updateUrl();
+    });
+    map.on('move', async () => {
+        updateUrl();
+    });
+    map.on('moveend', async () => {
```
HarelM, line 49: Is async needed?

HarelM: Also, I think you can inline the update url method into here for better readability.

HarelM: And use only move event...?

birkskyum: If does feels a lot more responsive if it also update when you let go of the map. If it's only 'move', it'll update on 'movestart', but not until you let go and start a new drag

test/examples/hash-router.html
```diff
@@ -0,0 +1,52 @@
...
+        hash: true, // <- Enable hash routing
+        style: 'https://demotiles.maplibre.org/style.json',
+        center: [0, 0],
+        zoom: 1,
+        maplibreLogo: true
+    });
+
+    // Below can be ignored. It just prints the url to an overlay on top of the map
+    function updateUrl() {
+        setTimeout(() => {
```
HarelM, line 43: setInterval instead?

HarelM (approved):

Issue #4703, opened 2024-09-13
closest terrain tile detail level seems incorrect when camera pitch is too high


maxmueh:
-->

**maplibre-gl-js version**: 4.5.2 and also in the maplibre 3d terrain example 

**browser**: chrome

### Steps to Trigger Behavior

 1. incline camera as much as possible
 2. fly slowly over a mountain top
 3. might need a few tries

### Link to Demonstration

Reproducable in the 3d terrain example: https://maplibre.org/maplibre-gl-js/docs/examples/3d-terrain/  

### Expected Behavior
The closer you get, the higher the detail of the tile

### Actual Behavior
While moving slowly over the peak of a mountain with an inclined camera, the closer you get, the higher gets the resolution of the tile. 

When reaching the peak, a tile with a lower resolution is loaded though, instead of the highest resolution.

It seems that since the tile is 'under' the camera and not in camera direction, it is not recognized as being close. 
This is also reproducable in regions without mountains, but the effect gets more obvious here.  

https://github.com/user-attachments/assets/fe589c80-4f9b-4aa7-9466-665ddaecaeb1

HarelM:
I believe that level of details is calculated for the camera center and not for every part of the scene separately.
Feel free to dig into this, a PR would be welcomed!

clement-igonet:
Hi [maxmueh](https://github.com/maxmueh)

   I was not able to reproduce.
   Could you have a try on last release ?

PR #4741, opened 2024-09-23

NathanMOlson:
This (failing) unit test demonstrates the issue reported here: https://github.com/maplibre/maplibre-gl-js/issues/3983 and https://github.com/maplibre/maplibre-gl-js/issues/4703

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Fix circle won't render on mesa 24.1 with AMD GPU ([#4062](https://github.com/maplibre/maplibre-gl-js/issues/4062))
 - Fix hash router for urls ending with a hashtag ([#4730](https://github.com/maplibre/maplibre-gl-js/pull/4730))
 - Replace rollup-plugin-sourcemaps with rollup-plugin-sourcemaps2 ([#4740](https://github.com/maplibre/maplibre-gl-js/pull/4740))
+- Fix level of detail at high pitch angle ([#3983](https://github.com/maplibre/maplibre-gl-js/issues/3983))
 - _...Add new stuff here..._
 
 ## 4.7.0
--- a/src/geo/transform.test.ts
+++ b/src/geo/transform.test.ts
@@ -333,6 +333,22 @@ describe('transform', () => {
         });
 
     });
+    
+    test('maxzoom-0', () => {
+        const options = {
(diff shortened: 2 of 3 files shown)


NathanMOlson:
This fixes https://github.com/maplibre/maplibre-gl-js/issues/3983 and improves https://github.com/maplibre/maplibre-gl-js/issues/4703 by calculating the desired zoom level for each tile independently. Each tile zoom is calculated by scaling  the center point zoom by the ratio of (3D) distance-to-tile to distance-to-center. Old behavior is preserved when there is no terrain and pitch < 60 degrees.

Before at 70 degrees pitch:
before_lodfix
After:
after_lodfix

Before at 80 degrees pitch:
before_lodfix_80
After:
after_lodfix_80

HarelM:
We are working on the globe branch where we moved a lot of the tile fetching logic to a different file.
There are two options here from my point of view:
1. Merge this to main and then merge/rewrite this in globe branch.
2. Merge this to globe branch and wait for version 5 release.

Let me know which works better for you.

Can you also show before and after of the showtilesboundries when terrain is disabled?

NathanMOlson:
The previously-posted images are with terrain disabled. Here is the before/after with terrain enabled:

Before:
before_lodfix_terrain
After:
after_lodfix_terrain

NathanMOlson:
> Thanks for taking the time to open this PR! We are working on the globe branch where we moved a lot of the tile fetching logic to a different file. There are two options here from my point of view:
> 
> 1. Merge this to main and then merge/rewrite this in globe branch.
> 2. Merge this to globe branch and wait for version 5 release.
> 
> Let me know which works better for you.

I just took a look at the `globe` branch and it looks like there's been some extensive rearrangement. So I think this will need a rewrite either way. Given that there will be no new 4.x releases, I don't see a reason to merge it to `main`. I'll work on moving it to `globe`.

HarelM:
I can carve out a new release if 4.x if this is something you absolutely need, but I would prefer to simply move this to 5.x if this is not urgent for you as it might take around 2 month to release a stable 5.x version.

NathanMOlson:
I'm happy to move this to 5.x.

NathanMOlson:
Superceded by https://github.com/maplibre/maplibre-gl-js/pull/4750

Issue #4717, opened 2024-09-17
Add API support for adjusting camera roll angle, extend pitch angle range


ssokol:
## User Story

As a developer I can programmatically adjust the roll angle of the camera from 0 (level) to +/- 180° and I can programmatically adjust the pitch angle from 0° (looking down) to 180° (looking directly up) so that I can accurately represent the spatial orientation of the observer when displaying maps with 3D terrain.

## Rationale

I am looking to use MapLibre in an aviation application to implement what is commonly know as "Synthetic Vision" - a feature which provides a virtual "view from the cockpit". To be able to accurately represent the this view, I need to be able to adjust the roll angle of the camera to match the roll angle of the airplane. You can see a similar system at a distinct roll angle here:

https://blog.foreflight.com/wp-content/uploads/2014/12/800x600-refined-terrain.png

From what I can find, the ```transformCameraUpdate``` callback currently allows programmatic control of bearing, center, elevation, pitch, and zoom (with pitch limited to 0° - 60°). Perhaps this could be extended to include the roll parameter and to accept pitch  values from 0 (straight down) to 180 (straight up).

It might also be useful to allow for all of these parameter to be incorporated into the parameter objects for ```map.flyTo()``` and ```map.easeTo()```. The way the Synthetic Vision app works, a server feeds situation data (location, altitude, track, speed, etc.) over a WebSocket at between 10-20 Hz. The handler for the WebSocket moves the camera using a string of very short duration flyTo or easeTo operations. Adding the ability to update situational parameters directly would simplify the process.

## Impact

This won't have any real impact on users who only need a basic down-and-forward level view, but it will go a long way towards making MapLibre more useful for certain simulations, games, and situational awareness applications.

HarelM:
Thanks for taking the time to open this issue.
In general, I think this can be a nice (although niche) addition to this lib.
The main caveat I can think of is which tile to fetch and show.
But feel free to push this forward.

ssokol:
Thanks! 

I'm slowly making my way through the library - long time Javascript user, first time Typescript user. I hope to be able to submit a PR, but it will probably need some help from those more fluent in the language. One question for you...

I haven't been able to find a way to manually set the camera altitude in feet or meters above median sea level. I kind of expected the easeTo and flyTo functions to have an altitude parameter, indicating the camera altitude at the end of the move.

[...]

PR #4771, opened 2024-09-28
Camera Roll

NathanMOlson:
@ssokol @Samarth1696

This is the first part of https://github.com/maplibre/maplibre-gl-js/issues/4717: adding camera roll.

In this PR, `roll` is defined as rotation about the camera boresight. Rotation order is `bearing`, `pitch`, `roll`. `easeTo()` and `flyTo()` are implemented using linear interpolation of Euler angles.

You can play with this feature here, using the controls in the top right corner: https://nathanmolson.github.io/camera_roll

There is some weirdness near +/- 90 degrees roll, but I'd like to get https://github.com/maplibre/maplibre-gl-js/pull/4750 merged before attempting to address it.

### DIFF
--- a/src/geo/projection/globe_transform.ts
+++ b/src/geo/projection/globe_transform.ts
@@ -93,6 +93,9 @@ export class GlobeTransform implements ITransform {
     setPitch(pitch: number): void {
         this._helper.setPitch(pitch);
     }
+    setRoll(roll: number): void {
+        this._helper.setRoll(roll);
+    }
     setFov(fov: number): void {
         this._helper.setFov(fov);
     }
@@ -180,6 +183,9 @@ export class GlobeTransform implements ITransform {
     }
     get pitch(): number {
         return this._helper.pitch;
+    }
+    get roll(): number {
+        return this._helper.roll;
     }
     get bearing(): number {
         return this._helper.bearing;
@@ -628,6 +634,7 @@ export class GlobeTransform implements ITransform {
         this._globeProjMatrixInverted = createMat4f64();
         mat4.invert(this._globeProjMatrixInverted, globeMatrix);
         mat4.translate(globeMatrix, globeMatrix, [0, 0, -this.cameraToCenterDistance]);
+        mat4.rotateZ(globeMatrix, globeMatrix, -this.roll * Math.PI / 180);
         mat4.rotateX(globeMatrix, globeMatrix, -this.pitch * Math.PI / 180);
         mat4.rotateZ(globeMatrix, globeMatrix, -this.angle);
         mat4.translate(globeMatrix, globeMatrix, [0.0, 0, -globeRadiusPixels]);
@@ -655,6 +662,7 @@ export class GlobeTransform implements ITransform {
         const zero = createVec3f64();
         this._cameraPosition = createVec3f64();
         this._cameraPosition[2] = this.cameraToCenterDistance / globeRadiusPixels;
+        vec3.rotateZ(this._cameraPosition, this._cameraPosition, zero, this.roll * Math.PI / 180);
         vec3.rotateX(this._cameraPosition, this._cameraPosition, zero, this.pitch * Math.PI / 180);
         vec3.rotateZ(this._cameraPosition, this._cameraPosition, zero, this.angle);
         vec3.add(this._cameraPosition, this._cameraPosition, [0, 0, 1]);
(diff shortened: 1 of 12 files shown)


src/geo/projection/globe_transform.ts
```diff
@@ -628,6 +634,7 @@ export class GlobeTransform implements ITransform {
         this._globeProjMatrixInverted = createMat4f64();
         mat4.invert(this._globeProjMatrixInverted, globeMatrix);
         mat4.translate(globeMatrix, globeMatrix, [0, 0, -this.cameraToCenterDistance]);
+        mat4.rotateZ(globeMatrix, globeMatrix, -this.roll * Math.PI / 180);
```
HarelM, line 637: rotateZ is called one with roll and a second time with this.angle. what's the difference?

HarelM:
There's a question of how to change roll based of gesture or keyboard/mouse keys. 
The demo looks nice 🙂

src/geo/projection/globe_transform.ts
NathanMOlson: The first `rotateZ()` performs the roll rotation, the second `rotateZ()` performs the heading rotation.

The two calls to `rotateZ()` are different because of the `rotateX(pitch)` between them. If `pitch` == 0, then there is no difference between roll and bearing. This is the singular point in the Euler angle representation.

NathanMOlson:
> There's a question of how to change roll based of gesture or keyboard/mouse keys.

My suggestion is to not change the current gesture control. Adding an third degree of freedom when the input device (touchscreen or mouse) has only 2 degrees of freedom can make it hard to achieve a desired camera orientation. I think for the use cases discussed so far, calling `setRoll()` programmatically is the best way to interact with roll.

HarelM:
This can be a good middle ground at this point in time.

Samarth1696:
This is a great demo for understanding!

I will also try to create a demo for the motion control but it will need some time for me.

src/geo/projection/mercator_camera_helper.ts
```diff
@@ -155,13 +160,25 @@ export class MercatorCameraHelper implements ICameraHelper {
...
                 tr.setZoom(interpolates.number(startZoom, endZoom, k));
             }
-            if (!quat.equals(startRotation, endRotation)) {
-                let rotation: quat = new Float64Array(4) as any;
-                quat.slerp(rotation, startRotation, endRotation, k);
-                const eulerAngles = getRollPitchBearing(rotation);
-                tr.setRoll(eulerAngles.roll);
-                tr.setPitch(eulerAngles.pitch);
-                tr.setBearing(eulerAngles.bearing);
+            if (this.useSlerp) {
```
NathanMOlson, line 163: This section shows the difference between linear interpolation of Euler angles and spherical linear interpolation (slerp). I intend to take this switch out and only include the slerp version before this PR gets merged, but for now this switch is used in the comparison demo here: https://nathanmolson.github.io/slerp

NathanMOlson:
I updated the code to do spherical linear interpolation instead  of linear interpolation of Euler angles. This is a change to the existing behavior of `easeTo()` and `flyTo()`. For single-axis rotations, (pure pitch or pure bearing), there is zero difference. For combined pitch+bearing rotations, the difference is very small for small rotations, and grows with the size of the rotation. To compare the two implementations side-by-side, I made a demo showing both versions (existing behavior on top, spherical linear interpolation on bottom). Use the "Big Move" button in the upper right corner.

https://nathanmolson.github.io/slerp

Even for this very large rotation, I can't see the difference between the two algorithms. @HarelM I'd like to switch to the spherical linear interpolation method despite the fact that it changes existing behavior. IMO the change is 1) not noticeable, 2) more correct, and 3) allows a consistent implementation when roll is added.

Samarth1696:
You did the right thing of using SLERP. Did you use linear interpolation in `camera_roll`?
https://nathanmolson.github.io/camera_roll/


https://github.com/user-attachments/assets/fa4bcee8-026c-48ab-bd91-3fe2d14ea528

I am getting different orientation after using those 4 buttons. Is it a gimbal lock or a `rotateZ()` problem or I am misunderstanding something?

HarelM:
I think this was mistakenly closed since globe branch was merged... weird...

NathanMOlson:
Reopened here: https://github.com/maplibre/maplibre-gl-js/pull/4780

Samarth1696:
> Yes, the camera_roll demo uses linear interpolation of Euler angles. That is why you see the unnecessary spin at at about 0:10 in your video.

I was asking about different orientation I am getting after 0:10. I am moving the mouse from top to bottom which should change the pitch from top to bottom but I am seeing the change sideways.

NathanMOlson:
> I was asking about different orientation I am getting after 0:10. I am moving the mouse from top to bottom which should change the pitch from top to bottom but I am seeing the change sideways.

I see what you mean. This is because the camera is rolled 90 degrees. The mouse movements do not change camera roll, and do not rotate with the camera (vertical mouse movements still control pitch).

Issue #4770, opened 2024-09-27
`queryRenderedFeatures` gets very slow with complex maps due to O(n^2) layerId lookup in `loadMatchingFeature`


tomhicks:
I know exactly how to fix this: we need to pass `layers` as a `Set` rather than a list and use `.has` instead of `arrayIntersects` and `filterLayerIDs.indexOf(layerID)`.

The problem here is that this is likely to alter the signature of some parts of MapLibre that might be considered public. For instance, although this is accessed usually via `map.queryRenderedFeatures` there are other entry points such as via the `Tile` class.

How should I proceed in making a PR here? We could accept either and check if it's an Array or Set all the way down?

**maplibre-gl-js version**: 4.7

**browser**: All

### Steps to Trigger Behavior

 1. Have a complex map with lots of layers (100s of layers are needed)
 2. Use query rendered features and look at the performance trace

### Link to Demonstration

An immense amount of time being spent in `loadMatchingFeatures`
<img width="607" alt="image" src="https://github.com/user-attachments/assets/950ac26d-a417-4113-ad11-77e81bab47f4">

### Expected Behavior

It shouldn't get quadratically slower as layers are added.

### Actual Behavior

It gets quadratically slower as layers are added.

HarelM:
Map is the relevant public API, tile is less interesting.
If you have a good solution you are welcome to open a PR.
We are also nearing version 5, so we might be able to introduce a breaking change.

PR #4777, opened 2024-09-30
Use sets to speed up queryRenderedFeatures

tomhicks:
## What this does

Fixes https://github.com/maplibre/maplibre-gl-js/issues/4770

Uses a Set instead of an Array for internals that traverse the "filter layer IDs" that a user passes when querying rendered features. This speeds up `queryRenderedFeatures` significantly when we have a large number of layers, as `Array.prototype.indexOf` becomes slow with large numbers of elements.

We convert to a Set as soon as possible, which means that some other semi-public APIs (e.g. `Tile::queryRenderedFeatures`) have changed to require a Set instead of an Array.

## Launch Checklist

- [X] Link to related issues.
- [X] Post benchmark scores.


src/style/style.ts
```diff
@@ -1433,6 +1434,12 @@ export class Style extends Evented {
         // LayerSpecification is serialized StyleLayer, and this casting is safe.
         const serializedLayers = this._serializedAllLayers() as {[_: string]: StyleLayer};
 
+        const layersAsSet = params.layers instanceof Set ? params.layers : Array.isArray(params.layers) ? new Set(params.layers) : undefined;
```
HarelM, line 1437: Wouldn't this code create performance issues when calling `queryRenderedFeatures` "fast"? For example when using mouse hover?

src/source/query_features.ts
```diff
@@ -7,6 +7,8 @@ import type {FilterSpecification} from '@maplibre/maplibre-gl-style-spec';
 import type {MapGeoJSONFeature} from '../util/vectortile_to_geojson';
 import type Point from '@mapbox/point-geometry';
 import {mat4} from 'gl-matrix';
+import {FeatureIndex} from '../data/feature_index';
```
HarelM, line 10: `import type ...`?

src/source/query_features.ts
```diff
@@ -102,7 +108,7 @@ export function queryRenderedFeatures(
     for (const tileIn of tilesIn) {
         renderedFeatureLayers.push({
             wrappedTileID: tileIn.tileID.wrapped().key,
-            queryResults: tileIn.tile.queryRenderedFeatures(
+            queryResults: (tileIn.tile as Tile).queryRenderedFeatures(
```
HarelM, line 111: Why was this added?

src/source/query_features.ts
```diff
@@ -16,7 +18,7 @@ export type QueryRenderedFeaturesOptions = {
      * An array of [style layer IDs](https://maplibre.org/maplibre-style-spec/#layer-id) for the query to inspect.
      * Only features within these layers will be returned. If this parameter is undefined, all layers will be checked.
      */
-    layers?: Array<string>;
+    layers?: Array<string> | Set<string>;
```
HarelM, line 21: Please update the comment above this field.

HarelM:
Can you add some test for this?

src/source/query_features.ts
tomhicks: Because without it the queryRenderedFeatures call is undiscoverable by TypeScript. I found that in a few places I ended up with some uncovered TS (types resolving to `any`) and these assertions were the minimal change to get some amount of type safety back.

src/style/style.ts
tomhicks: This will always be faster than the resulting calls into the `queryRenderedFeatures` helper functions, as they have a tendency to iterate over themselves. This will be O(n) and makes the rest of the call stack O(n) (often a loop of length n with an O(1) lookup inside) as opposed to O(n^2).

Also, the user can escape this conversion if they really need to by passing their own Set.

HarelM: Fair enough.

src/source/query_features.ts
HarelM: Is it possible to improve the typing that lead to this problem instead of adding `as ...`?
I know it's a bigger change, I know, but probably better...?

src/source/query_features.ts
tomhicks: It might be a good idea to enable that check in the tsconfig to avoid having to manually inspect things like this.

src/style/style.ts
tomhicks: Here's an example showing this performance gain, even if we convert to a set every time. In MapLibre the gains would be even more than this because there are I think at least two times where the arrays are iterated in a nested fashion.

https://jsperf.app/xekege

src/source/query_features.ts
HarelM: I'm not familiar with it, feel free to push this forward.

src/data/feature_index.ts
```diff
@@ -196,7 +196,7 @@ export class FeatureIndex {
         ) => boolean | number) {
 
         const layerIDs = this.bucketLayerIDs[bucketIndex];
-        if (filterLayerIDs && !arraysIntersect(filterLayerIDs, layerIDs))
+        if (filterLayerIDs && !layerIDs.filter(i => filterLayerIDs.has(i)))
```
neodescis, line 199: Did you mean to check the length of the filtered `layerIDs` array here? The `.filter()` function will always return something truthy.

tomhicks: er yes this is supposed to be `some` or `every` - good spot. I'll also try to add a breaking test.

src/source/source_cache.ts
```diff
@@ -948,9 +948,15 @@ export class SourceCache extends Evented {
      * @param pointQueryGeometry - coordinates of the corners of bounding rectangle
      * @returns result items have `{tile, minX, maxX, minY, maxY}`, where min/max bounding values are the given bounds transformed in into the coordinate space of this tile.
      */
-    tilesIn(pointQueryGeometry: Array<Point>, maxPitchScaleFactor: number, has3DLayer: boolean): any[] {
-
-        const tileResults = [];
+    tilesIn(pointQueryGeometry: Array<Point>, maxPitchScaleFactor: number, has3DLayer: boolean) {
+        type TileResult = {
```
HarelM, line 952: Can this be defined outside the method? Preferably in the upper part of the file.

src/data/feature_index.ts
```diff
@@ -113,7 +113,7 @@ export class FeatureIndex {
     ): {[_: string]: Array<{featureIndex: number; feature: GeoJSONFeature}>} {
         this.loadVTLayers();
 
-        const params = args.params || {} as { filter: any; layers: string[]; availableImages: string[] },
+        const params = args.params || {} as QueryParameters['params'],
```
HarelM, line 116: Use a type for this, it will help both here and in the above definition...

HarelM:
Besides my nit-picking, this looks good.
THANKS!

tomhicks:
I've tried enabling `noImplicitAny` to highlight the places where `any`s are getting into the code and it's a lot of places, so I won't be including that in this PR.

I will look at opening another one or two with [`noImplicitAny`](https://www.typescriptlang.org/tsconfig/#noImplicitAny) and [`verbatimModuleSyntax`](https://www.typescriptlang.org/tsconfig/#verbatimModuleSyntax) which should reduce nits and improve type safety.

HarelM:
No explicit any is a great idea, but as you said, should be out of scope of this PR.

src/source/source_cache.ts
```diff
@@ -948,9 +948,15 @@ export class SourceCache extends Evented {
...
+            tile: Tile;
+            tileID: OverscaledTileID;
+            queryGeometry: Array<Point>;
+            cameraQueryGeometry: Array<Point>;
+            scale: number;
+        };
+        const tileResults: Array<TileResult> = [];
```
neodescis, line 959:
```suggestion
        const tileResults: TileResult[] = [];
```
I think we should prefer [] over Array<>, at least for simple types

src/data/feature_index.ts
```diff
@@ -196,7 +196,7 @@ export class FeatureIndex {
         ) => boolean | number) {
 
         const layerIDs = this.bucketLayerIDs[bucketIndex];
-        if (filterLayerIDs && !arraysIntersect(filterLayerIDs, layerIDs))
+        if (filterLayerIDs && !layerIDs.some(id => filterLayerIDs.has(id)))
```
ibesora, line 199: Awesome work @tomhicks !

Have you considered the alternative approach of converting `bucketLayerIDs[bucketIndex]`to a `Set` instead?
It wouldn't require any change to any of the public signatures and get similar results by reversing the arrays in your condition to something like 
```typescript
if (filterLayerIDs && !filterLayerIDs.some(id => layerIDs.has(id)))
```

ibesora: Oh, although that would get similar results, your following improvement would not be possible 🤦 
Nothing to see here.

tomhicks: It may be beneficial to convert them both to sets at some point and use [Set.prototype.isDisjointFrom](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/isDisjointFrom) but that's very new so would need a shim for older browsers.

Also depends on whether there are any genuine array semantics necessary, such as mapping, filtering, etc.

tomhicks:
One thing I'm not quite sure of here is how to take this to completion by including some kind of regression test for performance.

The unit tests for `style.queryRenderedFeatures` use a very heavily stubbed map which wouldn't show the performance benefit. I notice there are benchmarks in `test/bench` but I'm not sure how to set up a very heavy map that would show this benefit, again.

I wonder if someone could help me do this? @ibesora as you're familiar with the "bike lanes" layer, this perf issue pops up if you've got say 100 of the bike lanes layers and you're querying around northern Europe. The density of features seems to be a factor aside from just number of layers.

src/data/feature_index.ts
tomhicks: FYI here are some performance comparisons for set intersection implementations: https://jsperf.app/modede

It might be worth considering this when there are other cases of indexOf usage where the size of the arrays can get very large. In smaller cases, the overhead of generating and accessing a Set will be slower than indexOf, so it really depends whether the dimensions are likely to get large or can get large.

ibesora:
@tomhicks I just pushed a commit to your branch that adds a benchmark. 
Instead of going all the way through from `style.queryRenderedFeatures` it just tests `featureIndex.loadMatchingFeature` which is where the performance changes happened.

Here it is a comparison without and with your improvements. This is testing a tile with 100 layers filtering by 100 non-existing layer ids, which is the worst case scenario we had before.
Screenshot 2024-10-04 at 10 12 17
tomsImprovements

Your implementation is basically an order of magnitude faster than what we currently have 💪

tomhicks:
thanks for adding that @ibesora - good to know it shows up in benchmarks, too.

I think this is pretty much good to go now?
- all the requested changes are addressed
- it performs better
- the important public APIs aren't changed
- it has tests that cover the new Array-or-Set APIs

src/source/source_cache.ts
```diff
@@ -1100,3 +1099,11 @@ function compareTileId(a: OverscaledTileID, b: OverscaledTileID): number {
 function isRasterType(type) {
     return type === 'raster' || type === 'image' || type === 'video';
 }
+
+type TileResult = {
```
HarelM, line 1103: Can this be moved to the beginning of the file please?

HarelM (approved):

src/source/source_cache.ts
tomhicks: uff, just pushed. thought i'd pushed this already

HarelM:
Can you add a changelog item?

tomhicks:
@HarelM changelog item added. I think that's all the things from the checklist.

HarelM (approved):

HarelM:
THANKS!

PR #4778, opened 2024-09-30
Globe: symbol and coveringTiles optimizations

kubapelc:
This PR ports symbol rendering optimizations from non-globe main branch to current globe-based main and adds some other globe-specific optimizations on top:

- Added a fast path for symbol projection for mercator projection, comparable to MapLibre 4.7 in performance - this just fixes a performance regression that was present in the globe branch.
- Improved collision box placement performance by not calling `isTilePositionOccluded` (the information was already computed beforehand).
- ~~Improved globe's coveringTiles performance by caching tile AABBs instead of recomputing them every time.~~
- ~~Fixed globe's coveringTiles incorrectly returning hundreds of hidden tiles as visible if the camera is sufficiently pitched.~~
- ~~Improved globe's coveringTiles tile visibility accuracy, reducing the number of returned tiles and thus improving performance.~~
- Added a `SymbolCollisionBoxGlobe` benchmark for symbol placement under globe projection. This is the same as `SymbolCollisionBox`, just with globe projection used instead of mercator.
- Fixes bad line scaling near the poles
- Fixed globe loading tiles at unnecessarily high zoom level when pitched

Below are benchmark results for mercator projection. Here "main" essentially represents globe branch without optimizations ported over, thus it is relatively slow.

mercator symbol placement benchmark

These next two figures are the new globe placement benchmark, first on the current main branch, then on this PR branch.

globe placement on current main

globe placement on this PR branch


src/geo/projection/globe_covering_tiles.ts
```diff
@@ -19,6 +20,78 @@ type CoveringTilesStackEntry = {
...
+ * A more accurate function for determining whether a tile is hidden below horizon.
+ * @param cameraX - Camera mercator X (0..1)
+ * @param cameraY - Camera mercator Y (0..1)
+ * @param tileX - Tile X (0..2^z)
+ * @param tileY - Tile Y (0..2^z)
+ * @param tileZ - Tile Z (zoom)
+ * @param plane - Horizon plane.
+ * @returns True if the tile is entirely below horizon.
+ */
+function tileBelowHorizon(cameraX: number, cameraY: number, tileX: number, tileY: number, tileZ: number, plane: vec4): boolean {
```
HarelM, line 38: I think this method deserves some unit tests :-)

src/geo/projection/globe_covering_tiles.ts
```diff
@@ -99,11 +172,20 @@ function getWrap(centerCoord: MercatorCoordinate, tileX: number, tileSize: numbe
     return 0;
 }
 
+// Tile AABBs are static, cache them!
+const tileAabbCache: Map<string, Aabb> = new Map();
```
HarelM, line 176: Can this grow and create a memory issue? Do we need to clean this as a certain point in time?

src/geo/projection/globe_transform.test.ts
```diff
@@ -439,7 +439,7 @@ describe('GlobeTransform', () => {
             const globeTransform = createGlobeTransform(globeProjectionMock);
 
             expect(globeTransform.getGlobeViewAllowed()).toBe(true);
-            expect(globeTransform.useGlobeControls).toBe(true);
+            expect(globeTransform['_globeRendering']).toBe(true);
```
HarelM, line 442: Is there a better way to test this by not using an internal field?

src/geo/projection/globe_transform.ts
```diff
@@ -232,6 +232,7 @@ export class GlobeTransform implements ITransform {
 
     private _projectionMatrix: mat4 = createIdentityMat4f64();
     private _globeViewProjMatrix: mat4 = createIdentityMat4f64();
+    private _globeViewProjMatrix32f: mat4;
```
HarelM, line 235: Why is this extra matrix is needed? Can't we use the 64 bit matrix?

HarelM:
I've added a few comments. thanks for this!
There is a PR that also looks at covering tiles here, it would be great to get your input there:
- #4750

(it's currently closed due to globe branch deletion, it should be opened soon).

HarelM:
New PR is here: 
- https://github.com/maplibre/maplibre-gl-js/pull/4779

@nathanmolson - can you check that the covering tiles optimization and improvements done here are incorporated in your PR - I would like to separate responsibilities: have Nathan's PR address covering tiles stuff and have this PR focus on symbol optimizations.

src/geo/projection/globe_covering_tiles.ts
```diff
@@ -171,28 +253,37 @@ export function getTileAABB(tileID: {x: number; y: number; z: number}): Aabb {
...
  * @returns 0 is not visible, 1 if partially visible, 2 if fully visible.
  */
-function isTileVisible(frustum: Frustum, plane: vec4, x: number, y: number, z: number): IntersectionResult {
-    const tileID = {x, y, z};
-    const aabb = getTileAABB(tileID);
+function isTileVisible(frustum: Frustum, plane: vec4, x: number, y: number, z: number, cameraX: number, cameraY: number): IntersectionResult {
+    // First test if the tile is below horzion.
+    // This alone cannot determine whether the tile is only partially hidden,
+    // so the AABB-plane test below is still needed.
+    if (tileBelowHorizon(cameraX, cameraY, x, y, z, plane)) {
```
NathanMOlson, line 282: `tileBelowHorizon()` is a fairly expensive function that is now called for every candidate tile. It will only result in early return for tiles below horizon, which will be a small fraction of candidate tiles. Has this been profiled to actually be faster?

kubapelc:
@HarelM @NathanMOlson thank you for your feedback! I will address it and improve this PR in a week or two, sadly I need to focus on other things in the meantime. I think this PR will not create merge conflicts with the terrain tile LOD PR, since that changes different parts of the covering tiles logic.

NathanMOlson:
> can you check that the covering tiles optimization and improvements done here are incorporated in your PR

The two optimizations for covering tiles in this PR are

1. Caching of tile AABBs. This needs changes to address unbounded memory growth.
2. Early return if `tileBelowHorizon()`. I'm not convinced this is actually an improvement.

I don't think these changes are ready to be incorporated in my PR.

src/geo/projection/globe_covering_tiles.ts
kubapelc: It can, good catch. I will definitely address this. First idea that comes to mind is to only cache AABBs of tiles computed in the previous frame, since the set of tested tiles is unlikely to change significantly between frames.

HarelM:
I would advise to remove covering tiles related logic from this PR and optionally add it to @NathanMOlson's PR if applicable.

src/geo/projection/globe_covering_tiles.ts
kubapelc: I've reverted all coveringTiles changes for now, including this method. I will try to improve performance in a different way, most likely without using this method.

src/geo/projection/globe_covering_tiles.ts
kubapelc: Reverted. I will implement a better cache eventually.

src/geo/projection/globe_transform.test.ts
kubapelc: I've now exposed `_globeRendering` publicly under the name `IsGlobeRendering`.

src/geo/projection/globe_utils.ts
```diff
@@ -43,11 +43,26 @@ export function angularCoordinatesRadiansToVector(lngRadians: number, latRadians
     return vec;
 }
 
-export function projectTileCoordinatesToSphere(inTileX: number, inTileY: number, tileID: {x: number; y: number; z: number}): vec3 {
-    const mercator = tileCoordinatesToMercatorCoordinates(inTileX, inTileY, tileID);
-    const angular = mercatorCoordinatesToAngularCoordinatesRadians(mercator.x, mercator.y);
-    const sphere = angularCoordinatesRadiansToVector(angular[0], angular[1]);
-    return sphere;
+export function projectTileCoordinatesToSphere(inTileX: number, inTileY: number, tileIdX: number, tileIdY: number, tileIdZ: number): vec3 {
```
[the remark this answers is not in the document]
kubapelc: Added!

kubapelc:
I have reverted most of coveringTiles changes, with one exception at line 84:

```ts
return Math.min(centerDist, cameraDist) * 2 <= radiusOfMaxLvlLodInTiles * tileSize; // Multiply distance by 2, because the subdivided tiles would be half the size
```

This is a simple bugfix, before this change globe displayed unnecessarily many tiles at high zoom levels. This alone improves performance a lot when camera is pitched.

I will try to implement a better tile culling by using general bounding frustums instead of AABBs. Frustums can approximate the actual tile shape much more tightly, thus leading to more discarded tiles, thus improving performance. A frustum-frustum visibility test should not be much slower than frustum-AABB test, but I will measure the times and create a benchmark. I will also implement a tile bounding volume cache that does not leak memory. I will do that as a separate PR.

src/geo/projection/globe_utils.ts
```diff
@@ -188,3 +212,15 @@ export function interpolateLngLatForGlobe(start: LngLat, deltaLng: number, delta
         );
     }
 }
+
+export function clampLngLat(val: LngLat, min: LngLat, max: LngLat): LngLat {
```
HarelM, line 216: It might be better to move this method to the lnglat file...?

src/geo/projection/mercator_transform.ts
```diff
@@ -197,8 +197,8 @@ export class MercatorTransform implements ITransform {
     private _fogMatrix: mat4;
 
     private _posMatrixCache: {[_: string]: mat4};
-    private _fogMatrixCache: {[_: string]: mat4};
     private _alignedPosMatrixCache: {[_: string]: mat4};
+    private _fogMatrixCache: {[_: string]: mat4};
```
HarelM, line 201: Can you see if you can move things a bit to reduce changes that are "not real"?

src/geo/projection/mercator_transform.ts
```diff
@@ -406,7 +405,7 @@ export class MercatorTransform implements ITransform {
         const fogMatrix = calculateTileMatrix(unwrappedTileID, this.worldSize);
         mat4.multiply(fogMatrix, this._fogMatrix, fogMatrix);
 
-        cache[posMatrixKey] = fogMatrix;
+        cache[posMatrixKey] = new Float32Array(fogMatrix);
```
HarelM, line 408: Why is this needed? We just created this matrix, didn't we?

src/geo/projection/globe_transform.ts
```diff
@@ -295,13 +300,13 @@ export class GlobeTransform implements ITransform {
...
+    public get projectionMatrix(): mat4 { return this.IsGlobeRendering ? this._projectionMatrix : this._mercatorTransform.projectionMatrix; }
 
-    public get modelViewProjectionMatrix(): mat4 { return this._globeRendering ? this._globeViewProjMatrixNoCorrection : this._mercatorTransform.modelViewProjectionMatrix; }
+    public get modelViewProjectionMatrix(): mat4 { return this.IsGlobeRendering ? this._globeViewProjMatrixNoCorrection : this._mercatorTransform.modelViewProjectionMatrix; }
 
-    public get inverseProjectionMatrix(): mat4 { return this._globeRendering ? this._globeProjMatrixInverted : this._mercatorTransform.inverseProjectionMatrix; }
+    public get inverseProjectionMatrix(): mat4 { return this.IsGlobeRendering ? this._globeProjMatrixInverted : this._mercatorTransform.inverseProjectionMatrix; }
 
-    public get useGlobeControls(): boolean { return this._globeRendering; }
+    public get useGlobeControls(): boolean { return this.IsGlobeRendering; }
```
HarelM, line 309: Is this public getter still needed? I saw that it was removed from the transform interface...?

src/geo/projection/globe_transform.ts
```diff
@@ -528,6 +533,14 @@ export class GlobeTransform implements ITransform {
         return !this.isSurfacePointVisible(angularCoordinatesToSurfaceVector(location));
     }
 
+    public tileCoordinatesOccluded(inTileX: number, inTileY: number, canonicalTileID: {x: number; y: number; z: number}): boolean {
```
HarelM, line 536: I couldn't find where this is being used as part of this PR. Am I missing anything?

HarelM:
Added a few more comments. Otherwise looks good enough to be merged. Thanks!

src/geo/projection/globe_utils.ts
kubapelc: This function is no longer in use, so I removed it.

src/geo/projection/mercator_transform.ts
kubapelc: Done!

src/geo/projection/mercator_transform.ts
kubapelc: This was another attempt at speeding things up by using float32. Reverted.

src/geo/projection/globe_transform.ts
kubapelc: It is not, removed. I also renamed `IsGlobeRendering` to `isGlobeRendering` (proper camelCase).

src/geo/projection/globe_transform.ts
kubapelc: It isn't used anymore. Removed.

HarelM:
Can you update the branch from main to resolve conflicts?
I believe other comments were addressed.

kubapelc:
Should be fixed now, I also fixed the changelog wording and the failing render test (reverted the expected image to its state before this PR).

HarelM (approved):

PR #4779, opened 2024-09-30

NathanMOlson:
Reopening accidentally closed PR: https://github.com/maplibre/maplibre-gl-js/pull/4750

This fixes https://github.com/maplibre/maplibre-gl-js/issues/3983 and improves https://github.com/maplibre/maplibre-gl-js/issues/4703 by calculating the desired zoom level for each tile independently. Each tile zoom is calculated by scaling the center point zoom by the ratio of (3D) distance-to-tile to distance-to-center. For Mercator transform, the old behavior is preserved when there is no terrain and pitch < 60 degrees. For Globe transform, the new behavior applies in all cases.

Before:

before_terrain
before

After:

after2_terrain
after2

Here's my understanding of the previous implementation:
image
image

And here's a drawing of how it works now (in crude 3d):

image

Here's a description of how it works.

The center point is rendered at the requested zoom level.
the 3D distance from the camera to each candidate tile is calculated and the desired zoom for that tile is the requested zoom scaled by ratio of camera-to-center distance to camera-to-tile distance. If the candidate tile is lower resolution than desired, the tile is split.
Note that rather than use the true 3D distance to each tile (accounting for terrain), it is assumed that all tiles are at the same elevation as the center point. This prevents "resolution explosion" when the camera is very close to a hill, and keeps the resolution variation from map center smooth.


src/geo/transform_interface.ts
```diff
@@ -41,6 +41,11 @@ export type CoveringTilesOptions = CoveringZoomOptions & {
      * When terrain is present, tile visibility will be computed in regards to the min and max elevations for each tile.
      */
     terrain?: Terrain;
+    /**
+     * Controls how tiles are loaded at high pitch angles. Higher numbers cause fewer, lower resolution
+     * tiles to be loaded. A reasonable range for this parameter is [0,2].
+     */
+    pitchBehavior?: number;
```
HarelM, line 48: As stated before, maybe the right approach would be to give a zoom range for pitch - when 1 is all tiles at the same zoom as the center and higher number means higher variability.
Zoom is more in user terminology space.

src/geo/projection/globe_transform.ts
```diff
@@ -87,6 +87,10 @@ export class GlobeTransform implements ITransform {
     setRenderWorldCopies(renderWorldCopies: boolean): void {
         this._helper.setRenderWorldCopies(renderWorldCopies);
     }
+    setPitchBehavior(pitchBehavior: number): void {
+        this._helper.setPitchBehavior(pitchBehavior);
+        this._mercatorTransform.setPitchBehavior(pitchBehavior);
```
HarelM, line 92: Why do we need to set both? are they not using the same helper?

src/geo/projection/globe_covering_tiles.ts
```diff
@@ -215,28 +201,23 @@ function isTileVisible(frustum: Frustum, plane: vec4, x: number, y: number, z: n
...
-    if (options.maxzoom !== undefined && z > options.maxzoom) {
-        z = options.maxzoom;
-    }
-
-    const numTiles = Math.pow(2, z);
+export function globeCoveringTiles(frustum: Frustum, plane: vec4, cameraCoord: MercatorCoordinate, centerCoord: MercatorCoordinate, tileSize: number, zoom: number, pitch: number, fov: number, pitchBehavior: number, options: CoveringTilesOptions): OverscaledTileID[] {
+    let nominalZ = (options.roundZoom ? Math.round : Math.floor)(zoom + scaleZoom(tileSize / options.tileSize));
+    const minZoom = options.minzoom || 0;
+    const maxZoom = options.maxzoom !== undefined ? options.maxzoom : nominalZ + 3;
+    console.log(zoom, nominalZ, tileSize, options.tileSize, maxZoom);
```
HarelM, line 208: remove console.log

src/geo/projection/globe_covering_tiles.ts
```diff
@@ -215,28 +201,23 @@ function isTileVisible(frustum: Frustum, plane: vec4, x: number, y: number, z: n
...
-
-    if (options.minzoom !== undefined && z < options.minzoom) {
-        return [];
-    }
-    if (options.maxzoom !== undefined && z > options.maxzoom) {
-        z = options.maxzoom;
-    }
-
-    const numTiles = Math.pow(2, z);
+export function globeCoveringTiles(frustum: Frustum, plane: vec4, cameraCoord: MercatorCoordinate, centerCoord: MercatorCoordinate, tileSize: number, zoom: number, pitch: number, fov: number, pitchBehavior: number, options: CoveringTilesOptions): OverscaledTileID[] {
```
HarelM, line 204: Looks like there some similarities between the two methods, do you think there's a way to make the code using some strategy pattern somehow and use common code? Do you think you can allow overriding this behavior from the outside somehow?

CHANGELOG.md
```diff
@@ -58,6 +59,8 @@ using `transformCameraUpdate` caused the `maxBounds` to stop working just for ea
 
 ### ✨ Features and improvements
 
+- Merge atmosphere an sky implementation ([#3888](https://github.com/maplibre/maplibre-gl-js/issues/3888))
```
HarelM, line 62: This is an incorrect merge...?

NathanMOlson: Fixed.

src/geo/transform_interface.ts
NathanMOlson: I don't like that idea. If we load all tiles at the same zoom as the center when the horizon is visible, we will load tens or hundreds of thousands of tiles in some cases.

You can see what that would be like by changing `pitchBehavior` to less than 0. https://nathanmolson.github.io/lodfix_demo Here's a screenshot when `pitchBehavior` = -1.
image

I don't think we ever want this.

src/geo/projection/globe_transform.ts
NathanMOlson: I don't thinks so. See line 278 and 285. `GlobeTransform` has its own helper, which it does not share with `this._mercatorTransform`. Unless I'm missing something?

src/geo/projection/covering_tiles.ts
```diff
@@ -0,0 +1,165 @@
...
+
+export type CoveringTilesStackEntry = {
+    zoom: number;
+    x: number;
+    y: number;
+    wrap: number;
+    fullyVisible: boolean;
+};
+/**
+ * Returns the distance from the point to the tile
```
HarelM, line 22: Can these 3 methods be converted to an interface that can be passed to the covering tiles method?

src/geo/projection/covering_tiles.ts
```diff
@@ -0,0 +1,165 @@
...
+ * @param tileID - Tile x, y and z for zoom.
+ * @param wrap - wrap number of the tile.
+ * @param elevation - camera center point elevation.
+ * @param options - CoveringTilesOptions.
+ */
+type GetTileAABBFunc = (tileID: {x: number; y: number; z: number}, wrap: number, elevation: number, options: CoveringTilesOptions) => Aabb;
+
+/**
+ * A simple/heuristic function that returns whether the tile is visible under the current transform.
+ * @returns 0 is not visible, 1 if partially visible, 2 if fully visible.
```
HarelM, line 44: I would like to the `IntersectionResult` enum instead of duplicating its definition.

src/geo/projection/covering_tiles.ts
```diff
@@ -0,0 +1,165 @@
...
+ */
+export function isTileVisible(frustum: Frustum, plane: vec4, aabb: Aabb): IntersectionResult {
+
+    const frustumTest = aabb.intersectsFrustum(frustum);
+    if (!plane) {
```
HarelM, line 49: Does this mean plane is an optional parameter? In that case I would move it to be the last parameter and mark it with "?".

src/geo/projection/covering_tiles.ts
```diff
@@ -0,0 +1,165 @@
...
+    const nominalZ = Math.min(Math.max(0, desiredZ), maxZoom);
+
+    const numTiles = Math.pow(2, nominalZ);
+    const cameraPoint = [numTiles * cameraCoord.x, numTiles * cameraCoord.y, 0];
+    const centerPoint = [numTiles * centerCoord.x, numTiles * centerCoord.y, 0];
+    const distanceToCenter2d = Math.hypot(centerCoord.x - cameraCoord.x, centerCoord.y - cameraCoord.y);
+    const distanceZ = Math.abs(centerCoord.z - cameraCoord.z);
+    const distanceToCenter3d = Math.hypot(distanceToCenter2d, distanceZ);
+
+    const newRootTile = (wrap: number): any => {
```
HarelM, line 78: Return type should be `CoveringTilesStackEntry` instead of any?

src/geo/projection/globe_transform.ts
HarelM: There are not many methods (if at all) that there is a need call a set for both the helper and the mercator transform, feels like a smell to me, but I might be wrong...

src/geo/projection/covering_tiles.ts
NathanMOlson: done.

src/geo/projection/covering_tiles.ts
NathanMOlson: done.

src/geo/projection/covering_tiles.ts
NathanMOlson: done.

src/geo/projection/covering_tiles.ts
NathanMOlson: done.

src/geo/transform_interface.ts
HarelM: Yes, this is not a desired behavior, I agree, but I still want this property to be in user terms and not arbitrary number...

src/geo/projection/covering_tiles.ts
```diff
@@ -0,0 +1,185 @@
...
+
+type CoveringTilesStackEntry = {
+    zoom: number;
+    x: number;
+    y: number;
+    wrap: number;
+    fullyVisible: boolean;
+};
+
+export interface CoveringTilesDetails {
```
HarelM, line 22: "Details" seems a bit off to me, I thought about "context" but it's not exactly it too...
CoveringTilesSetting maybe, IDK...

HarelM: Maybe all the method parameters (except the trasform) can be added to this object and call it CoveringTilesOptions like we do for other objects...?

NathanMOlson: > Maybe all the method parameters (except the trasform) can be added to this object and call it CoveringTilesOptions like we do for other objects...?

We already have a `CoveringTilesOptions`. What should happen to that object, which needs to be created at a high level that doesn't know about the `CoveringTilesDetails` or other method parameters?

HarelM: In general, I'd like to move the covering tiles outside the scope of the transform, i.e. it should not be a method of the trasform but something that receives a transform and being called where needed, but it might be out of scope for this PR.

NathanMOlson:
I've been working to improve the parameters used to control the tile-loading behavior at high pitch angles.

In addition to the previously demonstrated, unfortunately named, `pitchTileLoadingBehavior`, I've added `tileZoomDeadband`, which works as follows: If z14 is requested at the center point, any tile whose computed zoom is within `tileZoomDeadband` of z14 will use z14. So if `tileZoomDeadband` is 1, and the computed zoom for a tile is z13-z15, it will use z14. If the computed zoom is z12, it will use z13, if the computed zoom is z11, it will use z12, etc.

This provides a way to do what @HarelM requested, which is to be able to reduce the amount of variability in loaded zoom levels.

I also tried doing a weighted average of the computed zoom level and the center zoom level, where the weighting was controlled by a `tileZoomVariability` parameter. However, that approach worked less well. The response to the behavior was less uniform, and any value that had an effect loaded too many tiles near the horizon.

The two parameters are demonstrated here: https://nathanmolson.github.io/lodfix_demo

Here is how the number of tiles computed by `coveringTiles()` changes with these two parameters:

image

Notice the behavior change when the heuristic horizon comes into view at 69 degrees pitch. Here's the same plot, zoomed in to show detail in " fully supported" range of 0-60 degrees pitch:

image

The plots show that decreasing tile zoom variability (by increasing `tileZoomDeadband`) causes more tiles to be loaded.

For me, the defaults `pitchTileLoadingBehavior=1`, `tileZoomDeadband=0` result in a good compromise between tile variability and number of tiles loaded. I could also be persuaded to set the default `tileZoomDeadband` to 0.5 or 1.

Also worth noting: The plots were created with the "special case" (constant zoom level if no terrain and pitch < 60) ignored. With the right selection of parameters, we could match the current behavior for the special case without explicitly enforcing it. This would allow a smooth transition from the "special case" to outside the special case, where there is currently a noticeable discrete transition.  I'm not recommending this, it's just an idea. My recommendation is instead to get this working how we like and separately decide whether or not to preserve the special case.

HarelM:
Thanks a lot for pushing this forward!
I've played with the demo you provided in order to see if I understand how every parameter controls the tile loading but unfortunately it wasn't intuitive when to change each parameter and how each one of them affects the map - i.e. I couldn't "predict" what will happen with each parameter change, all I could do is play with it and see how it changes.
I also think that two confusing parameters are worse than one :-)
I don't have a good solution to this.
The only improvement I could think of is using a `enum` + `number` value for the "behavior" so that you can name `0` to be `LowZoomLevelAtTheHorizon`, 2 to be `HighZoomLevelAtTheHorizon` and leave it to be also set as a number so you could manipulate it as you need in extreme cases.
I think it makes sense to remove the 60 pitch edge case, but I think the current behavior that the entire map has a single zoom level is a good behavior we should keep - i.e. change the code so that pitch 60 using the default parameters value will behave as it behaves before this fix/change.

NathanMOlson:
@HarelM what if the enum values were `uniformTileScreenWidth`, `uniformTileScreenArea`, and `uniformTileScreenHeight`? That's basically what 0, 1 and 2 do.

Any ideas what I can do with `tileZoomDeadband`? The best way to understand it is: If you increase `tileZoomDeadband`, the region that shares zoom level with the center grows outward in both directions.

I would also be in favor of getting rid of (meaning hardcoding to default value) one or both of these parameters. Then we don't have to worry about users understanding them.

I think 60 degrees is too high to show at constant zoom level. At 60 degrees, the top of the viewport is at 78 degrees from nadir. At 60 degrees, the screen width of tiles at the top of the viewport is less than 1/3 the screen width of tiles at the bottom of the viewport. To me, it seems that constant zoom level 1) loads too many tiles, 2) doesn't render distant tiles at an appropriate (lower) zoom level. I think 45 degrees would be a more appropriate cutoff point.

HarelM:
I would say that the best way to be able to get rid of these parameters would be to allow supplying an override for this method - this way if someone would like to change the behavior they should write a small method to return a different results for covering tiles (we have that for transformcameraUpdate, transfromRequest, addProtocol and other hooks).
If we can facilitate this, then we can leave out all this configuration and allow the user a full control over the method.
It might not fit as part of this PR, but I think it might the best approach.
Let me know what you think.

NathanMOlson:
> I would say that the best way to be able to get rid of these parameters would be to allow supplying an override for this method - this way if someone would like to change the behavior they should write a small method to return a different results for covering tiles (we have that for transformcameraUpdate, transfromRequest, addProtocol and other hooks).
> If we can facilitate this, then we can leave out all this configuration and allow the user a full control over the method.
> It might not fit as part of this PR, but I think it might the best approach.
> Let me know what you think.

I think "write your own implementation of `coveringTiles()`" is too onerous of a task for library users, to the point where it will almost never happen in practice. However, all the "interesting" stuff happens in the few lines where we compute `thisTileDesiredZ`. Maybe we could allow an override of `calculateTileZoom()`? The only challenge would be figuring out what the inputs arguments would be. The current algorithm would use 

```
calculateTileZoom(requestedCenterZoom: number,
    distanceToTile2D: number,
    distanceToTileZ: number,
    distanceToCenter3D: number,
    cameraVFOV: number);
```

but other implementations might want other info.

Implementing this function would be a lot easier than implementing all of `coveringTiles()`, but I still have the feeling it would be a feature that doesn't get much use.

HarelM:
@msbarry Any chance for your input here? I know you had some issues with contour plugin and covering tiles algorithm so maybe you can share what might have helped you in the past or what might help you in the future.

msbarry:
> @msbarry Any chance for your input here? I know you had some issues with contour plugin and covering tiles algorithm so maybe you can share what might have helped you in the past or what might help you in the future.

The pieces required for the contour source are:

1. expressing the ideal covering tiles (and backup tiles?) for a source, it's configuration, and how its used
2. building the staged tile pyramid for each usage of the source usage that provides backup tiles when ideal ones are missing
3. requesting/caching/deduplicating requests for tiles from a source to serve all usages of it

Right now those are all kind of coupled into one ball of logic in source_cache that's not flexible enough to share DEM tiles across hillshade and vector contour usages. It seems like this PR could help break (1) out. (1) should eventually be configurable by source/usage in addition to transform, but it seems like they would most likely just be based on this transform algorithm with a zoom offset for raster/vector/overzoomed.

rbrundritt:
Would occlusion culling be an additional potential performance enhancement related to this scenario? 

If the 3D bounding box of a tile is hidden behind other tiles, we can skip rendering it until it comes into view. This will likely play more of a role when elevations are being used that's likely when it would be most useful due to more complex meshes being used. In 2D, depending on pitch and camera elevation, it may be possible to skip the tile that's directly under the camera, however I suspect that will rarely be the case. 

A similar optimization may be to have an option to limit the depth of tiles rendered in the distance. For example, a person standing on the ground typically can see about 3 miles before distance objects start to be hidden by the curvature of the earth. Here is a table that relates viewable distance of the horizon relative to elevation above the ground: http://www.totally-cuckoo.com/distance_visible_to_the_horizon.htm (sorry for the horrible UI on that site, not sure why anyone would choose that combo).

Perhaps these are future things to look into but thought that since this is fresh in your mind, might be a good time to see if you think there may be some potential benefits with these.

NathanMOlson:
> A similar optimization may be to have an option to limit the depth of tiles rendered in the distance.

This is currently governed by what I call "the horizon heuristic", which is implemented in `getMercatorHorizon()` in `mercator_utils.ts`.  The current heuristic is not based on the geometry of the Earth. We could change it to better match the actual geometry, but we should be be careful to maintain one of the advantages of the Mercator transform: you can see the whole world at once, which is certainly not possible with the actual geometry :) Maybe the horizon could get progressively more accurate as the camera moves closer to the ground. I think updates to the horizon heuristic are independent of the changes proposed here, and outside the scope of this PR.

rbrundritt:
Ok, just brainstorming. I do like the infinite view option of Mercator at times, particularly for business intelligence type visualizations. At the moment I'm thinking from the perspective of a game developer and how fog is used to help limit the viewable distance so that far off meshes don't have to be rendered. Perhaps that makes more sense for globe view, or a separate option in the future.

src/source/source.ts
```diff
@@ -116,6 +117,10 @@ export interface Source {
      * Allows to execute a prepare step before the source is used.
      */
     prepare?(): void;
+    /**
```
NathanMOlson, line 120: @HarelM What do you think of this approach?

src/geo/projection/covering_tiles.ts
```diff
@@ -0,0 +1,231 @@
+import {OverscaledTileID} from '../../source/tile_id';
+import {Aabb, Frustum, IntersectionResult} from '../../util/primitives';
+import {vec2, vec4} from 'gl-matrix';
+import {CoveringTilesOptions, IReadonlyTransform} from '../transform_interface';
```
HarelM, line 4: Shouldn't `CoveringTilesOptions` be part of this file?

src/geo/projection/covering_tiles.ts
```diff
@@ -0,0 +1,231 @@
...
+    x: number;
+    y: number;
+    wrap: number;
+    fullyVisible: boolean;
+};
+
+/**
+ * Function to define how tiles are loaded at high pitch angles
+ * @param requestedCenterZoom - the requested zoom level, valid at the center point.
+ * @param distanceToTile2D - 2D distance from the camera to the candidate tile, in mercator units.
```
HarelM, line 26: What is mercator units?

src/geo/projection/covering_tiles.ts
```diff
@@ -0,0 +1,231 @@
...
+ * @param distanceToTileZ - vertical distance from the camera to the candidate tile, in mercator units.
+ * @param distanceToCenter3D - distance from camera to center point, in mercator units
+ * @param cameraVFOV - camera vertical field of view, in degrees
+ * @return the desired zoom level for this tile. May not be an integer.
+ */
+export type CalculateTileZoomFunction = (requestedCenterZoom: number,
+    distanceToTile2D: number,
+    distanceToTileZ: number,
+    distanceToCenter3D: number,
+    cameraVFOV: number) => number;
```
HarelM, line 36:
```suggestion
    cameraVertivalFOV: number) => number;
```

src/source/source.ts
HarelM: Ahh, this is interesting! @msbarry what do you think?
I tend to think this has the potential to allow defining the relevant tiles per source, and this is not an API that is widely used although public.

src/source/source_cache.ts
```diff
@@ -626,7 +626,8 @@ export class SourceCache extends Evented {
                 maxzoom: this._source.maxzoom,
                 roundZoom: this.usedForTerrain ? false : this._source.roundZoom,
                 reparseOverscaled: this._source.reparseOverscaled,
-                terrain
+                terrain,
+                calculateTileZoom: this._source.calculateTileZoom
```
HarelM, line 630: What would it take to call `coveringTiles` method and passing it the transform object instead of how it is now? Feels like this is doable after all this changes, isn't it?

src/geo/projection/globe_covering_tiles.ts
```diff
@@ -186,110 +167,20 @@ export function getTileAABB(tileID: {x: number; y: number; z: number}): Aabb {
...
-            const childY = (y << 1) + (i >> 1);
-            const childZ = it.zoom + 1;
-            stack.push({zoom: childZ, x: childX, y: childY, fullyVisible});
-        }
-    }
-
-    return result.sort((a, b) => a.distanceSq - b.distanceSq).map(a => a.tileID);
+export function globeCoveringTiles(transform: IReadonlyTransform, frustum: Frustum, plane: vec4, cameraCoord: MercatorCoordinate, centerCoord: MercatorCoordinate, options: CoveringTilesOptions): OverscaledTileID[] {
+    const allowVariableZoom = transform.coveringZoomLevel(options) > 4;
+    const details: CoveringTilesDetails = {
```
HarelM, line 179: Can't we simply create a static class/object to hold these methods? You already defined an interface, so why not implement it?
Feels weird to create this object again and again as it is a stateless object, right?

src/geo/projection/globe_covering_tiles.ts
```diff
@@ -103,7 +84,7 @@ function getWrap(centerCoord: MercatorCoordinate, tileX: number, tileSize: numbe
  * Returns the AABB of the specified tile. The AABB is in the coordinate space where the globe is a unit sphere.
  * @param tileID - Tile x, y and z for zoom.
  */
-export function getTileAABB(tileID: {x: number; y: number; z: number}): Aabb {
+export function getTileAABB(tileID: {x: number; y: number; z: number}, _wrap: number, _elevation: number, _options: CoveringTilesOptions): Aabb {
```
HarelM, line 87: Is this used outside this file that it needs to be exported?

src/geo/projection/covering_tiles.ts
HarelM: I think `helper` is probably a better name than details, although both aren't very descriptive.
Maybe `CoveringTilesDetailsProvider`?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - ⚠️ Fix order of normalizeSpriteURL and transformRequest in loadSprite ([#3897](https://github.com/maplibre/maplibre-gl-js/issues/3897))
 - ⚠️ Remove unminified prod build ([#4906](https://github.com/maplibre/maplibre-gl-js/pull/4906))
 - Fix issue where raster tile source won't fetch updates following request error ([#4890](https://github.com/maplibre/maplibre-gl-js/pull/4890))
+- Fix level of detail at high pitch angle ([#3983](https://github.com/maplibre/maplibre-gl-js/issues/3983))
```
HarelM, line 11:
```suggestion
- ⚠️ Fix level of detail at high pitch angle by changing which tiles to load ([#3983](https://github.com/maplibre/maplibre-gl-js/issues/3983))
```

src/geo/transform_interface.ts
HarelM: This was removed, so there's no more confusion to the end user. Great! Thanks!

src/geo/projection/globe_transform.ts
HarelM: This was removed, so this comment is no longer relevant, Thanks!

HarelM:
I think this looks good now, besides a few minor comments.
If we can also nail removing the `coveringTiles` method from the transform interface to a method that receives `IReadonlyTransform` it will be great!
It might be an interesting case for the suggested `CalculateTileZoomFunction` to see if it can facilitate the difference between globe an mercator? This is just a thought to see if the provided interface can cover this scenario, or if more parameters are needed. IDK, not mandatory, just a thought experiment I guess...

src/geo/projection/covering_tiles.ts
NathanMOlson: Done.

src/geo/projection/globe_covering_tiles.ts
NathanMOlson: It is used by unit tests in `globe_covering_tiles.test.ts`.

NathanMOlson:
> It might be an interesting case for the suggested `CalculateTileZoomFunction` to see if it can facilitate the difference between globe an mercator?

Not as currently implemented. Some of the inputs to `CalculateTileZoomFunction` are calculated differently depending on the projection.

src/geo/projection/globe_covering_tiles.ts
NathanMOlson: I added a `const` object. I'm not familiar with the appropriate use of `static` in Typescript so let me know if there is a better practice.

src/geo/projection/covering_tiles.ts
NathanMOlson: "mercator units" are described in `custom_style_layer.ts` line 96. Here's the most relevant part of the description:

```
The spherical mercator coordinate `[0, 0]` represents the top left corner of the mercator world
and `[1, 1]` represents the bottom right corner. A box with identical x, y, and z lengths in mercator
units would be rendered as a cube. {@link MercatorCoordinate.fromLngLat} can be used to project
a `LngLat` to a mercator coordinate.
```

For this function, the units don't actually matter as long as `distanceToTile2D`, `distanceToTileZ`, and `distanceToCenter3D` are all given in the same units. I think it's best to specify units though? Or we could change the inputs to be ratios (`distanceToTile2D/distanceToCenter3D` and `distanceToTileZ/distanceToCenter3D`) but that seems less clear.

HarelM: Units are very important, I just didn't know what it meant. 
If there's an easy way to link to what you wrote here it will be great, or link to a class/type that is relevant. 
Totally optional.

src/source/source_cache.ts
HarelM: Can we add a method to the Tranform called getCoveringTilesDetailsProvider?
Or have transform implement anthor interface - the CoveringTilesDetails and pass that to the covering tiles method? 
Let's do that, I think it's worth the effort.

src/geo/projection/globe_covering_tiles.ts
HarelM: This should be a class that is returned by the relevant transform, the trasform, we can use the transform factory to maintain its lifecycle I guess.

src/source/source_cache.ts
NathanMOlson: I don't understand.

> Can we add a method to the Tranform called getCoveringTilesDetailsProvider?

What is a `CoveringTilesDetailsProvider`?

> Or have transform implement anthor interface - the CoveringTilesDetails and pass that to the covering tiles method?

Does that mean

```
export interface ITransform extends IReadonlyTransform, ITransformMutators {}
```

becomes 

```
export interface ITransform extends IReadonlyTransform, ITransformMutators, CoveringTilesDetails  {}
```

???

src/geo/projection/globe_covering_tiles.ts
NathanMOlson: > we can use the transform factory to maintain its lifecycle I guess

I don't understand. What is the transform factory?

HarelM: Protection factory, my mistake: 
https://github.com/maplibre/maplibre-gl-js/blob/d67fc043bc29778bf951fe4ecafbf2c7c5f93f7d/src/geo/projection/projection_factory.ts

src/source/source_cache.ts
HarelM: Yeah, I'm not sure it's a good idea to add another interface to transform.
CoveringTilesDetails should be renamed to CoveringTilesDetailsProvider.
And maybe get the to right instance using the protection factory.

birkskyum:
Closes #4940

birkskyum:
Closes #4940

src/geo/projection/globe_transform.ts
```diff
@@ -695,11 +692,11 @@ export class GlobeTransform implements ITransform {
             return this._mercatorTransform.coveringTiles(options);
```
NathanMOlson, line 692: @HarelM this seems like another sticking point. `globe` transform has an internal switch between mercator and globe transform, it seems like the `CoveringTilesDetailsProvider` needs to switch at the same time. To me that means making the `CoveringTilesDetailsProvider` a member variable of the transform.

src/source/source_cache.ts
NathanMOlson: See https://github.com/maplibre/maplibre-gl-js/pull/4779/commits/d23cc33ce95952e782c7fcbfc61de18a8e7b30c2

This doesn't use the projection factory because `GlobeTransform` needs to switch between `GlobeCoveringTilesDetailsProvider` and `MercatorCoveringTilesDetailsProvider`.

src/geo/projection/globe_covering_tiles.ts
NathanMOlson: See comment above.

src/geo/projection/globe_transform.ts
```diff
@@ -293,6 +292,7 @@ export class GlobeTransform implements ITransform {
         this._globeness = globeProjectionEnabled ? 1 : 0; // When transform is cloned for use in symbols, `_updateAnimation` function which usually sets this value never gets called.
         this._projectionInstance = globeProjection;
         this._mercatorTransform = new MercatorTransform();
+        this._coveringTilesDetailsProvider = new GlobeCoveringTilesDetailsProvider();
```
HarelM, line 295: Should this be injected instead of instantiated in the constructor?

src/geo/projection/covering_tiles.ts
```diff
@@ -0,0 +1,290 @@
...
+ * @param cameraVerticalFOV - camera vertical field of view, in degrees
+ * @return the desired zoom level for this tile. May not be an integer.
+ */
+export type CalculateTileZoomFunction = (requestedCenterZoom: number,
+    distanceToTile2D: number,
+    distanceToTileZ: number,
+    distanceToCenter3D: number,
+    cameraVerticalFOV: number) => number;
+
+export interface CoveringTilesDetailsProvider {
```
HarelM, line 76: Should this class be in a different file?

src/geo/projection/globe_covering_tiles.ts
```diff
@@ -49,247 +36,139 @@ function distanceToTileWrapX(pointX: number, pointY: number, tileCornerX: number
...
-
-// Returns the wrap value for a given tile, computed so that tiles will remain loaded when crossing the antimeridian.
-function getWrap(centerCoord: MercatorCoordinate, tileX: number, tileSize: number): number {
-    const distanceCurrent = distanceToTileSimple(centerCoord.x, tileX, tileSize);
-    const distanceLeft = distanceToTileSimple(centerCoord.x, tileX - 1.0, tileSize);
-    const distanceRight = distanceToTileSimple(centerCoord.x, tileX + 1.0, tileSize);
-    const distanceSmallest = Math.min(distanceCurrent, distanceLeft, distanceRight);
-    if (distanceSmallest === distanceRight) {
-        return 1;
+export class GlobeCoveringTilesDetailsProvider implements CoveringTilesDetailsProvider {
```
HarelM, line 39: Should this be in a different file?

HarelM: Or is this the main class now in this file?
If so, consider changing the file name.

src/geo/projection/mercator_covering_tiles.ts
```diff
@@ -1,140 +1,44 @@
...
-import {CoveringTilesOptions, IReadonlyTransform} from '../transform_interface';
+import {IReadonlyTransform} from '../transform_interface';
+import {CoveringTilesDetailsProvider, CoveringTilesOptions} from './covering_tiles';
 
-type CoveringTilesResult = {
-    tileID: OverscaledTileID;
-    distanceSq: number;
-    tileDistanceToCamera: number;
-};
+export class MercatorCoveringTilesDetailsProvider implements CoveringTilesDetailsProvider {
```
HarelM, line 8: Same comment as previous one, I think this file needs a rename...?

src/geo/transform_interface.ts
```diff
@@ -284,27 +253,16 @@ export interface IReadonlyTransform extends ITransformGetters {
...
     getVisibleUnwrappedCoordinates(tileID: CanonicalTileID): Array<UnwrappedTileID>;
 
-    /**
-     * Returns a list of tile coordinates that when rendered cover the entire screen at an optimal detail level.
-     * Tiles are ordered by ascending distance from camera.
-     * @param options - Additional options - min & max zoom, terrain presence, etc.
-     * @returns Array of OverscaledTileID. All OverscaledTileID instances are newly created.
-     */
-    coveringTiles(options: CoveringTilesOptions): Array<OverscaledTileID>;
+    getCameraFrustum(): Frustum;
```
HarelM, line 263: Please add some tsdocs comments to better explain what these methods are for.

src/geo/projection/globe_transform.ts
HarelM: Is this still relevant?

HarelM:
This looks good! Seems like there's some issue with the render tests.
I've added minor comments, but otherwise, this can be merged I believe.

src/geo/projection/globe_transform.ts
NathanMOlson: I've resolved this by making the `CoveringTilesDetailsProvider` a member variable of the transform.

HarelM: Thought so, thanks for the info!

src/geo/projection/mercator_covering_tiles.ts
NathanMOlson: done

src/geo/projection/globe_transform.ts
NathanMOlson: I don't think so.

src/geo/projection/covering_tiles.ts
NathanMOlson: Done.

birkskyum:
@NathanMOlson , I think it's necessary to add a `npm cache clear --force` to the github ci to resolve this:

see [StackOverflow](https://stackoverflow.com/questions/77200076/npm-failed-to-remove-some-directories-when-running-npm-install)

```
npm WARN cleanup Failed to remove some directories [
npm WARN cleanup   [
npm WARN cleanup     'D:\\a\\maplibre-gl-js\\maplibre-gl-js\\node_modules\\jest-environment-jsdom\\node_modules\\jsdom',
npm WARN cleanup     [Error: EPERM: operation not permitted, rmdir 'D:\a\maplibre-gl-js\maplibre-gl-js\node_modules\jest-environment-jsdom\node_modules\jsdom\lib'] {
npm WARN cleanup       errno: -404[8](https://github.com/maplibre/maplibre-gl-js/actions/runs/11635159170/job/32403911284?pr=4779#step:5:9),
npm WARN cleanup       code: 'EPERM',
npm WARN cleanup       syscall: 'rmdir',
npm WARN cleanup       path: 'D:\\a\\maplibre-gl-js\\maplibre-gl-js\\node_modules\\jest-environment-jsdom\\node_modules\\jsdom\\lib'
npm WARN cleanup     }
npm WARN cleanup   ],
```

src/geo/projection/covering_tiles_details_provider.ts
```diff
@@ -0,0 +1,32 @@
+import {Aabb} from '../../util/primitives';
+import {MercatorCoordinate} from '../mercator_coordinate';
+import {IReadonlyTransform} from '../transform_interface';
+import {CoveringTilesOptions} from './covering_tiles';
+
+export interface CoveringTilesDetailsProvider {
```
HarelM, line 6: This is `MercatorCoveringTilesDetailsProvider`, right?

src/geo/projection/globe_covering_tiles_details_provider.ts
```diff
@@ -0,0 +1,175 @@
...
+        smallestDistance = Math.min(smallestDistance, distanceToTileWrapX(pointX, pointY, tileCornerX, tileCornerY, tileMercatorSize));
+        // Up
+        smallestDistance = Math.min(smallestDistance, distanceToTileWrapX(pointX, pointY, tileCornerX + halfWorld, -tileCornerY - tileMercatorSize, tileMercatorSize));
+        // Down
+        smallestDistance = Math.min(smallestDistance, distanceToTileWrapX(pointX, pointY, tileCornerX + halfWorld, worldSize + worldSize - tileCornerY - tileMercatorSize, tileMercatorSize));
+
+        return smallestDistance;
+    }
+
+    // Returns the wrap value for a given tile, computed so that tiles will remain loaded when crossing the antimeridian.
```
HarelM, line 67: Use TsDocs style comments, these are visible in IDEs and are generally a good practice.

src/geo/projection/globe_transform.ts
HarelM: Why not?

src/geo/projection/covering_tiles_details_provider.ts
HarelM: My mistake, this is the interface, it has the right name.

src/geo/projection/mercator_covering_tiles_details_provider.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export class MercatorCoveringTilesDetailsProvider implements CoveringTilesDetailsProvider {
+
+    distanceToTile2d(pointX: number, pointY: number, tileID: {x: number; y: number; z: number}, aabb: Aabb): number {
+        const distanceX = aabb.distanceX([pointX, pointY]);
+        const distanceY = aabb.distanceY([pointX, pointY]);
+        return Math.hypot(distanceX, distanceY);
+    }
+
+    // Returns the wrap value for a given tile, computed so that tiles will remain loaded when crossing the antimeridian.
```
HarelM, line 17: Use Tsdocs style comments instead.

src/geo/projection/globe_transform.ts
HarelM: The factory is the one that creates both globe and mercator transform, it can create the relevant object needed to be injected as well.
MetcatorTransfrom also probably needs to be injected in order to receive the right CoveringTiles in the constructor.
This is how inversion of control works basically, and since we are building transform correctly, I think this is a step in the right direction.

HarelM:
I left a few last minor comments.
Looks good otherwise, please check why a render test is failing.

HarelM:
@birkskyum, where do you see the error you described in the CI?

birkskyum:
@HarelM it was in the windows workflow that appear to be running correctly now, after a few re-runs.

birkskyum:
I just want to double check, does this fix the discrepancy in tile loading between a Flat Terrain and No Terrain?

**No Terrain:**

Screenshot 2024-11-02 at 20 26 40

**Flat Terrain (Elevation 0):** (I mocked it)
Screenshot 2024-11-02 at 20 23 16

**Terrain (Appear to be Loading the same OSM tiles as the Flat Terrain):**
Screenshot 2024-11-02 at 20 23 32

birkskyum:
I tried the PR , and it does improve the consistency between the terrain and the mercator, but I don't think I understand why there is a difference between the two, in areas where the terrain has no elevation.

This PR


https://github.com/user-attachments/assets/7d2d01d0-e460-4a94-9049-193e8036d7a9


Main branch

https://github.com/user-attachments/assets/e6cd90aa-ae85-4d83-a37c-db0422f45d74

src/geo/projection/globe_covering_tiles_details_provider.ts
NathanMOlson: Done

src/geo/projection/mercator_covering_tiles_details_provider.ts
NathanMOlson: Done

src/geo/projection/globe_transform.ts
NathanMOlson: `GlobeTransform` always needs a `GlobeCoveringTilesDetailsProvider`, `MercatorTransform` always needs a `MercatorCoveringTilesDetailsProvider`. Constructing the `CoveringTilesDetailsProvider` here in the constructor means it will always be correct.

If we inject a `CoveringTilesDetailsProvider` from outside, we are adding lines of code and increasing the likelihood the the wrong object will be injected. Also, moving the construction of this object elsewhere makes the code harder to follow.

I don't see how injecting the `CoveringTilesDetailsProvider` is a win.

test/integration/render/tests/projection/globe/terrain/line-translate/expected.png
NathanMOlson, line 1: @birkskyum This PR dramatically changes the result of this new render test. To my eye the results look "better" but I'm ignorant about what they should look like. Do you think the new results are acceptable?

birkskyum: I think it's definitely an improvement

birkskyum: I had a PR on the way that added a background that would have hidden this improvement, but I've remove the background again so this difference will be visible

HarelM (approved):
THANKS!!!!

PR #4780, opened 2024-09-30
Camera Roll

NathanMOlson:
Reopening accidentally closed PR: https://github.com/maplibre/maplibre-gl-js/pull/4771

This is the first part of https://github.com/maplibre/maplibre-gl-js/issues/4717: adding camera roll.

In this PR, roll is defined as rotation about the camera boresight. Rotation order is bearing, pitch, roll. `easeTo()` and `flyTo()` are implemented using ~~linear interpolation of Euler angles~~ spherical linear interpolation.

You can play with this feature here, using the controls in the top right corner: https://nathanmolson.github.io/camera_roll

There is some weirdness near +/- 90 degrees roll, but I'd like to get https://github.com/maplibre/maplibre-gl-js/pull/4750 merged before attempting to address it. Near 90 degrees, it appears that tiles are being selected with a much lower zoom level than they should be.

You can also play with camera roll on the globe projection here: https://nathanmolson.github.io/globe_roll/


ssokol:
I'm not qualified to comment on the details of the code, but in terms of meeting the requirements for the bounty, this is spot on. As far as I am concerned, the requirement is met.

src/geo/projection/mercator_transform.ts
```diff
@@ -531,7 +537,8 @@ export class MercatorTransform implements ITransform {
         // 1 Z unit is equivalent to 1 horizontal px at the center of the map
         // (the distance between[width/2, height/2] and [width/2 + 1, height/2])
         const groundAngle = Math.PI / 2 + this._helper._pitch;
-        const fovAboveCenter = this._helper._fov * (0.5 + offset.y / this._helper._height);
+        const zfov = this._helper._fov * (Math.abs(Math.cos(this._helper._roll)) * this._helper._height + Math.abs(Math.sin(this._helper._roll)) * this._helper._width) / this._helper._height;
```
HarelM, line 540: Can this use getters instead of underscore fields?

src/geo/transform_helper.ts
```diff
@@ -164,6 +169,7 @@ export class TransformHelper implements ITransformGetters {
         this._angle = -thatI.bearing * Math.PI / 180;
         this._fov = thatI.fov * Math.PI / 180;
         this._pitch = thatI.pitch * Math.PI / 180;
+        this._roll = thatI.roll * Math.PI / 180;
```
HarelM, line 172: This is strange, for other fields as well, not just roll... I would expect a straight assignment without angle conversion...

HarelM:
Overall looks good! 
Can you add a render test that shows both straight text and tilted horizon?

src/geo/projection/mercator_transform.ts
NathanMOlson: Sure, I'll do that.

Getters are in degrees, underscore fields are in radians.

I was following a patter I saw in other places in the code where underscore fields are used for trig functions, presumably to avoid a call to `degreesToRadians()`.

src/geo/transform_helper.ts
NathanMOlson: Getters are in degrees, underscore fields are in radians. I haven't changed that in this PR, but it sure introduces a lot of places for units errors.

NathanMOlson:
> Overall looks good! Can you add a render test that shows both straight text and tilted horizon?

Will do.

src/geo/projection/mercator_transform.ts
```diff
@@ -568,6 +575,7 @@ export class MercatorTransform implements ITransform {
 
         mat4.scale(m, m, [1, -1, 1]);
         mat4.translate(m, m, [0, 0, -this._cameraToCenterDistance]);
+        mat4.rotateZ(m, m, -this._helper._roll);
```
NathanMOlson, line 578: @HarelM Do you want this changed to use a getter also? what about the following 2 unchanged lines?

HarelM: I honestly don't know. 
Maybe the internal fields should be renamed to rollInRadians so that we'd know from reading the code what to expect? 
The public API doesn't need to change to facilitate that...

HarelM: Or we can add getters only that are without the conversation? IDK...
I don't like all this undocumented know how...

src/geo/projection/mercator_transform.ts
NathanMOlson: Done, in this particular spot.

src/geo/projection/mercator_transform.ts
NathanMOlson: Let me know. I can change the internal field names if that's what you want, but if I don't know what to do my strategy is "don't change it" :)

NathanMOlson: OK, I changed the private fields from `_roll`, `_pitch`, ... to `_rollInRadians`, ....
I added direct getters for these fields in the `TransformHelper`, as well as in the `TransformInterface`.
I changed uses of these fields to use the getter instead of going though `_helper` or converting from degrees.
I replaced the ambiguous field `_angle` with `_bearingInRadians`, which is its negative.

HarelM:
CC: @louwers - note that there are tests here that are related to roll which is not supported in native and might cause problems with parity testing. Can you suggest a way to void it? Should all new tests be in a partent folder called `roll` or something?

@NathanMOlson I see that the style that is used in the render test has a `roll` property in the root of the object.
This is in continue to the monthly meeting yerterday.
Do you think it should be added to the root style spec object:
https://maplibre.org/maplibre-style-spec/root/
I would expect that adding it to the root object in the style.json file as you did, without changing the style spec validation, should fail, shouldn't it?

src/geo/projection/mercator_camera_helper.ts
```diff
@@ -149,11 +153,21 @@ export class MercatorCameraHelper implements ICameraHelper {
             if (isZooming) {
                 tr.setZoom(interpolates.number(startZoom, endZoom, k));
             }
-            if (startBearing !== options.bearing) {
-                tr.setBearing(interpolates.number(startBearing, options.bearing, k));
-            }
-            if (startPitch !== options.pitch) {
-                tr.setPitch(interpolates.number(startPitch, options.pitch, k));
+            if (!quat.equals(startRotation, endRotation)) {
```
HarelM, line 156: Is this code identical to the code in globe_camera_helper.ts?
Can we extract it to somewhere if so?

src/geo/projection/mercator_transform.ts
```diff
@@ -619,7 +637,7 @@ export class MercatorTransform implements ITransform {
         // of the transformation so that 0°, 90°, 180°, and 270° rasters are crisp, and adjust the shift so that
         // it is always <= 0.5 pixels.
         const xShift = (this._helper._width % 2) / 2, yShift = (this._helper._height % 2) / 2,
-            angleCos = Math.cos(this._helper._angle), angleSin = Math.sin(this._helper._angle),
+            angleCos = Math.cos(this.bearingInRadians), angleSin = Math.sin(-this.bearingInRadians),
```
HarelM, line 640: Missing a "-" here for the first `cos` method or is it OK since it's a `cos` method?

src/render/program/sky_program.ts
```diff
@@ -8,22 +8,31 @@ import {getMercatorHorizon} from '../../geo/projection/mercator_utils';
...
 });
 
-const skyUniformValues = (sky: Sky, transform: IReadonlyTransform, pixelRatio: number): UniformValues<SkyUniformsType> => ({
-    'u_sky_color': sky.properties.get('sky-color'),
-    'u_horizon_color': sky.properties.get('horizon-color'),
-    'u_horizon': (transform.height / 2 + getMercatorHorizon(transform)) * pixelRatio,
-    'u_sky_horizon_blend': (sky.properties.get('sky-horizon-blend') * transform.height / 2) * pixelRatio,
-});
+const skyUniformValues = (sky: Sky, transform: IReadonlyTransform, pixelRatio: number): UniformValues<SkyUniformsType> => {
+    const cosRoll = Math.cos(transform.roll * Math.PI / 180.0);
```
HarelM, line 25: Can this use the `InRadians` getter?

src/symbol/projection.ts
```diff
@@ -115,16 +120,54 @@ export function getGlCoordMatrix(
...
         }
+        mat4.scale(m, m, [pixelsToTileUnits, pixelsToTileUnits, 1]);
         return m;
     } else {
         return transform.pixelsToClipSpaceMatrix;
     }
 }
 
+export function getTileSkewMatrix(transform: IReadonlyTransform): mat2 {
+    const cosRoll = Math.cos(degreesToRadians(transform.roll));
```
HarelM, line 138: Can we use `rollInRadians`?

src/symbol/projection.ts
```diff
@@ -115,16 +120,54 @@ export function getGlCoordMatrix(
...
+        vec2.scale(vecEast, vecEast, 1 / vecEastLen);
+    }
+
+    const skew = mat2.create();
+
+    skew[0] = vecEast[0];
+    skew[1] = vecEast[1];
+    skew[2] = vecSouth[0];
+    skew[3] = vecSouth[1];
+    return skew;
```
HarelM, line 168: Consider returning two vectors instead of a mat2, as I don't think the mat2 is used as is, and returning 2 vectors might be a bit more readable.

louwers (changes requested):
Changes to the Style Specification should first go through a [design proposal process](https://github.com/maplibre/maplibre-style-spec/issues/new?assignees=&labels=&projects=&template=design_proposal.md&title=) in the MapLibre Style Spec repo.

NathanMOlson:
> @NathanMOlson I see that the style that is used in the render test has a `roll` property in the root of the object.
> This is in continue to the monthly meeting yerterday.
> Do you think it should be added to the root style spec object:
> https://maplibre.org/maplibre-style-spec/root/

I do not require roll in the root style spec object, as I will be continuously setting `roll` using `map.jumpTo()`. However, it is needed for tests, and pitch and bearing are in the root style spec object, so I guess it makes sense for roll to join them. I'll make a design proposal.

> I would expect that adding it to the root object in the style.json file as you did, without changing the style spec validation, should fail, shouldn't it?

I believe it works due to this magic section of code (map.ts line 719):

```
        this.on('style.load', () => {
            if (this.transform.unmodified) {
                this.jumpTo(this.style.stylesheet as any);
            }
        });

File: src/geo/projection/mercator_transform.ts (hunk 5 above)
Comment (NathanMOlson) [reply src/geo/projection/mercator_transform.ts:640]: `cos()` is symmetric: `cos(x) === cos(-x)`

File: src/symbol/projection.ts (hunk 7 above)
Comment (NathanMOlson) [reply src/symbol/projection.ts:138]: Done.

File: src/render/program/sky_program.ts (hunk 6 above)
Comment (NathanMOlson) [reply src/render/program/sky_program.ts:25]: Done.

File: src/symbol/projection.ts (hunk 8 above)
Comment (NathanMOlson) [reply src/symbol/projection.ts:168]: This is used as a `mat2` in projection.ts line 102.

File: src/geo/projection/mercator_camera_helper.ts (hunk 4 above)
Comment (NathanMOlson) [reply src/geo/projection/mercator_camera_helper.ts:156]: Yes. Done.

**Review by louwers — DISMISSED** [review]

**Comment by louwers** [comment]
Understood!

**Review by louwers — COMMENTED** [review]
Understood!

**Comment by HarelM** [comment]
We should change the above code you mention to be a lot more specific, this `as any` is a bad idea, we should create a obejct with the relevant parameters from the stylespec instance, this way setting roll would not be possible without a spec change.

File: src/geo/projection/mercator_camera_helper.ts (hunk 4 above)
Comment (HarelM) [reply src/geo/projection/mercator_camera_helper.ts:156]: Did you forget to push a commit?

File: src/symbol/projection.ts (hunk 8 above)
Comment (HarelM) [reply src/symbol/projection.ts:168]: I think you use this in two different places in a different way.
I think a two points vector is a better API to understand what is happening here and the mat2 can be constructed in the relevant place.

**Comment by HarelM** [comment]
Another thing I just thought about is a dedicated control for this.
All other viewport possible changes can be updated using the `NavigationControl`.
I think there should be some options to allow changing the roll from a dedicated button in the navigation control if the user sets it to true, like the visualize pitch and show compass.
See here:
https://maplibre.org/maplibre-gl-js/docs/API/type-aliases/NavigationControlOptions/
Also, there is a non-interactive map flag that need to support this as well in case there is a control, I think, not sure.

Otherwise, after adding the roll to the spec, updating the spec package here and addressing the above, I think this can be merged.

File: src/geo/projection/mercator_camera_helper.ts (hunk 4 above)
Comment (NathanMOlson) [reply src/geo/projection/mercator_camera_helper.ts:156]: I don't think so. I moved the spherical interpolation code into `updateRotation()`, which is called on the next line.

```if (!quat.equals(startRotation, endRotation))```

is still duplicated but I don't think it's a good candidate for extracting.

File: src/symbol/projection.ts (hunk 8 above)
Comment (NathanMOlson) [reply src/symbol/projection.ts]: Done.

**Comment by NathanMOlson** [comment]
> Otherwise, after adding the roll to the spec, updating the spec package here and addressing the above, I think this can be merged.

@HarelM Is the NavigationControl update required to merge this? I don't intend to ever use the NavigationControl.

**Comment by HarelM** [comment]
I believe so, also an example page with this new functionality, preferably with a control the users can interact with.

**Comment by NathanMOlson** [comment]
OK, Here's a demo of the updated NavigationControl. https://nathanmolson.github.io/camera_roll

1. The compass icon now indicates the roll angle (when `visualizeRoll` is `true`).
2. The user can control the roll angle by pressing `Ctrl` and right-click and dragging.

@HarelM if this is acceptable I will add unit tests and an example page. Or I can extend an existing example (maybe `navigation`, `3d-terrain`, or `set-perspective`?)

File: src/ui/control/navigation_control.ts [hunk 9]
```diff
@@ -24,12 +24,17 @@ type NavigationControlOptions = {
...
+     */
+    visualizeRoll?: boolean;
 };
 
 const defaultOptions: NavigationControlOptions = {
     showCompass: true,
     showZoom: true,
-    visualizePitch: false
+    visualizePitch: false,
+    visualizeRoll: true
```
HarelM, line 37: Should probably be `false` by default...?

src/ui/control/navigation_control.ts
```diff
@@ -93,11 +98,19 @@ export class NavigationControl implements IControl {
     };
 
     _rotateCompassArrow = () => {
-        const rotate = this.options.visualizePitch ?
-            `scale(${1 / Math.pow(Math.cos(this._map.transform.pitch * (Math.PI / 180)), 0.5)}) rotateX(${this._map.transform.pitch}deg) rotateZ(${this._map.transform.angle * (180 / Math.PI)}deg)` :
-            `rotate(${this._map.transform.angle * (180 / Math.PI)}deg)`;
-
-        this._compassIcon.style.transform = rotate;
+        if (this.options.visualizePitch) {
```
HarelM, line 101: consider using if statement and return to reduce indentation and complexity:
```ts
if (this.options.visualizePitch && this.options.visualizeRoll) { 
   ...
   return;
}
if (something else) {
   ...
   return;
}
```

src/ui/handler/mouse.ts
```diff
@@ -90,3 +94,24 @@ export const generateMousePitchHandler = ({enable, clickTolerance, pitchDegreesP
...
+    enable?: boolean;
+}): MouseRollHandler => {
+    const mouseMoveStateManager = new MouseMoveStateManager({
+        checkCorrectEvent: (e: MouseEvent): boolean =>
+            (DOM.mouseButton(e) === RIGHT_BUTTON && e.ctrlKey),
+    });
+    return new DragHandler<DragRollResult, MouseEvent>({
+        clickTolerance,
+        move: (lastPoint: Point, point: Point) =>
+            ({rollDelta: (point.x - lastPoint.x) * rollDegreesPerPixelMoved}),
```
HarelM, line 110: According to what I tested, I think this needs to be in the opposite direction - i.e. when moving the mouse right I expect the right part to be "lowered".
Not sure if it's exactly here or not.

HarelM:
I think the demo look.
Please make sure to merge main and resolve conflicts.
I think the [navigation](https://maplibre.org/maplibre-gl-js/docs/examples/navigation/) example should be altered to add pitch and roll visualization and enable roll there as well.

HarelM:
Please also see if you can update the style-spec package that was just released and remove the `as any` we talked about.

src/ui/control/navigation_control.ts
NathanMOlson: When pitch = 0, roll and bearing are identical. If `visualizeRoll` is false, the compass icon might display differently when the map is displayed exactly the same, but the internal  representation is different. (For example, `pitch=0, bearing=0, roll=45` would show the same map view as `pitch=0, bearing=45, roll=0`, but would show a different compass icon if `visualizeRoll` is false).

For this reason, I think the default value should be `true`. In a typical use case, where roll never deviates from 0, `visualizeRoll` has no impact.

src/ui/handler/mouse.ts
NathanMOlson: I do not think the behavior should change. As currently implemented, at pitch = 0, roll and bearing are the same, and roll control behaves the same as bearing control. Your proposed change would cause identical mouse movements to move bearing and roll in opposite directions, which seems unexpected and confusing.

src/ui/control/navigation_control.ts
NathanMOlson: Done.

src/ui/handler/mouse.ts
HarelM: I see what you mean about mouse movement for bearing.
If you are looking at the map from the top then dragging above the center of the map behaves "opposite", as if the gesture is expecting to be done below the center of the map to feel intuitive.
This is not the case for mobile as you have two fingers which defines the rotation better.
I believe both roll and bearing mouse gestures needs to be changed to be "better".
Something like taking into consideration the point where the gesture started and calculating the angle from the map center or view center for the rotation.
Here's a movie I made, I was not aware that this is the case, it doesn't make any sense...

https://github.com/user-attachments/assets/1b35b4e3-cf90-4b78-819e-b4186a850e84

I don't know if this is out of scope for this PR, but I wouldn't like to pile up on the "bad" behavior.
Since we are still in 5.x pre-release, this can be a breaking change that can be introduced at this point.

src/ui/map.ts
```diff
@@ -711,7 +724,8 @@ export class Map extends Camera {
 
         this.on('style.load', () => {
             if (this.transform.unmodified) {
-                this.jumpTo(this.style.stylesheet as any);
+                const coercedOptions = pick(this.style.stylesheet, ['center', 'zoom', 'bearing', 'pitch', 'roll']) as CameraOptions;
```
HarelM, line 727: Is it possible to make this a bit more strict somehow?
Maybe by creating an object of type `CameraOptions`?
Something like:
```ts
const coercedOptions: CameraOptions = {
  center: this.style.stylesheet.center,
  ...
}
```

src/ui/control/navigation_control.ts
HarelM: I guess since roll is disabled by default then this won't be a problem, right?

HarelM:
Last two comments hopefully before this can get merged.
Thanks a lot for answering my annoying code review comments, I know this can be irritating and I appreciate your openness and patience.

src/ui/map.ts
NathanMOlson: That's what I tried at first. I got a ton of unit test failures and I don't understand how they are related, so I copied a working example from `camera.ts`.

Here's an example of what happens using your approach (there are other failed test suites as well):

```
                const coercedOptions: CameraOptions = {
                    center: new LngLat(this.style.stylesheet.center[0], this.style.stylesheet.center[1]),
                    zoom: this.style.stylesheet.zoom,
                    bearing: this.style.stylesheet.bearing,
                    pitch: this.style.stylesheet.pitch,
                    roll: this.style.stylesheet.roll
                };
```
                
=>

npx jest src/ui/map_tests/map_layer.test.ts --reporters=default
 FAIL   unit  src/ui/map_tests/map_layer.test.ts (75.728 s)
  ✕ #moveLayer (5003 ms)
  ✕ #getLayer (5001 ms)
  ✕ #removeLayer restores Map#loaded() to true (5003 ms)
  #getLayersOrder
    ✕ returns ids of layers in the correct order (5003 ms)
  #setLayoutProperty
    ✕ sets property (5001 ms)
    ✓ throw before loaded (8 ms)
    ✕ fires an error if layer not found (5002 ms)
    ✕ fires a data event (5002 ms)
    ✕ sets visibility on background layer (5002 ms)
    ✕ sets visibility on raster layer (5004 ms)
    ✕ sets visibility on video layer (5004 ms)
    ✕ sets visibility on image layer (5004 ms)
  #getLayoutProperty
    ✕ fires an error if layer not found (5003 ms)
  #setPaintProperty
    ✕ sets property (5002 ms)
    ✕ #3373 paint property should be synchronized with an update (5001 ms)
    ✓ throw before loaded (5 ms)
    ✕ fires an error if layer not found (5002 ms)

  ● #moveLayer

    thrown: "Exceeded timeout of 5000 ms for a test.
    Add a timeout value to this test to increase the timeout, if this is a long-running test. See https://jestjs.io/docs/api#testname-fn-timeout."

      17 | });
      18 |
    > 19 | test('#moveLayer', async () => {
         | ^
      20 |     const map = createMap({
      21 |         style: extend(createStyle(), {
      22 |             sources: {

      at Object.<anonymous> (src/ui/map_tests/map_layer.test.ts:19:1)

src/ui/handler/mouse.ts
NathanMOlson: I think that's out of scope fir this PR.

src/ui/map.ts
NathanMOlson: @HarelM this works: 
```
                let coercedOptions: CameraOptions = {}
                if (this.style.stylesheet.center != undefined) {
                    coercedOptions.center = new LngLat(this.style.stylesheet.center[0], this.style.stylesheet.center[1]);
                }
                if (this.style.stylesheet.zoom != undefined) {
                    coercedOptions.zoom = this.style.stylesheet.zoom;
                }
                if (this.style.stylesheet.bearing != undefined) {
                    coercedOptions.bearing = this.style.stylesheet.bearing;
                }
                if (this.style.stylesheet.pitch != undefined) {
                    coercedOptions.pitch = this.style.stylesheet.pitch;
                }
                if (this.style.stylesheet.roll != undefined) {
                    coercedOptions.roll = this.style.stylesheet.roll;
                }
```

Do you like this better than the current `pick()`? Or do you have any idea how to turn this into something prettier?

src/ui/control/navigation_control.ts
NathanMOlson: Correct.

src/ui/map.ts
HarelM: I've looked into it.
I think pick is doing these if in a shorter why and is typesafe enough.
I would like the `CameraOptions` to be defined before the `pick` but I see the Stylespec's `center` is defined as a `number[]` and the `CameraOptions` is expecting `center` to be `[number, number]`.
I think this might be solvable by adjusting the style-spec package exported types.
What do you think?

HarelM: In any case, this is not a showstopper for this PR, this is nit picking in the nice-to-have realm.
If you feel like improving this bit feel free to open a PR etc.

HarelM (approved):

src/ui/map.ts
NathanMOlson: We should probably figure out what to do about center-point elevation before changing the style-spec package exported types. https://github.com/maplibre/maplibre-style-spec/issues/851

HarelM: These doesn't have to be tied together, we can improve the style spec typings to supports current implementation and change that when needed in the future.

PR #4794, opened 2024-10-02
use platform instead of jest.done

### LINKED ISSUE #4793: test: Use await/async instead of jest done
This was resolved in 
- #4684 

New cases done() were re-introduced in this PR since:
- #3963 

So there is a bit more cleanup here.

birkskyum:
Small cleanup after the globe merge.

Close #4793

### DIFF
--- a/src/ui/camera.test.ts
+++ b/src/ui/camera.test.ts
@@ -2659,7 +2659,7 @@ describe('#easeTo globe projection', () => {
             camera.easeTo({center: [100, 0], duration: 0});
         });
 
-        test('pans eastward across the antimeridian', done => {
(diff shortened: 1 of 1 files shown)


HarelM (approved):
Is there a lint rule we can use to enforce this somehow?

birkskyum:
there is a no-done-callback rule [here](https://github.com/jest-community/eslint-plugin-jest/blob/main/docs/rules/no-done-callback.md), but I haven't managed to make it take effect, but it would be helpful.

Also the jest done() doesn't exist in vitest and other alternatives, so that would be a very definitive way to guard against it too.

PR #4822, opened 2024-10-11
Add CORS error detection and convert the TypeError to AJAXError

jonathanlurie:
## Launch Checklist

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with a compliant license) - if you are not sure about this, please ask!
 - [ ] Link to related issues.
 - [ ] Write tests for all new functionality. **Read bellow**
 - [x] Post benchmark scores.

## Description
This PR addresses how failed HTTP requests due to CORS policy are handled. So far, if a request fails because of CORS, it does not return a response, instead, it makes the browser throw a generic TypeError ( [see here](https://github.com/maplibre/maplibre-gl-js/blob/v5.0.0-pre.1/src/util/ajax.ts#L158) ). As a result, a map instance subscribing to the `"error"` event would catch a generic error that does not include the faulty URL. It is better to handle a CORS error with an AJAXError, that contains information about the URL and the cause of the actual problem.  

This is especially useful in the context of loading a style.json so that if a given style is not reachable because of CORS, it is possible to catch this error with the same efficiency as if fetching a style.json would yield an HTTP 5xx status code, since in both cases the request would fails due to server configuration.

## Before / After
With this minimalist snippet:
```js
  container: 'map',
  style: "https://example.com/style.json",
});

map.on("error", (err) => {
  console.log(">>> ", err);
}) 
```
Here is the console output,
- **BEFORE:**
Screenshot 2024-10-11 at 09 59 01

- **AFTER:**
Screenshot 2024-10-11 at 10 03 41

## Test
I tried adding another test to `src/util/ajax.test.ts` but it is impossible to simulate a CORS with the `Nise` package's `FakeServer`, since a CORS error is not part of the server response but is due to the browser blocking it.

## Other info
- This makes no change to public API.
- this does not affect performance


HarelM:
Thanks for taking the time to report this issue!
Can this be tested using integration tests in browser or unit tests?

src/util/ajax.ts
```diff
@@ -295,3 +310,27 @@ export const getVideo = (urls: Array<string>): Promise<HTMLVideoElement> => {
...
+ * Check whether the provided error is due to CORS policy
+ */
+function isCorsError(error: unknown): boolean {
+    if (error instanceof TypeError) {
+        // Common CORS error messages can include:
+        // - "Failed to fetch"
+        // - "Network request failed"
+        // - "CORS error"
+        // - "Cross-Origin Request Blocked"
+        const corsErrorMessages = [
```
neodescis, line 324: Two of these are pretty generic... are we certain they are limited to CORS?

jonathanlurie: You're actually right, the message "Failed to fetch" can occur when there is a DNS issue and/or if the URL is malformed or invalid. I believe those case should also be dealt with the same way and still should throw an AJAXError, but I will adjust the messages and comments

jonathanlurie: This is now updated.

jonathanlurie:
> Thanks for taking the time to report this issue! Can this be tested using integration tests in browser or unit tests?

This is quite challenging to test since the fake server (`nise`) cannot simulate CORS issue or DNS resolution failures. I haven't find a way to test that yet.

src/util/ajax.ts
```diff
@@ -155,7 +155,22 @@ async function makeFetchRequest(requestParameters: RequestParameters, abortContr
...
 
-    const response = await fetch(request);
+    // When the error is due to CORS policy, DNS issue or malformed URL, the fetch call does not resolve but throws a generic TypeError instead.
+    // It is preferable to throw an AJAXError so that the Map event "error" can catch it and still have
+    // access to the faulty url.
+    let response: Response;
+    try {
+        response = await fetch(request);
+    } catch (e) {
+        if (isNetworkError(e)) {
```
HarelM, line 165: Why do we care if this is a network error then? If the missing information is that we want to get the url for the relevant fetch behavior, why not simply wrap any exception here with AJAXError?

HarelM:
Yes, all the tests set `fetch to null`, so this can't be really tested.

src/util/ajax.ts
jonathanlurie: Yes, that's a good point, we could do that. I am not aware of all the possible scenarios in which `fetch` would throw, but if it does, that's fair to call it an AJAXError regardless of the reason and to make the faulty URL part of it.

jonathanlurie:
For some weird reason, this PR is now coming with other things from the main branch. Closing it and reopening it clean.

jonathanlurie:
Clean version here https://github.com/maplibre/maplibre-gl-js/pull/4931 , sorry for that!

PR #4825, opened 2024-10-13
Improve terrain rendering when using globe projection

### LINKED ISSUE #4792: tiles are not displayed correctly when both terrain and the new globe projection are used
If you use the new `globe` projection, things display properly until you enable `terrain`.
A `hillshade` layer from the same `raster-dem` source works, but not the `terrain` option.

This is using the 5.0.0-pre.1 alpha release.

https://codepen.io/programingjd/pen/jOgWepY

You can toggle the terrain from the checkbox and see how simply resetting terrain to null makes the raster tiles display correctly.

This is not for raster tiles only, the same thing happens with vector tiles.

ibesora:
This PR fixes #4792 
When terrain is used we first render all layers on an offscreen framebuffer and then use it as a texture on the terrain.
When using globe in addition to terrain, the wrong matrix was being used so tiles were projected in the wrong place in the offscreen framebuffer. This fixes it on raster layers by using the mercator fallback matrix when rendering to the offscreen buffer. I'll tackle vector layers in another PR.
Note that there can be further improvements on terrain by using an orthographic projection when rendering to the offscreen buffer but I'll tackle that in another PR.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


src/geo/transform_interface.ts
```diff
@@ -423,8 +423,10 @@ export interface IReadonlyTransform extends ITransformGetters {
      * Generates a `ProjectionData` instance to be used while rendering the supplied tile.
      * @param overscaledTileID - The ID of the current tile.
      * @param aligned - Set to true if a pixel-aligned matrix should be used, if possible (mostly used for raster tiles under mercator projection).
+     * @param ignoreTerrainMatrix - Set to true if the terrain matrix should be ignored.
+     * @param ignoreGlobeMatrix - Set to true if the globe projection matrix should be ignored.
      */
-    getProjectionData(overscaledTileID: OverscaledTileID, aligned?: boolean, ignoreTerrainMatrix?: boolean): ProjectionData;
+    getProjectionData(overscaledTileID: OverscaledTileID, aligned?: boolean, ignoreTerrainMatrix?: boolean, ignoreGlobeMatrix?: boolean): ProjectionData;
```
HarelM, line 429: Looks like a lot of parameter, consider using a type here instead of all these parameters...

src/render/render_to_texture.ts
```diff
@@ -177,13 +178,14 @@ export class RenderToTexture {
...
-                    const coords = layer.source ? this._coordsDescendingInv[layer.source][tile.tileID.key] : [tile.tileID];
+                    const coords = layer.source ? this._coordsAscending[layer.source][tile.tileID.key] : [tile.tileID];
                     painter.context.viewport.set([0, 0, obj.fbo.width, obj.fbo.height]);
                     painter._renderTileClippingMasks(layer, coords, true);
                     painter.renderLayer(painter, painter.style.sourceCaches[layer.source], layer, coords);
-                    if (layer.source) tile.rttCoords[layer.source] = this._coordsDescendingInvStr[layer.source][tile.tileID.key];
+                    if (layer.source) tile.rttCoords[layer.source] = this._coordsAscendingStr[layer.source][tile.tileID.key];
                 }
             }
+            painter.isRenderingToTexture = false;
```
HarelM, line 188: I'm not sure this is the only place that this needs to be set, and thus creates a potential risk to "forget" to set it back to false. 
Is there another solution to this? Maybe pass it to the painter somehow instead of "globally" setting it?

HarelM:
Don't forget to add a few render tests to make sure this behavior does not break in the future...

src/geo/transform_interface.ts
ibesora: I've added a type to encapsulate all those parameters and changed where those were used.

HarelM (approved):
Consider adding a unit test maybe, or another render test that show the issue better.
Otherwise, this is approved.

ibesora:
> I'll tackle vector layers in another PR.

This was merged via https://github.com/maplibre/maplibre-gl-js/pull/4977

PR #4838, opened 2024-10-15
Globe custom layers: draw globe into depth buffer

### LINKED ISSUE #4817: 3d models are not correctly hidden using the globe view
When using a 3d model on the globe view, the model is not correctly occluded when it goes to the back side of the globe.

https://github.com/user-attachments/assets/825a4dd4-0564-4205-8fe8-15c54aa41ee5

**maplibre-gl-js version**: 5.0.0-pre.1

**browser**: Chrome

### Steps to Trigger Behavior

Using the globe-3d-model example
 1. Move the camera to make the 3d model get behind the horizon

### Expected Behavior
The 3d model should be occluded

### Actual Behavior
You can see the 3d model right through the globe

kubapelc:
This PR fixes #4817 

A 3D model occluded by the globe

Fill extrusion and 3D custom layer now use a depth buffer with globe rendered into it for occlusion. All layer blending should work just like it does with fill-extrusion without globe. Removed ray-sphere intersection test from fill extrusion pixel shader, since regular depth testing is used now.

All below is OUTDATED, only left for documentation.

> A new render pass is added: "3d". It executes after all the other passes ("offscreen", "opaque", "translucent"). When globe is enabled, the globe is drawn into the depth buffer before this "3d" pass. Custom layers that are marked as being 3D then use this render pass. This pass needs to be the last pass, since it replaces the depth buffer contents that is used by other layers for optimization.
> 
> This also means that other layers cannot be blended on top of custom 3D layers (only other custom 3D layers can).
> 
> Fill-extrusion does not use this pass at the moment, since other layers can be blended on top of it. So fill-extrusion still does ray-sphere intersection in the shader.
> 
> There are definitely ways to make both fill extrusion and 3D custom layers use the depth buffer and still allow other layers to be drawn on top, but I think those would be unnecessarily complicated and would not be compatible with the current optimization where opaque pixels do not get overdrawn with underlying translucent layers. Fast drawing of regular 2D maps should take priority, IMO.
> 
> Note that the model must be drawn on top of all other layers, including symbols, in order to use globe's depth buffer:
> 
> 
> 
> I'm not sure if there is a good solution to this, or if this even had a good solution before the globe. I guess symbols could either always be overdrawn by the model or the model could always be overdrawn by the symbols...


src/render/painter.ts
```diff
@@ -412,6 +440,10 @@ export class Painter {
         return new DepthMode(func || this.context.gl.LEQUAL, mask, [depth, depth]);
     }
 
+    depthModeFor3D(): Readonly<DepthMode> {
```
HarelM, line 443: `depthModeFor3D` => `createDepthModeFor3D`?

src/render/painter.ts
```diff
@@ -308,6 +308,34 @@ export class Painter {
         }
     }
 
+    /**
+     * Fills the depth buffer with the geometry of all supplied tiles.
+     * Does not change the color buffer or the stencil buffer.
+     */
+    _renderTilesDepthBuffer() {
```
HarelM, line 315: Does this have a performance or memory impact?
The given case is somewhat of an edge case from my perspective and I wonder if this should be "always on".

HarelM:
Thanks for taking the time to solve this issue.
Can you add a few tests to cover this scenario?

src/render/painter.ts
```diff
@@ -545,6 +577,25 @@ export class Painter {
             drawAtmosphere(this, this.style.sky, this.style.light);
         }
 
+        // 3D pass ===============================================
```
HarelM, line 580: I wonder if these passes should be split into method with the relevant name to avoid these comments...?

src/render/painter.ts
kubapelc: It definitely has some performance impact. It is only enabled for globe, but I will also make sure it only runs when needed by custom layers.

kubapelc:
I've discovered that MapLibre already has a somewhat hidden mechanism for drawing any layer on top of a 3D layer, and I managed to use that for both 3D custom tiles and for fill-extrusion, which simplifies the code a lot and fixes the layering! I will update this PR tomorrow.

src/render/painter.ts
kubapelc: Fixed, I named it `getDepthModeFor3D` and also renamed "depthModeForSublayer" for consistency.

src/render/painter.ts
kubapelc: This is fixed now, globe depth is only ever rendered if any layer needs it.

src/render/painter.ts
kubapelc: I've ended up removing the "3d" pass altogether, thus reverting the method to mostly its previous state.

kubapelc:
I've added a render test that covers this bug.

test/integration/render/run_render_tests.ts
```diff
@@ -423,8 +423,149 @@ async function getImageFromStyle(styleForTest: StyleWithTestData, page: Page): P
             }
         }
 
+        class Tent3DGlobe {
```
HarelM, line 426: Do you think this can move to a new js file and be loaded as a string or as a reference script?
I feel like this is a bit too much for this file.
Although in the past I tried to make it a single file for everything to run as part of it, I think the nullIsland and the other custom layer interfaces deserve a file of their own.
What do you think?

HarelM:
Looks good, added a minor comment.
@louwers do these render test need to be moved to a different folder to avoid problems with native pairity?

test/integration/render/run_render_tests.ts
kubapelc: Yes, I think that is a good idea.

kubapelc: I'm not sure how to do that, simply moving the classes into separate files and importing then fails when running the render tests on "x is not defined", where x is the custom layer class. I've also tried to use factory functions instead of classes, it fails in the same way.

HarelM: This code is running in the browser and that's why it's in the evaluate method.
You'd need to move it to a different file and tell puppeteer to import it, much like how we import the maplibre-gl.js file in the tests.
See here:
https://github.com/maplibre/maplibre-gl-js/blob/da7885aca06f8d679f2f4e2b71f143b66283b793/test/integration/render/run_render_tests.ts#L922

HarelM: It's not a critical change request, I can do it in a later commit if needed.

louwers:
Ideally they are moved somewhere under the globe hierarchy.

kubapelc:
I've moved the new tests to `projection/globe/custom`.

ibesora (commented):
Thanks for tackling this @kubapelc!
I've taken a look at the rendering-side changes and they make sense.

HarelM:
@louwers let me know if this is OK in terms of location.
@kubapelc please resolve conflicts

I'll merge this shortly.

HarelM (approved):

HarelM:
CC: @ibesora - feel free to review and merge if this is OK by you.

louwers:
Thanks!

ibesora (approved):
This looks good from my end.
Approving and merging

PR #4851, opened 2024-10-17
Pitch > 90 degrees

### LINKED ISSUE #3683: Change altitude of look at point?
## User Story
As a developer where I am making 3d buildings with toggleable floors I would benefit from having a way of changing the altitude of camera look at/rotation around point, I think it is called the focal point of the camera.

## Rationale
Because having a 3d map with toggleable floors, either tough extrusions or three js, it would be nice to be able to change the altitude of the camera look at point. I know this is possible today because of the terrain and when you have different altitudes in the terrain this works. I would appreciate if someone who is more familiar with the codebase could help me find where this calculations of the altitude happens and maybe I can work to implement something myself from there. 

## Impact
Would make 3d worlds more enjoyable and easier to navigate for everyone!

NathanMOlson:
Allow pitch angle > 90 degrees. https://github.com/maplibre/maplibre-gl-js/issues/4717

Fixes #3683

Changes:

1. Add `centerClampedToGround` configuration variable. The default is true, which maintains previous behavior. To use pitch angles higher than 90 degrees, this must be set to true, and the elevation must be provided. The elevation must be floating in space above the ground to keep the camera from going below ground.
2. Add `elevation` to `JumpToOptions`. This is needed to set camera pitch > 90 degrees, as the center point must be above the camera, which must be above the terrain.
3. Extend `maxPitch` to 180.
4. Change `recalculateZoom()` to keep the camera position constant. This means that it now adjusts the center point, in addition to `zoom` and `elevation`. This is important to keep the camera from jumping when terrain is updated behind the scenes. This is also the biggest change and the one with the most potential for unwanted ripples.

demo: https://nathanmolson.github.io/camera-centric
image

TODO
 - [ ] Figure out whether `elevation` needs to be added to style spec. https://github.com/maplibre/maplibre-style-spec/issues/851

 - [X] Link to related issues.
 
 
 Additional details about the center point elevation:
The `Transform` variables `center`, `elevation`, `zoom`, `pitch`, `bearing`, and `fov` control the location of the camera indirectly.

 `elevation` sets the height of the "center point" above sea level. In the typical use case (`centerClampedToGround = true`), the library modifies `elevation` in an attempt to keep the center point always on the terrain (or 0 MSL if no terrain is enabled).

`zoom` sets the distance from the center point to the camera (in conjunction with `fovInRadians`, which is currently hardcoded).

Together, `zoom`,  `elevation`, and `pitch` set the altitude of the camera.

image

To allow pitch > 90, the "center point" must be placed off of the ground. This will allow the camera to stay above the ground when it pitches above 90. This requires setting `centerClampedToGround = false`

image

The same math applies whether the center point is on terrain or not, and whether the camera is above or below the ground: 

image
image

In most cases, having the camera underground is undesirable.

To help users position the camera, a new function `calculateCameraOptionsFromCameraLngLatAltRotation()` has been added to `Camera`.


HarelM:
Is it possible to move some parts of this to a plugin? Or maybe place that in an example instead?
I would like to keep changes to the public API as minimal as possible for what I believe is a niche requirement, if possible.

NathanMOlson:
> Is it possible to move some parts of this to a plugin? Or maybe place that in an example instead?
> I would like to keep changes to the public API as minimal as possible for what I believe is a niche requirement, if possible.

Certainly the `jumpToLLA()` function and it's supporting changes can be pulled out to a plugin.

I think the other changes (change in `recalculateZoom()`, increase of `maxPitch` to `180`, and addition of `elevation` to `JumpToOptions`) need to stay. The only change affecting the public API would be the addition of `elevation` to `JumpToOptions`.

HarelM:
How does elevation and zoom correlate? Can one set an elevation value that will change the zoom value?
It might be better to start with a design before jumping to implementation to see that there is an alignment on what should be the solution...

NathanMOlson:
> How does elevation and zoom correlate? Can one set an elevation value that will change the zoom value?

 `elevation` sets the height of the "center point" above sea level. When 3D terrain is used, this is the height of the terrain at the center point, otherwise it is zero. To allow pitch > 90, I intend to position the "center point" floating in space above the terrain. If the center point remains tied to terrain, then pitch > 90 will cause the camera to move under the terrain.

`zoom` sets the distance from the center point to the camera (in conjunction with `fovInRadians`, which is currently hardcoded).

Together, `zoom`,  `elevation`, and `pitch` set the altitude of the camera.

image

To allow pitch > 90, I want to move the "center point" off of the ground. This will allow the camera to stay above the ground when it pitches above 90.

image

> It might be better to start with a design before jumping to implementation to see that there is an alignment on what should be the solution...

Where/how should I present the design?

HarelM:
The design can be presented either as an initial comment of an issue, and initial comment in a PR or a discussion.
I would advise to present it before diving into code changes as it makes changes harder when you "fall in love" with a solution :-)

In any case, the diagram above makes things a lot more clearer.
The first question that comes to mind though is how would this support pitch of 180 degrees? Elevation would be set to infinity?

test/integration/render/run_render_tests.ts
```diff
@@ -535,6 +535,9 @@ async function getImageFromStyle(styleForTest: StyleWithTestData, page: Page): P
                     case 'pauseSource':
                         map.style.sourceCaches[operation[1]].pause();
                         break;
+                    case 'setElevation':
```
HarelM, line 538: I'm not very keen on changing the render test - it seems odd to me...

src/ui/map.ts
```diff
@@ -3194,12 +3205,14 @@ export class Map extends Camera {
...
             this.terrain.sourceCache.update(this.transform, this.terrain);
             this.transform.setMinElevationForCurrentTile(this.terrain.getMinTileElevationForLngLatZoom(this.transform.center, this.transform.tileZoom));
-            if (!this._elevationFreeze) {
+            if (!this._elevationFreeze && this.transform.pitch < 89) {
                 this.transform.setElevation(this.terrain.getElevationForLngLatZoom(this.transform.center, this.transform.tileZoom));
             }
         } else {
             this.transform.setMinElevationForCurrentTile(0);
-            this.transform.setElevation(0);
+            if (this.transform.pitch < 89) {
```
HarelM, line 3213: < 90? <= 89?

src/ui/map.ts
```diff
@@ -212,6 +212,11 @@ export type MapOptions = {
      * @defaultValue [0, 0]
      */
     center?: LngLatLike;
+    /**
+     * The elevation of the initial geographical centerpoint of the map, in meters above sea level. If `elevation` is not specified in the constructor options, MapLibre GL JS will look for it in the map's style object. If it is not specified in the style, either, it will default to `0`.
```
HarelM: This needs to be part of the spec for this comment to be correct I suppose...?

src/geo/projection/globe_transform.ts
```diff
@@ -827,6 +827,10 @@ export class GlobeTransform implements ITransform {
         };
     }
 
+    calculateCenterFromCameraLLA(ll: LngLat, alt: number, bearing?: number, pitch?: number): {center: LngLat; elevation: number; zoom: number} {
```
HarelM, line 830: LLA? I would like to keep TLAs to a minimum if possible.

src/geo/projection/mercator_transform.ts
```diff
@@ -273,31 +273,27 @@ export class MercatorTransform implements ITransform {
...
-        // Since pixelPerMeter = mercatorZfromAltitude(1, center.lat) * worldSize:
-        const requiredWorldSize = requiredPixelPerMeter / mercatorZfromAltitude(1, center.lat);
-        // Since worldSize = this.tileSize * scale:
-        const requiredScale = requiredWorldSize / this.tileSize;
-        const zoom = scaleZoom(requiredScale);
+        // Find the current camera position
+        const origPixelPerMeter = mercatorZfromAltitude(1, this.center.lat) * this.worldSize;
+        const cameraToCenterDistanceMeters = this._cameraToCenterDistance / origPixelPerMeter;
+        const origCenterMerc = MercatorCoordinate.fromLngLat(this.center, this.elevation);
+        const cameraMerc = camMercFromCenterAndRotation(this.center, this.elevation, this.pitch, this.bearing, cameraToCenterDistanceMeters);
```
HarelM, line 287: Merc?

NathanMOlson:
> The first question that comes to mind though is how would this support pitch of 180 degrees? Elevation would be set to infinity?

No need to set elevation to infinity. It just needs to be set high enough that the camera is above the terrain.

image

src/geo/projection/mercator_utils.ts
```diff
@@ -82,7 +82,10 @@ export function unprojectFromWorldCoordinates(worldSize: number, point: Point):
  * @returns Horizon above center in pixels.
  */
 export function getMercatorHorizon(transform: {pitch: number; cameraToCenterDistance: number}): number {
-    return Math.tan(Math.PI / 2 - transform.pitch * Math.PI / 180.0) * transform.cameraToCenterDistance * 0.85;
+    if (transform.pitch > 85.01) {
```
HarelM, line 85: What does 85.01 means?

src/shaders/terrain.fragment.glsl
```diff
@@ -20,10 +20,10 @@ vec4 linearToGamma(vec4 color) {
 
 void main() {
     vec4 surface_color = texture(u_texture, vec2(v_texture_pos.x, 1.0 - v_texture_pos.y));
-    if (v_fog_depth > u_fog_ground_blend) {
+    float blend_color = smoothstep(0.0, 1.0, max((v_fog_depth - u_horizon_fog_blend) / (1.0 - u_horizon_fog_blend), 0.0));
```
HarelM, line 23: Can you elaborate a bit on this change? It's hard for me to fully understand this...

src/ui/map.ts
```diff
@@ -2024,7 +2031,9 @@ export class Map extends Camera {
             if (this.painter.renderToTexture) this.painter.renderToTexture.destruct();
             this.painter.renderToTexture = null;
             this.transform.setMinElevationForCurrentTile(0);
-            this.transform.setElevation(0);
+            if (this.transform.pitch <= 89) {
```
HarelM, line 2034: These changes look a bit messy, is there a way to improve reuse in this file?

src/shaders/terrain.fragment.glsl
NathanMOlson: If fog is transparent (`fog_horizon_color_linear.a == 0`), skip the fog calculations and don't modify the surface color.

This prevents fog from ever doing anything if it is disabled.

src/geo/projection/globe_transform.ts
NathanMOlson: Changed LLA -> LngLatAlt

test/integration/render/run_render_tests.ts
NathanMOlson: Is there another way to set the elevation in a render test, without adding it to the style spec?

src/geo/transform_interface.ts
```diff
@@ -377,10 +377,19 @@ export interface IReadonlyTransform extends ITransformGetters {
...
     getCameraAltitude(): number;
 
+    /**
+     * Given the camera position (lng, lat, alt), calculate the center point and zoom level
+     * @param ll - lng, lat of the camera
+     * @param alt - altitude of the camera above sea level, in meters
+     * @param bearing bearing of the camera, in degrees
+     * @param pitch pitch angle of the camera, in degrees
+     */
+    calculateCenterFromCameraLLA(ll: LngLat, alt: number, bearing?: number, pitch?: number): {center: LngLat; elevation: number; zoom: number};
```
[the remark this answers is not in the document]
NathanMOlson: Yes, `Camera::calculateCameraOptionsFromTo()`. That function takes center lng, lat, alt, and camera lng, lat alt, and outputs the zoom and camera rotation.

This function takes camera lng, lat, alt, and rotation and outputs center point and zoom.

Incidentally, `Camera::calculateCameraOptionsFromTo()` is where I took the `Merc` notation from; it uses `fromMerc` and `toMerc` to represent the camera and center positions in Mercator coordinates.

src/ui/map.ts
NathanMOlson: How about replacing `setElevation()` with `suggestElevation()` (or `setElevationIfCenterPointBelowHorizon()`) which would only set the elevation if the center point is below the horizon?

src/geo/projection/mercator_transform.ts
NathanMOlson: Short for Mercator, following the precedent established in `Camera::calculateCameraOptionsFromTo()`

src/geo/projection/mercator_utils.ts
NathanMOlson: It means that degreesToRadians(radiansToDegrees(85)) > 85, and I didn't want to change the behavior of the existing 85 degree test cases...

I'll clean this up.

test/integration/render/run_render_tests.ts
HarelM: If you have a public API that if part of the map class you can call it, see a few lines below this code.

src/geo/transform_interface.ts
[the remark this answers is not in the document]
HarelM: Shouldn't we expand the method to support this as well instead of adding a new one?
Or is it in the wrong place or something?

src/ui/map.ts
HarelM: The "...IfCenter.." one is self explanatory, so I would go with that.

src/geo/projection/mercator_transform.ts
HarelM: I would change both t use the full word.

HarelM:
Is it possible to clean the git history as well? Sorry for not picking on that, but it makes review harder because I need to expand every time I look at this PR as git thinks there's a lot of commits...

src/geo/transform_interface.ts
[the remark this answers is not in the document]
NathanMOlson: `calculateCenterFromCameraLngLatAlt()` is used by `recalculateZoomAndCenter()`, which is currently a `Transform` function, so it currently needs to stay here.

We could move `recalculateZoomAndCenter()` to `Camera`. It accesses a lot of `Transform` internals though, so I think it's in a good spot.

Regardless of location, I don't think we should combine `calculateCameraOptionsFromTo()` with `calculateCenterFromCameraLngLatAlt()` because they have different inputs and outputs.

HarelM:
Can you elaborate a bit of the center change that is added.
I know that we already suffer from the zoom change at the end of a panning when terrain is on, can you explain what is expected to change here? 
A video or drawing works be great. 
Thanks!!

src/geo/transform_interface.ts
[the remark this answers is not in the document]
HarelM: They are very similar as far as I understand...?
Maybe they can use one another or be a symmetrical counterpart like encode and decode URL...?

src/geo/projection/mercator_transform.ts
NathanMOlson: Done.

test/integration/render/run_render_tests.ts
NathanMOlson: Thanks. Done.

src/geo/projection/mercator_utils.ts
NathanMOlson: OK, cleaned up. The old horizon heuristic had the horizon angle go to 90 degrees when pitch went to 90 degrees, which caused som geometry problems at pitch angles near 90 degrees. The new heuristic works the same up to about 83.5 degrees, than starts to deviate so that the horizon never goes above 89 degrees.

src/ui/map.ts

src/ui/map.ts
NathanMOlson: Done

test/integration/render/tests/terrain/fog/expected.png
NathanMOlson, line 1: There is a hole in the terrain on the right side of the image, causing a bit of sky to be rendered underground. This is caused by missing terrain tiles in the repo. Note this issue affects `main` also, so I'm not sure why we don't see terrain hole in `main` and yet the render test passes...

Assuming this is not an intentional test point, it seems to me that the best solution would be to add the missing terrain tiles to the repo. Where can I find them?

This issue is not caused by the changes in this branch. It has become apparent because the horizon moved slightly, caused by the change to the Mercator horizon heuristic, which affects pitch angles > 83.5 degrees.

I've added a separate bug report here: https://github.com/maplibre/maplibre-gl-js/issues/4859

test/integration/render/tests/sky/basic/expected.png
NathanMOlson, line 1: High-pitch render tests have changed slightly due to the change in Mercator horizon heuristic that affects pitch angles > 83.5.

package.json
```diff
@@ -82,7 +82,7 @@
     "autoprefixer": "^10.4.20",
     "benchmark": "^2.1.4",
     "canvas": "^2.11.2",
-    "cspell": "^8.15.4",
```
HarelM, line 85: I don't think this is related to this PR, please check if you need to update the branch or something...

src/geo/projection/mercator_utils.ts
```diff
@@ -82,7 +82,8 @@ export function unprojectFromWorldCoordinates(worldSize: number, point: Point):
  * @returns Horizon above center in pixels.
  */
 export function getMercatorHorizon(transform: {pitch: number; cameraToCenterDistance: number}): number {
-    return Math.tan(Math.PI / 2 - transform.pitch * Math.PI / 180.0) * transform.cameraToCenterDistance * 0.85;
+    return transform.cameraToCenterDistance * Math.min(Math.tan(degreesToRadians(90 - transform.pitch)) * 0.85,
+        Math.tan(degreesToRadians(89 - transform.pitch)));
```
HarelM, line 86: Why 89 and not 90?
Also, what will happen for pitch = 90 (or 89 in this case)?

src/geo/transform_helper.ts
```diff
@@ -190,6 +190,12 @@ export class TransformHelper implements ITransformGetters {
         this._minElevationForCurrentTile = ele;
     }
 
+    setElevationIfCenterPointBelowHorizon(elevation: number): void {
+        if (this.pitch <= 89.0) {
```
HarelM, line 194: Same as above, I don't understand why 89 was selected and not 90...?

src/geo/transform_interface.ts
```diff
@@ -177,6 +177,15 @@ interface ITransformMutators {
     setCenter(center: LngLat): void;
     setElevation(elevation: number): void;
     setMinElevationForCurrentTile(elevation: number): void;
+
+    /**
+     * Sets the transform's center elevation above sea level, in meters.
+     * Only takes effect if the current center point is below the horizon.
+     * If the current center point is above the horizon, setting the elevation to ground level will cause
+     * the camera to move belo the map.
```
HarelM, line 185:
```suggestion
     * the camera to move below the map.
```

HarelM: Not sure I follow why this should be done only if the center is below the horizon? Can't we define the elevation in all cases and treat it the same way?

src/ui/map.ts
```diff
@@ -2048,7 +2055,7 @@ export class Map extends Camera {
                 } else if (e.dataType === 'source' && e.tile) {
                     if (e.sourceId === options.source && !this._elevationFreeze) {
                         this.transform.setMinElevationForCurrentTile(this.terrain.getMinTileElevationForLngLatZoom(this.transform.center, this.transform.tileZoom));
-                        this.transform.setElevation(this.terrain.getElevationForLngLatZoom(this.transform.center, this.transform.tileZoom));
+                        this.transform.setElevationIfCenterPointBelowHorizon(this.terrain.getElevationForLngLatZoom(this.transform.center, this.transform.tileZoom));
```
HarelM, line 2058: Looks like this is only used in this class, it might be better to create a private method here instead of adding another method to the transform?
I think it would be better to no occupy the transform if we are crossing a certain pitch...?
IDK, just a nit-picking thought...

test/integration/render/tests/sky/pitch95-roll135/style.json
```diff
@@ -0,0 +1,40 @@
+{
```
HarelM, line 1: @louwers - do you have any preferences where to place these render tests so that you'd know they are not supported in native as this is a new feature for gl-js and I don't want the parity tests to fail on this.
@NathanMOlson I think you'd need to update the [pitch](https://maplibre.org/maplibre-style-spec/root/#pitch) definition in the spec to add which version support what.

test/integration/render/tests/sky/pitch95-roll135/style.json
```diff
@@ -0,0 +1,40 @@
...
+    "glyphs": "local://glyphs/{fontstack}/{range}.pbf",
+    "timeout": 60000,
+    "metadata": {
+        "test": {
+            "height": 512,
+            "width": 512,
+            "maxPitch": 95,
+            "operations": [
+                ["setCenterElevation", 10000],
+                ["setPitch", 95],
```
HarelM, line 13: Can't we use the `pitch` and `roll` spec definitions here instead of calling these methods?

src/ui/map.ts
```diff
@@ -371,7 +376,7 @@ const defaultMinPitch = 0;
 const defaultMaxPitch = 60;
 
 // use this variable to check maxPitch for validity
-const maxPitchThreshold = 85;
+const maxPitchThreshold = 180;
```
HarelM, line 379: There's a need to update the comment about max pitch, right?
https://maplibre.org/maplibre-gl-js/docs/API/type-aliases/MapOptions/

src/shaders/terrain.fragment.glsl
HarelM: I would consider moving this to a different PR then to minimize the scope of this PR, if possible.

test/integration/render/tests/terrain/fog/expected.png
HarelM: I'm not sure I know where the original tiles have been coming from, it's probably a long time ago.
Feel free to see if you can find a DEM tile for this place which is license free, or move the scene to demotiles DEM location.

test/integration/render/tests/sky/basic/expected.png
HarelM: I see, thanks for the info, I think it should be noted in the changelog entry that this is a breaking change, see how we highlighted those in previous major releases.

HarelM:
I see that this is still in Draft, please ping me when it's ready for review, although I already added some comments, sorry about that...

test/integration/render/tests/sky/pitch95-roll135/style.json
louwers: Should fail I think.

package.json
NathanMOlson: Reverted.

src/shaders/terrain.fragment.glsl
NathanMOlson: Reverted.

src/ui/map.ts
NathanMOlson: Done.

src/geo/transform_interface.ts
NathanMOlson: This function is called in `Map` in 1) when enabling/disabling terrain, 2) in `_render()`. It is used to set the center point to ground level (terrain height if enabled, 0 if not). If pitch is above the horizon, we should not set the elevation to ground level because this will move the camera below the ground.

src/geo/projection/mercator_utils.ts
NathanMOlson: Here's a graph comparing the old heuristic, which was only valid for pitch <=85, to the new heuristic:

image

In the old heuristic, at 90 degrees pitch, the distance to horizon becomes infinite, which wreaks all sorts of havoc in `_calcMatrices()`. I chose 89 degrees pitch as an arbitrary "close to 90 but less than 90" limit for the horizon angle. tan(89 degrees) = 57.3, so we're getting into angle were very small angle changes result in large distance changes. 89 degrees seemed like a reasonable cutoff to me, but the exact value doesn't really matter. If we made the cutoff 89.25 or higher, it would leave the heuristic unchanged for pitch angles <= 85 degrees, and not require changes to the high-pitch render tests. 

tan(89) = 57.3
tan(89.25) = 76.4
tan(89.5) = 114.6
tan(89.75) = 229.18

Any thoughts on changing the cuttoff point?

src/geo/transform_helper.ts
NathanMOlson: See comments re: `getMercatorHorizon()`. I have replaced the hardcoded 89 with constant `maxMercatorHorizonAngle`.

test/integration/render/tests/terrain/fog/expected.png
NathanMOlson: I think the fact that the scene includes the Dead Sea may be important for reproducing the problem.

test/integration/render/tests/sky/pitch95-roll135/style.json
NathanMOlson: Yes, we can. I've made that change for the `sky` render tests. When terrain is present, the order of setting these matters, so I've left the `terrain/pitch95` test unchanged.

src/ui/map.ts

test/integration/render/tests/sky/basic/expected.png
NathanMOlson: It doesn't have to be a breaking change is we set the horizon angle limit to 89.25 degrees or higher. Is "avoid breaking changes" a goal here? If so how important is it?

NathanMOlson:
> I see that this is still in Draft, please ping me when it's ready for review, although I already added some comments, sorry about that...

No problem, I appreciate the feedback and the earlier the better.

test/integration/render/tests/sky/pitch95-roll135/style.json
HarelM: Why does the order matter?

src/geo/transform_interface.ts
HarelM: But if I want to set the elevation for pitch < 90? Will it be ignored in that case? From API perspective I guess we should be able to "unset" it so that it behaves automatically, or define a value that should be honored. 
I'm not sure how to do that from API perspective, but we should consider the case when someone would like to look at the understand maybe. IDK, just a thought...

src/geo/projection/mercator_utils.ts
HarelM: No need. Just create a variable called horizon cittoff or something and add this explanation to it.

test/integration/render/tests/terrain/fog/expected.png
HarelM: Oh, if it's from the dead sea then I probably brought it from israelhiking.osm.org.il, feel free to bring in a few more if they are needed...

test/integration/render/tests/sky/basic/expected.png
HarelM: We are at 5.x pre release phase, now is the time to break things, assuming we want to break them (on purpose).
If we do break things we need to tell the users that we have done so using the changelog. 
We try not to break things too often - hopefully less than twice a year release a major version.

NathanMOlson: I don't have a good reason to change the behavior at  pitch<= 85 degrees pitch, so I've updated the horizon cutoff to 89.25 degrees, which leaves the behavior unchanged at pitch <=85 degrees.

test/integration/render/tests/sky/pitch95-roll135/style.json
NathanMOlson: > I think you'd need to update the [pitch](https://maplibre.org/maplibre-style-spec/root/#pitch) definition in the spec to add which version support what.

Like this? https://github.com/maplibre/maplibre-style-spec/pull/860

louwers: Oh, right. If you place it under `sky` then it won't show up as a separate disparity to track.

It should be placed somewhere else.

src/geo/projection/mercator_transform.ts
```diff
@@ -273,31 +273,27 @@ export class MercatorTransform implements ITransform {
         return mercatorCoveringTiles(this, options, this._invViewProjMatrix);
     }
 
-    recalculateZoom(terrain: Terrain): void {
-        const origElevation = this.elevation;
-        const origAltitude = Math.cos(this.pitchInRadians) * this._cameraToCenterDistance / this._helper._pixelPerMeter;
-
+    recalculateZoomAndCenter(terrain?: Terrain): void {
```
NathanMOlson, line 276: `recalculateZoomAndCenter()` is called after a map move. The goal is to move the center point onto the terrain without changing the camera position. Here's an example of how it works when pitch = 0:
image
1. The center is on terrain.
2. The map is moved. The center elevation has not changed, but the terrain elevation has. The center is no longer on terrain.
3. `recalculateZoomAndCenter()` is called. The center elevation is set to terrain elevation, and the zoom is updated to keep the camera at the same altitude.

The old version worked fine when pitch = 0, but not when the map was pitched. Here's an example with pitch:
image
1. The center is on terrain.
2. The map is moved. The center elevation has not changed, but the terrain elevation has. The center is no longer on terrain.
3a. (old version, in red) `recalculateZoomAndCenter()` is called. The center elevation is set to terrain elevation, and the zoom is updated to keep the camera at the same altitude. The center position remains the same, causing the camera to move.
3b. (old version, in red) `recalculateZoomAndCenter()` is called. The center point, elevation, and zoom are updated to keep the camera at the same longitude, latitude, and altitude.

The unit test updates make it clear how this changed. The old unit test expects `center` to be unchanged after `recalculateZoomAndCenter()`, the new unit test expects `getCameraLngLat()` to be unchanged after `recalculateZoomAndCenter().

src/geo/projection/mercator_utils.ts
NathanMOlson: Added the variable and explanation.

src/geo/transform_interface.ts
NathanMOlson: The existing behavior is to always clamp the center point to the ground. This definitely won't work if we want pitch > 90 and to have the camera above the ground. As you say, there may be other uses where we want to set the center point to a specific elevation.

I added a flag `centerClampedToGround` (default true) that controls whether the automatic elevation adjustment is active. This adds an extra burden on users who want pitch >90 (they have to set `centerClampedToGround` and provide the center point elevation), but works in more general cases than the "if center point below horizon" check.

src/geo/projection/mercator_transform.ts
HarelM: Thanks! These drawings are great!
My concern is that the center point is a core concept, I tend to think that a lot of logic people built on top of maplibre is dependent on the fact that the center won't change, but I don't have data to support this.
From what you wrote it seems that the following bug still exists:
- https://github.com/maplibre/maplibre-gl-js/issues/3872

Is this the case and this change solves this?

HarelM: Wait, I see that the old code change the center too, right? So this is mainly a change to the name to reflect what it does (it also changes the elevation as far at I can tell, we might want to add that to the method name), right?
It might be interesting to add a developer guide md file under developer-guides with the elevation, terrain, camera movement etc with all these nice drawing that you made for future readers.

src/geo/projection/mercator_utils.ts
```diff
@@ -1,13 +1,21 @@
...
 import {MAX_VALID_LATITUDE, UnwrappedTileIDType, zoomScale} from '../transform_helper';
 import {LngLat} from '../lng_lat';
 import {MercatorCoordinate, mercatorXfromLng, mercatorYfromLat} from '../mercator_coordinate';
 import Point from '@mapbox/point-geometry';
 import type {ProjectionData} from './projection_data';
 
+/*
+* The maximum angle to use for the Mercator horizon. This must be less than 90
+* to prevent errors in `MercatorTransform::_calcMatrices()`. It shouldn't be too close
+* to 90, or the distance to the horizon will become vary large, unnecessarily increasing
```
HarelM, line 14:
```suggestion
* to 90, or the distance to the horizon will become very large, unnecessarily increasing
```

src/geo/transform_interface.ts
```diff
@@ -377,10 +378,19 @@ export interface IReadonlyTransform extends ITransformGetters {
...
     getCameraAltitude(): number;
 
+    /**
+     * Given the camera position (lng, lat, alt), calculate the center point and zoom level
+     * @param ll - lng, lat of the camera
+     * @param alt - altitude of the camera above sea level, in meters
+     * @param bearing - bearing of the camera, in degrees
+     * @param pitch - pitch angle of the camera, in degrees
+     */
+    calculateCenterFromCameraLngLatAlt(ll: LngLat, alt: number, bearing?: number, pitch?: number): {center: LngLat; elevation: number; zoom: number};
```
HarelM, line 392:
```suggestion
    calculateCenterFromCameraLngLatAlt(lngLat: LngLat, alt: number, bearing?: number, pitch?: number): {center: LngLat; elevation: number; zoom: number};
```

src/geo/transform_interface.ts
```diff
@@ -377,10 +378,19 @@ export interface IReadonlyTransform extends ITransformGetters {
...
 
     /**
-     * The altitude of the camera above the center of the map in meters.
+     * The altitude of the camera above the sea level in meters.
      */
     getCameraAltitude(): number;
 
+    /**
+     * Given the camera position (lng, lat, alt), calculate the center point and zoom level
+     * @param ll - lng, lat of the camera
```
HarelM, line 387:
```suggestion
     * @param lngLat - lng, lat of the camera
```

src/geo/transform_interface.ts
```diff
@@ -177,6 +177,7 @@ interface ITransformMutators {
     setCenter(center: LngLat): void;
     setElevation(elevation: number): void;
     setMinElevationForCurrentTile(elevation: number): void;
+
```
HarelM, line 180:
```suggestion
```

src/ui/camera.ts
```diff
@@ -308,6 +312,14 @@ export abstract class Camera extends Evented {
      */
     transformCameraUpdate: CameraUpdateTransformFunction | null;
 
+    /**
+     * If true, the elevation of the center point will automatically be set to the terrain elevation
```
HarelM, line 316: Mark it with `@internal`?

src/ui/camera.ts
```diff
@@ -308,6 +312,14 @@ export abstract class Camera extends Evented {
...
+     * (or zero if terrain is not enabled). If false, the elevation of the center point will default
+     * to sea level and will not automatically update. Defaults to true. Needs to be set to false to
```
HarelM, line 318: What will happen in the dead sea? Where elevation is negative?

src/ui/camera.ts
```diff
@@ -1065,7 +1123,9 @@ export abstract class Camera extends Evented {
 
     _finalizeElevation() {
         this._elevationFreeze = false;
-        this.transform.recalculateZoom(this.terrain);
+        if (this.centerClampedToGround) {
+            this.transform.recalculateZoomAndCenter(this.terrain);
```
HarelM, line 1127: If I want to support both pitch > 90 and terrain, and I want the camera to keep looking at the same location when panning the map and the pitch is under 90, what should I be doing? Do I need extra configurations?

src/ui/map.ts
```diff
@@ -153,12 +153,12 @@ export type MapOptions = {
      */
     maxZoom?: number | null;
     /**
-     * The minimum pitch of the map (0-85). Values greater than 60 degrees are experimental and may result in rendering issues. If you encounter any, please raise an issue with details in the MapLibre project.
+     * The minimum pitch of the map (0-180). Values greater than 60 degrees are experimental and may result in rendering issues. If you encounter any, please raise an issue with details in the MapLibre project.
```
HarelM, line 156: Can we remove the experimental comment now? I think we are basically making this production ready with this PR and the tiles selection, right?

src/geo/projection/mercator_transform.ts
```diff
@@ -523,21 +519,64 @@ export class MercatorTransform implements ITransform {
         return result;
     }
 
+    calculateCenterFromCameraLngLatAlt(ll: LngLat, alt: number, bearing?: number, pitch?: number): {center: LngLat; elevation: number; zoom: number} {
```
HarelM, line 522: This method needs more coverage as it is not fully covered by tests.

src/geo/projection/mercator_transform.ts
```diff
@@ -523,21 +519,64 @@ export class MercatorTransform implements ITransform {
...
         this._helper._pixelPerMeter = mercatorZfromAltitude(1, this.center.lat) * this.worldSize;
 
         // Calculate the camera to sea-level distance in pixel in respect of terrain
-        const cameraToSeaLevelDistance = this._cameraToCenterDistance + this._helper._elevation * this._helper._pixelPerMeter / Math.cos(this.pitchInRadians);
+        const limitedPitchRadians = degreesToRadians(Math.min(this.pitch, maxMercatorHorizonAngle));
+        const cameraToSeaLevelDistance = Math.max(this._cameraToCenterDistance / 2, this._cameraToCenterDistance + this._helper._elevation * this._helper._pixelPerMeter / Math.cos(limitedPitchRadians));
         // In case of negative minimum elevation (e.g. the dead see, under the sea maps) use a lower plane for calculation
-        const minElevation = Math.min(this.elevation, this.minElevationForCurrentTile);
-        const cameraToLowestPointDistance = cameraToSeaLevelDistance - minElevation * this._helper._pixelPerMeter / Math.cos(this.pitchInRadians);
+        const minElevation = Math.min(this.elevation, this.minElevationForCurrentTile, this.getCameraAltitude() - 100);
```
HarelM, line 578: `- 100`?

src/geo/projection/mercator_transform.ts
```diff
@@ -552,13 +591,13 @@ export class MercatorTransform implements ITransform {
         // Find the distance from the center point to the horizon
         const horizon = getMercatorHorizon(this);
         const horizonAngle = Math.atan(horizon / this._cameraToCenterDistance);
-        const fovCenterToHorizon = 2 * horizonAngle * (0.5 + offset.y / (horizon * 2));
+        const fovCenterToHorizon = horizonAngle > degreesToRadians(90 - maxMercatorHorizonAngle) ? 2 * horizonAngle * (0.5 + offset.y / (horizon * 2)) : degreesToRadians(90 - maxMercatorHorizonAngle);
```
HarelM, line 594: `degreesToRadians(90 - maxMercatorHorizonAngle)` can be moved to a variable? Not that I mind much that it is calculated twice (although a good reason by itself), but setting it to a variable will better indicate what it means.

src/geo/projection/mercator_transform.ts
```diff
@@ -552,13 +591,13 @@ export class MercatorTransform implements ITransform {
...
         const topHalfSurfaceDistanceHorizon = Math.sin(fovCenterToHorizon) * lowestPlane / Math.sin(clamp(Math.PI - groundAngle - fovCenterToHorizon, 0.01, Math.PI - 0.01));
 
         // Calculate z distance of the farthest fragment that should be rendered.
         // Add a bit extra to avoid precision problems when a fragment's distance is exactly `furthestDistance`
         const topHalfMinDistance = Math.min(topHalfSurfaceDistance, topHalfSurfaceDistanceHorizon);
-        this._farZ = (Math.cos(Math.PI / 2 - this.pitchInRadians) * topHalfMinDistance + lowestPlane) * 1.01;
+        this._farZ = Math.min(100 * this._cameraToCenterDistance, (Math.cos(Math.PI / 2 - limitedPitchRadians) * topHalfMinDistance + lowestPlane) * 1.01);
```
HarelM, line 600: Is this `100` related to the previous hard coded 100?

src/geo/projection/mercator_transform.ts
```diff
@@ -797,3 +844,14 @@ export class MercatorTransform implements ITransform {
         return this.calculatePosMatrix(tileID);
     }
 }
+
+function cameraMercatorCoordinateFromCenterAndRotation(center: LngLat, elevation: number, pitch: number, bearing: number, distance: number): MercatorCoordinate {
```
HarelM, line 848: for non class methods there's a utils for mercator transform I believe, can this be placed there?

src/geo/projection/mercator_transform.ts
```diff
@@ -523,21 +519,64 @@ export class MercatorTransform implements ITransform {
...
+        } else {
+            distanceToCenterMeters = -altitudeAGL / dzNormalized;
+        }
+
+        let metersPerMercUnit = altitudeFromMercatorZ(1, camMercator.y);
+        let centerMercator: MercatorCoordinate;
+        let dMercator: number;
+        let iter = 0;
+        const maxIter = 10;
+        do {
```
HarelM, line 547: This looks like a "complicated" algorithm, can you better explain how this method is operating?

test/integration/render/tests/sky/pitch95-roll135/style.json
HarelM: @NathanMOlson let's place high (>60) pitch related render test in a folder named `high-pitch` so that we can track it differently that sky as those are different features.

HarelM:
Ok, I think I went over most of the changes in this PR.
It would be great to move the drawings to the initial comment in this PR so that future readers will have a better picture of why this change was introduced and what is the design behind it.
There's also the question if center can be a 3 numbers array instead of two in terms of public API instead of adding an elevation field.
I don't have strong feelings either way, I'll post it on slack and see if I can get some feedback.

src/ui/camera.ts
NathanMOlson: No, you don't need any extra configuration.

`recalculateZoomAndCenter()` only forces the center point to be on terrain. If we don't call it at all, center point, elevation, and zoom are unchanged, so the camera position is unchanged as well.

src/ui/camera.ts
NathanMOlson: If terrain is enabled and `clampCenterToGround` is true, the center elevation with be clamped to the ground and be < 0 at the Dead Sea, just like before this PR.

src/geo/projection/mercator_transform.ts
NathanMOlson: The old code does set the variable `center`, but does not change it in a way that keeps camera position constant. The unit test demonstrates this.

NathanMOlson: > My concern is that the center point is a core concept, I tend to think that a lot of logic people built on top of maplibre is dependent on the fact that the center won't change, but I don't have data to support this.

I agree that center point is a core concept, but in my opinion it changes too much currently. It is constantly trying to move to match the current terrain via cals to `setElevation()` and `recalculateZoomAndCenter()`, which causes all sorts of trouble. This PR gives a mode (`centerClampedToGround == false`) where the center point is more constant.

> From what you wrote it seems that the following bug still exists:
> 
> * [Unwanted pan-jump as requested pan action finishes (3D terrain) #3872](https://github.com/maplibre/maplibre-gl-js/issues/3872)
> 
> Is this the case and this change solves this?

I have not observed that bug. The video in the bug shows a much bigger jump than I would expect to be caused by the old `recalculateZoom()`, which likely only causes visible jumps at very high pitch angles.

src/geo/projection/mercator_transform.ts
NathanMOlson: Moved.

src/ui/map.ts
NathanMOlson: Removed experimental comment.

src/geo/projection/mercator_transform.ts
NathanMOlson: Done.

src/geo/projection/mercator_transform.ts
NathanMOlson: Done.

src/ui/camera.ts
NathanMOlson: Done

src/geo/projection/mercator_transform.ts
NathanMOlson: Gave this magic number a name.

src/geo/projection/mercator_transform.ts
NathanMOlson: Gave this quantity a name.

src/geo/projection/mercator_transform.ts
NathanMOlson: Nope, not at all. I've given `100 * this._cameraToCenterDistance` a name.

src/geo/transform_interface.ts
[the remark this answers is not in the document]
NathanMOlson: I've added `calculateCameraOptionsFromCameraLngLatAltRotation()` to complement the existing `calculateCameraOptionsFromTo()`.

test/integration/render/tests/terrain/fog/expected.png
NathanMOlson: Let's handle this in https://github.com/maplibre/maplibre-gl-js/issues/4859, outside this PR.

src/geo/transform_interface.ts
NathanMOlson: I've added the `center-point` example to demonstrate this use case.

test/integration/render/tests/sky/pitch95-roll135/style.json
NathanMOlson: Done.

test/integration/render/tests/sky/pitch95-roll135/style.json
NathanMOlson: I removed the setPitch call. With it present, we were getting into `_elevateCameraIfInsideTerrain()` which changes the results of the render test.

src/geo/projection/mercator_transform.test.ts
```diff
@@ -449,31 +450,66 @@ describe('transform', () => {
...
             getElevationForLngLatZoom: () => 200,
             pointCoordinate: () => null
         };
-        transform.recalculateZoom(terrain as any);
+        transform.recalculateZoomAndCenter(terrain as any);
         expect(transform.getCameraAltitude()).toBeCloseTo(expectedAltitude, 10);
         expect(transform.zoom).toBe(14);
 
-        // expect new zoom because of elevation change
+        // expect new zoom and center because of elevation change
```
HarelM, line 466: These comments are probably an indication that this test needs to be split into multiple tests.

src/ui/camera.ts
```diff
@@ -368,6 +381,48 @@ export abstract class Camera extends Evented {
...
+
+    /**
+     * Sets the value of `centerClampedToGround`.
+     *
+     * If true, the elevation of the center point will automatically be set to the terrain elevation
+     * (or zero if terrain is not enabled). If false, the elevation of the center point will default
+     * to sea level and will not automatically update. Defaults to true. Needs to be set to false to
+     * keep the camera above ground when pitch \> 90 degrees.
+     */
+    setCenterClampedToGround(centerClampedToGround: boolean): void {
```
HarelM, line 422: If the above is a getter, shouldn't this be a setter? or both not a getter or a setter?
I prefer that this will be a symmetrical if possible.

test/examples/center-point.html
```diff
@@ -0,0 +1,85 @@
...
+
+      let labelLayerId;
+      for (let i = 0; i < layers.length; i++) {
+          if (layers[i].type === 'symbol' && layers[i].layout['text-field']) {
+              labelLayerId = layers[i].id;
+              break;
+          }
+      }
+
+      map.addSource('openmaptiles', {
```
HarelM, line 46: I'm guessing most of the definition of the scene is with this source and the below layer.
Can we remove the style, the code that finds the layer id and simply set the source and the buildings layer to simplify the example?

src/ui/handler_manager.ts
```diff
@@ -527,7 +527,11 @@ export class HandlerManager {
         if (this._map.cameraHelper.useGlobeControls && !tr.isPointOnMapSurface(around)) {
             around = tr.centerPoint;
         }
-        const preZoomAroundLoc = tr.screenPointToLocation(panDelta ? around.sub(panDelta) : around);
+        // If we are rotating about the center point, avoid numerical issues near the horizon by using the transform's
+        // center directly, instead of computing it from the screen point
+        const preZoomAroundLoc = around.distSqr(tr.centerPoint) < 1.0e-2 ?
```
HarelM, line 532: This needs to be covered by a test I believe. Same goes for the other place that has a similar logic.

src/ui/camera.ts
```diff
@@ -1065,7 +1145,9 @@ export abstract class Camera extends Evented {
 
     _finalizeElevation() {
         this._elevationFreeze = false;
-        this.transform.recalculateZoom(this.terrain);
+        if (this.centerClampedToGround) {
```
HarelM, line 1148: Can this `if` be moved inside the `recalculateZoomAndCenter` to avoid sparkling these if around the code? Or it doesn't make sense?
I feel like these if will surely cause a bug in the future for someone not adding it in the right place...

src/ui/camera.ts
HarelM: From the example you provided, in order to be able to see pitch > 85 or 90 I need to set `enterClampedToGround = false`, right?
If this is the case but I still want center to clamp to the ground when the pitch is under 85, what should I be doing? How would I go about and configure this?

src/ui/camera.ts
```diff
@@ -919,13 +978,34 @@ export abstract class Camera extends Evented {
...
-            center: toMerc.toLngLat(),
+            center: toMercator.toLngLat(),
             zoom,
             pitch,
             bearing
         };
     }
 
+    /**
+     * Given a camera position and rotation, calculates zoom and center point and returns them as {@link CameraOptions}.
```
HarelM, line 989: Reading the description of both methods, I still don't understand the difference and when should I use each.
Can the comments improve this to highlight which should be used when, also maybe add an example section like there are for other public methods for both of them to highlight the difference?

src/geo/projection/mercator_transform.ts
```diff
@@ -523,21 +525,71 @@ export class MercatorTransform implements ITransform {
...
+        } else {
+            distanceToCenterMeters = -altitudeAGL / dzNormalized;
+        }
+
+        // The mercator transform scale changes with latitude. At high latitudes, there are more "Merc units" per meter
+        // than at the equator. We treat the center point as our fundamental quantity. This means we want to convert
+        // elevation to Mercator Z using the scale factor at the center point (not the camera point). Since the center point is
+        // initially unknown, we compute it using the scale factor at the camera point. This gives us a better estimate of the
+        // center point scale factor, which we use to recompute the center point. We repeat until the error is very small.
+        // This typically takes about 5 iterations.
```
HarelM, line 553: Is there a way to calculate this without an iterative approach? This seems like a heavy calculation that is being made often...

src/ui/camera.ts
NathanMOlson: I think these should stay out here. I don't think `Transform` should care about `centerClampedToGround`. If somebody tells `Transform` to change elevation, it should do it without caring why.

src/geo/projection/mercator_transform.ts
```diff
@@ -331,9 +328,13 @@ export class MercatorTransform implements ITransform {
                 return coordinate;
             }
         }
+        return this.screenPointToMercatorCoordinateAtZ(p);
+    }
+
+    screenPointToMercatorCoordinateAtZ(p: Point, mercatorZ?: number): MercatorCoordinate {
```
HarelM, line 334: Should this be moved to the utils file as well?

src/ui/camera.ts
NathanMOlson: Made symmetric.

src/geo/projection/mercator_transform.test.ts
NathanMOlson: I split the test.

src/geo/projection/mercator_transform.ts
```diff
@@ -273,36 +273,33 @@ export class MercatorTransform implements ITransform {
...
-        const requiredAltitude = origAltitude + origElevation - elevation;
-        // Since altitude = Math.cos(this._pitchInRadians) * this.cameraToCenterDistance / pixelPerMeter:
-        const requiredPixelPerMeter = Math.cos(this.pitchInRadians) * this._cameraToCenterDistance / requiredAltitude;
-        // Since pixelPerMeter = mercatorZfromAltitude(1, center.lat) * worldSize:
-        const requiredWorldSize = requiredPixelPerMeter / mercatorZfromAltitude(1, center.lat);
-        // Since worldSize = this.tileSize * scale:
-        const requiredScale = requiredWorldSize / this.tileSize;
-        const zoom = scaleZoom(requiredScale);
+        // Find the current camera position
+        const origPixelPerMeter = mercatorZfromAltitude(1, this.center.lat) * this.worldSize;
```
HarelM, line 284: `originalPixelPerMeter`?

test/examples/center-point.html
NathanMOlson: I simplified the example, I believe I removed what you requested.

HarelM: Can you remove the addSource and addLayer and simply define a style with those two?

src/ui/camera.ts
HarelM: Fair enough.

src/geo/projection/mercator_transform.ts
```diff
@@ -552,13 +604,15 @@ export class MercatorTransform implements ITransform {
...
-        const fovCenterToHorizon = 2 * horizonAngle * (0.5 + offset.y / (horizon * 2));
+        const minFovCenterToHorizonRadians = degreesToRadians(90 - maxMercatorHorizonAngle);
+        const fovCenterToHorizon = horizonAngle > minFovCenterToHorizonRadians ? 2 * horizonAngle * (0.5 + offset.y / (horizon * 2)) : minFovCenterToHorizonRadians;
         const topHalfSurfaceDistanceHorizon = Math.sin(fovCenterToHorizon) * lowestPlane / Math.sin(clamp(Math.PI - groundAngle - fovCenterToHorizon, 0.01, Math.PI - 0.01));
 
         // Calculate z distance of the farthest fragment that should be rendered.
         // Add a bit extra to avoid precision problems when a fragment's distance is exactly `furthestDistance`
         const topHalfMinDistance = Math.min(topHalfSurfaceDistance, topHalfSurfaceDistanceHorizon);
-        this._farZ = (Math.cos(Math.PI / 2 - this.pitchInRadians) * topHalfMinDistance + lowestPlane) * 1.01;
+        const maxRenderDistancePixels = 100 * this._cameraToCenterDistance;
```
HarelM, line 614: This is to avoid rendering "too far" when near the horizon, right?

HarelM: Wait, why is this needed if you have a limited pitch anyway?

developer-guides/center-point.md
```diff
@@ -0,0 +1,32 @@
...
+This guide describes how camera position is calculated from the center point, zoom,  and camera rotation.
+The `Transform` variables `center`, `elevation`, `zoom`, `pitch`, `bearing`, and `fov` control the location of the camera indirectly.
+
+ `elevation` sets the height of the "center point" above sea level. In the typical use case (`centerClampedToGround = true`), the library modifies `elevation` in an attempt to keep the center point always on the terrain (or 0 MSL if no terrain is enabled). When `centerClampedToGround = false`, the user provides the elevation of the center point.
+
+`zoom` sets the distance from the center point to the camera (in conjunction with `fovInRadians`, which is currently hardcoded).
+
+Together, `zoom`,  `elevation`, and `pitch` set the altitude of the camera:
+
+Pseudocode (implemented in `MercatorTransform::_calcMatrices()`):
```
HarelM, line 12: I think this part would be better of with typescript code and typescript highlighting.
You can reference the relevant part of the code even.

src/ui/camera.ts
NathanMOlson: I've updated the comments and added examples to both.

test/examples/center-point.html
NathanMOlson: Ah, I get it. Done.

HarelM: Thanks!

src/geo/projection/mercator_transform.ts
NathanMOlson: No. It uses `Transform` member variables `_pixelMatrixInverse` and `worldSize`.

src/geo/projection/mercator_transform.ts
NathanMOlson: It wasn't needed. Removed.

developer-guides/center-point.md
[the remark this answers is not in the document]
NathanMOlson: done.

src/geo/projection/mercator_transform.ts
NathanMOlson: If I'm not missing something, there's 1 `cos`, 1 `atan`, 1 `exp`, and ~20 arithmetic operations per loop. That doesn't seem too heavy to me? I can't think of a way to do it without iteration, but if it's too heavy we could reduce the accuracy threshold, which is currently set at 10^-12 m. This would result in fewer iterations being performed.

src/ui/camera.ts
NathanMOlson: In that case, I think you would need to manually toggle `centerClampedToGround`.

If you want to switch from `centerClampedToGround`, to looking above the horizon, you will need to call `map.setCenterElevation()` anyway. You can call `map.setCenterClampedToGround(false)` immediately prior.

We could automatically call `map.setCenterClampedToGround(false)` inside `map.setCenterElevation()`, but I don't like that, it seems too unexpected. Although the current alternative is also unexpected (you call `map.setCenterElevation()`, but the value you set is overwritten next time the map moves).

In use cases that need to look above the horizon, I think it will typically be much better to specify the camera position and rotation, and use `calculateCameraOptionsFromCameraLngLatAltRotation()` to calculate the center point. Then you never need to worry about elevation at all.

src/ui/handler_manager.ts
NathanMOlson: I added unit tests and the coverage report now shows these as fully covered.

src/ui/camera.ts
HarelM: What I had in mind is the case where you use two fingers swipe to change the pitch, in a case of max pitch 180 I thought about a scenario the this would naturally continue pitching without any extra configuration, I believe this is currently not supported and I wonder what one would need to do in order to support it easily, and if this should be part of this library or a plugin. 
My gut feeling is that it should simply work.
What do you think?

NathanMOlson: I think the best way to support that currently is to `setCenterClampedToGround(false)` and set the center elevation to a fixed altitude above ground, like in the `center-point` demo. This provides continuous pitch past 90 degrees with two-finger swipe. The 2 issues with this are 1) If you pitch far enough, the camera will go underground, 2) You can't zoom in past the the point where the camera would cross the center point. To alleviate #1, you want the center point high. To alleviate #2, you want the center point low. If you can find a compromise point that works for you, this should work.

To make the scenario you describe seamless, I would suggest making the two-finger swipe rotate about the camera position instead of about the center point. This could be implemented as a plugin, which would make use of `calculateCameraOptionsFromCameraLngLatAltRotation()`.

src/geo/projection/mercator_transform.ts
HarelM: Fair enough, let's try this and see how it behaves...

HarelM:
@kubapelc @ibesora would you like to go over this PR as well?
I think it's currently in a good state to be merged.
It is only an opt-in behavior basically that currently can't really be used by current handlers, which I'm not sure I know how I feel about it...
@NathanMOlson does this PR requires the following PR?
- #4779

HarelM (approved):

NathanMOlson:
> @NathanMOlson does this PR requires the following PR?

No, those changes are independent.

developer-guides/center-point.md
```diff
@@ -0,0 +1,32 @@
+# How Camera position is calculated
+
+This guide describes how camera position is calculated from the center point, zoom,  and camera rotation.
```
ibesora, line 3:
```suggestion
This guide describes how camera position is calculated from the center point, zoom, and camera rotation.
```

src/geo/projection/mercator_transform.ts
```diff
@@ -523,21 +525,71 @@ export class MercatorTransform implements ITransform {
         return result;
     }
 
+    calculateCenterFromCameraLngLatAlt(ll: LngLat, alt: number, bearing?: number, pitch?: number): {center: LngLat; elevation: number; zoom: number} {
+        bearing = bearing !== undefined ? bearing : this.bearing;
+        pitch = pitch !== undefined ? pitch : this.pitch;
```
ibesora, line 530: Can we use local variables instead of assigning to the input ones here?

src/style/style.ts
```diff
@@ -747,6 +747,7 @@ export class Style extends Evented {
                 case 'setZoom':
                 case 'setBearing':
                 case 'setPitch':
+                case 'setRoll':
```
ibesora, line 750: Why are we adding this here?

ibesora (commented):
I've left some minor comments around the code, please take a look.
One thing that worries me is that using your example I'm able to go below earth. Shouldn't that be impossible?

Screenshot 2024-10-23 at 10 55 00

src/style/style.ts
HarelM: This was missing from a previous PR I believe.

src/geo/projection/mercator_transform.ts
NathanMOlson: Done.

NathanMOlson:
> One thing that worries me is that using your example I'm able to go below earth. Shouldn't that be impossible?

I don't think it should be fundamentally impossible.

You can make it impossible by setting `maxPitch` to 90, or by setting the right combination of `maxPitch`, `minZoom`, and `elevation` (for example, setting `minZoom` to 14.75 in the example, where `maxPitch=105` and `elevation=541`).

NathanMOlson:
> > One thing that worries me is that using your example I'm able to go below earth. Shouldn't that be impossible?
> 
> I don't think it should be fundamentally impossible.

Maybe a good compromise between "going underground is impossible" and "no fundamental limits on going underground" is "mouse/touch navigation never puts the camera underground".

Here's a proof of concept that ignores mouse/touch commands that result in the camera being underground. It works in the `center-point` demo (which doesn't have terrain).

Changes to `mercator_camera_helper.ts`:

```diff
     handleMapControlsRollPitchBearingZoom(deltas: MapControlsDeltas, tr: ITransform): void {
         if (deltas.bearingDelta) tr.setBearing(tr.bearing + deltas.bearingDelta);
-        if (deltas.pitchDelta) tr.setPitch(tr.pitch + deltas.pitchDelta);
+        if (deltas.pitchDelta) {
+            const originalPitch = tr.pitch;
+            tr.setPitch(tr.pitch + deltas.pitchDelta);
+            if (tr.getCameraAltitude() < 0) {
+                tr.setPitch(originalPitch);
+            }
+        }
         if (deltas.rollDelta) tr.setRoll(tr.roll + deltas.rollDelta);
-        if (deltas.zoomDelta) tr.setZoom(tr.zoom + deltas.zoomDelta);
+        if (deltas.zoomDelta) {
+            const originalZoom = tr.zoom;
+            tr.setZoom(tr.zoom + deltas.zoomDelta);
+            if (tr.getCameraAltitude() < 0) {
+                tr.setZoom(originalZoom);
+            }
+        }
     }
```

To fully work, this would need to support terrain, and would probably have some glitches due to the fact that terrain is constantly changing as the resolution of queried tiles changes.  In the terrain case, it might be better to either accept/reject the entire `deltas` struct instead of the individual compoonents.

HarelM:
For terrain the following method is being used to make sure the camera is not inside the terrain:
https://github.com/maplibre/maplibre-gl-js/blob/491ac04f27c6bba1bf4f3e4b15cea0ddffcff16d/src/ui/camera.ts#L1100-L1113
Which is being used by a global transform update for every movement of the camera:
https://github.com/maplibre/maplibre-gl-js/blob/491ac04f27c6bba1bf4f3e4b15cea0ddffcff16d/src/ui/camera.ts#L1122

I'm guessing this should be similar, or the `elevateCameraIfInsideTheTerrain` can be updated to take into account when the pitch is causing the camera to maybe go underground...

NathanMOlson:
We could easily update `_elevateCameraIfInsideTheTerrain()` to handle the no-terrain case.

There are two things I don't like about applying the limit there.

1. It applies to all transform updates, making it impossible to place the camera underground in cases where that is the desired behavior.
2. At that point we have lost the reason _why_ the camera went underground. We just have a transform that puts the camera underground, and we always "fix" it the same way: by raising the camera altitude. This feels strange when the camera entered the ground due to zooming out, but rather than simply limiting the zoom, the correction primarily uses pitch. I would prefer if I try to zoom out to place the camera underground, my zoom change is rejected instead of causing a pitch change.

HarelM:
For me it feels like solving the same issue in different places, so I prefer to have a single point that handles this, it doesn't have to be the current solution.

src/ui/camera.ts
```diff
@@ -1098,9 +1203,9 @@ export abstract class Camera extends Evented {
      * @param tr - The transform to check.
      */
     _elevateCameraIfInsideTerrain(tr: ITransform) : { pitch?: number; zoom?: number } {
-        const cameraLngLat = tr.screenPointToLocation(tr.getCameraPoint());
+        const cameraLngLat = tr.getCameraLngLat();
```
HarelM, line 1206: Is it possible to rapidly return in case of no terrain and "regular" pitch?

NathanMOlson: Yes. It also requires "regular" center elevation.

PR #4852, opened 2024-10-17
reduce calls to _render() caused by jumpTo()

NathanMOlson:
1. Set `fadeDuration` to 0.
2. Don't call `_render()` on `'move'` event
3. On `'moveend'` event, only call `_render()` if all tiles are loaded
4. On `'data'` event, only call `_render()` if all tiles are loaded
5. If `_placementDirty` is true, a re-render will be required. In this case, return from `_render()` early, before the expensive call to `painter.render()`


NathanMOlson:
Oops, I intended to make this PR in my own fork. This is not intended to be merged.

Moved to: https://github.com/NathanMOlson/maplibre-gl-js/pull/2

Issue #4862, opened 2024-10-20
Render test passes even though actual image is different from expected.


NathanMOlson:
The terrain/fog render test passes, even though the actual image is different from the expected image.

Expected:
![expected](https://github.com/user-attachments/assets/e8565e36-72f1-4f08-a2ba-efa93ad2330a)

Actual:
![actual](https://github.com/user-attachments/assets/2d7a93bd-b021-41ee-abd8-8c4478f00ae9)

Actual (drawn over a red background to demonstrate the transparent "hole" near the horizon):
![actual_red_background](https://github.com/user-attachments/assets/b7fc8fb1-a56c-4502-b8d9-dd9297731b4f)

Computed Diff:
![diff](https://github.com/user-attachments/assets/5ab6c2f7-59a1-4702-9653-23aa5c70f69e)

HarelM:
The render test are using an external library to compare images: `pixelmatch` with some configuration.
The configuration is there to prevent false positive test failures.
Feel free to investigate what's causing this...

NathanMOlson:
> Feel free to investigate what's causing this...

The root cause of this bug is that `pixelmatch` does not detect any difference between white and transparent pixels: https://github.com/mapbox/pixelmatch/issues/26

HarelM:
Ahh, in that case we should make sure either not to use white or use a solid non-white background. 
Another approach we can try is contribute the fix to picelmatch, but I don't have high hopes for this route.

NathanMOlson:
I submitted a PR to pixelmatch to fix this issue: https://github.com/mapbox/pixelmatch/pull/142. I think it would be a good idea to incorporate this fix, as it may affect other render tests in our test suite.

HarelM:
Worst case we can `patch-package` it I guess.
Let's wait a few days and see where this lands.

NathanMOlson:
https://github.com/mapbox/pixelmatch/pull/142 has now been merged upstream in pixelmatch repo.

melitele:
@HarelM is there a reason this is still open? The fix for comparison is in `pixelmatch` [7.0.0](https://github.com/mapbox/pixelmatch/commits/v7.0.0/) and `maplibre-gl-js` is using `pixelmatch` [7.1.0](https://github.com/maplibre/maplibre-gl-js/blob/ae6d184767f2e70c144b125948632b3de82451ea/package.json#L106)

HarelM:
Thanks for pinging, no reason, closing.

Issue #4859, opened 2024-10-19
Terrain "hole" rendered when terrain tiles are missing


NathanMOlson:
This issue appears in the `terrain/fog` render test. The render test is passing in `main`, although the actual image is different from expected. Perhaps that is a separate bug?

Expected:
![expected](https://github.com/user-attachments/assets/e8565e36-72f1-4f08-a2ba-efa93ad2330a)

Actual:
![actual](https://github.com/user-attachments/assets/2d7a93bd-b021-41ee-abd8-8c4478f00ae9)

Actual (drawn over a red background to demonstrate the transparent "hole" near the horizon):
![actual_red_background](https://github.com/user-attachments/assets/b7fc8fb1-a56c-4502-b8d9-dd9297731b4f)

Computed Diff:
![diff](https://github.com/user-attachments/assets/5ab6c2f7-59a1-4702-9653-23aa5c70f69e)

This seems to be caused by missing terrain tiles in the render test. When tiles are missing, they are filled in with "empty" tiles. This means that the terrain doesn't line up at the boundaries between tiles. When neighboring tiles are of the same zoom level, `backfillBorder()` is used to force them to match at the edges. This causes some visual discontinuities, but at least it prevents holes. You can see what should happen at the edge of valid terrain data in the official 3D terrain example:
![image](https://github.com/user-attachments/assets/2ab0c3d6-74a8-47d2-bf97-405ef1be8fa1)

I think the problem is that at high pitch angles, neighboring tiles do not share a zoom level, and thus `backfillBorder()` is not being called, leaving a discontinuity between neighboring tiles. It's also possible that "empty" tiles are handles differently than valid "zero" tiles.

HarelM:
All terrain tiles have a border to prevent holes in the terrain, see:
- https://github.com/maplibre/maplibre-gl-js/issues/1602

But for missing tiles I don't know what the behavior is.

Can you clarify the issue here, I'm not sure I understand the problem statement.

NathanMOlson:
Sure. I've removed my thoughts about what is causing this so the problem statement is more clear, and added a separate issue for the problem with render test infrastructure: https://github.com/maplibre/maplibre-gl-js/issues/4862

ibesora:
What are your expectations on this? AFAIK, there's no hole it's just that it's getting rendered as if there were no elevation

NathanMOlson:
[...]

PR #4864, opened 2024-10-20
Fix render test false negative

NathanMOlson:
Fixes https://github.com/maplibre/maplibre-gl-js/issues/4862. Merging this change will cause the `terrain/fog` render test to fail (as it should), due to https://github.com/maplibre/maplibre-gl-js/issues/4859.

The pixelmatch library does not detect a difference between transparent and white. To work around this, the horizon and fog colors have been changed to not be white.

 - [X] Link to related issues.


HarelM (approved):

HarelM:
I think we should not merge a failing render test though, we should update the image or fix the root cause of this test as oart of this PR.

HarelM:
@NathanMOlson any updates on this PR? Can we move it forward somehow?

NathanMOlson:
This PR can be moved forward by fixing #4859. I don't know how to fix #4859.

Also, a better solution than the one in this PR would be to update the pixelmatch library to include the changes here: https://github.com/mapbox/pixelmatch/pull/142. These changes have been approved by the maintainer, but have not been merged because they add processing time. I do not see a way to fix the pixelmatch library without adding processing time. In my opinion, MapLibre would be better served with a version of pixelmatch that detects transparency differences, even if it is slightly slower.

HarelM:
In the past, a similar problem was solved by adding a border around the tiles to stitch different zoom level for tiles when terrain is on.
Is it possible that the border needs to be "bigger" to solve this issue (the tile is very far away).
I'm not sure it's the right fix, but wondering if it works solve the issue...

NathanMOlson:
https://github.com/mapbox/pixelmatch/pull/142 has been merged in the pixelmatch library.

HarelM:
I believe I addressed the change here: 
- #5541 

Is there anything else needed as part of this PR?

NathanMOlson:
@HarelM #5541 updated several render test so that the expected result includes visible defects.

[high-pitch/pitch95](https://github.com/maplibre/maplibre-gl-js/pull/5541/files#diff-95b0ebb4be9f6a1baa81abd3e110c480bdc2d9d9a78595cff8955e09af45f44e) has 3 white pixels near the horizon in the upper left.
[terrain/fog](https://github.com/maplibre/maplibre-gl-js/pull/5541/files#diff-c03216948d286f7a75efe7989059eea03e911e836c77233194e571fc9517f387) has a transparent section near the horizon in the upper right
[terrain/symbol-add-terrain](https://github.com/maplibre/maplibre-gl-js/pull/5541/files#diff-6549580c263746ac6d01bbaa0adb0f173e9af124d5e3456ecbbc0c5a41fe428e) the tile "skirt" is visible

It feels wrong to test against defective results, would it be better to remove/disable these tests? I think the fact that prior to #5541, the expected results didn't have the defects, suggests that these are regressions.

#5541 fixes the problem that I was pointing out with this PR, which is that pixelmatch could pass render tests that should fail. However, it does not fix the defective terrain/fog rendering.

I'm fine with this being closed, or left open as a demonstration of the rendering defect in terrain/fog.

HarelM:
The render tests simply define the current behavior, if there's a fix that changes the results of the render test, we'll update the image, that's fine. 
It's even better to see that if a fix was done the relevant render test works change for the best. 

As for skipped tests, there were a bunch of skipped tests when I started maintaining this library, and I removed all of them as keeping skipped tests doesn't really help anyone. 

If you want to indicate there's a behavior that needs fixing, an issue with a simple reproduction (using the style.json for the render test even) would be a much better approach.

HarelM:
As far as I can tell the following issue is the tracking one:
- #4859

Closing.

PR #4877, opened 2024-10-22
Update package to support multi features

### LINKED ISSUE #3516: Filter can't be applied to type: MultiPolygon
Perhaps this is a feature request, but it seems like an existing function is incomplete.

The scenario is "Add multiple geometries from one GeoJSON source" (https://maplibre.org/maplibre-gl-js/docs/examples/multiple-geometries/) but where a Feature is a MultiPolygon.

When applying styling using a filter to a Feature that is of type = MultiPolygon:

`
map.addLayer({
            'id': 'park-boundary',
            'type': 'fill',
            'source': 'national-park',
            'paint': {
                'fill-color': '#888888',
                'fill-opacity': 0.8
            },
            'filter': ['==', '$type', 'MultiPolygon']
        });
`

an error is thrown:

`Error: layers.park-boundary.filter[2]: expected one of [Point, LineString, Polygon], "MultiPolygon" found`

MultiPolygons are styled correctly when not using the filter.

HarelM:
## Launch Checklist

Fixes #3516
Incorporates the following PR from style-spec repo:
- https://github.com/maplibre/maplibre-style-spec/pull/519
Fixes:
- #3516

This basically updates the spec package.

 - [x] Link to related issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## main
 
 ### ✨ Features and improvements
+- ⚠️ Changed `geometry-type` to identify "Multi-" features ([#4877](https://github.com/maplibre/maplibre-gl-js/pull/4877))
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
         "@mapbox/unitbezier": "^0.0.1",
         "@mapbox/vector-tile": "^1.3.1",
         "@mapbox/whoots-js": "^3.1.0",
-        "@maplibre/maplibre-gl-style-spec": "^20.4.0",
+        "@maplibre/maplibre-gl-style-spec": "^21.0.0",
         "@types/geojson": "^7946.0.14",
         "@types/geojson-vt": "3.2.5",
         "@types/mapbox__point-geometry": "^0.1.4",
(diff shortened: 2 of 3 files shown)


sbachinin (approved):

HarelM:
This was the community decision.

wipfli:
If I remember correctly, the behavior of the ["geometry-type"] expression was changed to align it with what the specification says it does. And the change solved some problems related to that misalignment. So if we were to revert, we would still have the old problems and spec that does not match the implementation.
How would you solve the issues around ["geometry-type"] @hyperknot ?

hactar:
> Please consider reverting this breaking change in 5.0.1, before millions of MapLibre maps break across the internet.

Hmm. Changes that are not backwards compatible always suck, but in order to continue development in an adequate speed, sometimes these must be done. The established way of doing this is increasing the major version number, and warning about this in the documentation/changelog. The old version should still continue to be available for people who can't change immediately (version 4.7.1). All of this was done in this case.

I would hope that the "millions of maps" have their auto-update system set up correctly so that they don't fetch version 5 automatically as is best practice? And if they don't, isn't limiting do version a version below version 5 not a single line change they could do to quickly fix their broken map?

prusswan:
Apparently this causes multi geometries to not display (e.g. protomaps/PMTiles#511). While major version upgrades can be expected to have breaking changes, this may require a more prominent warning elsewhere (I saw #5290, but this came *after* the v5.0.0 tag so people who check by tags would not have seen it).

hyperknot:
About the lack of warnings:
- The official linting tool, `gl-style-migrate` has shown absolutely nothing in versions 19, 20, 21 or even currently, version 22.
- There have been no console warnings either during the lifecycle of 4.x or even now in 5.x

Basically it's almost impossible to figure out why maps broke, without any depreciation warnings displayed to the user.

prusswan:
Also, the poll (maplibre/maplibre-style-spec#536 - not exactly the place most people will go) only had 13 responses, so the results are not really meaningful. Majority of users have yet to consider the issue and even understand the actual impact.

PR #4878, opened 2024-10-22
Ib/remove sky on low zoom globe

### LINKED ISSUE #4853: Sky shouldn't be visible on low zoom in Globe-Mercator projection
When using a globe, and going to a high pitch, the non-globe sky  is shown

Demo:
https://cartes.app/?style=transports&choix+du+style=oui#2.08/66.22/-14.38/47/85

Screenshot 2024-10-18 at 00 21 59

ibesora:
This PR fixes #4853 by using a black background when using the globe view and blending in the sky as you zoom in to a mercator view.

https://github.com/user-attachments/assets/7293d467-0cd1-40e3-acb2-7c8de7a0ccb7

@birkskyum is something like this what you had in mind?

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


birkskyum:
@ibesora  Yes, this is much better! I did find it a bit abrupt to hit the threshold and have all or no sky, as shown in the video, so it might be better with a more gradual transition. If it complicates things a lot that might be better to consider for a future improvement.

https://github.com/user-attachments/assets/b51fd67e-d0e7-4785-97a7-e298fb417876

zerda-ocm:
does "black background" mean, that something like 
`body { background-image: url(starfield.png); }`
wouldn't be visible anymore?

birkskyum:
@zerda-ocm , yes

see here: https://codepen.io/birkskyum-1471370946/pen/mdNwGOo?editors=1010

zerda-ocm:
black_atmos

birkskyum:
> So, kind of still visible, but I get a black atmosphere...

I see. Generally, while the background-image work-ish for simple cases, I can imagine we'll eventually sacrifice it for a in-canvas skybox that'll respond to rotation etc.

https://github.com/user-attachments/assets/0c95f404-67f7-45e3-a8d7-a4fdc4c67383

src/render/draw_sky.ts
```diff
@@ -43,13 +43,13 @@ export function drawSky(painter: Painter, sky: Sky) {
...
     const gl = context.gl;
 
-    const skyUniforms = skyUniformValues(sky, painter.style.map.transform, painter.pixelRatio);
-
     const depthMode = new DepthMode(gl.LEQUAL, DepthMode.ReadWrite, [0, 1]);
     const stencilMode = StencilMode.disabled;
     const colorMode = painter.colorModeForRenderPass();
     const program = painter.useProgram('sky');
 
+    const skyUniforms = skyUniformValues(sky, painter.style.map.transform, painter.pixelRatio);
```
HarelM, line 51: Does moving this actually changes something? If not I would consider reverting this...

src/shaders/sky.fragment.glsl
```diff
@@ -16,4 +17,5 @@ void main() {
             gl_FragColor = u_sky_color;
         }
     }
+    gl_FragColor = mix(gl_FragColor, vec4(vec3(0.0), 1.0), u_sky_blend);
```
HarelM, line 20: I would consider moving this part into the "if"s as I think in some cases the sky is not drawn at all, or not drawn over the entire screen, and so the blend needs to be only on the sky itself...?
I'm not very good with the shaders, but I think there might be issues here.
Also it might be better to maybe project the sky mesh using the projection data somehow? Or doing some calculations to improve the sky-horizon rendering as previously the sky horizon was an easy straight line, but now with globe and roll the code is more complicated I guess...
IDK...

test/integration/render/tests/sky/globe/expected.png
HarelM, line 1: Due to pixel match issues it is advised not to have white parts in the picture, this is a new issue we found out about recently.

HarelM:
I think this looks good in general, but we might consider improving it a bit.
The fix is fairly small, so we should consider merging it as is to improve the current situation and later or find a more complex and full solution, your call @ibesora.

src/render/draw_sky.ts
ibesora: You are right

src/shaders/sky.fragment.glsl
ibesora: > I would consider moving this part into the "if"s as I think in some cases the sky is not drawn at all, or not drawn over the entire screen, and so the blend needs to be only on the sky itself...?

Sky is rendered as a full-screen quad positioned in the background so all the screen pixels should have a color. Fragment shader not outputting a color is an undefined behavior according to spec. It works now because we are rendering tiles on top of pixels where that happens.
But this is specifically about pixels that are not covered by the globe and that are below the horizon. In that case, we want those to be black. 

> improve the sky-horizon rendering as previously the sky horizon was an easy straight line, but now with globe and roll the code is more complicated I guess...

I don't think sky-horizon rendering should work with globe but we'll definitely need to make it work with roll. Will do that in another PR. Does that sound good?

HarelM: Not sure I understand why. I mean when seeing the full globe, yes the sky is not well defined, but when you zoom enough and the horizon is still a bit curved I do think there's a merit to seeing the sky without some black artifact...

ibesora:
> I did find it a bit abrupt to hit the mercator threshold and have all or no sky, as shown in the video, so it might be better with a more gradual transition, or since it's mainly problematic in the horizon maybe it can be hidden with fog. If it complicates things a lot that might be better to consider for a future improvement.

@birkskyum  Yeah, it uses the existing globe-to-mercator animation transition, which lasts 0.5 seconds. I think there needs to be a polish pass on all those things so I'll happily tackle it in another PR.

> It looks like the mercator horizon is quite a lot closer than the globe horizon in the repro case i look.

Yeah, transitioning from globe to mercator doesn't give you exact results and it's really visible on the horizon. I'll try to fix it in another PR.

> does "black background" mean, that something like
body { background-image: url(starfield.png); }
wouldn't be visible anymore?

@zerda-ocm I just changed it so it blends with the background instead of going to black. I think this is a better solution until we support having a starry background ourselves

https://github.com/user-attachments/assets/f98dd559-4232-4278-b4d5-541ceb1cb98e

src/shaders/sky.fragment.glsl
ibesora: I can see that. Do you think the zoom level at which we start showing the sky should be configurable?

HarelM: Since we don't define when globe turns into Mercator (spec-wise) I think we should not burden the user with this definition. 
Is there a definition of camera distance from the ground and pitch where sky is visible (for real life I mean) that we can use? Or a close heuristic for that? 
It not, you can choose an arbitrary value of globeness and use that I guess...

birkskyum (approved):

HarelM:
@ibesora can you updated the failing render test expected image?

ibesora:
> @ibesora can you update the failing render test expected image?

I'm working on it. Unfortunately it's not that easy, the expected image on windows and ubuntu is completely blank and I'm trying to figure out what's happening. I can reproduce that same behaviour on ubuntu when running tests although when running an example, everything seems to work as expected

test/integration/render/tests/sky/globe/style.json
```diff
@@ -0,0 +1,42 @@
...
+  "sprite": "local://sprites/sprite",
+  "glyphs": "local://glyphs/{fontstack}/{range}.pbf",
+  "timeout": 60000,
+  "metadata": {
+      "test": {
+          "height": 512,
+          "width": 512,
+          "maxPitch": 85,
+          "operations": [
+              ["wait"]
```
HarelM, line 12: Try using `idle`, or add some wait with more time to see if it passes the CI. It might be some timing issues...

birkskyum:
> > @ibesora can you update the failing render test expected image?
> 
> I'm working on it. Unfortunately it's not that easy, the expected image on windows and ubuntu is completely blank and I'm trying to figure out what's happening. I can reproduce that same behaviour on ubuntu when running tests although when running an example, everything seems to work as expected

The globe is a bit heavy, so it might be caused by the tests snapshotting too early sometimes (flaky), before the render is complete. It might resolve the issue to increase the time before the snapshow happen to let the render finish.

ibesora:
> @ibesora , it seems like the sky is only rendered from around zoom 12, and the test is zoom 11. Changing the test to zoom 12 makes the sky render in the test as expected, so maybe we can just have two tests to capture the transition, one at zoom 11 (no sky visible) and one at zoom 12 (sky visible)

The intention of the test was to catch the no visible sky when globe is shown, the problem is that in the captures for ubuntu and windows, there's no globe being rendered at all

HarelM:
You can try opening codespaces and see if you can reproduce the issue there to better understand what doesn't work...

ibesora:
> You can try opening codespaces and see if you can reproduce the issue there to better understand what doesn't work...

It's weird because changing the test, positioning the camera in another place made it work. 
I'll dig deeper to see what happens and fix it in another PR

birkskyum:
Sounds good, seems like this PR is ready to be merged then.

birkskyum:
@ibesora  just to clarify, you consider this PR ready to be merged in it's current state, and the additional investigation/fix you mention is another PR against main at a later point?

ibesora:
> @ibesora just to clarify, you consider this PR ready to be merged in it's current state, landing these improvements, and the additional investigation/fix you mention is another PR against main at a later point?

Yup, that’s what I meant, sorry. 
The test failure is not related to this fix, I can reproduce it on `main`

HarelM:
Let's merge this, if needed we can continue fixing issue as version 5 is still in pre-release.

PR #4885, opened 2024-10-23
Add example of Three.js model with shadow

birkskyum:
None of the 3D model examples we have contain a shadow. This example is a more advanced modification of simple Three.js model example, enabling a shadow, to demonstrate that we do support this and to give people a proper starting point for with shadows.

This has been requested quite a few times - the latest slack thread on the topic is [here](https://osmus.slack.com/archives/C03TFH5NE83/p1729649981995159).

Screenshot 2024-10-23 at 17 15 46
Screenshot 2024-10-23 at 17 17 35

PR #4888, opened 2024-10-24
Use ESM and latest three.js 0.169.0

birkskyum:
This updates all the Three.js examples to the latest version 0.169.0

This requires the examples to be updated to ESM / importmap, which is the only distribution of Three.js since 0.160.0.

Resolves the issues mentioned in #4885 :
- remove `map = window.map`
- write "using three.js" instead of "with three.js"
- latest three.js version 0.169.0 instead of 2y old 0.147.0


HarelM (approved):
Thanks!

PR #4890, opened 2024-10-24
Allow raster source tile updates to re-trigger tile request that failed previously

wagewarbler:
Events where tile sources are updated should be handled differently from other types of source updates. Currently if a tile errors, it won't be re-requested presumably on account of not wanting to waste traffic on a previously failed fetch. The fetch may no longer be pointless however if the "source's source" has been updated. In this situation, old tile requests that failed are no longer relevant. It's in effect like a new source in the old source's place.

This PR fixes a bug where fetches that previously failed prevent further fetches with with new tile URLs.

This could likely be applied to vector tiles as well if the approach makes sense!

A map loaded with failed tile requests. Previously a change to the date wouldn't trigger new fetches. With this change in place, new tiles are requested as expected:

[Screencast from 2024-10-24 16-25-07.webm](https://github.com/user-attachments/assets/af66a905-6505-4be6-ab6d-ef762a1cc41e)

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


src/source/source_cache.test.ts
```diff
@@ -532,6 +532,33 @@ describe('SourceCache / Source lifecycle', () => {
 
     });
 
+    test('does reload errored tiles, if event is source data change', () => {
+        const transform = new Transform();
```
wagewarbler, line 536: Am I understanding correctly that there's an error here? This is really just a copy/paste of the previous test with a small event arg change to demonstrate a downstream change in behavior. There shouldn't be any reason for an error on `Transform`

HarelM: Is your branch up to date? There's no more Trasform class...

src/source/source_cache.ts
```diff
@@ -919,7 +921,8 @@ export class SourceCache extends Evented {
         // for sources with mutable data, this event fires when the underlying data
         // to a source is changed. (i.e. GeoJSONSource#setData and ImageSource#serCoordinates)
         if (this._sourceLoaded && !this._paused && e.dataType === 'source' && eventSourceDataType === 'content') {
-            this.reload();
+            const sourceDataChanged = Boolean(e.sourceDataChanged);
```
HarelM, line 924: Is there a problem with the data that it needs casting?

src/source/source_cache.ts
```diff
@@ -253,7 +253,9 @@ export class SourceCache extends Evented {
         this._cache.reset();
 
         for (const i in this._tiles) {
-            if (this._tiles[i].state !== 'errored') this._reloadTile(i, 'reloading');
+            if (sourceDataChanged || this._tiles[i].state !== 'errored') {
```
HarelM, line 256: Are all changes to source require a reload?

src/source/raster_tile_source.ts
```diff
@@ -133,7 +133,9 @@ export class RasterTileSource extends Evented implements Source {
 
         callback();
 
-        this.load();
+        const sourceDataChanged = true;
```
HarelM, line 136: I'm not sure this extra variable is needed, simply call load(true)

HarelM:
Can you better describe the scenario that you would change av raster source definition and expect it to be reloaded? Wouldn't removing the source and adding a different one is a better approach?

src/source/raster_tile_source.ts
wagewarbler: Yep, just a code habit. Prob because I don't love single boolean args for functions. It lacks description. Will update

src/source/source_cache.ts
wagewarbler: The code prior to this PR is written so that any source changes trigger a reload. This PR adds the behavior that amounts to re-requesting errored tiles if the `sourceDataChanged` flag is passed, which in this case only happens when `RasterTileSource > setTiles` is called. I think any other case that doesn't modify the tile source(s) wouldn't warrant attempting to reload previously errored tiles

src/source/source_cache.ts
wagewarbler: Depends on comfort levels with using `undefined` as a boolean value. Of course I know it's treated as `false` in JS/TS, just a question of explicitness preference. This value will frequently come through as `undefined` since it's optional. Since you know the code base best, what's your preference? Leave as `undefined`, double negation (`!!e.sourceDataChanged`), or something else?

wagewarbler:
> Can you better describe the scenario that you would change the raster source definition and expect it to be reloaded? 

Sure. Updating a date (or any attribute) filter for an existing source. For instance going from this:

https://<some-base-url>/tiles/WebMercatorQuad/4/5/5?datetime=2024-**10-03**T16:30:00Z/2024-**10-03**T17:30:00Z

To this:

https://<some-base-url>/tiles/WebMercatorQuad/4/5/5?datetime=2024-**10-04**T16:30:00Z/2024-**10-04**T17:30:00Z

> Wouldn't removing the source and adding a different one is a better approach?

That's a valid question but possibly beyond the scope of this PR. This PR is just attempting to fix behavior that effectively derails a raster tile source from showing any further tiles once it gets a `404` or other error. That said, I think your question is a valid one in big picture kind of way and could potentially spawn further work. Correct me if I'm wrong, but currently this lib is set up so that cross-fading is only possible when navigating within an existing layer source. In other words, zooming in and out or calling `RasterTileSource > setTiles` as we added in #3208. A different approach would be to remove an old layer source and add a new one, but without the cross fading this would be a jerky experience for the user.

HarelM:
Yeah, it might have been that the original design expected the source to be immutable or sometime similar, which we broke with the addition of setTiles.
Other sources are not really immutable, so fixing this probably makes more sense now. 
Thanks for the context!

src/source/source_cache.ts
wagewarbler: Yep, you're right!

HarelM:
Please add a changelog entry and I'll merge this.

HarelM (approved):

HarelM:
Can you resolve conflict? Probably related to changelog file...

PR #4897, opened 2024-10-25
Fixes scale control for globe on zoom out

### LINKED ISSUE #4804: Scale shows wrong numbers when zooming out.
**maplibre-gl-js version**: 5.0.0-pre.1

**browser**: chrome, edge

### Steps to Trigger Behavior

 1. add a scale to the map
```
var scale = new maplibregl.ScaleControl({
  maxWidth: 80,
  unit: "imperial",
});
map.addControl(scale);
```
 2a. move or zoom slightly and observe the scale
 2b. zoom out a lot and observe the scale

### Link to Demonstration

https://jsfiddle.net/zLhtufqk/

### Expected Behavior

Scale shows distance according to scale of the globe

### Actual Behavior

The scale shows for example "300mi", after a move or zoom event (even a minor one), the scale jumps to for example "1000mi". Zooming further out, the scale shows smaller and smaller numbers until the whole earth is only 1 mile wide, Zooming in seem to work fine (presumably, because the projections gets closer to mercator.)

pcardinal:
## Launch Checklist

- #4804

Fixes #4804

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


pcardinal:
**Before modifications:**

image

image

image



**After modification`:**

image

image

image



**ready for review**

pcardinal:
**ready for review**

HarelM:
Can you please add a changelog item?
I'm not sure a test is really required here as you only change the place where scale is being measured...

pcardinal:
Before doing that, I found another little problem to solve

birkskyum:
@pcardinal a problem with this pr, or is it another ticket?

pcardinal:
I do not understand the request from Harel?
What does he mean with this phrase "Can you please add a changelog item?"

birkskyum:
It's just to write a line in the top part changelog.md with a link to this pr

birkskyum:
You can just insert this under the `bug fixes`

`- Fixes scale control for globe on zoom out ([#4897](https://github.com/maplibre/maplibre-gl-js/pull/4897))`

pcardinal:
**For reference**

The Scale Bar on a MapLibre GL JS Globe: An Approximation

While it's true that a globe represents a 3D spherical surface, and a scale bar is typically associated with 2D planar maps, MapLibre GL JS does provide a way to display a scale bar on a globe. However, it's important to understand the limitations of this feature.

Why Display a Scale Bar on a Globe?

Visual Reference: The scale bar serves as a quick visual reference for estimating distances.
User Familiarity: Users are accustomed to seeing scale bars on maps, and they can provide a sense of scale even on a globe.
Compromise: While not perfectly accurate, the scale bar offers a useful approximation.
Interpreting the Scale Bar
The scale bar on a MapLibre GL JS globe should be interpreted as an approximation rather than an exact measurement. Its accuracy is influenced by:

Location on the Globe: Distortions introduced by map projections are more pronounced at the poles than at the equator.
Zoom Level: The scale bar adjusts based on the zoom level.
Limitations of the Scale Bar

Distortions: Distances measured on the screen do not precisely match real-world distances, especially at higher latitudes.
Terrain: The scale bar doesn't account for terrain elevation, which can affect distance calculations in mountainous regions.
In essence, the scale bar on a MapLibre GL JS globe provides a general sense of scale but should not be relied upon for precise measurements. It's a useful visual aid but is subject to the inherent limitations of representing a spherical surface on a 2D screen.

For accurate measurements, it is recommended to use specialized geospatial measurement tools designed for 3D environments.

_( gemini.google.com)_

.gitignore
```diff
@@ -40,3 +40,7 @@ npm-debug.log
 /.vs/
 .*.swp
 junit.xml
+test.html
+package-lock.json
+.vscode/launch.json
+package-lock.json
```
birkskyum, line 46: The package-lock.json shouldn't be in the .gitignore

pcardinal:
Hi Byrk, 
it seems i don't have permission to change the change.log file

Also I "fucked up" my local Git environment and it is possible that I lost my link with the PR. The code has been finalized but I may not be able to update the current PR

HarelM:
Let me know if you need help with this.

pcardinal:
https://github.com/maplibre/maplibre-gl-js/pull/4969#issuecomment-2454901007

package-lock.json
HarelM, line 1: Please revert this file.

pcardinal:
I will give a check about the pitch

package-lock.json
HarelM: You haven't changed any package, so package-lock file should stay the same.

pcardinal: The package-lock.json file changed when I ran "npm install" the first time. I don't know anything I can do about it.

pcardinal: "The package-lock.json shouldn't be in the .gitignore".  package-lock.json is not in .gitignore any more

HarelM: You can download the file from main branch and override yours and check it in to revert the change...

.gitignore
```diff
@@ -40,3 +40,7 @@ npm-debug.log
 /.vs/
 .*.swp
 junit.xml
+test.html
```
HarelM, line 43: This shouldn't be here as well. 
Please revert changes to gitignire file.

HarelM:
I see that the files are OK now, thanks.
Let me know what you find out in regards to pitch.
Better to have it fixed in a single PR although this PR by itself has a value.

pcardinal:
Finally,the value of the pitch has no major impact on the value of the scale display.

HarelM:
Have you tested with pitch > 70? I find this statement odd...

birkskyum:
@HarelM , it's true that the scale only really makes sense when you know where on the screen it's measuring. What @pcardinal effectively say is that reference implementations, such as  [Google Maps](https://www.google.com/maps/@13.0131665,-18.1242011,13217392a,42.8y,0.06t/data=!3m1!1e3?entry=ttu&g_ep=EgoyMDI0MTEwNi4wIKXMDSoASAFQAw%3D%3D), completely disregard both the globe transform **and** the pitch in the scale control in the lower right of the screen.

pcardinal:
"Have you tested with pitch > 70? I find this statement odd..."

I have tested with pitch values 0, 85 and 90. There is no real problem when the center of the screen is above the horizon of  the globe. The function map.unproject(x,y) returns a coordinate value that is ok for the calculations.
```
    let left = map.unproject([x - optWidth / 2, y]);
    let right = map.unproject([x + optWidth / 2, y]);
```

HarelM:
As far as I understand, the bug was that we measured the scale in the corner of the screen, which wasn't giving good results for globe, when "nothing" was in the corner of the screen. 
Now that we measured near the center (if my understanding is correct) we get better results for globe (yay!) but it might cause worse results for high pitch (since the scale is usually at the corner of the screen, where we don't measure) and there is a difference between the pixel-per-meter in the corner vs in the center. 
So my question is if we have a good idea how to solve this. 
The current code is obviously an improvement and we can merge it as us, but since we are looking at this code, it would be a good idea to try and see what else we can solve without a lot of effort.

pcardinal:
The scale on this map is true mostly on the center of the map because the distorsion is maximum to the poles (at the poles, in fact there is no possible measure). For the globe you will have the same type of situations. With the pitch you still have distorsions. 

image

pcardinal:
The bug was the point used to calculate the scale. Before the changes, this point was the intersection of the horizontal line in the middle of the screen with the left side of the screen. This point was sometimes outside the globe and sometimes inside the globe. Additionally, a second point is calculated to the right of this first point with a distance representing the length of the scale bar. This second point could be outside the globe or inside the globe independently of the first point. After the changes, the center of the screen is used for the basis of the calculations, so that the left and right points are at a symmetrical position relative to the center and are both, at the same time, inside or outside the globe.

The position of the scale bar in the screen has nothing to do with the value of the scale. 

In summary, it is impossible to have a scale representative of the entire map (globe) displayed, mainly at small scale because the distortions are then the greatest.

HarelM (approved):

PR #4899, opened 2024-10-25
Accounting for IDs of cluster features when the promoteID properties is set

popkinj:
This is to address the issue of loosing the ID for clusters when the `promoteId` property is set. Talked about [here](https://github.com/mapbox/mapbox-gl-js/issues/10094).

## Launch Checklist

 - [X] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
Please add a test that covers this scenario and a changelog entry. 
If there's an open issue in the repo please link to it.

popkinj:
> Please add a test that covers this scenario and a changelog entry. If there's an open issue in the repo please link to it.
Hey @HarelM 

Just a quick sanity check... So I'm guessing I'll:

- Make a new folder [here](https://github.com/maplibre/maplibre-gl-js/tree/main/test/integration/render/tests/feature-state) called something like _promote-id-cluster_. 
- Write a style to have one clustered feature and one un-clustered. 
- Set the promoteId to some attribute. 
- Style the two features based on ID? If an ID is missing it would be one colour. If the ID is there it would be another (correct) colour. 
- Then finally create an _expected.png_ file as the control.

I'm I heading down the right direction.

HarelM:
I would look into writing a unit test and not an integration test. i.e. Look for other tests for this class and add the relevant scenario you fixed in the test. 
It is best to check that the test fails without the fix you did to make sure the test actually tests your fix.

src/data/feature_index.test.ts
```diff
@@ -0,0 +1,71 @@
+import {FeatureIndex} from './feature_index';
+import {OverscaledTileID} from '../source/tile_id';
+import {VectorTileFeature} from '@mapbox/vector-tile';
```
HarelM, line 3: Import type?

src/data/feature_index.test.ts
```diff
@@ -0,0 +1,71 @@
...
+describe('FeatureIndex', () => {
+    describe('getId', () => {
+        const tileID = new OverscaledTileID(0, 0, 0, 0, 0);
+        
+        test('converts boolean ids to numbers', () => {
+            const featureIndex = new FeatureIndex(tileID, 'someProperty');
+            const feature = {
+                id: true,
```
HarelM, line 11: "true" feels like a weird choice, can this be an number instead?

HarelM:
Overall, this looks great, thanks!

src/data/feature_index.test.ts
popkinj: Nice catch. Fixed.

src/data/feature_index.test.ts
popkinj: I believe it was a 1 before... switching back.

HarelM:
According to the coverage report, it looks like the test is not covering the code that you added.
Can you check that removing your fix causes the test to fail?
I've seen in the past that the coverage report is not accurate, but I want to make sure this is not the case here...

popkinj:
> According to the coverage report, it looks like the test is not covering the code that you added. Can you check that removing your fix causes the test to fail? I've seen in the past that the coverage report is not accurate, but I want to make sure this is not the case here...

Yes.... Looking into that now. One of the three tests is not behaving how I thought. Maybe that's the issue.
Removing the fix doesn't seem to trip the last of the three tests.

popkinj:
Yeah... Think I was trying to get to fancy with my tests... I've pared it down to only test the three lines.

popkinj:
> According to the coverage report, it looks like the test is not covering the code that you added. Can you check that removing your fix causes the test to fail? I've seen in the past that the coverage report is not accurate, but I want to make sure this is not the case here...

Removing my code now causes my test to fail. However I'm not seeing any changes in the report.
Hmmm.... Maybe I don't understand how the code coverage works. Am I missing some Codecov settings somewhere?

HarelM:
Can you run the coverage locally and see if this is covered?

HarelM:
And also make sure you push your changes in case you haven't by mistake.

popkinj:
> Can you run the coverage locally and see if this is covered?

Good thinking... That's much better then waiting for the pipeline to finish haha.

popkinj:
How do you run the coverage locally @HarelM . Is it something like `npm run codegen`?

HarelM:
`npm run test-unit-ci` IIRC

popkinj:
> `npm run test-unit-ci` IIRC

Hmmmm... That worked well.
In the output report it's not mentioning my new lines as uncovered:

  feature_index.ts                                |   41.22 |    68.75 |   36.36 |   41.22 | 109-172,175-253,258-285,288-295,319-324,326-338,340-342                                                               

If I'm not mistaken... the new code I added are on lines 305-307. Does this mean it worked?

popkinj:
Hey @HarelM . 
Is the GH action really running out of resources... as the error suggests? Or is it something I introduced?

HarelM:
You need to increase the build size expected value a bit.

HarelM (approved):

PR #4914, opened 2024-10-28
Add setVerticalFieldOfView() to public API

NathanMOlson:
## Launch Checklist

This PR adds `setVerticalFieldOfView()` to the public API. This allows different camera focal length to be used, like this:

https://github.com/user-attachments/assets/61529ae2-6981-44ef-ac44-313d21052cd8

This is useful for applications that control the camera position: https://github.com/maplibre/maplibre-gl-js/issues/4717

 - [X] Link to related issues.


NathanMOlson:
@wipfli I think you were asking about this at the last MapLibre-GL-JS meeting?

src/ui/camera.ts
```diff
@@ -550,6 +550,40 @@ export abstract class Camera extends Evented {
...
+    /**
+     * Returns the map's current vertical field of view, in degrees.
+     *
+     * @returns The map's current vertical field of view.
+     * @example
+     * ```ts
+     * map.getFov();
+     * ```
+     */
+    getFov(): number { return this.transform.fov; }
```
HarelM, line 562: Should the public API use FOV, Fov or FieldOfView?

HarelM:
I'm not sure how comfortable I am with exposing this. Can we gather more feedback on this?
Found this article from a few years back, which is somewhat similar to what is requested here I guess...
https://mikevanriel.com/2019/01/04/setting-the-field-of-view-fov-in-mapbox-gl-js/

src/ui/camera.ts
NathanMOlson: I don't know. I picked `setFov` because that is what was used in `TransformInterface`.

I don't really care about Fov/FOV/FieldOfView, but if we change this I would like to make it clear that this is the vertical field of view (as opposed to horizontal or diagonal).

src/ui/camera.ts
```diff
@@ -550,6 +550,40 @@ export abstract class Camera extends Evented {
         return this;
     }
 
+    /**
+     * Returns the map's current vertical field of view, in degrees.
```
HarelM, line 554: Is there a default? Should there be one?

src/geo/transform_helper.ts
```diff
@@ -317,13 +317,13 @@ export class TransformHelper implements ITransformGetters {
         return this._fovInRadians;
     }
     get fov(): number {
-        return this._fovInRadians / Math.PI * 180;
+        return radiansToDegrees(this._fovInRadians);
     }
     setFov(fov: number) {
-        fov = Math.max(0.01, Math.min(60, fov));
-        if (this._fovInRadians === fov) return;
+        fov = clamp(fov, 0.1, 150);
```
HarelM, line 323: Does this change the possible range? why was 60 replaced with 150?

HarelM:
@louwers can this be changed in the Android or native code? I've seen some mapbox old docs that reference this I think.

src/geo/transform_helper.ts
NathanMOlson: Yes it does. I replaced 60 with 150 because I personally want to use FOV higher than 60, and I didn't see any reason to make this small. Stuff starts to get crazy at 180, but 150 looks fine.

src/ui/camera.ts
HarelM: Let's use get/setVerticalFieldOfView then, I prefer to be explicit.

src/ui/camera.ts
NathanMOlson: The default is 36.87 degrees, hardcoded in `transform_helper.ts` line 149.

src/geo/transform_helper.ts
HarelM: Are there method changing this now or is this a "hidden" feature?

src/ui/camera.ts
HarelM: I would add a comment about the defualt then, i.e. `@default ...`.

src/geo/transform_helper.ts
NathanMOlson: This is currently a "hidden" feature.

src/ui/camera.ts
NathanMOlson: Done.

src/geo/transform_helper.ts
NathanMOlson: Changing the 60 degree limit is the most important part of the PR for me. If it is changed, and `setFov` is kept out of the public API, I can use the hidden feature. I would prefer to use the public API though.

src/ui/camera.ts
NathanMOlson: That puts the comment pretty far from the actual definition. It seems like the comment is at high risk of getting out of sync with the code...

I think if I add a `@defaultValue` comment it would make more sense on the `set` than on the `get`, right?

HarelM: I don't know, maybe both...?

louwers:
@HarelM MapLibre Native has a hardcoded fov.

src/ui/camera.ts
NathanMOlson: Done.

wipfli:
Thanks for sharing! Does this work as well for terrain3d and globe?

HarelM (approved):
I hope this won't bite is in the future...

PR #4926, opened 2024-10-29
Globe + 3d terrain + satellite example

birkskyum:
Combining Hybrid Satellite + Globe + 3d Terrain, to see how photorealistic maplibre gl js can be.

Demo:
https://codepen.io/birkskyum-1471370946/pen/xxvWmoG

Affected by
- #4927 

Screenshot 2024-10-29 at 14 50 44


test/examples/globe-3d-terrain-satellite.html
```diff
@@ -0,0 +1,98 @@
...
+            height: 100%;
+        }
+    </style>
+</head>
+
+<body>
+    <div id="map"></div>
+    <script type="module">
+
+        async function downloadStyleObject(url) {
```
HarelM, line 29: Can we simplify this by just writing a short inline style without using an external style?

birkskyum: The hybrid satellite layer from maptiler has many layers, I'm not sure how I can achieve the same with a short inline style.

HarelM: I would recommend using a 2-5 critical layers for the example, instead of complicating it with a style that you override...

birkskyum: Okay, I've now removed ~ 10 layers and inlined the 4 main onces from maptiler:
- city label
- road line
- road label
- sat raster tiles

The example still demonstrate what it's supposed to. The line count grew from 89 to 249 lines.

birkskyum: > The layers are still somewhat complicated - the idea of the example is to focus on what you would like to highlight and not how to write styles.

I fully agree with this, which is why I've reverted the inline again. Having most of style in the maptiler url better show what I'm trying to highlight, which is the full extend of maplibre globe+terrain+hillshade+hybrid rendering capabilities, instead of focusing on how to write inline styles simpler, which cuts into the visual output I'm trying to highlight.

HarelM: Can you add a style to the demo tiles repo with what's needed for this example?

birkskyum: Yes, the full style is here:
- https://github.com/maplibre/demotiles/pull/30

birkskyum: I don't know if it's really worth adding a new style to demotiles just for these 30L of style modification, but i've reached out about the licensing, to figure if it's actually possible for us to do so.

HarelM: Let's wait for a response, I think using a style file is a lot cleaner for what we would like to showcase here.

kubapelc:
We have never actually implemented globe support for terrain, so if this works, it works by accident.

PR #4937, opened 2024-10-30
Globe covering tiles optimization

kubapelc:
Improves the performance of globe's `coveringTiles` function, adds benchmarks to measure globe `coveringTiles` performance, also likely slightly improves mercator's `coveringTiles` performance (not measured, but some of the changes should help there too).

Improvements:

- Adds a cache for tile AABBs to avoid recomputing them every time. The cache is cleared at the beginning of every frame, only keeping the tile AABBs that were accessed in the last frame. Any two consecutive frames should mostly be viewing the same set of tiles, so most of the cache should be reused. (Only the AABB cache idea from #4778 is implemented, not the accurate tile-frustum intersections.)
- Improves performance of plane-AABB intersection test
- Improves frustum-AABB performance by applying the pervious optimization to the frustum's 6 planes
- Simplifies the final test in frustum-AABB intersection by precomputing the frustum's bounding box

## Performance

There are two benchmarks, one with the camera looking at the globe from above, one with a pitched camera. The camera slowly pans to the east during the benchmark. Here are the rough contributions of individual changes:

| Change         | Unpitched | Pitched |
| :------------- | --------: | ------: |
| Baseline       |   13.5 ms | 21.0 ms |
| Cached AABBs   |    9.7 ms | 14.0 ms |
| +FastAABBplane |    8.8 ms | 12.2 ms |
| +FastFrustum   |    7.4 ms |  9.8 ms |
| +FrustumBounds |    7.4 ms |  9.5 ms |


HarelM:
The following PR is nearing its merge:
- https://github.com/maplibre/maplibre-gl-js/pull/4779

If you need, I can create a branch in this repo for collaboration, so you could open this PR against the updates made in the above PR, but otherwise, I prefer to have the above PR merged first to avoid conflicts on the hard work @NathanMOlson did.

kubapelc:
Lets wait for #4779 to be merged first.

birkskyum:
@kubapelc , does this affect tile loading of the globe when the terrain3d is active too?

kubapelc:
> does this affect tile loading of the globe when the terrain3d is active too?

It should not, this PR does not change functionality of tile culling, just makes it faster.

But globe coveringTiles so far assumes that there is no terrain and does not take tile min/max elevation into account when generating AABBs. I think it should not be an issue for the purposes of porting terrain to globe.

I'd like to redo globe coveringTiles to not use AABBs, since they do not approximate tiles on a sphere very well, and use oriented bounding boxes instead. I would add support for tile min/max elevation along with this.

birkskyum:
@kubapelc , great - note the terrain is fully working in here on globe (it's only the heatmap i'm working on), except for an issue of the poles #4984 
- #4977

HarelM:
Converting tiles have been merged, this can now proceed.

HarelM:
Can you add some tests and a changelog entry?

kubapelc:
> Can you add some tests and a changelog entry?

I've added both. Most of the changes were already covered by existing tests, I've added tests for aabb-plane intersection and for the aabb cache.

src/geo/projection/globe_covering_tiles_details_provider.ts
```diff
@@ -38,6 +38,26 @@ function distanceToTileWrapX(pointX: number, pointY: number, tileCornerX: number
 }
 
 export class GlobeCoveringTilesDetailsProvider implements CoveringTilesDetailsProvider {
+    private _cachePrevious: Map<string, Aabb> = new Map();
```
HarelM, line 41: Maybe we can move some of this logic to a dedicated AABBCache class?

kubapelc: Done.

src/util/primitives.ts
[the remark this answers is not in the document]
kubapelc: Done.

src/util/primitives.ts
```diff
@@ -8,7 +8,7 @@ export const enum IntersectionResult {
 
 export class Frustum {
 
-    constructor(public points: vec4[], public planes: vec4[]) { }
+    constructor(public points: vec4[], public planes: vec4[], public aabb: Aabb) { }
```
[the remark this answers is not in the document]
kubapelc: It is used outside, by the AABB class to perform AABB-frustum intersection test.

src/util/primitives/aabb_cache.ts
```diff
@@ -0,0 +1,53 @@
...
+    constructor(aabbFactory: AabbFactory) {
+        this._aabbFactory = aabbFactory;
+    }
+
+    /**
+     * Prepares AABB cache for next frame. Call at the beginning of a frame.
+     * Any tile accesses in the last frame is kept in the cache, other tiles are deleted.
+     * @returns 
+     */
+    newFrame() {
```
HarelM, line 21: Can this get a better name maybe? `switchCacheIfDirty`? I'm not sure I fully follow the concept of two "dictionaries" in order to represent a cache, is this in order for the cache clean to be fast?

src/util/primitives/aabb_cache.ts
HarelM, line 1: Not sure I fully understand the idea of primitives folder, but I don't have strong feelings against it...

src/util/primitives/aabb_cache.ts
HarelM: Also, according to the coverage report, this method is not fully covered, but I don't fully trust the coverage report.

HarelM:
This looks straight forward and good, thanks!
Added a few minor comments.

src/util/primitives/aabb_cache.ts
kubapelc: I see it as a folder for classes related to bounding primitives and frustum culling, which was the contents of the old primitives.ts file. I would like to add a new "oriented bounding box" primitive in the future.

src/util/primitives/aabb_cache.ts
kubapelc: I think current "newFrame" is a better name, since the purpose of the cache is to take advantage of the fact that the set of visible tiles between any two consecutive frames (excluding camera jumps) will be very similar, so the AABBs can be reused. But any AABB that was not accessed during the current frame will be deleted. So the cache just needs to be notified that a new frame begins.

kubapelc: > this in order for the cache clean to be fast?

Yes, but I'm not sure if it is the best approach. I think (hope) that this shouldn't really matter for performance. An alternative would be to keep a flag for each AABB whether it was accessed during the current frame and at the beginning of new frame delete all AABBs that were not accessed.

PR #4945, opened 2024-10-31
[example] Animate symbol to follow mouse

birkskyum:
This is an example that show how to animate a symbol to follow the mouse around:

Demo: https://codepen.io/birkskyum-1471370946/pen/wvVjXxY

This functionality was requested on slack [here](https://osmus.slack.com/archives/C03TFH5NE83/p1730379151240459), and I think it's a good compliment to the other examples we have that all have pre-defined paths (along line, along route, in circle)

## Launch Checklist


test/examples/animate-symbol-to-follow-mouse.html
```diff
@@ -0,0 +1,72 @@
...
+        });
+
+
+        map.addLayer({
+            'id': 'point',
+            'source': 'point',
+            'type': 'symbol',
+            'layout': {
+                'icon-image': 'airport_15',
+                'icon-rotate': ['get', 'bearing'],
```
HarelM, line 56: Remove bearing? I haven't seen this used here...

test/examples/animate-symbol-to-follow-mouse.html
```diff
@@ -0,0 +1,72 @@
...
+                'icon-rotation-alignment': 'map',
+                'icon-overlap': 'always',
+                'icon-ignore-placement': true
+            }
+        });
+
+        function animateMarker(timestamp) {
+            map.getSource('point').setData(getPoint());
```
HarelM, line 64: Shouldn't this line be called inside the `mousemove` handler instead of this loop of animation frame?
I think it can simpify the reading a bit...?

test/examples/animate-symbol-to-follow-mouse.html
```diff
@@ -0,0 +1,72 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>Animate symbol to follow the mouse</title>
```
HarelM, line 4: Should we link from the API docs to this example?
There are some `@see` comments in the code to showcase a usage of a method or event.
Should this be the case for this example as well?

birkskyum:
@HarelM , you're right, I managed to simplify the example a lot, by removing the animation loop in favor of adding the logic to mousemove

test/examples/animate-symbol-to-follow-mouse.html
```diff
@@ -0,0 +1,74 @@
...
+            const pointSource = map.getSource('point');
+
+            if (pointSource && lngLat.lng && lngLat.lat) {
+                pointSource.setData({
+                    'type': 'Point',
+                    'coordinates': [lngLat.lng, lngLat.lat]
+                });
+            }
+        });
+        map.addControl(new maplibregl.GlobeControl());
```
HarelM, line 48: Is this relevant here?

birkskyum: It makes it clear how it'll behave on the globe - it clamps, but still track the mouse:

https://github.com/user-attachments/assets/4964120e-6e9c-4246-bfb1-1716dddf014b

HarelM: Oh, great!

HarelM (approved):

PR #966, opened 2022-02-08
Migrate query tests from puppeteer to playwright

birkskyum:
We currently use both selenium and puppeteer, for the following reasons:

- selenium offers cross-browser support
- puppeteer is maintained by chrome team and doesn't need to run an additional server, and is much more ergonomical.

Fortunately, playwright is an evolution after puppeteer (and made by main puppeteer devs) almost with same api, and with cross-browser support. 

I think we'll be better off just using playwright throughout, as it works well with jest as test runner, so this pr ports the query tests.

PR #4949, opened 2024-10-31
[investigate] Replace Puppeteer with Playwright (Vitest Browser Mode?)

birkskyum:
Ongoing work on improving the test suite.

Initially we had Selenium in Query / Browser suite, and custom mocks of webgl1 in Render suite.

#### February 2022 - **Browser, Query**: Selenium to Playwright
- https://github.com/maplibre/maplibre-gl-js/pull/974
- https://github.com/maplibre/maplibre-gl-js/pull/966

#### May 8th 2023 - **Browser, Query**: Playwright to Puppeteer
Google chrome made a splash. Until this point headless and headful chrome were two different browsers, and now they cleaned it up so the headless became a "real chrome", which is a lot better. Only puppeteer supported this initially, so we switched to be able to leverage it in (query + browser):
- https://github.com/maplibre/maplibre-gl-js/pull/2494

#### May 9th 2023 - **Render**: Custom Webgl1 Mocks to Puppeteer:
- https://github.com/maplibre/maplibre-gl-js/pull/2497

#### Now - **Browser, Query, Render**: Puppeteer to Playwright
This is now old history, and since, playwright has due to a nicer dx, and multi-browser support (firefox / webkit / chromium), managed to become the No. 1 browser runner, and it's possible to it thoughout.

As an example of the better DX, this is i.e. a diff from browser.test.ts:

Puppeteer:
```ts
await page.emulateMediaFeatures([
    {name: 'prefers-reduced-motion', value: 'reduce'},
]);
```

Playwright:
```ts
await page.emulateMedia({reducedMotion: 'reduce'});
```

Screenshot 2024-10-31 at 18 51 53

Rendet test HTML reporting works - here a full run on a mac.

Screenshot 2024-10-31 at 20 01 09

## Launch Checklist


HarelM:
Looks like a good direction, let's make sure the coverage is working as expected as I had to fight with it back then...

birkskyum:
@HarelM , how to I check the coverage? is it specific to the render test suite?

HarelM:
If this is only a refactoring, the coverage should stay the same between main branch and this one. 
You can open an empty PR and compare the report in codecov, probably selectively.

birkskyum:
It seems like the render tests take 2.5x longer to run with the current state of this PR (26 min. vs ~ 11 min), so that has to be reduced significantly before this PR can be merged.

HarelM:
Yes, that's a concern...

birkskyum:
The performance overhead appear to come from improved isolation / cross-browser architechture, it just is a bit steep price. I'll revisit this when the vitest browser mode is stable, as it might bring forward some options for performance improvements:
- https://vitest.dev/guide/browser/

birkskyum:
The test suite for Integration / Unit is now Vitest.

That means we can explore Vitest Browser for render tests:
- https://vitest.dev/guide/browser/

HarelM:
@birkskyum can this PR be closed?
I'm trying to clean up the PRs to better reflect what needs a review and what is currently being worked on.

birkskyum:
This is something I'm working on, attempting to make a better setup for the tests that need real browsers. I wanted to give vitest [a shot](https://github.com/maplibre/maplibre-gl-js/pull/4949#issuecomment-2494945957), to see how it compares.

There's no need for a review at this stage, and likely won't be for quite a while, so I can continue locally. Basically, the only benefits of a draft pr is to be able to check the actual impact CI times (but chances are that if it's faster locally, it'll be faster on CI), and to have the work happen in the open.

unphased:
Curious -- is it possible to fake clicks with maplibre and playwright? I am trying to hack some simple playwright smoke tests and it wasn't clear if this is possible. At least for now it is okay to deal with my testing need as i can just bring the code that the click handler runs into `page.evaluate`.

Separately seems like an OCR layer and image based click offset computation ability would be the way to go to empower playwright better to drive canvas based UIs.

Issue #4950, opened 2024-10-31
Sprite url normalization breaks some url encoding / signing schemes


jcary741:
**maplibre-gl-js version**: 5.0.0-pre.4 / main

**browser**: Chrome

### Expected Behavior
I'm using Amazon Location Service-hosted maps, which require url signing (AWS Signature Version 4) for all requests. Upon `transformRequest`, they expect the url not to materially change. 

### Actual Behavior
Following #3898, `normalizeSpriteURL()` creates a new URL object to parse the URL, then converts it back to string. Something in this process appears to mangle encoding of the signature, or possibly materially changes the url itself in such a way that it is no longer valid.

### Possible solution
Instead of having `normalizeSpriteUrl` be an automatic, internal step, I suggest having a default version of `transformRequest` which checks for the 'Sprite' resource type and applies the "normalization" process. However, if the developer sets their own `transformRequest`, this can be overridden. 

Recent changes to MapLibre are looking awesome! Keep up the good work! <3

HarelM:
Thanks for taking the time to report this issue and try out the latest version!
I'm not sure I fully understand what you are suggesting or what exactly the problem is.
I would advise to open a PR with the relevant test and a possible fix.

jcary741:
Heh, I actually just noticed your comment [here](https://github.com/maplibre/maplibre-gl-js/pull/3898#pullrequestreview-2386276048), which is essentially what is going wrong here. Reviewing the several issues / PRs relating to this topic, I can see how nuanced this discussion is. I'll see what I can do about putting together a PR shortly.

jcary741:
#4962 will resolve this issue, primarily by reverting #3898. One thing I'll mention here in case anyone else is dealing with signed URLs is that, as it turns out, the issue wasn't encoding related as I had thought, instead it was simply that the signature was being generated *before* ".json" was appended to the sprite URL pathname, thereby invaliding the signature.

HarelM:
Closing this, please let me know if this is still relevant and I'll reopen.

PR #3923, opened 2024-03-31
removed url validation form normalizeSpriteURL

Kai-W:
Temporary Fix for normalizeSpriteURL not working with relative URLs
See Issue https://github.com/maplibre/maplibre-gl-js/issues/3897
Permanent Fix with API Breaking Changes: https://github.com/maplibre/maplibre-gl-js/pull/3898

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.
 - [x] Link to related issues.
 - [ ] Include before/after visuals or gifs if this PR includes visual changes.
 - [x] Write tests for all new functionality.
 - [ ] Document any changes to public APIs.
 - [ ] Post benchmark scores.
 - [x] Add an entry to `CHANGELOG.md` under the `## main` section.

PR #4962, opened 2024-11-02
Fix URL parsing in normalizeSpriteURL

jcary741:
## Launch Checklist

 - [x] Link to related issues.

Related issues: #3897, #182, #4950
Related PRs: #3898, #3923, #645

This PR reverts #3898, which introduces issues with use cases where `transformRequest()` is used for URL signing (see #4950). It also disallows relative URLs for sprite URLs moving forward. However, since #182 and #3897 were opened, [transformStyleFunction](https://maplibre.org/maplibre-gl-js/docs/API/type-aliases/TransformStyleFunction/) has been introduced (back in 2022), which can be used by developers to pre-transform relative URLs to whatever is most appropriate. Notably, this can support both 1) ArcGIS [Vector Tile Style](https://developers.arcgis.com/rest/services-reference/enterprise/vector-tile-style.htm) -style relative URLs at a separate domain than the map is being used from, and 2) relative URLs hosted at the same domain as the map. This PR adds examples for how to use `transformStyle` to make URLs absolute. I don't think it is exhaustive, but definitely covers the ArcGIS Vector Tile Style cases.

Also, `normalizeSpriteURL()` now also throws a detailed error message to guide developers how to solve the issue.


jcary741:
Oh, do I need to re-build docs & commit, or does the CI do that somewhere?

HarelM:
Is referring the previous PR the only solution to this issue? 
Cc: @Kai-W

jcary741:
It's hard to spot through the diffs, but this does preserve the usage of `new URL()` to parse URLs, which was introduced by #3898.

jcary741:
Oh, just re-read your question and see I misunderstood. Yes, I think this is the only solution here. On the net, we remove one breaking change (the re-ordering of normalizeSpriteURL and transformRequest), for another (requiring URLs to be absolute, and directing developers to correct the issue as needed).

Kai-W:
I see the problem and i didn't know signed URls were a thing.
Using TransformStyleFunction could work for most cases. I probably missed that function because it is part of setStyle not part of the constructor. In the docs the constructor only mentions transformRequest. TransformStyle is not and Option of the constructor.  If this is the intended way of dealing with relative urls, then transformStyle sould also be added to the Constructor as well.

An other solution could be to add an additional transform request.  I expected transformRequest to be able to handle relative URLs before they get validated and for the signing URLs usecase the URLS should be absolute, validated and the extension .json/.png should be appended. Calling transofrm request before normalizeSpriteURL with type "Sprite" and 2 times after the normalisation with "SpriteImage" & "SpriteJSON" could be an other Option. Shouldn't break existing code and gives the flexibilty to transform all Requests before the url gets Normalized.

HarelM:
@Kai-W I'm not sure I fully understand what you suggest. 
If you can this of a different way to solve this problem can you please elaborate or open a different PR to showcase the code?

jcary741:
@Kai-W I considered that approach, but making multiple calls with the same URL could just as easily cause issues for developers who choose to implement non-idempotent callbacks for `transformRequest`. I do like your idea of adding `transformStyle` to the map constructor, and I think it would actually also make sense to have it persisted on the Style class instance (where right now it apparently must be passed with each `setStyle` call). I'd be happy to submit a separate PR if @HarelM agrees.

HarelM:
Sure, you can create a different PR for style transform, I'll need to see the ergonomics around it to decide if it makes sense, and what would happen when you set both this transform and pass a function to `setStyle`, but I think it can be a good addition.

But let's close on this PR first. What is the current status? Can you guys agree on the needed solution?

Kai-W:
@jcary741 solution works for me and an additionale PR for a persisted transformStyle sounds good as well

HarelM (approved):

HarelM:
This is not a breaking change to version 4, right? Just a breaking version between the pre-releases...?

jcary741:
It's breaking between pre-releases, but not with version 4. It does make the requirement for sprite URLs to be absolute (post transformStyle) explicit, but I don't think that worked previously anyway based on #182.

src/style/load_sprite.ts
```diff
@@ -16,9 +16,14 @@ export type LoadSpriteResult = {
 }
 
 export function normalizeSpriteURL(url: string, format: string, extension: string): string {
-    const parsed = new URL(url);
-    parsed.pathname += `${format}${extension}`;
-    return parsed.toString();
+    try {
+        const parsed = new URL(url);
```
raboczi, line 20: Would there be bad consequences for changing this line to `const parsed = new URL(url, location.href);` so that stylesheets could be relative?  It's a real nuisance that a static site can't be redeployed without editing this one file.

HarelM:
@Kai-W any objections to merge this?

Kai-W:
nop works for me

HarelM (approved):

PR #5020, opened 2024-11-08
Refactor usage of implicit parameter to an explicit one when rendering

ibesora:
This PR aims to refactor our codebase a little bit in order to move away of what, IMHO, are code smells.

First, some context. 
The [Painter](https://github.com/maplibre/maplibre-gl-js/blob/main/src/render/painter.ts#L69) object mostly contains properties that describe how things are going to be rendered via the underlying rendering API (i.e. vertex and index buffers, stencil buffer definitions, depth range, etc.) but it also contains the `Style` object which in itself describes how things should look like. 
`draw_*` functions (i.e. [draw_background](https://github.com/maplibre/maplibre-gl-js/blob/b6e3b551e85b4fe42211226f75c0991cba698418/src/render/draw_background.ts#L15)) get both the `Painter` object and `StyleLayer` object, which is basically the subset of the `Style` that we need to render a given layer.

There are various places in the `draw_*` functions where we use `painter.style` to get information about the current rendering mode:
- We use `painter.style.map.terrain` to know if we are rendering with terrain
- We use `painter.style.projection.name` to know if we are rendering using `globe` or `mercator` projections

and change the drawing behaviour according to that. Since those properties are deep into the `Painter` object, you might get two similar-looking `draw_*` calls with very different results. 

This PR replaces checks for terrain rendering via the `painter.style` property with an extra `renderLayer` parameter that tells a given `draw_*` function if it's being rendered to a texture.

A follow-up idea would be to remove `style` entirely from the `Painter` object and add all the properties that are needed for the various `draw_*` functions into the `Painter` object itself.

Cc @birkskyum since we talked about this.

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/geo/projection/globe_transform.test.ts
+++ b/src/geo/projection/globe_transform.test.ts
@@ -27,7 +27,7 @@ function planeDistance(point: Array<number>, plane: Array<number>) {
 }
 
 function createGlobeTransform(globeProjection: GlobeProjection) {
-    const globeTransform = new GlobeTransform(globeProjection);
+    const globeTransform = new GlobeTransform(globeProjection, true);
     globeTransform.resize(640, 480);
     globeTransform.setFov(45);
     return globeTransform;
@@ -46,14 +46,14 @@ describe('GlobeTransform', () => {
             expectToBeCloseToArray(projectionData.tileMercatorCoords, [0.5, 0, 0.5 / EXTENT, 0.5 / EXTENT]);
         });
 
-        test('Globe transition is not 0 when not ignoring the globe matrix', () => {
+        test('Globe transition is 0 when not applying the globe matrix', () => {
             const projectionData = globeTransform.getProjectionData({overscaledTileID: new OverscaledTileID(1, 0, 1, 1, 0)});
-            expect(projectionData.projectionTransition).not.toBe(0);
+            expect(projectionData.projectionTransition).toBe(0);
         });
 
-        test('Ignoring the globe matrix sets transition to 0', () => {
-            const projectionData = globeTransform.getProjectionData({overscaledTileID: new OverscaledTileID(1, 0, 1, 1, 0), ignoreGlobeMatrix: true});
-            expect(projectionData.projectionTransition).toBe(0);
+        test('Applying the globe matrix sets transition to something different than 0', () => {
+            const projectionData = globeTransform.getProjectionData({overscaledTileID: new OverscaledTileID(1, 0, 1, 1, 0), applyGlobeMatrix: true});
+            expect(projectionData.projectionTransition).not.toBe(0);
         });
     });
 
--- a/src/render/draw_background.ts
+++ b/src/render/draw_background.ts
@@ -12,13 +12,13 @@ import type {BackgroundStyleLayer} from '../style/style_layer/background_style_l
 import {OverscaledTileID} from '../source/tile_id';
 import {coveringTiles} from '../geo/projection/covering_tiles';
 
-export function drawBackground(painter: Painter, sourceCache: SourceCache, layer: BackgroundStyleLayer, coords: Array<OverscaledTileID>, renderFlags: RenderOptions) {
+export function drawBackground(painter: Painter, sourceCache: SourceCache, layer: BackgroundStyleLayer, coords: Array<OverscaledTileID>, renderOptions: RenderOptions) {
     const color = layer.paint.get('background-color');
     const opacity = layer.paint.get('background-opacity');
 
     if (opacity === 0) return;
 
-    const {isRenderingToTexture} = renderFlags;
+    const {isRenderingToTexture} = renderOptions;
     const context = painter.context;
     const gl = context.gl;
     const projection = painter.style.projection;
--- a/src/render/draw_background.ts
+++ b/src/render/draw_background.ts
@@ -6,13 +6,13 @@ import {
     backgroundPatternUniformValues
 } from './program/background_program';
 
-import type {Painter, RenderFlags} from './painter';
+import type {Painter, RenderOptions} from './painter';
 import type {SourceCache} from '../source/source_cache';
 import type {BackgroundStyleLayer} from '../style/style_layer/background_style_layer';
 import {OverscaledTileID} from '../source/tile_id';
 import {coveringTiles} from '../geo/projection/covering_tiles';
 
-export function drawBackground(painter: Painter, sourceCache: SourceCache, layer: BackgroundStyleLayer, coords: Array<OverscaledTileID>, renderFlags: RenderFlags) {
+export function drawBackground(painter: Painter, sourceCache: SourceCache, layer: BackgroundStyleLayer, coords: Array<OverscaledTileID>, renderFlags: RenderOptions) {
     const color = layer.paint.get('background-color');
     const opacity = layer.paint.get('background-opacity');
 
--- a/src/render/draw_background.ts
+++ b/src/render/draw_background.ts
@@ -12,7 +12,7 @@ import type {BackgroundStyleLayer} from '../style/style_layer/background_style_l
 import {OverscaledTileID} from '../source/tile_id';
 import {coveringTiles} from '../geo/projection/covering_tiles';
 
-export function drawBackground(painter: Painter, sourceCache: SourceCache, layer: BackgroundStyleLayer, coords?: Array<OverscaledTileID>) {
+export function drawBackground(painter: Painter, sourceCache: SourceCache, layer: BackgroundStyleLayer, coords?: Array<OverscaledTileID>, isRenderingToTexture: boolean = false) {
     const color = layer.paint.get('background-color');
     const opacity = layer.paint.get('background-opacity');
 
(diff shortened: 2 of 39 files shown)


birkskyum:
@ibesora , thanks! I think this is much better. I have an ask - is there a way we can flip this around? So instead of enabling lots of matrix transformations by defualt, and then actively disabling the ones we don't want, that we instead pass the transformations that we **_do_** want?

HarelM:
I think this is great!
I would also look in the code and find the places where "this" is being passed down. Solving those places will prevent this behavior I believe where you look deep into objects to find references to other objects. 

Since we are in the middle of a breaking change version release, now would be a great time to move objects around and clean some historical bad decisions...

src/render/painter.ts
```diff
@@ -640,34 +640,34 @@ export class Painter {
...
             case 'raster':
                 drawRaster(painter, sourceCache, layer as any, coords, isRenderingToTexture);
                 break;
             case 'background':
-                drawBackground(painter, sourceCache, layer as any, coords);
+                drawBackground(painter, sourceCache, layer as any, coords, isRenderingToTexture);
                 break;
             case 'custom':
-                drawCustom(painter, sourceCache, layer as any);
+                drawCustom(painter, sourceCache, layer as any, isRenderingToTexture);
```
HarelM, line 670: Can you fix all these "as any" while you change this code?

HarelM (approved):

src/render/draw_heatmap.ts
```diff
@@ -179,7 +179,7 @@ function renderHeatmapTerrain(painter: Painter, layer: HeatmapStyleLayer, coord:
     context.activeTexture.set(gl.TEXTURE1);
     colorRampTexture.bind(gl.LINEAR, gl.CLAMP_TO_EDGE);
 
-    const projectionData = transform.getProjectionData({overscaledTileID: coord, ignoreTerrainMatrix: !isGlobe});
+    const projectionData = transform.getProjectionData({overscaledTileID: coord, ignoreTerrainMatrix: !isGlobe, ignoreGlobeMatrix: isRenderingToTexture});
```
birkskyum, line 182: Can we do the same for ignoreTerrainMatrix ? 


```ts
const isGlobe = painter.style.projection.name === 'globe';

ignoreTerrainMatrix: !isGlobe,
```

birkskyum:
We need to pull out the check on the "globe" string for anywhere that uses the globe, because the projection "vertical-perspective" is also using the same logic, just without the transition to mercator.

src/render/draw_heatmap.ts
birkskyum: We need to pull out the check on the "globe" string for anywhere that uses the globe, which is all ths, because the projection "vertical-perspective" is also using the same logic, just without the transition to mercator.

ibesora:
I've updated this PR with some of your suggestions:
- To remove reading `globe` from the style as suggested [here](https://github.com/maplibre/maplibre-gl-js/pull/5020#discussion_r1835143304), instead of adding another boolean to all the `draw_*` functions, I moved those into a type called `RenderFlags` that will be able to grow with more flags.
- I also added type predicate functions to change remove the `layer as any` casts.

I'll tackle removing `style` from the `Painter` object in another PR

src/render/painter.ts
```diff
@@ -62,6 +72,11 @@ type PainterOptions = {
     fadeDuration: number;
 };
 
+export type RenderFlags = {
```
HarelM, line 75: I think "Options" is used in the code. I would consider renaming this to "RenderOptions".

src/render/painter.ts
```diff
@@ -633,41 +649,41 @@ export class Painter {
...
-    renderLayer(painter: Painter, sourceCache: SourceCache, layer: StyleLayer, coords: Array<OverscaledTileID>, isRenderingToTexture: boolean = false) {
+    renderLayer(painter: Painter, sourceCache: SourceCache, layer: StyleLayer, coords: Array<OverscaledTileID>, renderFlags: RenderFlags) {
         if (layer.isHidden(this.transform.zoom)) return;
         if (layer.type !== 'background' && layer.type !== 'custom' && !(coords || []).length) return;
         this.id = layer.id;
 
-        switch (layer.type) {
-            case 'symbol':
-                drawSymbols(painter, sourceCache, layer as any, coords, this.style.placement.variableOffsets);
+        switch (true) {
```
HarelM, line 657: I find this odd TBH, maybe a if ...else if... Will be better here? It might be just that I'm not used to it... IDK.

src/render/draw_background.ts
```diff
@@ -6,13 +6,13 @@ import {
...
 
-import type {Painter, RenderFlags} from './painter';
+import type {Painter, RenderOptions} from './painter';
 import type {SourceCache} from '../source/source_cache';
 import type {BackgroundStyleLayer} from '../style/style_layer/background_style_layer';
 import {OverscaledTileID} from '../source/tile_id';
 import {coveringTiles} from '../geo/projection/covering_tiles';
 
-export function drawBackground(painter: Painter, sourceCache: SourceCache, layer: BackgroundStyleLayer, coords: Array<OverscaledTileID>, renderFlags: RenderFlags) {
+export function drawBackground(painter: Painter, sourceCache: SourceCache, layer: BackgroundStyleLayer, coords: Array<OverscaledTileID>, renderFlags: RenderOptions) {
```
HarelM, line 15: Don't forget to change the parameter name as well - renderOptions...

HarelM:
LGTM, thanks!

src/geo/projection/globe_transform.test.ts
```diff
@@ -46,14 +46,14 @@ describe('GlobeTransform', () => {
             expectToBeCloseToArray(projectionData.tileMercatorCoords, [0.5, 0, 0.5 / EXTENT, 0.5 / EXTENT]);
         });
 
-        test('Globe transition is not 0 when not ignoring the globe matrix', () => {
```
birkskyum, line 49: Thanks you, this is a lot cleaner.

From Triple negation
`        test('Globe transition is not 0 when not ignoring the globe matrix', () => {`

To Single negation
`         test('Globe transition is 0 when not applying the globe matrix', () => {`

src/render/painter.ts
ibesora: Yeah, to be fair it was a little bit weird. Changed it to an `if...else` block and it feels better.

src/render/draw_background.ts
```diff
@@ -47,7 +47,7 @@ export function drawBackground(painter: Painter, sourceCache: SourceCache, layer
     for (const tileID of tileIDs) {
         const projectionData = transform.getProjectionData({
             overscaledTileID: tileID,
-            ignoreGlobeMatrix: globeWithTerrain
+            applyGlobeMatrix: !isRenderingToTexture
```
birkskyum, line 50: is an `applyTerrainMaterix: true` missing here?

src/render/draw_circle.ts
```diff
@@ -80,7 +81,7 @@ export function drawCircles(painter: Painter, sourceCache: SourceCache, layer: C
         const terrainData = painter.style.map.terrain && painter.style.map.terrain.getTerrainData(coord);
         const uniformValues = circleUniformValues(painter, tile, layer, translateForUniforms, radiusCorrectionFactor);
 
-        const projectionData = transform.getProjectionData({overscaledTileID: coord});
+        const projectionData = transform.getProjectionData({overscaledTileID: coord, applyGlobeMatrix: !isRenderingToTexture});
```
birkskyum, line 84: Is an applyTerrainMatrix missing here too?

src/render/draw_background.ts
birkskyum: Or maybe background isn't affected by terrain?

src/render/draw_fill.ts
```diff
@@ -106,10 +108,9 @@ function drawFillTiles(
 
         updatePatternPositionsInProgram(programConfiguration, fillPropertyName, constantPattern, tile, layer);
 
-        const globeWithTerrain = painter.style.map.terrain && painter.style.projection.name === 'globe';
         const projectionData = transform.getProjectionData({
             overscaledTileID: coord,
-            ignoreGlobeMatrix: globeWithTerrain
+            applyGlobeMatrix: !isRenderingToTexture
```
birkskyum, line 113: Here also, I'd expect applyTerrainMatrix: true

src/render/draw_hillshade.ts
```diff
@@ -73,12 +75,10 @@ function renderHillshade(
         context.activeTexture.set(gl.TEXTURE0);
         gl.bindTexture(gl.TEXTURE_2D, fbo.colorAttachment.get());
 
-        const globeWithTerrain = painter.style.map.terrain && painter.style.projection.name === 'globe';
-
         const projectionData = transform.getProjectionData({
             overscaledTileID: coord,
             aligned: align,
-            ignoreGlobeMatrix: globeWithTerrain
+            applyGlobeMatrix: !isRenderingToTexture
```
birkskyum, line 81: here too, applyTerrainMatrix: true

PR #5022, opened 2024-11-08
Organize projection folder

birkskyum:
The projection folder is getting a bit unwieldy, growing with a bunch of the same files for each new projection, and mixing the projection-specific and projection-unspecific files. This PR is just bucketing them

before
before

after
after

## Launch Checklist


birkskyum:
can ways

PR #5027, opened 2024-11-09
Fix renderWorldCopies regression

NathanMOlson:
This fixes #5024.

 - [X] Link to related issues.


birkskyum:
Thanks @NathanMOlson  - I tried it, but I don't think it's sufficient.

HarelM:
Please add the relevant tests so this won't break in the future...

birkskyum (approved):
That does appear to work like I'd expect it to

birkskyum:
i'm trying to figure how we can test this. The render test suite has a style.json set initially, but this bug show up after changing the style.

HarelM:
@NathanMOlson any updates on this?

NathanMOlson:
> @NathanMOlson any updates on this?

No. I haven't figured out how to write a test for this.

HarelM:
Do you need my help for this? A unit test works probably be sufficient I guess...

HarelM:
The copies are shown, but not at the appropriate size, once you pan the map a bit, the size of the map changes to fit the screen:
image

So I believe this is not fully fixed. I'm working on a test, maybe a render test, to showcase this issue.

HarelM:
It has something to do with the zoom change between globe and mercator, and probably a constraint around the zoom which is not being respected when setting the projection from globe to mercator...

HarelM:
I'll merge it to a branch under maplibre to allow collaboration.

Issue #5004, opened 2024-11-06
Toggle for Terrain collision detection, to disable it


birkskyum:
Suggested [here](https://github.com/maplibre/maplibre-gl-js/pull/4551#issuecomment-2451287527) by @Samarth1696  

> I would like to disable this feature for the use case for aircraft. Currently motion control works with `jumpTo()` in every frame. From what I have seen from my work of motion control that the camera goes under the terrain regardless of `elevateCameraIfInsideTerrain` is being called, this is because of `jumpTo()` I think. This feature would be correct for the use case for normal users but things might go wrong for aircraft motion control use case.
> 
> We can have a on and off switch for this feature where the user can choose between going under the terrain and not going under it. I know for the default UX no one wants to go under the Terrain but we have to find a solution for the aircraft use case.

pronitdas:
@Samarth1696

PR #5028, opened 2024-11-10
Toggle to allow Camera Below the Surface

Samarth1696:
Solution to #5004.

before `terrainCollision: true`:

https://github.com/user-attachments/assets/fd56cc62-72f8-4cff-bb65-5521a01177e3

after `terrainCollision: false`:

https://github.com/user-attachments/assets/0fa2fc31-d9a3-41f9-8dbe-fa0b598d0b56

Just added small changes.

## Launch Checklist

 - [x] Link to related issues.


Samarth1696:
> I think I would recommend something more descriptive, terrainCollision sounds like a callback and not a boolean value...

I also thought of that. Maybe straight away call it as `goUnderTheTerrain`?

> Also, is there another way to achieve this without adding this to the public API?

Ok, noted. Users will just have to use the setter to enable it.

HarelM:
I thought about something like `allowTerrainCollision` or `allowCameraBelowSurface`, IDK... you can post on slack to get better ideas for names I guess...

> Ok, noted. Users will just have to use the setter to enable it.

If we have a setter I think would should allow initializing the map with it, but I'm asking if there's even something else that requires even less public API changes?

Samarth1696:
Based on this conversation I think `allowCameraBelowSurface` will look more sensible: https://github.com/maplibre/maplibre-gl-js/pull/4851#pullrequestreview-2387677714

> If we have a setter I think would should allow initializing the map with it, but I'm asking if there's even something else that requires even less public API changes?

I have reverted changes in `Map`. Let me know if this is what you wanted.

HarelM:
Map inherits from camera, so ant public method in camera is also public for map. I'm trying to see if there's a way to avoid exposing this somehow...?

Samarth1696:
Maybe set the variable as `private`? I can't think of any other idea than this..

Samarth1696:
Just letting you know I will be working on hacks to add it somehow on the plugin side. Also I was wrong, even with `_elevateCameraIfInsideTerrain` being returned early the camera was elevating up above the terrain surface. I will find a way to stop that also if I can.. idk..

Sorry for any false arguments that may have been irritating for you.. I am still learning you know... :/

HarelM:
No problem.
We have discussed this issue yesterday in the monthly meeting. 
A proposal that was discussed was to fire an event when this occurs to let outside code decide what to do and how to react to this situation. 
I think this will allow better ergonomics around this and won't change the public API dramatically.

Samarth1696:
We can close this one if it is no longer needed..

HarelM:
Thanks for the update! Closing.

PR #5043, opened 2024-11-13
Vertical alignment in `format` expression

stanislawpuda-tomtom:
Related to issue: [maplibre-style-spec](https://github.com/maplibre/maplibre-style-spec/issues/832)

This PR introduces implementation of vertical align property in format expression. It enables to specify how each section should be positioned in relation to biggest element in line. There are three possible options:
- "bottom" _default_: text baseline or image bottom are in line - this is current behaviour.
- "center": image center or text center are in line.
- "top": image top and text top are in 

## How to compare

To easily compare between current behaviour and new changes we build small comparison app. Please download `test.zip`, unpack and run eg. with VSC Live Server extension.
- `maplibre-gl-dev-MAIN.js` file contains current maplibre-gl-js version. We've changed export variable to `maplibreglMain`.
- `maplibre-gl-dev.js` contains maplibre-gl-js with changes from this PR.

Download test package: 
[test.zip](https://github.com/user-attachments/files/17736654/test.zip)

## Visual comparison

https://github.com/user-attachments/assets/69016864-69e8-41f6-9b9d-2ac35ab55ee8

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

Authors: @zbigniewmatysek-tomtom @stanislawpuda-tomtom


src/symbol/shaping.ts
```diff
@@ -81,23 +81,28 @@ class SectionOptions {
     fontStack: string;
     // Image options
     imageName: string | null;
+    // Common options
+    verticalAlign: 'bottom' | 'top' | 'center';
```
HarelM, line 85: Can this use a type? Is there a type defined in the spec package or is there a need to define a type here for bottom top and center?

src/symbol/shaping.ts
```diff
@@ -81,23 +81,28 @@ class SectionOptions {
...
 
     constructor() {
         this.scale = 1.0;
         this.fontStack = '';
         this.imageName = null;
+        this.verticalAlign = 'bottom';
     }
 
-    static forText(scale: number | null, fontStack: string) {
+    static forText(scale: number | null, fontStack: string, verticalAlign: 'bottom' | 'top' | 'center' | null) {
```
HarelM, line 94: See previous comment about the type.

src/symbol/shaping.ts
```diff
@@ -769,7 +861,7 @@ function align(positionedLines: Array<PositionedLine>,
     if (maxLineHeight !== lineHeight) {
         shiftY = -blockHeight * verticalAlign - SHAPING_DEFAULT_OFFSET;
     } else {
-        shiftY = (-verticalAlign * lineCount + 0.5) * lineHeight;
+        shiftY = -verticalAlign * lineCount * lineHeight + 0.5 * lineHeight;
```
HarelM, line 864: Can you explain this change? Mathematically, they are not equivalent, obviously...

HarelM:
This needs to be verified in terms of performance, as I think this is a "hot" path.
Can you please post the benchmark scores?
Also, can you make sure the new code is covered with tests? I think codecov is complaining... (you need to wait for the build to end so that all coverage files get uploaded).

src/symbol/shaping.ts
```diff
@@ -182,9 +187,25 @@ class TaggedString {
         return this.sectionIndex.reduce((max, index) => Math.max(max, this.sections[index].scale), 0);
     }
 
+    getMaxImageSize(imagePositions: {[_: string]: ImagePosition}) {
```
HarelM, line 190: Add return type to this method please.

src/symbol/shaping.ts
```diff
@@ -585,6 +606,57 @@ function getAnchorAlignment(anchor: SymbolAnchor) {
...
+    const lineContentHeight = Math.max(maxGlyphSize, maxImageHeight * layoutTextSizeFactor);
+    const lineContentWidth = Math.max(maxGlyphSize, maxImageWidth * layoutTextSizeFactor);
+
+    return {lineContentWidth, lineContentHeight};
+}
+
+function getVerticalAlignFactor(
+    verticalAlign: VerticalAlign
+) {
+    if (verticalAlign === 'top') {
```
HarelM, line 626: [nit-picking] I think switch-case would be clear here

src/symbol/shaping.ts
```diff
@@ -585,6 +606,57 @@ function getAnchorAlignment(anchor: SymbolAnchor) {
...
+
+function getRectAndMetrics(
+    glyphPosition: GlyphPosition,
+    glyphMap: {
+        [_: string]: {
+            [_: number]: StyleGlyph;
+        };
+    },
+    section: SectionOptions, codePoint: number
+) {
```
HarelM, line 643: Add return type to method signature.

src/symbol/shaping.ts
```diff
@@ -585,6 +606,57 @@ function getAnchorAlignment(anchor: SymbolAnchor) {
...
+    let rect: Rect = null;
+    let metrics: GlyphMetrics = null;
```
HarelM, line 645: I would omit declaring these variable and simply return the relevant pair in the relevant if statement, as these are not used elsewhere besides in the return value.

src/symbol/shaping.ts
stanislawpuda-tomtom: Yes, there is a type in MapLibre Style Spec. Updated!

src/symbol/shaping.ts
stanislawpuda-tomtom: Fixed.

src/symbol/shaping.ts
```diff
@@ -607,38 +679,42 @@ function shapeLines(shaping: Shaping,
     layoutTextSizeThisZoom: number) {
 
     let x = 0;
-    let y = SHAPING_DEFAULT_OFFSET;
+    let y = 0;
```
HarelM, line 682: I'm not sure I fully follow all the logic here unfortunatelly, but I think this method can be split into smaller methods, similar to what you did with the other small parts.
I think `baselineOffset` calculation can be a candidate to this sort of split...?

src/symbol/shaping.ts
```diff
@@ -739,9 +830,8 @@ function shapeLines(shaping: Shaping,
 function justifyLine(positionedGlyphs: Array<PositionedGlyph>,
     start: number,
     end: number,
-    justify: 1 | 0 | 0.5,
-    lineOffset: number) {
-    if (!justify && !lineOffset)
+    justify: 1 | 0 | 0.5) {
+    if (!justify)
```
HarelM, line 834: Does this if means `justify === 0`? because it looks like it can't be null, can it?

HarelM:
I've added a few minor commnet, this looks good otherwise assuming it doesn't have a performance impact.
If this does affect performance, I would consider moving some code only to places where use a different vertical alignment than baseline.

src/symbol/shaping.ts
stanislawpuda-tomtom: They are equivalent: `(a + b) * x = a * x + b * x` - or I missed sth?

stanislawpuda-tomtom: It was easier to understand for me that way because it is now: minus verticalHeight * height of all lines + half of one line.

HarelM: Or, yeah, sorry, my bad. then why is the change needed? readability?

HarelM: Ohh, OK.

HarelM:
Please note that there are render test failing, with a good error message I must add :-)

HarelM:
CC: @louwers please note that there are some render tests here that won't match native due to this new feature.
Please let us know if they need to be placed in a different directory or otherwise this is fine as is. Thanks!

src/symbol/shaping.ts
stanislawpuda-tomtom: You're correct, changed to `justify === 0`.

src/symbol/shaping.ts
stanislawpuda-tomtom: Good point, I refactored this fn, it is now easier to read.

src/symbol/shaping.ts
stanislawpuda-tomtom: Done.

src/symbol/shaping.ts
stanislawpuda-tomtom: Done.

src/symbol/shaping.ts
```diff
@@ -585,6 +606,55 @@ function getAnchorAlignment(anchor: SymbolAnchor) {
...
+    line: TaggedString,
+    layoutTextSizeFactor: number
+) {
+    const maxGlyphSize = line.getMaxScale() * ONE_EM;
+    const {maxImageWidth, maxImageHeight} = line.getMaxImageSize(imagePositions);
+
+    const lineContentHeight = Math.max(maxGlyphSize, maxImageHeight * layoutTextSizeFactor);
+    const lineContentWidth = Math.max(maxGlyphSize, maxImageWidth * layoutTextSizeFactor);
+
+    return {lineContentWidth, lineContentHeight};
```
HarelM, line 620: Is there a `size` object/type you can return instead?

src/symbol/shaping.ts
```diff
@@ -675,36 +749,49 @@ function shapeLines(shaping: Shaping,
                 const size = imagePosition.displaySize;
                 // If needed, allow to set scale factor for an image using
                 // alias "image-scale" that could be alias for "font-scale"
-                // when FormattedSection is an image section.
-                section.scale = section.scale * ONE_EM / layoutTextSizeThisZoom;
+                // when FormattedSection is an image section.A
```
HarelM, line 752: `section.A`?

src/symbol/shaping.ts
```diff
@@ -713,24 +800,26 @@ function shapeLines(shaping: Shaping,
...
     const {horizontalAlign, verticalAlign} = getAnchorAlignment(textAnchor);
-    align(shaping.positionedLines, justify, horizontalAlign, verticalAlign, maxLineLength, maxLineHeight, lineHeight, height, lines.length);
+    align(shaping.positionedLines, justify, horizontalAlign, verticalAlign, maxLineLength, maxLineHeight, lineHeight, y, lines.length);
 
-    shaping.top += -verticalAlign * height;
-    shaping.bottom = shaping.top + height;
+    // Calculate the bounding box
+    // shaping.top & shaping.left already include text offset (text-radial-offset or text-offset)
+    shaping.top += -verticalAlign * y;
+    shaping.bottom = shaping.top + y;
```
HarelM, line 822: This is a 100+ lines method, can we split it a bit more?

src/symbol/shaping.ts
HarelM: Preferably a size type or something...

HarelM:
I see that the coverage for the switch-case isn't great, can you check that please?

src/symbol/shaping.ts
stanislawpuda-tomtom: Some leftover - removed.

src/symbol/shaping.ts
stanislawpuda-tomtom: I split this fn into two, which return a number and their names are much more descriptive.

stanislawpuda-tomtom: I had to revert that commit - place where it was calculated was important, as some of the value changes later in the code.

stanislawpuda-tomtom: I improved naming of the return type:
- `lineContentWidth` -> `verticalLineContentWidth`,
- `lineContentHeight` -> `horizontalLineContentHeight`.

I don't think there is a type that fits here, I think object with descriptive props are the best option here.

src/symbol/shaping.ts
stanislawpuda-tomtom: Return type is derived from fn return.
Regarding the type itself - I considered image `Size` type, but I'm not sure if it is a good fit. Here I don't return size of a single image, but max width and max height of any image - those two values might be from two different images. I think `maxImageWidth` and `maxImageHeight` describe better what is returned.

src/symbol/shaping.ts
stanislawpuda-tomtom: I split `shapeLines` function - I moved `shapeImageSection` and `shapeTextSection` outside. I think it is better that way, because at least it is clear which values are taken and output from those code fragments.

In general I think this code requires some more serious refactoring - but at least it is a bit better now. Such refactoring should rather happen as a separate PR.

src/symbol/shaping.ts
stanislawpuda-tomtom: Finally I created a type: `LineShapingSize`.

stanislawpuda-tomtom:
Below I post the benchmark results:
LayerSymbol
LayerSymbolWithIcons
LayerSymbolWithSortKey
LayerTextWithVariableAnchor

src/symbol/shaping.ts
HarelM: Better to have this explicitly defined instead of implicitly.

src/symbol/shaping.ts
```diff
@@ -585,6 +606,73 @@ function getAnchorAlignment(anchor: SymbolAnchor) {
     return {horizontalAlign, verticalAlign};
 }
 
+type LineShapingSize = {
```
HarelM, line 609: Can you move the type definition to the begging of the file please?

src/symbol/shaping.ts
```diff
@@ -713,35 +782,114 @@ function shapeLines(shaping: Shaping,
...
-    shaping.bottom = shaping.top + height;
+    // Calculate the bounding box
+    // shaping.top & shaping.left already include text offset (text-radial-offset or text-offset)
+    shaping.top += -verticalAlign * y;
+    shaping.bottom = shaping.top + y;
     shaping.left += -horizontalAlign * maxLineLength;
     shaping.right = shaping.left + maxLineLength;
 }
 
+type ShapingSectionAttributes = {
```
HarelM, line 809: Move this to the beggining of the file please.

src/symbol/shaping.ts
```diff
@@ -607,104 +695,85 @@ function shapeLines(shaping: Shaping,
...
+        const lineShapingSize = calculateLineContentSize(imagePositions, line, layoutTextSizeFactor);
+
         for (let i = 0; i < line.length(); i++) {
             const section = line.getSection(i);
             const sectionIndex = line.getSectionIndex(i);
             const codePoint = line.getCharCode(i);
+
             let baselineOffset = 0.0;
-            let metrics = null;
+            let metrics: GlyphMetrics | null = null;
```
HarelM, line 732: I would consider replacing these 3 parameters with a single one of type `ShapingSectionAttributes` - this way there's no need to copy the information into these 3, unless they are needed from a loop perspective, but that can be solved as well, I guess.

src/symbol/shaping.ts
```diff
@@ -607,104 +695,85 @@ function shapeLines(shaping: Shaping,
...
-                    const glyph = glyphs && glyphs[codePoint];
-                    if (!glyph) continue;
-                    metrics = glyph.metrics;
-                }
-
-                // We don't know the baseline, but since we're laying out
-                // at 24 points, we can calculate how much it will move when
-                // we scale up or down.
-                baselineOffset = (lineMaxScale - section.scale) * ONE_EM;
+                const sectionAttributes = shapeTextSection(section, codePoint, vertical, lineShapingSize, glyphMap, glyphPositions);
```
HarelM, line 737: This looks a lot better, thanks!

HarelM:
Added a few nitpicking, otherwise looks good, thanks!

src/symbol/shaping.ts
```diff
@@ -607,104 +695,85 @@ function shapeLines(shaping: Shaping,
...
-                verticalAdvance = metrics.advance;
-
-                // Difference between height of an image and one EM at max line scale.
-                // Pushes current line down if an image size is over 1 EM at max line scale.
-                const offset = vertical ? size[0] * section.scale - ONE_EM * lineMaxScale :
-                    size[1] * section.scale - ONE_EM * lineMaxScale;
-                if (offset > 0 && offset > lineOffset) {
-                    lineOffset = offset;
-                }
+                section.scale = section.scale * layoutTextSizeFactor;
```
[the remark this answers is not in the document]
stanislawpuda-tomtom: From technical perspective yes, but in my opinion we shouldn't do that. Modifying `section.scale` is a side effect, and if we want to improve code readability, it would be better to keep it outside for two reasons:
- It is easier to understand what fn does if it either return some value, or modifies something. If it does both, then readability suffers.
- From `shapeLines` it is clear right away, that the value of `section.scale` was changed

src/symbol/shaping.ts
stanislawpuda-tomtom: Done.

src/symbol/shaping.ts
stanislawpuda-tomtom: Done.

src/symbol/shaping.ts
stanislawpuda-tomtom: Done.

src/symbol/shaping.ts
stanislawpuda-tomtom: Done, I have added a type.

stanislawpuda-tomtom:
I checked performance benchmarks once again.
LayerSymbol
LayerSymbolWithIcons
LayerSymbolWithSortKey
LayerTextWithVariableAnchor

src/symbol/shaping.ts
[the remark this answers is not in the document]
HarelM: Fair enough.

src/symbol/shaping.ts
```diff
@@ -607,104 +705,76 @@ function shapeLines(shaping: Shaping,
...
+
+                sectionAttributes = shapeImageSection(section, vertical, lineMaxScale, lineShapingSize, imagePositions);
+                if (!sectionAttributes) continue;
```
HarelM, line 755: Is it possible to move these two lines outside the if statement by any chance?
the `if ... continue` and the imageOffest update?

src/symbol/shaping.ts
```diff
@@ -607,104 +705,76 @@ function shapeLines(shaping: Shaping,
...
+                imageOffset = Math.max(imageOffset, sectionAttributes.imageOffset);
             }
 
+            const {rect, metrics, baselineOffset} = sectionAttributes;
```
HarelM, line 759: Are these variables used later on in the method that this deconstruction is needed? Otherwise, I would recommend to simply use `rect: sectionAttributes.rect` inside the push call.

src/symbol/shaping.ts
stanislawpuda-tomtom: The only way I can see for now is:
```TypeScript
if (section.imageName) {
    shaping.iconsInText = true;
    // If needed, allow to set scale factor for an image using
    // alias "image-scale" that could be alias for "font-scale"
    // when FormattedSection is an image section.
    section.scale = section.scale * layoutTextSizeFactor;
}

const sectionAttributes: ShapingSectionAttributes = section.imageName ?
    shapeImageSection(section, vertical, lineMaxScale, lineShapingSize, imagePositions) :
    shapeTextSection(section, codePoint, vertical, lineShapingSize, glyphMap, glyphPositions);
if (!sectionAttributes) continue;

if (section.imageName) {
    imageOffset = Math.max(imageOffset, sectionAttributes.imageOffset);
}
```
and I don't like it for two reasons:
- instead of checking if section is an image once, we do it 3 times
- in my opinion readability isn't any better, in original solution we check if section is an image an perform some operations in that scope - now it is scattered.

src/symbol/shaping.ts
stanislawpuda-tomtom: `metrics` are used later on.

HarelM (approved):

test/integration/render/tests/text-field/formatted-vertical-align-line/style.json
```diff
@@ -0,0 +1,452 @@
...
+          "format",
+          ["image", "government_icon"], 
+          { "vertical-align": "bottom" },
+          " ",
+          {},
+          "Ag",
+          { "font-scale": 1.2, "vertical-align": "bottom" },
+          " ",
+          {},
+          "Baseline",
```
HarelM, line 436: "Bottom" maybe?

HarelM:
Did a last review and added a minor comment about the render test.
Looks good otherwise.

test/integration/render/tests/text-field/formatted-vertical-align-line/style.json
stanislawpuda-tomtom: Changed.

PR #1998, opened 2022-12-26
change the cluster properties on the sources option

IhsenBen:
## Launch Checklist

 - I Confirm that **my changes do not include backports from Mapbox projects** (unless with compliant license) 
 - fix issue #1384 and #429 
 -  enable users to change cluster properties on the source by adding setClusterOptions in [src/source/geojson-source.ts](https://github.com/maplibre/maplibre-gl-js/blob/v2.4.0/src/source/geojson_source.ts#L68-L346)
 - `setClusterOptions(cluster: boolean, clusterRadius?: number; clusterMaxZoom?: number})`  The value of the clusters properties.
 
Exemple:
```js
map.addSource('earthquakes', {
                type: 'geojson',
                data:
                    'https://maplibre.org/maplibre-gl-js-docs/assets/earthquakes.geojson',
                cluster: true,
            });
///////////////
         map.getSource('earthquakes').setClusterOptions(cluster:false);
        map.getSource('earthquakes').setClusterOptions(clust: true, clusterRadius: 80, clusterMaxZoom: 50 });
/////////////
```

PR #5055, opened 2024-11-17
Fix cluster radius scaling in setClusterOptions

ciscorn:
The calculation for `superclusterOptions.radius` in `GeoJSONSource.setClusterOptions` was missing the necessary scaling, resulting in incorrect values.

For reference, see the relevant initialization code:

https://github.com/maplibre/maplibre-gl-js/blob/bc432051b36bb3425a3c70cab79124698976934d/src/source/geojson_source.ts#L160  
https://github.com/maplibre/maplibre-gl-js/blob/bc432051b36bb3425a3c70cab79124698976934d/src/source/geojson_source.ts#L185

Note: `setClusterOptions` was introduced in #1998.

## Launch Checklist

 - [x] Link to related issues.

### DIFF
--- a/src/source/geojson_source.test.ts
+++ b/src/source/geojson_source.test.ts
@@ -2,6 +2,7 @@ import {Tile} from './tile';
 import {OverscaledTileID} from './tile_id';
 import {GeoJSONSource, GeoJSONSourceOptions} from './geojson_source';
 import {IReadonlyTransform} from '../geo/transform_interface';
+import {EXTENT} from '../data/extent';
 import {LngLat} from '../geo/lng_lat';
 import {extend} from '../util/util';
 import {Dispatcher} from '../util/dispatcher';
@@ -231,7 +232,7 @@ describe('GeoJSONSource#update', () => {
             sendAsync(message: ActorMessage<any>) {
                 expect(message.type).toBe(MessageType.loadData);
                 expect(message.data.geojsonVtOptions).toEqual({
-                    extent: 8192,
+                    extent: EXTENT,
                     maxZoom: 10,
                     tolerance: 4,
                     buffer: 256,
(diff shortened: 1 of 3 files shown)


HarelM:
Thanks! I would advise to add a test to this case or modify an existing one. 
Also consider making a small method out of this if it is used both for initialization and update.

ciscorn:
@HarelM
Thanks. I've updated the expected values in the existing test and refactored the logic into a separate method.

HarelM:
Thanks!
Can you add a changelog item?

ciscorn:
@HarelM  added!

src/source/geojson_source.test.ts
```diff
@@ -259,8 +260,8 @@ describe('GeoJSONSource#update', () => {
                 expect(message.data.superclusterOptions).toEqual({
                     maxZoom: 12,
                     minPoints: 3,
-                    extent: 8192,
-                    radius: 1600,
+                    extent: EXTENT,
+                    radius: 100 * EXTENT / 512,
```
HarelM, line 264: What is the hard coded 512?

src/source/geojson_source.ts
```diff
@@ -182,7 +182,7 @@ export class GeoJSONSource extends Evented implements Source {
                 maxZoom: options.clusterMaxZoom !== undefined ? options.clusterMaxZoom : this.maxzoom - 1,
                 minPoints: Math.max(2, options.clusterMinPoints || 2),
                 extent: EXTENT,
-                radius: (options.clusterRadius || 50) * scale,
+                radius: this._scaleClusterRadius(options.clusterRadius || 50),
```
HarelM, line 185: Can we use the same method to scale the `buffer` input field, and rename it accordingly?

HarelM:
Added a few nitpicking, looks good otherwise, THANKS!

src/source/geojson_source.ts
ciscorn: I renamed the method to `_pixelsToTileUnits` and applied it to scale those fields as well.

src/source/geojson_source.test.ts
ciscorn: I’ve replaced the hardcoded `512` with `source.tileSize`.

HarelM (approved):
Thanks!

PR #5072, opened 2024-11-19
WebGL matrix performance fix

### LINKED ISSUE #5070: Performance degradation in render time in 5.0.0 version
4.7.1 render time:
image

Chrome dev tools buffer usage:

image

5.0.0-pre.6 render time:

image

Chrome dev tools buffer usage:

image

Notice how buffer differs, 47 seconds and it's almost empty vs 4 seconds and it's full. Both tests are done in mercator projection.

**maplibre-gl-js version**: 5.0.0-pre6 but seems like every 5.0.0 has this issue

**browser**: Google Chrome Version 130.0.6723.117 (Official Build) (arm64)

### Steps to Trigger Behavior

 1. Open codesandbox, go to console -> performance -> start recording
 2. Move map for a second or two, stop recording
 3. Version 5.0.0 takes ~10x longer to render and fills dev tools buffer usage immediately.

### Link to Demonstration

https://codepen.io/kamil-sienkiewicz-asi/pen/qBeGYaJ?editors=1010

### Expected Behavior

Render time takes around ~1ms with basic layers from demotiles

### Actual Behavior

Render time takes from ~9ms to ~30ms with same layers from demotiles

kubapelc:
Fixes #5070 by passing float32 matrices to WebGL instead of float64. Note that the performance degradation only seemed to affect Chrome, not Firefox. Updates relevant unit & render tests.


ibesora:
Nice work and quick fix! 
Is there anything we should be wary of by going back to 32-bit matrices? What was the rationale on changing those to 64?

src/geo/projection/globe_transform.ts
```diff
@@ -1197,7 +1197,7 @@ export class GlobeTransform implements ITransform {
         const fallbackMatrixScaled = createMat4f64();
         mat4.scale(fallbackMatrixScaled, projectionData.fallbackMatrix, [EXTENT, EXTENT, 1]);
 
-        projectionData.fallbackMatrix = fallbackMatrixScaled;
+        projectionData.fallbackMatrix = new Float32Array(fallbackMatrixScaled);
```
HarelM, line 1200: Does it make sense to create a 64f matrix and then save it as 32? It's it easier to create a 32 from the start or does it help precision and worth the conversion?

src/geo/projection/mercator_transform.ts
```diff
@@ -416,7 +416,7 @@ export class MercatorTransform implements ITransform {
         const tileMatrix = calculateTileMatrix(tileID, this.worldSize);
         mat4.multiply(tileMatrix, aligned ? this._alignedProjMatrix : this._viewProjMatrix, tileMatrix);
 
-        cache[posMatrixKey] = tileMatrix;
+        cache[posMatrixKey] = new Float32Array(tileMatrix); // Must be 32 bit floats, otherwise WebGL calls in Chrome get very slow.
```
HarelM, line 419: Same question as above.

HarelM:
Added a small question around object creation...
Looks good otherwise.

src/geo/projection/mercator_transform.ts
kubapelc: In this particular spot I added a cache for both float32 and float64 matrices, since float32 is used for rendering and float64 for symbol placement.

kubapelc:
Firefox benchmark:

Firefox

Chrome:

Chrome

It seems that the float32 fix works, but there is still a performance regression somewhere.

src/geo/projection/mercator_transform.ts
```diff
@@ -210,9 +210,11 @@ export class MercatorTransform implements ITransform {
     private _pixelMatrixInverse: mat4;
     private _fogMatrix: mat4;
 
-    private _posMatrixCache: {[_: string]: mat4};
-    private _fogMatrixCache: {[_: string]: mat4};
-    private _alignedPosMatrixCache: {[_: string]: mat4};
+    private _posMatrixCacheF32: Map<string, mat4> = new Map();
```
HarelM, line 213: I would advise to create an object/type that has both 32 and 64 matrices to avoid cache misalignment.

HarelM:
Codecov is not running on PRs lately for somereason, I'll see if I can make it work again.
Otherwise I left a few minor comments.

src/geo/projection/mercator_transform.ts
kubapelc: I'm not sure what you mean exactly. As in not allowing the maps to have different sets of key-value pairs by accident?

HarelM: Instead of two maps holding the same keys, have one map holding an object with two fields.

kubapelc: Fixed!

src/geo/projection/mercator_transform.ts
```diff
@@ -224,7 +224,6 @@ export class MercatorTransform implements ITransform {
             calcMatrices: () => { this._calcMatrices(); },
             getConstrained: (center, zoom) => { return this.getConstrained(center, zoom); }
         }, minZoom, maxZoom, minPitch, maxPitch, renderWorldCopies);
-        this._clearMatrixCaches();
```
HarelM, line 227: Why was this removed?

kubapelc: The caches were previously plain objects and were not initialized with a value at their declaration, the clear method assigned `{}` to each, so it was called here in the constructor to inicialize them. Now the caches are maps and are inicialized at the declaration, making this call unneeded. Also `_clearMatrixCaches` now calls `clear()` on each cache field instead of assigning a new object to it.

HarelM (approved):
THANKS!

PR #5074, opened 2024-11-19
Improve rotate UX

HarelM:
This changes the drag-rotate on desktop to be around the center and respect the distance from the center.
It also improves the roll pixel factor.

The following comment started this:
https://github.com/maplibre/maplibre-gl-js/pull/4780#discussion_r1803869111

Before:

https://github.com/user-attachments/assets/1b35b4e3-cf90-4b78-819e-b4186a850e84

After:

https://github.com/user-attachments/assets/54faaecc-501e-41dc-a4f3-a385ac29b408

 - [x] Link to related issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## main
 
 ### ✨ Features and improvements
+- ⚠️ Change drag rotate behavior to be around the center of the screen ([#5074](https://github.com/maplibre/maplibre-gl-js/pull/5074))
 - Export `Event` class ([#5016](https://github.com/maplibre/maplibre-gl-js/pull/5016))
 - Support Vertical Perspective projection ([#5023](https://github.com/maplibre/maplibre-gl-js/pull/5023))
 - _...Add new stuff here..._
--- a/src/ui/camera.ts
+++ b/src/ui/camera.ts
@@ -1139,7 +1139,7 @@ export abstract class Camera extends Evented {
         this._rotating = this._rotating || (startBearing !== bearing);
         this._pitching = this._pitching || (pitch !== startPitch);
         this._rolling = this._rolling || (roll !== startRoll);
-        this._padding = !tr.isPaddingEqual(padding as PaddingOptions);
+        this._padding = !tr.isPaddingEqual(padding);
         this._zooming = this._zooming || easeHandler.isZooming;
         this._easeId = options.easeId;
         this._prepareEase(eventData, options.noMoveStart, currently);
@@ -1163,7 +1163,8 @@ export abstract class Camera extends Evented {
         return this;
     }
 
-    _prepareEase(eventData: any, noMoveStart: boolean, currently: any = {}) {
+    _prepareEase(eventData: any, noMoveStart: boolean, 
+        currently: { moving?: boolean; zooming?: boolean; rotating?: boolean; pitching?: boolean; rolling?: boolean} = {}) {
         this._moving = true;
         if (!noMoveStart && !currently.moving) {
             this.fire(new Event('movestart', eventData));
--- a/src/ui/control/navigation_control.ts
+++ b/src/ui/control/navigation_control.ts
@@ -190,7 +190,7 @@ class MouseRotateWrapper {
         const mapRotateTolerance = map.dragRotate._mouseRotate.getClickTolerance();
         const mapPitchTolerance = map.dragRotate._mousePitch.getClickTolerance();
         this.element = element;
-        this.mouseRotate = generateMouseRotationHandler({clickTolerance: mapRotateTolerance, enable: true});
+        this.mouseRotate = generateMouseRotationHandler({clickTolerance: mapRotateTolerance, enable: true, aroundCenter: false});
         this.touchRotate = generateOneFingerTouchRotationHandler({clickTolerance: mapRotateTolerance, enable: true});
         this.map = map;
         if (pitch) {
--- a/src/ui/handler/drag_handler.ts
+++ b/src/ui/handler/drag_handler.ts
@@ -1,5 +1,5 @@
 import {DOM} from '../../util/dom';
-import type Point from '@mapbox/point-geometry';
+import Point from '@mapbox/point-geometry';
 import {DragMoveStateManager} from './drag_move_state_manager';
 import {Handler} from '../handler_manager';
 
@@ -28,7 +28,7 @@ export interface DragRollResult extends DragMovementResult {
     rollDelta: number;
 }
 
-type DragMoveFunction<T extends DragMovementResult> = (lastPoint: Point, point: Point) => T;
+type DragMoveFunction<T extends DragMovementResult> = (lastPoint: Point, currnetPoint: Point, center: Point) => T;
 
 export interface DragMoveHandler<T extends DragMovementResult, E extends Event> extends Handler {
     dragStart: (e: E, point: Point) => void;
@@ -140,13 +140,13 @@ export class DragHandler<T extends DragMovementResult, E extends Event> implemen
             return;
         }
 
-        const movePoint = point['length'] ? point[0] : point;
+        const movePoint = Array.isArray(point) ? point[0] : point;
 
         if (!this._moved && movePoint.dist(lastPoint) < this._clickTolerance) return;
         this._moved = true;
         this._lastPoint = movePoint;
 
-        return this._move(lastPoint, movePoint);
+        return this._move(lastPoint, movePoint, new Point(window.innerWidth / 2, window.innerHeight / 2));
     }
 
     dragEnd(e: E) {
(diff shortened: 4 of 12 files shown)


HarelM:
CC: @NathanMOlson FYI.

birkskyum (approved):
It definitely behaves more like I'd expect it to at high pitch, where it was opposite before - thanks!

kubapelc:
I built maplibre from the latest main without checking new changes, so I had no idea this was implemented, and this was my first time experience: I noticed that map rotation behaviour felt very weird and seemingly inverted in the X axis. I could not figure out what was going on and how to rotate the map the way I wanted, the rotation seemingly had no relation to mouse movement, it accelerated randomly, and then stopped moving at some point. I thought it was a bug before finding this PR.

This new behaviour feels very unintuitive for me, especially since the rotation seems to depend on relative position to screen center, an invisible and somewhat arbitrary point. Both google maps and earth use the old way, and that feels much more consistent with mouse movement and simple to use for me.

When the mouse crosses near the screen center, the map flicks upside down:

https://github.com/user-attachments/assets/ea0e2ff7-e2b1-43c3-833b-3d568f18ace4

I might very well be an outlier, but it may be good to see how more people feel about this change. Maybe posting a link to a demo with the new behaviour in the OSM slack would work?

HarelM:
@kubapelc I've opened the following PR to address your concern:
- #5104 

Please let me know if you think this is enough.

zta6:
Just chiming in to agree with @kubapelc in saying that this doesn't feel like an improvement to me. I assumed it was a bug too, and after playing with it for 20 minutes I still can't get it to do what I want. 

I understand that any UX change results in some change aversion but I don't think that's what I'm feeling unfortunately. 

With the previous controls the mental model was simple. Move the mouse laterally and the map will rotate. Move the mouse vertically and the map will pitch. 

With these new controls I can't tell what inputs are required to get the map to turn the way that I want it to turn. I understand what you were going for with 'grabbing' a piece of terrain and moving it. But I don't think it really works in 3 dimensions, and even if it would work I'm not sure you've achieved the 'grabbing' feeling you were looking for. Even just watching your video - your mouse quickly separates itself from the location you clicked or 'grabbed'. 

As an example of the complexity of this mental model - Try to rotate the map 360 degrees in one continuous gesture without drastically changing the pitch. I can't do it for the life of me. These controls don't feel orthogonal to me and non-orthogonal controls don't tend to be considered easy to use.

zta6:
>Dragging the map vertically pitches the map exactly as before, dragging the map horizontally rorates the map,
Except that things break when you put your cursor near the vertical center of the map, because the code can't decide which direction a drag should make the map rotate so it ends up rotating the map an unpredictable amount in an unpredictable direction. And IMO near the centerline is a pretty natural place to put your cursor when attempting to rotate the map. Even if it's not, crossing the vertical centerline mid-rotation is also something that's bound to happen and the rotation reversing direction when that happens doesn't make any sense. 

I can move my cursor 10cm horizontally and have the map rotate <10 degrees. And then on another attempt, or when the cursor is on a different part of the screen I can move my cursor an 2cm and have the map turn 180 degrees. As a user I can't tell if I need to move my cursor 1 cm or 10cm+ to achieve the outcome I'm looking for. 

https://github.com/user-attachments/assets/7b210752-c757-4a58-9d86-1a944a8db50e

birkskyum:
If the change brings improvements to some cases, and regressions in others, then it's probably safest to revert it until we know how to only target the behavior that was problematic before.

gustavo80br:
Hey folks, arriving late to the discussion. I had to change to v5.0.0-pre releases, to take the improvement #3854, which is really necessary. I had to stop at v5.0.0-pre.6 because this rotation change is creating multiple undesired behaviors.

If you rotate the map around the vertical middle of the screen, while you are moving the mouse to the same direction, rotation changes to the other side for a reason I was not able to understand why. I suggestion that his feature to have a toggle, so that people can choose to turn it on or off, because for me this broke the lib completely, if this stays on v5 I will not be able to upgrade.

HarelM:
The before and after videos at the top pf this thread explains very well, from my point of view, why this change was introduced.
Moving a map, that rotates around the center, to one direction while the rotation of the map is in opposite direction to the mouse movement is a bad user experience.
Reverting this PR will still keep this bad behavior, it's not a solution.
If you would like to suggest a solution, feel free to do so using a PR which addresses the issues that were solved in this PR and creates a reasonable experience as well.
Adding more configurations doesn't sound like a great solution to me, BTW...

kubapelc:
Another suggestion: revert to the old behaviour, but with the addition of hiding the cursor while the map is being rotated. This might solve the original issue of users attempting to drag-rotate the map - hiding the cursor subtly implies that its position is irrelevant to rotation controls.

HarelM:
I belive hiding the cursor breaks some accessibility rules, also not a great user experience if you'd ask me...

birkskyum:
@gustavo80br , I struggle to imagine this - can you make a small video recording of it?

gustavo80br:
@birkskyum here is the video. You can see the mouse is moving in the same direction, but the rotation direction is changing. It happens when the mouse is close to the vertical middle. The actual behavior, that's present in the latest release, I see very intuitive and with no glitches like this. When I upgraded to the pre-release 9 the first thing I noticed was this rotation behavior change and started to look for some settings to revert it back, until I found this thread and rolled back to pre-release 6, and then to pre-release 5 as other rendering glitches were happening in pre-relase 6. @HarelM will see what I can do to submit a PR with suggestions.

https://github.com/user-attachments/assets/87eb3cd7-26ca-4210-951b-1ea7a9f56fc4

gustavo80br:
I have compared MapLibre actual rotation behavior with Google Earth. It's similar, the map rotates to the same direction that MapLibre does in the production release. Just the movement in Google Earth that is less sensitive to mouse movement and has some easing when you release the mouse. The actual rotation of MapLibre was intuitive to me, I just started to use it and already was familiarized with how to move the map perspective. After the change, navigating through the map became unpredictable.

I don't feel the pain that motivated this change. My way of thinking about rotation and pitch: move mouse up and down, changes the pitch; move mouse left/right, rotates the map to one side or the other. The pitch behavior did not change but now moving mouse left/right brings some unpredictable rotation change if you are around the middle of the screen.

HarelM:
There are edge case in any definition of movement, dragging near the center is where this solution falls short.

birkskyum:
For better or worse, it's definitely an adjustment. It's interesting how maps do this differently - I haven't thought about it before:

- google maps has up/down as pitch, right/left as bearing, pivot around center. Meaning when cursor is above center the rotation is opposite of mouse movement direction. I believe it's a lot like how maplibre was before.

- apple maps has a up/down as pitch, and right/left as bearing, but the pivot point for rotation is here the point where the drag starts - meaning the curser almost never "grab"/follow the map around.

justinclift:
Yeah, developing a new mapping application over the last few weeks and this new rotation behaviour is much less straight forward to work with, as least on macOS and iOS devices.

Will need to stick with 4.7.1 for now, then take a look at v5 again in a few weeks to see if it's been fixed. :smile:

axekan:
I agree with the other comments that this is way too subjective to be the default. This behavior should at least be configurable if it isn't already (I couldn't find any settings regarding this). 

The way it reverses direction above the mid point of the map doesn't make sense at all to me and causes too much chopping in rotation.

It is the opposite of how mapbox and google maps both work.

https://github.com/user-attachments/assets/efc0c1d4-7ec3-43c8-96b9-823a9ae17abb

### Google maps vs Maplibre comparsion:

https://github.com/user-attachments/assets/d3b37f06-ee80-4362-8b29-2436ec0787cc

https://github.com/user-attachments/assets/2ba7c52b-555f-4fb6-8cad-ce23b4bcb23f

kubapelc:
I'm having to rotate the map a lot lately while developing improved tile frustum culling, and after hours of interacting with the map I still can't intuitively get it to move how I want, and judging from the discussion here I am not alone in this.

I suggest to make this new behaviour optional, and make the old behaviour default. I would like to investigate this (mostly check that the `aroundCenter` workaround doesn't have other side effects) and make a PR to implement this. Would you be okay with this @HarelM ?

HarelM:
Changing the default can be made as part of a breaking change version, which I don't see a good reason to do at the moment. 
 
I generally don't like the old behavior, so I would prefer a better fix to the current behavior instead of just complicating the user with more options and flags...

justinclift:
> Changing the default can be made as part of a breaking change version, which I don't see a good reason to do at the moment.

Would giving end users a library that works by default be a good enough reason?

Because at the moment it's not the case, so everyone using 5.x.x in any serious way seems to be using the workaround.

That being said, the current broken behaviour is clearly not broken for yourself.  Is your setup different from the current approach in some way that could explain it?  ie maybe there's some other factor that needs to be considered too?

HarelM:
This now has become a heated subject, locking this thread.

PR #5092, opened 2024-11-21
Fix unwanted roll #5083

### LINKED ISSUE #5083: Map ends up with roll when rollEnabled=false
**maplibre-gl-js version**: 5.0.0-pre.3, 5.0.0-pre.7

**browser**: Chrome

### Steps to Trigger Behavior

 1. Pitch the map to approximately pitch=0.
 2. "Throw" the map. (Give it a quick rotation so that it continues rotating when you let go of the mouse button).
 3. Before the map stops rotating, interrupt the motion by clicking on the map again
 4. Pitch the map. You will see that the map is now rolled.

### Demonstration

https://github.com/user-attachments/assets/6fdda8ea-6550-4177-934e-8fbe30f7351a

### Expected Behavior
When `rollEnabled=false`, no user mouse input causes the map to roll.

### Actual Behavior
When `rollEnabled=false`, the map ends up rolled when rotations near pitch = 0 are interrupted.

NathanMOlson:
## Launch Checklist

Fix #5083.

Use spherical linear interpolation only when `startRoll != endRoll`. Otherwise, use linear interpolation of Euler angles.

 - [x] Link to related issues.


src/ui/camera.test.ts
```diff
@@ -1535,6 +1535,16 @@ describe('#flyTo', () => {
         expect(fixedLngLat(camera.getCenter())).toEqual({lng: 100, lat: 0});
     });
 
+    test('no roll when motion is interrupted', async () => {
```
NathanMOlson, line 1538: @HarelM Any idea why this test passes when it runs as part of the complete test suite? It fails (as expected) when I run it in isolation.

src/geo/projection/camera_helper.ts
```diff
@@ -6,6 +6,7 @@ import {PaddingOptions} from '../edge_insets';
 import {LngLatBounds} from '../lng_lat_bounds';
 import {getRollPitchBearing, RollPitchBearing, warnOnce} from '../../util/util';
 import {quat} from 'gl-matrix';
+import {interpolates} from '@maplibre/maplibre-gl-style-spec';;
```
HarelM, line 9:
```suggestion
import {interpolates} from '@maplibre/maplibre-gl-style-spec';
```

src/ui/camera.ts
```diff
@@ -746,6 +746,10 @@ export abstract class Camera extends Evented {
         return this;
     }
 
+    setRollEnabled(rollEnabled: boolean) {
```
HarelM, line 749: This world become part of the public API, is this intended?
Do we have similar API methods like this?
What about symmetry of getRollEnabled?

NathanMOlson: No, it is not intended that this become part of the public API. Is there a way to prevent that? Is adding `@internal` sufficient?

HarelM: peotected should probably do it, and in terms of docs internal should do.

NathanMOlson: OK, I've made `setRollEnabled()` `protected` and `@internal`. Do you still want a `getRollEnabled()` even though there is currently no use for one?

NathanMOlson: `protected` prevents the tests from running, so I'm going to have to do something else.

src/ui/map.ts
```diff
@@ -643,6 +643,7 @@ export class Map extends Camera {
         this._trackResize = resolvedOptions.trackResize === true;
         this._bearingSnap = resolvedOptions.bearingSnap;
         this._centerClampedToGround = resolvedOptions.centerClampedToGround;
+        this.cameraHelper.setRollEnabled(resolvedOptions.rollEnabled === true);
```
HarelM, line 646: Can someone set roll value while roll is disabled? How does these two interact?

NathanMOlson: `'rollEnabled` controls whether the mouse can be used to adjust roll. Regardless of the state of `rollEnabled`, roll can always be adjusted using `jumpTo()`, `easeTo()`, and `flyTo()`.

src/ui/camera.test.ts
```diff
@@ -822,12 +822,14 @@ describe('#easeTo', () => {
 
     test('rolls to specified roll', () => {
         const camera = createCamera();
+        camera.cameraHelper.setRollEnabled(true);
```
HarelM, line 825: Not sure I understand why this is needed here in regards to what you explained in the previous comment - if roll enabled is not required, why this test needs it?

NathanMOlson: Because after this change, the interpolation behaves differently depending on `rollEnabled`, and we need to test both paths.

NathanMOlson: an alternative would be to only use spherical linear interpolation if `startRoll != endRoll` (instead of depending on `rollEnabled`)

HarelM: This sounds like a good approach - to be as similar to what it was before the introduction of roll.
Less configuration is better I believe.

src/geo/projection/camera_helper.ts
```diff
@@ -100,23 +101,31 @@ export interface ICameraHelper {
...
-    // At pitch ==0, the Euler angle representation is ambiguous. In this case, set the Euler angles
-    // to the representation requested by the caller
-    if (k < 1) {
-        const rotation: quat = new Float64Array(4) as any;
-        quat.slerp(rotation, startRotation, endRotation, k);
-        const eulerAngles = getRollPitchBearing(rotation);
-        tr.setRoll(eulerAngles.roll);
-        tr.setPitch(eulerAngles.pitch);
-        tr.setBearing(eulerAngles.bearing);
+export function updateRotation(startRotation: quat, endRotation: quat, startEulerAngles: RollPitchBearing, endEulerAngles: RollPitchBearing, tr: ITransform, k: number, useSlerp: boolean) {
```
HarelM, line 110: It might be better to define a type for this method to pass the parameters as there are many parameters now, to many for my taste...

src/ui/camera.test.ts
NathanMOlson: Done.

HarelM:
We've just migrated the code to vitest, I'm trying to figure out if the coverage reports is accurate. 
Cc: @birkskyum

birkskyum:
> We've just migrated the code to vitest, I'm trying to figure out if the coverage reports is accurate. Cc: @birkskyum

It appear to be reporting at least some code as Covered on main
https://app.codecov.io/gh/maplibre/maplibre-gl-js/blob/main/src%2Fdata%2Ffeature_position_map.ts

And uncovered in this PR
https://app.codecov.io/gh/maplibre/maplibre-gl-js/pull/5092/blob/src/data/feature_position_map.ts

Despite being identical

HarelM:
The report comment is saying that there is a missing uploaded file (8 files instead of 9).
I'm rerunning the CI now, see if it will help.

HarelM:
@NathanMOlson can you address the last code review comment so I can merge this and release a new version?

src/geo/projection/camera_helper.ts
```diff
@@ -61,6 +62,16 @@ export type FlyToHandlerResult = {
     pixelPathLength: number;
 }
 
+export type UpdateRotationArgs = {
+    startRotation: quat;
```
HarelM, line 66: Isn't this a duplicate of euler angles but in a different type? Can we infer one from the other?

NathanMOlson: We can infer the quaternion from the Euler angles (but not vice-versa), but it is computationally expensive.

HarelM: If so, can we calculate it only inside this method instead of outside?

HarelM: Can we avoid calculation when not needed?

NathanMOlson: I moved the calculation inside this function. I don't know how to avoid recalculation without using global variables or moving this function into a class with member variables, neither of which seem worth it to me.

src/geo/projection/camera_helper.ts
```diff
@@ -100,23 +109,33 @@ export interface ICameraHelper {
  * Set a transform's rotation to a value interpolated between startRotation and endRotation
  * @param startRotation - the starting rotation (rotation when k = 0)
  * @param endRotation - the end rotation (rotation when k = 1)
+ * @param startEulerAngles - the starting Euler angles.
```
HarelM, line 112: Consider moving these comments to be in the type declaration instead of here.

src/geo/projection/camera_helper.ts
```diff
@@ -100,23 +109,33 @@ export interface ICameraHelper {
...
-        tr.setPitch(eulerAngles.pitch);
-        tr.setBearing(eulerAngles.bearing);
+export function updateRotation(args: UpdateRotationArgs) {
+    if (args.useSlerp) {
+        // At pitch ==0, the Euler angle representation is ambiguous. In this case, set the Euler angles
+        // to the representation requested by the caller
+        if (args.k < 1) {
+            const startRotation = rollPitchBearingToQuat(args.startEulerAngles.roll, args.startEulerAngles.pitch, args.startEulerAngles.bearing);
+            const endRotation = rollPitchBearingToQuat(args.endEulerAngles.roll, args.endEulerAngles.pitch, args.endEulerAngles.bearing);
+            const rotation: quat = new Float64Array(4) as any;
```
HarelM, line 125: In previous code there was a check if quat are equal, was it removed intentionally?

HarelM:
We have recently updated the coverage reporter, I hope it's accurate, if so, the tests need to cover more of the new code.

src/util/util.ts
```diff
@@ -916,6 +916,10 @@ export type RollPitchBearing = {
     bearing: number;
 };
 
+export function rollPitchBearingEqual(a: RollPitchBearing, b: RollPitchBearing): boolean {
```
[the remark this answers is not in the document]
NathanMOlson: This defines `rollPitchBearingEqual()` as true only if the representation is exactly the same. It returns `false` if, for example, `.a.bearing == 0` and `b.bearing = 360`. It would also be reasonable to return `true` in this condition, but I would prefer a different name for such a function, maybe `rollPitchBearingEquivalent()`.

The angles are already normalized in `easeTo()`/`flyTo()`, so the choice of whether or not to normalize here doesn't currently matter.

I prefer the current version but can change it if required.

src/geo/projection/camera_helper.ts
NathanMOlson: That check occurs in `mercator_camera_helper.ts` line 162 and `globe_transform_helper.ts` line 320. I don't think it ever occurred in this file?

NathanMOlson:
> We have recently updated the coverage reporter, I hope it's accurate, if so, the tests need to cover more of the new code.

I don't think the coverage reporter is accurate. I believe the new code has 100% coverage.

HarelM (approved):
Thanks!!

HarelM:
Coverage seems to work correctly (hopefully).

Issue #5025, opened 2024-11-09
Globe: Flickering and missing layers when panning/zooming


zerda-ocm:
**maplibre-gl-js version**: 5.0.0-pre.6

**browser**: bing, chrome

### Steps to Trigger Behavior

 1. create a vector map with line layers
 3. set projection to globe 
 3. zoom or move around

### Link to Demonstration

https://jsbin.com/zusijan/edit?html,output

https://github.com/user-attachments/assets/933bf854-c75a-483b-a4c9-c51784c16840

### Expected Behavior

No flickering and complete drawing of layers.

### Actual Behavior

This bug is apparently present since the merge of https://github.com/maplibre/maplibre-gl-js/issues/3983

birkskyum:
It says `pmtiles` not defined. Maybe there's a script tag missing?

HarelM:
You have linked an issue and not a PR. The relevant PR is:
https://github.com/maplibre/maplibre-gl-js/pull/4779

Please create a minimal reproduction of this issue, as the linked jsbin contains a lot of complicated logic around tile loading...

HarelM:
Cc: @NathanMOlson

zerda-ocm:
> Please create a minimal reproduction of this issue, as the linked jsbin contains a lot of complicated logic around tile loading...

https://jsbin.com/jakoxob/edit?html,console,output

HarelM:
This style still has too many layers, can you simplify it, and hopefully revive the PMTiles related code (as those can introduce noise)?

birkskyum:
as a sidenote - I don't understand where the js bin get the pmtiles script tag injected. I can execute the example in jsbin, but not locally yet, which makes it harder for me to debug

birkskyum:
It looks a lot like z-fighting to me.

birkskyum:
Do we see issues only with the `line` layer?

zerda-ocm:
> This style still has too many layers, can you simplify it, and hopefully revive the PMTiles related code (as those can introduce noise)?

I will try to replicate this with some geojson, but that will take some time.

zerda-ocm:
> Do we see issues only with the `line` layer?

I haven't seen this issue with other layer types.

russss:
I'm also seeing this issue on OpenInfraMap, which isn't using PMTiles. I believe it's only happening on line layers.

birkskyum:
I do see some flickering in `mercator` projection too:

[...]

PR #5094, opened 2024-11-21
Fix flickering on lines

ibesora:
This PR fixes line flickering as seen in https://github.com/maplibre/maplibre-gl-js/issues/5025
The problem was that we were calling a function that changed the stencil value used even if we were not using the result of it. Wrapping everything with an `if` block made it work

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


src/render/painter.ts
```diff
@@ -377,6 +377,9 @@ export class Painter {
      * mask area of tile overlapped by children tiles.
      * Stencil ref values continue range used in _tileClippingMaskIDs.
      *
+     * Attention: This function changes this.nextStencilID even if the result of it
```
HarelM, line 380: I have little faith someone will read this comment, and that it will reflect what the method is doing in the future. 

I would consider splitting it to getting current and advancing the next or rename it to reflect that it changes the state.

test/integration/render/tests/line-overlap/style.json
```diff
@@ -0,0 +1,52 @@
+{
+  "version": 8,
+  "sources": {
+    "labels": {
+      "type": "geojson",
+      "data": "https://zerda.de/labels_small.geojson"
```
HarelM, line 6: I think it will be better to inline the relevant geojson line to prevent this tests from failing in the future if this file is not available. 
Also we probably need only one line, so it should be straight forward...
It's also worth making sure the test fails without the fix.

HarelM (approved):
Nice!

PR #5097, opened 2024-11-22
Node 20 -> 22

birkskyum:
22 has been the lts for 7 months now, so i think we're plenty safe to use it. 

https://endoflife.date/nodejs

Requires canvas v3 and jsdom v26

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
I like it that there's a need to change only in one place! Having said that, CI need more tender love and care...

birkskyum:
@HarelM , CI is a lot better now.

Basically, 2 things.

1. The "canvas" [package](https://github.com/Automattic/node-canvas) was causing trouble with installation. I see this a lot. They are solving it with their v3, of which I here use a pre-release, which doesn't require manual rebuilds of the package on isntallation. It's being used only for the Offscreen Canvas tests below: 

https://github.com/maplibre/maplibre-gl-js/blob/30ab703678718ab230786ba55b6223f575769c0c/src/util/util.test.ts#L458-L462

https://github.com/maplibre/maplibre-gl-js/blob/30ab703678718ab230786ba55b6223f575769c0c/src/util/offscreen_canvas_distorted.test.ts#L6-L14

2. npm began complaining a lot here, leaving me only two options; add the --legacy-peer-deps flags, or move to a more modern package manager like pnpm.

.github/workflows/test-all.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       - uses: actions/setup-node@v4
         with:
           node-version-file: '.nvmrc'
-      - run: npm ci
+      - run: npm ci --legacy-peer-deps
```
HarelM, line 23: This should be added in a `.npmrc` file so that when working locally this is respected as well.

birkskyum: excellent!

HarelM:
Can we replace the canvas package with a vitest package somehow? I don't like using non-official packages...

Also, what's the error you get that needs legacy peer deps flag?

birkskyum:
The error is that JSDOM has a peer dependency on "canvas": 2.x, and this upgrades to canvas 3.x, which is breaking in the sense that it dropped support for Node 16, but JSDOM will still work with it.

.nvmrc
```diff
@@ -1 +1 @@
-20.10
\ No newline at end of file
+22
```
HarelM, line 1: This might cause CI failures if kept "too open ended", I prefer a specifc version so that CI won't suddenly fail for no obvious reason.

HarelM:
Is there a JSDom version that "is compatible" with canvas 3? Maybe a pre-release of JSDom?
Otherwise I would consider waiting for official releases of these packages...

.npmrc
```diff
@@ -1,2 +1,3 @@
 engine-strict = true
 lockfileVersion = 3
+legacy-peer-deps = true
```
HarelM, line 3: Can this be removed when using skia-canvas?

HarelM:
@birkskyum what's the status for this PR? Can we push it to finalization?

birkskyum:
Status is that the [JSDOM team](https://github.com/jsdom/jsdom/pull/3795), and us as well, would prefer to move forward only when a stable v3 of node-canvas is out.

The 3.0.0-rc2 came out in june, so it appeared to progress quite slowly, but interestingly the rc3 came out only 5 days ago.
- https://github.com/Automattic/node-canvas/releases

And there are indications it could move forward any day now.
> ([maintainer](https://github.com/Automattic/node-canvas/issues/2448#issuecomment-2520347619)) I've got some time carved out this weekend to try to get 3.0 released, though.

HarelM:
Looks like canvas 3.0 was released.
The PR is failing (due to jsdom dependency I guess), but maybe there a way forward now with this PR:
- https://github.com/maplibre/maplibre-gl-js/pull/5262

birkskyum:
The Peer deps are being updated in Jsdom 

- https://github.com/jsdom/jsdom/pull/3806

birkskyum:
Jsdom 26 is out now. It should unblock this

HarelM:
I've updated jsdom and canvas, let's see if this passes now.

HarelM (approved):

PR #5099, opened 2024-11-22
Only import types where possible, and add consistent-type-imports rule

birkskyum:
If only the type is used, there's no need to import the value of the classes etc.

This adds an eslint rule for it.

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


src/data/bucket/fill_extrusion_bucket.ts
```diff
@@ -1,7 +1,8 @@
 import {FillExtrusionLayoutArray, PosArray} from '../array_types.g';
 
 import {members as layoutAttributes, centroidAttributes} from './fill_extrusion_attributes';
-import {Segment, SegmentVector} from '../segment';
+import type {Segment} from '../segment';
```
HarelM, line 4: Is there a way to configure this rule to make a single import line for these kind of situations?

eslint.config.js
```diff
@@ -43,7 +43,7 @@ export default [
         rules: {
             'no-dupe-class-members': 'off',
             '@typescript-eslint/no-dupe-class-members': ['error'],
-
+            '@typescript-eslint/consistent-type-imports': ['error'],
```
HarelM, line 46: Is it possible to add a rule to sort imports based on "from" alphabetically?
It works remove the bed to think where to add the import at I think.

birkskyum: It's possible to configure to prefer inline, which I've done ow

src/data/bucket/fill_extrusion_bucket.ts
birkskyum: It's possible to configure to prefer inline, which I've done now,

HarelM (approved):
Assuming this was done automatically... So approved.

eslint.config.js
HarelM: It won't, but lint will just make sure the imports are ordered correctly, so there's no need to ever touch that part of the file.

birkskyum: just realized sort-imports are sorting on the values of the vars imported, no the path they're imported from

PR #5100, opened 2024-11-22
add retry for flaky test

birkskyum:
I believe the CI is breaking, because we have a flaky test.

This adds a few retries - I think it'll do.

### DIFF
--- a/src/util/actor.test.ts
+++ b/src/util/actor.test.ts
@@ -95,7 +95,7 @@ describe('Actor', () => {
         expect(gotAbortSignal).toBeTruthy();
     });
 
-    test('cancel a request that must be queued will not call the method at all', async () => {
+    test('cancel a request that must be queued will not call the method at all', {retry: 3}, async () => {
         const worker = workerFactory() as any as WorkerGlobalScopeInterface & ActorTarget;
         const actor = new Actor(worker, '1');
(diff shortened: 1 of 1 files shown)


HarelM:
I haven't seen this test fail in CI, is this related to the vitest change?

birkskyum:
I just saw it failed on the main branch, and there was a {retry:3} in the vitest PR that I put there 2 months ago, and I forgot why I did that

HarelM:
I don't see any failures on main branch here: 
https://github.com/maplibre/maplibre-gl-js/actions/runs/11980723237

Also, since this is a unit test and not an integration or end to end test, I prefer to better understand what's causing the test to fail instead of adding a retry...

HarelM:
I see that this test is flaky... :-/ Saw it both locally and on a CI run.
I'll look into understanding why it is flaky...

HarelM:
The problem is that the cancel message is sent after the message already arrived to the worker, which is a race condition.
I think that with jest the promise took a bit more time to start and thus the test completed successfully and I'm assuming vitest has a slightly different mechanism. There is also some uncertainty with the mock message bus used to emulate the worker thread on the same thread, which is not emulating stuff accurately. I'll continue investigating this...

PR #5104, opened 2024-11-24
Improve rotate experience when close to center

HarelM:
## Launch Checklist

This is a continue of the following PR, where a comment mention that the experience around the center it not great.
- #5074

I have reverted to experience around the center to be as it was before my change.
Up until 100 pixels from the center.
Simplified the navigation control to use a single handler.

 - [x] Link to related issues.


kubapelc (approved):
Works much better around the center!

I'm still not a fan of this control scheme as a whole though, I can't intuitively predict how will mouse movements translate to map rotation, even after playing with the map for a while and reading about how the new rotation controls work.

HarelM:
I think only you and I have rotated the map on the desktop and are aware of this issue :-) otherwise I would've expected other people to complain about the behavior...
Thanks for the review!

HarelM:
I see now that coverage is a bit lacking, I'll see if I can solve that.

HarelM:
@kubapelc let me know if you want to review this again, I've changed the navigation control heavily...

src/ui/control/navigation_control.test.ts
```diff
@@ -84,7 +84,7 @@ describe('NavigationControl', () => {
         const navRect = navButton.getClientRects();
 
         const buttonX = (navRect.x ?? 0) + (navRect.width ?? 0) / 2;
-        const buttonY = (navRect.y ?? 0) + (navRect.height ?? 0) / 2;
+        const buttonY = (navRect.y ?? 0) + (navRect.height ?? 0) / 2 - 1;
```
ibesora, line 87: What's this `-1` doing here?
Can we add a comment so it's clear what it's doing?

src/ui/control/navigation_control.test.ts
```diff
@@ -168,7 +168,7 @@ describe('NavigationControl', () => {
...
```
ibesora, line 171: Same here

src/ui/handler/drag_move_state_manager.ts
```diff
@@ -113,3 +113,31 @@ export class OneFingerTouchMoveStateManager implements DragMoveStateManager<Touc
         return this._isOneFingerTouch(e) && this._isSameTouchEvent(e);
     }
 }
+
+export class MouseOrTouchMoveStateManager implements DragMoveStateManager<MouseEvent | TouchEvent> {
```
ibesora, line 117: This is great. Much nicer than what we had before

ibesora (approved):
I agree with @kubapelc that the handling is not great via mouse because you don't get an intuition of what will happen but this is an improvement over what we have currently.

Added a couple of minor comments but approving either way

src/ui/control/navigation_control.test.ts
HarelM: Instead of dragging from the exact center of the button where it won't change the rotation I moved the drag to be one pixel lower.

Issue #1224, opened 2022-05-04
Certain lat/lngs, zoom and styles throws error: Cannot read properties of null (reading 'point')


willheslam:
**maplibre-gl-js version**: 2.1.9

**browser**: Chrome 100.0.4896.127 on OSX

### Steps to Trigger Behavior

Certain lng lat coordinates, with certain zoom settings, e.g.
```
{
        style: 'https://api.maptiler.com/maps/basic/style.json?key=get_your_own_OpIi9ZULNHzrESv6T2vL'
        center: [4.908376, 52.366014],
        zoom: 20
}
```
cause maplibre to throw an error whilst rendering:
```
Uncaught TypeError: Cannot read properties of null (reading 'point')
```

It seems to be dependent on specific latlngs and zoom and styles.
For example, zoom 19 or 21 doesn't trigger the error, nor does 19.9 or 20.6 (but 20.5 does)
Using the 'streets' style 
`'https://api.maptiler.com/maps/streets/style.json?key=get_your_own_OpIi9ZULNHzrESv6T2vL'`
doesn't trigger the error but 'basic' does.

Here's a screenshot from jsfiddle:
 
<img width="1792" alt="Screenshot 2022-05-04 at 09 40 33" src="https://user-images.githubusercontent.com/5377213/166649175-bba11651-9b03-4754-8d68-5748ce1345a9.png">

and here's a screenshot from the app I first encounted this bug with (using react-map-gl as a layer on top of maplibre-gl)

<img width="972" alt="Screenshot 2022-05-04 at 09 43 03" src="https://user-images.githubusercontent.com/5377213/166649285-e17554b6-853e-4614-b1e1-12c452f52b93.png">

It's specifically coming from these lines:
```
    for (const glyph of placedGlyphs) {
        addDynamicAttributes(dynamicLayoutVertexArray, glyph.point, glyph.angle);
    }
```
Logging out `placedGlyphs` results in
```
[
    {
        "point": {
            "x": 12304.450428946711,
            "y": 23700.84536864106
        },
        "angle": 8.59940111024864,
        "path": [
            {
                "x": 12320,
                "y": 23684
            },
            {
                "x": 12304.450428946711,
                "y": 23700.84536864106
            }
        ]
    },
   ...
    null,
    null,
    null,
    null,
   ...
    {
        "point": {
            "x": 12352.048734854137,
            "y": 23649.28053724135
        },
        "angle": 2.316215803069055,
        "path": [
            {
[...]

PR #5118, opened 2024-11-26
update glyph placement to verify all glyphs are valid

ragnarok56:
Resolves https://github.com/maplibre/maplibre-gl-js/issues/1224

Addresses an issue with glyphs where first/last glyphs are assumed to be sufficient to place all remaining glyphs using `placeGlyphAlongLine` which _can_ return a `null` value and throws an exception later in `placeGlyphsAlongLine`.  

## Launch Checklist

 - [x] Link to related issues.
 - [x] Post benchmark scores.


HarelM:
It would be great if you could create a test for this. 
Hopefully a render test to showcase the issue and the solution. 
Let me know if you need help with wiring the test.

ragnarok56:
Trying to come up with a render test that can recreate it, but having some issues identifying the exact scenario that causes it to fail.  I don't have access to the environment this is was occurring in right now.  I'm also not certain if it 100% happens on just rendering or if it was being triggered from zooming in/out of the area and maybe that is part of the calculus involved, if so then a render test may not be sufficient by itself?

Also not at all familiar with how `placeGlyphAlongLine` works and how to trigger the error case where one of the symbols fails to be placed along a line after the first/last check is done.

HarelM:
If you can't create a test to reproduce it, it will be impossible in the future to make sure this won't break again...
Also it will make testing that this actually fixes the issue harder...

bgeihsgt:
Commenting on this as our user hit this from time to time and would love if this could get merged. Were you guys able to work out how to test this?

HarelM:
I haven't seen any updates regarding a test to showcase this issue and the relevant fix...

ragnarok56:
## background
For more context on where I've run into this issue, we're running [maplibre inside of deck.gl](https://deck.gl/docs/developer-guide/base-maps/using-with-maplibre#example-reverse-controlled) as a static basemap layer.  We use osm vector tiles and [osm bright](https://github.com/openmaptiles/osm-bright-gl-style/blob/master/style.json) style.  The error is reliably reproduced only when transitioning between zoom levels and certain features with [multiline text labels](https://github.com/openmaptiles/osm-bright-gl-style/blob/master/style.json#L1871).  We mitigated the issue by just removing the newline, so this doesn't actually occur for us (that we're aware of) anymore, see my initial [comment](https://github.com/maplibre/maplibre-gl-js/issues/1224#issuecomment-2499623512) on #1224.  When we load the same style and map data using [tileserver-gl](https://github.com/maptiler/tileserver-gl) and zoom in/out over the same features that cause the problem it works fine.  

My theory is that there is some condition where deck.gl managed zoom transitions with maplibre-gl end up triggering the condition exhibited by the issue, but maplibre-gl itself doesn't cause the scenario (at least reliably)

## what this pr is addressing
I was unable to create an integration render test that could replicate the error case.  That isn't to say that there isn't an issue, there is still a logic flaw in the code that this PR is addressing by adding a guard clause that mimics other scenarios 

 `placeGlyphAlongLine` _can_ return null
https://github.com/maplibre/maplibre-gl-js/blob/main/src/symbol/projection.ts#L839
```
        // offset does not fit on the projected line
        if (currentIndex < lineStartIndex || currentIndex >= lineEndIndex)
            return null;
```

every other invocation of this function has a guard check to see if it returned null, and ends up returning a `{notEnoughRoom: true}` result to short circuit the handling... except for:
https://github.com/maplibre/maplibre-gl-js/blob/main/src/symbol/projection.ts#L469
```
            placedGlyphs.push(placeGlyphAlongLine(fontScale * glyphOffsetArray.getoffsetX(glyphIndex), lineOffsetX, lineOffsetY, flip, symbol.segment,
                lineStartIndex, lineEndIndex, projectionContext, rotateToLine));
```
where it ends up throwing an exception later where `glyph` ends up being null
```
for (const glyph of placedGlyphs) {
        addDynamicAttributes(dynamicLayoutVertexArray, glyph.point, glyph.angle);
    }
```

Perhaps this is where a unit test is needed instead of an integration render test?  I don't understand enough of the internals to craft the test data that can force `placeGlyphAlongLine` to enter the failure case where it returns null from the starting point of a render test.

HarelM:
Sure, if a render/integration test is something we have s hard time creating, a unit test will be enough, but a test is still needed to prevent this from happening in the future.

HarelM:
The relevant tests were added using this PR to the fork here:
- https://github.com/maplibre/maplibre-gl-js/pull/5866
- https://github.com/ragnarok56/maplibre-gl-js/pull/1

HarelM (approved):

Issue #5133, opened 2024-11-28
map.easeTo does not save padding in globe projection


kamil-sienkiewicz-asi:
I want to be able to smoothly set padding for map (use case where im opening panel from the left side and want globe to be centered in visible space between floating HTML elements), it works fine in mercator projection, however in globe projection seems like `easeTo` does nothing, padding remains the same.

**maplibre-gl-js version**: 5.0.0-pre.8

**browser**: Google Chrome Version 131.0.6778.86 (Official Build) (arm64)

### Steps to Trigger Behavior

 1. Call map.easeTo() with padding in globe projection

### Link to Demonstration
https://jsbin.com/hihisehaku/edit?html,css,js,output

![Image](https://github.com/user-attachments/assets/53580a90-654e-4862-b1a5-512da2ae66e5)

### Expected Behavior

map.easeTo({padding:{}}) will smoothly add padding to map

### Actual Behavior

map.easeTo({padding:{}}) does nothing in globe projection

pjamessteven:
possibly related to https://github.com/maplibre/maplibre-gl-js/issues/4891

Issue #4891, opened 2024-10-25
flyTo() with padding is somewhat broken with Globe


pjamessteven:
**maplibre-gl-js version**: 5.0.0-pre.3

**browser**: Firefox/Chrome/Safari

### Steps to Trigger Behavior

Use flyTo with set padding. The first time, padding will be applied properly. Now use flyTo again, with a different padding value. This time, the newly specified padding is not applied and flyTo uses the original padding value. 

In other words, above zoom level 0 and when the globe is active, you have to flyTo the same location twice to get the padding to be applied correctly. This isn't the case when globe is not active.

I can use map.setPadding, but then I don't get a smooth transition. 

### Link to Demonstration

Without Globe (expected behavior):

https://jsbin.com/rejijay/edit?html,output

With Globe (borked):

https://jsbin.com/wihekiq/edit?html,output

HarelM:
Is this related to the following discussion by any chance? 
https://github.com/maplibre/maplibre-gl-js/discussions/2984

From the above description it sounds like a regression big, but just to be on the safe side...

HarelM:
You would use the jsbin like you would use other versions, pre-releases versions are uploaded to npm.

pjamessteven:
Hey HareIM, 

Thanks for that, didn't realise the prereleases were on npm. I don't believe it's related to that issue you mentioned. I've created a simple example that demonstrates the problem here. When the globe is active, you have to flyTo the same location twice to get the padding to be applied correctly. This isn't the case when globe is not active.

Without Globe: 

https://jsbin.com/rejijay/edit?html,output

With Globe:

https://jsbin.com/wihekiq/edit?html,output

HarelM:
As far as I can tell, the issue is that globe does the padding differently than mercator.
From what I saw in the code, mercator uses `pointAtOffset` to update the padding while calling flyTo, while globe is trying to do it prior to returning the ease function, which is not the same and probably what causes the bug, it basically ignores the pointAtOffset...
https://github.com/maplibre/maplibre-gl-js/blob/dd17405c8995085c5c5132006a2a179076df7efc/src/geo/projection/globe_camera_helper.ts#L411-L418
vs
https://github.com/maplibre/maplibre-gl-js/blob/dd17405c8995085c5c5132006a2a179076df7efc/src/geo/projection/mercator_camera_helper.ts#L219-L222

Cc: @kubapelc

kamil-sienkiewicz-asi:
[...]

PR #5134, opened 2024-11-28
kamil/fix-ease-to-padding-transition-on-globe-projection

kamil-sienkiewicz-asi:
## Launch Checklist

Fixes https://github.com/maplibre/maplibre-gl-js/issues/5133

Kapture 2024-11-28 at 14 36 15

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


kamil-sienkiewicz-asi:
@HarelM Hey, soo i've tried to debug this issue that i've created today and what I noticed was that padding wasn't really interpolated on every `easeTo` step as it is in mercator projection. At first glance seems like copying solution from mercator fixed the issue 🤔 . But I'm not familiar with codebase, so if you could let me know if maybe something else need adjustment along with that PR I would be grateful 🙏

piotrszacinski (approved):

HarelM:
I remember there was an issue with adding padding, but I'm not sure what it was...
I'll see if I can find similar issues related to it.

HarelM:
I think it might be related to the following issue: 
https://github.com/maplibre/maplibre-gl-js/issues/4891

Padding is applied differently.
Can you check if the linked issue is solved/changed/the same?

HarelM:
@NathanMOlson @kubapelc can you guys take a look at this?

kamil-sienkiewicz-asi:
[Kapture 2024-11-28 at 16.17.47.webm](https://github.com/user-attachments/assets/92473bcf-deea-48b2-9c81-eef9153c9327)

@HarelM I don't see a difference between pre.3 and my branch regarding [this issue](https://github.com/maplibre/maplibre-gl-js/issues/4891) that you've linked. 
To add to that, I think padding is applied correctly (see debug padding), but position to which map should center isn't. Assumption is that consecutive calls of `flyTo` takes into consideration previous padding(?) which would explain why `flyTo` to California ends up near Hawaii. And if I go even further by flying to both of those positions over and over again without moving map it's shifting to the left and bottom on every call and eventually breaks and lands somewhere in Asia.

kamil-sienkiewicz-asi:
@HarelM Hey, I think it's ready to review, not sure why codecov/project is shouting at me tho 🤔

src/geo/projection/globe_camera_helper.ts
```diff
@@ -288,9 +289,7 @@ export class GlobeCameraHelper implements ICameraHelper {
 
         const clonedTr = tr.clone();
         clonedTr.setCenter(constrainedCenter);
-        if (doPadding) {
-            clonedTr.setPadding(options.padding);
-        }
+
```
HarelM, line 292: Is "doPadding" variable still needed?

HarelM: It is later, nevermind...

HarelM:
My main concern about this code which I think I recently had a look at is the following comment: 
https://github.com/maplibre/maplibre-gl-js/blob/e25a8d5b557edf9c78f76d4837b9ad1f6380296e/src/geo/projection/mercator_camera_helper.ts#L172

I'm not sure what to make of it and which scenarios we didn't tests as part of this PR...

kamil-sienkiewicz-asi:
@HarelM ah I see why that was needed on mercator, unfortunately I don't have enough knowledge about codebase yet to say something valuable why this works on globe out of the box, I mainly relied on tests, but I could add some more tests if you can figure out some scenarios where it would break 👌

HarelM:
If I'm looking at the mercator code I see some references to offset that can be used (although I'm not sure I understand what's the difference between offset and padding).
I'm also not very knowledgeable in this area of the code, but I would expect some complexities when setting both for globe, can you see if you can find scenarios where this might work differently between mercator and globe?

kamil-sienkiewicz-asi:
[Kapture 2024-11-29 at 16.54.33.webm](https://github.com/user-attachments/assets/19526be9-4791-4d55-8bc2-42eb9d73665e)

Mercator and globe with padding and offset are behaving the same. LngLat values that you see printed out in are unique map.getCenter values (unique to 8th decimal place) - so it doesn't behave weirdly on every consecutive call. I've also tested pitch, roll, elevation, offset, padding and bearing together, and seems like these combinations yield same `map.getCenter()` (to some decimal place, I assume that its a approximation error) in both projections

@HarelM

HarelM (approved):
Let's hope this won't create other issues, thanks!!

PR #5139, opened 2024-12-01
Facilitate for globe transition expression refactoring

### LINKED ISSUE #5114: Cannot go from `vertical-perspective` to `globe` projection
**maplibre-gl-js version**:  5.0.0-pre.7

**browser**: Chrome Version 130.0.6723.117 (Official Build) (arm64) (MacOS, M2)

### Steps to Trigger Behavior

 1. `map.setProjection({type: "vertical-perspective"})`
 2. `map.setProjection({type: "globe"})`
 3. `map.getProjection()` still shows `.type` being `"vertical-perspective"` , instead of `"globe"`

### Expected Behavior
After calling `map.setProjection({type: "vertical-perspective"})` and then `map.setProjection({type: "globe"})`, we should get a `.type` value being `"globe"` when calling `map.getProjection()`.

### Actual Behavior
After calling `map.setProjection({type: "vertical-perspective"})` and then `map.setProjection({type: "globe"})`, we still have a `.type` value of `"vertical-perspective"` when calling `map.getProjection()`.  

 This means the animation is not possible from `vertical-perspective` directly to `globe` (where the animation should actually hardly show, since both are globe)

### LINKED ISSUE #5117: Incorrect custom layer render methods argument `args.defaultProjectionData.mainMatrix` value using globe projection with high zoom level
**maplibre-gl-js version**: 5.0.0-pre.7

**browser**: Chrome Version 131.0.6778.86 (Official Build) (x86_64) (MacOS, Intel)

### Steps to Trigger Behavior

 1. Use `test/examples/globe-3d-model.html`
 2. Zoom to 3D model until globe projection change to mercator.

### Link to Demonstration

https://jsbin.com/bedahot/edit?html,output

### Expected Behavior

3D model should be visible on all zoom levels.

### Actual Behavior

3D model disappear on zoom level 12 (when projection is changed to mercator).

https://github.com/user-attachments/assets/09a67c45-609c-47ba-b7f5-d316b2c0296b

I was able to make it work using:
```
// const defaultProjectionData = args.defaultProjectionData;
const defaultProjectionData = args.defaultProjectionData.projectionTransition === 1
  ? this.map.transform.getProjectionDataForCustomLayer()
  : this.map.transform._mercatorTransform.getProjectionDataForCustomLayer();

const matrix = defaultProjectionData.mainMatrix;
```

HarelM:
## Launch Checklist

- Fixes #5114
- Fixes #5117

This is a draft PR at this point to see what still need to be fixed and for others to see.
CC: @birkskyum @ibesora @kubapelc 

The idea behind this refactoring:
1. Move all logic that is common for vertical-perspective and mercator to the transform helper.
2. Create a new class to hold all the logic of transitioning while relaying on the two transforms to do their part when called.
3. Remove the events and define the default expression for globe


src/util/util.ts
```diff
@@ -406,6 +406,16 @@ export function nextPowerOfTwo(value: number): number {
     return Math.pow(2, Math.ceil(Math.log(value) / Math.LN2));
 }
 
+/**
+ * Computes scaling from zoom level.
+ */
+export function zoomScale(zoom: number) { return Math.pow(2, zoom); }
```
birkskyum, line 412: instead of zoomScale+scaleZoom, can we call these two functions something more explicit like getScaleFromZoom?

HarelM: Sure, I don't have strong feelings either way, I simply moved them to a place that doesn't create circular references...

birkskyum: ah, it was already in the transform_helper - thought it was new functions.

birkskyum:
As far as I can tell it's very much in the right direction for the globe (being a mercator<->verticalPerspective lerp) to contain a mercatorTransform and now also a verticalPerspectiveTransform, and call into those, instead of having too much logic of it's own.

HarelM:
I think I won't be able to understand what I changed to break the globe view just from looking at the code.
I'll create a new branch and make sure I do it one step at a time hopefully without breaking stuff on the way.

HarelM:
Surprisingly I think I managed to find what created the issues in the branch after trying to recreate it in another branch...
I'll see if I can fix the remaining failing unit test, but otherwise this might be ready for review.
@kubapelc I found strange behavior in the `getProjectionData` and `getProjectionDataForCustomLayer` - both methods don't really follow the "if globe do that otherwise do what we done in mercator".
Last two commits are basically reverting the logic there, but I still find it odd.
Any chance you can take a look and let me know if there are improvements that can be done?

kubapelc:
The new `getProjectionData` on globe transform look good to me! As for `getProjectionDataForCustomLayer` it is somewhat broken in main right now, I'm fixing it to resolve #5117. But I have some render tests failing after the fix, and I haven't figured it out yet.

HarelM:
I think you can simply fix it by opening a PR to this branch, it would be better to avoid the need for me to resolve conflicts.
Is it possible to get a fallbackMatrix without the code from mercator transform? Or should I move part of the code from mercator transfrom to the helper to calculate the fallbackMatrix in case of globe/vertical-perspective?
There is also the same question for prefetchTiles and a few more other places where I couldn't find a proper solution.
If you could take a look at those it would be great.
I'm also thinking about splitting the camera helper the same way I did for the globe transform, it creates a clear separation of code and is very easy to see where issues exists this way.

src/geo/projection/vertical_perspective_transform.ts
```diff
@@ -0,0 +1,999 @@
...
+        scaleVec[1] = globeRadiusPixels;
+        scaleVec[2] = globeRadiusPixels;
+
+        // Keep a atan-correction-free matrix for transformations done on the CPU with accurate math
+        mat4.rotateX(globeMatrixUncorrected, globeMatrix, this.center.lat * Math.PI / 180.0);
+        mat4.rotateY(globeMatrixUncorrected, globeMatrixUncorrected, -this.center.lng * Math.PI / 180.0);
+        mat4.scale(globeMatrixUncorrected, globeMatrixUncorrected, scaleVec); // Scale the unit sphere to a sphere with diameter of 1
+        this._globeViewProjMatrixNoCorrection = globeMatrixUncorrected;
+
+        mat4.rotateX(globeMatrix, globeMatrix, this.center.lat * Math.PI / 180.0);
```
kubapelc, line 480: GPU atan correction is missing here. Maybe it got deleted by accident?

Originally this line was:

```
mat4.rotateX(globeMatrix, globeMatrix, this.center.lat * Math.PI / 180.0 - this._globeLatitudeErrorCorrectionRadians);
```

HarelM: I wanted to move it to the globe code and out of vertical perspective probably forgot to add it where needed.
Thanks! I'll see if there's a good way to solve this, maybe pass this parameter to the method somehow...

kubapelc: I think you can pass `GlobeProjection` object to vertical_perspective_transform and copy the code around `_globeLatitudeErrorCorrectionRadians` there too.

HarelM: I'm not sure this is the right way to "share data", I've seen this with useGlobeControls and I'm still thinking if there's a better way to handle this, maybe thought the helper, I'm not sure...
There's also a lot of duplicated code to create each transform, maybe we can expose the helper (and call it `data`, IDK) and avoid the need to wrap every getting in the helper as part of the transform code.
Not sure, I'm still thinking about it, this is a good chance to see if the design is extendable enough as this PR adds a "third" transform basically.

src/geo/projection/globe_transform.ts
```diff
@@ -278,97 +244,48 @@ export class GlobeTransform implements ITransform {
...
         this._helper = new TransformHelper({
             calcMatrices: () => { this._calcMatrices(); },
             getConstrained: (center, zoom) => { return this.getConstrained(center, zoom); }
         });
-        this._globeProjectionAllowed = globeProjectionEnabled;
-        this._globeness = globeProjectionEnabled ? 1 : 0; // When transform is cloned for use in symbols, `_updateAnimation` function which usually sets this value never gets called.
+        this._globeness = 1; // When transform is cloned for use in symbols, `_updateAnimation` function which usually sets this value never gets called.
         this._projectionInstance = globeProjection;
         this._mercatorTransform = new MercatorTransform();
+        this._verticalPerspectiveTransform = new VeritcalPerspectiveTransform();
```
ibesora, line 260: `VeritcalPerspectiveTransform` contains a typo. Should be `VerticalPerspectiveTransform`

kubapelc:
> Is it possible to get a fallbackMatrix without the code from mercator transform?

I think it makes sense to query mercator transfrom to get the fallback matrix. since it is the exact matrix which is used when mercator is active. Globe mostly needs it to do the projection transition.

> There is also the same question for prefetchTiles and a few more other places where I couldn't find a proper solution.

I think `precacheTiles` is ok in its current state, since there is no equivalent for globe transform. Mercator needs to compute a separate matrix for each tile, globe does not. (But I'm honestly not sure this precaching step even helps performance, I think maplibre did do it originally, but it was a side effect of some other function which was obsoleted and removed by the globe changes, iirc., and so I left it in.)

I'll go over all the mercator transform usages.

> I'm also thinking about splitting the camera helper the same way I did for the globe transform

Sounds good!

src/geo/projection/globe_transform.ts
```diff
@@ -278,97 +244,48 @@ export class GlobeTransform implements ITransform {
...
-        copy[2] = this._cameraPosition[2];
-        return copy;
-    }
+    public get inverseProjectionMatrix(): mat4 { return this.isGlobeRendering ? this._verticalPerspectiveTransform.inverseProjectionMatrix : this._mercatorTransform.inverseProjectionMatrix; }
 
-    get cameraToCenterDistance(): number {
-        // Globe uses the same cameraToCenterDistance as mercator.
-        return this._mercatorTransform.cameraToCenterDistance;
-    }
+    public get cameraPosition(): vec3 { return this.isGlobeRendering ? this._verticalPerspectiveTransform.cameraPosition : this._mercatorTransform.cameraPosition; }
```
ibesora, line 284: Instead of having this `this.isGlobeRendering ? ... : ...` repeated in every function, can we maybe go a further step abstracting and have something like `this._actualTransform` that changes between `this._verticalPerspectiveTransform` and `this._mercatorTransform` depending on the value of `this.isGlobeRendering`?

src/geo/projection/globe_transform.ts
```diff
@@ -455,233 +372,55 @@ export class GlobeTransform implements ITransform {
...
     getProjectionData(params: ProjectionDataParams): ProjectionData {
-        const {overscaledTileID, aligned, applyTerrainMatrix, applyGlobeMatrix} = params;
-        const data = this._mercatorTransform.getProjectionData({overscaledTileID, aligned, applyTerrainMatrix});
-
-        // Set 'projectionMatrix' to actual globe transform
-        if (this.isGlobeRendering) {
-            data.mainMatrix = this._globeViewProjMatrix32f;
-        }
+        const mercatorProjectionData = this._mercatorTransform.getProjectionData(params);
+        const vertialPerspectiveProjectionData = this._verticalPerspectiveTransform.getProjectionData(params);
```
ibesora, line 376: Typo on `vertical` again

src/geo/projection/globe_transform.ts
```diff
@@ -455,233 +372,55 @@ export class GlobeTransform implements ITransform {
...
 
     public getPixelScale(): number {
-        return lerp(this._mercatorTransform.getPixelScale(), 1.0 / Math.cos(this.getAnimatedLatitude() * Math.PI / 180), this._globeness);
+        return lerp(this._mercatorTransform.getPixelScale(), this._verticalPerspectiveTransform.getPixelScale(), this._globeness);
     }
 
     public getCircleRadiusCorrection(): number {
-        return lerp(this._mercatorTransform.getCircleRadiusCorrection(), Math.cos(this.getAnimatedLatitude() * Math.PI / 180), this._globeness);
+        // HM TODO: there was a "double" interpolation here which was removed. Check if it's needed.
+        return lerp(this._mercatorTransform.getCircleRadiusCorrection(), this._verticalPerspectiveTransform.getCircleRadiusCorrection(), this._globeness);
```
kubapelc, line 401: The double interpolation here was correct. The function `getAnimatedLatitude()` would only have an effect on latitude if the globe was moved such that the map center was very close to mercator range around +-85°, because globe allows the user to pan much closer to poles than mercator, so when transitioning to mercator the latitude is smoothly interpolated to the max latitude mercator allows.

So in this case we interpolate circle correction from globe value to mercator value, but at the same time we also interpolate globe's latitude, hence the double interpolation.

src/geo/projection/globe_transform.ts
```diff
@@ -455,233 +372,55 @@ export class GlobeTransform implements ITransform {
...
     }
 
     public getPixelScale(): number {
-        return lerp(this._mercatorTransform.getPixelScale(), 1.0 / Math.cos(this.getAnimatedLatitude() * Math.PI / 180), this._globeness);
+        return lerp(this._mercatorTransform.getPixelScale(), this._verticalPerspectiveTransform.getPixelScale(), this._globeness);
     }
 
     public getCircleRadiusCorrection(): number {
-        return lerp(this._mercatorTransform.getCircleRadiusCorrection(), Math.cos(this.getAnimatedLatitude() * Math.PI / 180), this._globeness);
+        // HM TODO: there was a "double" interpolation here which was removed. Check if it's needed.
```
ibesora, line 400: I'd say so. The center latitude changes a little bit between mercator and globe transforms.
Without the other `lerp` we are just using one of them, although in practice it should produce similar results

src/geo/projection/vertical_perspective_transform.ts
```diff
@@ -0,0 +1,999 @@
...
+    tMin: number;
+    /**
+     * The ray parameter for intersection that is "more" along the ray direction.
+     * Note that this value can be negative, meaning that this intersection occurred before the ray's origin.
+     * The intersection point can be computed as `origin + direction * tMax`.
+     */
+    tMax: number;
+} | null;
+
+export class VeritcalPerspectiveTransform implements ITransform {
```
ibesora, line 43: I think this is where most of the typos are coming from

ibesora (commented):
I like where this is going, nice work!
Added some comments but feel free to ignore them (except for the typo ones!)

src/geo/projection/globe_transform.ts
HarelM: Done in 31e592234b2dcabae7cbda983e6ae30eea662dfc, thanks for the idea!

src/geo/projection/globe_transform.ts
HarelM: Thanks! fixed in 95910ce191e5dbc457038be0dfef7ccb9a73f2cb

src/geo/projection/vertical_perspective_transform.ts
HarelM: Yup, fixed as well, thanks!

src/geo/projection/vertical_perspective_transform.ts
HarelM: I believe this was addressed in 2dcbea9fb2429b30629cd5d5ed56efa3d272658a, but I'm not sure I have a good way to test this, so let me know if you think this is the right solution...

src/geo/projection/globe_transform.ts
HarelM: A more interesting question, which I'm having a hard time finding the answer to, is how to utilize the vertical perspective and mercator transform values to do the interpolation?

HarelM:
@kubapelc I'm looking into leveraging the following code to calculate the current globeness:
```ts
type ProjectionProps = {
    type: DataConstantProperty<ProjectionDefinition>;
}

type ProjectionPossiblyEvaluated = {
    type: ProjectionDefinitionSpecification;
}

const properties: Properties<ProjectionProps> = new Properties({
    'type': new DataConstantProperty(styleSpec.projection.type as StylePropertySpecification)
});

export class ProjectionTransition extends Evented {
    properties: PossiblyEvaluated<ProjectionProps, ProjectionPossiblyEvaluated>;

    _transitionable: Transitionable<ProjectionProps>;
    _transitioning: Transitioning<ProjectionProps>;

    constructor(projection?: ProjectionSpecification) {
        super();
        this._transitionable = new Transitionable(properties);
        this.setProjection(projection);
        this._transitioning = this._transitionable.untransitioned();
        this.recalculate(new EvaluationParameters(0));
    }

    setProjection(projection?: ProjectionSpecification) {
        this._transitionable.setValue('type', projection.type);
    }

    updateTransitions(parameters: TransitionParameters) {
        this._transitioning = this._transitionable.transitioned(parameters, this._transitioning);
    }

    hasTransition() {
        return this._transitioning.hasTransition();
    }

    recalculate(parameters: EvaluationParameters) {
        this.properties = this._transitioning.possiblyEvaluate(parameters);
    }
}

//... usage:
new ProjectionTransition({
    type: ["interpolate", ["linear"], ["zoom"],
    10, "vertical-perspective", 12, "mercator"]
});
```

This leverages maplibre's expression evaluation capabilities and it looks like it is working as expected, and also removes the code related to animation and time tracking. I do see code related to updating the zoom and center when the globeness is not equal, can you elaborate on what it does?

src/geo/projection/vertical_perspective_transform.ts
kubapelc: I've tested it on my hardware that does have inaccurate atan and it works. Thanks!

src/geo/projection/globe_transform.ts
kubapelc: It was originally designed with the assumption that the map will always use a single projection, and the transition would be a very time-limited animation. The possibility for the map to be stable mid-transition breaks this. I'm not sure how to handle this...

birkskyum: My coarse understanding is that we basically have to take full control of globeness, and always adjust it from the outside with a new setter like `setGlobeness(transitionValue)`

kubapelc: No, the problem is that sometimes globe and mercator map center lnglats can differ (because mercator generally allows a smaller range of latitudes than globe). This lerp smoothly moved the globe's center (see `_computeGlobenessAnimation`) to match mercator during the transition animation, but it is unclear what to do now that there is no animation and the user can do anything to change globeness.

Also, I guess there was originally an unneeded double interpolation after all, I haven't noticed until now that `_computeGlobenessAnimation` also interpolates the center itself.

kubapelc:
> I do see code related to updating the zoom and center when the globeness is not equal, can you elaborate on what it does?

See https://github.com/maplibre/maplibre-gl-js/pull/5139#discussion_r1867808966

We need to somehow smoothly move globe's map center to match mercator, it is not clear how now that the transition is controlled by an arbitrary expression.

Maybe we can interpolate the center towards mercator every time that globeness decresases compared to the last update, and do nothing if it increases? This would make zooming in smoothly interpolate the center, and zooming out would not affect it.

src/geo/projection/globe_transform.ts
HarelM: > Does this need us to take full control of globeness, and keep running a setGlobeness(transitionValue)

I think that I'll move the code that is related to calculating the current globeness out of the transform code, it needs to use the expression evaluation logic and it fits in the projection class I think, similar to sky, light and other root style spec elements.
For every frame that globeness will be recalculated it will be passed to the transfroms which will recalculate matrices I guess.
This way the globe tranasform, will only be a wrapper around mercator and vertical perspective, without any other logic. I think it can be made to work to look good.
I'll work on it in the coming days and I hope I'll manage to create a good design and implementation for this.

src/geo/projection/globe_transform.ts
HarelM: See the following comment in @kubapelc review summary:
https://github.com/maplibre/maplibre-gl-js/pull/5139#discussion_r1867808966

HarelM:
I'll create a PR against this branch with the changes related to the globeness calculations outside the transform class so we can discuss the changes.
@ibesora @kubapelc I've left the comments above unresolved, let me know if you want me to close them as resolved or if you prefer to do it so I'd know which comments I still need to address.

src/geo/projection/globe_transform.ts
HarelM: > It was originally designed with the assumption that the map will always use a single projection, and the transition would be a very time-limited animation. The possibility for the map to be stable mid-transition breaks this. I'm not sure how to handle this...

I now come to understand this better - the expression lets the projection be in a "middle" state between mercator and vertical-perspective.
This wasn't possible until these changes that I'm introducing because the transition between the projections happened in an animation sequence and always ended up in either globeness 0 or 1, but the map was never in a state where globness is "freezed" in the middle between 0 and 1. The value was transitioning smoothly once the zoom crossed 12 to either direction.

I tried to find an equivalent expression but couldn't. having said that, I don't want to invent an expression only for this problem, so I guess we'll need to see what are the limitations of being in a "middle" of transition and what problems arise, I think it might not be as bad as it might sound, but we'll see.

HarelM:
This is now ready for review and afterwards merged to main branch.
This is almost the last change related to version 5, so after this is merged we can release version 5 (there will be other bug fixes afterwards, but no breaking changes).

ibesora (approved):

HarelM:
@Pessimistress I was about to completely remove the event of projection change here as the only place after these code change that this event is fired is when setting the projection, which is a programmatic operation, and raising an event for something you called seems weird to me...
Please let me know what event would help and when to raise it.
I'll be merging this branch shortly.

PR #5150, opened 2024-12-03
Globe custom layer fixes

### LINKED ISSUE #5117: Incorrect custom layer render methods argument `args.defaultProjectionData.mainMatrix` value using globe projection with high zoom level

kubapelc:
Fixes #5117
- fixes 3D model disappearing at high zooms when globe is used #5117 + adds a render test for this issue
- fixes 3D model disappearing during projection transition, adds internal API for setting nearZ and farZ on transforms to help with this
- fixes the `globe-custom-tiles.html` example
- fixes that some examples were using internal API, exposes the needed functions in custom layer args


src/geo/projection/globe_transform.ts
```diff
@@ -530,20 +547,16 @@ export class GlobeTransform implements ITransform {
...
-        // Note that the regular projection matrices do not need to be modified, since the rescaling happens by setting
-        // the `u_projection_tile_mercator_coords` uniform correctly.
-        const fallbackMatrixScaled = createMat4f32();
-        mat4.scale(fallbackMatrixScaled, projectionData.fallbackMatrix, [EXTENT, EXTENT, 1]);
-
-        projectionData.fallbackMatrix = fallbackMatrixScaled;
-        return projectionData;
+        const mercatorData = this._mercatorTransform.getProjectionDataForCustomLayer(applyGlobeMatrix);
+
+        if (this.isGlobeRendering) {
```
HarelM, line 552: This if can be inverted to make this method less complex a bit.

src/style/style_layer/custom_style_layer.ts
```diff
@@ -105,6 +106,21 @@ export type CustomRenderMethodInput = {
...
      */
     defaultProjectionData: ProjectionData;
+
+    /**
+     * Returns a transform matrix for rendering a 3D model seamlessly with the map when using `defaultProjectionData.mainMatrix`.
+     * Works for any projection.
+     * @param location - The location of the model.
+     * @param altitude - The altitude of the model in meters above sea level. Optional.
+     */
+    getMatrixForModel(location: LngLatLike, altitude?: number): mat4;
```
HarelM, line 116: It would be better if we could pass the relevant data here instead of passing an object with methods.
If the requirement for these methods is the examples that we created, it might not be an actual need to "commit" to this public API, and might be better to simplify the examples?

test/examples/globe-3d-model.html
```diff
@@ -129,8 +129,8 @@
 
                 // We can use this API to get the correct model matrix.
                 // It will work regardless of current projection.
-                // See MapLibre source code, files mercator_transform.ts or globe_transform.ts, function "getCustomLayerArgs".
-                const modelMatrix = map.transform.getMatrixForModel(modelOrigin, modelAltitude);
+                // See MapLibre source code, file "draw_custom.ts".
+                const modelMatrix = args.getMatrixForModel(modelOrigin, modelAltitude);
```
HarelM, line 133: Can we pass other parameters as `args` and have a more complicated calculation here instead of providing a method?

src/shaders/_projection_globe.vertex.glsl
```diff
@@ -110,8 +110,12 @@ vec4 interpolateProjection(vec2 posInTile, vec3 spherePos, float elevation) {
 vec4 interpolateProjectionFor3D(vec2 posInTile, vec3 spherePos, float elevation) {
     vec3 elevatedPos = spherePos * (1.0 + elevation / GLOBE_RADIUS);
     vec4 globePosition = u_projection_matrix * vec4(elevatedPos, 1.0);
-    vec4 fallbackPosition = u_projection_fallback_matrix * vec4(posInTile, elevation, 1.0);
-    return mix(fallbackPosition, globePosition, u_projection_transition);
+    if (u_projection_transition < 0.999) {
```
HarelM, line 113: I would consider inverting this if in both places in this file
```
if (u_projection_transition > 0.999)
   return globePosition
...
constinue here the more complicated code...
```

src/geo/transform_helper.ts
```diff
@@ -508,7 +508,8 @@ export class TransformHelper implements ITransformGetters {
             mat4.translate(m, m, [-1, -1, 0]);
             mat4.scale(m, m, [2 / this._width, 2 / this._height, 1]);
             this._pixelsToClipSpaceMatrix = m;
-            this._cameraToCenterDistance = 0.5 / Math.tan(this._fovInRadians / 2) * this._height;
+            const halfFov = this.fovInRadians / 2;
+            this._cameraToCenterDistance = 0.5 / Math.tan(halfFov) * this._height;
```
HarelM, line 512: Was there an actual change here?

src/geo/transform_interface.ts
```diff
@@ -249,6 +259,12 @@ export interface IReadonlyTransform extends ITransformGetters {
     get nearZ(): number;
     get farZ(): number;
 
+    /**
+     * Overriding values to use for near and far Z instead of what the transform would normally compute.
+     * If undefined, the transform will compute its ideal values.
+     */
+    get nearZFarZOverride(): NearZFarZ | undefined; // Note: this is intentionally left out of TransformHelper, since the override should not always be inherited to nested transforms.
```
HarelM, line 266: nearZ and farZ already have setters. Maybe it would be cleaner to have a boolean saying `allwaysCalculateNearAndFarZ`? This way when setting to true is will calculate them and when set to false it will not update them as part of the calcMatrices?
This is still a rough idea... Let me know what you think.

src/geo/projection/globe_transform.ts
```diff
@@ -228,6 +230,8 @@ export class GlobeTransform implements ITransform {
     private _projectionInstance: GlobeProjection;
     private _globeLatitudeErrorCorrectionRadians: number = 0;
 
+    private _nearZFarZOverride: NearZFarZ | undefined = undefined;
```
HarelM, line 233: I think this field should be stored in the helper if it is relevant to both transforms, which I think it is.

HarelM:
I've added some comments. Thanks!

HarelM:
I think the following PR:
- #5164 

Will cause this PR to not exactly match the changes I've made.
I can easily incorporate some of the fixes here, but I'm still having an issue with the API around setting far and near Z, I would  prefer to make the near/farZ either auto calculated or allow setting them from the outside - i.e. only add a flag to disable their calculations, what do you think?

src/geo/projection/globe_transform.ts
```diff
@@ -416,12 +425,20 @@ export class GlobeTransform implements ITransform {
         if (!this._helper._width || !this._helper._height) {
             return;
         }
-        if (this._mercatorTransform) {
-            this._mercatorTransform.apply(this, true);
-        }
         if (this._verticalPerspectiveTransform) {
             this._verticalPerspectiveTransform.apply(this, this._globeLatitudeErrorCorrectionRadians);
         }
+        if (this._mercatorTransform) {
```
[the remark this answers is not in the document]
kubapelc: Good point, indeed they will always be defined, removed.

src/geo/projection/globe_transform.ts
kubapelc: Makes sense, fixed!

src/style/style_layer/custom_style_layer.ts
kubapelc: This is a question on what kinds of custom layers should have nice helper functions and what should be left on the user to implement.

As for `getMatrixForModel` - this enables easily placing a model on a map, something I assume is common since there already were several examples doing this. While the matrix is relatively easy to compute for mercator and globe, the beauty of this API is that it will work for *any* projection, without the user needing to change anything! Alternative would be forcing the user to more or less reimplement every projection.

`getProjectionData` is somewhat new, but is very useful for the case when a user wants to draw tiles with entirely custom shaders, and also hides a lot of projection related complexity. Alternative would be having the example use internal transform API, or having the user duplicate all the complexity needed to draw a tile consitently with the rest of MapLibre.

I think both of these make a lot of sense to keep.

test/examples/globe-3d-model.html
kubapelc: (see the anwser above - https://github.com/maplibre/maplibre-gl-js/pull/5150#discussion_r1879569205 )

src/geo/transform_helper.ts
kubapelc: This is cherrypicked from your commit `0eaabba5` in `refactor-globe-transform` branch, I expected you would be merging changes from that branch anyway.

src/shaders/_projection_globe.vertex.glsl
kubapelc: Good point, fixed!

src/geo/transform_interface.ts
kubapelc: Near and far Z do not have setters as far as I know. Adding setters for them and having the default be that the underlying values are recomputed every frame feels confusing. The current API is imo much clearer: either the override is not set and and near and far Z is automatic, or it is set and near and far Z has the override values.

kubapelc:
I can help with merge conflicts with #5164 if there are any. The changes don't seem to overlap too much.

src/style/style_layer/custom_style_layer.ts
HarelM: This input object is passed on every render call of the custom layer, I would prefer to pass on the relevant matices according to the current state instead of passing a method as part of this object, is it possible?

test/examples/globe-3d-model.html
HarelM: This is the same discussion, let keep the conversation in one thread, closing this for now.

src/geo/transform_helper.ts
HarelM: Updating the branch now should remove this change I believe...?

src/geo/projection/globe_transform.ts
```diff
@@ -550,14 +548,14 @@ export class GlobeTransform implements ITransform {
...
-            const globeData = this._verticalPerspectiveTransform.getProjectionData({overscaledTileID: new OverscaledTileID(0, 0, 0, 0, 0), applyGlobeMatrix});
-            globeData.tileMercatorCoords = [0, 0, 1, 1];
-            globeData.fallbackMatrix = mercatorData.mainMatrix;
-            return globeData;
-        } else {
+        if (!this.isGlobeRendering) {
             return mercatorData;
         }
+
+        const globeData = this._verticalPerspectiveTransform.getProjectionData({overscaledTileID: new OverscaledTileID(0, 0, 0, 0, 0), applyGlobeMatrix});
```
HarelM, line 550: `_verticalPerspectiveTransform` has a `getProjectionDataForCustomLayer` method, I would advise to use it and alter the resulting object if needed instead of using a different method of vertical perspective.
Of course vertical perspective can use this method under the hood to reduce code duplication.

src/geo/transform_interface.ts
```diff
@@ -86,6 +91,8 @@ export interface ITransformGetters {
      * The distance from the camera to the center of the map in pixels space.
      */
     get cameraToCenterDistance(): number;
+
+    get nearZFarZOverride(): NearZFarZ | undefined;
```
HarelM, line 95: Is this used anywhere or added just for "symmetry"?

kubapelc: The `apply` function on TransformHelper needs to read this.

src/geo/projection/globe_transform.ts
kubapelc: Good point, fixed!

src/geo/transform_interface.ts
HarelM: Maybe we can rename `setNearFarZOveride` to `setNearFarZ`
Calling this method would internally store a boolean called `avoidNearFarZCalculation` and update the current values of `nearZ` and `farZ`.
calling `setNearFarZOvveride` with undefined will set this avoid... to false.
What I basically want to achieve here is that near and far z are stored in a single place and not in "two" places.

HarelM:
I've added a few comments.
Also there's a need to resolve conflicts on this branch.
Let me know when those can be fixed and I'll merge this to the upstream branch and merge that branch into main.

src/geo/transform_interface.ts
HarelM: See my comment here:
https://github.com/maplibre/maplibre-gl-js/pull/5150#discussion_r1879802288

src/style/style_layer/custom_style_layer.ts
kubapelc: It is not, both functions depend on map projection which cannot be described with matrices alone.

HarelM: The methods are basically wrapping the transform methods, which in turn return a matrix, or projection data object.
Why it's not possible to simply pass the return value to the these methods instead of passing a method that wraps the transform object?
I'm probably missing some basic stuff here, but still...

kubapelc: Goal is to pass a function to the custom layer, not data. We do not know in advance if and where custom layer would place a 3D model, so we cannot pass data. We also cannot really describe the function of lnglat+altitude->matrix with data alone, since for globe it involves trigonometry, for mercator it involves angular to mercator coordinate conversion and scaling. Similarly for ProjectionData - it is tile dependent, and we do not know what custom tiles the custom layer might draw. And it has very different contents, calculated in different ways depending on projection.

src/geo/transform_interface.ts
kubapelc: I like having the distinction of automatic values in nearZ/farZ and user supplied values in nearZFarZOverride, changing the sematics of nearZ/farZ with a flag feels confusing.

HarelM: Hmm... I'm still having issues with this API, but it's internal, so we can always change it in the future.
Does it make sense to move the farZ and nearZ storing to the helper? Currently only nearFarZOverride is there, which I find a bit odd.

src/style/style_layer/custom_style_layer.ts
HarelM: Ahh, I see, these are helper functions to allow asking a question about every location, not just camera center or POV.
Do you think it makes sense to add those methods to the map public API instead of here?
I would want to keep this args are pure data only and maybe allow some map API to get data on other places like we have `getElevation` on the map public API. What do you think?

kubapelc: I think the map API would be a good place for `getMatrixForModel`, but the matrix it returns is really only useful in combination with the custom layer matrix passed in args. I'm also not sure what to do with `getProjectionData` in this regard, it is also not useful outside of custom layers and is very specific to rendering custom tiles (this is why I originally made the example call the internal API).

Suggestion: we can add an object to map that contains these custom layer utility function. Something like `map.customLayerUtils.getMatrixForModel` or `map.customLayerHelper.getMatrixForModel`. This would make it clear to the user that these functions won't be of use outside custom layers and make custom layer args only contain data.

src/geo/transform_interface.ts
kubapelc: TransformHelper mostly contains values which are either the transform state or which it computes on its own, near/farZ would be outliers there, though that is not an issue.

But I feel like both ways of doing things (current "nearFarZoverride" or having a flag) are almost equivalent with no clear better choice. If you think having the flag is better in the long run, then I'll refactor it.

src/style/style_layer/custom_style_layer.ts
HarelM: I don't have a good idea honestly. 
Let's remove thos methods for now and use map.transform in the examples.
If this solves the problem we can always add these methods to the public API if the demand is strong enough and there are actual use cases that people are scared will get broken. 
Maybe they will have completely different requirements and this won't help anyway and we are only thinking this is what's needed.

src/geo/transform_interface.ts
kubapelc: Done.

src/style/style_layer/custom_style_layer.ts
kubapelc: Done.

src/geo/projection/mercator_transform.ts
```diff
@@ -541,46 +545,50 @@ export class MercatorTransform implements ITransform {
...
-        // The larger the value of nearZ is
-        // - the more depth precision is available for features (good)
-        // - clipping starts appearing sooner when the camera is close to 3d features (bad)
-        //
-        // Other values work for mapbox-gl-js but deck.gl was encountering precision issues
-        // when rendering custom layers. This value was experimentally chosen and
-        // seems to solve z-fighting issues in deck.gl while not clipping buildings too close to the camera.
-        this._nearZ = this._helper._height / 50;
+
+        if (this._helper.autoCalculateNearFarZ) {
```
HarelM, line 549: I was under the impression other variables are used later on, since this is not the case I would consider moving this to an internal method, it will be a lot more readable.

src/geo/projection/mercator_transform.ts
```diff
@@ -238,16 +235,23 @@ export class MercatorTransform implements ITransform {
...
+        this._helper.apply(that, constrain, forceOverrideZ);
     }
 
     public get cameraPosition(): vec3 { return this._cameraPosition; }
     public get projectionMatrix(): mat4 { return this._projectionMatrix; }
     public get modelViewProjectionMatrix(): mat4 { return this._viewProjMatrix; }
     public get inverseProjectionMatrix(): mat4 { return this._invProjMatrix; }
-    public get nearZ(): number { return this._nearZ; }
-    public get farZ(): number { return this._farZ; }
+    public get nearZ(): number { return this._helper.nearZ; }
```
HarelM, line 246: Move these methods to be with the other helper methods in the file

HarelM:
I'll merge this and fix what's left in the other branch, THANKS!!!

PR #5158, opened 2024-12-05
Added export to coveringTiles function

jonathanlurie:
## Launch Checklist

 - [ ] Link to related issues.
 - [x] Post benchmark scores.

## Description
Before the work on globe, this function used to be a method of the `Transform` instance, naturally exposed at `map.transform.coveringTiles(...)`, but with the internal change related to globe, it is no longer the case.  

This PR consists in exposing publically again the function `coveringTiles(...)` that exists internally. This PR does not make any change to the exposed function. 

## Goal
Exposing this function is very handy for external plugins or any application-level code that needs to know what tiles are currently on screen.

## usage

```ts
import maplibregl from "maplibre-gl";

// ... instantiate a map...

const listOfTiles = maplibregl.coveringTiles(map.transform, {tileSize: map.transform.tileSize});
// This returns a list of tile information, including the canonical IDs {z, x, y}
```

From the console, it looks like that:
Screenshot 2024-12-05 at 15 37 07

Screenshot 2024-12-05 at 15 37 53

## Benchmark
af far as I could test, it's pretty fast and it sounds reasonable to call it as part of a `.render()` method of custom layers if need be:
Screenshot 2024-12-05 at 15 42 56


HarelM:
It is advised to write some rationality to a PR to better explain what it solves...

jonathanlurie:
@HarelM apologies, it was actually a draft PR but I forgot to convert to draft...

HarelM:
The problem from my point of view that transform was not part of the official public API. alAdding this to the public API will be a maintenance overhead which I would like to keep for "hard" requirements.
So this PR needs a better justification from my point of view.

jonathanlurie:
The ´.transform´ was always a public attribute of the ´Map´ class, it just happened to not be well documented, but definitely public API by design. The coveringTiles() function (and feature) was publicly exposed and it's no longer the case, so it is a regression. I probably wouldn't be the only one complaining about this and it forces people to duplicate some internal Maplibre code in their codebase while it used to be for free

HarelM:
Nope, sorry, the trasnform was not "not well documented", it was not part of the public API by design and was removed from the docs intentionally. It was accessible, much like other fields are, but it doesn't make them part of the public API.
I'm still looking for a proper justification, the above feels more like ranting to be honest...

jonathanlurie:
Not sure how to make a justification "proper" enough without sounding like "ranting", but here is my specific use case: i am creating a custom layer that is composed of tiles (like actual webgl polygons) , in other words a tiled layer but with custom shaders. This tiled layer is fed by textures at different zoom levels and the tiling needs to follow the tiling from Maplibre so that the custom tiled layer is updated accordingly. 

With mercator only, it's doable without calling the builtin coveringTiles function because the math are quite simple, but for globe, it's a real limitation to no longer have access to this. My use case is not outstandingly original, there are many people creating custom tiled layers out there and we are all super excited to see how those will look like on globe, except the function is no longer exposed.

HarelM:
I've looked at the code again, making coveringaTiles method public also means all the input classes and interfaces are to make public which increases the maintainability constrains and increases the piblic API scope a lot. I would recommend finding something more "simple" to add to the public API instead.

jonathanlurie:
Ok, thanks for your review. I'll try to find a way to expose this feature with a minimal footprint on the public API.

PR #5159, opened 2024-12-06
Refactor `Projection` and other classes to the new concept

HarelM:
!! THIS BRANCH IS BROKEN !! 

I'll keep it a while longer until I'll make the relevant changes in the parent branch

This is a step in the same direction as 
- #5139 

It changes the projection objects to be similar to how transform was changed to accommodate for a way to split the vertical perspective and mercator logic on one hand but also unite them in a kind of wrapper.


HarelM:
This is not longer relevant, I have incorporated all the changes I wanted.

PR #5162, opened 2024-12-06
Refactor camera helper to split the logic for vertical-perspective and mercator

HarelM:
## Launch Checklist

This is in the same direction as but only for the camera helpers for this PR as the other PR was too big and too complicated.
- #5139


src/geo/projection/projection_factory.ts
```diff
@@ -36,11 +38,10 @@ export function createProjectionFromName(name: ProjectionSpecification['type']):
         }
         case 'vertical-perspective':
         {
-            const proj = new GlobeProjection();
             return {
-                projection: proj,
+                projection: new GlobeProjection(),
```
birkskyum, line 42: This should use the new VP projection class

HarelM: Sure, I want to make everything work first.

HarelM: Fixed. Thanks.

Issue #3947, opened 2024-04-08
WebGL 1 phase out


wipfli:
This is a tracking issue to discuss the removal of WebGL 1. At the moment, there are no concrete plans of doing this as far as I understand, but it is good to have a centralized place for the arguments on removing WebGL 1...

wipfli:
@HarelM you asked previously:

> Is there a way you can think of to allow moving the webgl1 code into a plugin somehow leaving the main version using only webgl2 but allowing this plugin to try and "help" in case someone needs webgl1?

I think the answer is no. To really benefit from WebGL 2 one has to write different shaders as far as I understand.

CommanderStorm:
I think this is the correct link for said feature: https://caniuse.com/webgl2
=> currently `96.1%`-`98.1%` of users have webgl2. The remaining users are in a few small clusters (0.8% IOS mobile, 0.5% IE) and in the weird browsers.

zhangyiatmicrosoft:
Wait ... I thought webgpu is the future and webgl 1 and 2 are going away anyways ...
Unless webgl2 offers significant perf boost and the migration effort is relatively small, why do we need it?

birkskyum:
We currently have superficial webgl2 support, with a webgl1 fallback, and this fallback is kinda blocking us from adopting webgl2-only features, like layout qualifiers, to begin refactoring of the codebase towards something a little closer to what webgpu will require. Thus dropping webgl1 could narrow the gap and act as a stepping stone. We will need to keep some webgl support for quite a while, so dropping webgl1 would make the adapter layer simpler.

case in point, when the "modularization" refactor started in MapLibre Native to preparing for adding Metal, the [first thing that happened](https://github.com/maplibre/maplibre-native/pull/995/files) was upgrading the OpenGL implementation from OpenGL 2 (~webgl1) to OpenGL 3+ (~webgl2) as it was easier to then abstract over Metal & OpenGL3 than Metal & OpenGL2.

birkskyum:
Related to:
- #4511

HarelM:
I've discussed this in the last open visualization summit in London with the deck.gl team, I believe the way forward world be to use luma.gl from the deck.gl stack to abstract the webgl layer to allow future supports of webgpu. This however, requires to sunset webgl 1 as they did in deck.gl as well.
They also rewrote all the shaders, which is something I would love to avoid, especially for day1. 
[...]

Issue #5135, opened 2024-11-28
Add powerPreference support to MapLibre GL


alexkn80:
**Description:**  
Currently, MapLibre GL does not allow direct configuration of the `powerPreference` attribute, which is part of the WebGL context creation settings. This feature is essential for optimizing performance and energy efficiency on various devices, such as mobile phones or laptops.  

In particular, some Intel-based MacBook Pro models experience significant performance degradation when `powerPreference` is not explicitly set to `high-performance`. Allowing developers to configure this setting directly would help mitigate such issues.  

**Proposed solution:**  
Add a new `powerPreference` option to the `Map` constructor. This option would directly influence the WebGL context creation and allow developers to optimize their application's performance based on their specific needs.  

Example usage:  

```javascript
const map = new maplibregl.Map({
  container: 'map',
  style: 'https://demotiles.maplibre.org/style.json',
  center: [0, 0],
  zoom: 2,
  powerPreference: 'high-performance', // Optional: 'default' | 'low-power'
});
```  

**Expected behavior:**  
- If `powerPreference` is provided, MapLibre GL passes it to the WebGL context during initialization.  
- If not specified, the default browser behavior (`default`) is used.  

**Benefits:**  
- Improves usability for developers by providing a clear and direct way to control GPU performance preferences.  
- Enhances support for power-sensitive devices and applications requiring high-performance rendering.  
- Mitigates performance issues on Intel-based MacBook Pro models by enabling explicit GPU preference.  

**References:**  
- [[MDN WebGLRenderingContext](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext)](https://developer.mozilla.org/en-US/docs/Web/API/WebGLRenderingContext)  
- [[WebGL Specification](https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.2)](https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.2)  
- Related closed PR: [[Add powerPreference to Map options #2213](https://github.com/maplibre/maplibre-gl-js/pull/2213)](https://github.com/maplibre/maplibre-gl-js/pull/2213)  

**Additional context:**  
Many modern web applications need to balance between performance and energy efficiency. Adding `powerPreference` as a standalone option aligns MapLibre GL with current WebGL best practices and addresses performance bottlenecks on specific hardware, such as Intel-based MacBook Pro models.

HarelM:
I belive so, yes.

[...]

PR #5165, opened 2024-12-08
[draft] refactor: remove webgl1 in favor of webgl2

0xFA11:
This is an incomplete draft PR as a follow-up to the WebGL1 removal discussion here #3947

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/build/generate-shaders.ts
+++ b/build/generate-shaders.ts
@@ -34,29 +34,7 @@ console.log(`Finished converting ${fragment.length} fragment shaders`);
 
 function glslToTs(code: string, type: 'fragment'|'vertex'): string {
     code = code
-        .trim(); // strip whitespace at the start/end
-
-    // WebGL1 Compat -- Start
-
-    if (type === 'fragment') {
-        code = code
-            .replace(/\bin\s/g, 'varying ') // For fragment shaders, replace "in " with "varying "
-            .replace('out highp vec4 fragColor;', '');
-    }
-
-    if (type === 'vertex') {
-        code = code
-            .replace(/\bin\s/g, 'attribute ') // For vertex shaders, replace "in " with "attribute "
-            .replace(/\bout\s/g, 'varying '); // For vertex shaders, replace "out " with "varying "
-    }
-
-    code = code
-        .replace(/fragColor/g, 'gl_FragColor')
-        .replace(/texture\(/g, 'texture2D(');
-
-    // WebGL1 Compat -- End
-
-    code = code
+        .trim() // strip whitespace at the start/end
         .replace(/\s*\/\/[^\n]*\n/g, '\n') // strip double-slash comments
         .replace(/\n+/g, '\n') // collapse multi line breaks
         .replace(/\n\s+/g, '\n') // strip indentation
--- a/src/render/program.ts
+++ b/src/render/program.ts
@@ -72,6 +72,7 @@ export class Program<Us extends UniformBindings> {
         }
 
         const defines = configuration ? configuration.defines() : [];
+        defines.unshift('#version 300 es\n');
         if (showOverdrawInspector) {
             defines.push('#define OVERDRAW_INSPECTOR;');
         }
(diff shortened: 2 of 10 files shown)


src/ui/map.ts
```diff
@@ -3056,9 +3056,7 @@ export class Map extends Camera {
             }
         }, {once: true});
 
-        const gl =
-            this._canvas.getContext('webgl2', attributes) as WebGL2RenderingContext ||
-            this._canvas.getContext('webgl', attributes) as WebGLRenderingContext;
+        const gl = this._canvas.getContext('webgl2', attributes) as WebGL2RenderingContext;
```
0xFA11, line 3059: This works just fine in browser but we need to also upgrade test infrastructure to WebGL2 supported headless canvas.

Currently, `vitest-webgl-canvas-mock` package does not come with WebGL2 support out of the box.
It's also somewhat old (last update ~2 years, ~4-5k weekly downloads).
We might consider `vitest-canvas-mock` as a replacement but I haven't tested whether or not it supports WebGL2 capable headless canvas.
We might also consider skia-canvas as a cairo replacement as a backend but for now, I'm just skipping these topics to isolate and focus on WebGL1 -> WebGL2 transition.

src/shaders/sky.vertex.glsl
```diff
@@ -1,5 +1,5 @@
-attribute vec2 a_pos;
+in vec2 a_pos;
```
0xFA11, line 1: Syntax changes like these should've been made in the original shader code in the first place.
Simply put, they should've been written in WebGL2 syntax and rely on WebGL1 Compact translation code.
If you were to try to run maplibre-gl-js exclusively targeting WebGL2, you'd bump into issues with these shaders written in WebGL1 syntax.
(Maybe we never actually had WebGL2 exclusive tests in order to validate all shaders, but hey, we're here now!)

src/render/program.ts
```diff
@@ -72,6 +72,7 @@ export class Program<Us extends UniformBindings> {
         }
 
         const defines = configuration ? configuration.defines() : [];
+        defines.unshift('#version 300 es\n');
```
0xFA11, line 75: This is required to be the first line in WebGL2 shaders, in every shader source code.

build/generate-shaders.ts
```diff
@@ -34,29 +34,7 @@ console.log(`Finished converting ${fragment.length} fragment shaders`);
...
-        code = code
-            .replace(/\bin\s/g, 'attribute ') // For vertex shaders, replace "in " with "attribute "
-            .replace(/\bout\s/g, 'varying '); // For vertex shaders, replace "out " with "varying "
-    }
-
-    code = code
-        .replace(/fragColor/g, 'gl_FragColor')
-        .replace(/texture\(/g, 'texture2D(');
-
-    // WebGL1 Compat -- End
```
0xFA11, line 57: Removing WebGL1 Compat translator because we would no longer need to downgrade/rewrite WebGL2 shaders in WebGL1.

src/ui/map.ts
HarelM: The only tests that should be using a real canvas are render tests, and they are running in the browser. 
The units are basically using mocks in order to run without failing, they don't use the canvas API to draw, so the unit tests should not care about the webgl2 change.

src/shaders/sky.vertex.glsl
ibesora: > If you were to try to run maplibre-gl-js exclusively targeting WebGL2, you'd bump into issues with these shaders written in WebGL1 syntax.

How come? [WebGL2 fully supports WebGL 1 shading language](https://registry.khronos.org/webgl/specs/latest/2.0/#4.3)

Although I agree we should force shaders to be written in GLSL 300

ibesora:
This seems like a good starting point on what we'd need to deprecate WebGL1.
I was going to tackle that as part of https://github.com/maplibre/maplibre-gl-js/issues/5135. 
Do you mind me doing that work?

0xFA11:
@ibesora — I think we wouldn't completely remove WebGL1 support but instead, implement a WebGL2 to WebGL1 fallback depending on what's available at runtime on the canvas, rather than a complete WebGL1 removal.

there are some conversations and context here #5166 and here #3947

PR #5166, opened 2024-12-08
fix: convert webgl1 shaders to webgl2

0xFA11:
## Launch Checklist

We are already relying on WebGL2 to WebGL1 conversion at build time (`generate-shaders.ts`).
However, some shaders are written in WebGL1 syntax but they should've been in WebGL2.
Removing WebGL1 Compat code in `glslToTs` function easily breaks maplibre because these shaders are not WebGL2 friendly atm.

 - [ ] ~Write tests for all new functionality.~


HarelM:
Is there a lint rule we can add to make sure this won't happen in the future?

0xFA11:
> Is there a lint rule we can add to make sure this won't happen in the future?

unfortunately not at the moment but I'm planning to introduce webgl2 specific integration test on puppeteer in the future when we come back with webgl2 option to test against webgl2 shader compatibility.

HarelM:
I thought about a test as well, but I wasn't sure if all the types of shaders were needed to make sure all the shaders are correctly written, which can increase the CI dramatically, which I would like to avoid. 
A sanity test can be helpful, but a linter is probably the right solution, but I can't find a good open source linter for glsl in npm...

0xFA11:
1 single additional puppeteer test case just for webgl2 sanity check wouldn't be too bad, would it? :)

HarelM:
It wouldn't hurt 😁

HarelM:
Let me know if you want to add it here, if not I'll merge this. THANKS!

0xFA11:
> Let me know if you want to add it here, if not I'll merge this. THANKS!

I think we can merge this, I wouldn't add it here because currently no way to enforce WebGL2 only shaders, we will follow-up on our discussion over #3947 this week.

PR #5186, opened 2024-12-11
Avoid constraining with Mercator transform when Globe is used

ibesora:
This fixes a subtle bug I found while trying to create a test for another issue.
When a Map is created and before the style is loaded, a Mercator transform is created as a default. When a style that uses globe is loaded, a new transform is created. This is not a problem in general but it is when setting center and zoom values that are valid when using globe and not when using Mercator.

What this PR does is avoid constraining the center and zoom values when a style is defined until the style is loaded and we can be sure we are constraining them with the correct transform.

TLDR:
Before: Notice how both the hash and the camera position change after a reload when using invalid Mercator values

https://github.com/user-attachments/assets/3ec31e6e-1e7c-4609-9574-cecadc51fbda

After: Notice how the hash and the camera are in the same position after a reload

https://github.com/user-attachments/assets/e513cfc9-d3fc-4758-add7-8ae67b4cec9e

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -725,7 +725,11 @@ export class Map extends Camera {
             }
         }
 
-        this.resize();
+        // When no style is set or it's using the default mercator projection, we can constrain the camera.
+        // When a style is set with other projections though, we can't constrain the camera until the style is loaded
+        // and the correct transform is used. Otherwise, valid points in the desired projection could be rejected
+        const shouldConstrainUsingMercatorTransform = typeof resolvedOptions.style === 'string' || !resolvedOptions.style || !resolvedOptions.style?.projection || resolvedOptions.style?.projection?.type === 'mercator';
+        this.resize(null, shouldConstrainUsingMercatorTransform);
 
         this._localIdeographFontFamily = resolvedOptions.localIdeographFontFamily;
         this._validateStyle = resolvedOptions.validateStyle;
@@ -739,6 +743,8 @@ export class Map extends Camera {
             this.addControl(new LogoControl(), resolvedOptions.logoPosition);
 
         this.on('style.load', () => {
+            // If we didn't constrain the camera before, we do it now
+            if (!shouldConstrainUsingMercatorTransform) this._resizeTransform();
             if (this.transform.unmodified) {
                 const coercedOptions = pick(this.style.stylesheet, ['center', 'zoom', 'bearing', 'pitch', 'roll']) as CameraOptions;
                 this.jumpTo(coercedOptions);
@@ -878,7 +884,7 @@ export class Map extends Camera {
      * if (mapDiv.style.visibility === true) map.resize();
      * ```
      */
-    resize(eventData?: any): Map {
+    resize(eventData?: any, constrainTransform = true): Map {
         const dimensions = this._containerDimensions();
         const width = dimensions[0];
         const height = dimensions[1];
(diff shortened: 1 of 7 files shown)


HarelM:
Let's wait a bit with merging this as I want to finalize the transform and projection changes and merge them to main. I hope that's OK.

HarelM:
The relevant PR was merged, I think you can go ahead and update this branch...

src/ui/map.ts
```diff
@@ -720,7 +720,11 @@ export class Map extends Camera {
             }
         }
 
-        this.resize();
+        // When no style is set or it's using the default mercator projection, we can constrain the camera.
+        // When a style is set with other projections though, we can't constrain the camera until the style is loaded
+        // and the correct transform is used. Otherwise, valid points in the desired projection could be rejected
+        const shouldConstrainUsingMercatorTransform = typeof resolvedOptions.style === 'string' || !resolvedOptions.style || !resolvedOptions.style?.projection || resolvedOptions.style?.projection?.type === 'mercator';
```
HarelM, line 726: Can this condition be simplified using some more `?` operator and negate the expression?

HarelM:
Is there simpler way to achieve this? It seems a bit hacky with the resize and style.load events...

src/ui/map.ts
```diff
@@ -909,6 +915,15 @@ export class Map extends Camera {
         return this;
     }
 
+    _resizeTransform() {
```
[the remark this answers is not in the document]
ibesora: Just wanted to not duplicate work being done. I think the chances of the container having changed between map load and style load are slim for example. Also we would be firing multiple events instead of one.
But it would be simpler, that's true. Let me know if that's the way I should proceed with this.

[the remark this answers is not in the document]
HarelM: Can this method be used inside `resize` method maybe?
I'm looking to remove some code duplication here...

src/ui/map.ts
```diff
@@ -720,7 +720,11 @@ export class Map extends Camera {
             }
         }
 
-        this.resize();
+        // When no style is set or it's using something other than the globe projection, we can constrain the camera.
+        // When a style is set with other projections though, we can't constrain the camera until the style is loaded
+        // and the correct transform is used. Otherwise, valid points in the desired projection could be rejected
+        const shouldConstrainUsingMercatorTransform = typeof resolvedOptions.style === 'string' || !(resolvedOptions.style?.projection?.type === 'globe');
```
HarelM, line 726: Not sure I fully follow this check:
`typeof resolvedOptions.style === 'string'`
What will happen if this points to a URL of a style that has a globe projection enabled?

ibesora: It will constrain the camera with the Mercator projection as it did before. It's either that or wait until the style has been loaded to do the first constrain of the values, but knowing it can be a URL, it might take some time until that happens.
This check is there to use the globe projection if we have all the info needed on map creation, if not, it will keep working as it was.

HarelM: Can this be further simplified to something like:
`resolvedOptions.style?.projection?.type !== 'globe'` or is typescript complaining in this case?

src/ui/map.ts
```diff
@@ -909,6 +914,15 @@ export class Map extends Camera {
         return this;
     }
 
+    _resizeTransform(constrainTransform = true) {
+        const dimensions = this._containerDimensions();
```
HarelM, line 918: Is it possible to do something like `const [width, height] = this._containerDimensions();`? This will allow removing a bit more duplication from both this method and the resize method.
This is mainly nit-picking...

src/ui/map.ts
ibesora: Yeah, it's complaining about `style` being possibly a `string` and thus not being able to access `projection`

HarelM (approved):

PR #5196, opened 2024-12-12
Add contextAttributes map option

ibesora:
This PR fixes https://github.com/maplibre/maplibre-gl-js/issues/5135 by adding a `contextAttributes` property to the map constructor that supports all [WebGLContextAttributes](https://registry.khronos.org/webgl/specs/latest/1.0/#5.2), so now you can do something like:

```js
  container: 'map',
  center: [0, 0],
  zoom: 2,
  contextAttributes: {
    powerPreference: 'default',
    antialias: true
});
```

Note that some `WebGLContextAttributes` can't be overwritten because Maplibre relies on it. These are `alpha`, `depth`, `stencil` and `premultipliedAlpha` which are always set to `true`.

This PR also sets the default WebGL `powerPreference` to `high-performance` instead of `default`, which defaults to using dedicated GPUs on multi-GPU systems.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ### ✨ Features and improvements
 - Add support for projection type expression as part of a refactoring of the transfrom and projection classes ([#5139](https://github.com/maplibre/maplibre-gl-js/pull/5139))
+- ⚠️ Support setting WebGL context options on map creation ([#5196](https://github.com/maplibre/maplibre-gl-js/pull/5196))
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### ✨ Features and improvements
 - Add support for projection type expression as part of a refactoring of the transfrom and projection classes ([#5139](https://github.com/maplibre/maplibre-gl-js/pull/5139))
-- ⚠️ Support setting WebGL context options on map creation ([#5196](https://github.com/maplibre/maplibre-gl-js/pull/5196))
+- ⚠️ Support setting WebGL context options on map creation ([#5196](https://github.com/maplibre/maplibre-gl-js/pull/5196)). Previously supported WebGL context options like `antialias`, `preserveDrawingBuffer` and `failIfMajorPerformanceCaveat` must now be defined inside the `canvasContextAttributes` object on `MapOptions`.
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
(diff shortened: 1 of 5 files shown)


ibesora:
I'm working on enabling setting which WebGL version do you want to use but since that requires some changes on how shaders are compiled, I'm working on a branch on top of this to keep changes separated

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ### ✨ Features and improvements
 - Add support for projection type expression as part of a refactoring of the transfrom and projection classes ([#5139](https://github.com/maplibre/maplibre-gl-js/pull/5139))
+- ⚠️ Support setting WebGL context options on map creation ([#5196](https://github.com/maplibre/maplibre-gl-js/pull/5196))
```
HarelM, line 5: This is a breaking change so elaborating how to migrate will be useful here.

src/ui/map.ts
```diff
@@ -3038,13 +3036,14 @@ export class Map extends Camera {
...
+        // We use the values provided in the map constructor for the rest of context attributes
         const attributes = {
+            ...this._contextAttributes,
             alpha: true,
-            stencil: true,
             depth: true,
-            failIfMajorPerformanceCaveat: this._failIfMajorPerformanceCaveat,
-            preserveDrawingBuffer: this._preserveDrawingBuffer,
-            antialias: this._antialias || false
+            stencil: true,
```
HarelM, line 3045: Maybe allow setting themn like the other parameters and issue a warning if they are not as requested?

src/ui/map.ts
```diff
@@ -113,21 +113,14 @@ export type MapOptions = {
      */
     logoPosition?: ControlPosition;
     /**
-     * If `true`, map creation will fail if the performance of MapLibre GL JS would be dramatically worse than expected
-     * (i.e. a software renderer would be used).
-     * @defaultValue false
+     * Set of WebGLContextAttributes that are applied to the WebGL context of the map.
```
HarelM, line 116: Can't we simply link to MDN docs on webglcontextattributes?

HarelM:
I've added a few comments. 
From my point of view this can be merged without the addition of "forcing" webgl2 as this PR has its merit.

src/ui/map.ts
ibesora: Most of the layer types won’t work with alpha, depth or stencil disabled. I’m not sure why anyone would do that and get broken results…

CHANGELOG.md
ibesora: Oh. don't mind. I've added it to the changelog following what was done some time ago

src/ui/map.ts
HarelM: Fair enough.

HarelM (approved):

PR #5198, opened 2024-12-12
Dual-Stack WebGL Runtime with WebGL2 to WebGL1 Fallback

0xFA11:
## Launch Checklist

As a follow-up to #3947, this PR moves WebGL2 to WebGL1 shader conversion to happen at runtime depending on the initialized WebGL context from the canvas. (This PR also builds on prior work started here #2656 originally but implements it as an actual fallback rather than forced WebGL2 to WebGL1 conversion).

Put more simply, if canvas is running in a WebGL2 environment, we do not convert original vertex and fragment shaders from WebGL2 to WebGL1 — otherwise, if canvas is running in a non-WebGL2 environment (WebGL1 assumed), we convert vertex and fragment shader sources from WebGL2 syntax to WebGL1 syntax at runtime.

On browsers with no WebGL2 support (which are <2% globally), shader conversion will happen but for the vast majority of the user base (>98% globally), we will still be able to offer WebGL2 environment and capabilities within maplibre-gl-js.

This dual-stack approach will hopefully allow more sophisticated and modern shaders written directly in maplibre stack.

 - [x] Link to related issues.
 - ~Include before/after visuals or gifs if this PR includes visual changes.~
 - ~Document any changes to public APIs.~
 - ~Post benchmark scores.~


HarelM:
Cc: @ibesora - make sure you don't do the same work as this PR does...

src/render/program.ts
```diff
@@ -82,8 +86,23 @@ export class Program<Us extends UniformBindings> {
...
 
-        const fragmentSource = defines.concat(shaders.prelude.fragmentSource, projectionPrelude.fragmentSource, source.fragmentSource).join('\n');
-        const vertexSource = defines.concat(shaders.prelude.vertexSource, projectionPrelude.vertexSource, source.vertexSource).join('\n');
+        let fragmentSource = defines.concat(shaders.prelude.fragmentSource, projectionPrelude.fragmentSource, source.fragmentSource).join('\n');
+        let vertexSource = defines.concat(shaders.prelude.vertexSource, projectionPrelude.vertexSource, source.vertexSource).join('\n');
+
+        if (!isWebGL2(gl)) {
+            // WebGL1 Compatibility
+            // Convert WebGL2 shader sources to WebGL1
+            fragmentSource = fragmentSource
```
HarelM, line 95: Can this be moved to a method or a different file? Preferably with the relevant tests to cover the scenarios here...

ibesora:
> Cc: @ibesora - make sure you don't do the same work as this PR does...

Too late @HarelM ! I'll stop that work and keep an eye on this PR

src/shaders/shaders.test.ts
```diff
@@ -0,0 +1,68 @@
...
+    test('built-in shaders should be written in WebGL2', () => {
+        const shaderFiles = globSync('./*.glsl');
+        for (const shaderFile of shaderFiles) {
+            const shaderSource = fs.readFileSync(shaderFile, 'utf8');
+            expect(shaderSource.includes('attribute')).toBe(false);
+            expect(shaderSource.includes('varying')).toBe(false);
+            expect(shaderSource.includes('gl_FragColor')).toBe(false);
+            expect(shaderSource.includes('texture2D')).toBe(false);
+        }
+    });
```
0xFA11, line 52: this should prevent people from writing shaders in WebGL1 syntax accidentally.

src/shaders/shaders.test.ts
```diff
@@ -0,0 +1,68 @@
...
+            }`;
+        const fragmentSourceWebGL1 = `
+            void main() {
+                gl_FragColor = vec4(1.0, 0.0, 0.0, 1.0);
+            }`;
+        const vertexSourceTranspiled = transpileToWebGL1('vertex', vertexSourceWebGL2);
+        const fragmentSourceTranspiled = transpileToWebGL1('fragment', fragmentSourceWebGL2);
+        expect(vertexSourceTranspiled.trim()).equals(vertexSourceWebGL1.trim());
+        expect(fragmentSourceTranspiled.trim()).equals(fragmentSourceWebGL1.trim());
+    });
```
0xFA11, line 40: we expect WebGL2 to WebGL1 transpiled shaders to be identical (after whitespace trimming).
this should prevent unexpected or accidental WebGL2->WebGL1 transpiler changes.

src/shaders/shaders.test.ts
```diff
@@ -0,0 +1,68 @@
+import {describe, test, expect} from 'vitest';
+import {transpileToWebGL1} from './shaders';
+import {globSync} from 'glob';
+import fs from 'fs';
+
+describe('Shaders', () => {
+    test('`transpileToWebGL1()` should throw for unexpected type', () => {
+        //@ts-expect-error
+        expect(transpileToWebGL1('invalid', '')).toThrowError();
+    });
```
0xFA11, line 10: we could rely on typescript enforcing types and not allowing anything other than 'vertex' | 'fragment' passed in as a parameter but since it's plain javascript at runtime, it is still possible to pass a string other than 'vertex' or 'fragment'. implementation throws an error if that's the case at runtime, this test simply tests that specific case.

src/shaders/shaders.test.ts
```diff
@@ -0,0 +1,68 @@
...
+            const vertexShaderTranspiled = transpileToWebGL1('vertex', vertexShaderSource);
+            expect(vertexShaderSource.trim()).not.equals(vertexShaderTranspiled.trim());
+        }
+        const fragmentShaderFiles = globSync('./*.fragment.glsl');
+        for (const fragmentShaderFile of fragmentShaderFiles) {
+            const fragmentShaderSource = fs.readFileSync(fragmentShaderFile, 'utf8');
+            const fragmentShaderTranspiled = transpileToWebGL1('fragment', fragmentShaderSource);
+            expect(fragmentShaderSource.trim()).not.equals(fragmentShaderTranspiled.trim());
+        }
+    });
```
0xFA11, line 67: we assume original built-in shaders are going to be in WebGL2 syntax (due to previous test above) and WebGL2 to WebGL1 transpiled output MUST be different after conversion.

src/shaders/shaders.ts
```diff
@@ -77,52 +76,51 @@ export type PreparedShader = {
...
+    projectionErrorMeasurement: prepare(projectionErrorMeasurementFrag, projectionErrorMeasurementVert),
+    atmosphere: prepare(atmosphereFrag, atmosphereVert),
+    sky: prepare(skyFrag, skyVert),
 };
 
-// Expand #pragmas to #ifdefs.
-
-function compile(fragmentSource: string, vertexSource: string): PreparedShader {
+/** Expand #pragmas to #ifdefs, extract attributes and uniforms */
+function prepare(fragmentSource: string, vertexSource: string): PreparedShader {
```
0xFA11, line 117: a small rename and proper jsdoc comment on the function.

src/shaders/shaders.test.ts
0xFA11, line 67: we assume original built-in shaders are going to be in WebGL2 syntax (due to previous test above) and WebGL2 to WebGL1 transpiled output MUST be different after conversion.

src/shaders/shaders.ts
0xFA11, line 117: a small rename and proper jsdoc comment on the function.

src/shaders/shaders.ts
```diff
@@ -131,7 +129,7 @@ function compile(fragmentSource: string, vertexSource: string): PreparedShader {
         if (operation === 'define') {
             return `
 #ifndef HAS_UNIFORM_u_${name}
-varying ${precision} ${type} ${name};
+in ${precision} ${type} ${name};
```
0xFA11, line 132: now that we assume original shaders to be in WebGL2, we should update our attribute and uniform extraction/scanning code.

src/shaders/shaders.ts
```diff
@@ -154,8 +152,8 @@ uniform ${precision} ${type} u_${name};
                 return `
 #ifndef HAS_UNIFORM_u_${name}
 uniform lowp float u_${name}_t;
-attribute ${precision} ${attrType} a_${name};
-varying ${precision} ${type} ${name};
+in ${precision} ${attrType} a_${name};
+out ${precision} ${type} ${name};
```
0xFA11, line 156: ^ same as above

src/shaders/shaders.ts
```diff
@@ -185,7 +183,7 @@ uniform ${precision} ${type} u_${name};
                 return `
 #ifndef HAS_UNIFORM_u_${name}
 uniform lowp float u_${name}_t;
-attribute ${precision} ${attrType} a_${name};
+in ${precision} ${attrType} a_${name};
```
0xFA11, line 186: ^ same as above

src/shaders/shaders.ts
```diff
@@ -213,5 +211,23 @@ uniform ${precision} ${type} u_${name};
...
+            .replace(/fragColor/g, 'gl_FragColor')
+            .replace(/texture\(/g, 'texture2D(');
+    }
+    if (type === 'vertex') {
+        return src
+            .replace(/\bin\s/g, 'attribute ')
+            .replace(/\bout\s/g, 'varying ')
+            .replace(/texture\(/g, 'texture2D(');
+    }
+    throw new Error(`Unexpected shader type: ${type}`);
```
0xFA11, line 232: this function is trying to convert (transpile, source-to-source conversion) WebGL2 shader source to WebGL1. originally moved from `generate-shaders.ts` but also had a minor refactor and second look to validate it is working as expected.

src/render/program.ts
```diff
@@ -72,6 +73,9 @@ export class Program<Us extends UniformBindings> {
         }
 
         const defines = configuration ? configuration.defines() : [];
+        if (isWebGL2(gl)) {
+            defines.unshift('#version 300 es');
+        }
```
0xFA11, line 78: this is a WebGL2 requirement in every shader source code, if we're on WebGL2.

src/render/program.ts
```diff
@@ -82,8 +86,13 @@ export class Program<Us extends UniformBindings> {
...
 
-        const fragmentSource = defines.concat(shaders.prelude.fragmentSource, projectionPrelude.fragmentSource, source.fragmentSource).join('\n');
-        const vertexSource = defines.concat(shaders.prelude.vertexSource, projectionPrelude.vertexSource, source.vertexSource).join('\n');
+        let fragmentSource = defines.concat(shaders.prelude.fragmentSource, projectionPrelude.fragmentSource, source.fragmentSource).join('\n');
+        let vertexSource = defines.concat(shaders.prelude.vertexSource, projectionPrelude.vertexSource, source.vertexSource).join('\n');
+
+        if (!isWebGL2(gl)) {
+            fragmentSource = transpileToWebGL1('fragment', fragmentSource);
+            vertexSource = transpileToWebGL1('vertex', vertexSource);
+        }
```
0xFA11, line 95: if we're not on WebGL2, we transpile shader sources from WebGL2 to WebGL1 at runtime on the fly.

src/shaders/shaders.test.ts
```diff
@@ -0,0 +1,68 @@
+import {describe, test, expect} from 'vitest';
+import {transpileToWebGL1} from './shaders';
+import {globSync} from 'glob';
+import fs from 'fs';
+
+describe('Shaders', () => {
+    test('`transpileToWebGL1()` should throw for unexpected type', () => {
+        //@ts-expect-error
+        expect(() => transpileToWebGL1('invalid', '')).toThrow();
+    });
```
0xFA11, line 10: we could rely on typescript enforcing types and not allowing anything other than 'vertex' | 'fragment' passed in as a parameter but since it's plain javascript at runtime, it is still possible to pass a string other than 'vertex' or 'fragment'. implementation throws an error if that's the case at runtime, this test simply tests that specific case.

src/shaders/shaders.ts
```diff
@@ -213,5 +211,23 @@ uniform ${precision} ${type} u_${name};
         }
     });
 
-    return {fragmentSource, vertexSource, staticAttributes, staticUniforms};
+    return {fragmentSource, vertexSource, staticAttributes: vertexAttributes, staticUniforms: shaderUniforms};
+}
+
+/** Transpile WebGL2 shader source to WebGL1 */
+export function transpileToWebGL1(type: 'fragment' | 'vertex', src: string): string {
```
HarelM, line 218: Why not split this method into two methods?
I don't see anything shared code between the two flows...

src/shaders/shaders.test.ts
HarelM: I think typescript is enough for this since this is an internal method.

src/shaders/shaders.test.ts
HarelM: Great! Thanks!
It might be better to move it to the integration test folder or build tests folder as this is reading from files.

src/shaders/shaders.test.ts
HarelM: I'm not sure this is a valuable enough test given the two tests above, I would consider removing it...

HarelM:
I've added a few minor comments, looks good otherwise, thanks!!

0xFA11:
> I've added a few minor comments, looks good otherwise, thanks!!

thank you for the code review @HarelM, I tried to address your feedback and I think it should be good to go now 🚀

HarelM (approved):

birkskyum:
> This dual-stack approach will hopefully allow more sophisticated and modern shaders written directly in maplibre stack.

@0xFA11 is it correctly understood that we still need, in addition to this PR, a proper abstraction of webgl1 and webgl2 as separate pipelines in order to better leverage webgl2? I mean, it must put certain limits on the set of webgl2 features that can be used, if there's a requirement to be convert it all back to webgl1 at runtime.

0xFA11:
> > This dual-stack approach will hopefully allow more sophisticated and modern shaders written directly in maplibre stack.
> 
> @0xFA11 is it correctly understood that we still need, in addition to this PR, a proper abstraction of webgl1 and webgl2 as separate pipelines in order to better leverage webgl2? I mean, it must put certain limits on the set of webgl2 features that can be used, if there's a requirement to be convert it all back to webgl1 at runtime.

So the basic idea behind this PR (and previous WebGL2 Shaders fix) was that, we should allow and choose WebGL2 shaders (with 300 es) over WebGL1 with a fallback if the canvas context was not initialized with WebGL2RenderingContext.

Currently, all built-in shaders in maplibre-gl-js are written with simple WebGL1 compatible features. We do not have any WebGL2 specific shaders or features in the codebase. For instance, we wanted to use pixel derivatives in order to draw outlines on buildings edges (custom fill-extrusion) and wanted to make them super sharp, we needed WebGL2 shaders (version 300 es) for that.

I would not suggest writing shaders with WebGL2 exclusive features for built-in shaders unless we completely drop WebGL1 — because it'd be a nightmare and arguably unreliable to port or emulate WebGL2 features on a WebGL1 stack.

This PR, at least, pushed WebGL1 fallback to happen at runtime only when necessary, which means that we could simply ignore and drop support for WebGL1 in our custom projects without asking maplibre-gl-js to also drop WebGL1 support. Previously, this was not trivial because maplibre-gl-js was downgrading (transpiling) shaders to WebGL1 anyway and we needed to modify maplibre-gl-js slightly to run WebGL2 with our WebGL2-only shaders.

_(FWIW, there are several WebGL2 features which were available on WebGL1 with extensions over time but they're now just available by default without extensions in WebGL2, that's great too)_

> a proper abstraction of webgl1 and webgl2 as separate pipelines

In general, we could implement 2 separate rendering pipelines for WebGL1 and WebGL2. It sounds attractive in theory but in practice, I see very little value in maintaining a full-fledged, separate pipeline just for WebGL1, especially since WebGL2 is mostly backwards compatible with WebGL1 anyway.

Some of this line of thought was discussed in #3947 and @HarelM didn't like the idea of dropping WebGL1 entirely in the end. For us and our products, we wanted to write our custom shaders in WebGL2 environment with WebGL2 shaders, without needing to bring in other additional renderers on top (like deck.gl etc.). A great next step evolution for us would be WebGPU (because it's now also enabled by default on iOS 18.2 onwards) but that needs an entire separate discussion anyway.

birkskyum:
@0xFA11 , thanks for clarifying, I definitely think this is a good step, and I really appreciate some much-needed progress in this area after it has been a bit stuck for two years.

I know I ask a lot of questions, but I think I'm finally understanding the key motivations here.

1. Was there really no way to skip the webgl2->webgl1 shader downgrade for _custom shaders_ until this point? For some reason I'd convinced myself all this time that we only affected the _built-in shaders_, and that there maybe was a way to injecting webgl2-type _custom shaders_ as long as we ran a webgl2 canvas, but now it's dawning on me that there's just _shaders_ - a single pipeline.

2. Is the main benefiter of this change the group of projects that choose to support only webgl2, because they're allowed to leverage all kinds of convenient webgl2-exclusive features now?

0xFA11:
That's correct @birkskyum and I can name a few: pixel derivatives, all vertex data in one buffer rather than chunks with 16-bit limit (e.g. for fill-extrusion vertices), MRT (Multiple Render Targets). There is simply no way to write WebGL1 compatible shaders for these things and now, we can write them side-by-side with built-in shaders and simply do not support WebGL1 at all.

I think I saw an earlier version of your comment, something like "couldn't you disable webgl2->webgl1 conversion and write webgl2 version 300 es shaders anyway?" — yes, we could do that for "custom style layer" shaders, however, that wouldn't be enough because we still needed "bucket", render passes and directly process data coming from the layer (let's say building from openmaptiles vector in mbtiles) to pass the entire vertex array in one go while still doing that work tile by tile in workers, similar to how maplibre-gl-js is doing.

For just a few extra visual enhancements, we could use custom style layer approach and implement the interface, write our WebGL2 version 300 es shaders there but for things applied to the fundamental rendering of tiles in the basemap, we needed this change.

We honestly think we don't actually need any additional renderers on top of maplibre-gl-js, like how deck.gl renders on top of maplibre-gl-js and treating it as a "basemap". We think we can just directly write fancy shaders and visualizations within maplibre-gl-js rendering pipeline.

We also think of contributing some of our work back if we can generalize them enough, hopefully we can. (smooth rounded edges on fill-extruded buildings, pre-computed -one at runtime- ambient occlusion on buildings, more accurate directional light for sun, pre-computed -one at runtime- cascaded shadows for buildings, loading OBJ and/or GLTF/GLB objects -without threejs or any other 3rd party lib-, to name a few). We can't promise if we can actually land them in a generalized way but we can promise that we will try do contribute them back if we can :)

birkskyum:
Awesome! I think I know a webgl2-only project that'll be stoked to learn of this!

PR #5205, opened 2024-12-14
Fix NavigationControl compass regression

raboczi:
## Launch Checklist

See https://github.com/maplibre/maplibre-gl-js/discussions/5195 for discussion.

I think the regression was caused by "TouchEvent" only being defined on Chromium-derived and/or touchscreen-equipped browsers, such that the drag handler would fail on Firefox and Safari on my Mac laptop.

This PR does two things:
* Adds a `typeof TouchEvent !== 'undefined'` check to prevent a runtime error when it's undefined.
* Consolidates the type checks into an `_onMouseEventOrTouchEvent` method to reduce duplication, possibly at the cost of clarity.

 - [X] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/ui/handler/drag_move_state_manager.ts
+++ b/src/ui/handler/drag_move_state_manager.ts
@@ -122,7 +122,7 @@ export class MouseOrTouchMoveStateManager implements DragMoveStateManager<MouseE
 
     _onMouseEventOrTouchEvent(e: MouseEvent | TouchEvent, onMouseEvent: (MouseEvent) => any, onTouchEvent: (TouchEvent) => any) {
         if (e instanceof MouseEvent) return onMouseEvent(e as MouseEvent);
-        if ('TouchEvent' in window && e instanceof TouchEvent) return onTouchEvent(e as TouchEvent);
+        if (window.TouchEvent && e instanceof TouchEvent) return onTouchEvent(e as TouchEvent);
     }
 
     startMove(e: MouseEvent | TouchEvent) {
--- a/src/ui/handler/drag_move_state_manager.ts
+++ b/src/ui/handler/drag_move_state_manager.ts
@@ -120,24 +120,38 @@ export class MouseOrTouchMoveStateManager implements DragMoveStateManager<MouseE
         private oneFingerTouchMoveStateManager = new OneFingerTouchMoveStateManager()
     ) {}
 
+    _onMouseEventOrTouchEvent(e: MouseEvent | TouchEvent, onMouseEvent: (MouseEvent) => any, onTouchEvent: (TouchEvent) => any) {
+        if (e instanceof MouseEvent) return onMouseEvent(e as MouseEvent);
+        if ("TouchEvent" in window && e instanceof TouchEvent) return onTouchEvent(e as TouchEvent);
+    }
+
     startMove(e: MouseEvent | TouchEvent) {
-        if (e instanceof MouseEvent) this.mouseMoveStateManager.startMove(e);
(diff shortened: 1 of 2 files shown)


raboczi:
The `'TouchEvent' in window && ...` fix I have here is based on the advice I found at https://stackoverflow.com/questions/3390396/how-can-i-check-for-undefined-in-javascript, but I tried and `window.TouchEvent && ...` as suggested in https://stackoverflow.com/questions/27313488/touchevent-not-working-in-firefox-and-other-website-browser also works in both Firefox and Safari.

HarelM:
Thanks!!

Other places in the code are using: typeof ...
Like here: 
https://github.com/maplibre/maplibre-gl-js/blob/f1a70c0257f7ad7fc70e2a19b83a09a169406f89/src/util/offscreen_canvas_supported.ts#L5

Let's keep it similar and avoid the need to write "window.".

raboczi:
> Let's keep it similar and avoid the need to write "window.".

Rewrote it with `typeof TouchEvent !== 'undefined'`.  Still seems to work.

src/ui/handler/drag_move_state_manager.ts
```diff
@@ -120,24 +120,38 @@ export class MouseOrTouchMoveStateManager implements DragMoveStateManager<MouseE
         private oneFingerTouchMoveStateManager = new OneFingerTouchMoveStateManager()
     ) {}
 
+    _onMouseEventOrTouchEvent(e: MouseEvent | TouchEvent, onMouseEvent: (MouseEvent) => any, onTouchEvent: (TouchEvent) => any) {
+        if (e instanceof MouseEvent) return onMouseEvent(e as MouseEvent);
```
HarelM, line 124: I belive there should be inference due to "instanceof" so I belive "as ..." is not needed...?

src/ui/handler/drag_move_state_manager.ts
```diff
@@ -120,24 +120,38 @@ export class MouseOrTouchMoveStateManager implements DragMoveStateManager<MouseE
         private oneFingerTouchMoveStateManager = new OneFingerTouchMoveStateManager()
     ) {}
 
+    _onMouseEventOrTouchEvent(e: MouseEvent | TouchEvent, onMouseEvent: (MouseEvent) => any, onTouchEvent: (TouchEvent) => any) {
```
HarelM, line 123: Can this method get a better name maybe? Something like executeRelevantHandler or executeBasedOnEventType
Something a bit more specific to explain why this method receives two other methods.

HarelM:
Overall this looks great, thanks! 
If there's a way to reproduce this issue in a simple unit test that can be great.

src/ui/handler/drag_move_state_manager.ts
raboczi: You are correct: the explicit cast is unneeded.

src/ui/handler/drag_move_state_manager.ts
raboczi: Honestly, I thought `_onMouseEventOrTouchEvent` was overly verbose but pretty explicit.  I've chosen the shorter of your two proposals.

HarelM:
Looks good! Can you add a changelog item?
I'm not sure how to properly test this, so if you find a way great, otherwise I'll merge as is.

raboczi:
> Looks good! Can you add a changelog item?

Done!

>I'm not sure how to properly test this, so if you find a way great, otherwise I'll merge as is.

I can't run the existing test suite locally, so I wouldn't dare to try adding my own new stuff to it.

HarelM (approved):

Issue #5194, opened 2024-12-12
 `tests\projection\globe\terrain\fill-planet-pole` test is not stable


HarelM:
I've seen in previous CI runs that fill-planet-pole test is flickering.
@kubapelc how valuable is this test?
From CI runs I see that there is a problem around the pole where the image doesn't look good in all kinds of scenarios, I've also seen this on my dev env using a Mac.

I think the easiest approach would be to remove it, but maybe a different test can provide a similar value for this case...
CC: @birkskyum @ibesora

HarelM:
Yes, sorry, forgot that there is more than one test with the same folder postfix.
The test is failing here:
https://github.com/maplibre/maplibre-gl-js/actions/runs/12293860857/job/34307425714

I've updated the issue's title.

HarelM:
It can exist in the git history, we can create an issue and add the style.json file there, or we can keep this issue open and remove the test for now until we decide to solve this issue.
But I have removed ignored tests in the past as it is a bad practice, so I would like to avoid adding skipped tests...

birkskyum:
Okay, we can do that. I'm not trying to be argumentative at all, or be the judge what's bad practice. I genuinely find it difficult to figure what's good practice for the main branch in this CI/CD pipeline era. The formal training I've had in the area of code testing was dominated by recommendations of following a TDD doctrine, and in that mindset it seems completely opposite to delete a failing test solely for the reason that correctly flagging a bug. With TDD I'd always have this list of failing tests ahead of me until the project was done. Now we have a situation where that might still be possible to some extend on a feature branch, but the main branch has to be kept CI-green at all times, otherwise it'll block all incoming PRs, pre-releases etc. 

If it was possible, I'd probably prefer to have the CI on main branch run all the test (so none are skipped/ignored), but allow a certain tests that hit bugs or lack implementation to fail silently.

HarelM:
TDD means you are writing tests first and then making them pass.
Keeping a failing tests only gets you to ignore the CI results as you are saying that "this is probably just another flaky test".
You are right that the best practice is to fix the test, but I'm not sure we can do that ATM and it's best not to keep flaky tests.
We can open a branch, add that test there and return to that branch when we have the time to fix that, but I'm not sure this would work either.
So I prefer not flaky tests over completeness of coverage I guess is what I'm saying.

PR #5232, opened 2024-12-17
Fix globe terrain holes at poles

### LINKED ISSUE #4984: Mesh is broken near the pole on globe with terrain
There's a hole down though planet if Terrain3D is active on the Globe, which is making some glitchy behavior.

This can be seen in the `heatmap-terrain.html` example.

This is not fixed in:
- #4977 

Arguably easiest to start from this branch
- #4977 

Screenshot 2024-11-04 at 12 29 29

### LINKED ISSUE #4990: Globe with/without terrain can zoom too far out
A globe with terrain will by default allow zooming to negative zoom levels, resulting in some undefined behavior. 

The globe will show artifacts, and then jump back to the minZoom.
There should be a hard limit on how far a user is allowed to zoom out.

https://github.com/user-attachments/assets/7f757c47-5620-4b5a-885b-f2c70d577b6f

Screenshot 2024-11-04 at 22 26 46

Repro:
- Open hybrid-terrain example
- Zoom out

kubapelc:
- Fixes #4984
- Fixes #4990

Also fixes an artifact when globe terrain was zoomed out too much. Brings back the globe/terrain/fill-planet-pole test (see #5194), it should no longer be unstable.


src/render/terrain.ts
```diff
@@ -108,7 +109,7 @@ export class Terrain {
      * GL Objects for the terrain-mesh
      * The mesh is a regular mesh, which has the advantage that it can be reused for all tiles.
      */
-    _mesh: Mesh;
+    _meshDict: { [key: string]: Mesh } = {};
```
HarelM, line 112: `meshCache`?

HarelM:
Does it fix the following issue as well? 
- #4990

kubapelc:
> Does it fix the following issue as well?

It fixes the geometry sticking out of the planet, but not the zoom controls/behaviour.

HarelM:
I'll mark it as fixed as I think this is the main concern in this issue, for the problem of min zoom another issue can be opened if someone cares about it enough...

HarelM (approved):
Thanks!!

kubapelc:
I've just noticed that the pole mesh is also used on mercator and it doesn't look too good, I'll try to fix that.

PR #5235, opened 2024-12-18
Atmosphere blending fix

### LINKED ISSUE #5230: Globe sky color
Hello,

I know the Globe is not stable yet, but I would like to ask if there is a plan for fix or implementation of [sky](https://maplibre.org/maplibre-style-spec/sky/) for the globe projection.

I was playing with `5.0.0-pre.9` and looks like the sky specification is working only for mercator projection. There is some sky default for globe projection, see pictures:

overview:
Screenshot 2024-12-16 at 19 22 58

detail (crop):
Screenshot 2024-12-16 at 19 22 58 copy

_Originally posted by @slaviczavik in https://github.com/maplibre/maplibre-gl-js/discussions/5223_

kubapelc:
Fixes the black banding seen around atmopshere. Fixes #5230

Before:

After:


src/shaders/atmosphere.fragment.glsl
```diff
@@ -131,7 +133,7 @@ vec4 atmosphere(vec3 r, vec3 r0, vec3 pSun, float iSun, float rPlanet, float rAt
 
     // Calculate opacity
     //float opacity = exp(-(length(kRlh) * iOdRlh + kMie * iOdMie));
-    float opacity = min(0.75, exp(-(length(kRlh) * length(totalRlh) + kMie * length(totalMie))));
+    float opacity = exp(-(length(kRlh) * length(totalRlh) + kMie * length(totalMie)));
```
HarelM, line 136: Is the comment above this line still relevant? Can it be removed?

HarelM:
Should there be a test similar to the screen shot you have provided in the initial comment of this thread?

src/shaders/atmosphere.fragment.glsl
kubapelc: I'm not sure, it's not my code, but I guess it can be removed. The used line and commented line both approximate opacity in a slightly different ways.

kubapelc:
> Should there be a test similar to the screen shot you have provided in the initial comment of this thread?

Existing tests cover this case well - the black bands were visible in the original expected images.

kubapelc:
@HarelM also note that by removing the `min(..., 0.75)` thing in the shader the atmosphere no longer (arbitrarily) darkens the map - see the test old/new test image comparison, it is pretty visible there. But the darkening did make the non-illuminated side of the globe darker, making it look kinda like lighting. People might request that at some point.

HarelM (approved):
Thanks!

HarelM:
Yes, I saw the diff in the images, the 0.75 might have been a bit too dark, I think that in the new images it's clear that there is a shadow on one hand but you can see the underlaying details on the other hand.

HarelM:
Ah. forgot to ask about a changelog entry.
Can you please add that?

kubapelc:
Added!

PR #5239, opened 2024-12-18
Expose `coveringTiles()` in transform objects

jonathanlurie:
## Launch Checklist

## Description
Before the work on globe, the function `.coveringTiles()` used to be a method of the `Transform` object, it was very convenient to access the list of tiles (with z, x and y indices) that were currently displayed. This PR aims at bringing this function back to the `map.transform` object.

## Motivations
We (at MapTiler but we are not the only ones) are creating tiled layers to display raster data. These tiles are WebGL polygons with the same footprints as regular raster tiles, just with custom shaders. When only using Mercator it was fairly easy to figure out the math to know what tile should be displayed, but with the globe view, it is much more difficult. Since the list of tiles is computed internally in Maplibre, it would be great to expose it so that third-party plugins can use it.

Contrary to a previous PR I submitted (https://github.com/maplibre/maplibre-gl-js/pull/5158), this one does not add any function or object to the official public API.

## Usage
```ts
const tileList = map.transform.coveringTiles({tileSize: 512});
```

The `tileSize` attribute is the only one mandatory (as in the internal `coveringTiles()` function). Since it's the default size is always `512`, we could make it the default value if not provided.


HarelM:
This is a change in the opposite direction of what we are looking to do.
Please also read the following PR comments and problems:
- #5158

I would advise to think of a better way to expose this.
The transform class is not an external API nor we are planning for it to be so.

HarelM:
Closing this for now as there was no response, if this is still relevant please let me know and I'll reopen.

PR #5253, opened 2024-12-20
Add support for 204 in render tests

HarelM:
## Launch Checklist

In continue to the last messages in 
- #2497 

I have added support for returning 204 in render tests.
I haven't found an easy way to do that using the `st` package.

### DIFF
--- a/test/integration/render/run_render_tests.ts
+++ b/test/integration/render/run_render_tests.ts
@@ -301,7 +301,7 @@ async function getImageFromStyle(styleForTest: StyleWithTestData, page: Page): P
                 }`;
 
                 const fragmentSource = `#version 300 es
-                
+
                 out highp vec4 fragColor;
                 void main() {
                     fragColor = vec4(1.0, 0.0, 0.0, 1.0);
@@ -452,7 +452,7 @@ async function getImageFromStyle(styleForTest: StyleWithTestData, page: Page): P
                 // Inject MapLibre projection code
                 ${shaderDescription.vertexShaderPrelude}
                 ${shaderDescription.define}
-                
+
                 in vec3 a_pos;
 
                 void main() {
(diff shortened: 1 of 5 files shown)


sbachinin (approved):

test/integration/render/run_render_tests.ts
```diff
@@ -982,15 +982,56 @@ async function executeRenderTests() {
...
 
-    const browser = await puppeteer.launch({headless: !options.openBrowser, args: ['--enable-webgl', '--no-sandbox',
-        '--disable-web-security']});
+    const browser = await puppeteer.launch({
+        headless: !options.openBrowser, 
+        args: [
+            '--enable-webgl', 
+            '--no-sandbox',
+            '--disable-web-security'
+        ]});
```
ibesora, line 991: Can we move these definitions in a function somewhere? It seems error prone to have different ways of configuring puppeteer across the code base

ibesora (approved):
Added a small nit but feel free to ignore it

test/integration/render/run_render_tests.ts
HarelM: We can probably create something in test utils class, but it's just a bit different from test suite to test suite, so I don't know...

HarelM: Also this was moved to a different PR as main is currently broken, so I needed to merge this fast.
Feel free to open a PR to improve on this.

HarelM:
I'm still waiting to see what the author of `st` will write before I merge this...

xabbu42:
Looking at this code and the answer from the author of st I may be able to make a shorter fix. I'll try to have pull request in the next days. I'm sorry for causing you extra work @HarelM but I would not have had the right direction without it.

HarelM:
Don't worry about the extra work, you made a good observation I missed when reviewing the tests change. 
If you can configure st to return 204 then we are even 🤪

HarelM:
Solved differently by #5269.

PR #5266, opened 2024-12-24
Better error reporting

### LINKED ISSUE #4532: v4.0.0 silently swallows errors in load event
**maplibre-gl-js version**: 4.5.1

**browser**: safari

### Steps to Trigger Behavior

 1. throw any error or reference an uninitialized variable in the `on("load",...)` event.
 2. No error is printed to the console.

### Link to Demonstration

Paste this in on JSBin: https://jsbin.com/

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <title>Display a map</title>
    <meta charset='utf-8'>
    <link rel='stylesheet' href='https://unpkg.com/maplibre-gl@4.5.1/dist/maplibre-gl.css' />
    <script src='https://unpkg.com/maplibre-gl@4.5.1/dist/maplibre-gl.js'></script>
    <style>
    </style>
</head>
<body>
<div id="map"></div>
<script>
        zoom: 1, // starting zoom
        maplibreLogo: true
    });
  
    map.on("load", () => {
      console.log(variable_does_not_exist);
    })
</script>
</body>
</html>
```

### Expected Behavior

On the last version of 3.x, `console.error` is called with `ReferenceError: Can't find variable: variable_does_not_exist`

```html
<!DOCTYPE html>
<html lang="en">
<head>
    <title>Display a map</title>
    <meta charset='utf-8'>
    <link rel='stylesheet' href='https://unpkg.com/maplibre-gl@3.6.2/dist/maplibre-gl.css' />
    <script src='https://unpkg.com/maplibre-gl@3.6.2/dist/maplibre-gl.js'></script>
    <style>
    </style>
</head>
<body>
<div id="map"></div>
<script>
        zoom: 1, // starting zoom
        maplibreLogo: true
    });
  
    map.on("load", () => {
      console.log(variable_does_not_exist);
    })
</script>
</body>
</html>
```

### Actual Behavior

Nothing is printed to the console.

### LINKED ISSUE #4613: After version 4.5.0 requests with 404 are not emitted in map.on('error')
maplibre-gl-js version: 4.6.0
browser: Chrome

### Steps to Trigger Behavior

1. Create a MapLibre map with vector tiles that have data only up to a certain zoom level (e.g., zoom level 16).
2. Attach an event listener to capture errors:
```
map.on('error', (e) => {
    console.log('ERROR', e);
});

```
3. Zoom in past the available tile data (e.g., zoom beyond level 16 where 404 responses are expected for missing tiles).

### Expected Behavior

- When a tile request results in a 404 response due to missing data at higher zoom levels, the map.on('error') event should emit the error as it did in versions prior to 4.5.0.
- This allowed overzooming tiles and keeping the previous zoom level’s data visible.

### Actual Behavior

- In version 4.6.0, map.on('error') no longer emits any error when tile requests result in 404 responses.
- This causes tiles with no data at certain zoom levels to disappear.

### Reproduction
1. Surf to https://jsbin.com/puwuwob/2/edit?html,output
2. Zoom to level 19
3. Change the version to 4.4 to see how it worked in the past and how it doesn't work in recent versions (tested up until 5.2)

ibesora:
This PR avoids suppressing any error that's not an abort error or a "Framebuffer not complete" error.
While doing that I found multiple instances of errors that were being thrown and swallowed across the codebase and had to fix some tests.

- Fixes #4532 
- Fixes #4613

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/render/painter.ts
+++ b/src/render/painter.ts
@@ -492,7 +492,7 @@ export class Painter {
         const coordsAscending: {[_: string]: Array<OverscaledTileID>} = {};
         const coordsDescending: {[_: string]: Array<OverscaledTileID>} = {};
         const coordsDescendingSymbol: {[_: string]: Array<OverscaledTileID>} = {};
-        const renderOptions: RenderOptions = {isRenderingToTexture: false, isRenderingGlobe: style.projection.transitionState > 0};
+        const renderOptions: RenderOptions = {isRenderingToTexture: false, isRenderingGlobe: style.projection?.transitionState > 0};
 
         for (const id in sourceCaches) {
             const sourceCache = sourceCaches[id];
@@ -539,7 +539,7 @@ export class Painter {
         }
 
         // Execute offscreen GPU tasks of the projection manager
-        this.style.projection.updateGPUdependent({
+        this.style.projection?.updateGPUdependent({
             context: this.context,
             useProgram: (name: string) => this.useProgram(name)
         });
--- a/src/style/style.ts
+++ b/src/style/style.ts
@@ -203,7 +203,7 @@ export class Style extends Evented {
     glyphManager: GlyphManager;
     lineAtlas: LineAtlas;
     light: Light;
-    projection: Projection;
+    projection: Projection | undefined;
     sky: Sky;
 
     _frameRequest: AbortController;
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -3222,14 +3222,14 @@ export class Map extends Camera {
             this.style.update(parameters);
         }
 
-        const globeRenderingChaged = this.style.projection?.transitionState > 0 !== isGlobeRendering;
+        const globeRenderingChanged = this.style.projection?.transitionState > 0 !== isGlobeRendering;
         this.style.projection?.setErrorQueryLatitudeDegrees(this.transform.center.lat);
         this.transform.setTransitionState(this.style.projection?.transitionState, this.style.projection?.latitudeErrorCorrectionRadians);
 
         // If we are in _render for any reason other than an in-progress paint
         // transition, update source caches to check for and load any tiles we
         // need for the current transform
-        if (this.style && (this._sourcesDirty || globeRenderingChaged)) {
+        if (this.style && (this._sourcesDirty || globeRenderingChanged)) {
             this._sourcesDirty = false;
             this.style._updateSources(this.transform);
         }
@@ -3248,7 +3248,7 @@ export class Map extends Camera {
             }
         }
 
-        this._placementDirty = this.style && this.style._updatePlacement(this.transform, this.showCollisionBoxes, fadeDuration, this._crossSourceCollisions, globeRenderingChaged);
+        this._placementDirty = this.style && this.style._updatePlacement(this.transform, this.showCollisionBoxes, fadeDuration, this._crossSourceCollisions, globeRenderingChanged);
 
         // Actually draw
         this.painter.render(this.style, {
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -3381,7 +3381,11 @@ export class Map extends Camera {
                 PerformanceUtils.frame(paintStartTimeStamp);
                 this._frameRequest = null;
                 this._render(paintStartTimeStamp);
-            }).catch(() => {}); // ignore abort error
+            }).catch((error: Error) => {
+                if (error.message.startsWith('WebGL:')) {
+                    throw error;
+                }
+            }); // ignore abort error
         }
     }
 
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -3387,7 +3387,7 @@ export class Map extends Camera {
                 if (!isAbortError(error) && !isFramebufferNotCompleteError(error)) {
                     throw error;
                 }
-            }); // ignore abort error
+            });
         }
     }
 
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -64,6 +64,7 @@ import {MercatorTransform} from '../geo/projection/mercator_transform';
 import {type ITransform} from '../geo/transform_interface';
 import {type ICameraHelper} from '../geo/projection/camera_helper';
 import {MercatorCameraHelper} from '../geo/projection/mercator_camera_helper';
+import {isAbortError} from '../util/abort_error';
 
 const version = packageJSON.version;
 
@@ -3382,7 +3383,7 @@ export class Map extends Camera {
                 this._frameRequest = null;
                 this._render(paintStartTimeStamp);
             }).catch((error: Error) => {
-                if (error.message.startsWith('WebGL:')) {
+                if (!isAbortError(error)) {
                     throw error;
                 }
             }); // ignore abort error
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -65,7 +65,7 @@ import {type ITransform} from '../geo/transform_interface';
 import {type ICameraHelper} from '../geo/projection/camera_helper';
 import {MercatorCameraHelper} from '../geo/projection/mercator_camera_helper';
 import {isAbortError} from '../util/abort_error';
-import { isFramebufferNotCompleteError } from '../util/framebuffer_error';
+import {isFramebufferNotCompleteError} from '../util/framebuffer_error';
 
 const version = packageJSON.version;
 
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -65,6 +65,7 @@ import {type ITransform} from '../geo/transform_interface';
 import {type ICameraHelper} from '../geo/projection/camera_helper';
 import {MercatorCameraHelper} from '../geo/projection/mercator_camera_helper';
 import {isAbortError} from '../util/abort_error';
+import { isFramebufferNotCompleteError } from '../util/framebuffer_error';
 
 const version = packageJSON.version;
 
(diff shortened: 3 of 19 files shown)


src/render/painter.ts
```diff
@@ -539,7 +539,7 @@ export class Painter {
         }
 
         // Execute offscreen GPU tasks of the projection manager
-        this.style.projection.updateGPUdependent({
+        this.style.projection?.updateGPUdependent({
```
ibesora, line 542: `this.style.projection` is only defined after the style is loaded. This was throwing an error in multiple places due to the code accessing an `undefined` property

src/ui/map.ts
```diff
@@ -3381,7 +3383,11 @@ export class Map extends Camera {
                 PerformanceUtils.frame(paintStartTimeStamp);
                 this._frameRequest = null;
                 this._render(paintStartTimeStamp);
-            }).catch(() => {}); // ignore abort error
+            }).catch((error: Error) => {
+                if (!isAbortError(error) && !isFramebufferNotCompleteError(error)) {
```
ibesora, line 3387: Don't like swallowing "framebuffer is not complete" errors but I haven't been able to make some tests work doing it any other way. Mocking `WebGLContext`, which would be the best solution, doesn't seem to work on some instances.
Although we swallow that error, I think this is an improvement over not reporting any of them

src/ui/map_tests/map_events.test.ts
```diff
@@ -959,6 +960,19 @@ describe('map events', () => {
...
+
+        const loadHandler = vi.fn(() => {
+            throw new Error('Error in load handler');
+        });
+
+        map.on('load', loadHandler);
+        await sleep(1);
+
+        expect(loadHandler).toThrowError();
+    });
```
ibesora, line 974: This test checks for the error reported in https://github.com/maplibre/maplibre-gl-js/issues/4532

src/ui/map_tests/map_events.test.ts
```diff
@@ -1049,6 +1060,37 @@ describe('map events', () => {
...
+
+            const errorHandler = vi.fn();
+            map.on('error', errorHandler);
+
+            map.triggerRepaint();
+            await sleep(100);
+
+            expect(errorHandler).toHaveBeenCalledTimes(1);
+          
+        });
```
ibesora, line 1093: This one checks https://github.com/maplibre/maplibre-gl-js/issues/4613

src/util/evented.ts
```diff
@@ -124,7 +124,11 @@ export class Evented {
         const type = event.type;
 
         if (this.listens(type)) {
-            (event as any).target = this;
+            try {
+                (event as any).target = this;
+            } catch (_) { // eslint-disable-line @typescript-eslint/no-unused-vars
+                // On some type of events, setting target raises a TypeError
+            }
```
ibesora, line 131: I also don't like this but I haven't find any other way to fix the error.
Javascript was complaining with `TypeError: Cannot set property target of #<Event> which has only a getter` while running some tests

src/util/test/util.ts
```diff
@@ -223,3 +223,33 @@ export function expectToBeCloseToArray(actual: Array<number>, expected: Array<nu
         expect(actual[i]).toBeCloseTo(expected[i], precision);
     }
 }
+
+export function createTerrain() {
```
HarelM, line 227: Return type should be `Terrain` to avoid any need to cast the response to `any`.

src/util/evented.ts
```diff
@@ -124,7 +124,11 @@ export class Evented {
         const type = event.type;
 
         if (this.listens(type)) {
-            (event as any).target = this;
+            try {
+                (event as any).target = this;
+            } catch (_) { // eslint-disable-line @typescript-eslint/no-unused-vars
+                // On some type of events, setting target raises a TypeError
```
HarelM, line 130: Not sure I fully understand this, can you please elaborate?

src/ui/marker.test.ts
```diff
@@ -1003,10 +994,7 @@ describe('marker', () => {
             .setLngLat([0, 0])
             .addTo(map);
 
-        map.terrain = {
-            getElevationForLngLatZoom: () => 0,
-            depthAtPoint: (p) => p.y === 256 ? .95 : .92 // return "far" given the marker's center coord; return "near" otherwise
-        } as any as Terrain;
+        map.terrain = {...createTerrain(), depthAtPoint: (p) => p.y === 256 ? .95 : .92} as any as Terrain;
```
HarelM, line 997: Instead of using `{... }` use `map.terrain.depthAtPoint = ...` so that it's clear what you are overriding here.
You can also allow passing it to the `createTerrain` method, but I think it will only complicate it too much.

src/ui/map.ts
```diff
@@ -3381,7 +3383,11 @@ export class Map extends Camera {
...
+                    throw error;
+                }
+            }); // ignore abort error
```
HarelM, line 3390: I think the comment is no longer needed since the code is now self explenatory.

src/style/style.ts
```diff
@@ -203,7 +203,7 @@ export class Style extends Evented {
     glyphManager: GlyphManager;
     lineAtlas: LineAtlas;
     light: Light;
-    projection: Projection;
+    projection: Projection | undefined;
```
HarelM, line 206: Is it possible to avoid all the `?` checks somehow?

test/build/sourcemaps.test.ts
```diff
@@ -84,6 +84,6 @@ describe('main sourcemap', () => {
         const s1 = setMinus(actualEntriesInSourcemapJSON, expectedEntriesInSourcemapJSON);
         expect(s1.length).toBeLessThan(5);
         const s2 = setMinus(expectedEntriesInSourcemapJSON, actualEntriesInSourcemapJSON);
-        expect(s2.length).toBeLessThan(16);
+        expect(s2.length).toBeLessThan(17);
```
HarelM, line 87: Why was this changed?

src/render/painter.ts
HarelM: How is this different from `light` property?

src/util/evented.ts
HarelM: I would look into fixing those tests instead of the code, although I haven't seen these console errors when I ran any of the tests locally.
Where did you see this error?

ibesora: Those errors were getting swallowed by the `catch` in map.ts. Just catching `Abort` errors is what made them show, although they were already present.
I've pushed a fix that removes this try/catch. We were using browser `Event` instead of `Maplibre `Event` in some tests and that was triggering the error

HarelM: Ah, makes sense, thanks!

src/render/painter.ts
ibesora: It's the same thing really but since `light` is only used when rendering globe and that's only set after the style is loaded, there's no error raised. 
The problem with `this.style.projection` is that it's used in `render` which can happen before the style has been loaded

test/build/sourcemaps.test.ts
ibesora: I was compiling with `npm run start` and this one was failing. Compiling with `npm run build-dist` regenerated the sourcemap and made it go away

src/style/style.ts
ibesora: I tried initializing the style projection with `mercator` and let it change to `vertical-perspective` if needed when the style is finally loaded (similar to what we do with `transform`) but that broke in lots of other different ways. I can take a deeper look at that if you prefer it that way though

HarelM: Are all the places where you added "?" mandatory?

ibesora: Yup, I basically kept adding them as I ran the tests

HarelM: But doesn't this mean the tests need to be updated to overcome this issue instead of adding it in the code, where this might not be a valid code path?

HarelM:
I've tried to test this code with:
`'circle-translate': [100, 100]` and it doesn't look like this fully solve this.
I'm guessing there's still a need to fine tune the relevant matrices etc, which is very confusing to me to be honest...

src/style/style.ts
ibesora: From a typing point of view I'd argue that not having `undefined` in there is wrong, since the `projection` object is not being created on the `style` constructor which means that we might be adding code that relies on that without TS complaining. Leaving that aside, I can't see how this might not be a valid code path if it's being triggered in some of our existing tests that only use the public APIs.

The following snippet triggers that error:
```js
const style = {
        'version': 8,
        'sources': {},
        'layers': []
    }
    container: 'map',
    style
});

map.setStyle(style, {
    transformStyle: (prevStyle, nextStyle) => ({
        ...nextStyle,
    })
});
```

HarelM: Fair enough.

src/ui/map.ts
```diff
@@ -3220,9 +3222,9 @@ export class Map extends Camera {
             this.style.update(parameters);
         }
 
-        const globeRenderingChaged = this.style.projection.transitionState > 0 !== isGlobeRendering;
-        this.style.projection.setErrorQueryLatitudeDegrees(this.transform.center.lat);
-        this.transform.setTransitionState(this.style.projection.transitionState, this.style.projection.latitudeErrorCorrectionRadians);
+        const globeRenderingChaged = this.style.projection?.transitionState > 0 !== isGlobeRendering;
```
HarelM, line 3225: Chaged should be changed, not fully related to this change obviously...

HarelM (approved):

Issue #5255, opened 2024-12-21
Globe Projection Bug: Circle Layers have smaller hover/click area than visually represented


ingalls:
**maplibre-gl-js version**: 5.0.0-pre10

**browser**: Reproduced in Chrome and Firefox on Ubuntu 24.10

### Steps to Trigger Behavior

 1. Create a Circle Layer with Icons OR actual circles
 2. When using in mercator mode the hover/click events match the rendered symbol's bounds
 3. Turning on globe projects seems to only allow a very small point within the rendered circle or icon that hovers or clicks

### Link to Demonstration

https://codepen.io/ingalls-the-bold/pen/jENmLGe

### Expected Behavior

MouseEnter/Leave & Click covers entire rendered feature

### Actual Behavior

MouseEnter/Lease & Click is much smaller than actual rendered feature in the new Globe projection

HarelM:
Thanks for taking the time to report this issue!
I think this is related to missing implementation in queryRenderFeatures for globe.
https://github.com/maplibre/maplibre-gl-js/blob/cf8e2ecd8b3b46588220733eca1364199c6cef77/src/source/query_features.ts#L61-L69
CC: @kubapelc

HarelM:
This seems to apply only for circles.
While this bug needs a solution, since it does not apply to every type of layer, I'm reducing its severity.

bp2030:
Clusters with globe projection gives a very small clickable area as well. I'm guessing this is linked to this bug.

evoactivity:
If you zoom out and zoom back in the points become pretty much impossible to click on.

ingalls:
I'll reward a $500 bounty via Github Sponsors for anyone that cuts a PR that fixes this issue and is merged/released by the maintainers of the Maplibre GL JS project by the end of March, 2025.

PR #5267, opened 2024-12-24
Transform specific `projectPoint`

ibesora:
This PR is my first take on fixing https://github.com/maplibre/maplibre-gl-js/issues/5255
It moves circle-layer specific projection code to each transform to make it transform-aware. 
Globe projection is implemented using the view projection matrix, which works as a simple approximation 

Before (Notice how the cursor changes to a hand when the underlying feature is detected)

https://github.com/user-attachments/assets/3b93b415-ab2a-4f6a-9361-69142c1c3369

After

https://github.com/user-attachments/assets/21fb77af-b291-4c05-bbc5-76e48ce8500f

@kubapelc I'd love if you could take this PR and follow along with a more exact implementation if you have time

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/geo/projection/globe_transform.ts
+++ b/src/geo/projection/globe_transform.ts
@@ -317,6 +317,10 @@ export class GlobeTransform implements ITransform {
         return this.currentTransform.projectTileCoordinates(x, y, unwrappedTileID, getElevation);
     }
 
+    public projectPoint(p: Point, pixelPosMatrix: mat4): Point {
+        return this.currentTransform.projectPoint(p, pixelPosMatrix);
+    }
+
     private _calcMatrices(): void {
         if (!this._helper._width || !this._helper._height) {
             return;
--- a/src/geo/projection/mercator_transform.ts
+++ b/src/geo/projection/mercator_transform.ts
@@ -779,6 +779,11 @@ export class MercatorTransform implements ITransform {
         };
     }
 
+    projectPoint(p: Point, pixelPosMatrix: mat4) {
+        const point = vec4.transformMat4([] as any, [p.x, p.y, 0, 1], pixelPosMatrix);
+        return new Point(point[0] / point[3], point[1] / point[3]);
+    }
+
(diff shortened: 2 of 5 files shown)


HarelM:
While this may solve this case, I'm not sure it will solve the cases for other layers.
When I looked into the code, what I found problematic was the following method:
https://github.com/maplibre/maplibre-gl-js/blob/9fc7e7311b327e5ee89e4c320e943e5330cc6d71/src/source/query_features.ts#L61-L70

It is part of the code path that causes this bug as far as I understand.
The method is not typed so it passes well for transform that has the `calculatePosMatrix` method, which is basically the mercator transform.
I was hoping that solving this method and adding the relevant posMatrix would solve this.
I'll see if I can create a failing test never the less as it will be helpful regardless of the solution.

HarelM:
I've tried to test this code with:
'circle-translate': [100, 100] and it doesn't look like this fully solve this.
I'm guessing there's still a need to fine tune the relevant matrices etc, which is very confusing to me to be honest...

kubapelc:
I think that instead of correctly projecting all features to screen space for globe, a better appoach would be to project the query geometry to mercator coordinates or tile space and approximate it with some polygon (since it will get curved by the projection), and then query features against that. The reason is that projecting from tile space to globe is slower than projecting from tile space to mercator coordinates, so we want to prefer the other projection where possible, and bulk of the projecting will be done for the features.

HarelM:
The question is about all kind of transformation related to screen space like circle radius, circle offset, the padding that is used in pixels, all of those are in screen space so there is a conversation back and forth. 
These are defined in the style.json and most of them are in pixels. 

I added some comments recently to help understand in which domain the coordinates are, but it's confusing. 
Can we use a method instead of a matrix to do the transformations or will this be too slow?

HarelM:
This is now being addressed by another PR.
Closing.

PR #5269, opened 2024-12-26
Return 204 for some missing files in tiles/sparse204.

xabbu42:
This way we can use st passthrough to generate the 204 results necessary for testing. Also allows to just symlink sparse204 to sparse as we no longer need any empty files.

This is an alternative to https://github.com/maplibre/maplibre-gl-js/pull/5253.

## Launch Checklist

 - [x] Link to related issues.


test/integration/render/run_render_tests.ts
```diff
@@ -990,12 +990,22 @@ async function executeRenderTests() {
...
-        })
-    );
+    const mount = st({
+        path: 'test/integration/assets',
+        cors: true,
+        passthrough: true,
+    });
+    const server = http.createServer((req, res) => {
+        mount(req, res, () => {
+            if (req.url.includes('/sparse204/1-')) {
```
HarelM, line 1000: Can we rename it to something a bit more generic? Maybe like "empty" or "empty-file"?

xabbu42: well not all files in sparse204 are empty. the idea is that sparse contains incomplete tiles and sparse204 incomplete tiles returning 204 for the missing tiles.

Issue #5248, opened 2024-12-19
Vector tiles disappear when more detailed tile is being fetched


kamil-sienkiewicz-asi:
**maplibre-gl-js version**: 5.0.0-pre.10

**browser**: chrome

### Steps to Trigger Behavior

 1. Enter webpage with maplibre
 2. Set throttling to networking in chrome dev tools / have slow internet connection
 3. Sometimes tiles are completely disappearing and only background is rendered

### Link to Demonstration

https://codepen.io/kamil-sienkiewicz-asi/pen/zxOZeaP

### Expected Behavior

Less detailed vector tile is rendered until more detailed tile is completely fetched

### Actual Behavior

Less detailed vector tiles is removed at the start of fetch of new tile - which results in some period of time without any tile at all.

https://github.com/user-attachments/assets/796e46ea-beb1-45b7-94c9-4d2fab8a009c

kamil-sienkiewicz-asi:
@HarelM I mean, sort of, I think it's somewhere in-between bug and a feature request, I think that on mercator projection tiles are behaving in more "natural" way, where on globe they can disappear without clear reason, especially when camera is facing poles.

On my video I have no clue why tiles disappeared when I zoomed in, on mercator they would just wait for higher resolution to be fetched, where on globe I got empty sphere

HarelM:
Ahh, so this is a globe only related problem, interesting...

PR #5271, opened 2024-12-26
Fix children not retained when wrapped

ibesora:
This PR fixes https://github.com/maplibre/maplibre-gl-js/issues/5248

Looking at it, when globe is being used, some tileIDs coordinates are wrapped and we were only using unwrapped tile ids to check if a parent tile had loaded children. The fix here is just to add a check for the canonical key, in addition to the existing one.

Before

https://github.com/user-attachments/assets/fabfc02b-ae6f-410e-840c-f5f052efa242

After

https://github.com/user-attachments/assets/9d2769c3-eb47-4b57-9cbf-fb33df747cb7

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/source/source_cache.test.ts
+++ b/src/source/source_cache.test.ts
@@ -14,6 +14,7 @@ import {TileBounds} from './tile_bounds';
 import {sleep} from '../util/test/util';
 import {type TileCache} from './tile_cache';
 import {MercatorTransform} from '../geo/projection/mercator_transform';
+import {GlobeTransform} from '../geo/projection/globe_transform';
 
 class SourceMock extends Evented implements Source {
     id: string;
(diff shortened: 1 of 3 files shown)


src/source/source_cache.test.ts
```diff
@@ -851,6 +852,43 @@ describe('SourceCache#update', () => {
         sourceCache.onAdd(undefined);
     }));
 
+    test('retains children tiles for pending parents', () => new Promise<void>(done => {
```
HarelM, line 855: Is there a way to avoid "done" in tests? I'm trying to remove usage of those as best as I can, but it's progressing slow, so I want to at least avoid adding more...

ibesora: Removed! I just copied what other tests were doing but it seems it's not needed. I've checked that the test passes and fails correctly without it

HarelM (approved):

kamil-sienkiewicz-asi:
Hey @ibesora thanks for the fix, however it seems like similar issue is still reproducible

https://codepen.io/kamil-sienkiewicz-asi/pen/zxOZeaP

[Kapture 2025-01-07 at 11.39.01.webm](https://github.com/user-attachments/assets/f0f7ccbb-2dec-47a1-9629-a56539be9602)

Take a look at top right part of the visible globe, when im moving camera closer and closer to the north pole, one of the tiles decides to reload(?) which results in blinking

cc @HarelM

PR #5276, opened 2024-12-29
Improve typing and change StyleLayer query call interface

### LINKED ISSUE #5255: Globe Projection Bug: Circle Layers have smaller hover/click area than visually represented

HarelM:
## Launch Checklist

This is a pure refactoring step to add more types to the query code flows.
It also add the globe projection matrix to be passed down to the style layers that needs it, but it doesn't solve any problems with exiting code (mainly circle layer).

 - [x] Link to related issues.


HarelM:
I've done a very simple change in the position matrix that is passed to the `queryIntersectsFeature` of every layer.
This seems to still pass the tests.
I now see that there are missing tests for query stuff related to globe (`test/integration/query`).
I'm still not sure how to "see" what I'm writing in terms how to present on the map the intersection perimeter, so I'm still in very early stages of this fix.
Any help would be greatly appreciated.
Cc: @ibesora @kubapelc

HarelM:
I was able to leverage the following method from the transform class:
`projectTileCoordinates` to project tile coordinates to screen coordinates, but I'm not sure it a great solution.
image

In order to use it I need to pass down the tileID, which is not a big deal since the tile is the caller in this case, and it know its own ID.
I can check it in if this would help anyone to better understand what I did.
I think it is similar to @ibesora fix here:  #5267, and it break the same way as this PR when adding `circle-translate`...

HarelM:
Ignore my last comment, it breaks mercator related circle stuff...

HarelM:
This is now ready for review, I'll add comments where I changed something that is not only types.

src/source/query_features.ts
```diff
@@ -122,19 +132,7 @@ export function queryRenderedFeatures(
...
-            const state = sourceCache.getFeatureState(feature.layer['source-layer'], feature.id);
-            feature.source = feature.layer.source;
-            if (feature.layer['source-layer']) {
-                feature.sourceLayer = feature.layer['source-layer'];
-            }
-            feature.state = state;
-        });
-    }
-    return result;
+    return convertFeaturesToMapFeatures(result, sourceCache);
```
HarelM, line 135: I've moved this code to the end of the file to be somewhat reused.

src/style/style_layer.ts
```diff
@@ -56,16 +94,7 @@ export abstract class StyleLayer extends Evented {
...
-        queryGeometry: Array<Point>,
-        feature: VectorTileFeature,
-        featureState: FeatureState,
-        geometry: Array<Array<Point>>,
-        zoom: number,
-        transform: IReadonlyTransform,
-        pixelsToTileUnits: number,
-        pixelPosMatrix: mat4
-    ): boolean | number;
+    queryIntersectsFeature?(params: QueryIntersectsFeatureParams): boolean | number;
```
HarelM, line 97: This an interface change, but it only moves the method parameters to a "type".

src/source/query_features.ts
```diff
@@ -55,18 +62,21 @@ export type QuerySourceFeatureOptions = {
...
+function getPixelPosMatrix(transform: IReadonlyTransform, tileID: OverscaledTileID) {
     const t = mat4.create();
     mat4.translate(t, t, [1, 1, 0]);
     mat4.scale(t, t, [transform.width * 0.5, transform.height * 0.5, 1]);
-    if (transform.calculatePosMatrix) { // Globe: TODO: remove this hack once queryRendererFeatures supports globe properly
-        return mat4.multiply(t, t, transform.calculatePosMatrix(tileID.toUnwrapped()));
-    } else {
-        return t;
-    }
+    const projectionData = transform.getProjectionData({overscaledTileID: tileID});
```
HarelM, line 78: This is the main change in this code which uses the projection data from the transform and adds typings to this method.
As far as I can see this looks OK.
I have removed the `toUnwrapped()` which I'm not sure how it affects the mercator stuff, but the tests do not seem to fail, so I'm not sure if it was needed or not...

ibesora: I think `toUnwrapped` still needs to be there. We are using `tileID.key` as the key into a cache of matrices. Those matrices are the same wether it's wrapped or not but removing `toUnwrapped` will make the cache possibly double in size

ibesora (approved):
Added a minor nit, otherwise this is a nice code improvement!

HarelM:
The problem is that get projection data doesn't support passing unwrapped. I tried changing that but it didn't go well. 
This part needs more work anyway, so I'll revert that and push only the type changes and minor refactoring ATM...

src/source/query_features.ts
HarelM: I've reverted this.

Issue #355, opened 2021-09-18
[accessibility] Interactive markers should have a default accessible name


Malvoz:
Interactive markers (e.g. markers that bind to popups) should have a default [accessible name](https://www.w3.org/TR/accname/#dfn-accessible-name), e.g. `aria-label="Marker"` if not specified.

Related WCAG 2.1 Success Criterion: [4.1.2 Name, Role, Value](https://www.w3.org/TR/WCAG21/#name-role-value) (this issue is to address the <q>Name</q> part of the Success Criterion for interactive markers).

<hr>

Furthermore, ideally developers should ensure that each marker has a unique and descriptive name, such that the user can know what the marker is visually highlighting (in the case of markers binding to popups it'd also allow <span title="Assistive technology">AT</span> users to know what it's for before opening the popup). For example, the marker in https://maplibre.org/maplibre-gl-js-docs/example/set-popup/ would ideally be named "Washington Monument" or similar. Perhaps developers can be warned if markers are not uniquely named?

<hr>

(The same issue was also reported with MapBox: https://github.com/mapbox/mapbox-gl-js/issues/11030.)

PR #5298, opened 2025-01-06
Feature: only add ARIA-LABEL if the element does not already has one

### LINKED ISSUE #5623: [accessibility] Marker should not set element `aria-label` if it is already set
## User Story

Imagine you want to render a marker with a link element like 

```html
<a href="http://foobar.com" aria-label="Foobar" />
```

If you specify the `aria-label`, it should not be overwritten by maplibre.

liorchamla:
- Fixes #5623

Hey folks, this is a wonderful project ! So first thing first, thanks to all contributors, the library is great. We are working a lot on a11y since a week, and the problem we have with `Marker` is that it always has the same `aria-label` derived from a default string `Marker.Title`. But in our use case, we want Markers to be self explainatory to our audience.

Our custom HTML element for the marker has a good `aria-label` but it is replaced automaticly when the Marker is added to the map (`addTo()` method).

So my proposition here is to say : If the element already has an `aria-label` attribute that makes sense in the userland context, let's not replace it with the default Marker.Title.

# Edit
By the way, I just found that there was an issue related (sort of) to my PR : https://github.com/maplibre/maplibre-gl-js/issues/355


liorchamla:
By the way, I can't understand why spellcheck is failing in the CI while it works fine inside my CodeSpace :'( 

If you have any advice on what is missing ?

src/ui/marker.ts
```diff
@@ -316,7 +316,11 @@ export class Marker extends Evented {
     addTo(map: Map): this {
         this.remove();
         this._map = map;
-        this._element.setAttribute('aria-label', map._getUIString('Marker.Title'));
+
+        // If the element does not have an aria-label, we must provide it with the default Marker.Title
```
HarelM, line 320: No need for this comment, but please add a unit test that cover this scenario.

HarelM:
This is related to awesome-maplibre project spelling issue there, I'll see if I can fix this quickly.

src/ui/marker.test.ts
```diff
@@ -258,8 +258,18 @@ describe('marker', () => {
         const marker = new Marker().setLngLat([0, 0]).addTo(map);
 
         expect(marker.getElement().getAttribute('aria-label')).toBe('alt title');
+
+        const customHtmlElement = document.createElement('div');
```
HarelM, line 262: Better to create a new test for it instead of using an existing one.

HarelM (approved):

PR #5317, opened 2025-01-08
made the v5 geometry type change more explicitely documented.

CommanderStorm:
As discussed in the discussion just now, I think a bit of a longer wording for this might be more clear in how it breaks.
Currently this point is somewhat simple to gloss over if reading quickly ^^

To be clear:
Would be fine with closing this PR if you think the docs are currently sufficient.
We are on the same team ^^

PS: Could we also change the [release notes here](https://github.com/maplibre/maplibre-gl-js/releases/tag/v5.0.0)?

## Launch Checklist

 - [x] Link to related issues.
 - [x] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,11 +11,6 @@
 
 ### ✨ Features and improvements
 
-- ⚠️ `StyleLayer`'s `queryIntersectsFeature` method parameters were moved to `QueryIntersectsFeatureParams`. ([#5276](https://github.com/maplibre/maplibre-gl-js/pull/5276)) Wrap the method parameters with `{}` to solve this
-- ⚠️ Support setting WebGL context options on map creation ([#5196](https://github.com/maplibre/maplibre-gl-js/pull/5196)). Previously supported WebGL context options like `antialias`, `preserveDrawingBuffer` and `failIfMajorPerformanceCaveat` must now be defined inside the `canvasContextAttributes` object on `MapOptions`.
-- ⚠️ Change the return type of `on` method to return a `Subscription` to allow for easy unsubscribe ([#5080](https://github.com/maplibre/maplibre-gl-js/pull/5080)). `map.on('x').on('y')` => `map.on('x'); map.on('y');`.
-- ⚠️ Change drag rotate behavior to be around the center of the screen ([#5074](https://github.com/maplibre/maplibre-gl-js/pull/5074))
-- ⚠️ Return actual altitude from queryTerrainElevation + Pass non-translated matrices to custom layer on mercator map ([#3854](https://github.com/maplibre/maplibre-gl-js/pull/3854))
 - ⚠️ Changed `geometry-type` to identify "Multi-" features ([#4877](https://github.com/maplibre/maplibre-gl-js/pull/4877)). Use `$type` which has no "Multi-" support or use `in` expression to get the previous behavior.  
   Example:  
   This change means that a filter like `["==", ["geometry-type"], "Polygon"]` would not display `MultiPolygon` features any more.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -16,12 +16,14 @@
 - ⚠️ Change the return type of `on` method to return a `Subscription` to allow for easy unsubscribe ([#5080](https://github.com/maplibre/maplibre-gl-js/pull/5080)). `map.on('x').on('y')` => `map.on('x'); map.on('y');`.
 - ⚠️ Change drag rotate behavior to be around the center of the screen ([#5074](https://github.com/maplibre/maplibre-gl-js/pull/5074))
 - ⚠️ Return actual altitude from queryTerrainElevation + Pass non-translated matrices to custom layer on mercator map ([#3854](https://github.com/maplibre/maplibre-gl-js/pull/3854))
-- ⚠️ Changed `geometry-type` to identify "Multi-" features ([#4877](https://github.com/maplibre/maplibre-gl-js/pull/4877)). Use `$type` for non "Multi-" support or use `in` expression to get the previous behavior.
-Example:
-```diff
-- "filter": ["==", ["geometry-type"], "Polygon"]
-+ "filter": ["==", "$type", "Polygon"]
-```
+- ⚠️ Changed `geometry-type` to identify "Multi-" features ([#4877](https://github.com/maplibre/maplibre-gl-js/pull/4877)). Use `$type` which has no "Multi-" support or use `in` expression to get the previous behavior.
```
HarelM, line 19: Can we move it to be the first item on the list of breaking changes?

CommanderStorm: Good point

CHANGELOG.md
```diff
@@ -11,17 +11,19 @@
 
 ### ✨ Features and improvements
 
+- ⚠️ Changed `geometry-type` to identify "Multi-" features ([#4877](https://github.com/maplibre/maplibre-gl-js/pull/4877)). Use `$type` which has no "Multi-" support or use `in` expression to get the previous behavior.  
+  Example:  
+  This change means that a filter like `["==", ["geometry-type"], "Polygon"]` would not display `MultiPolygon` features any more.
```
HarelM, line 16:
```suggestion
  ⚠️ This change means that a filter like `["==", ["geometry-type"], "Polygon"]` would not display `MultiPolygon` features any more.
```

HarelM:
Looks like npm and girhub aren't communicating well now.
I'll update the release notes as well with this change.

HarelM (approved):
Thanks!

slaviczavik:
Just a note the `["==", "$type", "Polygon"]` expression cannot be used together with the expression syntax defined by `ExpressionSpecification` type. What is workaround for this?

HarelM:
`in`, `match`, `case` can be used as well, it's just a bit longer and usually `$type` is enough... Feel free to improve the comment if you feel it is still lacking.

slaviczavik:
> `in`, `match`, `case` can be used as well, it's just a bit longer and usually `$type` is enough... Feel free to improve the comment if you feel it is still lacking.

Thanks for quick answer. I would like to use expression with `match` operator something like this:
```
[
  "all",
  [
    "==",
    [
      "geometry-type"
    ],
    "Polygon"
  ],
  [
    "match",
    [
      "get",
      "brunnel"
    ],
    [
      "bridge",
      "tunnel"
    ],
    true,
    false
  ]
]
  ```

What would be correct syntax with `$type` attribute?

**Comment by HarelM** [comment]
I meant to say that you can replace $type with `in`, `match`, `case` to support "non-deprecated" expressions.
Something like the following, I haven't tested it so it might be a bit off...
```diff
-[
-    "==",
-    [
-      "geometry-type"
-    ],
-    "Polygon"
-  ],
+[
+    "in",
+    [
+      "geometry-type"
+    ],
+    [["literal"], ["Polygon", "MultiPolygon"]]
+  ],
```

prusswan:
Could the docs/examples be updated as well? I also ended up with @HarelM's last example, it looks rather convoluted but more understandable than the `$type` version which seems to be getting deprecated (but why?).

PR #5330, opened 2025-01-12
Drag globe near poles with versor

### LINKED ISSUE #5296: Globe panning is reverted when North pole facing the user
See video: moving the mouse on the left of the globe and to the bottom rotates the globe clockwise, unexpectedly:

https://github.com/user-attachments/assets/23d4fd55-1cae-4506-92e2-0ed36c298c0b

A way to solve it would rely on something like versor (https://github.com/d3/versor), tracking the rotation with quaternions.

**maplibre-gl-js version 5.0.0**:

### Expected Behavior

Homogenous panning all over the globe, non-blocking at the poles (for instance: https://observablehq.com/@jcolot/versor-panning)

### Actual Behavior

Unexpected behaviours related to the poles

jcolot:
## Launch Checklist

- fixes #5296

Adds an opt-in quaternion (versor) globe panning mode via `DragPanOptions.fixedBearing`:

```js
map.dragPan.enable({fixedBearing: false});
```

When `fixedBearing: false`, dragging rotates the globe with a single quaternion rotation that couples center and bearing, so panning stays smooth near and across the poles (where the bearing-preserving approach degenerates), at the cost of letting the bearing change during the drag. Pan inertia for this mode is computed in angular space and eased with a quaternion slerp, so flings keep moving even when the fling target leaves the visible globe.

**The default drag behavior is unchanged** — with `fixedBearing` unset (or `true`), globe panning still uses the existing delta-based path. The versor rotation lives in a standalone `versorSetLocationAtPoint` helper; `ITransform.setLocationAtPoint` keeps its original signature.

Also included:
- Hardening of the fixed-bearing `setLocationAtPoint` solver: a near-pole fallback to pure longitude rotation, and clamping to the tangent point when no exact bearing-preserving solution exists (previously the drag would freeze).
- `angleToRotateBetweenVectors2D` now clamps the dot product before `acos`, fixing NaN results from rounding errors. Note: this changes the `flyTo` "noop with offset" globe behavior — the offset target computation previously relied on an unclamped `acos` returning NaN; the flight now collapses to the same result as the equivalent `easeTo` (test expectation updated from lng `+84.495` to `-84.495`).
- A new example, `pan-a-globe-with-a-free-bearing`, toggling the option near the North Pole (docs image included).

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


src/geo/projection/vertical_perspective_transform.ts
```diff
@@ -49,187 +50,247 @@ export class VerticalPerspectiveTransform implements ITransform {
     get pixelsToClipSpaceMatrix(): mat4 {
         return this._helper.pixelsToClipSpaceMatrix;
     }
+
```
HarelM, line 53: Better to avoid these kind of changes as part of this PR, I would recommend moving this to a different PR if this is truly needed.

kubapelc:
I played around with this for a while, here are some thoughts on the controls:

- I think it would be very good UX-wise to find a way to make the whole screen usable for panning, not just the part covered by the globe.
- The map still rolls slightly when panning around Europe at zoom levels around 5 to 7. At this point the map looks pretty flat, maybe it would be more intuitive to transition to fixed-bearing panning at this point?
- The bearing change from the new controls sometimes clashes with the "snap to north" behaviour MapLibre does (eg. in the case described above). Maybe it should be disabled if fixed bearing panning isn't in use?

It works quite nicely overall!

NazCodeland:
Looking forward to this!

HarelM:
@jcolot any updates on this?
Are you planning on moving this PR for review or should I close it?

jcolot:
@HarelM  Sorry for this unfinished PR. If it's ok, I will try to work on it this month. From what I remember from when I stopped working on it is that it would maybe be nice to have two options for dragging: The classic one, and the one I introduce with versor calculations. What still needed to be done was fixing the easing, which should not be the same when using versor, and add tests. Happy to discuss.

clement-igonet:
The versor panning here is really useful for near-pole dragging (#8003). Since this PR has been idle, I rebased its core onto main as a reduced, focused version in #8017 (versor drag + `fixedBearing` option + example + tests, credit to you). Happy to fold in any changes you'd prefer, or close mine if you'd rather pick this back up.

HarelM:
I'm not sure which PR to review.
@clement-igonet @jcolot please align on which PR should be used and close the other one. Thanks!

jcolot:
Closing in favour of #8017. The work here is merged into that branch via clement-igonet/maplibre-gl-js#1, so it is all in one place now. Thanks @clement-igonet for picking it up.

PR #5343, opened 2025-01-13
Add Feature Properties Transform

neodescis:
This is a furtherance of @wipfli's https://github.com/maplibre/maplibre-gl-js/pull/4199, with the aim to solve https://github.com/maplibre/maplibre-gl-js/issues/4198.

Things I've done since the original:
- Tweaked transform function signature so it returns transformed properties instead of modifying them in situ.
- Added Re-encoding of "rawTileData" using `vtpbf` after feature transforms to ship it back to the main thread. By doing this, `queryRenderedFeatures()` now works as expected.
- Updated example to add a click event handler to demonstrate the functional `queryRenderedFeatures()` call, and also to transform a feature from a GeoJSON source.
- Investigated transforming geometries as well. Unfortunately this is very complex given that a GeoJSON geometry would need to be converted back to an MVT geometry representation.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


src/source/vector_tile_worker_source.ts
```diff
@@ -133,7 +133,7 @@ export class VectorTileWorkerSource implements WorkerSource {
             try {
                 const result = await parsePromise;
                 // Transferring a copy of rawTileData because the worker needs to retain its copy.
-                return extend({rawTileData: rawTileData.slice(0)}, result, cacheControl, resourceTiming);
+                return extend({rawTileData: result.rawTileData ?? rawTileData.slice(0)}, result, cacheControl, resourceTiming);
```
neodescis, line 136: rawTileData already existed in the parse() results type, but it wasn't actually used in WorkerTile before this

neodescis:
> Thank you for taking this up @neodescis! I don't know if we need a getFeaturePropertiesTransform actually since the user always sets it. Regarding raw tile data I am not sure, but I would do the least amount of work for normal operations, and do some extra work if needed if someone calls queryRenderedFeatures.

I left the getter in for symmetry, but I can take it out if that's what's wanted. As for queryRenderedFeatures, I believe this updating of rawTileData needs to be done up-front, since the data needs to make its way back to the main thread (ultimately into the FeatureIndex), and queryRenderedFeatures is synchronous.

src/source/worker_tile.ts
```diff
@@ -188,12 +213,30 @@ export class WorkerTile {
         }
 
         this.status = 'done';
+
+        // If any features were transformed, re-encode to rawTileData to override the original
+        let transformedRawData: ArrayBuffer = null;
+        if (Object.keys(transformedFeatures).length) {
+            const transformedTile: VectorTile = {
```
neodescis, line 220: This is a bit funky, but the VectorTile classes don't allow you to modify their data directly. Instead, this is overriding the layer's function that retrieves its features, so vtpbf can process using the updated data.

neodescis:
I've spent some time investigating geometry overrides, and I don't think the effort is worth it:
- The transformed geometries would still be limited to a given tile, which makes overriding them awkward.
- For the API to be usable, an override would need to be in the form of a GeoJSON Geometry, but converting from that to the vector tile geometry format is complex and I haven't been able to find anything that can be leveraged to do it directly. geojson-vt is close, but its inner workings are not exposed.

neodescis:
On the other hand, I've also looked at how this works with `type: 'geojson'` sources, and it seems to pretty much work the same as vector tiles. So, I think that just leaves tests at this point.

src/source/worker_tile.ts
```diff
@@ -99,6 +109,24 @@ export class WorkerTile {
             for (let index = 0; index < sourceLayer.length; index++) {
                 const feature = sourceLayer.feature(index);
                 const id = featureIndex.getId(feature, sourceLayerId);
+                if (WorkerTile.featurePropertiesTransform) {
+                    const {geometry} = feature.toGeoJSON(this.tileID.canonical.x, this.tileID.canonical.y, this.tileID.canonical.z);
```
neodescis, line 113: I believe the z here should be canonical and not the overscaled value, since this is used to convert the MVT geometry to GeoJSON. Is that correct?

neodescis:
@HarelM, I have tests working now. The example also works, but is maybe a little complex since it encompasses a GeoJSON source and vector tiles? Also, I'm not sure exactly how to document this API, since it's only available in the worker thread.

HarelM:
I'll review this in the coming days thoroughly.
Can you elaborate why there's a need to re-encode the tiles? 
More so, if this is the case, does it have better performance than addProtocol (which can be made to run in a worker thread as well)?

neodescis:
Currently the "rawTileData" is shipped back to the main thread, and it is passed into the FeatureIndex there to be used by queryRenderedFeatures. The raw data is read every time queryRenderedFeatures is called. You can see this in `FeatureIndex.query()` which calls `loadVTLayers()`.

So, if we want to transform features AND have queryRenderedFeatures return them as they have been transformed, the raw data needs to be updated.

Now, I fully recognize that, from a design perspective, we may not actually want to handle things this way. If we think that `addProtocol` is the appropriate path for achieving something like this, I'm fine with that. If that is the case, I would be happy to work up an example that does so in a separate pull request. I believe performance should be similar either way, since the vector tiles will need to be decoded, modified, and re-encoded in either approach. I have not yet tried `addProtocol` for my particular use case, so I haven't been able to compare them yet.

neodescis:
Well, I went down the `addProtocol` route to do this instead, and I think I like it better. I guess I didn't realize that requests were dispatched back to the main thread to take advantage of a protocol added there. This is really nice, because then I don't need to import a script into the workers! I still have to parse and re-encode the vector tile, but it's much more centralized.

So, my recommendation would be to create an example that does this using `addProtocol` and abandon this (and @wipfli's) pull requests. I can create such a pull request in the coming days.

HarelM:
I still think it would be great to have an example of addProtocol in a worker if we don't have such an example yet, but yes, addProtocol is very powerful.

neodescis:
Closing this in favor of https://github.com/maplibre/maplibre-gl-js/pull/5370

Issue #5341, opened 2025-01-13
Add button Option for DragPan Enable or Map Configuration


gottesman:
## Why this Feature Request?

It would be beneficial to allow developers to customize the mouse button used for the dragPan interaction in MapLibre GL JS. Currently, the dragPan handler is tied to the left mouse button, with no built-in support for customizing this behavior. Adding a button option to the dragPan.enable() method or allowing configuration during map initialization would enable greater flexibility for developers.

This feature is already implemented in some other mapping libraries, making it an expected functionality for developers transitioning to MapLibre GL JS.

## Proposed Solution

1. Introduce an optional button parameter in the dragPan.enable() method, which specifies the mouse button to trigger the drag pan interaction. For example:

```
map.dragPan.enable({ button: 1 }); // Enables middle mouse button for dragPan
```

2. Alternatively, add a configuration option in the map's initialization settings:

```
const map = new maplibregl.Map({
    container: 'map',
    style: 'https://demotiles.maplibre.org/style.json',
    center: [0, 0],
    zoom: 2,
    dragPanButton: 1 // Enables middle mouse button for dragPan
});
```

## Use Cases

Applications requiring non-standard drag interactions, such as:

- Using the middle mouse button for drag panning to avoid conflicts with other left-button interactions.
- Customizing drag interactions to align with user preferences or specific application requirements.

## Workaround Used Now

Currently, for my temporary implementation, I'm using custom event handling. Although not perfect, performant, or native, it works (for anyone trying to implement this unofficially):

First I created a variable for the handler:
```
const dragPan = {
    enabled: true,
    isDragging: false,
    startPos: null,
    lastPos: null,
    velocity: [0, 0],
    inertiaAnimation: null,
    dragButtons: { 0:true, 1: true },
    decay : 0.8,
    minVelocity : 0.1,
};
```
Then I disabled the dragPan functionality after initializing the map:
```
map.dragPan.disable();
```

Finally, I added the implementation:
```
const canvas = map.getCanvas();
canvas.addEventListener('mousedown', (e) => {
    if (e.button in dragPan.dragButtons && dragPan.enabled) {
        dragPan.isDragging = true;
        dragPan.startPos = [e.clientX, e.clientY];
        dragPan.lastPos = dragPan.startPos;
        dragPan.velocity = [0, 0];
        canvas.style.cursor = 'grabbing';
        e.preventDefault();

        if (dragPan.inertiaAnimation) {
[...]

PR #5354, opened 2025-01-14
Enable Multi-MouseButton Support for dragPan

gottesman:
This PR enhances the functionality of dragPan in MapLibre GL JS, addressing feature request #5341.
It introduces the ability to configure one or multiple mouse buttons for the dragPan interaction using the [MouseEvent.button](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button) property.

Previously, dragPan was limited to left mouse button, restricting user interaction and personalization. This improvement provides greater flexibility for users who want to customize the drag behavior of the map.

## Examples

**1. Configuring on Map Initialization:**

``` javascript
const map = maplibregl.Map({
    container: 'map',
    dragPan: {
        buttons: [0,1]  // Enable left button (0) and middle button (1)
    }
});
```

**2. Updating on DragPan Enable:**

``` javascript
map.dragPan.enable({
    buttons: [1]  // Enable only middle button (1)
})
```

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/ui/handler/drag_move_state_manager.ts
+++ b/src/ui/handler/drag_move_state_manager.ts
@@ -1,12 +1,14 @@
 import {DOM} from '../../util/dom';
 
 const LEFT_BUTTON = 0;
+const MIDDLE_BUTTON = 1;
 const RIGHT_BUTTON = 2;
 
 // the values for each button in MouseEvent.buttons
 const BUTTONS_FLAGS = {
     [LEFT_BUTTON]: 1,
-    [RIGHT_BUTTON]: 2
+    [MIDDLE_BUTTON]: 4,
+    [RIGHT_BUTTON]: 2,
 };
 
 function buttonNoLongerPressed(e: MouseEvent, button: number) {
@@ -36,14 +38,16 @@ export interface DragMoveStateManager<E extends Event> {
     isValidStartEvent: (e: E) => boolean;
     isValidMoveEvent: (e: E) => boolean;
     isValidEndEvent: (e?: E) => boolean;
+    updateButtons?: (buttons: number[]) => void;
 }
 
 export class MouseMoveStateManager implements DragMoveStateManager<MouseEvent> {
     _eventButton: number | undefined;
-    _correctEvent: (e: MouseEvent) => boolean;
+    _correctEvent: (e: MouseEvent, buttons?: number[]) => boolean;
+    _buttons: number[] = [LEFT_BUTTON];
 
     constructor(options: {
-        checkCorrectEvent: (e: MouseEvent) => boolean;
+        checkCorrectEvent: (e: MouseEvent, buttons?:number[]) => boolean;
     }) {
         this._correctEvent = options.checkCorrectEvent;
     }
@@ -58,7 +62,7 @@ export class MouseMoveStateManager implements DragMoveStateManager<MouseEvent> {
     }
 
     isValidStartEvent(e: MouseEvent) {
-        return this._correctEvent(e);
+        return this._correctEvent(e, this._buttons);
     }
 
     isValidMoveEvent(e: MouseEvent) {
@@ -74,6 +78,10 @@ export class MouseMoveStateManager implements DragMoveStateManager<MouseEvent> {
     isValidEndEvent(e: MouseEvent) {
         const eventButton = DOM.mouseButton(e);
         return eventButton === this._eventButton;
+    }
+
+    updateButtons(buttons: number[]) {
+        this._buttons = buttons;
     }
 }
 
@@ -116,7 +124,7 @@ export class OneFingerTouchMoveStateManager implements DragMoveStateManager<Touc
 
 export class MouseOrTouchMoveStateManager implements DragMoveStateManager<MouseEvent | TouchEvent> {
     constructor(
-        private mouseMoveStateManager = new MouseMoveStateManager({checkCorrectEvent: () => true}), 
+        private mouseMoveStateManager = new MouseMoveStateManager({checkCorrectEvent: () => true}),
         private oneFingerTouchMoveStateManager = new OneFingerTouchMoveStateManager()
     ) {}
 
--- a/src/ui/handler_inertia.ts
+++ b/src/ui/handler_inertia.ts
@@ -40,7 +40,6 @@ export type InertiaOptions = {
     easing: (t: number) => number;
     deceleration: number;
     maxSpeed: number;
-    buttons?: number[]; // buttons that can be used to trigger pan
 };
 
 export class HandlerInertia {
--- a/src/ui/handler_inertia.ts
+++ b/src/ui/handler_inertia.ts
@@ -40,6 +40,7 @@ export type InertiaOptions = {
     easing: (t: number) => number;
     deceleration: number;
     maxSpeed: number;
+    buttons?: number[]; // buttons that can be used to trigger pan
 };
 
 export class HandlerInertia {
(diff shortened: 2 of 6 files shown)


HarelM:
How will this behave when rotating with right click (default behavior) and setting drag pan to right click?

src/ui/handler_inertia.ts
```diff
@@ -40,6 +40,7 @@ export type InertiaOptions = {
     easing: (t: number) => number;
     deceleration: number;
     maxSpeed: number;
+    buttons?: number[]; // buttons that can be used to trigger pan
```
HarelM, line 43: Is this really needed here?

src/ui/handler/drag_move_state_manager.ts
```diff
@@ -58,7 +62,7 @@ export class MouseMoveStateManager implements DragMoveStateManager<MouseEvent> {
     }
 
     isValidStartEvent(e: MouseEvent) {
-        return this._correctEvent(e);
+        return this._correctEvent(e, this._buttons);
```
HarelM, line 65: Instead of passing buttons it might be better to pass a predicate to override the default one - what happens if someone wants to have right click with control button for panning?

src/ui/handler_inertia.ts
gottesman: Oh, my bad. It's not, so this can be ignored. Probably a remaining code while working out my solution. Thank you

src/ui/handler/drag_move_state_manager.ts
HarelM: _correctEvent is basically what you want to override. 
I would try and find a way to pass it down instead if adding a buttons parameter and pass that down...

HarelM:
Any updates on the feedback and questions I asked?

jiacona:
Hi @HarelM, I also need this functionality and am most of the way to an implementation that addresses your comments here. Should I just open a new PR?

HarelM:
My main concern about this feature is button collision when you override one gesture and "cancel" other gesture.

jiacona:
Yes, I can see how you could run into problems if the predicate you use conflicts with another handler. Is it rotate, pitch, and roll you are concerned about?

I was thinking it would be simple enough to allow specifying a `checkCorrectEvent` predicate for each of the different handlers. That way if you want to customise one, you can also customise any others that might conflict with your new predicate.

HarelM:
There's a few more handlers: zoom, context menu, scroll, two finger gesture/block.
I think adding configuration for all of those might be complicated, but it's possible to add a map from handler to predicate as part of the map options, similar to localization (to some extent).
I'm not sure it would be very useful, but I guess it's an option...

jiacona:
I'm wondering if we're making this all more complicated than it needs to be. I really only need dragPan on middle mouse (in addition to the existing buttons), and I suspect that is all the original issue author wanted as well. What would you think about dropping the idea of button configuration and just make dragPan handled on middle-mouse by default?

HarelM:
I think that you can hack the following in theory (I haven't fully tested it):
`(map.dragPan._mousePan as any)._moveStateManager._correctEvent =...`
It is very specific to how maplibre is currently structured, but it might just work.
Making a patch that only support this use case without defining a proper API to support this will introduce backward compatibility issues in the future I believe.

jiacona:
> I think that you can hack the following in theory (I haven't fully tested it): `(map.dragPan._mousePan as any)._moveStateManager._correctEvent =...` It is very specific to how maplibre is currently structured, but it might just work. 

Ah yes, interesting. I would be pretty hesitant to reach into the internals like that without a proper API.

> Making a patch that only support this use case without defining a proper API to support this will introduce backward compatibility issues in the future I believe.

This is the minimal change I am proposing. Are there reasons not to want to support drag pan with middle mouse by default?

```diff
diff --git a/src/ui/handler/drag_move_state_manager.ts b/src/ui/handler/drag_move_state_manager.ts
index 5ef48fb72..a8b5b0097 100644
--- a/src/ui/handler/drag_move_state_manager.ts
+++ b/src/ui/handler/drag_move_state_manager.ts
@@ -1,12 +1,14 @@
 import {DOM} from '../../util/dom';

 const LEFT_BUTTON = 0;
+const MIDDLE_BUTTON = 1;
 const RIGHT_BUTTON = 2;

 // the values for each button in MouseEvent.buttons
 const BUTTONS_FLAGS = {
     [LEFT_BUTTON]: 1,
-    [RIGHT_BUTTON]: 2
+    [RIGHT_BUTTON]: 2,
+    [MIDDLE_BUTTON]: 4,
 };

 function buttonNoLongerPressed(e: MouseEvent, button: number) {
diff --git a/src/ui/handler/mouse.ts b/src/ui/handler/mouse.ts
index 30dd66471..1e712a6cd 100644
--- a/src/ui/handler/mouse.ts
+++ b/src/ui/handler/mouse.ts
@@ -23,6 +23,7 @@ export interface MousePitchHandler extends DragMoveHandler<DragPitchResult, Mous
 export interface MouseRollHandler extends DragMoveHandler<DragRollResult, MouseEvent> {}

 const LEFT_BUTTON = 0;
+const MIDDLE_BUTTON = 1;
 const RIGHT_BUTTON = 2;

 const assignEvents = (handler: DragHandler<DragPanResult, MouseEvent>) => {
@@ -39,7 +40,9 @@ export function generateMousePanHandler({enable, clickTolerance}: {
     enable?: boolean;
 }): MousePanHandler {
     const mouseMoveStateManager = new MouseMoveStateManager({
-        checkCorrectEvent: (e: MouseEvent) => DOM.mouseButton(e) === LEFT_BUTTON && !e.ctrlKey,
+        checkCorrectEvent: (e: MouseEvent) =>
+            (DOM.mouseButton(e) === LEFT_BUTTON && !e.ctrlKey) ||
+            (DOM.mouseButton(e) === MIDDLE_BUTTON),
     });
     return new DragHandler<DragPanResult, MouseEvent>({
         clickTolerance,
```

HarelM:
I don't think this is the same as the original issue, I'm also not sure why you'd need to support both left click and middle click to drag...

jiacona:
Yes, you are correct. And as I've been exploring this more for our own use I've run into the issue you were worried about with conflicting handlers.

For now I'm using your suggestion to do this
```ts
(map.dragPan._mousePan as any)._moveStateManager._correctEvent =...
```

It doesn't quite work on it's own if you want middle mouse support because middle mouse is not considered valid by `MoveStateManager.isValidMoveEvent` so I had to patch that as well:
```ts
export const LEFT_BUTTON = 0
export const MIDDLE_BUTTON = 1
export const RIGHT_BUTTON = 2

export const BUTTONS_FLAGS = {
  [LEFT_BUTTON]: 1,
  [MIDDLE_BUTTON]: 4,
  [RIGHT_BUTTON]: 2,
}

const manager = (map.dragPan as any)._mousePan._moveStateManager
manager.isValidMoveEvent = function (e: MouseEvent): boolean {
  if (this._eventButton === undefined) return false

  const flag = BUTTONS_FLAGS[this._eventButton]
  if (flag === undefined) return false

  return valid = e.buttons !== undefined && (e.buttons & flag) === flag
}
```

I'll see if a reasonable pattern emerges for setting groups of handlers in a coherent way as I'm implementing it for our use case.

johncarmack1984:
The open question since 2025-11 is the API shape (a predicate rather than a button list), and nobody has proposed one; #5341 stays open for that. The _correctEvent override in the thread is what works today. This one can be closed.

PR #5370, opened 2025-01-17
Add example using addProtocol to transform vector tile feature properties

neodescis:
Provides an example of how to solve https://github.com/maplibre/maplibre-gl-js/issues/4198 using addProtocol.

IMO, this should be favored over https://github.com/maplibre/maplibre-gl-js/pull/5343 and https://github.com/maplibre/maplibre-gl-js/pull/4199.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


neodescis:
@wipfli, check this out. This is how I solved the problem using addProtocol. I believe this is cleaner than our other approach. I'm curious to hear what you think!

test/examples/feature-properties-transform.html
```diff
@@ -0,0 +1,65 @@
...
+
+        const map = new maplibregl.Map({
+            container: 'map',
+            style: 'https://demotiles.maplibre.org/style.json',
+            center: [8, 47],
+            zoom: 5,
+            hash: 'map'
+        });
+
+        map.setTransformRequest((url, resourceType) => {
```
HarelM, line 58: Shouldn't this return something in "regular" cases?

HarelM (approved):
This is a very nice examples!

HarelM:
It would be interesting to see how this example can be done in a worker thread instead of the main thread, but this is a "nice to have".
In any case, I think the other two PRs can be closed, this is a good enough way to solve this I think...

test/examples/feature-properties-transform.html
neodescis: If a request transform doesn't return, or returns undefined, then nothing happens. I can add the `return undefined` to be clear though.

wipfli:
Amazing, thanks for writing this example @neodescis. It great to see how this works with add protocol, and also it is interesting to have now an example which uses `import` in the html. I always wondered how to do this...

PR #5138, opened 2024-12-01
Rollback to mapbox gl draw example

birkskyum:
I think it's best to roll back this for now. I learned a ton about the library in the process of migrating, and want to reconsider the approach taken.

This mapbox gl draw examples still has it's turf.js dependency upgraded compared to before.

## Launch Checklist

 - [X] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!

PR #5393, opened 2025-01-23
update mapbox gl draw example

andrewharvey:
## Launch Checklist

As raised via a comment in #5138, the [mapbox-gl-draw example](https://maplibre.org/maplibre-gl-js/docs/examples/mapbox-gl-draw/) currently provides an example of using mapbox-gl-draw however the description mentioned maplibre-gl-draw.

This PR updates the description to match the example given.

We also update to mapbox-gl-draw 1.5.0 since from this version the CANVAS class is exposed which we need to override to match the maplibre class. This is necessary for the delete a vertex feature to work and possibly other features.

Is this done for examples as well?

### DIFF
--- a/test/examples/mapbox-gl-draw.html
+++ b/test/examples/mapbox-gl-draw.html
@@ -31,10 +31,10 @@
         font-size: 13px;
     }
 </style>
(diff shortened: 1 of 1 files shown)


HarelM (approved):
Thanks!
This will be available when the next release goes out.

Issue #5394, opened 2025-01-23
Add getBounds() method to geoJSON sources


AbelVM:
As requested in the Slack thread linked below

https://osmus.slack.com/archives/C01G3D28DAB/p1737627086709479

The code would be as follows

```javascript
async function _getbounds() {
	const bounds = new maplibregl.LngLatBounds();
	const data = await this.getData();
	const features = data.features;
	features.forEach(function(feature) {
		const c = feature.geometry.coordinates.flat(Infinity);
		for(let i = 0; i < c.length - 1; i += 2){
			bounds.extend([c[i], c[i+1]]);
		}	
	});
	return bounds;
}

maplibregl.GeoJSONSource.prototype.getBounds = _getbounds;
```
It works regardless the geometry type, or even with mixed geometry types. Testing it with the [Add multiple geometries from one GeoJSON source](https://maplibre.org/maplibre-gl-js/docs/examples/multiple-geometries/)

```javascript
const mysource = map.getSource('national-park');
> $ {id: 'national-park', type: 'geojson', minzoom: 0, maxzoom: 18, tileSize: 512, …}

const bounds = await mysource.getBounds()
> V {_sw: Dc, _ne: Dc}

JSON.stringify(bounds)
> '{"_sw":{"lng":-121.577415,"lat":40.409124},"_ne":{"lng":-121.246768,"lat":40.584978}}'
```

**EDITED:** simpler for loop, more performant

HarelM:
I think this can be made to be an example, not sure it's worth the maintenance overhead. But I might be too strict... Your call...

AbelVM:
IMHO, it's quite a trivial block of code that 

* shouldn't require much maintenance, as its dependencies are quite limited to core classes and methods that shouldn't change in time
* easy to implement in trunk
* it's compatible with any type of geoJSON geometries, even when mixed in the same file or geometry collections
* enables the classical feature `zoom to layer` for geoJSON sources (the only source which bounds might be known without messing with the arbitrary metadata)

From my point of view, the benefits of adding this method are way above the costs 😄 

**NOTE:** I've updated the snippet I made yesterday in a 5-minute break so that it can avoid extra, unneeded, steps

AbelVM:
Faster (and ES6+) version

```javascript
async function _getbounds() {
	const bounds = new maplibregl.LngLatBounds();
	const data = await this.getData();
	const coordinates = data.features
[...]

PR #5403, opened 2025-01-24
Added geoJSON source getBounds method

AbelVM:
Resolves https://github.com/maplibre/maplibre-gl-js/issues/5394


src/source/geojson_source.test.ts
```diff
@@ -506,3 +507,23 @@ describe('GeoJSONSource#serialize', () => {
...
     });
 });
+
+describe('GeoJSONSource#getBounds', () => {
+    function createSource(opts?) {
+        opts = opts || {};
+        opts = extend(opts, {data: hawkHill});
+        return new GeoJSONSource('id', opts, wrapDispatcher({
+            sendAsync(_message) {
+                return new Promise((resolve) => {
```
HarelM, line 517: There's a sleep utility for that...

src/source/geojson_source.ts
```diff
@@ -249,6 +250,24 @@ export class GeoJSONSource extends Evented implements Source {
...
+    async getBounds(): Promise<LngLatBounds> {
+        const bounds = new LngLatBounds();
+        const data = await this.getData();
+        const coordinates = data.features
+            .map(f => f.geometry.coordinates.flat(Infinity))
+            .reduce((acc, cur) => [...acc, ...cur],[]);
+        for(let i = 0; i < coordinates.length - 1; i += 2){
+            bounds.extend([coordinates[i], coordinates[i+1]]);
+        }
+        return new Promise((resolve, reject) => resolve(bounds));
```
HarelM, line 268: Why but simply return the bounds here?

HarelM:
There are problems with transpiling the code, and some nitpicking, but otherwise this looks good, thanks!

src/source/geojson_source.test.ts
AbelVM: I just copied & pasted the function from `GeoJSONSource#setData` test :sweat_smile: . I'm gonna change it right now

src/source/geojson_source.ts
AbelVM: As the method is `async` it looks I'm forced to return a `Promise` :shrug:

HarelM: No, you can return a regular value, async means it will be converted to a promise...

AbelVM: Roger that

AbelVM:
Changes committed and changelog message added :nerd_face:

package.json
```diff
@@ -78,7 +78,7 @@
     "@typescript-eslint/eslint-plugin": "^8.21.0",
     "@typescript-eslint/parser": "^8.21.0",
     "@vitest/coverage-v8": "3.0.3",
-    "@vitest/ui": "3.0.3",
+    "@vitest/ui": "3.0.4",
```
HarelM, line 81: This is causing issues due to incorrect peer dependencies...

AbelVM: Hmmm, that line is not mine! :thinking:  I just pulled from `main` before committing, so it looks like it took it from there

HarelM: This change isn't in main...

AbelVM: Fixed. Maybe a `dependabot` glitch on my fork

AbelVM:
Ok, something is weird here. It looks like `getData()` is coded to return `GeoJSON.GeoJSON` but it's returning `GeoJSON<Geometry, { [name: string]: any; }>` instead, breaking the tests.

Any help is welcome

HarelM:
GeoJson.Geojson is both feature collection and feature. You need to check which one you received and act accordingly.

AbelVM:
Ok, it looks fine now

src/source/geojson_source.ts
```diff
@@ -249,6 +251,48 @@ export class GeoJSONSource extends Evented implements Source {
...
+    /**
+     * Allows to get the source's actual boundaries.
+     *
+     * @returns a promise which resolves to the source's actual boundaries
+     */
+
+    getBounds(): LngLatBounds {
+        type justGeometry = Point | MultiPoint | LineString | MultiLineString | Polygon | MultiPolygon;
+        const bounds = new LngLatBounds();
+        const data = this._data as GeoJSON.GeoJSON;
```
HarelM, line 263: This casting is invalid in case geojson is a url and not an object - these types of casing are dangerous...

src/source/geojson_source.ts
```diff
@@ -249,6 +251,48 @@ export class GeoJSONSource extends Evented implements Source {
...
+        let coordinates: any[];
```
HarelM, line 264: This shouldn't be "any".

src/source/geojson_source.ts
```diff
@@ -249,6 +251,48 @@ export class GeoJSONSource extends Evented implements Source {
...
+                    }
+                });
+                break;
+            case 'GeometryCollection':
+                coordinates = data.geometries.map((g: justGeometry) => g.coordinates);
+                break;
+            default:
+                coordinates = data.coordinates;
+        }
+        if(!!!coordinates) return bounds;
```
HarelM, line 288: This is not needed in case of typescript... just check if the array is empty or not...

HarelM:
Seems like the test coverage of the code is a bit lacking...

HarelM:
I guess you should turn off dependabot on your fork...

AbelVM:
Sorry for the misleading comment, it was more of a self-note than a PR comment.

And damned dependabot, it sure was on by default

src/source/geojson_source.ts
AbelVM: I'm having blocking issues with this.getData() in this context as it resolves immediately without error and returns nothing, while this._data contains the right data

AbelVM:
Once I solve the getData issue, missing unit tests will be added

HarelM:
getData is suppose to send a message to the worker and get back the data from there, that's why it's async, it might need work to work correctly in the unit test.

AbelVM:
Any clue on how to move on? I've been following the red thread across the code base, but I haven't found the actual reason why `getData` gets resolved but empty in this context. Quite a show stopper :sweat:

HarelM:
Try adding a test to the browser.test.ts (integration tests) instead of unit test, it might be easier...

AbelVM:
Same results

HarelM:
Can't help with a code I can't see, try checking it in so that I can take a look...

AbelVM:
Sorry, I tested it locally. Now its uploaded

test/integration/browser/browser.test.ts
```diff
@@ -6,6 +6,58 @@ import type {AddressInfo} from 'net';
 import type {default as MapLibreGL, Map} from '../../../dist/maplibre-gl';
 import {sleep} from '../../../src/util/test/util';
 
+import {type Actor} from '../../../src/util/actor';
```
HarelM, line 9: These should not be added.
You should write a test as if you write code inside the browser - create a map, add a geojson source, add a layer, and query the bounds of that source.

AbelVM:
Ok, now everything related to this feature is working as expected :tada: 

Coverage is low because there's no unit test as they're run as integration tests. I have no idea how to improve that, as `getData()` doesn't work within unit tests

test/integration/browser/browser.test.ts
```diff
@@ -5,6 +5,7 @@ import http, {type Server} from 'http';
 import type {AddressInfo} from 'net';
 import type {default as MapLibreGL, Map} from '../../../dist/maplibre-gl';
 import {sleep} from '../../../src/util/test/util';
+import type {LngLatBounds} from '../../../src/geo/lng_lat_bounds';
```
HarelM, line 8: Don't import from source, import from dist like the other imports.

test/integration/browser/browser.test.ts
```diff
@@ -448,4 +449,213 @@ describe('Browser tests', () => {
...
+        let testbounds: LngLatBounds;
+        const bounds = await page.evaluate(async () => {
+            testbounds = new maplibregl.LngLatBounds([-122.49378204345702, 37.82880236636284, -122.48339653015138, 37.83381888486939]);
+            const map = new maplibregl.Map({
+                container: 'map',
+                style: 'https://demotiles.maplibre.org/style.json',
+                center: [10, 10],
+                zoom: 10
+            });
+            map.on('load', () => {
```
HarelM, line 463: Use `await once(...)` to make less indentation and make sure the test waits from the map to load.

test/integration/browser/browser.test.ts
```diff
@@ -448,4 +449,213 @@ describe('Browser tests', () => {
...
+                map.addSource('testid', {
+                    type: 'geojson',
+                    data: {
+                        'type': 'FeatureCollection',
+                        'features': [{
+                            'type': 'Feature',
+                            'properties': {},
+                            'geometry': {
+                                'type': 'LineString',
+                                'coordinates': [
```
HarelM, line 473: Can the geometry be simplified (2-3 points)?

test/integration/browser/browser.test.ts
```diff
@@ -448,4 +449,213 @@ describe('Browser tests', () => {
         expect(center.lng).toBeCloseTo(11.39770);
         expect(center.lat).toBeCloseTo(47.29960);
     });
+
+    test('GeoJSONSource#getBounds: FeatureCollection', async () => {
+        let testbounds: LngLatBounds;
+        const bounds = await page.evaluate(async () => {
+            testbounds = new maplibregl.LngLatBounds([-122.49378204345702, 37.82880236636284, -122.48339653015138, 37.83381888486939]);
```
HarelM, line 456: Use simpler numbers (less digits after the dot) to improve test readability.

test/integration/browser/browser.test.ts
AbelVM: Sure, but I've just reused the data in the unit tests

HarelM: Then the data there probably needs simplification as well...

test/integration/browser/browser.test.ts
AbelVM: One of the existing imports comes from dist and the other one from source, what's the rule here? (for future me reference :sweat_smile: )

HarelM: `sleep` is a test util, so it needs to be coming from the test utils code, other stuff (actual production code) should come from dist, for this integration test only, for unit tests, everything should be coming from the code.

AbelVM: Roger that!

test/integration/browser/browser.test.ts
AbelVM: done

test/integration/browser/browser.test.ts
AbelVM: Done

test/integration/browser/browser.test.ts
AbelVM: done

test/integration/browser/browser.test.ts
AbelVM: It looks like `await map.once('load');` doesn't work as expected. 

- [ `.once()` only works with `MapLayerEventType`, and `load` is a MapEventType, so... not supported](https://maplibre.org/maplibre-gl-js/docs/API/classes/Map/#once)
- [`.on()` works with `MapEventType`, but the listener is compulsory and returns a Subscription, not a Promise](https://maplibre.org/maplibre-gl-js/docs/API/classes/Map/#on)

I need to move back to `map.on('load' =>{...})`

(and we might need to change the behavior of `once()` to reflect the one of `on()`)

HarelM: I'm not sure what you are talking about, the unit tests are filled with `await once('load');`, here's one for example:
https://github.com/maplibre/maplibre-gl-js/blob/bedcb8622337d73875638d165f10b0397f18d775/src/ui/map_tests/map_basic.test.ts#L80

AbelVM: It might look like some common methods behave differently, hanging on the test context: unit, or integration tests. 

Just in this PR, two methods, `getData()` and `map.once()`, are behaving differently

HarelM: on amd once are different...
Using await on(...) shouldn't work as it's not designed to work this way. 
Regarding the tests failures after the once change, the error message indicate that the expected object is "{}" which I find odd, regardless of the test failure. I would advise to try and run this code is a jsbin and see how it works, and after it works as expected, copy it to this file...

HarelM: getData is using messages to communicate with workers, so it you want to use that, you'll need to mock the workers communication since it's running in node environment and not in the browser.
When running things in the browser I expect it to just work, like it does in the examples we have in the docs, which uses this kind of await once code...

AbelVM: Both `bounds` and `testbounds` are being used with their default values because `await map.once()` is not working as expected and... just doesn't wait :sweat_smile: So the flow reaches the assessment but the code to get the bounds has not run

The problem I found with the tests is that running them locally or even building sample code, behaves differently than within the contexts that Jenkins/GH Actions/Whatever uses. I get different errors, messages, and so on.  That's why I'm sending too many commits, to getting actionable information from the online tests that I don't get locally

src/source/geojson_source.ts
```diff
@@ -249,6 +251,48 @@ export class GeoJSONSource extends Evented implements Source {
...
+     *
+     * @returns a promise which resolves to the source's actual boundaries
+     */
+
+    async getBounds(): Promise<LngLatBounds> {
+        type justGeometry = Point | MultiPoint | LineString | MultiLineString | Polygon | MultiPolygon;
+        const bounds: LngLatBounds = new LngLatBounds();
+        const data: GeoJSON.GeoJSON = await this.getData();
+        let coordinates: number[];
+        if (!!data) return bounds;
```
HarelM, line 265: Not sure this is what I would expect as a return value...

test/integration/browser/browser.test.ts
HarelM: This is not resolved. 
Please try and use codespaces if you can't reproduce the issue locally...

test/integration/browser/browser.test.ts
HarelM: I don't see it... Did you forget to push?

AbelVM: :thinking: 

It's pushed https://github.com/AbelVM/maplibre-gl-js/blob/main/test/integration/browser/browser.test.ts

src/source/geojson_source.test.ts
```diff
@@ -12,6 +12,8 @@ import {SubdivisionGranularitySetting} from '../render/subdivision_granularity_s
 import {type ActorMessage, MessageType} from '../util/actor_messages';
 import {type Actor} from '../util/actor';
 import {MercatorTransform} from '../geo/projection/mercator_transform';
+/* import {LngLatBounds} from '../geo/lng_lat_bounds';
```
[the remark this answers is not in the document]
AbelVM: you're right :sweat_smile:

src/source/geojson_source.ts
AbelVM: IMHO, if the source has no data, its bounding box would be empty

test/integration/browser/browser.test.ts
AbelVM: Any help is welcome, I wasn't able to make the integration tests work using `await map.once()`. 

In that very same file, the pattern is `map.once('load', callback)` (even when `load` event is not supported, as per the docs, by `map.once`) https://github.com/maplibre/maplibre-gl-js/blob/6c54ed30a253f65c8ecaf346d14c21fcd1d60fcb/test/integration/browser/browser.test.ts#L53

AbelVM:
Greenlighted!

src/source/geojson_source.ts
```diff
@@ -249,6 +251,48 @@ export class GeoJSONSource extends Evented implements Source {
...
     }
 
+    /**
+     * Allows to get the source's actual boundaries.
+     *
+     * @returns a promise which resolves to the source's actual boundaries
+     */
+
+    async getBounds(): Promise<LngLatBounds> {
+        type justGeometry = Point | MultiPoint | LineString | MultiLineString | Polygon | MultiPolygon;
```
HarelM, line 261: This should be placed before the class definition and should start with a capital letter.

src/source/geojson_source.ts
```diff
@@ -249,6 +251,48 @@ export class GeoJSONSource extends Evented implements Source {
...
+        const bounds: LngLatBounds = new LngLatBounds();
+        const data: GeoJSON.GeoJSON = await this.getData();
```
HarelM, line 263: Not sure the type is needed here as it is inferred from the return type of the method.

src/source/geojson_source.ts
```diff
@@ -249,6 +251,48 @@ export class GeoJSONSource extends Evented implements Source {
...
+                    coordinates = data.geometry.geometries.map((g: justGeometry) => g.coordinates).flat(Infinity) as number[];
+                }else{
+                    coordinates = data.geometry.coordinates.flat(Infinity) as number[];
+                }
+                break;
+            case 'FeatureCollection':
+                coordinates = data.features.map(f => {
+                    if (f.geometry.type === 'GeometryCollection') {
+                        return f.geometry.geometries.map((g: justGeometry) => g.coordinates);
+                    }else{
```
HarelM, line 278: I'm not sure else is needed as you return inside the if...

src/source/geojson_source.ts
```diff
@@ -249,6 +251,48 @@ export class GeoJSONSource extends Evented implements Source {
...
+                        return f.geometry.coordinates;
+                    }
+                }).flat(Infinity) as number[];
```
HarelM, line 281: Why is casting needed at the end of each `flat`?

test/integration/browser/browser.test.ts
HarelM: There's still a lot of coordinates here...

HarelM: image

HarelM:
I've opened the following PR with all the relevant changes I believe:
- https://github.com/maplibre/maplibre-gl-js/pull/5575

Please review it so I can close this one.

test/integration/browser/browser.test.ts
AbelVM: Hmmm, my use cases are always down to 10^-6 and I am reusing the test data that was already there.

HarelM: I've opened a PR to replace this PR, feel free to review it.

src/source/geojson_source.ts
AbelVM: Got a lot of typing errors because of the fuzzy definition of getData results typing, so I had to enforce it

src/source/geojson_source.ts
AbelVM: Geometry Collection needs an extra loop

AbelVM:
What's wrong with this PR to close it?

HarelM:
Some of the comments were not addressed and from the looks of all the cases that were added to the integration tests it seems that the right place to add the tests is unit tests and not integration tests, and I didn't want to burden you with all this effort so I went ahead and opened a new PR based on your code.

AbelVM:
LGTM! 

On the other hand, the testing environment might need more documentation or just a quick guide. V.g.: the way you override the map object with a dummy one solves all the blocking stuff I found with unit tests

HarelM:
I actually only added tests to existing file...
Stubbing object like that is a common practice in unit tests...
If you think the docs can be improved feel free to open a PR, I'm already too involved with this project to know what's not documented well...

AbelVM:
Not stubbing, but the object you used to do so. No event listeners or other indirect approaches like I was trying to implement following the existing tests.

I'd love to help improve the docs, but without knowing that kind of not-exposed APIs it's a bit... blocking :sweat_smile:

HarelM:
I simply copied a different test that used to test getData and changed the logic a bit, nothing fancy...

HarelM:
This was merged and released. Thanks for pushing this through!

PR #5405, opened 2025-01-25
fix: #5404 cjk keep up right bugfix (suggestion) - draft version

### LINKED ISSUE #5404: [Bug] There is an issue with text rotation when using text-keep-upright
**maplibre-gl-js version**:  5.0.1

**browser**: chrome (latest)

### Steps to Trigger Behavior
1. Set `symbol-placement` in `layout` to `line`
2. Set `text-keep-upright` in `layout` to `true`
3. Check the map
4. The text rotates to follow the line, but numbers do not rotate as expected

### Link to Demonstration
https://jsbin.com/hucaripeja/

Image

### Expected Behavior
- `text-keep-upright` should provide a consistent display when `symbol-placement` is set.
- It would be ideal if numbers between CJK characters were oriented in the same direction as the CJK characters.

### Actual Behavior
- The behavior differs depending on whether the text is in a CJK language.

Image

jklee84:
Resolve #5404 

I noticed a long history in this PR. 
While finding a better solution might be more challenging, I think the primary issue lies in mixed cases like `반포대로21길`.

https://github.com/mapbox/mapbox-gl-js/pull/3438

## Before
스크린샷 2025-01-25 오후 10 52 07

## After
스크린샷 2025-01-25 오후 10 51 40

## Passed all unit tests
스크린샷 2025-01-25 오후 10 56 32


test/examples/cjk-keep-up-right.html
```diff
@@ -0,0 +1,159 @@
+<!DOCTYPE html>
```
HarelM, line 1: I think this would be better as a render test instead of an example.

HarelM:
@1ec5 your review here would be greatly appreciated.

jklee84:
> @1ec5 your review here would be greatly appreciated.

@HarelM @1ec5 
My purpose was not to merge the code. If my PR can provide even a small insight, I would be deeply grateful for that. :)

Issue #5420, opened 2025-01-28
[label] Full-width hyphen in Japanese `ー` is accidently disappeared


Kanahiro:
**maplibre-gl-js version**: latest

**browser**: Chrome

### Steps to Trigger Behavior

 1. write style to show label, without "text-font" or with "text-font" not including `ー`
 2. show labels including full-width hyphen in Japanese `ー`
 3. `ー` is not shown.

### Link to Demonstration

Following two are using same style.

#### GL JS v4

https://jsbin.com/wosopasika/edit?html,output

![Image](https://github.com/user-attachments/assets/3994193a-f2c1-4c2b-983d-4edc9a7e8552)

#### GL JS v3

https://jsbin.com/qiduxehete/edit?html,output

![Image](https://github.com/user-attachments/assets/73aa8749-8eab-44e6-aeb5-a53a1d3ae8de)

- You can see `A-BーC` in v3 but in v4 `A-BC`.
- It seems that there might be some regression from v3 to v4.

### Expected Behavior

v3 is correct.

### Actual Behavior

`ー` is missing. It occurs also on latest version.

HarelM:
Thanks for taking the time to report this issue! 
Any chance you could take a look into this issue?

Kanahiro:
Yes, I'm researching this now.👍

Kanahiro:
When `text-font` isn't defined OR defined font doesn't exist in glyphs, `getGlyphs` returns null.
This seems to occur this problem.

https://github.com/maplibre/maplibre-gl-js/blob/bedcb8622337d73875638d165f10b0397f18d775/src/render/glyph_manager.ts#L44

HarelM:
That can be a result of my refactoring of the callback hell, including the glyphs code, which was a nightmare to read and understand...

Kanahiro:
The problem is here:

https://github.com/maplibre/maplibre-gl-js/blob/bedcb8622337d73875638d165f10b0397f18d775/src/render/glyph_manager.ts#L129

This regexp judges whether a character is CJK or not.

Japansese full-width dash `ー` is not included `sc=Hira` or `sc=Kana`. It is included in `sc=Common`.
`sc=Common` includes very broad characters: some Japanese symbols `「` `☆` and some half-width symbols like whitespace.

`sc=Common` includes not only Japanese symbols but also some symbols used globally.
[...]

PR #5421, opened 2025-01-28
Fix/cjk unicode lookup

Kanahiro:
## Launch Checklist

 - [x] Link to related issues.

- Issue: #5420 

## before

- `あーい、う。` becomes `あいう`. some symbols are ignored.

Screenshot 2025-01-29 at 9 38 24

## after

Screenshot 2025-01-29 at 9 36 30


src/render/glyph_manager.ts
```diff
@@ -119,14 +120,16 @@ export class GlyphManager {
...
 
     _doesCharSupportLocalGlyph(id: number): boolean {
-        // The CJK Unified Ideographs blocks and Hangul Syllables blocks are
-        // spread across many glyph PBFs and are typically accessed very
-        // randomly. Preferring local rendering for these blocks reduces
-        // wasteful bandwidth consumption. For visual consistency within CJKV
-        // text, also include any other CJKV or siniform ideograph or hangul,
-        // hiragana, or katakana character.
         return !!this.localIdeographFontFamily &&
-            /\p{Ideo}|\p{sc=Hang}|\p{sc=Hira}|\p{sc=Kana}/u.test(String.fromCodePoint(id));
```
1ec5, line 129: I think we should still include this regular expression as an additional case. `Ideo` covers a lot of sinoxenic writing systems besides Han and kana.

HarelM:
Please add a render test to make sure this behavior is tested and kept in the future. 
Also looks like a render test failed due to the changes made, please are that you take care of it as well.

Kanahiro:
@1ec5 @HarelM 
Thanks for reviewing!
I improved mycodes and added tests. Please check it.

Kanahiro:
`Tests / Render tests (windows-latest, 0) (pull_request) ` faled but rendered image of label could have difference between Ubuntu and Windows. I'm not sure how can I clear this. Please advice me if anyone have knowledge.

Kanahiro:
I can't run `$env:UPDATE=$true; npm run test-render -- text-local-ideographs/cjk-symbols`.
Could anyone run and generate `expected-windows.png`?

HarelM:
You can download the render test html report from the failing CI run, open it in the browser and right click to save the image and check it in. 
I guess this option should be mentioned in the readme as well...

HarelM:
You don't have to create all the image, just those that are different between OSs.
In this case, because it is using system fonts you get different results, this is not true for most cases but we have the relevant infrastructure to solve this.

Kanahiro:
@HarelM Thanks for advice! fixed

src/render/glyph_manager.ts
```diff
@@ -119,14 +120,20 @@ export class GlyphManager {
...
+        (/\p{Ideo}|\p{sc=Hang}|\p{sc=Hira}|\p{sc=Kana}/u.test(String.fromCodePoint(id)) ||
+        (
+            // fallback: RegExp can't cover all cases. refer Issue #5420
+            unicodeBlockLookup['CJK Unified Ideographs'](id) ||
+            unicodeBlockLookup['Hangul Syllables'](id) ||
+            unicodeBlockLookup['Hiragana'](id) ||
+            unicodeBlockLookup['Katakana'](id) || // includes "ー"
+            // memo: these symbols are not all. others could be added if needed.
```
1ec5, line 131: Should we use more of the blocks that are listed for ideographic line breaking behavior?

https://github.com/maplibre/maplibre-gl-js/blob/470eeb7a82bc50179b1c199720a02b04b1bb461a/src/util/script_detection.ts#L76-L89

src/render/glyph_manager.ts
```diff
@@ -119,14 +120,20 @@ export class GlyphManager {
     }
 
     _doesCharSupportLocalGlyph(id: number): boolean {
-        // The CJK Unified Ideographs blocks and Hangul Syllables blocks are
-        // spread across many glyph PBFs and are typically accessed very
-        // randomly. Preferring local rendering for these blocks reduces
-        // wasteful bandwidth consumption. For visual consistency within CJKV
-        // text, also include any other CJKV or siniform ideograph or hangul,
-        // hiragana, or katakana character.
```
1ec5, line 127: I think this comment is still accurate, even with your fix. It explains why this function returns true for, say, halfwidth forms but not, Nushu, even though both have relatively few codepoints.

src/render/glyph_manager.ts
Kanahiro: I'm not familiar with a functionality of codes you mentioned but I guess those are not linked to this pull request. This codes judge whether a character should be loaded from system fonts or not. The codes you refered are not so and listed blocks can be different from this pr pointed.
Are these right?

src/render/glyph_manager.ts
Kanahiro: fixed

src/render/glyph_manager.ts
Kanahiro: Thanks. I think so and other many symbols can be listed to my codes. If a cost to do so is small, I am agree with adding all blocks we need.
However I'm not sure how much is the cost and why developers havn't done so. Then I patched a regression and added blocks I think useful for Japanese users.
If other blocks are needed, it is better new patches are made, IMO.

1ec5 (approved):

src/render/glyph_manager.ts
```diff
@@ -126,7 +127,19 @@ export class GlyphManager {
...
+        (
+            // fallback: RegExp can't cover all cases. refer Issue #5420
+            unicodeBlockLookup['CJK Unified Ideographs'](id) ||
+            unicodeBlockLookup['Hangul Syllables'](id) ||
+            unicodeBlockLookup['Hiragana'](id) ||
+            unicodeBlockLookup['Katakana'](id) || // includes "ー"
+            // memo: these symbols are not all. others could be added if needed.
+            unicodeBlockLookup['CJK Symbols and Punctuation'](id) || // 、。〃〄々〆〇〈〉《》「...
+            unicodeBlockLookup['Halfwidth and Fullwidth Forms'](id) // ！？＂＃＄％＆...
+        )
```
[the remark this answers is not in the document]
Kanahiro: Thx! fixed

HarelM (approved):

Issue #5432, opened 2025-01-30
GeolocateControl marker shrinks when tilting and zooming map


tricaric:
**maplibre-gl-js version**: 5.1.0

**browser**: Chrome/Firefox/Safari on Mac, and Safari/iOS

### Steps to Trigger Behavior

 1. open the "Locate the user" example page: https://maplibre.org/maplibre-gl-js/docs/examples/locate-user/
 2. click on the "find my location" button on top-right of the map, and give permission to locate
 3. wait for the map to zoom on user position, and then zoom more to have the marker clearly visible
 4. alternate zooming and tilting of the map, and you will notice that the marker accuracy disk appears to shrink when you tilt the map, and then go back to its normal size when observing vertically the map

Of course here I am not talking about the expected change in apparent size due to the change of viewing angle, but to the sudden shrinking and growing of the marker after tilting and zooming, as it is clear when compared to the nearby buildings.

### Link to Demonstration

https://github.com/user-attachments/assets/3aca9e2f-58f9-42ab-b9f5-ba6448f13854

### Expected Behavior

The marker should not shrink when tilting the map, it should stay fixed at the size given by the accuracy of the position of the user.

### Actual Behavior

The marker appears to shrink when tilting the map, and then appears to grow back to its normal size when observing the map vertically.

HarelM:
Thanks for taking the time to report this issue. 
Seems like the style of the accuracy circle should be set to viewport and not map.
PR is welcomed.

geekdenz:
I think we can close this as I cannot reproduce this issue. If you like I can make a video confirming this.

HarelM:
Yes, please do.

HarelM:
A short test I made seems like this is still an issue in the docs site and also locally in the latest version of main.

PR #5437, opened 2025-01-31
improve GeolocateControl accuracy circle radius computation

tricaric:
Now the GeolocateControl marker is using the center position to calculate the pixel scale, and is updating calculations at each move/rotate/zoom of the map, as those change the position of the accuracy circle, and this affects its pixel size depending on map pitch. This also improves the case of moving the marker "far away" in tilted map (final part of videos)

This should improve/fix https://github.com/maplibre/maplibre-gl-js/issues/5432 

Videos show the GeolocateControl before and after the fix.

https://github.com/user-attachments/assets/6e8a56ad-2a22-4dbe-964f-8a2b2063f21f

https://github.com/user-attachments/assets/d09f419c-8945-4080-bd02-630a3724b4b7

## Launch Checklist

 - [x] Link to related issues.

### DIFF
--- a/src/ui/control/geolocate_control.ts
+++ b/src/ui/control/geolocate_control.ts
@@ -303,6 +303,8 @@ export class GeolocateControl extends Evented implements IControl {
         }
 
         DOM.remove(this._container);
+        this._map.off('move', this._onMove);
+        this._map.off('rotate', this._onRotate);
         this._map.off('zoom', this._onZoom);
         this._map = undefined;
         numberOfWatches = 0;
@@ -374,6 +376,11 @@ export class GeolocateControl extends Evented implements IControl {
             this._finish();
 
             return;
+        }
+
+        if (this.options.showAccuracyCircle) {
+            // this is needed by _updateCircleRadius()
+            this._lastKnownPosition = position;
         }
 
         if (this.options.trackUserLocation) {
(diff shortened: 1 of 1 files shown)


HarelM:
This looks a lot better, thanks! 
Do you think the animation of the blue circle should also have a tilt like the accuracy circle?

tricaric:
> Do you think the animation of the blue circle should also have a tilt like the accuracy circle?

I think that the central pulsating dot is good as it is, no need to tilt it.

HarelM:
But it looks weird when tilting the map, doesn't it?

tricaric:
> But it looks weird when tilting the map, doesn't it?

Yes in tilted maps, or in the globe when near the edge of the globe, it may appear as it's a bit off compared to the accuracy circle. I think that this could be a option of GeolocateControl, to have it as it is now, or to flatten it on the map like the accuracy circle.

src/ui/control/geolocate_control.ts
```diff
@@ -460,16 +467,30 @@ export class GeolocateControl extends Evented implements IControl {
...
-        const bounds = this._map.getBounds();
-        const southEastPoint = bounds.getSouthEast();
-        const northEastPoint = bounds.getNorthEast();
-        const mapHeightInMeters = southEastPoint.distanceTo(northEastPoint);
-        const mapHeightInPixels = this._map._container.clientHeight;
-        const circleDiameter = Math.ceil(2 * (this._accuracy / (mapHeightInMeters / mapHeightInPixels)));
+        if (!this._lastKnownPosition) return;
+        const center = new LngLat(this._lastKnownPosition.coords.longitude, this._lastKnownPosition.coords.latitude);
+        const point = this._map.project(center);
+        const delta_x_pixels = 10; // arbitrary small offset, in x direction because there is no map tilting in the x direction
```
geekdenz, line 473: Should this constant be declared at the beginning of the file? Maybe it could be a bit more descriptive.
Should it be capitalized?

How about `const PIXEL_SCALE_DENOMINATOR_X_DELTA = 10`?

src/ui/control/geolocate_control.ts
```diff
@@ -303,6 +303,8 @@ export class GeolocateControl extends Evented implements IControl {
         }
 
         DOM.remove(this._container);
+        this._map.off('move', this._onMove);
+        this._map.off('rotate', this._onRotate);
```
geekdenz, line 307: Good that these are cleaned up.

geekdenz (commented):
Good work! This seems to solve the issue. I tested it briefly.

geekdenz:
How about the checks not being successful? Can I see them succeed on another PR or elsewhere?

HarelM:
main is always green, every PR that is merged is done only if all the tests are passing. 
You can try and open a PR with a simple change to see that it passes.

geekdenz:
> main is always green, every PR that is merged is done only if all the tests are passing. 
> You can try and open a PR with a simple change to see that it passes.

Thanks for the clarification.

HarelM:
Any updates on this? I have 3 PRs on the same issue and I'm not sure which one of them to merge if at all...

geekdenz:
I haven't forgotten about this but didn't have headspace and time to progress mine. However, I will probably have time this week to have a look.

Do we want to go the GeoJSON route or would you want an alternative approach that takes less code? That would be another PR though.

HarelM:
I think this one:
- #5543

Is the most promising, but it needs some work to reduce bundle size I believe...

HarelM:
@tricaric @geekdenz can we consolidate on a single PR to solve this issue?
If this is good enough and doesn't increase bundle size by much we can use this PR and later on decide if we want to improve it with a geojson source etc...
What do you think?

geekdenz:
I can work on solving the outstanding soon on whichever PR you think is best.

HarelM:
Let's start with this one, we can alway improve it in the future.

geekdenz:
OK. Thanks @HarelM .

To confirm: Do you want me to checkout https://github.com/tricaric/maplibre-gl-js/tree/fix-GeolocateControl-accuracy-circle and make a commit on top of it on a branch in my repo and do another PR?

HarelM:
Yes. I'll close these two PRs so that I won't get confused.

PR #5440, opened 2025-02-01
Fix GeolocateControl accuracy circle using turf.circle and GeoJSON

tricaric:
This PR replaces PR https://github.com/maplibre/maplibre-gl-js/pull/5437 and improves/fixes https://github.com/maplibre/maplibre-gl-js/issues/5432 in many more edge cases.

The original problem, still present in https://github.com/maplibre/maplibre-gl-js/pull/5437 was the use of a _circleElement. When there is significant map tilting, or when using the globe and approaching the edge of the map, there is no reasonable way to properly scale the element and make it work. The pixel scale calculated only at the center of the marker is not sufficient, and the element itself is too rigid to follow the curvature of the globe. The three videos below are mainline, PR  https://github.com/maplibre/maplibre-gl-js/pull/5437, and the current PR.

https://github.com/user-attachments/assets/bd238a2d-4661-4afa-bb9d-65d46499437f

https://github.com/user-attachments/assets/b0ba2b05-8074-4e71-ba43-e5f6b46cd01c

https://github.com/user-attachments/assets/1ee654b0-72a8-4c6a-a9b7-c2936d4609e4

With this PR, the _circleElement is replaced with a layer with dynamic GeoJSONSource, with data generated using a single turf/circle call at each position update. This removes all calls to _updateCircleRadius() at each user interaction, which may improve performance. The resulting accuracy circle seems to behave exactly as it should.

I had to comment out the tests on _circleElement because it's gone. Perhaps the tests can be recovered/updated somehow.

The code to add/remove the GeoJSONSource and the layer probably needs more love, it works as is but if one clicks too quickly on the GeolocateControl button, with trackUserLocation=true, the appearance/disappearance of the marker and circle is sometimes broken.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/ui/control/geolocate_control.test.ts
+++ b/src/ui/control/geolocate_control.test.ts
@@ -494,6 +494,7 @@ describe('GeolocateControl with no options', () => {
         expect(geolocate._watchState).toBe('BACKGROUND');
     });
 
+    /*
     test('accuracy circle not shown if showAccuracyCircle = false', async () => {
         const geolocate = new GeolocateControl({
             trackUserLocation: true,
@@ -516,7 +517,9 @@ describe('GeolocateControl with no options', () => {
         await zoomendPromise;
         expect(!geolocate._circleElement.style.width).toBeTruthy();
     });
+    */
 
+    /*
     test('accuracy circle radius matches reported accuracy', async () => {
         const geolocate = new GeolocateControl({
             trackUserLocation: true,
@@ -557,7 +560,9 @@ describe('GeolocateControl with no options', () => {
         await zoomendPromise;
         expect(geolocate._circleElement.style.width).toBe('4996px');
     });
+    */
 
+    /*
     test('shown even if trackUserLocation = false', async () => {
         const geolocate = new GeolocateControl({
             trackUserLocation: false,
@@ -580,7 +585,9 @@ describe('GeolocateControl with no options', () => {
         await zoomendPromise;
         expect(geolocate._circleElement.style.width).toBeTruthy();
     });
+    */
 
+    /*
     test('shown even if trackUserLocation = false', async () => {
         const geolocate = new GeolocateControl({
             trackUserLocation: false,
@@ -603,6 +610,7 @@ describe('GeolocateControl with no options', () => {
         await zoomendPromise;
         expect(geolocate._circleElement.style.width).toBeTruthy();
     });
+    */
 
     test('Geolocate control should appear only once', async () => {
         const geolocateControl = new GeolocateControl({});
--- a/src/ui/control/geolocate_control.ts
+++ b/src/ui/control/geolocate_control.ts
@@ -10,6 +10,9 @@ import type {FitBoundsOptions} from '../camera';
 import type {IControl} from './control';
 import {LngLatBounds} from '../../geo/lng_lat_bounds';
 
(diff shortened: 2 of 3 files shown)


HarelM:
Can you please close the other PR then?

src/ui/control/geolocate_control.test.ts
```diff
@@ -494,6 +494,7 @@ describe('GeolocateControl with no options', () => {
         expect(geolocate._watchState).toBe('BACKGROUND');
     });
 
+    /*
```
HarelM, line 497: Please don't keep commented out tests. Either bring them back or completely remove them.

src/ui/control/geolocate_control.ts
```diff
@@ -10,6 +10,9 @@ import type {FitBoundsOptions} from '../camera';
 import type {IControl} from './control';
 import {LngLatBounds} from '../../geo/lng_lat_bounds';
 
+import * as turf from '@turf/circle';
```
HarelM, line 13: Can we use a "specific import" instead of "*"?

src/ui/control/geolocate_control.ts
```diff
@@ -447,32 +452,46 @@ export class GeolocateControl extends Evented implements IControl {
...
-            this._accuracyCircleMarker.setLngLat(center).addTo(this._map);
             this._userLocationDotMarker.setLngLat(center).addTo(this._map);
             this._accuracy = position.coords.accuracy;
             if (this.options.showUserLocation && this.options.showAccuracyCircle) {
-                this._updateCircleRadius();
+
+                this._accuracyCirclePolygon = turf.circle([position.coords.longitude, position.coords.latitude],  position.coords.accuracy, {steps: 64,
+                    units: 'meters'});
+
+                if (this._map.getSource('accuracy-circle')) {
```
HarelM, line 462: What will happen if a style has a source with this name?

src/ui/control/geolocate_control.ts
```diff
@@ -447,32 +452,46 @@ export class GeolocateControl extends Evented implements IControl {
...
-        const mapHeightInPixels = this._map._container.clientHeight;
-        const circleDiameter = Math.ceil(2 * (this._accuracy / (mapHeightInMeters / mapHeightInPixels)));
-        this._circleElement.style.width = `${circleDiameter}px`;
-        this._circleElement.style.height = `${circleDiameter}px`;
-    }
-
-    _onZoom = () => {
-        if (this.options.showUserLocation && this.options.showAccuracyCircle) {
-            this._updateCircleRadius();
+            if (this._map.loaded() && this._map.isStyleLoaded()) {
```
HarelM, line 491: If the style is still being loaded this might "leak"...

HarelM:
While I agree this approach is the correct way to handle this (I'm doing the same thing in my app), I think it might be problematic for some cases when switching styles and it might collide with sources or layers defined by users.
Maybe a guid should be used to avoid layers or sources collision.
Also the added package seems a bit big in terms of solving this.
I would consider maybe creating a plugin with these changes and see if people are using it enough...

geekdenz:
I believe this PR is stale. Unless someone stops me, I'll have a look to get the existing commits and address the raised issues.

PR #5441, opened 2025-02-01
Avoid setting marker opacity twice

erasta:
When creating a marker setOpacity was called twice, which results in calling _updateOpacity twice.
Also the opacity params are check twice for undefined.
This minor optimization solves the above.  

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


src/ui/marker.ts
```diff
@@ -169,7 +169,9 @@ export class Marker extends Evented {
         this._rotationAlignment = options && options.rotationAlignment || 'auto';
         this._pitchAlignment = options && options.pitchAlignment && options.pitchAlignment !== 'auto' ?  options.pitchAlignment : this._rotationAlignment;
         this.setOpacity(); // set default opacity
-        this.setOpacity(options?.opacity, options?.opacityWhenCovered);
+        if (options?.opacity !== undefined || options?.opacityWhenCovered !== undefined) {
```
HarelM, line 172: Why is this if needed? Why not simply pass the options' `opacity` and `opacityWhenCovered` as the method checks for undefined etc.

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ### ✨ Features and improvements
 
-- _...Add new stuff here..._
+- Avoid setting marker opacity twice.
```
HarelM, line 5: Please add a PR link like the other entries.

HarelM:
Thanks for taking the time to fix this.

src/ui/marker.ts
erasta: Thank you, this is because `setOpacity` invokes `_updateOpacity`, which in turn has some performance implications. This happens disregarding the values of `opacity` and `opacityWhenCovered`, even if those are both `undefined`.
https://github.com/maplibre/maplibre-gl-js/blob/81331f3ff8ebb33afc4b72d7e34d810f8eeb1508/src/ui/marker.ts#L874

Moreover, if one was `undefined` and the other wasn't, we still need to call the function.

CHANGELOG.md
erasta: Done

src/ui/marker.ts
HarelM: After your changes the call to seOpacity can still happen twice in case the options object has the relevant values. 
My question is why not simply call this method one time with the options object values and remove the if altogether?

erasta: Correct, that would be more elegant. However we now need to reset the members beforehand on constructor, so i changed the `if` accordingly.

src/ui/marker.ts
```diff
@@ -857,16 +856,19 @@ export class Marker extends Evented {
      * @param opacityWhenCovered - Sets the `opacityWhenCovered` property of the marker.
      */
     setOpacity(opacity?: string, opacityWhenCovered?: string): this {
-        if (opacity === undefined && opacityWhenCovered === undefined) {
+        // Reset opacity when called without params or from constructor
+        if (this._opacity === undefined || (opacity === undefined && opacityWhenCovered === undefined)) {
```
HarelM, line 860: I think this method can be simplified even further:
```ts
this._opacity = opacity !== undefined ? opacity : '1';
this._opacityWhenCovered = opacityWhenCovered!== undefined ? opacityWhenCovered : '0.2';
```

erasta: In the case of calling `marker.setOpacity(something);`:
- the expected result is that `this._opacityWhenCovered` will not be changed
- in the suggested code it will reset it to the default `'0.2'`

same goes for calling `marker.setOpacity(undefined, something);`

I agree that a oneliner shortening is possible, but it will be somewhat less-understandable.

HarelM (approved):

Issue #5327, opened 2025-01-10
Popup on hover example is not changing popup when mouse moves to another overlapping feature


matthiasschwarz:
**maplibre-gl-js version**: 5.0.0

**browser**: Chromium 131.0.6778.265

### Steps to Trigger Behavior

 1. Adjust zoom to have atleast two overlapping features
 2. Move mouse to a feature to show popup
 3. Move mouse to another feature without leaving any feature

### Link to Demonstration

[Popup on hover example](https://maplibre.org/maplibre-gl-js/docs/examples/popup-on-hover/)

### Expected Behavior

Popup changes to the topmost shown feature.

### Actual Behavior

https://github.com/user-attachments/assets/30ceb26b-36b6-4654-99a6-1c434209374e

Popup doesn't change when moving to another overlapping feature. This is caused by the mouseenter event not triggered again because we didn't leave the layer while moving to the other feature.

HarelM:
You are welcome to open a PR to improve the example.
Keep in mind that we do want to keep the examples short and precise, as they are not intended to be bug free, but more of an example of how something can be achieved.

matths:
I recently had exactly the same problem popping up when working on my use case. Here's a JSBin with the workaround to use the mousemove instead of the mouseenter event.

https://jsbin.com/gibipih/edit?html,output

I can make this available as PR, too.

HarelM:
> I can make this available as PR, too.

That would be great, thanks!

johncarmack1984:
The example switched to mousemove with feature tracking in #5442 (v5.1.1). This can be closed.

PR #5442, opened 2025-02-01
Fix example not changing popup when mouse moves to another overlapping feature

matths:
This PR fixes this issue https://github.com/maplibre/maplibre-gl-js/issues/5327
The documentation example Display a popup on hover.
The Popup doesn't change when moving to another overlapping feature. This is caused by the mouseenter event not triggered again because we didn't leave the layer while moving to the other feature.
As a fix we are using mousemove instead of mouseenter event and check if the mouse is over another feature and the popup content needs to change.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/test/examples/popup-on-hover.html
+++ b/test/examples/popup-on-hover.html
@@ -162,28 +162,26 @@
         // and use mousemove instead of mouseenter event
         let currentFeatureCoordinates = undefined;
         map.on('mousemove', 'places', (e) => {
-            if (e.features && e.features.length > 0) {
-                const featureCoordinates = e.features[0].geometry.coordinates.toString();
-                if (currentFeatureCoordinates !== featureCoordinates) {
-                    currentFeatureCoordinates = featureCoordinates;
+            const featureCoordinates = e.features[0].geometry.coordinates.toString();
+            if (currentFeatureCoordinates !== featureCoordinates) {
+                currentFeatureCoordinates = featureCoordinates;
 
-                    // Change the cursor style as a UI indicator.
-                    map.getCanvas().style.cursor = 'pointer';
+                // Change the cursor style as a UI indicator.
+                map.getCanvas().style.cursor = 'pointer';
 
-                    const coordinates = e.features[0].geometry.coordinates.slice();
-                    const description = e.features[0].properties.description;
+                const coordinates = e.features[0].geometry.coordinates.slice();
+                const description = e.features[0].properties.description;
 
-                    // Ensure that if the map is zoomed out such that multiple
-                    // copies of the feature are visible, the popup appears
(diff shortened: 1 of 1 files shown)


test/examples/popup-on-hover.html
```diff
@@ -158,30 +158,41 @@
...
         });
 
-        map.on('mouseenter', 'places', (e) => {
-            // Change the cursor style as a UI indicator.
-            map.getCanvas().style.cursor = 'pointer';
+        // Make sure to detect marker change for overlapping markers
+        // and use mousemove instead of mouseenter event
+        let currentFeatureCoordinates = undefined;
+        map.on('mousemove', 'places', (e) => {
+            if (e.features && e.features.length > 0) {
```
HarelM, line 165: Can this if be negated to improve readability?
Also the below one, so that the method would be less indented.

test/examples/popup-on-hover.html
```diff
@@ -158,30 +158,41 @@
...
-            popup.setLngLat(coordinates).setHTML(description).addTo(map);
+                    // Populate the popup and set its coordinates
+                    // based on the feature found.
+                    popup.setLngLat(coordinates).setHTML(description).addTo(map);
+                }
+            }
         });
 
         map.on('mouseleave', 'places', () => {
+            currentFeatureCoordinates = undefined;
```
HarelM, line 191: I would consider removing this event handler and moving this logic into the `mousemove` event in the case of "no features" (since there's an if there checking that I'm guessing it's a valid code flow).

HarelM:
Thanks for taking the time to fix this example!

test/examples/popup-on-hover.html
matths: I thought so as well, but this doesn't work out, because the `map.on('mousemove', 'places', () => {}` is not being run through when the mouse is moved outside a places feature, you know. ;) So we do not run in such case.
As mentioned before, I fought the exact same situation just a couple of hours/days ago for a task at my job.

HarelM: Then doesn't it make the `if (e.features && e.features.length > 0)` in the mousemove event redundant?

test/examples/popup-on-hover.html
matths: I think it's safe to remove the if condition altogether, as we only get a mousemove event in the places layer when we are over a feature of that layer.

test/examples/popup-on-hover.html
matths: I did.

HarelM (approved):
THANKS!

PR #5518, opened 2025-02-18
Fix 5486

### LINKED ISSUE #5486: rendering incomplete when using setPadding()
**maplibre-gl-js version**:

**browser**:

### Steps to Trigger Behavior

 1.
 2.
 3.

### Link to Demonstration
https://test.twin.search.ch/pub/maplibre_setpadding_5.1.0.html

NathanMOlson:
Fixes #5486.

Do not allow variable zoom tiles just because the top padding is non-zero. Only allow variable zoom at high pitch or when terrain is enabled.

 - [x] Link to related issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - _...Add new stuff here..._
 - Fix how padding is applied when using flyTo() with Globe ([#5406](https://github.com/maplibre/maplibre-gl-js/pull/5406))
 - Fix URL hash validation to support bearing range -180 to 180 ([#5461](https://github.com/maplibre/maplibre-gl-js/issues/5461))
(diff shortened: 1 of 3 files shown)


HarelM:
Thanks! Do you think a render test can help here?

NathanMOlson:
@HarelM No. The bug was in the way covering tiles were calculated when the top padding was non-zero. The new unit test covers that. The rendering problem shown in the original bug report was because the source did not contain the lower-zoom tiles that were being requested.

Current MapLibre behavior is not to request higher zoom tiles when the requested zoom is not available. This seems like a reasonable default to prevent an explosion of tile requests, but it may be worth looking into a way to change that behavior (outside the scope of this PR).

HarelM:
In a render test we can have the numbers source and make sure all the tiles are from the same zoom, which is what this fix is about as far as I understand.
Before this fix there's a way to receive different zoom level, right?
This is just a thought, I'm fine with merging this as is too.

PR #5532, opened 2025-02-21
Allow setting opacity of popup when behind surface (#5079)

### LINKED ISSUE #5079: Allow setting opacity of popup when behind surface
**maplibre-gl-js version**: 5.0.0-pre.6

**browser**: Google chrome Version 130.0.6723.117 (Official Build) (arm64)

### Steps to Trigger Behaviour

 1. Add popup
 2. Rotate globe so popup is behind sphere

### Link to Demonstration

https://jsbin.com/xihuyikuxo/3/edit?html,css,output

### Expected Behaviour

Popup is hidden when its behind the globe (?) Maybe it's a conscious decision that it shouldn't be hidden, but maybe adding some boolean value to popup config indicating if popup should respect globe (should hide) would be nice addition?

geekdenz:
* Resolves #5079
* add locationOccludedOpacity: number property as opacity
* listen on the map if this is set
* change opacity when behind the surface
* add to changelog

Before:

https://github.com/user-attachments/assets/f248d8d8-e3a4-4feb-b623-c1cec786c234

After:

https://github.com/user-attachments/assets/6c29bcf8-7f4e-4a14-8d82-03cc0bbe238e

## Launch Checklist

 - [X] Link to related issues.


HarelM:
I'm not sure the example is interesting enough to be added, to be honest...

kamil-sienkiewicz-asi:
Changing opacity to 0 will hide the popup, but it still will receive mouse events, which potentially could block movement of the map when user will click on it, right?

geekdenz:
> I'm not sure the example is interesting enough to be added, to be honest...

You're right. Should I change the existing example or just leave it?

HarelM:
Leave it. You can asd an example to the popup tsdocs if that makes sense, although I'm not sure it does...

geekdenz:
I think leave it is the best option. Too small a thing to make biggish changes.

geekdenz:
@HarelM 

Does this need integration or unit tests?

If not, it is ready for review.

HarelM:
Yes, a unit test would be enough I believe.

geekdenz:
I wonder if the default should be to hide the popup when it is occluded. That would be an easy change here now. What do you think @HarelM ?

HarelM:
Not sure, I think it might be considered a breaking change, so it would need a major version release, which is a bit too much for this feature.

geekdenz:
OK. Makes sense. It is ready for review then I believe.

src/ui/popup.test.ts
```diff
@@ -825,4 +825,20 @@ describe('popup', () => {
...
     });
+    test('Popup changes opacity when location behind globe', () => {
+        const map = createMap();
+
+        const popup = new Popup({locationOccludedOpacity: 0.2})
+            .setLngLat([0, 0])
+            .setText('Test')
+            .addTo(map);
+
+        map.on('load', () => {
```
HarelM, line 836: This is an async call. 
Please use "once" and await to make sure this part of the test is executed.

src/ui/popup.ts
```diff
@@ -88,6 +88,11 @@ export type PopupOptions = {
      * @defaultValue false
      */
     subpixelPositioning?: boolean;
+    /**
+     * Optional opacity when the location is behind the globe.
+     * If not supplied default to 1. Note the number is converted to a string.
```
HarelM, line 93: Why not allow defining a string then?

src/ui/popup.ts
```diff
@@ -88,6 +88,11 @@ export type PopupOptions = {
...
+     */
+    locationOccludedOpacity?: number;
```
HarelM, line 95: Please add a default value to the object above that defines the defaults.

src/ui/popup.ts
HarelM: Also use the `@defaultValue` like other options.

src/ui/popup.ts
```diff
@@ -226,6 +231,14 @@ export class Popup extends Evented {
         return this;
     }
 
+    _updateOpacity = () => {
+        if (this._map.transform.isLocationOccluded(this.getLngLat())) {
+            this._container.style.opacity = `${this.options.locationOccludedOpacity}`;
+        } else {
```
HarelM, line 237: Can this accidentally change the opacity in case someone sets a custom html using ``setHtml`?

src/ui/popup.ts
```diff
@@ -646,6 +659,11 @@ export class Popup extends Evented {
 
         DOM.setTransform(this._container, `${anchorTranslate[anchor]} translate(${offsetedPos.x}px,${offsetedPos.y}px)`);
         applyAnchorClass(this._container, anchor, 'popup');
+
+        // add opacity to popup if in globe projection and location is behind view
```
HarelM, line 663: Move this comment to be tsdoc style above the newly added method instead.

src/ui/popup.ts
geekdenz: Since we are defining `@defaultValue`, should we make it 0 to hide the popup when behind by default? That seems like it should be expected behaviour.

HarelM: Wouldn't this change the current behavior for people not setting this value and create breaking change?

geekdenz: Yes that's why I'm asking. But this issue was created in the expectation that the opacity should be lower behind the globe. Is globe a fairly recent change? 

Also I had a colleague point out that they would expect it to be zero behind by default. 

We can default it to 1 and get rid of an if statement also.

HarelM: Default should keep the current behavior, we can consider changing that in next major version release.

HarelM: Yes, globe was introduced early this year.

src/ui/popup.ts
geekdenz: No, I tried it:
image

HarelM: What happens if this html has opacity defined that is different than 1 or 0.2, wouldn't it get overwritten?

geekdenz: Yes, good catch. That makes it tricky.  Quite the corner case as well. 

So, I guess you are talking about this sort of scenario:

```typescript
popup.getElement().style.opacity = 0.7 // after popup initialized
```

HarelM: Yes, but also just sending an HTML that has different opacity than 1.
It is OK if the code changes the opacity if the relevant configuration is provided, but it shouldn't do anything if this configuration is not provided.

geekdenz: So, it is good enough of it changes opacity to "1" after it has been changed to something else like 0.2 when the property is sweet but not run the code when the custom opacity isn't set?

HarelM: Can we use undefined instead if an empty string?

src/ui/popup.ts
```diff
@@ -88,6 +89,12 @@ export type PopupOptions = {
      * @defaultValue false
      */
     subpixelPositioning?: boolean;
+    /**
+     * Optional opacity when the location is behind the globe.
+     * If not supplied default to 1. Note the number is converted to a string.
+     * @defaultValue 1
```
HarelM, line 95: Default value isn't 1, right?

CHANGELOG.md
```diff
@@ -1,11 +1,21 @@
...
+### 🐞 Bug fixes
+
+- _...Add new stuff here..._
+
 ### ✨ Features and improvements
 
 - _...Add new stuff here..._
 
-### 🐞 Bug fixes
+### ✨ Features and improvements
```
HarelM, line 11: This is probably a bad merge...

src/ui/popup.test.ts
```diff
@@ -825,4 +825,20 @@ describe('popup', () => {
...
     });
+    test('Popup changes opacity when location behind globe', async () => {
+        const map = createMap();
+
+        const popup = new Popup({locationOccludedOpacity: 0.2})
+            .setLngLat([0, 0])
+            .setText('Test')
+            .addTo(map);
+
+        await map.once('load', () => {
```
HarelM, line 836: Don't use await and lambda, just await and continue the rest of the method "regularly".

src/ui/popup.ts
```diff
@@ -226,6 +233,20 @@ export class Popup extends Evented {
...
+     * Add opacity to popup if in globe projection and location is behind view
+     */
+    _updateOpacity = () => {
+        if (this.options.locationOccludedOpacity === undefined) {
+            return;
+        }
+        if (this._map.transform.isLocationOccluded(this.getLngLat())) {
+            this._container.style.opacity = `${this.options.locationOccludedOpacity}`;
+        } else {
+            this._container.style.opacity = '';
```
HarelM, line 246: How about setting this to undefined instead of empty string?

CHANGELOG.md
HarelM: There are two titles if "✨ Features and improvements", can you please fix this?

src/ui/popup.ts
```diff
@@ -88,6 +89,12 @@ export type PopupOptions = {
      * @defaultValue false
      */
     subpixelPositioning?: boolean;
+    /**
+     * Optional opacity when the location is behind the globe.
+     * If not supplied default to 1. Note the number is converted to a string.
```
HarelM, line 94:
```suggestion
     * Note that if a number is provided, it will be converted to a string.
```

HarelM (approved):

PR #5543, opened 2025-02-24
Fix accuracy circle turf geojson

geekdenz:
This PR takes up some of the work that was already done here: 
https://github.com/maplibre/maplibre-gl-js/pull/5440

The goal not finalised yet, is to get to a mergable PR.

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


geekdenz:
@HarelM Is it reasonable to increase the bundle size with the small turfjs library?

expected 912743 to be less than 910819
It's about 2kb.

Later we may want to introduce parts of proj4js as well. I am wondering if it is a good way forward to accept small parts of other libraries in the build.

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Avoid setting marker opacity twice. ([#5441](https://github.com/maplibre/maplibre-gl-js/pull/5441))
 
 ### 🐞 Bug fixes
+- Fix location accuracy circle radius when moving view ([#5543](https://github.com/maplibre/maplibre-gl-js/pull/5543))
```
HarelM, line 18: This should not be here, but upwards instead...

package.json
```diff
@@ -55,6 +55,8 @@
     "@rollup/plugin-terser": "^0.4.4",
     "@rollup/plugin-typescript": "^12.1.2",
     "@stylistic/eslint-plugin-ts": "^4.0.1",
+    "@turf/circle": "^7.2.0",
```
HarelM, line 58: Shouldn't this be under dependencies instead of here?

src/ui/control/geolocate_control.test.ts
```diff
@@ -610,6 +588,11 @@ describe('GeolocateControl with no options', () => {
         map.addControl(geolocateControl);
         // adding and removing to verify there is no race condition, and it is just added once
         map.removeControl(geolocateControl);
+        await map.once('loaded', async () => {
```
HarelM, line 591: use:
```ts
await map.once('loaded');
const accuracyElementsOnPage = ...
```

src/ui/control/geolocate_control.ts
```diff
@@ -265,7 +267,7 @@ export class GeolocateControl extends Evented implements IControl {
     _watchState: 'OFF' | 'ACTIVE_LOCK' | 'WAITING_ACTIVE' | 'ACTIVE_ERROR' | 'BACKGROUND' | 'BACKGROUND_ERROR';
     _lastKnownPosition: any;
     _userLocationDotMarker: Marker;
-    _accuracyCircleMarker: Marker;
+    _accuracyCirclePolygon: any;
```
HarelM, line 270: Why `any`?

test/build/min.test.ts
```diff
@@ -38,7 +38,7 @@ describe('test min build', () => {
         const decreaseQuota = 4096;
 
         // feel free to update this value after you've checked that it has changed on purpose :-)
-        const expectedBytes = 909795;
+        const expectedBytes = 912743;
```
HarelM, line 41: I tend to think this is an overkill for such a small fix.
While the package size does grow over time, I'm not sure this component (which can be easily replaced by any app that requires better control over the UX) should cause this kind of change.
Can we write something simpler that doesn't require all the complexity of turf circle?
We did the same for simple ruler for example, where we took only the relevant functionality to reduce bundle increase...

package.json
geekdenz: I was thinking the same but it works and you cannot access turf outside of the build. But I will change it anyway. Maybe something to think about.

test/build/min.test.ts
geekdenz: Isn' t this the way this fix was intended though. I am not sure if it should have `turf` in the title if it is not a fix using turf.

Should there maybe be an entirely new PR then?

Also, I don't know how to easily achieve a circle without turf. But I can look into it if that is what is best and desired now. Please let me know what you decide. I would recommend we accept the small increase but I wouldn't insist if bundle size is this important.

HarelM:
@geekdenz any updates on this?

geekdenz:
> @geekdenz any updates on this?

@HarelM Do we want to use the GeoJSON way of this like here? I can work on the unit tests for the new code, but am reluctant to if the size is going to fail it.

HarelM:
If we use minimal code to achieve this, the size shouldn't change dramatically...
I'm not sure turf falls under the "minimal code" category...

geekdenz:
Well then this PR needs a rework. Because I took the minimal code out of turf and created a minimal library (functions) but the code is still too big. I'd need to use a different approach than GeoJSON.

HarelM:
This still uses a lot of turf code, which I'm not sure it's easy to tree shake, so I'm not sure it's minimal...

geekdenz:
Hang on @HarelM .

Do you want me to improve this PR i.e. my (@geekdenz) PR or checkout @tricaric 's code and improve on that?

HarelM:
As far as I understand and remember, this PR has an issue with the bundle size.
I think trying to solve that would be harder than finalizing @tricaric 's code.
It's your choice, but I think we can start with adding @tricaric 's code and then decide if trying to use a geojson source here is possible while still keeping the bundle size small...

geekdenz:
@HarelM 

I've started here: https://github.com/maplibre/maplibre-gl-js/pull/6190

It is still in draft, but I thought I'd let you know in case you didn't see it.

HarelM:
Thanks for the info! Closing this for now.

PR #5564, opened 2025-02-28
Cleanup popup tip on popup removal

kamil-sienkiewicz-asi:
Hey, I've noticed that when I call popup.remove(), there is still additional DOM node hanging and never removed which is a maplibregl-popup-tip. In this PR I'm cleaning this up.

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/ui/popup.test.ts
+++ b/src/ui/popup.test.ts
@@ -498,23 +498,6 @@ describe('popup', () => {
         expect(map.getContainer().querySelectorAll('.maplibregl-popup')).toHaveLength(1);
     });
 
-    test('Popup remove() clears _closeButton event listener', () => {
-        const map = createMap();
-
-        const popup = new Popup()
-            .setText('Test')
-            .setLngLat([0, 0])
-            .addTo(map);
-
-        expect(popup._closeButton).toBeDefined();
-
-        const removeEventListenerSpy = vi.spyOn(popup._closeButton, 'removeEventListener');
-
-        popup.remove();
-
-        expect(removeEventListenerSpy).toHaveBeenCalledWith('click', expect.any(Function));
-    });
-
     test('Popup#addTo is idempotent (#1811)', () => {
         const map = createMap();
 
--- a/src/ui/popup.test.ts
+++ b/src/ui/popup.test.ts
@@ -842,4 +842,19 @@ describe('popup', () => {
 
         expect(popup.getElement().style.transform).toBe('translate(-50%,-100%) translate(0px,1px)');
     });
+    test('Popup changes opacity when location behind globe', async () => {
+        const map = createMap();
+
+        const popup = new Popup({locationOccludedOpacity: 0.2})
+            .setLngLat([0, 0])
+            .setText('Test')
+            .addTo(map);
+
+        await map.once('load');
+        map.setProjection({
+            type: 'globe'
+        });
+        map.setCenter([180, 0]);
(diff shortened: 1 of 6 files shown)


HarelM:
Looks like tip is created as a child element of the container, isn't it? It should be removed when container is removed...?

kamil-sienkiewicz-asi:
@HarelM Well, yes, it's removed from DOM, but it's not removed from memory.
https://jsbin.com/dujimegufa/edit?html,css,output

There are two monkey-patches that are tracking event listeners and DOM nodes.

image

image

After my changes, Popup class is correctly removed from memory

HarelM:
I don't see `_tip` is used anywhere, can't we simply remove it (i.e. remove the variable and not store it).
Aslo, wouldn't `_tip = null` be enough to allow GC to reclaim memory?

kamil-sienkiewicz-asi:
@HarelM 
Ok I see what's going on, I got bamboozled by my monkey patches, if DOM node was not explicitly removed from DOM monkey patch did not detect it, that's why I thought that _tip is causing a memory leak. _tip is fine. However I noticed different memory leak, regarding _closeButton, event listener is never cleaned, so this PR still applies for "popup memory leak" but in different place.

src/ui/popup.test.ts
```diff
@@ -498,6 +498,23 @@ describe('popup', () => {
...
 
+    test('Popup remove() clears _closeButton event listener', () => {
+        const map = createMap();
+
+        const popup = new Popup()
+            .setText('Test')
+            .setLngLat([0, 0])
+            .addTo(map);
+
+        expect(popup._closeButton).toBeDefined();
```
HarelM, line 509: Are there better ways to test this?
This test is using some internal fields which I hope are not needed to test this functionality...

src/ui/popup.ts
```diff
@@ -274,6 +274,9 @@ export class Popup extends Evented {
         }
 
         if (this._map) {
+            if(this._closeButton){
```
HarelM, line 277: I think this can be moved outside the if of the map as map is not needed for this code flow, right?

kamil-sienkiewicz-asi: yep, moved

src/ui/popup.test.ts
kamil-sienkiewicz-asi: what's wrong with accessing private fields in tests? Not sure how I can check if event was removed without touching this field tho

HarelM: Private fields are supposed to be private, and in tests it's better to test the public API and allow changes to the internal fields without the need to change the tests.
In any case, since this is only a removeEventListener in the remove method I don't see a big value in this test if we can't test the public API surface.
I would opt to simply remove it.

HarelM: Please do, sorry for the extra work...

HarelM (approved):

johanrd:
ok, good call. Cheering for @kamil-sienkiewicz-asi to revisit this, as it would be a big win to remove the memory leak here! <3

HarelM:
I think you might register the event twice. 
Overall, the perform is when the close button is added and when the event is unregistered, it's not symmetrical, which is what is causing some of these issues...

HarelM:
But in any case, I added the tests related to this issue, so we can explore fixing the memory leak while keeping the original behavior.

PR #5598, opened 2025-03-07
Fix image rendered partially on terrain

### LINKED ISSUE #1559: Image in raster layer shows only partially on certain zoom levels when 3D terrain is enabled.

pstaszek:
- Fixes #1559 - image was rendered partially on terrain.

**Description of the cause.**
Image is always in always assigned to one tile, but the content could exceed that tile. This approach doesn't work for terrain. Content of each terrain tile is rendered separately, and for each terrain tile MapLibre renders only overlapping tiles (children, parents and same tile). This is why sometimes image is not rendered - because it is on the tile that does not belong to some terrain tile.

As the solution I added optional information to the OverscaledTileID what tiles on various zoom levels are overlapping with this tile. This way while computing tiles to render for each terrain tile, we are able to render image tile as well, even if canonical tile itself are not overlapping (but the content is).

Below is the diagram with overview of changes:
maplibre-terrain-image (1)

For fixing animated image it was necessary to clear all terrain tiles textures (changes in `src/ui/map.ts`).

I also found another way to fix this problem - however I'm not sure if it is doable that way. As the image is applied to single tile, we could do it to tile 0/0/0. However, this creates issues with unsufficient precision of transformation matrix, also there is another issue with terrain tiles, which I didn't figure out wih that approach.

Before:
image_clipped_on_terrain_before

After:
image_clipped_on_terrain_after

For testing you can use scripts provided in issue thread:
- static image: https://jsbin.com/yowixah/1/edit?html,output
- animated image: https://jsbin.com/juwikuq/4/edit?html,output

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


src/source/image_source.test.ts
```diff
@@ -1,16 +1,18 @@
...
 import {type RequestManager} from '../util/request_manager';
 import {sleep, stubAjaxGetImage} from '../util/test/util';
 import {Tile} from './tile';
-import {OverscaledTileID} from './tile_id';
+import {CanonicalTileID, OverscaledTileID} from './tile_id';
 import {type Texture} from '../render/texture';
 import type {ImageSourceSpecification} from '@maplibre/maplibre-gl-style-spec';
 import {MercatorTransform} from '../geo/projection/mercator_transform';
+import {MercatorCoordinate} from '../geo/mercator_coordinate';
+import {type CanonicalTileRange} from '../../dist/maplibre-gl';
```
HarelM, line 15: We should not import from dist folder.

src/source/image_source.test.ts
```diff
@@ -230,4 +232,47 @@ describe('ImageSource', () => {
...
+        const coords = [
+            new MercatorCoordinate(0, 0, 0),
+            new MercatorCoordinate(1, 0, 0),
+            new MercatorCoordinate(1, 1, 0),
+            new MercatorCoordinate(0, 1, 0)
+        ];
+
+        const ranges = getOverlappingTileRanges(coords);
+
+        const getMaxRange = (z: number) =>
```
HarelM, line 256: Can this method be placed outside the test?

src/source/tile_id.ts
```diff
@@ -8,6 +8,13 @@ import {type ICanonicalTileID, type IMercatorCoordinate} from '@maplibre/maplibr
 import {MAX_TILE_ZOOM, MIN_TILE_ZOOM} from '../util/util';
 import {isInBoundsForTileZoomXY} from '../util/world_bounds';
 
+export type CanonicalTileRange = {
```
HarelM, line 11: This looks like a bbox type, what's the difference?

src/source/tile_id.ts
```diff
@@ -96,6 +103,11 @@ export class OverscaledTileID {
      * The matrix should be float32 in order to avoid slow WebGL calls in Chrome.
      */
     terrainRttPosMatrix32f: mat4 | null = null;
+    /**
+     * This object is used to store the range of terrain tiles that overlap with this tile.
+     * It is relevant for image tiles, as the image exceeds single tile boundaries.
+     */
+    terrainTileRanges: {[zoom: string]: CanonicalTileRange};
```
HarelM, line 110: This feels a bit odd to me as this is not a "regular" property of a single tile, can this be stored somewhere else perhaps?

src/ui/map.ts
```diff
@@ -2086,7 +2086,13 @@ export class Map extends Camera {
                             this.transform.setElevation(this.terrain.getElevationForLngLatZoom(this.transform.center, this.transform.tileZoom));
                         }
                     }
-                    this.terrain.sourceCache.freeRtt(e.tile.tileID);
+
+                    const source = this.style.getSource(e.sourceId);
+                    if (source?.type === 'image') {
```
HarelM, line 2091: Is it possible to move this logic so that map doesn't need to do something special for the case of image source?

src/source/image_source.test.ts
pstaszek: Ofc, automatic import :).

src/source/image_source.test.ts
pstaszek: I actually removed it, this logic can be directly placed in the loop.

src/source/tile_id.ts
pstaszek: Values there reference tile X and Y number, while bbox I could find in the repo were abould LngLat. Unless I missed some other type.

src/source/tile_id.ts
pstaszek: Yeah, it was difficult to find such logical place. I chose `OverscaledTileID` for following reasons:
- it already contains specific information for terrain tile, it seems logical to me to keep them in the same place
- TileID-kind type seems valid for such information, as it describes placement of the tile. It is not about the data as an example.
- This is the type we provide to `TerrainSourceCache.getTerrainCoords` in `RenderToTexture.prepareForRender` method - which seems logical, and we don't have to change that.

src/ui/map.ts
pstaszek: Tbh this entire fragment of code seems a bit special - I will take a closer look on that, I would try to move it somewhere.

src/source/tile_id.ts
HarelM: I think there's a geojson.bbox type that is simply an array of 4 numbers, but it might not be a great fit here, IDK...

src/ui/map.ts
pstaszek: I wasn't able to find better placement for this logic.

We have to clear all terrain tiles for image source because we don't know which terrain tiles were affected, for two reasons:
- image can exceed its tile (however we can use new range object introduced in this PR),
- image could be moved from one tile to another, but we only get information about new tile.
I don't see the way to fix that without some more major refactoring.

What I tried:
- add `hasCustomTerrainRange` method to `OverscaledTileID`, then we can ready from tileId in `TerrainSourceCache.freeRtt` that this tile is in fact an image tile - but it is very implicit, the logic won't be clear in my opinion
- pass information about the source to `freeRtt` but it is also bad, as `freeRtt` method probably also should not be aware of source

This place seems the best, as it contains already logic that decides when to clear terrain tiles based on event-related circumstances - eg. when style changes, clear all.

src/source/tile_id.ts
pstaszek: Right, I missed that.

Hmm... I still think separate type here is a better fit. GeoJSON BBox implies that it contains geographic coordinates - longitude and latitude, while here we store canonical tile numbers. Also having values in an array is a bit less robust, that having them explicitly named.

Let me know if this rationale sounds good for you.

HarelM: I would change the properties names in that case:
`minTileX` etc.
I know the type has that name inside, but if we keep the type as is, minX isn't descriptive enough IMHO...

src/source/terrain_source_cache.test.ts
```diff
@@ -80,4 +80,117 @@ describe('TerrainSourceCache', () => {
...
+        test('should call _getTerrainCoordsForOversizedTile for tile with custom range', () => {
+            const tile = new OverscaledTileID(5, 0, 5, 17, 11);
+            tile.terrainTileRanges = {1: {minX: 0, maxX: 1, minY: 0, maxY: 1}};
+            tsc.getTerrainCoords(tile);
+            expect(getCoordsRegularTileSpy).not.toHaveBeenCalled();
+            expect(getCoordsOversizedTileSpy).toHaveBeenCalled();
+        });
+    });
+
+    describe('#_getTerrainCoordsForRegularTile', () => {
```
HarelM, line 113: Is it possible to test only the "public" API? I think that this method is a bit more internal...

src/source/terrain_source_cache.ts
```diff
@@ -177,6 +194,58 @@ export class TerrainSourceCache extends Evented {
         return coords;
     }
 
+    /**
+     * Searches for the corresponding current renderable terrain-tiles.
+     * Includes terrain tiles that are within terrain tile ranges of the tileID.
+     * @param tileID - the tile to look for
+     * @returns the tiles that were found
+     */
+    _getTerrainCoordsForOversizedTile(tileID: OverscaledTileID): Record<string, OverscaledTileID> {
```
HarelM, line 203: Some part of the code here looks very similar to the other method, is it possible to reduce duplication somehow?

HarelM:
Can you share some flow diagram that better explains the solution?
I saw the code changes, but it would be great if you could share the flow of data to better explain the solution.

src/source/terrain_source_cache.test.ts
pstaszek: Yes, it is possible. I've changed tests.

src/source/tile_id.ts
pstaszek: Done.

src/source/tile_id.ts
pstaszek: I moved `terrainTileRanges` to `ImageSource`. Please take a look, especially on `RenderToTexture.prepareForRender` if it looks better now.

pstaszek:
@HarelM
Sorry for long reply.
I created diagram - I was wondering what exactly to include on this diagram, and I chose to keep it simple. Let me know if it is sufficient.

maplibre-terrain-image (1)

src/source/terrain_source_cache.ts
pstaszek: I did try already, but I ended with complex tree of ifs, which was much worse than how it looks now. Unfortunately code that is similar is mixed with code that is different :/.

HarelM (approved):
Added a minor comment, approved otherwise.

HarelM:
Can you add the diagram to the first post of this thread for future readers?

src/source/image_source.ts
```diff
@@ -314,6 +331,47 @@ export function getCoordinatesCenterTileID(coords: Array<MercatorCoordinate>) {
         Math.floor((minY + maxY) / 2 * tilesAtZoom));
 }
 
+/**
+ * Given a list of coordinates, determine overlapping tile ranges for all zoom levels.
+ *
+ * @returns Overlapping tile ranges for all zoom levels.
+ * @internal
+ */
+function getOverlappingTileRanges(
```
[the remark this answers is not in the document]
pstaszek: Moved. I did it because there are two other functions in `image_source.ts` files which are used only in `ImageSource` class, so I followed the pattern.

pstaszek:
> Can you add the diagram to the first post of this thread for future readers?

Done.

HarelM:
Can you check the coverage report, for some reason it's missing 
Also, do you think a render test can help here to "keep" this behavior in the future?

pstaszek:
For the render test - yes, this is why I [added one](https://github.com/maplibre/maplibre-gl-js/pull/5598/files#diff-eb172235ab79bef5f80699ac812dd7ab24193bcec8096027830a3fa52fecffb5). I think that this kind of bug is easy to reintroduce in future, and render test should help to avoid that.

For the coverage test - it looks that something is certainly off, coverage report is not working (not only in this PR, but I also briefly checked other PRs, and there is no coverage report as well). When I click on the link to test it shows me last (merge main) change.

pstaszek:
I checked manually how coverage changes for each changed ts file (only unit tests):
```
src/render/render_to_texture.ts: 93.69% -> 93.85%
src/source/image_source.ts: 87.82% -> 89.83%
src/source/terrain_source_cache.ts: 66.96% -> 88.43%
src/ui/map.ts: 87.19% -> 86.9%
```

PR #5599, opened 2025-03-07
Fix circle hitbox in globe projection

### LINKED ISSUE #5255: Globe Projection Bug: Circle Layers have smaller hover/click area than visually represented

lucaswoj:
fix #5255

## Problem

When using `circle` layers in the `globe` projection, `queryRenderedFeatures` only works over tiny fraction of the circle rather than the entire rendered feature. This makes interactive maps with circle layers frustrating to use at low zoom levels.

## Solution

In the `CircleStyleLayer#queryRenderedFeatures` function, we convert tile coordinates into pixel coordinates. Before the `globe` projection was introduced, this operation could be represented by a [transformation matrix](https://en.wikipedia.org/wiki/Transformation_matrix) (`pixelPosMatrix`). However, with the `globe` transformation, this operation is not a linear transformation, as it requires trig (see [`projectTileCoordinatesToSphere`](https://github.com/maplibre/maplibre-gl-js/blob/75299d54466643629681f039e0a49f9c4d004ac3/src/geo/projection/globe_utils.ts#L54-L74)).

This PR refactors `CircleStyleLayer#queryRenderedFeatures` to use the `transform.projectTileCoordinates` API instead of `pixelPosMatrix`.

## Testing

- Added integration tests that verify circle hit testing works correctly in Globe projection
- Verified both exact center hits and edge hits work correctly at various zoom levels, pitches, and with 3d terrain 

## Launch Checklist

 - [x] Link to related issues.
 - [x] ~Post benchmark scores.~ _These should not change as a result_


lucaswoj:
@HarelM @kubapelc Would you take a look at this PR and let me know what you think? I understand a more in depth fix is needed to "fully" support `queryRenderedFeatures` in the `globe` projection but this is a non-intrusive fix the bug that's affecting users right now.

HarelM:
When I looked at this feature I saw that the offset code is also broken. 
Did you see what results you get for offset? 
In any case, this is better than current state so I'll probably merge this anyway.

lucaswoj:
Thanks for the review! 

I’ll look into whether circle-offset works on the glove projection, write a quick blurb on what it’d take to fully resolve the TODO comment, and get back to you early next week.

lucaswoj:
I'm sorry I haven't had a chance to follow up on this PR yet. Planning to do that soon!

src/source/query_features.ts
```diff
@@ -75,10 +76,11 @@ function getPixelPosMatrix(transform, tileID: OverscaledTileID) {
     const t = mat4.create();
     mat4.translate(t, t, [1, 1, 0]);
     mat4.scale(t, t, [transform.width * 0.5, transform.height * 0.5, 1]);
-    if (transform.calculatePosMatrix) { // Globe: TODO: remove this hack once queryRendererFeatures supports globe properly
+    if ('calculatePosMatrix' in transform) { // Globe: TODO: remove this hack once queryRendererFeatures supports globe properly
```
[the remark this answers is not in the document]
lucaswoj: Thanks for the suggestion, @HarelM HarelM! I've refactored this PR so that `CircleStyleLayer#queryIntersectsFeature` no longer uses `calculatePosMatrix` / `pixelPosMatrix`.

I think removing `calculatePosMatrix` from `FillExtrusionStyleLayer` would be fairly challenging and better suited for a separate PR. Once we complete that refactor, we can fully remove `calculatePosMatrix` from the codebase. Let me know if you'd like me to ticket that out.

lucaswoj:
> When I looked at this feature I saw that the offset code is also broken.
> Did you see what results you get for offset?
> In any case, this is better than current state so I'll probably merge this anyway.

I can repro this bug on pitched maps. The codepath that handles `circle-translate` (and all other `*-translate` props) within `queryRenderedFeatures` expects a linear transformation between tile coordinates and pixel coordinates, however this is not true in the `globe` projection. Unfortunately this will require a separate fix applied across all `StyleLayer` classes, which is better suited for a separate PR.

src/source/query_features.ts
```diff
@@ -129,7 +130,9 @@ export function queryRenderedFeatures(
                 params,
                 transform,
                 maxPitchScaleFactor,
-                getPixelPosMatrix(sourceCache.transform, tileIn.tileID))
+                getPixelPosMatrix(sourceCache.transform, tileIn.tileID),
+                getElevation.bind(undefined, tileIn.tileID.toUnwrapped()),
```
HarelM, line 134: Can we use lambda instead of `bind`?

src/style/style.ts
```diff
@@ -1467,7 +1467,8 @@ export class Style extends Evented {
                     serializedLayers,
                     queryGeometry,
                     paramsStrict,
-                    transform)
+                    transform,
+                    this.map.terrain ? this.map.terrain.getElevation.bind(this.map.terrain) : () => 0)
```
HarelM, line 1471: Same here - avoid using `bind`.

src/style/style_layer/circle_style_layer.ts
```diff
@@ -61,20 +63,21 @@ export class CircleStyleLayer extends StyleLayer {
...
 
         for (const ring of geometry) {
             for (const point of ring) {
 
-                const transformedPoint = alignWithMap ? point : projectPoint(point, pixelPosMatrix);
+                const transformedPoint = alignWithMap ? point : projectPoint(point, transform, unwrappedTileID, getElevation);
 
                 let adjustedSize = transformedSize;
-                const projectedCenter = vec4.transformMat4([] as any, [point.x, point.y, 0, 1], pixelPosMatrix);
+                const w = transform.projectTileCoordinates(point.x, point.y, unwrappedTileID, getElevation).signedDistanceFromCamera;
```
HarelM, line 75: Is this similar in terms of performance for Mercator? 
I'm OK with a performance degradation for globe to achieve accuracy, but I don't want to degrade Mercator performance if possible...

src/style/style_layer/circle_style_layer.ts
```diff
@@ -85,13 +88,16 @@ export class CircleStyleLayer extends StyleLayer {
     }
 }
 
-function projectPoint(p: Point, pixelPosMatrix: mat4) {
-    const point = vec4.transformMat4([] as any, [p.x, p.y, 0, 1], pixelPosMatrix);
-    return new Point(point[0] / point[3], point[1] / point[3]);
+function projectPoint(point: Point, transform: IReadonlyTransform, unwrappedTileID: UnwrappedTileID, getElevation: (x: number, y: number) => number): Point {
+    const projection = transform.projectTileCoordinates(point.x, point.y, unwrappedTileID, getElevation);
+    return new Point(
+        (projection.point.x * 0.5 + 0.5) * transform.width,
```
HarelM, line 94: Is this rounding? Can you elaborate on this math?

src/style/style_layer/circle_style_layer.ts
lucaswoj: Yes, when the `mercator` projection is used, this code is effectively unchanged. 

The `MercatorTransform#projectTileCoordinates` method simply calls `calculatePosMatrix` (whose result is cached) and then `transformMat4`. 

https://github.com/maplibre/maplibre-gl-js/blob/58f3a82d737e30d630105a43eaa1b91071eb242b/src/geo/projection/mercator_transform.ts#L764-L779

src/style/style.ts
lucaswoj: Done

src/source/query_features.ts
lucaswoj: Yup, done in the latest commit.

src/style/style_layer/circle_style_layer.ts
lucaswoj: I added some better identifier names and code comments to clarify what's happening here. Let me know if this is enough!

src/style/style_layer/circle_style_layer.ts
```diff
@@ -85,13 +88,19 @@ export class CircleStyleLayer extends StyleLayer {
     }
 }
 
-function projectPoint(p: Point, pixelPosMatrix: mat4) {
-    const point = vec4.transformMat4([] as any, [p.x, p.y, 0, 1], pixelPosMatrix);
-    return new Point(point[0] / point[3], point[1] / point[3]);
+function projectPoint(tilePoint: Point, transform: IReadonlyTransform, unwrappedTileID: UnwrappedTileID, getElevation: undefined | ((x: number, y: number) => number)): Point {
+    // Convert `tilePoint` from tile coordinates to clip coordinates.
+    const clipPoint = transform.projectTileCoordinates(tilePoint.x, tilePoint.y, unwrappedTileID, getElevation).point;
+    // Convert `clipPoint` from clip coordinates into pixel/screen coordinates.
```
HarelM, line 94: Ahh, I see.
Can this conversion logic be somewhere else maybe? Utils? Point?
Is it used somewhere else that we need to extract the logic to align implantation?

HarelM:
Added a small nit picking...

src/style/style_layer/circle_style_layer.ts
HarelM: I understand your concern about performance, is there a way to test the performance impact of the last change?
You could also return a point object from the method to reduce extra object creation I guess.

lucaswoj: Thanks for your feedback!

We have a `QueryBox` performance benchmark (run `npm run start-bench` and then open http://localhost:9966/test/bench/versions/index.html#QueryBox). 

It's an imperfect measure, and I view its results skeptically because:
1. I get inconsistent results between runs (despite a reported R^2 of 1.000)
2. It samples a reasonable number of `symbol` features but no `circle` features
3. No benchmark can fully reflect all real-world usage patterns

Nevertheless, I ran comparisons between the implementations:

**Commit 66f841c3a** with `projectClipSpaceToPixelSpace` utility:
Screenshot showing benchmark results
This shows a small but measurable slowdown (~5%) with the abstraction.

**Commit 1ffcb50** without the abstraction:
Screenshot showing benchmark results
No measurable performance difference here.

I've considered some design alternatives:
- Returning a `Point` object from `projectClipSpaceToPixelSpace` would solve the intermediate object issue for circle layers but would likely worsen it for symbol layers which need a `ProjectionPoint` instance
- Using an "output argument" on `projectClipSpaceToPixelSpace` (i.e. 0e765cb0f) might help performance but introduces an anti-pattern that reduces code readability

Given that this is performance-sensitive code with just 2 lines of simple arithmetic, I believe we should heed the [[Rule of Three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming))](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)), which cautions against premature abstraction.

Would you be comfortable proceeding with the 1ffcb50 approach for now? We can always revisit and refactor if we encounter more use cases for this transformation later. The highest priority is fixing this bug without introducing any performance regressions.

HarelM: Sure, 5% should be avoided given that the comment is descriptive enough and the code is 2 lines. 
Thanks for looking into this!

lucaswoj: Thanks! I have reverted to 1ffcb5094b617 and merged the latest changes from `main`. I believe this is ready to merge.

HarelM:
I really want to get coverage report for this PR.
I'll see if there's something I can do to solve this.
I think the query tests are not adding coverage to the report, but I'm not sure, need to check...

lucaswoj:
> I really want to get coverage report for this PR. I'll see if there's something I can do to solve this. I think the query tests are not adding coverage to the report, but I'm not sure, need to check...

A coverage report for this PR is posted at https://app.codecov.io/github/maplibre/maplibre-gl-js/commit/5cb217d86aea35acdccb0ed6bbb052788f228734/tree/src?dropdown=coverage

The tests in `test/integration/query` do not contribute to our code coverage report. We can see that [`LineStyleLayer#queryRenderedFeatures`](https://app.codecov.io/github/maplibre/maplibre-gl-js/blob/main/src%2Fstyle%2Fstyle_layer%2Fline_style_layer.ts#L95) and [`FillStyleLayer#queryRenderedFeatures`](https://app.codecov.io/github/maplibre/maplibre-gl-js/blob/main/src%2Fstyle%2Fstyle_layer%2Ffill_style_layer.ts#L45) both report as having no coverage despite good coverage in `test/integration/query`. 

In [`query.test.ts` we directly load `dist/maplibre-gl.js`](https://github.com/maplibre/maplibre-gl-js/blob/74e8f1addd8eaf05218f86b5448d51aa2e568b7a/test/integration/query/query.test.ts#L188) which is not instrumented for code coverage measurement.

lucaswoj:
> I'll see if I can add coverage to these tests to see the coverage for this PR, I hope this can be done in a few days, sorry for the delay...

Good luck. Keep me in the loop. There are some folks anxiously awaiting this bug fix.

HarelM:
I've opened the following PR:
- #5674

If this doesn't get merged within a day or two you can take the changes there and apply them locally to see that this doesn't degrade the coverage of the query tests.

HarelM (approved):

HarelM:
Let's wait for the missing coverage report and see that the changed code is covered.

lucaswoj:
Wow! That was fast! Thank you. Happy to address any gaps in coverage once the report is published.

HarelM:
The coverage report doesn't like this PR for some reason.
In any case, this looks like it's fully covered so that's great.
I'll be merging this now and release a new version soon.

HarelM (approved):

HarelM:
It would be great if you could also push forward the removal of the "TODO" that was discussed as part of this PR, in a different PR.

PR #5601, opened 2025-03-10
add migrateProjection to bench StubMap

LostDragonist:
Quick little PR to fix an issue with the Layout benchmarks. Trying to load the style would result in a silent exception that made the test just sit there and spin.


HarelM (approved):
Thanks!!

Issue #5588, opened 2025-03-05
icon-offset is incorrect at top and bottom of viewport when map is pitched


almccon:
I'm drawing oneway arrows for each lane in a road, so I'm using `icon-offset` to place them in each lane. They look fine when the map is not pitched, but when I tilt the map even a little bit, the line offset is too large at the top of the viewport (in the distance) and it is too small at the bottom (closer to the viewer). The offset is correct only right in the middle distance of the viewport.

**maplibre-gl-js version**:

5.2.0, also confirmed it previously existed as early as version 2

**browser**:

Firefox, Chrome

### Link to Demonstration

Minimal example:

https://codepen.io/almccon/pen/KwKwWaz

(Note, I am also doing a `line-offset` to draw dashed lines dividing each lane, and those lines seem to be offset correctly. It's only `icon-offset` that exhibits this problem. Also note that I am using a zoom interpolation for `icon-size`. When you zoom in and out in the example, it might appear that the incorrectness of the offset is changing as you zoom, but that's only because the icons are moving closer to or further from the top and bottom of the viewport)

Screenshot:

![Image](https://github.com/user-attachments/assets/c4951821-2a19-4e33-9804-c595e36579c4)

### Expected Behavior

The arrows should be perfectly placed in the center of each lane even when you tilt the map.

### Actual Behavior

In the distance, the arrows are offset too wide, and in the foreground they're not offset far enough.

zerda-ocm:
Might be related to a precision problem when overzooming tiles (triangles should be placed on the line):

![Image](https://github.com/user-attachments/assets/3fa70bb8-c2c1-45f6-8047-6a1d17235fe1)

Does your described behaviour also happen at zoom 14?

HarelM:
If I needed to guess, by the picture you sent, it looks like the pitch is not taken into consideration for the offset of the icon. 
Can you create a simple reproduction please? 
Preferably with as little layers as possible.

springmeyer:
I've taken a look at this and I think the problem is that icons (as well as text) have a `perspective_ratio` applied to their size on the GPU (in the shaders). 

https://github.com/maplibre/maplibre-gl-js/blob/2f5a47a72fb9dfaa0cdcbf4a68beb34dbe50cb54/src/shaders/symbol_icon.vertex.glsl#L64-L69

And this code does not play nicely with the expectations for precise offsetting in pitched scenarios.

This is currently done in these 3 files:

```bash
[...]

PR #5625, opened 2025-03-14
Disable icon scaling if offset is applied

springmeyer:
## Context

For pitched maps (and the navigation use case), maplibre makes on-the-fly adjustments to labels/icons with the goal of improving information density and readability. These adjustments change the size and placement of labels/icons in the far and near distance, by design, because their size is scaled differently depending on the distance from the center.

## Problem

However for precise displacement of icons with `icon-offset` this behavior is not desirable as @almccon highlights in maplibre/maplibre-gl-js#5588. In this case the icon density should not be altered. Nor should icon size be changed to be more readable (big in distance, or small in foreground) because that would cause misalignments with the linear geometries they should match.

## Solution

This PR allows for disabling of the scaling of icons when they are offset.

## Discussion

This is intended to be an initial fix so that testing can be done to ensure this has no unwanted side effects for Stamen's use case. This change may not work for stylesheets that have a use case where icon's should be offset and scaled ( discussions and feedback therefore arewelcomed if others have insight into this potential). So, this may break expectations for maps that have that usecase. In that case another option could explored. It would be to start respecting https://maplibre.org/maplibre-style-spec/layers/#icon-pitch-alignment in the icon shader. If `map` is set then the scaling could be disabled. This also might break some existing maps, but could at least allow icon scaling and offsetting to be controlled independently. What do others think? Would you prefer to see a PR of that instead of this approach?

## Launch Checklist

 - [x] Link to related issues.
 - [x] Include before/after visuals or gifs if this PR includes visual changes. (see https://github.com/maplibre/maplibre-gl-js/issues/5588)
 - [ ] Ensure this does not regress any known use cases or visual tests
 - [ ] Post benchmark scores.

/cc @ebrelsford


src/render/program/symbol_program.ts
```diff
@@ -26,6 +26,7 @@ export type SymbolIconUniformsType = {
     'u_texture': Uniform1i;
     'u_translation': Uniform2f;
     'u_pitched_scale': Uniform1f;
+    'u_is_offset': Uniform1i;
```
HarelM, line 29: What about using the pitched_scaled? Was it not meant for this scenario?

HarelM:
I tend to think that icon-pitch-aligment=map should facilitate for that, shouldn't it? 
Can you add some before and after visuals and also add render test(s)?

src/render/program/symbol_program.ts
springmeyer: Great idea, will take a look to ensure the Boolean is the same. Thanks!

springmeyer:
> I tend to think that icon-pitch-aligment=map should facilitate for that, shouldn't it? 
> 
> Can you add some before and after visuals and also add render test(s)?

I've thought the same, but not had a chance to test yet. Next step I've been thinking is to add some additional tests, just like you propose.

src/render/program/symbol_program.ts
springmeyer: I took at look at using `pitched_scaled`. My understanding is that this is a floating point that gets passed the value of `transform.getCircleRadiusCorrection()` which differs depending on the projection of the map. So this was not, as far as I can tell, meant to be used for controlling the scaling of icons and not specifically icons that are offset. The issue is different. It is that no scaling should be applied to icons that are offset, when the stylesheet author wants the icons to be offset in the same way as lines (which are not scaled). 

So, the deeper issue I think is that no existing parameter can be used, since this is a bit of a custom usecase that does not seem handled by the current code. Rather the current code assumes it is safe to scale icons and text irrespective of whether they are offset, which means their offset is impacted by the scaling.

Overall the more I look at this problem the more it seems like a new styling property will be needed to be able to allow the behavior @almccon desires without having unintended impacts to existing styles.

springmeyer:
> I tend to think that icon-pitch-aligment=map should facilitate for that, shouldn't it?

I took a closer look at this. `icon-pitch-aligment=map` is already being set in the test case I'm using. And I think the behavior is correct currently. One might be curious, could I hook into the scenario in which `icon-pitch-aligment` is set to `map` and disabling scaling? I looked at that and that would break too much existing behavior as it would disable scaling overall (across all icons and text aligned to the map whether offset or not) which I don't think would  be desirable/viable.

So as I mentioned above it is starting to seem that a solution to this problem that does not break existing use cases is tricky and likely only possible with a new styling property. Maybe something like `icon-pitch-aligment=map_unscaled`...

> Can you add some before and after visuals and also add render test(s)?

Did this in #5654, so that I could get the tests in before this lands.

HarelM:
Let's assume for a second that we are not taking into consideration backwards compatibility, and don't mind breaking existing styles.
Why is the current behavior of icon pitch alignment set to map is "correct"? Shouldn't it scale the same way lines are scaled?
It might be a good idea to talk about it in the monthly meeting to try and better explain the problem and your point of view...

springmeyer:
> It might be a good idea to talk about it in the monthly meeting to try and better explain the problem and your point of view...

Great idea.

> Shouldn't it scale the same way lines are scaled?

The short answer is no. By design, since https://github.com/maplibre/maplibre-gl-js/commit/88637983950bbd62fc36d23b9733e4b89bcc4dcd, text and icons are scaled differently than line geometries. They are made bigger in the distance and smaller in the foreground than the default scaling math would make them, for improved readability/look.

springmeyer:
> They are made bigger in the distance and smaller in the foreground than the default scaling math would make them, for improved readability/look.

Here are two examples that might help readers understand the underlying issue (and therefore hopefully be more able to weigh in on the right solution):

1. Image 1 "current.png" below includes a pitched view with unchanged, maplibre gl js.
 
current

2. Image 2, "attenuation-disabled.png". This has the same scaling for lines and all icons such that the extra attenuation (added in https://github.com/maplibre/maplibre-gl-js/commit/88637983950bbd62fc36d23b9733e4b89bcc4dcd) is removed completely. While the patch in this PR disables the attenuation only for offset icons, the patch used to create this example does it for all icons whether they were offset or not, just for the sake of the visual example here. Notice how the trees become more crowded. Some might look at this and say it looks fine, but nevertheless it would be a regression of current maps (which I've felt was not acceptable).

attenuation-disabled

HarelM:
In the image you shared - the after parking icon on the right looks small when closer and bigger when far (or are they at the same size but the effect makes it feel this way), can you explain why?
I'm not sure changing the map in a subtle way to fix a bug can be considered a breaking change, but if we decide to do so we need to make sure the community is aware of such a change.

springmeyer:
> In the image you shared - the after parking icon on the right looks small when closer and bigger when far (or are they at the same size but the effect makes it feel this way), can you explain why?

In the after image the parking icons are the same size near and far. The reason why is that I've disabled attentuation (no matter whether they are offset or not) to show what it looks like if the icons are not scaled.

> I'm not sure changing the map in a subtle way to fix a bug can be considered a breaking change, but if we decide to do so we need to make sure the community is aware of such a change.

Agree. However the more I've thought about this the more I see this less as a bug as more as an unanticipated usecase. Most users want scaling to be attenuated, while the usecase (as originally described in #5588) needs it disabled. So the question is how best to disable the scaling. I think the proper solution is probably to create a new style spec property so that scaling can be disabled without impacting or interacting with the existing use cases which the scaling addresses.

HarelM:
Having another style spec property might be confusing to grasp, so if possible I would consider avoiding it. 
I still don't fully understand why when setting an icon to pitch with the map it doesn't provide the required effect...?

springmeyer:
I'm going to close this PR now. My current efforts here have been useful to adequately enable the team at Stamen to workaround #5588. As I've mentioned above, the more ideal fix I see is a PR to introduce a new style spec property to explicitly disable attenuation that works independently of `icon-offset`. If anyone is interested in working on that, please reach out to @almccon!

Thanks so much @HarelM for the feedback and help on this!

HarelM:
@CommanderStorm let me know if you want to open a PR to implement this to align with native implementation.
I'll be happy to review it.

CommanderStorm:
Shrue, is simple enough of a fix. But lets merge https://github.com/maplibre/maplibre-gl-js/pull/7490 first

louwers:
We have releases out for iOS and Android with this change.

Will let you know if we get complaints from anyone.

HarelM:
Thanks for the update, we will include it as part of version 6 breaking changes.

HarelM:
@CommanderStorm @birkskyum let me know if any of you would like to push this forward as part of v6 breaking changes, this also aligns the behavior with native.

CommanderStorm:
I also don't think this needs to block V6.

To me this is pretty clearly a bug, so I think it can be addressed separately.

As I offered above, I can work on this after my current PR-stack for the change I need is done.

HarelM:
- Superseded by #7742

Issue #5616, opened 2025-03-11
SourceCache `_updateRetainedTiles` get type-error on checking `key` of undefined


jessekrubin:
Getting the `OverscaledTileID` often returns an array of len uno (1), but where the array values are checked, if the array is not len 4, then getting `key` on `undefined` (where you expect an `OverscaledTileID` gives a type error.

I think the problem is in the function (marked with `HERE HERE HERE`):

```ts

    _updateRetainedTiles(idealTileIDs: Array<OverscaledTileID>, zoom: number): {[_: string]: OverscaledTileID} {
        const retain: {[_: string]: OverscaledTileID} = {};
        const checked: {[_: string]: boolean} = {};
        const minCoveringZoom = Math.max(zoom - SourceCache.maxOverzooming, this._source.minzoom);
        const maxCoveringZoom = Math.max(zoom + SourceCache.maxUnderzooming,  this._source.minzoom);

        const missingTiles = {};
        for (const tileID of idealTileIDs) {
            const tile = this._addTile(tileID);

            // retain the tile even if it's not loaded because it's an ideal tile.
            retain[tileID.key] = tileID;

            if (tile.hasData()) continue;

            if (zoom < this._source.maxzoom) {
                // save missing tiles that potentially have loaded children
                missingTiles[tileID.key] = tileID;
            }
        }

        // retain any loaded children of ideal tiles up to maxCoveringZoom
        this._retainLoadedChildren(missingTiles, zoom, maxCoveringZoom, retain);

        for (const tileID of idealTileIDs) {
            let tile = this._tiles[tileID.key];

            if (tile.hasData()) continue;

            // The tile we require is not yet loaded or does not exist;
            // Attempt to find children that fully cover it.

            if (zoom + 1 > this._source.maxzoom) {
                // We're looking for an overzoomed child tile.
                const childCoord = tileID.children(this._source.maxzoom)[0];
                const childTile = this.getTile(childCoord);
                if (!!childTile && childTile.hasData()) {
                    retain[childCoord.key] = childCoord;
                    continue; // tile is covered by overzoomed child
                }
            } else {
                // check if all 4 immediate children are loaded (i.e. the missing ideal tile is covered)
                const children = tileID.children(this._source.maxzoom);

//=======================================
// HERE HERE HERE
//=======================================
                if (retain[children[0].key] &&
                    retain[children[1].key] &&
                    retain[children[2].key] &&
[...]

PR #5630, opened 2025-03-14
fix: update retained tiles bug 5616

jessekrubin:
Fix for #5616 

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 ### 🐞 Bug fixes
 
 - Fix intersection detection between MultiPolygons and Points ([#5590](https://github.com/maplibre/maplibre-gl-js/pull/5590))
+- Fix [#5616](https://github.com/maplibre/maplibre-gl-js/issues/5616) `_updateRetainedTiles` checking for children when children len is 1 overscaled tile "child"
 - _...Add new stuff here..._
 
 ## 5.2.0
--- a/src/source/source_cache.ts
+++ b/src/source/source_cache.ts
@@ -747,7 +747,8 @@ export class SourceCache extends Evented {
                 // check if all 4 immediate children are loaded (i.e. the missing ideal tile is covered)
                 const children = tileID.children(this._source.maxzoom);
 
-                if (retain[children[0].key] &&
+                if (children.length === 4 &&
+                    retain[children[0].key] &&
                     retain[children[1].key] &&
                     retain[children[2].key] &&
                     retain[children[3].key]) continue; // tile is covered by children


HarelM:
Can you check which version was the first version that has this issue? 
 I'll be surprised if this isn't a relatively new bug. There was a PR that changed the logic of retained tiles in recent versions...

jessekrubin:
Had to have been 5 as I only started getting these errors with v5+.

HarelM:
@wagewarbler can you check if this issue might have been introduced by your change somehow? 
I'm wondering if there's a better way to solve this besides adding a if statement check, since if this wasn't happening before it means that the content of the cache had changed and it is filled differently than before...

HarelM:
Seems like there's no response from @wagewarbler.
@jessekrubin any change you can try and see if there's a better approach to solving this besides adding this if?

wagewarbler:
@HarelM I've been catching up on the issue. It's not immediately clear to me how my PR would have affected this but I understand your point about the nature of the underlying cache potentially being an issue. At a high level, do you have a sense of how this tilt/angle differently utilizes the cache vs. the standard perpendicular view? Or does this require some deeper digging?

wagewarbler:
Ok, I can do some looking in parallel. Entirely your call here, but there is the option to spin off a separate issue to re-evaluate the cache model and flow and link related PRs (this one and my previous one for example). Then merge this one as a tightly scoped PR that prevents array element access that exceed the array's bounds.

jessekrubin:
Hi yall. I say merge it. Sorry I have not been more involved. I broke my leg a week ago :(

HarelM:
I would prefer to have a better understanding of this issue before merging this so that we don't just use "defensive" coding approach for no reason.
If we find out this is the only solution, then sure, we can merge it.
Let's wait for @wagewarbler analysis.

jessekrubin:
> I would prefer to have a better understanding of this issue before merging this so that we don't just use "defensive" coding approach for no reason.
> 
I feel ya but it is currently causing crashes. My guess is it has to do with at high pitches the child tiles get reduced/collapsed to single parent tiles.

Is it possible this collapsing happens multiple times? Or for tiles outside of the range of webmercator bounds (which might make sense at extreme tilts)?

HarelM:
I don't have a clear understand at the moment. 
If you need a quick fix, you can use `patch-package` with this fix while we decide how to proceed.

jessekrubin:
I think what happens is when at a high pitch, certain tiles are "collapsed" to a single over zoomed tile if they are super far out. 

The tiles that get collapsed are always far off in the distance. 

Either way checking that the array is length 1 or 4 seems like a simple sanity check and I would be very happy if this were merged.

For now I have just limited my usage to not allow a pitch greater than 60. Using patch package seems like a huge pita. 

@HarelM What do you need to merge this?

HarelM:
I prefer to have a better root cause analysis of the original issue or what might cause this, sorry for delaying this, but it's very easy to add this type of code, but extremely hard to remove it, so I'm trying to get a more concrete answer here.

MrBlenny:
I'm seeing the same issue with the ancient version - 1.14.0

```js
          var rn = Rt.children(this._source.maxzoom);
                                if (X[rn[0].key] && X[rn[1].key] && X[rn[2].key] && X[rn[3].key])
                                    continue
```

It is intermittent, it seems to happen about 1 in 5 times. I am seeing it when loading a `raster` layer on top of a vector map. Using react-map-gl, the offending layer is something like this.

```
      <Source
        id="radar"
        type="image"
        url={imageUrlForSomeBlob}
        coordinates={coordinates}
      >
        <Layer
          beforeId="some-other-layer"
          id="radar"
          type="raster"
          source="radar"
          paint={{
            'raster-fade-duration': 0,
            'raster-saturation': 1.0,
            'raster-opacity': 0.6,
            'raster-brightness-min': 1.0,
          }}
        />
      </Source>
```

My guess was it was something to do with the raster layer or the `beforeId`. There is also an occasion where the coordinates for the raster may be `[[0, 0],[0, 0],[0, 0],[0, 0]]`

Hopefully another useful datapoint for your investigation.

jessekrubin:
> @MrBlenny can you reproduce this issue without `react-map-gl`? If so, it means that this is not a new issue and the fix here is good enough.

I was able to point my repro at an older version and still see the issue.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### 🐞 Bug fixes
 
 - The state of `glPixelStore` is now properly cleaned after texture updates to avoid `glTexSubImage2D` calls made on the same gl context acting differently at random ([#5730](https://github.com/maplibre/maplibre-gl-js/pull/5730))
+- Fix [#5616](https://github.com/maplibre/maplibre-gl-js/issues/5616) `_updateRetainedTiles` checking for children when children len is 1 overscaled tile "child"
```
HarelM, line 12:
```suggestion
- Fix `_updateRetainedTiles` checking for children when children length is 1 overscaled tile "child" ([#5360](https://github.com/maplibre/maplibre-gl-js/pull/5630))
```

HarelM:
Is it possible to add a unit test to cover this issue?
After adding the test I will merge this.
THANKS!

wayofthefuture:
Just wanted to add that I am also consistently getting the error when using pmtiles protocol via both fetchsource and filesource implementations (see browser console):

https://codepen.io/wayofthefuture/pen/xbGBNKW

wayofthefuture:
> I prefer to have a better root cause analysis of the original issue or what might cause this, sorry for delaying this, but it's very easy to add this type of code, but extremely hard to remove it, so I'm trying to get a more concrete answer here.

@HarelM, @jessekrubin I did some research into this, and it appears that the possible cause is:

https://github.com/maplibre/maplibre-gl-js/blob/f16ccea3d2f9d6a7daf65744778aab4f37e717e4/src/source/tile_id.ts#L154-L169

If this.overscaledZ >= sourceMaxZoom, then children() only returns an array of 1 versus an array of 4.

The cause appears to come from the commit https://github.com/jessekrubin/maplibre-gl-js/commit/7da29b9f57192a0541f780f013937ced891ef5ee, the same author of this PR, in source_cache.ts, where this:

```javascript
                    this._findLoadedChildren(tileID, maxCoveringZoom, retain);
                    // check if all 4 immediate children are loaded (i.e. the missing ideal tile is covered)
                    const children = tileID.children(this._source.maxzoom);
                    for (let j = 0; j < children.length; j++) {
                        if (!retain[children[j].key]) {
                            covered = false;
                            break;
                        }
                    }
```

was changed to this:

```javascript
                // check if all 4 immediate children are loaded (i.e. the missing ideal tile is covered)
                const children = tileID.children(this._source.maxzoom);

                if (retain[children[0].key] &&
                    retain[children[1].key] &&
                    retain[children[2].key] &&
                    retain[children[3].key]) continue; // tile is covered by children
            }
```

You'll notice that in the prior version, if there was only 1 child it would not have thrown, as the loop would have stopped at 1, whereas in the new code, it assumes an array length of 4. **Also, it would have passed the test and continued**. The newer version throws if there is only 1 child, as children() returns an array of [1] or [4] in length.

https://github.com/jessekrubin/maplibre-gl-js/commit/7da29b9f57192a0541f780f013937ced891ef5ee#diff-d05dc4b64c243506ec6722d825bf22b6769b291c29e9e686d5906f876c85134aL591 (LINE 593)

If the single tile returned suffices as it did before, then this PR misses the case when there is 1 available tile, so it could instead be:

```javascript
//single tile representing an overscaled tile
if (children.length === 1 && children[0].key) {
  continue;
}
//all 4 tiles
else if (children.length === 4 && children[0].key && children[1].key && children[2].key && children[3].key) {
  continue;
}

```

Otherwise, if the 1 tile does not suffice, then the version in this PR requiring children.length === 4 may be correct.

jessekrubin:
Yes. That's my understanding.

wayofthefuture:
@jessekrubin Should it continue if there is only 1 child?

jessekrubin:
I don't remember tbh. I can take a look tomorrow.

I forget the flow but iirc but my gut says not as it's just the overzoomed tiles.

wayofthefuture:
Ok. In the original implementation it returned as covered for the overzoomed tile because the loop stopped at 1 and didn't break, in the subsequent version however it appears the overzoomed tile wasn't considered as it assumed an array length of 4 hence the throw. So when the original was done overzoomed was considered covered, and in bringing back that behavior, then could tile loading be improved...

HarelM:
- Closed in favor of: #6388

PR #5654, opened 2025-03-22
Add testcase for #5588

springmeyer:
This PR adds a new rendering test and makes no changes to the rendering library.

Therefore this PR is intended to show the current behavior described in #5588. My hope is that this can be merged first, and then I will followup with a separate PR to improve the rendering behavior (the PR is already started at #5625).

@ebrelsford @almccon


HarelM:
I don't mind merging this, but isn't this not the desired behavior?
You want this merged so later on the image can be updated to the desired behavior, right? 
If this is the case, I would recommend simply keeping it as part of the relevant PR that updates the behavior.

springmeyer:
Thanks for your thoughts @HarelM. I would be inclined to merge this first/standalone because after looking further into #5625, I'm seeing that it may not be the ideal approach. So, de-coupling seems ideal to be and if you are open to merging this beforehand that would be great.

HarelM:
Let's conclude the discussion in the other PR.
If we would like to keep this behavior as is then this test is important and should be kept to make sure we don't break existing styles, but if we decide that we want to change the behavior a bit, then this PR might be "unwanted".

springmeyer:
For now I've pulled the test case into #5625 to keep things together.

PR #5057, opened 2024-11-18
Improve innerHTML handling in code

HarelM:
## Launch Checklist

Improve code around innerHTML, it should not affect attribution in case everything is OK.

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.
 - [x] Write tests for all new functionality.
 - [ ] Add an entry to `CHANGELOG.md` under the `## main` section.

PR #5673, opened 2025-03-25
Fix excessive attribution control rerendering

viernullvier:
## The lead-up

In one of my projects I noticed a weird flickering on the attribution control that seemed to be related to rapid data updates. Notably, it only occurred in Firefox and not in Chromium, so I started to investigate.

## The issue

Since attribution strings can include untrusted HTML, PR #5057 introduces a bit of input sanitization to mitigate potential XSS attacks trough the `attribution` field. The attribution control includes a mechanism to suppress excessive re-renders, but #5057 introduced a bug: The suppression logic compares the cached **sanitized** result with the new **unsanitized** result. This will lead to excessive DOM updates if the sanitized attribution actually differs from the unsanitized one, for instance if it contains a `&copy;` HTML entity that gets transformed into a `©` literal.

## The fix

This PR reverts some of the changes in #5057: The attribution control will cache the unsanitized input again instead of the sanitized result and only call the sanitizer directly when updating the DOM. This might even bring a minor performance improvement since the sanitizer will no longer be called on every update, but only if the unsanitized input has been changed.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 - Fix intersection detection between MultiPolygons and Points ([#5590](https://github.com/maplibre/maplibre-gl-js/pull/5590))
 - Fix issue with image rendered partially on terrain tiles ([#1559](https://github.com/maplibre/maplibre-gl-js/issues/1559)).
+- Fix excessive attribution control rerendering ([#5673](https://github.com/maplibre/maplibre-gl-js/pull/5673))
 - _...Add new stuff here..._
 
 ## 5.2.0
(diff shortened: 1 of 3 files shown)


HarelM:
Thanks!
Ping me when this is ready for review as I don't get notified when a PR status is changed from drag to ready for review.

viernullvier:
@HarelM I think it's ready for review if you don't mind the lack of benchmarks - the headless benchmarks keep crashing on my machine and the in-browser ones are causing out-of-memory issues.

HarelM (approved):
LGTM THANKS!

Issue #5655, opened 2025-03-22
Popup detaches from marker during animation when map repeats horizontally


sankichi92:
**maplibre-gl-js version**: 5.2.0

**browser**: Chrome, Safari

### Steps to Trigger Behavior

1. Add a marker to the map and attach a popup to it.
2. Animate the marker continuously across the map.
3. Zoom out until the map horizontally repeats (world copy is visible).

### Link to Demonstration

https://jsbin.com/totafit/1/edit?html,output

### Expected Behavior

The popup moves synchronously with the marker.

### Actual Behavior

The popup is displayed separated from the marker by 360 degrees of longitude.

<img width="830" alt="Image" src="https://github.com/user-attachments/assets/4e8a23b1-2467-45cd-8818-24f50862cf90" />

### Related Issues

- https://github.com/maplibre/maplibre-gl-js/issues/2775

---

This behavior particularly impacts applications that animate markers for globally-moving objects (e.g., satellites, aircraft), such as [react-sat-map](https://github.com/sankichi92/react-sat-map), resulting in confusing and incorrect visual representation.

sankichi92:
Thank you for your comment!  I attempted to address this issue, but it turned out to be more complex than expected, and a straightforward solution wasn't immediately clear.

yuiseki:
I have confirmed that this Issue is not fixed under specific conditions even in the latest maplible-gl-js v5.5.0.

https://github.com/user-attachments/assets/2d0fc277-0ca1-473a-84a0-2a15fd492f22

However, I have also noticed that at certain screen resolutions or rendering aspect ratios, this problem does not reproduce itself and works perfectly as expected.

https://github.com/user-attachments/assets/354f639e-cbac-4f51-a5a4-b822e5546d50

It is indeed strange, but I would like to work on a fix, since I was able to reproduce this problem under certain situations.

EDIT: This issue does not occur when worlds fit on one screen; it only occurs when multiple worlds are looped on one screen.

Duly330AI:
## Investigation Summary

I investigated this issue and couldn't reproduce it on current \main\. I added comprehensive behavior-based tests in PR #6610 that verify:

- ✅ Popup follows a draggable marker on \setLngLat()\
- ✅ Survives rapid consecutive moves
- ✅ Remains attached during map zoom/pan
- ✅ Position updates without drag interaction

All tests pass, suggesting the bug may already be fixed.

## Environment Details Checked
Tests were written to cover typical scenarios:
[...]

PR #5709, opened 2025-04-02
Fix #5473 - Marker drag Coordinates give Longitude off by ±360° with Globe

### LINKED ISSUE #5473: Marker drag Coordinates give Longitude off by ±360° with Globe
**maplibre-gl-js version**: 5.1.0

**browser**: Chromium 132.0.6834.160

### Steps to Trigger Behavior

 1. Add Marker with `drag` event to a `globe` projection map
 2. Drag Marker around (e.g. within Germany)

https://github.com/user-attachments/assets/f979af5c-ebf2-432f-affa-f02c3606fce9

### Link to Demonstration

https://codesandbox.io/p/sandbox/maplibre-global-marker-drag-xnml7p

### Expected Behavior

Coordinates should be correct, as without `globe` projection.

### Actual Behavior

Sometimes Longitude is off by ±360°, other drag events produce the same behavior.

docentYT:
In marker update added check if projection is not equal to `globe`. Fixes https://github.com/maplibre/maplibre-gl-js/issues/5473

## Launch Checklist

### DIFF
--- a/src/ui/marker.ts
+++ b/src/ui/marker.ts
@@ -601,7 +601,7 @@ export class Marker extends Evented {
             this._map.once('render', this._update);
         }
 
-        if (this._map.transform.renderWorldCopies) {
+        if (this._map.transform.renderWorldCopies && this._map.getProjection().type != 'globe') {
             this._lngLat = smartWrap(this._lngLat, this._flatPos, this._map.transform);
         } else {
             this._lngLat = this._lngLat?.wrap();


docentYT:
> I'm not sure I understand why adding this condition to the if statement solves the problem as when you zoom in enough globe turns into mercator which I assume will create the same issue. Am I missing anything?

The problem is with the `smartWrap()` function, which can return values shifted by `n * 360^`. This is applicable to a flat map, where the map can repeat at small zooms, and is meant to allow markers to be moved across copies of the map. In the case of the sphere, this does not apply because the map loops itself. You can't set the `renderWorldCopies` option to false, because the map must repeat at the merge point on the sphere when zoomed in. However, the transition to the mercator occurs at large zooms, not small ones so the problem that occurs with small zooms should not occur in this case.

> Tests are missing.

I don't know how tests for this if statement should look like.

HarelM:
There are many tests here:
https://github.com/maplibre/maplibre-gl-js/blob/5a2be1e98893177e5b5fb915fff5604e237a196a/src/ui/marker.test.ts
I'm sure there are similar tests that can be used.
I wonder why I didn't get codecov report on this PR, hopefully will be reported soon to make sure the changed code was covered.

In any case, I think this might be better of inside the `smartWrap` method as I see similar code in `popup.ts` file:
https://github.com/maplibre/maplibre-gl-js/blob/5a2be1e98893177e5b5fb915fff5604e237a196a/src/ui/popup.ts#L623

HarelM:
I also think the right way to check this is the `allowWorldCopies` as can be seen here, `renderWorldCopies` was an initial stab at this, but later on, when switching between mercator and globe it needed to keep the original value that was set be the user, so this `allowWorldCopies` was added, this way projection name and existence is not required.
https://github.com/maplibre/maplibre-gl-js/blob/5a2be1e98893177e5b5fb915fff5604e237a196a/src/geo/projection/covering_tiles.ts#L190

HarelM:
Not entirely related, but worth mentioning the following issue:
- https://github.com/maplibre/maplibre-gl-js/issues/5655

I tend to think that there should be more shared code between the marker and the popup class in order to avoid these kind of problems...

docentYT:
You are right that the problem applies to popups as well.

I think I have found a deeper problem with the smartWrap function. I need to investigate this, so I'm marking the pull request as draft.

docentYT:
@HarelM I wrote additional tests for marker and popup.
In `smartWrap()` I check if `transform.getCoveringTilesDetailsProvider().allowWorldCopies()` is true before wrapping lngLat.

src/ui/marker.test.ts
```diff
@@ -1102,6 +1102,25 @@ describe('marker', () => {
         expect(marker._lngLat.lng).toBe(-179);
     });
 
+    test('Marker\'s lng is wrapped when slightly crossing 180 with zoomed out globe', async () => {
+        const map = createMap({width: 1024, renderWorldCopies: true});
+        await new Promise<void>(resolve => map.once('load', () => {
```
HarelM, line 1107: No need to provide a callback, simply await the once method.

src/util/smart_wrap.ts
```diff
@@ -34,18 +34,23 @@ export function smartWrap(lngLat: LngLat, priorPos: Point, transform: IReadonlyT
...
-    while (Math.abs(lngLat.lng - transform.center.lng) > 180) {
-        const pos = transform.locationToScreenPoint(lngLat);
-        if (pos.x >= 0 && pos.y >= 0 && pos.x <= transform.width && pos.y <= transform.height) {
-            break;
-        }
-        if (lngLat.lng > transform.center.lng) {
-            lngLat.lng -= 360;
-        } else {
-            lngLat.lng += 360;
+    if (transform.getCoveringTilesDetailsProvider().allowWorldCopies()) {
```
HarelM, line 37: Since this logic was added here, I would've expected it to be removed from marker and popup class...?

docentYT: `smartWrap` when `allowWorldCopies` is `false returns slightly different coordinates than the `wrap` itself, so I'm not sure if switching to only using `smartWrap` is desirable.

src/ui/marker.test.ts
```diff
@@ -1102,6 +1102,22 @@ describe('marker', () => {
         expect(marker._lngLat.lng).toBe(-179);
     });
 
+    test('Marker\'s lng is wrapped when slightly crossing 180 with zoomed out globe', async () => {
+        const map = createMap({width: 1024, renderWorldCopies: true});
+        await map.once('load', () => {
```
HarelM, line 1107: No need to use both `await` and callback. Use only `await`.

src/ui/popup.test.ts
```diff
@@ -345,6 +345,23 @@ describe('popup', () => {
         expect(popup._pos).toEqual(map.project([-5, 0]));
     });
 
+    test('Popup\'s lng is wrapped when slightly crossing 180 with zoomed out globe', async () => {
+        const map = createMap({width: 1024, renderWorldCopies: true});
+        await map.once('load', () => {
```
HarelM, line 350: Same as above

src/util/smart_wrap.ts
HarelM: The lower part of the code (else below) is using "regular" wrap.
What I would expect to see is that the code in popup.ts and marker.ts would remove the relevant ifs.
https://github.com/maplibre/maplibre-gl-js/blob/fa7a410c325b1c41c6c8e714721859df9b0beaee/src/ui/popup.ts#L623
https://github.com/maplibre/maplibre-gl-js/blob/fa7a410c325b1c41c6c8e714721859df9b0beaee/src/ui/marker.ts#L604

HarelM: True, but you can incorporate this logic into smartWrap I believe. 
As far as I remember, snartWrap is only used by marker and popup and you charge it to fit the logic, can't you?

docentYT: What about `!this._trackPointer` in popup?

https://github.com/maplibre/maplibre-gl-js/blob/d5990260dc05323748ea93f65a94d6f6952657bf/src/ui/popup.ts#L623

HarelM: Can be passed as additional parameter I guess, but I'm not sure what it serves as probably the previous position is what's interesting...

docentYT: So, what now?

HarelM: I suggest to move as much of the logic to the smart wrap method, it should be a small refactoring effort.

docentYT: What do you think about these changes? https://github.com/maplibre/maplibre-gl-js/pull/5709/commits/14bd15b1ea9165bee80fdfc9fa8895c75ee0fc23

HarelM: Looks great! Thanks!

HarelM (approved):

Issue #5650, opened 2025-03-21
LOD control at high pitch angles


JBLINZ:
## User Story

 As a map maker/owner, I would like to be able to control the LOD that MapLibre loads raster tiles and terrain-RGB tiles when the camera is at high pitch angles so that I can ensure high-quality data is viewable in the mid-ground

## Rationale

Essentially, We would like more control over the zoom level of tiles that are loaded when the camera is tilted.  For example, a setting (or range of settings) that allow us to force our map to load zoom level 8 tiles all the way to the boundary of 4-and 5. See image.

![Image](https://github.com/user-attachments/assets/04a790dd-4039-4c6a-9d9b-f684fcfcf3b3)

I understand the economy/performance driving the current logic in MapLibre, but I believe it incorrectly considers where map users generally focus in a 3D map view.  With 3D mapping, the mid and background are more regularly focal points than the foreground, therefore, I'd like to ensure detailed data is loaded in these areas.

NathanMOlson:
@HarelM please assign this to me. LINZ is funding me to complete this work.

NathanMOlson:
I've hacked together a demo for evaluating changes to the tile loading behavior here: https://nathanmolson.github.io/lod_control

This demo includes my two personal favorite parameters for adjusting the tile loading behavior: `TileZoomRate` and `TileZoomDeadband`.

`TileZoomRate` controls the rate at which tile zoom level varies with distance. This parameter is most useful between 0 and 2 (the default is 1). At 0, tiles are loaded with approximately constant screen X resolution. At 1, tiles are loaded with approximately constant screen area. At 2, tiles are loaded with approximately constant screen Y resolution.

Increasing `TileZoomRate` causes lower-zoom (coarser) tiles to be used, particularly near the horizon.

`TileZoomDeadband` sets a zoom deadband around the center point, where all tiles have the same zoom level. This parameter is sensible above 0. Values 5 or higher effectively disable variable zoom and will lead to very many tiles being loaded. The default is 0.

Increasing `TileZoomDeadband` causes higher-zoom (finer) tiles to be used, particularly near the center of the screen.

Source code for the demo is here:
https://github.com/NathanMOlson/NathanMOlson.github.io/tree/main/lod_control
https://github.com/NathanMOlson/maplibre-gl-js/tree/lod_control

[...]

PR #5719, opened 2025-04-06
LOD Control

NathanMOlson:
## Launch Checklist

The PR addresses #5650 by adding the following function to the public API:

```
setSourceTileLodParams(maxZoomLevelsOnScreen: number, tileCountMaxMinRatio: number, id?: string) : boolean
```

`maxZoomLevelsOnScreen` - The maximum number of distinct zoom levels allowed on screen at a time.  There will generally be fewer zoom levels on the screen, the maximum can only be reached when the horizon is at the top of the screen. Increasing the maximum number of zoom levels causes the zoom level to decay faster toward the horizon.
`tileCountMaxMinRatio` - The ratio of the maximum number of tiles loaded (at high pitch) to the minimum number of tiles loaded. Increasing this ratio allows more tiles to be loaded at high pitch angles. If the ratio would otherwise be exceeded, the zoom level is reduced uniformly to keep the number of tiles within the limit.

Previous related work: #4779
Related work in MapLibre Native: https://github.com/maplibre/maplibre-native/pull/2958

 - [X] Link to related issues.


NathanMOlson:
I'm very happy with the new `tileCountMaxMinRatio` parameter, I think it behaves as expected and sets an upper bound on the number of tiles that get loaded, which protects users against accidentally trying to load a huge number of tiles.

I'm less happy with the `maxZoomLevelsOnScreen` parameter. As pointed out in #5650, in the current implementation, the actual number of zoom levels does not match the parameter setting, even at high pitch. I prefer the parameter I previously suggested, `tileZoomRate`, which directly controls how fast the zoom level decays at high pitch angles (despite the fact that it is unintuitive). To me, it feels easier to understand how `tileZoomRate` affects the behavior when you play with the demo. Also, `tileZoomRate` definitely feels like the right thing to adjust on the backend, due to its elegant geometric interpretation.

So I don't think the choice of parameters is fully settled, but I think the implementation mechanics are ready to review in this PR.

NathanMOlson:
Marginally related: with the changes in #4779 and these changes, is it time to re-evaluate the horizon heuristic?

```
/**
 * Calculate pixel height of the visible horizon in relation to map-center (e.g. height/2),
 * multiplied by a static factor to simulate the earth-radius.
 * The calculated value is the horizontal line from the camera-height to sea-level.
 * @returns Horizon above center in pixels.
 */
export function getMercatorHorizon(transform: {pitch: number; cameraToCenterDistance: number}): number {
    return transform.cameraToCenterDistance * Math.min(Math.tan(degreesToRadians(90 - transform.pitch)) * 0.85,
        Math.tan(degreesToRadians(maxMercatorHorizonAngle - transform.pitch)));
}
```

When pitch goes up, the horizon moves further out. But in the real world, when pitch goes up, the camera is moving lower, which should cause the horizon to move closer! So the heuristic feels backward to me. I propose replacing it with a fixed horizon angle of 89.25 degrees (the current value of `maxMercatorHorizonAngle`. A fixed angle would still prevent geometry calculations from blowing up, and the current tile-loading scheme would keep the number of tiles bounded. We would get at most a couple of degrees added to the horizon in high-pitch cases.

Are there any other reasons behind the horizon heuristic that I am unaware of, and that would provide a good reason to keep it in its current form?

src/geo/projection/covering_tiles.ts
NathanMOlson, line 1: TODO: replace this numerical integration with a closed-form solution and document its origin.

src/geo/projection/covering_tiles.ts
```diff
@@ -127,6 +128,50 @@ function calculateTileZoom(requestedCenterZoom: number,
     return thisTileDesiredZ;
 }
 
+function intCosXToP(p:number, x1: number, x2: number): number {
```
NathanMOlson, line 131: TODO: replace this numerical integration with a closed-form solution and document its origin.

test/examples/lod-control.html
```diff
@@ -0,0 +1,120 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>LOD Control</title>
```
HarelM, line 4: Use Level of details, as LOD is a technical term that users might not be familiar with.

test/examples/lod-control.html
```diff
@@ -0,0 +1,120 @@
...
+                    type: 'raster',
+                    source: 'numbers'
+                }
+            ]
+        },
+        maxZoom: 22,
+        maxPitch: 85
+    }));
+
+    class LodControl {
```
HarelM, line 46: This is a bit long, consider using something more html descriptive like in the sky example:
https://github.com/maplibre/maplibre-gl-js/blob/56d5d5cdf9113ee0a5511751b7cda54c2e584f52/test/examples/sky-with-fog-and-terrain.html#L27

src/ui/map_tests/map_set_lod_params.test.ts
```diff
@@ -0,0 +1,54 @@
...
+        expect(map.getSource('source-id2').calculateTileZoom).toBeFalsy();
+        expect(map.setSourceTileLodParams(1, 1)).toBe(true);
+        expect(map.getSource('source-id1').calculateTileZoom).toBeTruthy();
+        expect(map.getSource('source-id2').calculateTileZoom).toBeTruthy();
+
+        done();
+    });
+}));
+
+test('set tile LOD params for a non-existent source', () => new Promise<void>(done => {
```
HarelM, line 40: Can this promise and `done` be replaced by async method?

src/geo/projection/covering_tiles.test.ts
```diff
@@ -318,6 +318,46 @@ describe('coveringTiles', () => {
                 new OverscaledTileID(7, 0, 7, 73, 35),
                 new OverscaledTileID(7, 0, 7, 72, 36)
             ]);
+        
+            let optionsWithTileLodParams = {
```
HarelM, line 322: This should be split into multiple tests, it's already too long as it is...

src/ui/map.ts
```diff
@@ -2176,6 +2177,43 @@ export class Map extends Camera {
...
+     * @returns True is success, false if failure (for example if the ID
+     * corresponds to no existing sources).
+     * @example
+     * ```ts
+     * let success = map.setSourceTileLodParams(4.0, 3.0, 'terrain');
+     * ```
+     * @see [Modify LOD behavior](https://maplibre.org/maplibre-gl-js/docs/examples/lod-control/)
+
+     */
+    setSourceTileLodParams(maxZoomLevelsOnScreen: number, tileCountMaxMinRatio: number, id?: string) : boolean {
```
HarelM, line 2201: What will happen if there's a style reload or similar behavior? would the developer need to call this method again?
What will happen if this is called before the source is loaded?

src/ui/map.ts
```diff
@@ -2176,6 +2177,43 @@ export class Map extends Camera {
...
+        if (id) {
+            const source = this.getSource(id);
+            if(!source) {
+                return false;
+            }
+            source.calculateTileZoom = getCalculateTileZoomFunctionFromParams(Math.max(1, maxZoomLevelsOnScreen), Math.max(1, tileCountMaxMinRatio));
```
HarelM, line 2207: I think the fact that this is all what's needed in order to make this work is amazing, it means that the infrastructure around it is good.

src/geo/projection/covering_tiles.ts
```diff
@@ -127,6 +128,50 @@ function calculateTileZoom(requestedCenterZoom: number,
...
+    const dx = (x2-x1)/N;
+    for( let i = 0; i < N; i++)
+    {
+        const x = x1 + (i+0.5)/10 * (x2-x1);
+        sum += dx*Math.pow(Math.cos(x), p);
+    }
+    return sum;
+}
+
+export function getCalculateTileZoomFunctionFromParams(maxZoomLevelsOnScreen: number, tileCountMaxMinRatio: number): CalculateTileZoomFunction {
```
HarelM, line 143: I would expect this method to be called for "regular" use case with "default" parameters - this would mean that the current method is not repeated here and that the code is not duplicated.

HarelM, line 143: Also consider a different name, like `create...` or `...Factory`.

test/examples/lod-control.html
NathanMOlson: Done.

src/ui/map_tests/map_set_lod_params.test.ts
NathanMOlson: Done.

src/geo/projection/covering_tiles.ts
NathanMOlson: Done.

src/ui/map.ts
NathanMOlson: If this is called before the source is loaded, it will return `false`.

If the source is re-constructed, the developer would need to call this method again.

src/geo/projection/covering_tiles.test.ts
NathanMOlson: Done.

test/examples/lod-control.html
NathanMOlson: Done.

src/geo/projection/covering_tiles.ts
```diff
@@ -97,35 +98,52 @@ export function isTileVisible(frustum: Frustum, aabb: Aabb, plane?: vec4): Inter
...
+        // if distance to candidate tile is a tiny bit farther than distance to center,
+        // use the same zoom as the center. This is achieved by the scaling distance ratio by cos(fov/2)
+        thisTileDesiredZ = thisTileDesiredZ + scaleZoom(distanceToCenter3D / distanceToTile3D / Math.max(0.5, Math.cos(degreesToRadians(cameraVerticalFOV / 2))));
+        thisTileDesiredZ += pitchTileLoadingBehavior * scaleZoom(Math.cos(thisTilePitch)) / 2;
+        return thisTileDesiredZ;
+    };
 }
+const defaultMaxZoomLevelsOnScreen = 9.314;
+const defaultTileCountMaxMinRatio = 3.0;
+const calculateTileZoom = createCalculateTileZoomFunction(defaultMaxZoomLevelsOnScreen, defaultTileCountMaxMinRatio);
```
HarelM, line 146: `defaultCalculateTileZoom`?

src/ui/map.ts
```diff
@@ -2176,6 +2177,43 @@ export class Map extends Camera {
...
+     * pitch == 0, and the largest effect when the horizon is visible on screen.
+     *
+     * @param maxZoomLevelsOnScreen - The maximum number of distinct zoom levels allowed on screen at a time.
+     * There will generally be fewer zoom levels on the screen, the maximum can only be reached when the horizon
+     * is at the top of the screen. Increasing the maximum number of zoom levels causes the zoom level to decay 
+     * faster toward the horizon.
+     * @param tileCountMaxMinRatio - The ratio of the maximum number of tiles loaded (at high pitch) to the minimum
+     * number of tiles loaded. Increasing this ratio allows more tiles to be loaded at high pitch angles. If the ratio
+     * would otherwise be exceeded, the zoom level is reduced uniformly to keep the number of tiles within the limit.
+     * @param id - The ID of the source to set tile LOD parameters for. All sources will be updated if unspecified.
```
HarelM, line 2191: `id` -> `sourceId`?

src/ui/map.ts
```diff
@@ -2176,6 +2177,43 @@ export class Map extends Camera {
...
+     * @returns True is success, false if failure (for example if the ID
```
HarelM, line 2192: Should it throw instead of returning false? What is the behavior for other methods?

src/ui/map_tests/map_set_lod_params.test.ts
```diff
@@ -0,0 +1,48 @@
+import {beforeEach, test, expect} from 'vitest';
+import {createMap, beforeMapTest} from '../../util/test/util';
+
+beforeEach(() => {
+    beforeMapTest();
+    global.fetch = null;
+});
+
+test('set tile LOD params for a specific source', async () => {
+    createMap({interactive: false}, (err, map) => {
```
HarelM, line 10: can the tests use this method without the callback?
I would like to minimize the callback code in the tests and in general.
you can use `await map.on('load');` instead of using the callback.
I'll see if I can completely remove this callback from the util method.

HarelM:
Added a few minor comments, this looks good otherwise.

src/ui/map_tests/map_set_lod_params.test.ts
NathanMOlson: Done.

src/ui/map.ts
NathanMOlson: Some methods, throw, others return `undefined` or an empty array. Given that this method does not have a natural return value, I've updated it to throw.

src/ui/map.ts

src/geo/projection/covering_tiles.ts
NathanMOlson: Done.

NathanMOlson:
@HarelM Thanks for the feedback.
@JBLINZ Does the proposed functionality (as demonstrated here: [demo](https://nathanmolson.github.io/lod_control2)) meet your needs? Should I move forward with this parameter set despite the shortcomings?

JBLINZ:
@NathanMOlson The proposed functionality meets my needs. 

I'll admit I'm blindly using the slider until I get the behaviour performance/economy I'm after. 

For me (a non-dev spatial analyst type) to intuitively understand what I'm controlling, you would have to hide 
 maxZoomLevelsOnScreen & tileCountMaxMinRatio and have a General LOD control (Less Tiles >>>>>> More Tiles) which influenced these setting behind the scenes. 
 
The only other aspect that could neaten the output of these changes is moving away from the radis-controlled zoom level break points. This often results in lower-res tiles being brought in on the outer edges of the map. 
 
 Example:
 
chrome_nP4uDA01pu

From a visual perspective, I prefer to use a straight line from the perpendicular to camera-view direction.  Something like this:
image
or 
image


This is certainly not a new requirment. Just feedback.

developer-guides/covering-tiles.md
```diff
@@ -0,0 +1,105 @@
...
+
+$h$, `distanceToTileZ`, camera altitude above tile
+
+$\theta$, `thisTilePitch`, pitch angle from camera to tile
+
+$A$, area (map units)
+
+$Z$, tile zoom level
+
+$S$, tile scale factor (pixels per map unit): $ S = 2^Z$
```
HarelM, line 16:
```suggestion
$S$, tile scale factor (pixels per map unit): $S = 2^Z$
```

src/geo/projection/covering_tiles.ts
```diff
@@ -96,36 +97,63 @@ export function isTileVisible(frustum: Frustum, aabb: Aabb, plane?: vec4): Inter
...
 }
+/**
+ * Definite integral of cos(x)^p. The analytical solution is described in `developer-guides/covering-tiles.md`,
+ * but here the integral is evaluated numerically.
+ * @param p - the power to raise cos(x) to inside the itegral
+ * @param x1 - the starting point of the integral.
+ * @param x2 - the ending point of the integral.
+ * @return the integral of cos(x)^p from x=x1 to x=x2
+ */
+function intCosXToP(p: number, x1: number, x2: number): number {
```
HarelM, line 108: `integralOfCosXByP`?

src/ui/map_tests/map_set_lod_params.test.ts
```diff
@@ -0,0 +1,48 @@
...
+});
+
+test('set tile LOD params for a specific source', async () => {
+    const map = createMap({interactive: false});
+    await map.once('style.load');
+
+    map.addSource('source-id1', {type: 'raster', url: ''});
+    map.addSource('source-id2', {type: 'raster', url: ''});
+
+    expect(map.getSource('source-id1').calculateTileZoom).toBeFalsy();
```
HarelM, line 16: I don't think flasy and truthy should be used here, instead maybe `defined` and `undefined`?

src/geo/projection/covering_tiles.ts
```diff
@@ -96,36 +97,63 @@ export function isTileVisible(frustum: Frustum, aabb: Aabb, plane?: vec4): Inter
...
-    let thisTileDesiredZ = requestedCenterZoom;
-    const thisTilePitch = Math.atan(distanceToTile2D / distanceToTileZ);
-    const distanceToTile3D = Math.hypot(distanceToTile2D, distanceToTileZ);
-    // if distance to candidate tile is a tiny bit farther than distance to center,
-    // use the same zoom as the center. This is achieved by the scaling distance ratio by cos(fov/2)
-    thisTileDesiredZ = requestedCenterZoom + scaleZoom(distanceToCenter3D / distanceToTile3D / Math.max(0.5, Math.cos(degreesToRadians(cameraVerticalFOV / 2))));
-    thisTileDesiredZ += pitchTileLoadingBehavior * scaleZoom(Math.cos(thisTilePitch)) / 2;
-    thisTileDesiredZ = thisTileDesiredZ + clamp(requestedCenterZoom - thisTileDesiredZ, -tileZoomDeadband, tileZoomDeadband);
-    return thisTileDesiredZ;
+export function createCalculateTileZoomFunction(maxZoomLevelsOnScreen: number, tileCountMaxMinRatio: number): CalculateTileZoomFunction {
```
HarelM, line 121: Should this be exported as part of the public API? I'm not sure what I think, but I'll be happy to discuss this.

HarelM:
I think there's a problem with latest codecov:
https://github.com/codecov/codecov-action/issues/1806

Let's give them a few hours to fix this.

src/geo/projection/covering_tiles.ts
NathanMOlson: I'd like to hide this stuff from users as much as possible.

src/geo/projection/covering_tiles.ts
NathanMOlson: Done

src/geo/projection/covering_tiles.ts
HarelM: Fair enough.

src/ui/map_tests/map_set_lod_params.test.ts
NathanMOlson: Done.

HarelM (approved):

HarelM:
Are there any plans to address the comments raised by @JBLINZ before I merge this and release a new version?

NathanMOlson:
> Are there any plans to address the comments raised by @JBLINZ before I merge this and release a new version?

No, I don't plan to address that at this time.

JBLINZ:
> No, I don't plan to address that at this time.
Totally fine.

HarelM:
I'll merge this after I release 5.3.1, so that I can backtrack issues better.
Please let me know how urgent this is, and I can release version 5.4.0 if needed, otherwise I'll wait a bit to see if other stuff gets in before releasing a new version.

PR #5730, opened 2025-04-07
Cleanup GL pixelStore state after texture upload

kubapelc:
Cleans up glPixelStore state after texture updates in MapLibre.

This is global GL state that affects how texture uploads work (glTexSubImage2D, etc.). Not cleaning it would cause glTexSubImage2D calls made by the user to randomly act differently, depending on whether MapLibre did any texture updates that frame or not. Note that the cleanup is only necessary in texture.ts, changing pixelStore state during rendering of an actual frame is okay, since painter calls `context.setDefault();` at the very end, which resets it to default.

### DIFF
--- a/src/render/texture.ts
+++ b/src/render/texture.ts
@@ -78,6 +78,10 @@ export class Texture {
         if (this.useMipmap && this.isSizePowerOfTwo()) {
             gl.generateMipmap(gl.TEXTURE_2D);
         }
+
+        context.pixelStoreUnpackFlipY.setDefault();
+        context.pixelStoreUnpack.setDefault();
+        context.pixelStoreUnpackPremultiplyAlpha.setDefault();
     }
 
     bind(filter: TextureFilter, wrap: TextureWrap, minFilter?: TextureFilter | null) {


kubapelc:
I've added a test that tests specifically that the `Texture` constructor and by extension `update()` function reset the state properly.

I'm not aware of any issue for this, I've encountered this problem at work and it was hard to debug.

src/render/texture.test.ts
```diff
@@ -0,0 +1,33 @@
...
+            // We test the Texture constructor's side effects
+            // eslint-disable-next-line @typescript-eslint/no-unused-vars
+            const _texture = new Texture(context, testImage, gl.RGBA, {premultiply});
+            expect(context.pixelStoreUnpack.current).toEqual(context.pixelStoreUnpack.default);
+            expect(context.pixelStoreUnpackFlipY.current).toEqual(context.pixelStoreUnpackFlipY.default);
+            expect(context.pixelStoreUnpackPremultiplyAlpha.current).toEqual(context.pixelStoreUnpackPremultiplyAlpha.default);
+        }
+    
+        test('premultiply=false', () => {
+            testTextureCreation(false);
```
HarelM, line 26: I would recommend moving the `new Texture` part to each test to make it more explicit and avoid passing a Boolean to the method.
Even it is repeating some code in the tests.

HarelM (approved):

PR #5731, opened 2025-04-07
Update run_render_tests.ts to allow creating images when missing

### LINKED ISSUE #5714: Workflow is broken for creating a new render test
Problem:

A developer should be able to create a new render test according to the documentation at https://github.com/maplibre/maplibre-gl-js/tree/main/test/integration#writing-new-tests.

But as far as I can tell, it is not possible to follow these instructions due to bugginess in the code of `/test/integration/render/run_render_tests.ts`.

The problem is that the logic inside [`compareRenderResults`](https://github.com/maplibre/maplibre-gl-js/blob/main/test/integration/render/run_render_tests.ts#L123-L200) is faulty. It seems to expect the diff image to exist, which is not possible when creating a test anew.

I would expect that the code would be able to handle the case where only a `style.json` exists (when in `UPDATE` mode), but it cannot. Short of some other script to create a new set of expected images, this one needs fixed.

**maplibre-gl-js version**: latest (replicated today with main at 7305186be21)

**browser**: N/A

### Steps to Trigger Behavior

You can trigger/replicate by pretending an existing test is being created anew by clearing all the `png` images and then attempting to recreate them:

For example, for the `icon-offset/literal` test, first we confirm it passes normally:

```
› npm run test-render -- icon-offset/literal          

> maplibre-gl@5.3.0 test-render
> node --no-warnings --loader ts-node/esm test/integration/render/run_render_tests.ts icon-offset/literal

1/1: passed tests/icon-offset/literal
1 passed (100.0%)
```

Then:

 1. Make the test appear new: `rm test/integration/render/tests/icon-offset/literal/*png`
 2. Then - try running the test with `UPDATE=1` in order to create the expected images: 
 ```
› UPDATE=1 npm run test-render -- icon-offset/literal      

> maplibre-gl@5.3.0 test-render
> node --no-warnings --loader ts-node/esm test/integration/render/run_render_tests.ts icon-offset/literal

 1/1: errored tests/icon-offset/literal Cannot read properties of undefined (reading 'width') 
Updated 0/1 tests, 1 errored.
```

Note that the error of `Cannot read properties of undefined (reading 'width')` indicates that the test runner failed to run rather than that the test expectation failed. It would be reasonable here for the test expectation to fail (until images are created for a new test). But it is unexpected that the test runner would fail. Because the outcome of the test runner failing is that no images were able to be created and therefore the testcase is still not working, despite having a valid style.

This can be seen by doing:

```
› ls test/integration/render/tests/icon-offset/literal     
style.json

```

### Expected Behavior

Per above I would expect the test-runner to generate `expected.png`, `diff.png`, and `actual.png`. It does not.

A patch like this gets things working in a more expected way:

```diff
--- a/test/integration/render/run_render_tests.ts
+++ b/test/integration/render/run_render_tests.ts
@@ -182,9 +182,11 @@ function compareRenderResults(directory: string, testData: TestData, data: Uint8
         }
     }
 
-    const diffBuf = PNG.sync.write(minDiffImg, {filterType: 4});
-
-    fs.writeFileSync(diffPath, diffBuf);
+    if (minDiffImg) {
+        const diffBuf = PNG.sync.write(minDiffImg, {filterType: 4});
+        fs.writeFileSync(diffPath, diffBuf);            
+        testData.diff = diffBuf.toString('base64');
+    }
     fs.writeFileSync(actualPath, actualBuf);
 
     testData.difference = minDiff;
@@ -194,9 +196,9 @@ function compareRenderResults(directory: string, testData: TestData, data: Uint8
         console.log(`Updating ${expectedPath}`);
         fs.writeFileSync(expectedPath, PNG.sync.write(actualImg));
     }
-
-    testData.expected = minExpectedBuf.toString('base64');
-    testData.diff = diffBuf.toString('base64');
+    if (minExpectedBuf) {
+        testData.expected = minExpectedBuf.toString('base64');        
+    }
 }
```

### Actual Behavior

No expected images are generated due to the test runner erroring mid-way through its operation.

HarelM:
## Launch Checklist

- Fixes #5714

Moving some logic of the render test into an if to avoid harness failure in case of missing expected image (the case of adding a new render test by creating a folder and placing a style.json file only).

 - [x] Link to related issues.

### DIFF
--- a/test/integration/render/run_render_tests.ts
+++ b/test/integration/render/run_render_tests.ts
@@ -182,9 +182,12 @@ function compareRenderResults(directory: string, testData: TestData, data: Uint8
         }
     }
 
-    const diffBuf = PNG.sync.write(minDiffImg, {filterType: 4});
-
-    fs.writeFileSync(diffPath, diffBuf);
+    if (minDiffImg) {
+        const diffBuf = PNG.sync.write(minDiffImg, {filterType: 4});
+        fs.writeFileSync(diffPath, diffBuf);
+        testData.diff = diffBuf.toString('base64');
+        testData.expected = minExpectedBuf.toString('base64');
+    }
     fs.writeFileSync(actualPath, actualBuf);
 
     testData.difference = minDiff;
@@ -194,9 +197,6 @@ function compareRenderResults(directory: string, testData: TestData, data: Uint8
         console.log(`Updating ${expectedPath}`);
         fs.writeFileSync(expectedPath, PNG.sync.write(actualImg));
     }
-
-    testData.expected = minExpectedBuf.toString('base64');
-    testData.diff = diffBuf.toString('base64');
 }
 
 /**


louwers (approved):

louwers:
We have a separate flag that enables updating.

PR #5742, opened 2025-04-11
Hypsometric Tint from terrain-RGB tiles

### LINKED ISSUE #5666: Hypsometric Tint from terrain-RGB tiles
## User Story

As a map maker/owner, I would like to render a hypsometric tint using my terrain-RGB tiles in MapLibre to deliver high-quality cartographic products.

## Rationale

Currently, I cannot render anything other than a hillshade using my terrain-RGB tiles, although there are many common outputs from this type of elevation data that could be useful. 

By introducing hypsometric tint rendering from terrain RGB tiles, similar to GDAL’s “color-relief”, MapLibre can extend its usefulness for advanced cartography, visualisation and basic analysis. 

Ideally, users could display terrain tiles with a hypsometric tint or any other customised relief shading in MapLibre GL JS using the map style. The new terrain display algorithm could be similar to [color-relief](https://gdal.org/en/stable/programs/gdaldem.html#color-relief) in GDAL.

This command outputs a 3-band (RGB) or 4-band (RGBA) raster with values are computed from the elevation and a text-based color configuration file, containing the association between various elevation values and the corresponding wished color. By default, the colors between the given elevation values are blended smoothly and the result is a nice colorized DEM

Here is a basic example combined with a dynamic hill-shade:

Image

NathanMOlson:
Resolves #5666 (Hypsometric Tint from terrain-RGB tiles). Supporting style spec changes are here: https://github.com/NathanMOlson/maplibre-style-spec/tree/color_relief

Demo here: https://nathanmolson.github.io/color_relief

The shader uses a binary search to find the adjacent color stops from the color ramp, which means that larger color palettes take longer to render (O(log(n))).

Examples: 

cr_linz
cr_geo
cr_iso
cr_rainbow
cr_cat

 - [X] Link to related issues.
 - [x] Post benchmark scores.


NathanMOlson:
Benchmarks: ColorReliefLoad takes approximately the same amount of time as HillshadeLoad:

bench_hillshade
bench_cr

src/style/style_layer/color_relief_style_layer.ts
```diff
@@ -0,0 +1,45 @@
...
+    paint: PossiblyEvaluated<ColorReliefPaintProps, ColorReliefPaintPropsPossiblyEvaluated>;
+
+    constructor(layer: LayerSpecification) {
+        super(layer, properties);
+        this._updateColorRamp();
+    }
+
+    _updateColorRamp() {
+        const expression = this._transitionablePaint._values['color-relief-color'].value.expression;
+        if (expression instanceof ZoomConstantExpression && expression._styleExpression.expression instanceof Interpolate) {
```
NathanMOlson, line 26: What to do if the expression is not an `interpolate` expression` that uses `elevation`? Currently, we silently fail to display the layer. `HeatmapStyleLayer` has similar requirements, and does not seem to validate its input. Should we throw an error here in the constructor? Or display a default? Or something else?

HarelM:
Thanks for this!
I've just released a new version of the style spec, can you use it here so I can review the code with more ease (it has test failures comments).

src/style/style_layer/color_relief_style_layer.test.ts
```diff
@@ -0,0 +1,48 @@
+import {describe, test, expect} from 'vitest';
+import {ColorReliefStyleLayer} from './color_relief_style_layer';
+import {Color, type LayerSpecification} from '@maplibre/maplibre-gl-style-spec';
+import {createStyleLayer} from '../create_style_layer';
+import {extend} from '../../util/util';
+import {type EvaluationParameters} from '../evaluation_parameters';
+
+function createLayerSpec(properties?): LayerSpecification {
```
HarelM, line 8: I think the name here is a bit generic but the content is specific, consider changing the name to be more specific, also you can return the relevant type instead of casting outside of this method.

src/style/style_layer/color_relief_style_layer.ts
HarelM: It is possible not to check the expression type? I prefer to have this less specific and more future-proof. I think we should imitate how it is in the heatmap layer.

src/shaders/color_relief.fragment.glsl
HarelM, line 1: @mwilsnd any chance you can review the glsl files?

HarelM:
I've added some minor comments, looks good otherwise.
Let's wait for Marc's review and merge it afterwards.

mwilsnd (commented):
How many stops would you expect a typical use case to have? You started with a color ramp texture before moving to a uniform array, presumably this was for performance reasons?

The shaders look reasonable to me, though I'm wondering more about the limitations of a texture driven approach.

src/style/style_layer/color_relief_style_layer.ts
NathanMOlson: The current implementation uses the color ramp in GLSL. So rather than using the public `Expression` interface (like Heatmap does in `renderColorRamp()`), I'm extracting the elevation and color tables from the `Expression` internals and passing them to the shader.

`Heatmap` evaluates the color ramp onto a uniform grid (using `expression.evaluate()` and then passes the rendered color ramp to the shader as a texture. The shader can then do a simple, fast texture sample to get the appropriate color. This is fast and works well for smoothly-varying color ramps, and was my initial implementation for `color-relief`. However, it does not support sharp breaks in the color ramp, which is a LINZ requirement (see "LINZ" and "categorical" style examples in the [demo](https://nathanmolson.github.io/color_relief)).

The specific line that requires the expression to be an `Interpolate` is line 28,

```
this.elevationStops = interpolater.labels;
```

This is used to extract the elevation table out of the expression, so it can be used in the shader.

src/style/style_layer/color_relief_style_layer.test.ts
NathanMOlson: I've change the name to be more specific.

Re: returning the relevant type, I don't see a place where we're casting outside of this method. If you're thinking of 

```
const colorReliefStyleLayer = layer as ColorReliefStyleLayer;
```

that is casting the return value of `createStyleLayer()`, which is an existing, generic function with an established return type that I don't think we should change.

src/style/style_layer/color_relief_style_layer.ts
```diff
@@ -0,0 +1,45 @@
...
+            const interpolater = expression._styleExpression.expression;
+            this.elevationStops = [];
+            this.colorStops = [];
+            for (const label of interpolater.labels) {
+                this.elevationStops.push(label);
+                this.colorStops.push(interpolater.evaluate({globals: {elevation: label}} as any));
```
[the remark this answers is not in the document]
NathanMOlson: Changed to avoid the "as any". I believe the array types are already fully defined (line 13-14):

```
    elevationStops: Array<number>;
    colorStops: Array<Color>;
```

NathanMOlson:
> How many stops would you expect a typical use case to have? You started with a color ramp texture before moving to a uniform array, presumably this was for performance reasons?

I expect a typical use case to have between 2 and a few tens of stops. My examples are all between 3 and 30.

I didn't move away from the colorramp texture for performance reasons, but to support sharp color breaks (see "LINZ" and "categorical" style examples in the [demo](https://nathanmolson.github.io/color_relief)).

src/shaders/color_relief.fragment.glsl
```diff
@@ -0,0 +1,41 @@
+uniform sampler2D u_image;
+uniform vec4 u_unpack;
+uniform float u_elevation_stops[NUM_ELEVATION_STOPS];
```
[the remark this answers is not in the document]
NathanMOlson: The minimum required value for GL_MAX_FRAGMENT_UNIFORM_VECTORS is [16](https://registry.khronos.org/OpenGL-Refpages/es2.0/xhtml/glGet.xml), so it does seem wise to add an upper limit. 

16 is too low to be a generic limit, so I'm thinking of adding a check against GL_MAX_FRAGMENT_UNIFORM_VECTORS, and if the color ramp is too long, re-sample it and print a warning "Color relief layer may not render properly". This warning would only show up on low-end systems, so might not be evident to a developer who doesn't have access to such a system.

Also, the current implementation requires the colorramp length to be 1 + a power of 2, (e.g 2, 3, 5, 9, 17, 33, 65, ...). I'll see if I can rework the binary search to avoid this requirement (which would make the actual limit 9 on a low end system GL_MAX_FRAGMENT_UNIFORM_VECTORS=16).

src/render/draw_color_relief.ts
```diff
@@ -0,0 +1,97 @@
...
+    depthMode: Readonly<DepthMode>,
+    colorMode: Readonly<ColorMode>,
+    useBorder: boolean,
+    isRenderingToTexture: boolean
+) {
+    const projection = painter.style.projection;
+    const context = painter.context;
+    const transform = painter.transform;
+    const gl = context.gl;
+    const defines = [`#define NUM_ELEVATION_STOPS ${layer.elevationStops.length}`];
```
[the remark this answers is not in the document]
NathanMOlson: Currently yes, if the `color-relief-color` expression is not an `Interpolate`. I'd like to address that in the constructor so we don't get this far in that case.

src/shaders/color_relief.fragment.glsl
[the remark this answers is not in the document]
NathanMOlson: I've updated to remap the color ramp if the length exceeds GL_MAX_FRAGMENT_UNIFORM_VECTORS. Note this will change the appearance of non-smooth colorramps when it happens.

src/render/draw_color_relief.ts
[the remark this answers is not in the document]
NathanMOlson: I've modified the constructor to ensure that `layer.elevationStops.length` is always at least 2.

mwilsnd:
> I didn't move away from the colorramp texture for performance reasons, but to support sharp color breaks (see "LINZ" and "categorical" style examples in the [demo](https://nathanmolson.github.io/color_relief)).

Right, I see your newer reply to Harel from earlier explaining the additional requirement. Looking at the defined stops for the LINZ and categorical ramps, your decision makes sense.

NathanMOlson:
> The shaders look reasonable to me, though I'm wondering more about the limitations of a texture driven approach.

The current approach could swap the uniform array for a texture without any conceptual changes. The main reason I chose a uniform array instead of a texture is for code clarity (I use a `UniformColorArray` instead of a `Texture`, which is basically just a binary blob. I think this better conveys the code intent, both in TypeScript and in the shader.) I also expect it to perform better.

I do see one advantage to using a texture, which is that the minimum GL_MAX_TEXTURE_SIZE is 64, compared to the minimum GL_MAX_FRAGMENT_UNIFORM_VECTORS, which is 16. I'm not sure how likely we are to see such low values for GL_MAX_FRAGMENT_UNIFORM_VECTORS, all the hardware I have access to has GL_MAX_FRAGMENT_UNIFORM_VECTORS = 1024.

So it feels like a trade-off between "code clarity and performance in typical use-cases and environments", vs "correctness on very limited hardware or color ramps with a huge number of colors".

src/shaders/color_relief.fragment.glsl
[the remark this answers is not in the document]
mwilsnd: Yes, the limit is indeed rather low for WebGL. For WebGL 2, following the specification should defer to ES 3.0 which specifies at least 224 vectors. In practice, both WebGL and WebGL 2 report 1024 on my desktop machines (laptops) and 256 on Snapdragon 8 Gen 2 and Snapdragon 865 Android devices. It stands to reason devices which only support the original WebGL minimum are rather dated by this point but I've been (unpleasantly) surprised before by strange hardware configurations.

It may be possible to win back some uniform space for the low end by packing the colors into a 1D RGBA atlas and sending the stops as pairs of (elevation, color ramp U coordinate), trading uniform space for some extra texture samples.

mwilsnd:
> So it feels like a trade-off between "code clarity and performance in typical use-cases and environments", vs "correctness on very limited hardware or color ramps with a huge number of colors".

I agree, and also expect there are devices still in use which run closer to the lower bound of these limits. It may be worthwhile to examine mobile device hardware surveys and inspect the feature sets supported by devices in the longer tail end of active support. WebGL 2, featuring a much more useful limit, offers a decent filter on supported minimums, but I'd be wary of anything that still requires original WebGL.

src/style/style_layer/color_relief_style_layer.ts
```diff
@@ -0,0 +1,69 @@
...
+    _transitioningPaint: Transitioning<ColorReliefPaintProps>;
+    paint: PossiblyEvaluated<ColorReliefPaintProps, ColorReliefPaintPropsPossiblyEvaluated>;
+
+    constructor(layer: LayerSpecification) {
+        super(layer, properties);
+        this._updateColorRamp();
+    }
+
+    _updateColorRamp() {
+        this.elevationStops = [];
```
HarelM, line 25: Can these two arrays be removed from this method so that this method will only return the relevant arrays calculations? i.e. make this method a bit more pure...

NathanMOlson: Done, and I defined a new `ColorRamp` type to hold the arrays.

NathanMOlson:
Marked as draft while I create a version that uses textures for comparison.

src/style/style_layer/color_relief_style_layer.test.ts
```diff
@@ -0,0 +1,121 @@
+import {describe, test, expect} from 'vitest';
+import {ColorReliefStyleLayer} from './color_relief_style_layer';
+import {Color, type LayerSpecification} from '@maplibre/maplibre-gl-style-spec';
+import {createStyleLayer} from '../create_style_layer';
+import {extend} from '../../util/util';
+import {type EvaluationParameters} from '../evaluation_parameters';
+
+function createColorReliefLayerSpec(properties?): LayerSpecification {
```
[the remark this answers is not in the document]
NathanMOlson: I added a type. I couldn't get anything simpler to pass `typecheck`.

NathanMOlson:
I got a version working with textures (#5913). For clarity, you can see the differences here: https://github.com/NathanMOlson/maplibre-gl-js/pull/5.

Here are the benchmarks for the version using uniforms:
uniform256
uniform2


And using textures:
texture256
texture2

The version using uniforms scores 2% faster on a 2-color color ramp, and 13% faster on a 256-color color ramp

NathanMOlson:
The newest version that uses textures is even closer in performance: 
image
image

src/style/style_layer/color_relief_style_layer.ts
```diff
@@ -9,61 +9,62 @@ import {warnOnce} from '../../util/util';
...
     constructor(layer: LayerSpecification) {
         super(layer, properties);
-        this._updateColorRamp();
+        this.colorRamp = this._updateColorRamp();
     }
 
-    _updateColorRamp() {
-        this.elevationStops = [];
-        this.colorStops = [];
+    _updateColorRamp() : ColorRamp {
```
HarelM, line 25: `createColorRamp()`? this doesn't update anything, right?

src/style/style_layer/color_relief_style_layer.ts
```diff
@@ -9,61 +9,62 @@ import {warnOnce} from '../../util/util';
...
         }
+        return colorRamp;
     }
     
     // Get the color ramp, enforcing a maximum length for the vectors. This modifies the internal color ramp,
     // so that the remapping is only performed once.
     getColorRamp(maxLength: number) : {elevationStops: Array<number>; colorStops: Array<Color>} {
-        if (this.elevationStops.length > maxLength) {
-            const remapStepSize = (this.elevationStops.length - 1)/(maxLength - 1);
+        if (this.colorRamp.elevationStops.length > maxLength) {
```
HarelM, line 52: Can you use ts-docs style comments instead of `//`?

HarelM: Also, can we know the max length when initializing the color ramp or this is a runtime value related to the gl context so that we need to rebuild the color ramp?

If it's the second one, why not rebuild the color ramp only as part of this method and not build it in the constructor?

src/style/style_layer/color_relief_style_layer.ts
NathanMOlson: Done.

src/style/style_layer/color_relief_style_layer.ts
NathanMOlson: > Can you use ts-docs style comments instead of `//` for the method?

Done.

> Also, can we know the max length when initializing the color ramp or this is a runtime value related to the gl context so that we need to rebuild the color ramp?

This is a runtime value related to the gl context.

> If it's the second one, why not rebuild the color ramp only as part of this method and not build it in the constructor?

Done.

NathanMOlson:
Superceded by #5913.

HarelM:
I'm a bit lost with all the branches and PRs.
Please let me know which one to review and close the others.

NathanMOlson:
@HarelM please review #5913.

PR #5746, opened 2025-04-12
Terrain fill transparency tile border fix

### LINKED ISSUE #5658: 3D terrain with partially transparent vector landcover shows intersecting opaque stripes from version 5.0.0 onwards
**Description**
Starting with MapLibre GL JS version 5.0.0, I can see visually intersecting “stripes” when rendering a 3D terrain scene with semi‐transparent forest polygons from custom .pbf vector tiles. These stripes look more opaque or darker where polygons overlap, creating a crossed “banding” pattern. Reverting to version 4.7.1 eliminates the issue.

Image

**Steps to Reproduce**

1. Use a custom 3D terrain source (RGB tiles in .png format).
2. Use a custom raster background (tiles in .jpg format) – not necessary to reproduce bug.
3. Add a custom vector landcover layer (forest) with partial transparency on the fill. For example:
```
map.addLayer({
        'id': 'forests',
        'type': 'fill',
        'source-layer': 'landcover',
        'source': 'vector',
        'minzoom': 11,
        'filter': ['==', ['get', 'type'], 'forest'],
        "paint": {
            'fill-color': '#076b1b',
            'fill-opacity': 0.2
        }
    });
```

4. Tilt/rotate the map in 3D mode.
5. Observe crossing stripes or darker bands where the partially transparent polygons overlap.

**Expected Behavior**
Image

A smooth, semi‐transparent rendering of the forest polygons in 3D, without any noticeable intersecting stripes (version 4.7.1).

**Actual Behavior**

Image

Distinct intersecting stripes of darker or more opaque color appear where polygons or tile boundaries meet (version 5.2.0).

**Versions Affected**

MapLibre GL JS: 5.2.0 and 5.0.0

**Environment**

Browser: Chrome - version 134.0.6998.118

OS: Windows 11

Framework: Angular 19 (though likely not framework‐specific)

**Additional Context**

- There are none console errors related with this issue.
- This bug seems related to how partial transparency is handled in 3D mode in the newer rendering pipeline, because if transparency were off map renders correctly.

Image
- removing custom raster baselayer does not change behavior:

Image

- Rolling back to MapLibre GL JS 4.7.1 resolves the visual artifact completely.

If more information or a minimal reproduction repository is needed, I can provide it.

Thank you!
Lukasz Stempek
[sharpmap.eu](https://sharpmap.eu/)

kubapelc:
Fixes #5658. Adds a render test that fails prior to this fix.


kubapelc:
I have no idea why the render tests are failing. The draw_fill.ts change triggers that, but it should not affect anything other than terrain tile pre-rendering.

HarelM:
Wasn't the code that was removed had been added in the past to solve some terrain+globe issues?

HarelM:
Those are nice tests to indicate this code was needed to support these scenarios.
Check out the following html report from the CI run:
[results.html.txt](https://github.com/user-attachments/files/19742191/results.html.txt)

Here's a very nice example out of the above report:
image

HarelM:
Can you elaborate on how this fixes the issue a bit?

HarelM (approved):

kubapelc:
Vector tiles usually have a bit of geometry beyond their border, so that occasional tiny gaps between tiles are filled. When that geometry is transparent, drawing two tiles next to each other would fill some pixels twice, as seen in the linked issue.

Normally this is avoided by using stencil masks, where the target area for every tile is marked with a unique value in stencil, and when drawing the actual tile, only pixels with this specific stencil value set to pass the stencil test. (The stencil masks may extend a bit beyond the tile borders to fill the mentioned gaps, but still, any pixel will only have a single value in the stencil buffer and will be drawn over by a single tile.)

When terrain tiles are prerendered, this stencil mask wasn't used, probably by accident, causing the extra overlapping geometry to become visible. This PR re-enables stencil masks for terrain render-to-texture, and fixes some bugs related to using stencil with terrain when globe is active.

PR #5765, opened 2025-04-16
Fix queryRenderedFeatures for globe view

### LINKED ISSUE #5476: Empty array queryRenderedFeatures on globe projection
**maplibre-gl-js version**: 5.0.1

**browser**: Google Chrome

### Steps to Trigger Behavior

 1.   map.addSource('grid', {
      type: 'vector',
      tiles: [url],
      promoteId: 'gobject_id',
      maxzoom: 5,
    });
 2. map
      .addLayer({
        id: 'grid-layer',
        type: 'fill',
        source: 'grid',
        'source-layer': 'martin.test_function_mvt',
        paint: {
          'fill-color': [
            'case',
            ['boolean', ['feature-state', 'hover'], false],
            MapColors.objectHover,
            MapColors.object,
          ],
          'fill-opacity': 0.5,
          'fill-outline-color': '#000000',
        },
      })
 3.  map.on('mousemove', 'grid-layer', (e) => {
      if (e.features && e.features.length > 0) {
        const feature = e.features.sort((a, b) => area(a) - area(b))[0];
       
          handleHover(feature);
          map.setFeatureState(
            { source: 'grid', sourceLayer: 'martin.test_function_mvt', id: feature.id },
            { hover: true }
          );
        
      }
    });

4. function 
  const handleHover = (feature: maplibregl.MapGeoJSONFeature) => {
    const features = map.current
      ?.queryRenderedFeatures({
        layers: ['grid-layer'],
      })
      ?.filter((item) => item.state.hover === true);
      // i removed the rest of logic of this function
}

msbarry:
## Launch Checklist

Attempt to fix #5476 by switching to unwrapped tile coordinates when checking if a point is inside a rendered tile when using globe view. 

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/geo/projection/mercator_transform.ts
+++ b/src/geo/projection/mercator_transform.ts
@@ -198,6 +198,9 @@ export class MercatorTransform implements ITransform {
     get renderWorldCopies(): boolean {
         return this._helper.renderWorldCopies;
     }
+    get allowRenderWorldCopies(): boolean {
+        return this.renderWorldCopies;
+    }
     get cameraToCenterDistance(): number { 
         return this._helper.cameraToCenterDistance;
     }
@@ -209,9 +212,6 @@ export class MercatorTransform implements ITransform {
     }
     public get autoCalculateNearFarZ(): boolean { 
         return this._helper.autoCalculateNearFarZ; 
-    }
-    get isGlobeRendering(): boolean {
-        return false;
     }
     setTransitionState(_value: number, _error: number): void {
         // Do nothing
--- a/src/geo/projection/mercator_transform.ts
+++ b/src/geo/projection/mercator_transform.ts
@@ -210,6 +210,9 @@ export class MercatorTransform implements ITransform {
     public get autoCalculateNearFarZ(): boolean { 
         return this._helper.autoCalculateNearFarZ; 
     }
+    get isGlobeRendering(): boolean {
+        return false;
+    }
     setTransitionState(_value: number, _error: number): void {
         // Do nothing
     }
--- a/src/geo/projection/mercator_transform.ts
+++ b/src/geo/projection/mercator_transform.ts
@@ -198,9 +198,6 @@ export class MercatorTransform implements ITransform {
     get renderWorldCopies(): boolean {
         return this._helper.renderWorldCopies;
     }
-    get allowRenderWorldCopies(): boolean {
-        return this.renderWorldCopies;
-    }
     get cameraToCenterDistance(): number { 
         return this._helper.cameraToCenterDistance;
     }
--- a/src/source/source_cache.test.ts
+++ b/src/source/source_cache.test.ts
@@ -1577,7 +1577,7 @@ describe('SourceCache#tilesIn', () => {
         sourceCache.onAdd(undefined);
     }));
 
-    test('globe wrap', () => new Promise<void>(done => {
+    test('globe wrap', () => async () => {
         const transform = new GlobeTransform();
         transform.resize(512, 512);
         transform.setZoom(1.05);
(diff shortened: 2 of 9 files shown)


src/source/source_cache.ts
```diff
@@ -993,7 +993,7 @@ export class SourceCache extends Evented {
                 // Tiles held for fading are covered by tiles that are closer to ideal
                 continue;
             }
-            const tileID = tile.tileID;
+            const tileID = transform.isGlobeRendering ? tile.tileID.unwrapTo(0) : tile.tileID;
```
msbarry, line 996: The issue was that this tile was wrapped so the tileSpaceBounds check below was always failing and no tiles matched.

src/geo/projection/mercator_transform.ts
```diff
@@ -210,6 +210,9 @@ export class MercatorTransform implements ITransform {
     public get autoCalculateNearFarZ(): boolean { 
         return this._helper.autoCalculateNearFarZ; 
     }
+    get isGlobeRendering(): boolean {
```
HarelM, line 213: I would like to avoid this is possible as it is specific to globe, which is not relevant to mercator, and creates a weird interface.
I think you can use the projetion's transition state?

src/source/source_cache.ts
HarelM: The tile gets wrapped only for globe but not for mercator?

src/geo/projection/mercator_transform.ts
msbarry: We need some way to tell whether the globe is rendering. I also found trying to use the globe view that application need to be able to tell if globe is rendering through the external API.  How else can we determine that internally and externally?

src/source/source_cache.ts
msbarry: That's what it looks like. If I change this line to just `const tileID = tile.tileID.unwrapTo(0);` then the mercator projection test for querying a point on other side of IDL returns an empty array.

src/geo/projection/mercator_transform.ts
HarelM: What about using `allowRenderWorldCopies()` and `renderWorldCopies`? wouldn't that be more accurate for unwrapping to 0?

src/source/source_cache.ts
HarelM: I guess this is a lesser evil than the rest... :-/

src/geo/projection/mercator_transform.ts
msbarry: `transform.renderWorldCopies` seems to always reflect what the user has set with `map.setRenderWorldCopies(...)`.  I could change `VerticalPerspectiveTransform.renderWorldCopies` to return false always and `GlobeTransform.renderWorldCopies` to return `this.isGlobeRendering ? false : this._helper.renderWorldCopies` and use that instead of isGlobeRendering and these tests pass, but then these tests fail so I'm not sure if this is the expected behavior?

https://github.com/maplibre/maplibre-gl-js/blob/14b641b6c31f47b0854dc90cacd98c43b2089631/src/geo/projection/globe_transform.test.ts#L577-L593

On a side note, is there a way to determine if globe is rendering through external API?

msbarry: OK I switched isGlobeRendering to allowRenderWorldCopies. How does that look?

Also `getProjection` always returns globe, but I need to know when it switches between mercator and vertical perspective projection (currently around zoom 12?)

HarelM: No need to implement something new, see here:
https://github.com/maplibre/maplibre-gl-js/blob/14b641b6c31f47b0854dc90cacd98c43b2089631/src/geo/projection/covering_tiles.ts#L218

Sorry, took me a second to find it as well...

HarelM: I don't think you have a public API, but you can use the following float in the meantime:
`map.style?.projection?.transitionState`, globe is when this is `> 0`

msbarry: Ah nice, `transform.getCoveringTilesDetailsProvider().allowWorldCopies()` seems to work!

src/source/source_cache.test.ts
```diff
@@ -1576,6 +1576,119 @@ describe('SourceCache#tilesIn', () => {
         });
         sourceCache.onAdd(undefined);
     }));
+
+    test('globe wrap', () => new Promise<void>(done => {
```
HarelM, line 1580: Is it possible to move this promise definition to inside the test instead of here? 
Maybe wrap the on(data...) in a promise using a method and await that? 

I want to reduce the usage of a "top level promise" in the tests as much as possible.

HarelM:
Added a small comment, thanks!!

HarelM (approved):
Nice solution with the while-await!

Issue #2653, opened 2023-06-09
webgl1 support and future


zhangyiatmicrosoft:
## User Story

Closely related to #2644 
I'd like to suggest keep webgl1 support for much longer period of time

## Rationale
**Fact 1:**
With an one-day experiment on Bing users with MapLibre 3.0.1 release, the error rate of unable to create webgl context increased 9 times from 0.14% to 1.3%.
Put into context, right now there are total of 1.2 million user sessions per day, and a new feature is being released to bring 9x more sessions to MapLibre.

In short, 10 million sessions per day are expected and 1.3% of that --- about 140K sessions would fail if webgl1 was removed.

It appears that the hardware situation of end consumers is not ready for this limitation. Not to mention enterprise customers who tend to be even slower to upgrade.

**Fact 2**
Webgl (1 or 2) itself is sunsetting, sadly, and being superseded by webgpu. 
Sure the process can take many years, but the trend is irreversible. Therefore, the investments on webgl2 only features and optimizations might not get the best return after all

## Suggestion
1) In short term restore the support of webgl1
2) In long term, invest more on flexible render subsystems: webgl1 and 2 and maybe webgpu, so the "upper part" can be easily plugged into any combination of the other three.
It is perfectly aligned with our bounty direction https://github.com/maplibre/maplibre/issues/273
#2606 

Thanks

birkskyum:
Thank you for the writeup, and the real-life test data - it's very valuable as always.

Regarding this bullet:

> Sure the process can take many years, but the trend is irreversible. Therefore, the investments on webgl2 only features and optimizations might not get the best return after all

The short- and long-term benefits to investing in webgl2 is that we can refactor the code in ways that will make it faster short-term by i.e. adopting uniform buffer objects, and when WebGPU eventually arrives it will make adoption easier because it has a similar concept.

We will of course consider this inquiry carefully, and it will be on the TSC agenda on Wednesday 14th June, but keep in mind that there also is a prioritization of resources taking place here. Our budget currently allow for just one part-time maintainer, and it puts certain restraints on our ability to serve the marginal 1% of users while also pushing the project forward in ways that can benefit the other 99%. We're in the process of raising more funds to better the situation.

zhangyiatmicrosoft:
Thank you for the insights. I'd suggest investing more on https://github.com/maplibre/maplibre/issues/273 and it will be win-win for everybody --- so that it is flexible enough to swap back and forth.
[...]

PR #5768, opened 2025-04-17
Additional Hillshade Methods

### LINKED ISSUE #5665: Additional Hill-shade Options
## User Story

As a map maker/owner, I would like to have multiple options when choosing the hill-shade/relief-shade type that MapLibre renders using my terrain-RGB tiles so that I can deliver high-quality cartographic products. 

## Rationale

Currently, only a standard hill-shade is available which is a great general hillshade.  

By introducing three additional algorithms for rendering hill-shades from terrain RGB tiles, MapLibre will be increasingly useful for advanced cartography.  GDAL currently allows 'Combined', 'Multidirectional' and 'Igor', among others. Choosing from these three options when using terrain RGB tiles would allow map makers to select the right theme and feel for their project. 

Further explanation of hill-shade types:

- [Combined](https://gdal.org/en/stable/programs/gdaldem.html#cmdoption-combined) combined shading, a combination of slope and oblique shading.
- [Multidirectional](https://gdal.org/en/stable/programs/gdaldem.html#cmdoption-multidirectional) multidirectional shading, a combination of hillshading illuminated from 225 deg, 270 deg, 315 deg, and 360 deg azimuth. Multidirectional hill-shading applies the formula of http://pubs.usgs.gov/of/1992/of92-422/of92-422.pdf.
- [Igor](https://gdal.org/en/stable/programs/gdaldem.html#cmdoption-igor) shading, which tries to minimize effects on other map features beneath. Igor's hill-shading uses a formula from Maperitive http://maperitive.net/docs/Commands/GenerateReliefImageIgor.html.

Image

Image

Image

GDAL’s alternate hillshade algorithms — from top to bottom: multidirectional, combined, and igor. Images derived from the [USGS National Map](https://apps.nationalmap.gov/downloader/).

NathanMOlson:
- Resolves #5665

Top to bottom:

1. Standard Hillshade method (unchanged legacy behavior)
2. Basic hillshade (uses a point source and a simple physics-based model, similar to GDAL's default algorithm)
3. Combined (similar to GDAL `-combined` option`)
4. Igor (similar to GDAL `-igor` option`)
5. Multidirectional (applies basic hillshade from multiple independent light sources). The example image shows the use of different color illumination sources to color based on aspect.

image
image
image
image
image

Also adds `hillshade-illumination-altitude` control.

Incidentally adds support for passing `float` or `Color` arrays to shaders, and for compiling shaders with arbitrary `#define`s.

 - [X] Link to related issues.
 - [x] Post benchmark scores.


NathanMOlson:
Performance is unimpacted.
image

src/render/painter.ts
```diff
@@ -716,7 +716,7 @@ export class Painter {
      * False by default. Use true when drawing with a simple projection matrix is desired, eg. when drawing a fullscreen quad.
      * @returns
      */
-    useProgram(name: string, programConfiguration?: ProgramConfiguration | null, forceSimpleProjection: boolean = false): Program<any> {
+    useProgram(name: string, programConfiguration?: ProgramConfiguration | null, defines?: Array<string> | null, forceSimpleProjection: boolean = false): Program<any> {
```
HarelM, line 719: Should the defines be an empty array by default?

src/render/painter.ts
```diff
@@ -741,7 +742,8 @@ export class Painter {
                 this._showOverdrawInspector,
                 useTerrain,
                 projectionPrelude,
-                projectionDefine
+                projectionDefine,
```
HarelM, line 745: Is it a problem if there are defines and projection defines? Is there a need to unite them or is it better if this method will take care of it?

src/render/program/hillshade_program.ts
```diff
@@ -58,22 +66,30 @@ const hillshadeUniformValues = (
...
         'u_latrange': getTileLatRange(painter, tile.tileID),
-        'u_light': [layer.paint.get('hillshade-exaggeration'), azimuthal],
-        'u_shadow': shadow,
-        'u_highlight': highlight,
-        'u_accent': accent
+        'u_exaggeration': layer.paint.get('hillshade-exaggeration'),
+        'u_altitudes': illumination.altitudeRadians,
+        'u_azimuths': illumination.directionRadians,
+        'u_accent': accent,
+        'u_method': method == 'combined' ? 1 :
```
HarelM, line 87: Use switch case for the method.

test/integration/render/tests/terrain/fill-extrusion-multipolygon/expected.png
HarelM, line 1: These looks a bit darker, is it intentional?

test/examples/hillshade.html
```diff
@@ -0,0 +1,183 @@
...
+            layers: [
+                {
+                    id: 'background',
+                    type: 'background',
+                    paint: {
+                        "background-color": "#808080"
+                    }
+                }
+            ],
+            terrain: {
```
HarelM, line 157: Is terrain really needed for this example?

test/examples/hillshade.html
HarelM, line 1: Overall this example is good, but it might be a bit too complicated to read.
Maybe have separate pages for each method and have a bit more static (i.e no UI controls to change all the values but rather show how it looks so people can copy and modify)?
IDK...

src/style/style_layer/hillshade_style_layer.test.ts
```diff
@@ -0,0 +1,139 @@
...
+        let layer = createStyleLayer(layerSpec);
+        expect(layer).toBeInstanceOf(HillshadeStyleLayer);
+        
+        let illumination = (layer as HillshadeStyleLayer).getIlluminationProperties();
+        expect(illumination.directionRadians).toEqual([degreesToRadians(6), degreesToRadians(7)]);
+        expect(illumination.altitudeRadians).toEqual([degreesToRadians(23), degreesToRadians(23)]);
+        expect(illumination.highlightColor).toEqual([Color.red, Color.red]);
+        expect(illumination.shadowColor).toEqual([Color.black, Color.black]);
+
+        layerSpec = createLayerSpec({
```
HarelM, line 88: Consider splitting this to a different test.

src/style/style_layer/hillshade_style_layer.ts
```diff
@@ -15,6 +17,49 @@ export class HillshadeStyleLayer extends StyleLayer {
 
     constructor(layer: LayerSpecification) {
         super(layer, properties);
+        this.recalculate({zoom: 0, zoomHistory: {}} as EvaluationParameters, undefined);
+        if (this.paint.get('hillshade-illumination-direction').values.length < 1) {
+            throw new Error('"hillshade-illumination-direction" cannot be an empty array');
```
HarelM, line 22: Are you sure the constructor should validate and throw? Is this behavior common to other layers as well?

src/shaders/hillshade.fragment.glsl
```diff
@@ -2,30 +2,54 @@ uniform sampler2D u_image;
 in vec2 v_pos;
 
 uniform vec2 u_latrange;
-uniform vec2 u_light;
-uniform vec4 u_shadow;
-uniform vec4 u_highlight;
+uniform float u_exaggeration;
 uniform vec4 u_accent;
+uniform int u_method;
+uniform float u_altitudes[NUM_ILLUMINATION_SOURCES];
```
HarelM, line 8: Is there a different way to solve this besides a "define"? How come other programs do not need this?

src/shaders/hillshade.fragment.glsl
```diff
@@ -43,8 +67,119 @@ void main() {
...
+void main() {
+    vec4 pixel = texture(u_image, v_pos);
+
+    // We divide the slope by a scale factor based on the cosin of the pixel's approximate latitude
+    // to account for mercator projection distortion. see #4807 for details
+    float scaleFactor = cos(radians((u_latrange[0] - u_latrange[1]) * (1.0 - v_pos.y) + u_latrange[1]));
+
+    vec2 deriv = ((pixel.rg * 8.0) - 4.0) / scaleFactor;
+
+    if(u_method == STANDARD)
```
HarelM, line 159: Can we use switch or it's not part of the language?

src/shaders/hillshade.fragment.glsl
HarelM, line 1: There's not much input I can give here, I hope the math is correct...

test/examples/hillshade.html
NathanMOlson: No, removed.

test/integration/render/tests/terrain/fill-extrusion-multipolygon/expected.png
NathanMOlson: I believe this is due to the change I made to use the actual tile size in the derivative calculation. Previously, the assumed tile size was hardcoded to 512.

src/render/program/hillshade_program.ts
NathanMOlson: Done.

src/render/painter.ts
NathanMOlson: There is no problem with having `defines` and `projectionDefines`. We could merge them into a single array before passing them to the `Program` constructor. I don't have a preference.

src/shaders/hillshade.fragment.glsl
HarelM: As in frame buffer object?
How would this be changed, if at all, if we move to webgl2 (let's say in the not so near future)?

src/shaders/hillshade.fragment.glsl
HarelM: Can you link to the code in gdal for reference?

src/render/painter.ts
NathanMOlson: Done.

test/examples/hillshade.html
NathanMOlson: This example is more suited to let people explore the different styles, rather than figure out how to add a hillshade layer to their map.

I've added a simple, static example to demonstrate the mechanics of adding hillshade to the map, while leaving this example for exploration of the different styles.

HarelM: Isn't maputnik the right place for these kind of tweaking?

src/style/style_layer/hillshade_style_layer.ts
NathanMOlson: I've created a [PR](https://github.com/maplibre/maplibre-style-spec/pull/1094) for the style spec if we want to go that direction. It seems like a good solution to me. If we merge that, we can remove the validation from the constructor.

NathanMOlson:
@JBLINZ This is almost ready to go. Can you review and confirm this meets your requirements? The latest demo is [here](https://nathanmolson.github.io/hillshade).

src/shaders/hillshade.fragment.glsl
NathanMOlson: Done.

test/examples/hillshade.html
NathanMOlson: Having this in the docs is easier to access than using maputnik, you don't have to hook up your own data sources (or know how to use maputnik). I modeled this example after the existing Sky, Fog, Terrain example.

Shall I remove this example?

src/shaders/hillshade.fragment.glsl
NathanMOlson: I don't think I would use a frame buffer object, I would use a texture like `terrain.getCoordsTexture()` does.

I don't expect WebGL2 would require any changes, but I'm not an expert.

To me, passing these values as uniforms makes it a lot more clear what's going on, because they maintain their types instead of being put into a raw buffer.

test/examples/hillshade.html
HarelM: Yes, I created the train/fog/sky example, and looking back I'm not sure it's a great example.
I don't mind that you can play with these, it's fine, but my main concern is that it's a bit long to read so as a reader it's hard to understand what is the core code of the example.

So I would suggest to just create a few static examples and the reader would be able to copy paste them to jsbin and play with the settings there, if they wouldn't want to use maputnik.

In general, you'll need to play with it and your data anyway to use it in production, so the main idea of the page is to tell users this exists, basically.

src/style/style_layer/hillshade_style_layer.ts
HarelM: Style spec is a better fit to this code I believe. Thanks!

src/shaders/hillshade.fragment.glsl
HarelM: Fair enough.

src/render/painter.ts
HarelM: We'll probably need to pass them if they will be part of the "other" defines, so let's keep it this way.
If we decide in the future to improve it, we can always do it...

test/integration/render/tests/terrain/fill-extrusion-multipolygon/expected.png
HarelM: In any case, since this is somewhat of a rendering change it needs to be emphasized in the changelog.
I think the default for raster-dem tileSize is 256, isn't it? Weird that this was hardcoded to 512...

src/style/style_layer/hillshade_style_layer.ts
NathanMOlson: I've removed the validation in this constructer, to be performed in the style spec by either [PR 1094](https://github.com/maplibre/maplibre-style-spec/pull/1094) or [PR 1095](https://github.com/maplibre/maplibre-style-spec/pull/1095)

NathanMOlson: Validation is now performed in the style spec (https://github.com/maplibre/maplibre-style-spec/pull/1095)

test/integration/render/tests/terrain/fill-extrusion-multipolygon/expected.png
NathanMOlson: > So in theory, if we were using 512 raster-dem tiles, the tests results would not have change?

Not quite. These are the only two render tests that _were_ using 512x512 hillshade tiles. I updated the constant (1.25 => 0.625) in hillshade.fragment.glsl line 49 so that the render tests using 256x256 tiles remained unchanged. The hillshade in these two render tests was very light and is now darker, so I think the behavior has "improved".

This change will cause the appearance of maps using 512x512 terrain tiles for hillshade to change. Maps using 256x256 tiles will not change.

After this change, switching between 256x256 and 512x512 hillshade tiles will not change the appearance of the map.



> In any case, since this is somewhat of a rendering change it needs to be emphasized in the changelog. I think the default for raster-dem tileSize is 256, isn't it? Weird that this was hardcoded to 512...

No, the default rester-dem `tileSize` is 512. Agreed that this is a rendering change that will potentially impact users, I'll highlight it in the changelog.

test/examples/hillshade.html
NathanMOlson: I've added a `multidirectional` example (which the the only method conceptually different from the others) and removed the style exploration example.

src/style/style_layer/hillshade_style_layer.ts
```diff
@@ -15,6 +17,36 @@ export class HillshadeStyleLayer extends StyleLayer {
...
+    }
+
+    getIlluminationProperties(): {directionRadians: number[]; altitudeRadians: number[]; shadowColor: Color[]; highlightColor: Color[]} {
+        const direction = this.paint.get('hillshade-illumination-direction').values;
+        const altitude = this.paint.get('hillshade-illumination-altitude').values;
+        const highlightColor = this.paint.get('hillshade-highlight-color').values;
+        const shadowColor = this.paint.get('hillshade-shadow-color').values;
+        // ensure all illumination properties have the same length
+        const numIlluminationSources = Math.max(direction.length, altitude.length, highlightColor.length, shadowColor.length);
+        for (let i = direction.length; i < numIlluminationSources; i++) {
```
HarelM, line 30: I would export this small for into a method as it's a bit hard to understand what it does just by reading this for.
And also, you are doing it a few times...

src/style/style_layer/hillshade_style_layer.ts
```diff
@@ -15,6 +17,36 @@ export class HillshadeStyleLayer extends StyleLayer {
...
+        }
+        for (let i = highlightColor.length; i < numIlluminationSources; i++) {
+            highlightColor.push(highlightColor[i-1]);
+        }
+        for (let i = shadowColor.length; i < numIlluminationSources; i++) {
+            shadowColor.push(shadowColor[i-1]);
+        }
+        const altitudeRadians = [];
+        const directionRadians = [];
+        for (let i = 0; i < numIlluminationSources; i++) {
```
HarelM, line 44: I think using `altitudeRadians.map(...)` would be more readable.

HarelM:
Added some minor nit-picking, otherwise, looks good.

src/style/style_layer/hillshade_style_layer.ts
NathanMOlson: Done.

src/style/style_layer/hillshade_style_layer.ts
NathanMOlson: Done.

HarelM (approved):

test/integration/render/tests/terrain/fill-extrusion-multipolygon/expected.png
HarelM: I'll merge this and release, let's see if anyone notice the change and have any issues.
The change is very subtle anyway...

wiesehahn:
While I like your graphic I am wondering why you use different light colors for the multidirectional hillshade? 
I think this is a rare use case and might confuse people not familiar with hillshade methods. I would rather use gdal defaults which uses same light color with multiple directions and altitudes. In this case the image would be comparable between all methods. Or am I seeing this wrong?

Regarding the general hillshade documentation I can't find any information how the hillshade is added/merged with the underlying data!?

wiesehahn:
Maybe you are right. I would tend to make it comparable in the documentation and provide different maps with different options as examples. 

In the image it looks like all options give a different shading for areas with no hillshade (e.g. open sea), is this correct? I guess we don't want to darken the entire viewport if there are no mountains.
Not entirely sure but I think with gdal defaults there should be no or only minimal differences if we apply them on a plane.

Plantain:
Since upgrading to maplibre-gl 5.5.0, we see a lot of errors like below in our sentry logs, mostly on older browsers:
Could not compile fragment shader: ERROR: 0:27: 'switch' : Illegal use of reserved word
ERROR: 0:27: 'switch' : syntax error

The backtrace is not very good but it suggests hillshade might be responsible.
aint.get("hillshade-highlight-color").values.length}`],m=e.useProgram("hillshade",null,!1,p),f=!e.options.moving;for(const p of r){const r=t {snip}

I suspect this change is likely the problem? Does this break with WebGL1 perhaps?

HarelM:
If you need to support old browsers I suggest to keep the maplibre version pinned. 
We do no plan to keep support for old browsers as we are aiming to add new features which require new browser capabilities (for example drop webgl 1 in favor of supporting the new webgpu capabilities).

Plantain:
Currently I am just speculating it is WebGL1 (or old-browser related) until I am able to find a live reproduction case. I now have 5.4.0 pinned for the meantime which makes it hard for me to continue observing in the wild.

However, I'm concerned by the policy to not support 'old' browsers. What is the threshold of 'old' browsers? 10%? 5%? 2%? I would suggest any non-hobby project building on maplibre-gl-js is a non-starter at >1%. What was the point in keeping WebGL1 in https://github.com/maplibre/maplibre-gl-js/pull/5198 if breaking WebGL1 support a few months later is now fine?

HarelM:
Actually, webgl 1 support drop was attempted two years ago and not a few month back as part of version 3.0.0 breaking changes.
Back then, there were some reports about issues with adoption and there was not a great motivation to really drop support for webgl 1.
This is no longer the case as we want to move to using luma.gl in the near future in order to better position ourselves for the future and allow better compatibility with deck.gl, and also start leveraging webgl2 capabilities.
Having said that, since you have a requirement to support old browsers (we are talking about 10 years old I believe) you are welcome to invest the needed time and resources to support this via either a branch, a fork, or however you see fit, but this is not part of the maintenance effort we are planning to invest in due to the reasons I listed above.

samuel-git:
Thanks for the awesome work on the additional hillshade methods and HareIM for maintaining this project.
I fear for larger sites using MaplibreGL dropping WebGL 1 support is not feasible yet as WebGL 2 became available in Safari [not earlier than September 2021](https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext).
However, pinning to 5.4.0 is good advice as we do not want to hinder development and understand the need to keep moving forward. Would be interesting to get some up to date stats from @zhangyiatmicrosoft for [WebGL 2 support](https://github.com/maplibre/maplibre-gl-js/issues/2653).

Plantain:
I opened https://github.com/maplibre/maplibre-gl-js/issues/5970 as after pinning 5.4.0 we are still seeing these in the wild.

Issue #4781, opened 2024-09-30
Globe allows zooming out further than minZoom


russss:
Firstly, thanks for all your hard work on the globe feature, I've just given it a go and it looks great!

I don't generate vector tiles for zoom levels below 2, as it's slow and there's not really any reason to zoom out that far (this is still the case with the globe view IMO). So I create the Map object with `minZoom: 2`.

With globe mode enabled and `minZoom: 2`:
* center the map over mid latitudes
* zoom all the way out to level 2
* pan north or south

The zoom level now goes below 2 and my map data disappears.

You can see this happening on [OpenInfraMap](https://openinframap.org/#2/26/12).

HarelM:
This is discussed here: 
https://osmus.slack.com/archives/C03TFH5NE83/p1727690088635139?thread_ts=1727690088.635139&cid=C03TFH5NE83

There are links there to comments here in GitHub. 
It's currently a limitation of how the globe work...

russss:
Aha, I had even read [the page](https://github.com/maplibre/maplibre-gl-js/blob/main/developer-guides/globe.md#zoom-behavior) about this but I clearly didn't remember it!

I think maplibre should probably emit a warning if `minZoom` is set at the same time as the globe projection is enabled. And this should be mentioned in the docs for `minZoom`.

From my perspective, I can probably generate these lower-zoom tiles now there's a good reason for them.

russss:
> I think maplibre should probably emit a warning if `minZoom` is set at the same time as the globe projection is enabled. And this should be mentioned in the docs for `minZoom`.

I fixed my tile server to return tiles at lower zoom levels, and I initially considered keeping `minZoom: 2` because it restricts you from zooming unnecessarily far out. But I decided that the way this interacts with panning is also a bit confusing, so in the end I scrapped the `minZoom`.

I also realised that if you have zoom-dependent layer visibility, then panning around the globe will also cause these layers to change visibility, which is slightly unexpected from the user's perspective.

I think I would ideally like to have a zoom parameter which acts more like altitude, at least at zoom levels where maplibre is in globe rather than mercator projection mode. OpenInfraMap does benefit from being able to exchange zoom levels with classic Web Mercator OSM maps, but only at higher zoom levels.

I think these are all minor non-blocker issues, but I just wanted to write my thoughts down somewhere.

HarelM:
[...]

Issue #5367, opened 2025-01-16
Wheel scrolling inconsistent, fix suggestion


Zaczero:
**maplibre-gl-js version**: 5.0.1

**browser**: Chromium

### Steps to Trigger Behavior

https://github.com/maplibre/maplibre-gl-js/blob/761a634e62630356ac57c38fd1ae69092ccc62d6/src/ui/handler/scroll_zoom.ts#L202

This line assumes that if two initial scrolls happen in quick succession, they indicate a trackpad. This leads to frequent false positives, especially when using a free-spinning mouse wheel (like the MX Master 3 series). I also encounter this issue on a regular mouse if I scroll quickly enough.

I believe a good solution would be to maintain a short history of deltaY values and check if they are all equal. The idea is similar to using `const wheelZoomDelta = 4.000244140625;`, but the problem is that it depends on the system's sensitivity settings.

![Image](https://github.com/user-attachments/assets/6a4b79a0-184c-4435-94e7-f98d5f5a387a)

Perhaps assume "wheel" by default and, when deltaY changes in a way that isn’t simply multiplication (for example, `109.09090672642736` and `-218.18181345285473` are effectively equal, notice small floating-point errors), switch to “trackpad” until a new scroll action begins. This would require storing just 1 variable (`minDeltaY = min(minDeltaY, abs(e.deltaY))`) instead of working with arrays.

### Expected Behavior

Wheel scrolling consistent

### Actual Behavior

Wheel scrolling inconsistent

HarelM:
I think this is a continue of the following issue:
- https://github.com/maplibre/maplibre-gl-js/issues/4362

I think we should continue the discussion there with all the relevant people involved.
Let me know if I should reopen this ticket or not based on what you read there.

Zaczero:
I saw that issue earlier and I don't think I have choppiness/lagging issues. The issue is about wrong scroll type detection causing scroll to sometimes be "normal" and sometimes to be super quick due to trackpad scaling being used incorrectly.

Zaczero:
How about the already proposed fix:
> Perhaps assume "wheel" by default and, when deltaY changes in a way that isn’t simply multiplication (for example, 109.09090672642736 and -218.18181345285473 are effectively equal, notice small floating-point errors), switch to “trackpad” until a new scroll action begins. This would require storing just 1 variable (minDeltaY = min(minDeltaY, abs(e.deltaY) / 2)) instead of working with arrays.

[...]

PR #5154, opened 2024-12-04
Fix smooth mouse wheel zooming 

Al-4SW:
Fixes #4362 

This PR addresses issues with smooth mouse scroll zooming.

### Issue: Stutter/'Lag' on zooms with multiple scroll wheel events

**Fix:** The condition for applying easing to 'wheel' events has been reverted back to v4.5.0. 

This fixes the issue where zoom events with multiple wheel events sometimes bypass the easing block in renderFrame() on the first renderFrame of subsequent wheelEvents (wheelEventTimeDiff===0), causing the stutter effect.

### Issue:  Jitter and Lag for fast 'free scrolling' wheel events (Particularly Chromium based browsers when DevTools is closed)

**Fix:** Wheel events with 'wheelEventTimeDiff' values of 0 are now used to calculate easing, with a new minimum 'minWheelEventTimeDiff' value applied.

This fixes the issue with small zoom level change caused by fast free scrolling wheels.  By normalising the 'wheelEventTimeDiff'  value used to calculate easing, we achieve smooth and responsive fast  zooms in Chrome, emulating the behaviour when DevTools is open  (see below).

### DevTools open / closed in Chrome exhibits different fast scroll performance

Test results below indicate the cause for this behaviour. The number of very fast (<1ms) 'lastWheelEventTimeDiff ' values goes from around 80% of all events when DevTools are closed, to 7-8% when open. When open, around 60% of all values are at least 5ms, and none are 0. 

 4.5.0 Smooth Zoom Tests (For Benchmark Comparison)
Screenshot from 2024-12-04 16-19-06

  5.0.0pre8 Smooth Zoom Tests
Screenshot from 2024-12-04 16-55-26

  Fix Smooth Zoom Tests
Screenshot from 2024-12-04 16-23-30

 Before - Multiple Stepped Wheel Scrolls
https://github.com/user-attachments/assets/87303797-cde0-4494-87fb-ce1b4e25ad5c

 After - Multiple Stepped Wheel Scrolls
https://github.com/user-attachments/assets/ade3ffcf-5b0d-4e9e-afea-5b79d3dfc29b

 Before - Free Scrolling Wheel
https://github.com/user-attachments/assets/8ae39794-814c-456f-91f5-fc26b478b91d

 After - Free Scrolling Wheel
https://github.com/user-attachments/assets/a8cc186c-269e-447b-a068-9f2fae432c67

[...]

PR #5775, opened 2025-04-18
Fix scroll min zoom on globe view

msbarry:
## Launch Checklist

Zoom can end up less than minZoom when using the globe view at high/low latitudes in order to keep the globe a consistent size (see #4781). But, currently when you scroll-zoom in globe view when zoom is less than minZoom, it snaps back to the min zoom:

https://github.com/user-attachments/assets/f98bc3a6-bc87-4b29-a5dc-3b5fa77a736f

This change makes scroll zooming keep the globe a consistent size no matter what latitude you are looking at:

https://github.com/user-attachments/assets/d0cec3c7-6000-469c-ab53-a5d5540651fd

 - [x] Link to related issues.


src/ui/handler/scroll_zoom.test.ts
```diff
@@ -553,6 +558,7 @@ describe('ScrollZoomHandler', () => {
     });
 
     test('Terrain 3D zoom is in the same direction when pointing above horizon or under horizon', () => {
+        vi.useFakeTimers();
```
msbarry, line 561: I noticed that this test kicked off an async task that failed during any subsequent test because `map.terrain` was missing a lot of fields, so use fake timers so setTimeout calls don't get run.

HarelM: Is there a different solution to this issue? I'd like to avoid these "hacks" and just make sure the test finishes and waits for the timers to finish...

src/ui/handler/scroll_zoom.ts
```diff
@@ -297,7 +297,7 @@ export class ScrollZoomHandler implements Handler {
             }
 
             const fromScale = typeof this._targetZoom !== 'number' ? tr.scale : zoomScale(this._targetZoom);
-            this._targetZoom = Math.min(tr.maxZoom, Math.max(tr.minZoom, scaleZoom(fromScale * scale)));
+            this._targetZoom = tr.getConstrained(tr.getCameraLngLat(), scaleZoom(fromScale * scale)).zoom;
```
HarelM, line 300: There are multiple bugs related to scroll behavior with different type of mouse devices, is there a way to make sure this doesn't break these?

src/ui/handler/scroll_zoom.test.ts
msbarry: Looks like this test had 2 issues:

1. terrain didn't have enough fields on it so the callback was failing. I found a `createTerrain()` test helper, if I switch to that instead of `map.terrain = { pointCoordinate: () => null } as any;`
2. It creates 2 maps, and only calls map.remove() on the second one. I added a map.remove() call to the first one

(1) might not be necessary given the fix in (2) but it seems cleaner to just use the shared test helper?

src/ui/handler/scroll_zoom.ts
msbarry: Looking through the `getConstrained` code in vertical perspective and mercator transform it looks like this should be fairly isolated from the different mouse device-specific handling in scroll zoom.

For [vertical perspective transform](https://github.com/maplibre/maplibre-gl-js/blob/c13bc9eb17c2a91fb198e705bd3b427d9d302a83/src/geo/projection/vertical_perspective_transform.ts#L639-L651) it does the same clamp operation, except adjusted based on camera latitude (I tested with different pitch/bearing/roll and it appears to work correctly.

On [mercator transform](https://github.com/maplibre/maplibre-gl-js/blob/c13bc9eb17c2a91fb198e705bd3b427d9d302a83/src/geo/projection/mercator_transform.ts#L447) it's just a clamp between min/max zoom, except it also adjusts the zoom based on lat/lng range.

I added another test scrolling out in globe view scrolling out using the magic mouse wheel  deltaY value too to be safe.

src/ui/handler/scroll_zoom.test.ts
```diff
@@ -20,6 +20,11 @@ beforeEach(() => {
     beforeMapTest();
 });
 
+afterEach(() => {
```
HarelM, line 23: Is this still needed after all the changes?

src/ui/handler/scroll_zoom.ts
HarelM: I don't know, here are some related links:
- #5367
- #4362
- #5154

Not sure if there's a good way to see if it changes the behavior there.
In any case, this looks good.

src/ui/handler/scroll_zoom.test.ts
msbarry: I don't need clearMocks, but I do still need useRealTimers. Without it, `await map.once('load');` never resolves. I'm guessing there's a setTimeout in there that never gets run since setTimeout is still mocked.

src/ui/handler/scroll_zoom.ts
msbarry: Those issues look pretty tricky... It would be nice if there was some way to record the events that get fired on some particular hardware setup so others can play them back to reproduce the behavior. I think the solution would be in this same file, but I think this change should be orthogonal.

HarelM: That's a nice idea, shouldn't be hard to achieve it. I'll keep that in mind for future reports.

src/ui/handler/scroll_zoom.test.ts
HarelM: But this is in the "afterEach"...
Also there's beforeMapTest in the before which should take care of most of the relevant mocking. 
There are a lot of tests that use await once('load').
Am I missing anything?

msbarry: It looks like this is the only test file that uses fake timers and await map.once: https://github.com/search?q=repo%3Amaplibre%2Fmaplibre-gl-js+map.once+AND+vi.useFakeTimers+path%3A**%2F*.test.ts&type=code

There's a vitest `clearMocks: true` option we could add to the config to automatically clear mocks between tests, but as far as I can tell there is no config option to tell it to automatically call useRealTimers between tests. The closest I could find was to move `vi.useRealTimers()` call into `beforeMapTest`

HarelM: ahh, this is what I was missing, can we avoid the usage of using fake timers then instead?

HarelM: Oh, this is not part of the changes you made... Hmm... Annoying...
If this is the case, I would consider setting fake timers in the beforeeach method and remove them in the aftereach method. this way we can keep the symmetry of the tests with regards to fake timers.
But in general, I would like to remove the need for it, since this is not needed in any other place in the tests...

msbarry: Looks like useFakeTimers was just to force the zoom/zoomEnd event to fire. I replaced them all with `await map.once('zoom/zoomend')` and they seem to work fine now.

HarelM: Nice!!! Thanks!

HarelM (approved):

msbarry:
Thanks!! Timing is not urgent, I have a fork with a few extra changes on it so I can just base it off of this commit.

PR #5798, opened 2025-04-23
Fix expiry data logic

### LINKED ISSUE #4837: Tile Expiry Handling Issue: cacheControl takes precedence over expires in raster_dem_tile_source.ts and raster_tile_source.ts
### Description:
In the current implementation of `raster_dem_tile_source.ts`, tile expiration is only handled if both `cacheControl` and `expires` are present. However, in cases where `cacheControl` is an empty string, the condition is not met, causing tiles not to be refreshed even when `expires` is set.

### Proposed Solution:
Update the condition as follows to handle tile expiry when either `cacheControl` or `expires` is present:

https://github.com/maplibre/maplibre-gl-js/blob/v4.7.1/src/source/raster_dem_tile_source.ts#L70-L72
https://github.com/maplibre/maplibre-gl-js/blob/v4.7.1/src/source/raster_tile_source.ts#L185-L187

if `cacheControl` value is emptyString `''` or `undefined` or `null` if state is always false. even if `expires` is vilified value

```typescript
// before
if (this.map._refreshExpiredTiles && response.cacheControl && response.expires) {
    tile.setExpiryData({cacheControl: response.cacheControl, expires: response.expires});
}
```
```typescript
// after
if (this.map._refreshExpiredTiles && (response.cacheControl || response.expires)) {
    tile.setExpiryData({cacheControl: response.cacheControl, expires: response.expires});
}
```

NathanMOlson:
Fixes #4837.


HarelM:
Should there be tests to cover these scenarios?

HarelM (approved):

src/source/raster_dem_tile_source.test.ts
```diff
@@ -156,4 +157,66 @@ describe('RasterDEMTileSource', () => {
...
+            actor: 1
+        } as any as Tile;
+        const expiryDataSpy = vi.spyOn(tile, 'setExpiryData');
+        const tilePromise = source.loadTile(tile);
+        server.respond();
+        await tilePromise;
+        expect(expiryDataSpy).toHaveBeenCalledTimes(1);
+    });
+
+    test('Tile expiry data is set when "Expires" is set but not "Cache-Control"', async () => {
```
HarelM, line 192: Can you add a test with empty string cache control as described in the linked issue?

PR #5799, opened 2025-04-23
Performance: Use 'If-Modified-Since' header when re-requesting expired tiles

NathanMOlson:
When a tile expires, it is re-requested. Before this PR, it is always transferred over the network. After this PR, the `If-Modified-Since` header is set in the re-request, so the server can send a [304 Not Modified](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/304) response if the tile has not been modified, instead of resending the tile.

This is primarily intended for use with tiles that have a very short expiration time (<10 seconds). This could be useful when the underlying data is changing rapidly.

 - [ ] Link to related issues.
 - [X] Implement for raster tile source.
 - [ ] Extend to all tile sources (vector, raster-dem).


HarelM:
Overall, this is a good idea. 
The complexity of data flow and handling of different response types is what you should be looking out for though (for example 404 overzooming, etc...).

HarelM:
@NathanMOlson are you planning to make this PR ready for review or should I close it?

NathanMOlson:
I am planning to make this ready at some point, but I am not actively working on it at the moment.

HarelM:
If this becomes relevant, do let me know and I'll reopen the PR, closing for now to reduce clutter.

PR #5805, opened 2025-04-24
Fix corrupted `docs/assets/cog.tif` file

oscarfonts:
Git interpreted the binary file as text  per `.gitattributes`, and corrupted it trying to normalize CR & LF.

Thus the [COG example](https://maplibre.org/maplibre-gl-js/docs/examples/add-cog/) is broken.

Declared *.tif files as binary and re-committed the sample file.


HarelM:
Thanks!

HarelM (approved):

PR #5806, opened 2025-04-25
Add refreshTiles() to map public API

### LINKED ISSUE #415: A reliable way to reload tiles from a vector tile source
## Existing Behavior

The current behavior seems to support this, but it is buggy for some reason.

https://github.com/maplibre/maplibre-gl-js/blob/9e6d57aaf7f77161c3d2f9dfd330a16227b60f19/src/source/vector_tile_source.ts#L136-L137

The above causes the layers that rely on the source to flash, and then reload. However, if this action (clearing the tiles & loading) is performed in quick succession (Even a couple seconds apart), the tiles do not actually reload at all. Why this is, I'm not sure, is there non-synchronous code running that takes it's time to actually finish up?

**Expected Solution Behavior:**

* Map layers do NOT flash
* Tiles are reliably re-fetched when action is performed over-and-over

**Observed Current Behavior:**

* Map layers flash
* If action is performed in quick succession, layers do not reload, _but they do still flash_
   *  There are no browser cache hits during this period.

## Motivation

If you need to update a vector tile source, you need a way to force the tiles to be re-fetched. 

Currently this can be hacked around by manually clearing tiles from a source's tilecache, but even that's buggy & unreliable.

For instance [this PR ](https://github.com/mapbox/mapbox-gl-js/commit/ed27d232d9c8f31013e01e11ae0f4e6c98607184) added some ability there, but it still doesn't entirely work.

**Use Cases:**
* Dynamic data
* User actions change map data
* Map data change events are pushed to the client (Client needs to reload tiles)

## Design Alternatives

Unknown?

## Design

Unknown?

### Mock-Up

**User Point of view:**
* `map.getSource('mySource').reload()`

### Concepts

Unknown

### Implementation

Unknown

### LINKED ISSUE #3419: Map Source Refresh
## User Story

 As a user I can delete a geometry in one of my layers of the map source via SQL command and force the maplibre GL JS map to refresh source, so that I can immediately see the deleted geometry object disappeared from my map.

## Rationale

 - Why is this feature needed?
 currently I need to wait for my martin server to reload the source so that the deleted object disappears. I have to manually force the refresh by moving around the map.
 - What are some example use cases?
 immediate update after insertion, deletion etc.
 - Can it be done today? Is there a workaround?
I searched ChatGPT, it tells me to reset my source of the map to the old source, but that would mean I have to reset all my layers for the source too, which is troublesome.

## Impact

Better UX for all users who need smooth experience with map apps built on Maplibre

NathanMOlson:
This PR adds `refreshTiles()` to the map public API.

- resolves #3419 
- resolves #415 

This allows tiles to be force-reloaded from outside MapLibre. The use case I've got in mind is that when tiles on the server change, a WebSocket message is sent to the client (using a to-be-developed plugin) notifying it of changed tiles. The plugin can then call `map.refreshTiles()` with the modified tile IDs, causing the map to reload the tiles and display the new data.

You can check out a demo [here](https://nathanmolson.github.io/live_tiles). Clicking on a tile will call `map.refreshTiles()` on the tile. I'm using `addProtocol()` to draw a timestamp (from `performance.now()`) on each tile, representing the time it was loaded. Zoom all the way out to see that it behaves correctly for wrapped tiles (all tiles sharing an x/y/z are updated when a tile is clicked). Zoom past z3 to see that it behaves correctly for overscaled tiles.

Marginally related to #5799, which is also intended to help with the use case where server-side data is changing rapidly.

 - [X] Link to related issues.

### DIFF
--- a/src/source/source_cache.ts
+++ b/src/source/source_cache.ts
@@ -19,7 +19,7 @@ import type {Style} from '../style/style';
 import type {Dispatcher} from '../util/dispatcher';
 import type {IReadonlyTransform, ITransform} from '../geo/transform_interface';
 import type {TileState} from './tile';
-import type {SourceSpecification} from '@maplibre/maplibre-gl-style-spec';
+import type {ICanonicalTileID, SourceSpecification} from '@maplibre/maplibre-gl-style-spec';
 import type {MapSourceDataEvent} from '../ui/events';
 import type {Terrain} from '../render/terrain';
 import type {CanvasSourceSpecification} from './canvas_source';
@@ -891,6 +891,13 @@ export class SourceCache extends Evented {
                 this._reloadTile(id, 'expired');
                 delete this._timers[id];
             }, expiryTimeout);
+        }
+    }
+
+    refreshTile(tileID: ICanonicalTileID)
+    {
+        if(this._isIdRenderable(tileID.key)) {
+            this._reloadTile(tileID.key, 'expired');
         }
     }
 
--- a/src/source/source_cache.ts
+++ b/src/source/source_cache.ts
@@ -900,7 +900,7 @@ export class SourceCache extends Evented {
     refreshTiles(tileIds: Array<ICanonicalTileID>) {
         for (const id in this._tiles) {
             if (this._isIdRenderable(id)){
-                for (let refreshId of tileIds) {
+                for (const refreshId of tileIds) {
                     if (refreshId.equals(this._tiles[id].tileID.canonical)) {
                         this._reloadTile(id, 'expired');
                     }
--- a/src/source/source_cache.ts
+++ b/src/source/source_cache.ts
@@ -894,6 +894,9 @@ export class SourceCache extends Evented {
         }
     }
 
+    /**
+     * Reload any currently renderable tiles that are match one of the incoming `tileId` x/y/z
+     */
     refreshTiles(tileIds: Array<ICanonicalTileID>) {
         for (const id in this._tiles) {
             if (this._isIdRenderable(id)){
--- a/src/source/source_cache.ts
+++ b/src/source/source_cache.ts
@@ -894,10 +894,15 @@ export class SourceCache extends Evented {
         }
     }
 
-    refreshTile(tileID: ICanonicalTileID)
-    {
-        if(this._isIdRenderable(tileID.key)) {
-            this._reloadTile(tileID.key, 'expired');
+    refreshTiles(tileIds: Array<ICanonicalTileID>) {
+        for (const id in this._tiles) {
+            if (this._isIdRenderable(id)){
+                for (let refreshId of tileIds) {
+                    if (refreshId.equals(this._tiles[id].tileID.canonical)) {
+                        this._reloadTile(id, 'expired');
+                    }
+                }
+            }
         }
     }
 
--- a/src/source/source_cache.ts
+++ b/src/source/source_cache.ts
@@ -902,11 +902,8 @@ export class SourceCache extends Evented {
             if (!this._isIdRenderable(id)) {
                 continue;
             }
-            for (const refreshId of tileIds) {
-                if (refreshId.equals(this._tiles[id].tileID.canonical)) {
-                    this._reloadTile(id, 'expired');
-                    break;
-                }
+            if (tileIds.some(tid => tid.equals(this._tiles[id].tileID.canonical))) {
+                this._reloadTile(id, 'expired');
             }
         }
     }
--- a/src/source/source_cache.ts
+++ b/src/source/source_cache.ts
@@ -899,11 +899,13 @@ export class SourceCache extends Evented {
      */
     refreshTiles(tileIds: Array<ICanonicalTileID>) {
         for (const id in this._tiles) {
-            if (this._isIdRenderable(id)){
-                for (const refreshId of tileIds) {
-                    if (refreshId.equals(this._tiles[id].tileID.canonical)) {
-                        this._reloadTile(id, 'expired');
-                    }
+            if (!this._isIdRenderable(id)) {
+                continue;
+            }
+            for (const refreshId of tileIds) {
+                if (refreshId.equals(this._tiles[id].tileID.canonical)) {
+                    this._reloadTile(id, 'expired');
+                    break;
                 }
             }
         }
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -2217,19 +2217,23 @@ export class Map extends Camera {
     /**
      * Triggers a reload of the selected tiles
      *
-     * @param tileIds - An array of tile IDs to be reloaded
      * @param sourceId - The ID of the source
+     * @param tileIds - An array of tile IDs to be reloaded. If not defined, all tiles will be reloaded.
      * @example
      * ```ts
-     * map.refreshTiles([{x:1024, y: 1023, z: 11}, {x:1023, y: 1023, z: 11}], 'satellite');
+     * map.refreshTiles('satellite', [{x:1024, y: 1023, z: 11}, {x:1023, y: 1023, z: 11}]);
      * ```
      */
-    refreshTiles(tileIds: Array<{x: number; y: number; z: number}>, sourceId: string) {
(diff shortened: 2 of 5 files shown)


src/ui/map.ts
```diff
@@ -2213,6 +2214,39 @@ export class Map extends Camera {
...
+     * @param x - The tile X coordinate
+     * @param y - The tile Y coordinate
+     * @param z - The tile Z coordinate
+     * @param sourceId - The ID of the source
+     * @example
+     * ```ts
+     * map.refreshTile(1024, 1023, 11, 'satellite');
+     * ```
+     */
+    refreshTile(x: number, y: number, z: number, sourceId: string) {
```
HarelM, line 2229: I think this is not needed given that you can pass sn array of one tile.

src/ui/map.ts
```diff
@@ -2213,6 +2214,39 @@ export class Map extends Camera {
...
+     * Triggers a reload of the selected tiles
+     *
+     * @param tileIds - An array of tile IDs to be reloaded
+     * @param sourceId - The ID of the source
+     * @example
+     * ```ts
+     * map.refreshTiles([{x:1024, y: 1023, z: 11}, {x:1023, y: 1023, z: 11}], 'satellite');
+     * ```
+     */
+    refreshTiles(tileIds: Array<{x: number; y: number; z: number}>, sourceId: string) {
```
HarelM, line 2242: Is there a need to refresh all the tiles for a given source - a refresh all kind of call? 

Also I would consider swapping the order if we want to support that by making the tiles array optional.

src/ui/map_tests/map_refresh_tiles.test.ts
```diff
@@ -0,0 +1,40 @@
...
+    await map.once('style.load');
+
+    map.addSource('source-id1', {type: 'raster', url: ''});
+    const spy = vi.fn();
+    map.style.sourceCaches['source-id1'].refreshTiles = spy;
+
+    expect(() => {map.refreshTiles([{x: 1024, y: 1023, z: 11}], 'source-id2');})
+        .toThrow('There is no source cache with ID "source-id2", cannot refresh tile');
+    expect(spy).toHaveBeenCalledTimes(0);
+    map.refreshTiles([{x: 1024, y: 1023, z: 11}], 'source-id1');
```
HarelM, line 21: This should be in a new test, shouldn't it?

HarelM:
This PR solves the following feature request, doesn't it?
- #3419

src/ui/map.ts
NathanMOlson: Removed.

src/ui/map_tests/map_refresh_tiles.test.ts
NathanMOlson: I split this test.

src/source/source_cache.ts
```diff
@@ -894,6 +894,21 @@ export class SourceCache extends Evented {
         }
     }
 
+    /**
+     * Reload any currently renderable tiles that are match one of the incoming `tileId` x/y/z
+     */
+    refreshTiles(tileIds: Array<ICanonicalTileID>) {
```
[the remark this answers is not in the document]
NathanMOlson: Non-existent source is handled separately with a `throw`.

Refresh requests for tiles that are not currently renderable are silently ignored.

src/ui/map.ts

NathanMOlson:
> This PR solves the following feature request, doesn't it?
> 
> * [Map Source Refresh #3419](https://github.com/maplibre/maplibre-gl-js/issues/3419)

I believe so, now that I've added a source-global refresh option.

HarelM:
There is a report about flashing tiles here: #415, does this PR solves the requested feature?

NathanMOlson:
> There is a report about flashing tiles here: #415, does this PR solves the requested feature?

I'm not sure. I've only tested with raster tiles. With raster tiles, I don't see any flashing and the refresh happens immediately.

In my implementation, `map.refreshTiles(sourceId)` is equivalent to `map.style.sourceCaches[sourceId].reload()` when `sourceId` is a valid source ID.

src/source/source_cache.ts
```diff
@@ -894,6 +894,21 @@ export class SourceCache extends Evented {
...
+        for (const id in this._tiles) {
```
HarelM, line 901: Is it possible to do some `.filter(...)` to the array to remove some indentation in this method?

src/ui/map_tests/map_refresh_tiles.test.ts
```diff
@@ -0,0 +1,46 @@
+import {beforeEach, test, expect, vi} from 'vitest';
+import {createMap, beforeMapTest} from '../../util/test/util';
+import {CanonicalTileID} from '../../source/tile_id';
+
+beforeEach(() => {
```
HarelM, line 5: It might be that other files are missing that too, but I think there should be a top level `describe` here, shouldn't it?

HarelM:
Added a minor nit-picking.
How confident are you with adding this functionality with all the reload flow happening in maplibre?
Or maybe the question is why wasn't this API introduced earlier if the implementation is so simple?

src/source/source_cache.ts
NathanMOlson: I removed some indentation without using `filter`.

src/ui/map_tests/map_refresh_tiles.test.ts
NathanMOlson: Done.

NathanMOlson:
> How confident are you with adding this functionality with all the reload flow happening in maplibre?

I'm confident that this works in my use-case, which is raster tiles. I'm not confident that it works for vector tiles or geojson sources.

HarelM:
In geojson you can simply update the data and it will reload I believe, for vector I hope it will work as expected. 
In any case, the public API is ok even if we decide the underlying implantation.

src/source/source_cache.ts
```diff
@@ -894,6 +894,23 @@ export class SourceCache extends Evented {
...
+            if (!this._isIdRenderable(id)) {
+                continue;
+            }
+            for (const refreshId of tileIds) {
```
HarelM, line 905: Wouldn't `if (tileIds.find(tid => tid.equals(this._tiles[id].tileID.canonical)) != null)` achieve the following lines?
Or am I misunderstanding what this `for` is doing?

NathanMOlson: Updated to remove for loop.

HarelM (approved):

russss:
Nice! I might give it a test with vector tiles...

etnav:
I tried this for both raster and vector tiles. Raster worked fine, however for vector tiles I couldn't make it work. When ever I changed the source, I called `map.refreshTiles()` with the source `id`, this updated all tiles as described by @NathanMOlson in the issue above for raster, but not vector. The only solution right now for me was to reload the page.

NathanMOlson:
I've reproduced this issue with vector tiles. Note that `map.refreshTiles()` works with vector tiles when you specify a list of tiles, but does not work when you do not specify a list of tiles.

The difference is: when you specify a list of tiles, `_reloadTile(id, 'expired')` gets called on each tile. When you do not specify a list of tiles, `reload()` is called, which calls `_reloadTile(id, 'reloading')` on each tile. If `'reloading`' is replaced with `'expired'` in `reload()`, `refreshTiles()` works with vector tiles as intended.

 I haven't figured out what the difference between `'reloading'` and `'expired'` is, or what sort of ripples might happen by changing `reload()` to use `'expired'`.

You can reproduce the issue by entering `map.refreshTiles('maplibre');` in the console [here](https://jsbin.com/luquhuvuke/edit?html,console,output).

NathanMOlson:
@etnav See #5875.

etnav:
> @etnav See #5875.
> 
> Until this is fixed, you can reload the whole layer by calling
> 
> ```
> map.refreshTiles(sourceId, map.style.sourceCaches[sourceId].getRenderableIds().map(id => map.style.sourceCaches[sourceId]._tiles[id].tileID.canonical));
> ```

Thanks for your effort. This approach works fine for now and I will also give your fix a try as well :)

PR #5811, opened 2025-04-25
Fix initial position of pointer-tracked popup

matthiasschwarz:
Fixes the initial position of a pointer-tracked popup.
Currently the pointer has no position when added to the map, which results in it being displayed in the top left corner until the mouse is moved.
As initial position is now used the latest mouse position.

## Launch Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Fix scroll min zoom on globe view ([#5775](https://github.com/maplibre/maplibre-gl-js/pull/5775))
 - ⚠️ Fix hillshade appearance change between 256x256 and 512x512 tiles. This will change the appearance of hillshade layers using 512x512 tiles. ([#5768](https://github.com/maplibre/maplibre-gl-js/pull/5768))
 - Fix tile expiry logic for raster and raster-dem tiles ([#5798](https://github.com/maplibre/maplibre-gl-js/pull/5798))
+- Fix initial position of pointer-tracked popup ([#5811](https://github.com/maplibre/maplibre-gl-js/pull/5811))
 - _...Add new stuff here..._
 
 ## 5.4.0
--- a/src/ui/handler_manager.ts
+++ b/src/ui/handler_manager.ts
@@ -150,6 +150,7 @@ export class HandlerManager {
         passive?: boolean;
         capture?: boolean;
     } | undefined]>;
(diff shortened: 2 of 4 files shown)


HarelM:
I would like to avoid updating handler manger for this, if possible. 
Can you think if a solution that can be encapsulated within the popup class? 

Also, can you share a video of the issue?

matthiasschwarz:
Since there is no intended way to determine the current mouse position without using the mouse events. There is no solution that only encapsulates the popup class. The current mouse position must be cached in the map so that it is available when the popup is added to the map. An instance of a popup could be created during a mouse event. At this point, it would be too late to wait for the next mouse event.

Would you prefer tracking the mouse position in the map instance via an event listener?

https://github.com/user-attachments/assets/e0b75252-d33e-44f8-ad80-ba571ec40c41

[Codesandbox]( https://codesandbox.io/p/devbox/intelligent-silence-pstl6g?workspaceId=ws_JFAgBUarjb6Lr5u9USgb4W)

HarelM:
but you are controlling when to add it to the map, why not set the initial location before adding it to the map?
I'm not following on the problem statement... What am I missing?

matthiasschwarz:
I found a workaround to provide an initial location. Calling `trackPointer` after adding the popup to the map allows to use an initial location.

```js
popup
  .setLngLat(e.lngLat)
  .addTo(map)
  .trackPointer();
```
[Codesandbox](https://codesandbox.io/p/devbox/stoic-bush-st756d?workspaceId=ws_JFAgBUarjb6Lr5u9USgb4W)

There should probably an example about the usage especially since the [documentation](https://maplibre.org/maplibre-gl-js/docs/API/classes/Popup/#trackpointer) shows calling `trackPointer` before adding the popup to the map.

HarelM:
To my understanding, track pointer should be used on the entire map and less as a hover effect on a marker.
Moreover, the concept of the popup changing position when the mouse moves over the maker is a very odd one from my point of view and a weird user experience.
I would consider simply adding a popup when mouse over and set the location of the marker to be a "staric" one. 
But I might not be familiar with the relevant use case. 
In any case, you are welcome to contribute a PR that updates the docs.

johncarmack1984:
The public API answer is in the thread: setLngLat, addTo, then trackPointer. The handler manager change was the part not wanted. I believe this one is closable.

PR #5812, opened 2025-04-25
Implement data-driven styling for line-dasharray

### LINKED ISSUE #1235: Implement data-driven styling for line-dash-array
I would like to assign different values to "line-dash-array" property based on the actual data.
Example:
`
'line-dasharray': [
      "case", 
      ['==', ['get', "my_prop"], 'my_value'],
      ['literal', [0,0]],
      ['literal', [2,3]]
]
`

This feature has already been added to Mapbox GL JS >= 2.3.0.
I know that there is a workaround (i.e. duplicate the layer and use a filter) but it would be nice to have this feature also in maplibre.

lucaswoj:
See also https://github.com/maplibre/maplibre-style-spec/pull/1100 

This PR resolves #1235 by making the `line-dasharray` property "data-driven", meaning it can now reference feature property values in expressions.

Due to limits in the expression system, dasharrays must be specified in the style using the `["literal", [...]]` syntax. Feature properties cannot directly contain array values.

Example:
```json
"paint": {
  "line-dasharray": [
    "case",
    ["==", ["get", "road_type"], "trail"], ["literal", [1, 1]],
    ["==", ["get", "road_type"], "highway"], ["literal", [4, 2]],
    ["literal", [2, 2]]
  ]
}
```

## Benchmarks

```
$ npm run benchmark

> maplibre-gl@5.7.1 benchmark
> node --no-warnings --loader ts-node/esm test/bench/run-benchmarks.ts

Starting headless chrome at: http://localhost:9966/test/bench/versions/index.html
                                                     v5.7.1                           main  data-driven-dasharray 9778a15 
                        Paint                    4.8350 ms                      4.7819 ms                      4.8240 ms 
                   QueryPoint                  156.4884 ms                    154.7769 ms                    156.3121 ms 
                     QueryBox                  104.8300 ms                    104.8764 ms                    105.0493 ms 
                       Layout                   74.0651 ms                     74.0231 ms                     74.1744 ms 
                    Placement                    3.3104 ms                      3.4451 ms                      3.3704 ms 
                     Validate                    0.7733 ms                      0.7800 ms                      0.7660 ms 
             StyleLayerCreate                    0.7400 ms                      0.7240 ms                      0.7211 ms 
               FunctionCreate                    0.0117 ms                      0.0125 ms                      0.0125 ms 
             FunctionEvaluate                    0.0020 ms                      0.0020 ms                      0.0021 ms 
             ExpressionCreate                    0.1495 ms                      0.1479 ms                      0.1490 ms 
           ExpressionEvaluate                    0.0088 ms                      0.0088 ms                      0.0088 ms 
               WorkerTransfer                    3.9750 ms                      3.8574 ms                      4.1199 ms 
                  PaintStates                   32.9472 ms                     32.6724 ms                     32.5642 ms 
          PropertyLevelRemove                   31.9636 ms                     31.9730 ms                     31.9314 ms 
           FeatureLevelRemove                   31.8360 ms                     31.8835 ms                     31.9515 ms 
            SourceLevelRemove                   31.9842 ms                     31.9564 ms                     31.9443 ms 
              LayerBackground                    0.5388 ms                      0.5388 ms                      0.5415 ms 
                  LayerCircle                    0.0838 ms                      0.0884 ms                      0.0893 ms 
                    LayerFill                    0.1350 ms                      0.1415 ms                      0.1418 ms 
           LayerFillExtrusion                    0.0916 ms⚠️                     0.0981 ms⚠️                     0.0991 ms 
                 LayerHeatmap                    0.5595 ms☠️                     0.5651 ms☠️                     0.5520 ms 
               LayerHillshade                    0.0689 ms                      0.0691 ms                      0.0726 ms 
      LayerColorRelief2Colors                    5.5683 ms☠️                     5.9341 ms                      6.2106 ms 
    LayerColorRelief256Colors                    6.2341 ms                      6.0844 ms                      5.5985 ms 
                    LayerLine                    0.0849 ms                      0.0919 ms⚠️                     0.0886 ms 
                  LayerRaster                    0.6773 ms                      0.6917 ms                      0.6740 ms 
                  LayerSymbol                    0.4103 ms                      0.4109 ms⚠️                     0.4143 ms⚠️
         LayerSymbolWithIcons                    0.4243 ms⚠️                     0.4256 ms⚠️                     0.4213 ms⚠️
  LayerTextWithVariableAnchor                    7.1703 ms☠️                     7.2015 ms                      7.2197 ms 
       LayerSymbolWithSortKey                    0.0535 ms⚠️                     0.0524 ms⚠️                     0.0518 ms⚠️
                         Load                   16.8075 ms☠️                    16.6567 ms⚠️                    16.7000 ms 
          LoadMatchingFeature                    0.0004 ms                      0.0004 ms                      0.0004 ms 
                 SymbolLayout                    4.0526 ms                      4.0474 ms                      4.0546 ms 
                 FilterCreate                    0.1777 ms                      0.1752 ms                      0.1752 ms 
               FilterEvaluate                    0.2753 ms                      0.2789 ms                      0.2780 ms 
                HillshadeLoad                  320.4750 ms☠️                   318.5486 ms                    323.5206 ms 
              ColorReliefLoad                  324.0647 ms                    324.0647 ms                    324.4353 ms 
                  CustomLayer                    0.0088 ms                      0.0088 ms                      0.0089 ms 
                      MapIdle                  322.4229 ms                    324.3457 ms                    324.0026 ms 
           SymbolCollisionBox                    3.6000 ms                      3.5871 ms⚠️                     3.5723 ms 
      SymbolCollisionBoxGlobe                   13.1642 ms                     13.2435 ms                     13.2263 ms 
                    Subdivide                   14.7933 ms                     14.6061 ms                     14.6397 ms 
           CoveringTilesGlobe                   10.1184 ms                     10.1171 ms                     10.4978 ms 
    CoveringTilesGlobePitched                   15.1567 ms                     14.7786 ms                     14.7753 ms 
        CoveringTilesMercator                    2.6828 ms                      2.6613 ms                      2.7006 ms 
 CoveringTilesMercatorPitched                    5.0028 ms                      5.0668 ms                      5.0054 ms 
```

See full results at [all.pdf](https://github.com/user-attachments/files/22372311/all.pdf)

### Phase 1: Basic Functionality

 - [x] Update `line_sdf.vertex.glsl` and `line_sdf.fragment.glsl`
 - [x] Update `line_program.ts`
 - [x] Update `program_configuration.ts`
    - [x] Register `line-dasharray` and its associated `varying`s under `attributeNameExceptions`
    - [x] Modify `CrossFadedConstantBinder` for line-dasharray support
    - [x] Modify `ProgramConfiguration` for line-dasharray support
 - [x] Update `style/properties.ts`
    - Review `CrossFadedDataDrivenProperty` to ensure support for data-driven `line-dasharray`
 - [x] Update `draw_line.ts`
    - [x] Add a call to `programConfiguration.setConstantDasharray` for `line-dasharray`
 - [x] Link to related issues.
 - [x] Write integration tests for all new functionality.
 - [x] Figure out how to use draft style spec version in CI 
 - [x] Get tests for constant line-dasharray passing again
 - [x] Double check understanding of how data is stored line line-atlas
 - [x] Trace codepath for line-cap property functions, determine how it should interact with line-dasharray
 - [x] Trace codepath for line-pattern property-functions, determine which parts of it can work for line-dash array
 - [x] Try to replace `lineDasharrayAttributes` with `patternAttributes` as support for the latter is better throughout the codebase 
 - [x] Ensure we are preserving fromScale and toScale functionality https://github.com/maplibre/maplibre-gl-js/pull/5812/files#r2064445752 
 - [x] Figure out where to add call to `LineAtlas#getDash`

### Phase 2: Pre-Code Review

 - [x] Fix failing render tests
 - [x] Resolve all PR comments from myself
 - [x] Remove unnecessary attributes from `dash_attributes.ts`
  - [x] Fix CI failures related to code coverage
 - [x] Post benchmark scores, investigate any regressions
    - [x] Figure out how to interpret emoji symbols in benchmarks (https://osmus.slack.com/archives/C01G3D28DAB/p1758057638986079)
 - [x] Update style spec and API documentation to advertise the new feature and explain gotchas like needing to use ["literal", [...]] syntax for array values
 - [x] Revise the PR description, add benchmark results and a technical overview to help reviewers understand the branch.
 - [x] Support round line caps w/ data-driven line-dasharray

### Phase 3: Code Review

_Underway_

- [x] Look into support for `"interpolate"` expressions with arrays. ([link](https://osmus.slack.com/archives/C01G3D28DAB/p1758778801576759?thread_ts=1758644582.948679&cid=C01G3D28DAB))
- [x] Look into support for mixing array literals and `null` in one expression. ([link](https://osmus.slack.com/archives/C01G3D28DAB/p1758778801576759?thread_ts=1758644582.948679&cid=C01G3D28DAB))
- [x] Test `global-state` expressions with array literals. ([link](https://osmus.slack.com/archives/C01G3D28DAB/p1758778801576759?thread_ts=1758644582.948679&cid=C01G3D28DAB))

### Phase 4: Post Code Review
_See https://github.com/maplibre/maplibre-gl-js/pull/5768 for similar workflow_

 - [x] Merge https://github.com/maplibre/maplibre-style-spec/pull/1100
 - [x] Publish new version of maplibre-style-spec
 - [x] Remove [`postinstall` script hack](https://github.com/maplibre/maplibre-gl-js/pull/5812/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R134) and update `package.json` to reference latest version of `maplibre-style-spec`
 - [ ] Merge this PR


1ec5:
> Due to limits in the expression system, arrays must be specified in the style using the ["literal", [...]] syntax. Feature properties cannot directly contain array values.

This would be a good use case for being able to form arrays dynamically: maplibre/maplibre-style-spec#950.

src/data/program_configuration.ts
```diff
@@ -479,6 +496,14 @@ export class ProgramConfiguration {
         }
     }
 
+    setConstantDasharrayPositions(dashTo: DashEntry, dashFrom: DashEntry) {
+        for (const property in this.binders) {
+            const binder = this.binders[property];
+            if (binder instanceof CrossFadedConstantBinder)
+                binder.setConstantDasharrayPositions(dashTo, dashFrom);
+        }
+    }
```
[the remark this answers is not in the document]
lucaswoj: We can get away with it in practice because lines cannot ever have a "pattern" and a "dasharray" at the same time.

package.json
```diff
@@ -138,6 +138,7 @@
     "vitest-webgl-canvas-mock": "^1.1.0"
   },
   "scripts": {
+    "postinstall": "cd .. && git clone https://github.com/lucaswoj/maplibre-style-spec.git && cd maplibre-style-spec && git checkout data-driven-line-dasharray && npm install && npm run build && npm run generate-typings && cd ../maplibre-gl-js && npm link ../maplibre-style-spec",
```
lucaswoj: Remove this after merging https://github.com/maplibre/maplibre-style-spec/pull/1100

src/data/bucket.ts
```diff
@@ -47,9 +48,9 @@ export type BucketFeature = {
     id?: any;
     readonly patterns: {
         [_: string]: {
-            'min': string;
-            'mid': string;
-            'max': string;
+            'min': any;
+            'mid': any;
+            'max': any;
```
lucaswoj: Come up with a dash-specific lookup table here

src/data/bucket/fill_bucket.ts
```diff
@@ -136,7 +136,7 @@ export class FillBucket implements Bucket {
     }
 
     addFeatures(options: PopulateParameters, canonical: CanonicalTileID, imagePositions: {
-        [_: string]: ImagePosition;
+        [_: string]: ImagePositionLike;
     }) {
```
lucaswoj, line 140: pass "dashes" as a separate parameter here, and everywhere ImagePositionLike is used.

src/data/bucket/pattern_bucket_features.ts
```diff
@@ -52,6 +65,65 @@ export function addPatternDependencies(type: string, layers: PatternStyleLayers,
...
+                        // Update the pattern feature to reference these new keys
+                        patternFeature.patterns[layer.id] = {min: minKey, mid: midKey, max: maxKey};
+                    }
+                }
+            }
+        }
+    }
+
+    return dasharrayPositions;
+}
```
lucaswoj: Factor all of this out into `line_bucket.ts`

src/data/program_configuration.ts
```diff
@@ -513,7 +514,7 @@ export class ProgramConfiguration {
         const result = [];
         for (const property in this.binders) {
             const binder = this.binders[property];
-            if (binder instanceof ConstantBinder || binder instanceof CrossFadedConstantBinder) {
+            if (property !== 'line-pattern' && (binder instanceof ConstantBinder || binder instanceof CrossFadedConstantBinder)) {
```
lucaswoj: Refactor this

src/data/program_configuration.ts
```diff
@@ -365,14 +366,14 @@ class CrossFadedCompositeBinder implements AttributeBinder {
...
+                imageMid.tlbr[0], imageMid.tlbr[1], imageMid.tlbr[2], imageMid.tlbr[3],
+                imageMin.tlbr[0], imageMin.tlbr[1], imageMin.tlbr[2], imageMin.tlbr[3],
                 imageMid.pixelRatio,
                 imageMin.pixelRatio,
             );
             this.zoomOutPaintVertexArray.emplace(i,
-                imageMid.tl[0], imageMid.tl[1], imageMid.br[0], imageMid.br[1],
-                imageMax.tl[0], imageMax.tl[1], imageMax.br[0], imageMax.br[1],
+                imageMid.tlbr[0], imageMid.tlbr[1], imageMid.tlbr[2], imageMid.tlbr[3],
+                imageMax.tlbr[0], imageMax.tlbr[1], imageMax.tlbr[2], imageMax.tlbr[3],
```
lucaswoj: Consider creating an entirely separate flow here for line-dasharray

src/render/draw_line.ts
```diff
@@ -69,6 +70,19 @@ export function drawLine(painter: Painter, sourceCache: SourceCache, layer: Line
             if (posTo && posFrom) programConfiguration.setConstantPatternPositions(posTo, posFrom);
         }
 
+        const constantDasharray = dasharray && dasharray.constantOr(null);
+        if (constantDasharray) {
+            const round = layer.layout.get('line-cap') === 'round';
+
+            const dashTo = painter.lineAtlas.getDash(constantDasharray.to, round);
+            const posTo = {tlbr: [0, dashTo.y, dashTo.height, dashTo.width], pixelRatio: 1};
```
lucaswoj: Ensure it's safe to set `pixelRatio: 1` here

src/render/draw_line.ts
```diff
@@ -78,17 +92,18 @@ export function drawLine(painter: Painter, sourceCache: SourceCache, layer: Line
         const pixelRatio = transform.getPixelScale();
 
         const uniformValues = image ? linePatternUniformValues(painter, tile, layer, pixelRatio, crossfade) :
-            dasharray ? lineSDFUniformValues(painter, tile, layer, pixelRatio, dasharray, crossfade) :
+            dasharray ? lineSDFUniformValues(painter, tile, layer, pixelRatio, dasharray.constantOr(null), crossfade) :
```
lucaswoj: Remove dasharray from this parameter list

src/render/draw_line.ts
```diff
@@ -78,17 +92,18 @@ export function drawLine(painter: Painter, sourceCache: SourceCache, layer: Line
...
 
         if (image) {
             context.activeTexture.set(gl.TEXTURE0);
             tile.imageAtlasTexture.bind(gl.LINEAR, gl.CLAMP_TO_EDGE);
             programConfiguration.updatePaintBuffers(crossfade);
-        } else if (dasharray && (programChanged || painter.lineAtlas.dirty)) {
+        } else if (dasharray) {
             context.activeTexture.set(gl.TEXTURE0);
             painter.lineAtlas.bind(context);
+            programConfiguration.updatePaintBuffers(crossfade);
```
lucaswoj, line 103: Revisit this. The final logic will depend on how much we want to share across between pattern and dasharray.

src/render/line_atlas.ts
```diff
@@ -180,7 +180,7 @@ export class LineAtlas {
         }
 
         const dashEntry = {
-            y: (this.nextRow + n + 0.5) / this.height,
+            y: ((this.nextRow + n + 0.5) / this.height) * 1024,
```
lucaswoj: Compute this constant at runtime, or find a better way to handle this. Ideally this number is representable as a uint16. 1024 = width * height * 2. (The factor of 2 allows us to encode the 0.5).

src/shaders/line_sdf.fragment.glsl
```diff
@@ -38,8 +42,10 @@ void main() {
     float sdfdist_a = texture(u_image, v_tex_a).a;
     float sdfdist_b = texture(u_image, v_tex_b).a;
     float sdfdist = mix(sdfdist_a, sdfdist_b, u_mix);
-    alpha *= smoothstep(0.5 - u_sdfgamma / floorwidth, 0.5 + u_sdfgamma / floorwidth, sdfdist);
+    float sdfgamma = u_sdfgammaratio / min(pattern_from.w, pattern_to.w);
+    alpha *= smoothstep(0.5 - sdfgamma / floorwidth, 0.5 + sdfgamma / floorwidth, sdfdist);
 
+    // fragColor = vec4(0.0, 0.0, sdfdist_a * 2.0, 1.0);
```
lucaswoj: Remove this

src/shaders/line_sdf.vertex.glsl
```diff
@@ -103,7 +106,12 @@ void main() {
...
 
-    v_tex_a = vec2(a_linesofar * u_patternscale_a.x / floorwidth, normal.y * u_patternscale_a.y + u_tex_y_a);
-    v_tex_b = vec2(a_linesofar * u_patternscale_b.x / floorwidth, normal.y * u_patternscale_b.y + u_tex_y_b);
+    float u_patternscale_a_x = u_tileratio / pattern_from.w / u_crossfade_from;
+    float u_patternscale_a_y = -pattern_from.z / 2.0;
+    float u_patternscale_b_x = u_tileratio / pattern_to.w / u_crossfade_to;
+    float u_patternscale_b_y = -pattern_to.z / 2.0;
+
+    v_tex_a = vec2(a_linesofar * u_patternscale_a_x / floorwidth, normal.y * u_patternscale_a_y + (float(pattern_from.y) / 1024.0));
+    v_tex_b = vec2(a_linesofar * u_patternscale_b_x / floorwidth, normal.y * u_patternscale_b_y + (float(pattern_to.y) / 1024.0));
```
lucaswoj: consider renaming some of these vars

src/source/tile.test.ts
```diff
@@ -48,9 +48,9 @@ describe('querySourceFeatures', () => {
         tile.querySourceFeatures(result, {sourceLayer: undefined, filter: ['!=', 'oneway', true]});
         expect(result).toHaveLength(0);
         result = [];
-        const polygon = {type: 'Polygon',  coordinates: [[[-91, -1], [-89, -1], [-89, 1], [-91, 1], [-91, -1]]]};
-        tile.querySourceFeatures(result, {sourceLayer: undefined, filter: ['within', polygon]});
-        expect(result).toHaveLength(1);
+        // TODO enable me
+        // tile.querySourceFeatures(result, {sourceLayer: undefined, filter: ['within', polygon]});
+        // expect(result).toHaveLength(1);
```
lucaswoj: fix this

src/source/worker_tile.ts
```diff
@@ -162,10 +165,23 @@ export class WorkerTile {
...
+        }
+
+        const [glyphMap, iconMap, patternMap, dasharrayMap] = await Promise.all([getGlyphsPromise, getIconsPromise, getPatternsPromise, getDasharrayPromise]);
+
         const glyphAtlas = new GlyphAtlas(glyphMap);
         const imageAtlas = new ImageAtlas(iconMap, patternMap);
 
+        // Process dasharray dependencies and generate dash positions
+        const dasharrayPositions = addDasharrayDependencies(buckets, dasharrayMap);
+        const allPatternPositions = {...imageAtlas.patternPositions, ...dasharrayPositions};
```
lucaswoj: Instead of combining these, consider passing them separately down through `addFeatures`

src/util/actor_messages.ts
```diff
@@ -76,9 +82,25 @@ export type GetGlyphsResponse = {
...
 
 /**
- * A response object returned when requesting images
+ * Parameters needed to get the line dashes
  */
-export type GetImagesResponse = {[_: string]: StyleImage};
+export type GetDashesParameters = {
+    dashes: Array<{
+        dasharray: Array<number>;
+        round: boolean;
```
lucaswoj, line 91: Factor this out into a shared type

src/data/bucket/pattern_bucket_features.ts
```diff
@@ -52,6 +65,65 @@ export function addPatternDependencies(type: string, layers: PatternStyleLayers,
...
+                const round = false; // TODO handle round caps if needed
+
+                const dasharrayPropertyValue = dasharrayProperty.value;
+                const min = {dasharray: dasharrayPropertyValue.evaluate({zoom: zoom - 1}, patternFeature, {}), round};
+                const mid = {dasharray: dasharrayPropertyValue.evaluate({zoom}, patternFeature, {}), round};
+                const max = {dasharray: dasharrayPropertyValue.evaluate({zoom: zoom + 1}, patternFeature, {}), round};
+
+                dasharrays.push(min);
+                dasharrays.push(mid);
+                dasharrays.push(max);
```
lucaswoj: Deduplicate dash definitions, probably want to create a unique string per.

src/data/bucket/pattern_bucket_features.ts
```diff
@@ -52,6 +65,65 @@ export function addPatternDependencies(type: string, layers: PatternStyleLayers,
             // save for layout
             patternFeature.patterns[layer.id] = {min, mid, max};
         }
+
+        // Handle data-driven dasharray
+        if (type === 'line') {
+            const dasharrayProperty = (layer.paint as PossiblyEvaluated<any, any>).get('line-dasharray');
+            if (dasharrayProperty && dasharrayProperty.value.kind !== 'constant') {
+                const round = false; // TODO handle round caps if needed
```
lucaswoj: Properly handle "round" here, including potential property functions.

src/data/bucket/pattern_bucket_features.ts
```diff
@@ -52,6 +65,65 @@ export function addPatternDependencies(type: string, layers: PatternStyleLayers,
...
+                        const minKey = `dash_${JSON.stringify(min)}_${round}_min`;
+                        const midKey = `dash_${JSON.stringify(mid)}_${round}_mid`;
+                        const maxKey = `dash_${JSON.stringify(max)}_${round}_max`;
+
+                        dasharrayPositions[minKey] = {tlbr: [0, dashMin.y, dashMin.height, dashMin.width], pixelRatio: 1};
+                        dasharrayPositions[midKey] = {tlbr: [0, dashMid.y, dashMid.height, dashMid.width], pixelRatio: 1};
+                        dasharrayPositions[maxKey] = {tlbr: [0, dashMax.y, dashMax.height, dashMax.width], pixelRatio: 1};
+
+                        // Update the pattern feature to reference these new keys
+                        patternFeature.patterns[layer.id] = {min: minKey, mid: midKey, max: maxKey};
```
lucaswoj: Apply similar logic here as in https://github.com/maplibre/maplibre-gl-js/pull/5812/files#r2173455337

pnorman:
It's great to see work here and I'm looking forwards to when the feature is merged!

lucaswoj:
Woohoo! All render tests are now passing. I think there are a few edge cases that are broken but untested. However this is a big step closer to marking this "ready for review."

src/render/program/line_program.ts
```diff
@@ -155,29 +156,20 @@ const lineSDFUniformValues = (
...
-    const widthB = posB.width * crossfade.toScale;
-
     return extend(lineUniformValues(painter, tile, layer, ratioScale), {
-        'u_patternscale_a': [tileRatio / widthA, -posA.height / 2],
-        'u_patternscale_b': [tileRatio / widthB, -posB.height / 2],
-        'u_sdfgamma': lineAtlas.width / (Math.min(widthA, widthB) * 256 * painter.pixelRatio) / 2,
+        'u_tileratio': tileRatio,
+        'u_crossfade_from': crossfade.fromScale,
+        'u_crossfade_to': crossfade.toScale,
+        'u_sdfgammaratio': painter.lineAtlas.width / 256,
```
lucaswoj: Should `painter.pixelRatio` be a factor in this equation? Or is `lineAtlas` totally independent of `pixelRatio`?

src/data/bucket/circle_bucket.ts
```diff
@@ -122,6 +122,7 @@ export class CircleBucket<Layer extends CircleStyleLayer | HeatmapStyleLayer> im
                 index,
                 geometry: needGeometry ? evaluationFeature.geometry : loadGeometry(feature),
                 patterns: {},
+                dashes: {},
```
lucaswoj: Can we make this property optional so that `circle_bucket`, `fill_bucket`, etc are untouched in this PR?

src/data/bucket/pattern_bucket_features.ts
```diff
@@ -26,13 +28,23 @@ export function hasPattern(type: string, layers: PatternStyleLayers, options: Po
             patterns[constantPattern.to] =  true;
             patterns[constantPattern.from] =  true;
         }
+
+        // Also check for data-driven dasharray
+        if (type === 'line') {
+            const dasharrayProperty = (layer.paint as PossiblyEvaluated<any, any>).get('line-dasharray');
+            if (dasharrayProperty && !dasharrayProperty.isConstant()) {
+                hasPattern = true;
```
lucaswoj: Why do we need to set `patterns[]` above but there's no equivalent `dashes[]` here?

src/data/bucket/pattern_bucket_features.ts
```diff
@@ -52,6 +64,62 @@ export function addPatternDependencies(type: string, layers: PatternStyleLayers,
...
+
+                const minKey = JSON.stringify(min);
+                const midKey = JSON.stringify(mid);
+                const maxKey = JSON.stringify(max);
+
+                dasharrays[minKey] = min;
+                dasharrays[midKey] = mid;
+                dasharrays[maxKey] = max;
+
+                patternFeature.dashes[layer.id] = {min: minKey, mid: midKey, max: maxKey};
```
lucaswoj: Would be wise to add a zoom function render test that exercises this codepath

src/data/bucket/pattern_bucket_features.ts
```diff
@@ -52,6 +64,62 @@ export function addPatternDependencies(type: string, layers: PatternStyleLayers,
             // save for layout
             patternFeature.patterns[layer.id] = {min, mid, max};
         }
+
+        // Handle data-driven dasharray
+        if (type === 'line') {
+            const dasharrayProperty = (layer.paint as PossiblyEvaluated<any, any>).get('line-dasharray');
+            if (dasharrayProperty && dasharrayProperty.value.kind !== 'constant') {
+                const round = false;
```
lucaswoj: Lookup this value at runtime, and add render tests for various cases.

src/data/bucket/pattern_bucket_features.ts
```diff
@@ -52,6 +64,62 @@ export function addPatternDependencies(type: string, layers: PatternStyleLayers,
...
+                dasharrays[maxKey] = max;
+
+                patternFeature.dashes[layer.id] = {min: minKey, mid: midKey, max: maxKey};
+            }
+        }
     }
     return patternFeature;
 }
+
+export function addDasharrayDependencies(buckets: {[_: string]: any}, dashes: GetDashesResponse): {[_: string]: ImagePositionLike} {
```
lucaswoj: Is there a better spot for this function to live?

src/data/program_configuration.ts
```diff
@@ -513,8 +513,19 @@ export class ProgramConfiguration {
...
+                if (property === 'line-dasharray' || property === 'line-pattern') {
+                    const lineDasharrayBinder = this.binders['line-dasharray'];
+                    const linePatternBinder = this.binders['line-pattern'];
+
+                    if ((lineDasharrayBinder instanceof ConstantBinder || lineDasharrayBinder instanceof CrossFadedConstantBinder) && (linePatternBinder instanceof ConstantBinder || linePatternBinder instanceof CrossFadedConstantBinder)) {
+                        result.push(...binder.uniformNames.map(name => `#define HAS_UNIFORM_${name}`));
+                    }
+                } else {
+
+                    result.push(...binder.uniformNames.map(name => `#define HAS_UNIFORM_${name}`));
```
lucaswoj: TODO tidy up this logic

src/render/program/line_program.ts
lucaswoj: Relatedly, do we even need this property if it's easily computable from `u_lineatlas_width?`

src/render/draw_line.ts
```diff
@@ -78,17 +88,18 @@ export function drawLine(painter: Painter, sourceCache: SourceCache, layer: Line
...
+            dasharray ? lineSDFUniformValues(painter, tile, layer, pixelRatio, crossfade) :
                 gradient ? lineGradientUniformValues(painter, tile, layer, pixelRatio, bucket.lineClipsArray.length) :
                     lineUniformValues(painter, tile, layer, pixelRatio);
 
         if (image) {
             context.activeTexture.set(gl.TEXTURE0);
             tile.imageAtlasTexture.bind(gl.LINEAR, gl.CLAMP_TO_EDGE);
             programConfiguration.updatePaintBuffers(crossfade);
-        } else if (dasharray && (programChanged || painter.lineAtlas.dirty)) {
+        } else if (dasharray) {
```
lucaswoj, line 98: Review this change, make sure we aren't causing any perf issues.

src/source/worker_tile.ts
```diff
@@ -162,10 +164,22 @@ export class WorkerTile {
...
+            getDasharrayPromise = actor.sendAsync({type: MessageType.getDashes, data: {dashes: dasharray, source: this.source, tileID: this.tileID, type: 'dasharray'}}, abortController);
+        }
+
+        const [glyphMap, iconMap, patternMap, dasharrayMap] = await Promise.all([getGlyphsPromise, getIconsPromise, getPatternsPromise, getDasharrayPromise]);
+
         const glyphAtlas = new GlyphAtlas(glyphMap);
         const imageAtlas = new ImageAtlas(iconMap, patternMap);
 
+        // Process dasharray dependencies and generate dash positions
+        const dasharrayPositions = addDasharrayDependencies(buckets, dasharrayMap);
```
lucaswoj: This feels weird. Why is this codepath so different from patterns otherwise?

src/shaders/line_sdf.vertex.glsl
```diff
@@ -103,7 +107,12 @@ void main() {
...
 
-    v_tex_a = vec2(a_linesofar * u_patternscale_a.x / floorwidth, normal.y * u_patternscale_a.y + u_tex_y_a);
-    v_tex_b = vec2(a_linesofar * u_patternscale_b.x / floorwidth, normal.y * u_patternscale_b.y + u_tex_y_b);
+    float u_patternscale_a_x = u_tileratio / pattern_from.w / u_crossfade_from;
+    float u_patternscale_a_y = -pattern_from.z / 2.0;
+    float u_patternscale_b_x = u_tileratio / pattern_to.w / u_crossfade_to;
+    float u_patternscale_b_y = -pattern_to.z / 2.0;
+
+    v_tex_a = vec2(a_linesofar * u_patternscale_a_x / floorwidth, normal.y * u_patternscale_a_y + (float(pattern_from.y) + 0.5) / u_lineatlas_height);
+    v_tex_b = vec2(a_linesofar * u_patternscale_b_x / floorwidth, normal.y * u_patternscale_b_y + (float(pattern_to.y) + 0.5) / u_lineatlas_height);
```
lucaswoj: refactor this for clarity

src/shaders/line_sdf.vertex.glsl
```diff
@@ -16,11 +16,11 @@ in vec4 a_data;
...
 uniform mediump float u_ratio;
 uniform lowp float u_device_pixel_ratio;
-uniform vec2 u_patternscale_a;
-uniform float u_tex_y_a;
-uniform vec2 u_patternscale_b;
-uniform float u_tex_y_b;
 uniform vec2 u_units_to_pixels;
+uniform float u_tileratio;
+uniform float u_crossfade_from;
+uniform float u_crossfade_to; // TODO: this is always 1, consider removing it
```
lucaswoj: Revisit this TODO

src/shaders/line_sdf.fragment.glsl
lucaswoj: Remove this

src/render/draw_line.ts
```diff
@@ -78,17 +88,18 @@ export function drawLine(painter: Painter, sourceCache: SourceCache, layer: Line
...
             context.activeTexture.set(gl.TEXTURE0);
             painter.lineAtlas.bind(context);
+            programConfiguration.updatePaintBuffers(crossfade);
```
lucaswoj, line 103: Review this change, make sure we aren't causing any perf issues.

src/data/bucket/bucket_line_atlas.ts
```diff
@@ -0,0 +1,33 @@
+import type {ImagePositionLike} from '../../render/image_atlas';
+import type {GetDashesResponse} from '../../util/actor_messages';
+
+export function createBucketLineAtlas(buckets: {[_: string]: any}, dashes: GetDashesResponse): {[_: string]: ImagePositionLike} {
```
lucaswoj: Either use a real LineAtlas in the worker or add a comment explaining why this case is different from ImageAtlas

src/data/bucket/fill_bucket.ts
```diff
@@ -26,7 +26,7 @@ import type {IndexBuffer} from '../../gl/index_buffer';
 import type {VertexBuffer} from '../../gl/vertex_buffer';
 import type Point from '@mapbox/point-geometry';
 import type {FeatureStates} from '../../source/source_state';
-import type {ImagePosition} from '../../render/image_atlas';
+import type {ImagePosition, ImagePositionLike} from '../../render/image_atlas';
```
lucaswoj: Eliminate ImagePositionLike. Pass around `ImagePosition` and `DashEntry` separately.

src/data/bucket/pattern_bucket_features.ts
```diff
@@ -52,6 +62,29 @@ export function addPatternDependencies(type: string, layers: PatternStyleLayers,
...
+                const midKey = JSON.stringify(mid);
+                const maxKey = JSON.stringify(max);
+
+                dasharrays[minKey] = min;
+                dasharrays[midKey] = mid;
+                dasharrays[maxKey] = max;
+
+                patternFeature.dashes[layer.id] = {min: minKey, mid: midKey, max: maxKey};
+            }
+        }
```
lucaswoj: Let's fully factor out `line-dasharray` logic from `pattern_bucket_features`. We can put it directly in `line_bucket` instead.

src/data/program_configuration.ts
```diff
@@ -149,13 +150,17 @@ class CrossFadedConstantBinder implements UniformBinder {
...
             uniformName === 'u_pattern_to' ? this.patternTo :
                 uniformName === 'u_pattern_from' ? this.patternFrom :
-                    uniformName === 'u_pixel_ratio_to' ? this.pixelRatioTo :
-                        uniformName === 'u_pixel_ratio_from' ? this.pixelRatioFrom : null;
+                    uniformName === 'u_dasharray_to' ? this.patternTo :
+                        uniformName === 'u_dasharray_from' ? this.patternFrom :
+                            uniformName === 'u_pixel_ratio_to' ? this.pixelRatioTo :
+                                uniformName === 'u_pixel_ratio_from' ? this.pixelRatioFrom :
+                                    uniformName === 'u_dash_pixel_ratio_to' ? this.pixelRatioTo :
+                                        uniformName === 'u_dash_pixel_ratio_from' ? this.pixelRatioFrom : null;
```
lucaswoj: Refactor this.

src/data/program_configuration.ts
```diff
@@ -365,14 +370,14 @@ class CrossFadedCompositeBinder implements AttributeBinder {
...
+                imageMid.tlbr[0], imageMid.tlbr[1], imageMid.tlbr[2], imageMid.tlbr[3],
+                imageMin.tlbr[0], imageMin.tlbr[1], imageMin.tlbr[2], imageMin.tlbr[3],
                 imageMid.pixelRatio,
                 imageMin.pixelRatio,
             );
             this.zoomOutPaintVertexArray.emplace(i,
-                imageMid.tl[0], imageMid.tl[1], imageMid.br[0], imageMid.br[1],
-                imageMax.tl[0], imageMax.tl[1], imageMax.br[0], imageMax.br[1],
+                imageMid.tlbr[0], imageMid.tlbr[1], imageMid.tlbr[2], imageMid.tlbr[3],
+                imageMax.tlbr[0], imageMax.tlbr[1], imageMax.tlbr[2], imageMax.tlbr[3],
```
lucaswoj: Undo these changes, localize new code to `CrossFadedDasharrayBinder`

src/data/program_configuration.ts
```diff
@@ -365,14 +370,14 @@ class CrossFadedCompositeBinder implements AttributeBinder {
         // unnecessary vertex data to the shaders, we determine which to upload at draw time.
         for (let i = start; i < end; i++) {
             this.zoomInPaintVertexArray.emplace(i,
-                imageMid.tl[0], imageMid.tl[1], imageMid.br[0], imageMid.br[1],
```
lucaswoj, line 375: Rename `CrossFadedCompositeBinder` to `CrossFadedPatternBinder`

src/data/program_configuration.ts
```diff
@@ -392,6 +397,84 @@ class CrossFadedCompositeBinder implements AttributeBinder {
     }
 }
 
+class CrossFadedDasharrayBinder implements AttributeBinder {
```
lucaswoj: Look for ways to share more code with `CrossFadedCompositeBinder` / `CrossFadedPatternBinder`

src/data/program_configuration.ts
```diff
@@ -683,6 +773,7 @@ function paintAttributeNames(property, type) {
         'text-halo-width': ['halo_width'],
         'icon-halo-width': ['halo_width'],
         'line-gap-width': ['gapwidth'],
+        'line-dasharray': ['dasharray_to', 'dasharray_from', 'dash_pixel_ratio_to', 'dash_pixel_ratio_from'],
```
lucaswoj: See if we can drop the `dash_pixel_ratio_` attributes entirely

src/source/worker_tile.ts
```diff
@@ -79,6 +80,7 @@ export class WorkerTile {
             iconDependencies: {},
             patternDependencies: {},
             glyphDependencies: {},
+            dasharrayDependencies: {} as {[key: string]: {round: boolean; dasharray: Array<number>}},
```
lucaswoj: Remove the `as` statement here, no other properties have one.

build/generate-struct-arrays.ts
```diff
@@ -146,7 +147,8 @@ const layoutAttributes = {
     heatmap: circleAttributes,
     line: lineLayoutAttributes,
     lineExt: lineLayoutAttributesExt,
-    pattern: patternAttributes
+    pattern: patternAttributes,
+    dasharray: dasharrayAttributes
```
lucaswoj: Rename this, and similar, to `dash` instead of `dasharray`. Reserve the latter for the literal array that specifies the dash style.

src/data/bucket.ts
```diff
@@ -27,6 +27,7 @@ export type PopulateParameters = {
     iconDependencies: {};
     patternDependencies: {};
     glyphDependencies: {};
+    dasharrayDependencies: {[key: string]: {round: boolean; dasharray: Array<number>}};
```
lucaswoj: Rename to `dashDependencies `

src/data/bucket/bucket_line_atlas.ts
```diff
@@ -0,0 +1,33 @@
...
+                            dasharrayPositions[midKey] = {tlbr: [0, dashMid.y, dashMid.height, dashMid.width], pixelRatio: 1};
+                            dasharrayPositions[maxKey] = {tlbr: [0, dashMax.y, dashMax.height, dashMax.width], pixelRatio: 1};
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    return dasharrayPositions;
```
lucaswoj: Rename dasharray to dash

src/data/bucket/pattern_bucket_features.ts
```diff
@@ -26,13 +26,23 @@ export function hasPattern(type: string, layers: PatternStyleLayers, options: Po
...
     }
 
     return hasPattern;
 }
 
 export function addPatternDependencies(type: string, layers: PatternStyleLayers, patternFeature: BucketFeature, zoom: number, options: PopulateParameters) {
     const patterns = options.patternDependencies;
+    const dasharrays = options.dasharrayDependencies;
```
lucaswoj: Rename to "dashes"

src/shaders/line_sdf.fragment.glsl
```diff
@@ -18,13 +18,17 @@ in float v_depth;
 #pragma mapbox: define lowp float opacity
 #pragma mapbox: define mediump float width
 #pragma mapbox: define lowp float floorwidth
+#pragma mapbox: define mediump vec4 dasharray_from
+#pragma mapbox: define mediump vec4 dasharray_to
```
lucaswoj, line 22: Rename to "dash"

src/source/worker_tile.ts
```diff
@@ -162,9 +164,19 @@ export class WorkerTile {
             getPatternsPromise = actor.sendAsync({type: MessageType.getImages, data: {icons: patterns, source: this.source, tileID: this.tileID, type: 'patterns'}}, abortController);
         }
 
-        const [glyphMap, iconMap, patternMap] = await Promise.all([getGlyphsPromise, getIconsPromise, getPatternsPromise]);
+        const dasharray = options.dasharrayDependencies;
+        let getDasharrayPromise = Promise.resolve<GetDashesResponse>({} as GetDashesResponse);
```
lucaswoj: Rename to "dash"

src/data/bucket/dash_attributes.ts
```diff
@@ -0,0 +1,9 @@
+import {createLayout} from '../../util/struct_array';
+
+export const dashAttributes = createLayout([
+    // [tl.x, tl.y, br.x, br.y] - coordinates for dasharray texture lookup
+    {name: 'a_dasharray_from', components: 4, type: 'Uint16'},
+    {name: 'a_dasharray_to', components: 4, type: 'Uint16'},
+    {name: 'a_dash_pixel_ratio_from', components: 1, type: 'Uint16'},
+    {name: 'a_dash_pixel_ratio_to', components: 1, type: 'Uint16'},
```
lucaswoj: We can reduce the size of this struct now that it's forked from pattern_attributes

pnorman:
Is there any way others can help with this right now?

lucaswoj:
Thanks for the offer @pnorman! Right now I'm investigating one lingering bug, related to rendering data-driven dash array with round caps, and then I'll be campaigning for a PR review. I can't think of anything that others could help with right now.

src/data/bucket/dash_attributes.ts
```diff
@@ -0,0 +1,7 @@
+import {createLayout} from '../../util/struct_array';
+
+export const dashAttributes = createLayout([
+    // [0, y, height, width]
+    {name: 'a_dasharray_from', components: 4, type: 'Uint16'},
+    {name: 'a_dasharray_to', components: 4, type: 'Uint16'},
+]);
```
lucaswoj, line 7: Since element 0 is always set to 0, let's take a stab at removing that element to reduce the buffer size.

lucaswoj:
This PR is ready for review!

Please don't hesitate to reach out to me in the OSM Slack with any questions. I appreciate that this is a lot of code to review. 🙇🏻

dschep:
> Due to limits in the expression system, dasharrays must be specified in the style using the ["literal", [...]] syntax. Feature properties cannot directly contain array values.

Does this apply to all use of `line-dasharray` or only within data driven expressions? If it's the former, that makes this a breaking change right? And it would be breaking almost all existing MapLibre styles(at least all styles that use `line-dasharray`, which is _a lot_ of them)

lucaswoj:
> > Due to limits in the expression system, dasharrays must be specified in the style using the ["literal", [...]] syntax. Feature properties cannot directly contain array values.
> 
> Does this apply to all use of `line-dasharray` or only within data driven expressions? If it's the former, that makes this a breaking change right? And it would be breaking almost all existing MapLibre styles(at least all styles that use `line-dasharray`, which is _a lot_ of them)

There's no breaking change here. The existing non-expression syntax like `"line-dasharray": [1, 2]` will still work. What I'm trying to say here is that there's no way to create an arbitrary array at runtime, like if you wanted to calculate the dash width / gap with from a feature property, `"line-dasharray": ["array", ["get", "foo"], ["get", "bar"]]`

package.json
```diff
@@ -131,6 +131,7 @@
     "vitest-webgl-canvas-mock": "^1.1.0"
   },
   "scripts": {
+    "postinstall": "cd .. && (git clone https://github.com/lucaswoj/maplibre-style-spec.git || true) && cd maplibre-style-spec && (git checkout data-drive-line-dasharray || git checkout main) && npm install && npm run build && npm run generate-typings && cd ../maplibre-gl-js && npm link ../maplibre-style-spec",
```
HarelM: This needs to be reverted, Let me know if you need me to release a new version of the style-spec to be able to revert this.

src/util/actor_messages.ts
```diff
@@ -7,6 +7,7 @@ import type {PluginState} from '../source/rtl_text_plugin_status';
 import type {LayerSpecification} from '@maplibre/maplibre-gl-style-spec';
 import type {OverscaledTileID} from '../source/tile_id';
 import type {GetResourceResponse, RequestParameters} from './ajax';
+import {type DashEntry} from '../render/line_atlas';
```
HarelM: nit picking: can this import be like the line above it where the type is outside?

src/style/style.ts
```diff
@@ -1883,6 +1888,14 @@ export class Style extends Evented {
         this.glyphManager.setURL(glyphsUrl);
     }
 
+    async getDashes(mapId: string | number, params: GetDashesParameters): Promise<GetDashesResponse> {
```
HarelM, line 1891: What part of this this method is async?

src/source/worker_source.ts
```diff
@@ -59,6 +60,7 @@ export type WorkerDEMTileParameters = TileParameters & {
 export type WorkerTileResult = ExpiryData & {
     buckets: Array<Bucket>;
     imageAtlas: ImageAtlas;
+    dashPositions: {[_: string]: DashEntry};
```
HarelM: `Record<string, DashEntry>`?

src/render/line_atlas.ts
```diff
@@ -180,8 +180,8 @@ export class LineAtlas {
         }
 
         const dashEntry = {
-            y: (this.nextRow + n + 0.5) / this.height,
-            height: 2 * n / this.height,
+            y: this.nextRow + n,
```
HarelM, line 183: Why was height removed? doesn't it create smaller numbers which allows better precision?

src/data/program_configuration.ts
```diff
@@ -321,7 +346,7 @@ class CompositeExpressionBinder implements AttributeBinder, UniformBinder {
     }
 }
 
-class CrossFadedCompositeBinder implements AttributeBinder {
+abstract class CrossFadedBinder<T> implements AttributeBinder {
```
HarelM, line 349: Can we do composition instead of inheritance? I find these abstract classes hard to "follow"...

src/data/program_configuration.ts
```diff
@@ -535,9 +610,10 @@ export class ProgramConfiguration {
                 for (let i = 0; i < binder.paintVertexAttributes.length; i++) {
                     result.push(binder.paintVertexAttributes[i].name);
                 }
-            } else if (binder instanceof CrossFadedCompositeBinder) {
-                for (let i = 0; i < patternAttributes.members.length; i++) {
-                    result.push(patternAttributes.members[i].name);
+            } else if (binder instanceof CrossFadedBinder) {
+                const attributes = binder.getVertexAttributes();
+                for (let i = 0; i < attributes.length; i++) {
```
HarelM: can we use `for ... of`?

src/data/program_configuration.ts
```diff
@@ -608,7 +684,7 @@ export class ProgramConfiguration {
     upload(context: Context) {
         for (const property in this.binders) {
             const binder = this.binders[property];
-            if (binder instanceof SourceExpressionBinder || binder instanceof CompositeExpressionBinder || binder instanceof CrossFadedCompositeBinder)
+            if (binder instanceof SourceExpressionBinder || binder instanceof CompositeExpressionBinder || binder instanceof CrossFadedBinder)
```
HarelM, line 687: Is it possible to check something else beside `instanceof`?

src/data/program_configuration.ts
HarelM: I would also consider moving this to a different file, I'm not sure I understand how it is related to "program configuration" (the file's name), as it seems a bit too specific...

src/data/bucket.ts
```diff
@@ -51,6 +53,13 @@ export type BucketFeature = {
             'max': string;
         };
     };
+    readonly dashes?: {
```
HarelM: Wasn't this defined elesewhere? Maybe in the spec package?

src/data/bucket/line_bucket.ts
```diff
@@ -597,3 +606,47 @@ export class LineBucket implements Bucket {
 }
 
 register('LineBucket', LineBucket, {omit: ['layers', 'patternFeatures']});
+
+function hasLineDasharray(layers: Array<LineStyleLayer>): boolean {
```
HarelM: Why not make them a private function of the `LineBucket` class?

src/data/bucket/line_bucket.ts
```diff
@@ -597,3 +606,47 @@ export class LineBucket implements Bucket {
...
+    for (const layer of layers) {
+        const dasharrayProperty = (layer.paint as PossiblyEvaluated<any, any>).get('line-dasharray');
```
HarelM: Can we better type these two "any"s?

src/data/bucket/line_bucket.ts
```diff
@@ -597,3 +606,47 @@ export class LineBucket implements Bucket {
...
+            const mid = {
+                dasharray: dasharrayProperty.value.evaluate({zoom}, bucketFeature, {}),
+                round
+            };
+            const max = {
+                dasharray: dasharrayProperty.value.evaluate({zoom: zoom + 1}, bucketFeature, {}),
+                round
+            };
+
+            const minKey = `${min.dasharray.join(',')},${min.round}`;
```
HarelM, line 639: Is color important to the key generation? line width? Other properties?

src/data/bucket/line_bucket.ts
```diff
@@ -145,7 +147,7 @@ export class LineBucket implements Bucket {
     }
 
     populate(features: Array<IndexedFeature>, options: PopulateParameters, canonical: CanonicalTileID) {
-        this.hasPattern = hasPattern('line', this.layers, options);
+        this.hasPattern = hasPattern('line', this.layers, options) || hasLineDasharray(this.layers);
```
HarelM: Not sure `hasPattern` is a good name after this change...?

.github/workflows/test-all.yml
```diff
@@ -217,6 +217,7 @@ jobs:
           path: coverage-merged
 
       - name: Code Coverage Annotation
+        if: github.event.pull_request.head.repo.full_name == github.repository
```
HarelM, line 220: This probably needs to be reverted...?

src/source/worker_tile.ts
```diff
@@ -177,12 +187,9 @@ export class WorkerTile {
                     canonical: this.tileID.canonical,
                     subdivisionGranularity: options.subdivisionGranularity
                 });
-            } else if (bucket.hasPattern &&
-                (bucket instanceof LineBucket ||
-                bucket instanceof FillBucket ||
-                bucket instanceof FillExtrusionBucket)) {
+            } else if (bucket.hasPattern && (bucket instanceof FillBucket || bucket instanceof FillExtrusionBucket || bucket instanceof LineBucket)) {
```
HarelM: Same here with the `instanceof`...?

src/style/style.test.ts
```diff
@@ -3378,4 +3378,33 @@ describe('Style.serialize', () => {
         expect(style.sky.properties.get('fog-color').g).toBe(1);
         expect(style.sky.properties.get('fog-color').r).toBe(0);
     });
+
+    test('Style.getDashes returns line atlas entries for dash patterns', async () => {
```
HarelM, line 3382: I'm not sure I understand how the dasharray get created. This test looks good otherwise.

HarelM:
I've added some minor comments I believe.
This looks good, especially the part where the tests cover all the newly added functionality.
Thanks a lot for pushing this forward!
Let me know what else you need from my side to push this through.

package.json
lucaswoj: Yes, we'll need to publish a new version of the style-spec before I can revert this. Can you do the release?

src/style/style.ts
lucaswoj: Nothing in this method is async. But `registerMessageHandler`, to which this method is passed, requires a `Promise` type. `setGlyphs` similarly is marked as `async` so that it returns a promise despite not doing any actual async operations. I'd prefer to stick with the existing code patterns in this file than try to tackle a refactor within this PR.

src/render/line_atlas.ts
lucaswoj: We need all values in `dashEntry` to be integers for packing into a vertex buffer of `Uint16`s. Nothing has functionally changed, we simply divide by `line_atlas_height` in the shader rather than right here.

src/data/program_configuration.ts
lucaswoj: Using `for ... of` is possible here but undesirable for two reasons:

 - `for ... of` has worse performance
 - `for ... of` typically implies iteration over an object instead of an array
 
 Do you have any specific concerns about using a "traditional" `for` loop here?

src/data/program_configuration.ts
lucaswoj: Rewriting to use composition would add a lot more boilerplate code, to the point that it might just be better to simply duplicate the logic. I'm against this refactor.

lucaswoj: If we move this class to a new file now, we'll loose the ability to see a PR diff against the existing implementation. I'd like to keep the ability to see that diff for now.

src/data/program_configuration.ts
lucaswoj: I can't think of a better way to write this logic without a much bigger refactor. Do you have any specific concerns about `instanceof`?

src/data/bucket.ts
lucaswoj: Yes. The style spec exports a similar `Feature` type that's used in some codepaths on the "main thread." This `BucketFeature` type is used exclusively within the WebWorkers.  The two are similar but not exactly cross-compatible.

src/data/bucket/line_bucket.ts
lucaswoj: Done

src/data/bucket/line_bucket.ts
lucaswoj: These are the only two arguments passed to [`LineAtlas#getDash`](https://github.com/lucaswoj/maplibre-gl-js/blob/3ce9efdcfe19b4c7e9a7909c8e6b9da6e76c320d/src/render/line_atlas.ts#L50). This key refers to an entry in `LineAtlas`.

src/data/bucket/line_bucket.ts
lucaswoj: Renamed

.github/workflows/test-all.yml
lucaswoj: Done

src/data/bucket/line_bucket.ts
lucaswoj: Done

src/source/worker_tile.ts
lucaswoj: I can't think of a better way to write this logic without a much bigger refactor. Do you have any specific concerns about `instanceof`?

src/style/style.test.ts
lucaswoj: We define two arbitrary dasharray "fixtures" for this test and pass them to `await style.getDashes('mapId', params)` to exercise all the code that adds them to the `LineAtlas`. Does that answer your question?

lucaswoj:
Thanks for the review @HarelM. I've gone through the review and responded to each comment. In cases where I was able to make the requested change, I resolved the comment. This is ready for another round of review.

src/style/style.test.ts
HarelM: Yes, thanks!

src/source/worker_tile.ts
HarelM: Only that it's not OOP and "smells".
But it can be done in a separate PR.

src/data/program_configuration.ts
HarelM: Only because the "i" is not used besides getting the attribute, so it's less "readable".
I prefer to use for-of when "i" is used only to get the element out of the array, which is the case here, right?

src/style/style.ts
HarelM: Fair enough

package.json
HarelM: Yes, I'll do it now. I also saw you added it to the to do list, so no need to keep this review comment open.

src/source/worker_tile.ts
HarelM: hasDependencies and instanceof seems like they serve a similar objective, can we improve hasDependencies to be an enum maybe and avoid the need for instanceof?

HarelM:
Can you fix the lint warning please?

src/data/bucket.ts
HarelM: This looks very very similar to the following type:
https://github.com/maplibre/maplibre-style-spec/blob/5f3539929fa2a001eb073502b05d6d3eaea03fd2/src/expression/index.ts#L43

Is there a way to reduce duplication?
At the very least I would consider defining the `min, mid, max` as a type to be used in all these places and replace the `[_: string]` with `Record` of this type.

src/data/bucket/line_bucket.ts
```diff
@@ -246,7 +255,7 @@ export class LineBucket implements Bucket {
         }
     }
 
-    addFeature(feature: BucketFeature, geometry: Array<Array<Point>>, index: number, canonical: CanonicalTileID, imagePositions: {[_: string]: ImagePosition}, subdivisionGranularity: SubdivisionGranularitySetting) {
+    addFeature(feature: BucketFeature, geometry: Array<Array<Point>>, index: number, canonical: CanonicalTileID, imagePositions: {[_: string]: ImagePosition}, dashPositions: {[_: string]: DashEntry}, subdivisionGranularity: SubdivisionGranularitySetting) {
```
HarelM: `Record` here too?

src/data/bucket/line_bucket.ts
```diff
@@ -594,6 +603,50 @@ export class LineBucket implements Bucket {
...
+        }
+        return false;
+    }
+
+    private addLineDashDependencies(layers: Array<LineStyleLayer>, bucketFeature: BucketFeature, zoom: number, options: PopulateParameters) {
+        const dashes = options.dashDependencies;
+
+        for (const layer of layers) {
+            const dasharrayProperty = layer.paint.get('line-dasharray');
+            if (dasharrayProperty && dasharrayProperty.value.kind !== 'constant') {
```
HarelM: nit picking: can we invert the if to reduce indentation?

src/data/bucket/line_bucket.ts
```diff
@@ -594,6 +603,50 @@ export class LineBucket implements Bucket {
...
+                const max = {
+                    dasharray: dasharrayProperty.value.evaluate({zoom: zoom + 1}, bucketFeature, {}),
+                    round
+                };
+
+                const minKey = `${min.dasharray.join(',')},${min.round}`;
+                const midKey = `${mid.dasharray.join(',')},${mid.round}`;
+                const maxKey = `${max.dasharray.join(',')},${max.round}`;
+
+                dashes[minKey] = min;
```
HarelM: Does this update `options.dashDependencies`? If so, I think it should use this object instead of a reference as it's not clear in this line what is being updated...

src/data/program_configuration.ts
lucaswoj: Done

src/data/bucket.ts
lucaswoj: Done

src/data/bucket/line_bucket.ts
lucaswoj: done

src/data/bucket/line_bucket.ts
lucaswoj: done

src/data/bucket/line_bucket.ts
lucaswoj: done

.github/workflows/test-all.yml
lucaswoj: I think we really do need this change. I can't get tests to pass without it. Tests are running in GitHub Actions on my fork, which cannot run this script that updates resources on the upstream Maplibre org repository.

lucaswoj:
I've done another round of code review changes. Again, I've left open any conversations that I wasn't able to obviously resolve.

src/data/bucket.ts
```diff
@@ -53,13 +54,7 @@ export type BucketFeature = {
...
         };
     };
-    readonly dashes?: {
-        [_: string]: {
-            'min': string;
-            'mid': string;
-            'max': string;
-        };
-    };
+    readonly dashes?: NonNullable<StyleFeature['dashes']>;
```
HarelM, line 57: What about using the same for patterns?

.github/workflows/test-all.yml
HarelM, line 220: This is probably the least problematic one as if all lines are covered this wouldn't fail, are you sure this creates problems?

src/data/bucket.ts
```diff
@@ -78,12 +82,12 @@ export type BucketFeature = {
...
 export interface Bucket {
     layerIds: Array<string>;
-    hasPattern: boolean;
+    hasDependencies: boolean;
     readonly layers: Array<any>;
     readonly stateDependentLayers: Array<any>;
     readonly stateDependentLayerIds: Array<string>;
     populate(features: Array<IndexedFeature>, options: PopulateParameters, canonical: CanonicalTileID): void;
-    update(states: FeatureStates, vtLayer: VectorTileLayer, imagePositions: {[_: string]: ImagePosition}): void;
+    update(states: FeatureStates, vtLayer: VectorTileLayer, imagePositions: {[_: string]: ImagePosition}, dashPositions: {[_: string]: DashEntry}): void;
```
HarelM: `Record`?

src/data/bucket.ts
lucaswoj: Refactoring the code that handles patterns is out of scope for this PR.

HarelM:
I've just released version 24.2.0 of the style-spec.

.github/workflows/test-all.yml
lucaswoj: I noticed that removing this `if:` condition in [d382706](https://github.com/maplibre/maplibre-gl-js/pull/5812/commits/d382706ce61cf775b87131d9cff81d5309e3faef) caused this step to fail.

- Our [code coverage reporting and enforcement step continues to work correctly](https://app.codecov.io/gh/maplibre/maplibre-gl-js/commit/39af6cd713694f50f3f77241a008abe7f0bc60de).
- The failure appears to be related to annotation permissions. Since this workflow is running from my fork, it doesn't have the necessary permissions to create annotations in the main MapLibre GL JS repository. The annotations are helpful but disabling them does not compromise our quality standards.
- Regarding the four statements marked as uncovered: these are actually covered by our render tests. I verified this by commenting them out and watching the render tests fail. This type of coverage reporting discrepancy is something we've seen before with our test setup. Addressing it would be outside this PR's scope.

Based on these factors, keeping the `if:` condition seems like the right approach here.

lucaswoj:
I've done another round of revisions, including updating `package.json` to reference the newly released style spec version.

birkskyum:
Is there a demo using this new feature anywhere?

lucaswoj:
> Is there a demo using this new feature anywhere? 

I don't have anything up online. Sorry!

HarelM:
I think the only thing left is the changelog. Can you please add to it and I'll merge this PR and release a new version.

lucaswoj:
Done

PR #5824, opened 2025-04-30
Disable geolocate control button when permission is denied and trackUserLocation is false

tsegarra:
If all three of the following conditions are true, then the geolocate control button should show as disabled, but doesn't:

1. The browser supports geolocation
2. The user denies geolocation permission when prompted
3. `trackUserLocation` is set to `false`

Instead, the button does nothing and appears unchanged. That's because the code for disabling the button only runs if `trackUserLocation` is `true`. I added a test for the new case and fixed it.


HarelM:
Thanks for taking the time to report and fix this issue.

HarelM (approved):

Issue #5834, opened 2025-05-02
Excessive tiles loaded in globe


NathanMOlson:
At z11, many more tiles are loaded by globe projection than mercator projection at some locations on the globe.

The bug is demonstrated by the failing unit tests here in #5835. ('zoom = 11, mid lat' and 'zoom = 11, high lat')

johncarmack1984:
Oriented-bounding-box culling in #5865 (v5.6.0) took care of this, tightened again in #7932. This can be closed.

PR #5835, opened 2025-05-02
Demonstrate #5834

NathanMOlson:
This PR contains failing unit tests that demonstrate #5834.

### DIFF
--- a/src/geo/projection/covering_tiles.test.ts
+++ b/src/geo/projection/covering_tiles.test.ts
@@ -65,11 +65,7 @@ describe('coveringTiles', () => {
                 new OverscaledTileID(3, 0, 3, 2, 3),
                 new OverscaledTileID(3, 0, 3, 2, 4),
                 new OverscaledTileID(3, 0, 3, 5, 3),
-                new OverscaledTileID(3, 0, 3, 5, 4),
-                new OverscaledTileID(3, 0, 3, 2, 2),
-                new OverscaledTileID(3, 0, 3, 2, 5),
-                new OverscaledTileID(3, 0, 3, 5, 2),
-                new OverscaledTileID(3, 0, 3, 5, 5),
+                new OverscaledTileID(3, 0, 3, 5, 4)
             ]);
         });
     
@@ -108,10 +104,8 @@ describe('coveringTiles', () => {
     
             expect(tiles).toEqual([
                 new OverscaledTileID(7, 0, 7, 64, 64),
-                new OverscaledTileID(7, 0, 7, 65, 63),
                 new OverscaledTileID(7, 0, 7, 64, 63),
                 new OverscaledTileID(7, 0, 7, 63, 63),
-                new OverscaledTileID(7, 0, 7, 64, 62),
                 new OverscaledTileID(10, 0, 10, 510, 512),
                 new OverscaledTileID(10, 0, 10, 511, 512),
                 new OverscaledTileID(10, 0, 10, 511, 513)
--- a/src/geo/projection/covering_tiles.test.ts
+++ b/src/geo/projection/covering_tiles.test.ts
@@ -171,6 +171,63 @@ describe('coveringTiles', () => {
                 new OverscaledTileID(0, 0, 0, 0, 0)
             ]);
         });
+
+        test('zoom = 11', () => {
+            const transform = new GlobeTransform();
+            transform.resize(128, 128);
+            transform.setZoom(11);
+            transform.setPitch(0);
+            transform.setCenter(new LngLat(-179.73, -0.087));
+    
+            const tiles = coveringTiles(transform, {
+                tileSize: 512,
+                minzoom: 0,
+                maxzoom: 15,
+                reparseOverscaled: true
+            });
+    
+            expect(tiles).toEqual([
+                new OverscaledTileID(11, 0, 11, 1, 1024)
+            ]);
+        });
+        
+        test('zoom = 11, mid lat', () => {
+            const transform = new GlobeTransform();
+            transform.resize(128, 128);
+            transform.setZoom(11);
+            transform.setPitch(0);
+            transform.setCenter(new LngLat(-179.73, 60.02));
+    
+            const tiles = coveringTiles(transform, {
+                tileSize: 512,
+                minzoom: 0,
+                maxzoom: 15,
+                reparseOverscaled: true
+            });
+    
+            expect(tiles).toEqual([
+                new OverscaledTileID(11, 0, 11, 1, 594)
+            ]);
+        });
+        
+        test('zoom = 11, high lat', () => {
+            const transform = new GlobeTransform();
+            transform.resize(128, 128);
+            transform.setZoom(11);
+            transform.setPitch(0);
+            transform.setCenter(new LngLat(-179.73, 85.028));
+    
+            const tiles = coveringTiles(transform, {
+                tileSize: 512,
+                minzoom: 0,
+                maxzoom: 15,
+                reparseOverscaled: true
+            });
+    
+            expect(tiles).toEqual([
+                new OverscaledTileID(11, 0, 11, 1, 1)
+            ]);
+        });
     });
 
     describe('mercator', () => {
@@ -449,6 +506,60 @@ describe('coveringTiles', () => {
             transform.setZoom(8);
             expect(coveringTiles(transform, options)).toEqual([
                 new OverscaledTileID(0, 0, 0, 0, 0)
+            ]);
+        });
+        
+        test('z11', () => {
+            const options = {
+                minzoom: 1,
+                maxzoom: 15,
+                tileSize: 512,
+                reparseOverscaled: true
+            };
+        
+            const transform = new MercatorTransform(0, 15, 0, 85, true);
+            transform.resize(128, 128);
+            transform.setZoom(11);
+            transform.setCenter(new LngLat(-179.73, -0.087));
+        
+            expect(coveringTiles(transform, options)).toEqual([
+                new OverscaledTileID(11, 0, 11, 1, 1024)
+            ]);
+        });
+        
+        test('z11, mid lat', () => {
+            const options = {
+                minzoom: 1,
+                maxzoom: 15,
+                tileSize: 512,
+                reparseOverscaled: true
+            };
+        
+            const transform = new MercatorTransform(0, 15, 0, 85, true);
+            transform.resize(128, 128);
+            transform.setZoom(11);
+            transform.setCenter(new LngLat(-179.73, 60.02));
+        
+            expect(coveringTiles(transform, options)).toEqual([
+                new OverscaledTileID(11, 0, 11, 1, 594)
+            ]);
+        });
+        
+        test('z11, high lat', () => {
+            const options = {
+                minzoom: 1,
+                maxzoom: 15,
+                tileSize: 512,
+                reparseOverscaled: true
+            };
+        
+            const transform = new MercatorTransform(0, 15, 0, 85, true);
+            transform.resize(128, 128);
+            transform.setZoom(11);
+            transform.setCenter(new LngLat(-179.73, 85.028));
+        
+            expect(coveringTiles(transform, options)).toEqual([
+                new OverscaledTileID(11, 0, 11, 1, 1)
             ]);
         });
     
--- a/src/geo/projection/covering_tiles.test.ts
+++ b/src/geo/projection/covering_tiles.test.ts
@@ -228,6 +228,44 @@ describe('coveringTiles', () => {
                 new OverscaledTileID(11, 0, 11, 1, 1)
             ]);
         });
+
+        test('zoom = 11, mid lat, mid lng', () => {
+            const transform = new GlobeTransform();
+            transform.resize(128, 128);
+            transform.setZoom(11);
+            transform.setPitch(0);
+            transform.setCenter(new LngLat(-58.97, 60.02));
+    
+            const tiles = coveringTiles(transform, {
+                tileSize: 512,
+                minzoom: 0,
+                maxzoom: 15,
+                reparseOverscaled: true
+            });
+    
+            expect(tiles).toEqual([
+                new OverscaledTileID(11, 0, 11, 688, 594)
+            ]);
+        });
+        
+        test('zoom = 11, mid lng', () => {
+            const transform = new GlobeTransform();
+            transform.resize(128, 128);
+            transform.setZoom(11);
+            transform.setPitch(0);
+            transform.setCenter(new LngLat(-58.97, -0.087));
+    
+            const tiles = coveringTiles(transform, {
+                tileSize: 512,
+                minzoom: 0,
+                maxzoom: 15,
+                reparseOverscaled: true
+            });
+    
+            expect(tiles).toEqual([
+                new OverscaledTileID(11, 0, 11, 688, 1024)
+            ]);
+        });
     });
 
     describe('mercator', () => {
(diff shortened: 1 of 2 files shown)


src/geo/projection/covering_tiles.test.ts
```diff
@@ -171,6 +171,63 @@ describe('coveringTiles', () => {
                 new OverscaledTileID(0, 0, 0, 0, 0)
             ]);
         });
+
+        test('zoom = 11', () => {
```
NathanMOlson, line 175: This test passes.

src/geo/projection/covering_tiles.test.ts
```diff
@@ -171,6 +171,63 @@ describe('coveringTiles', () => {
...
+                maxzoom: 15,
+                reparseOverscaled: true
+            });
+    
+            expect(tiles).toEqual([
+                new OverscaledTileID(11, 0, 11, 1, 1024)
+            ]);
+        });
+        
+        test('zoom = 11, mid lat', () => {
```
NathanMOlson, line 194: This test fails. It loads one z11 tile and one z7 tile, where Mercator only loads one z11 tile.

src/geo/projection/covering_tiles.test.ts
```diff
@@ -171,6 +171,63 @@ describe('coveringTiles', () => {
...
+                maxzoom: 15,
+                reparseOverscaled: true
+            });
+    
+            expect(tiles).toEqual([
+                new OverscaledTileID(11, 0, 11, 1, 594)
+            ]);
+        });
+        
+        test('zoom = 11, high lat', () => {
```
NathanMOlson, line 213: This test fails. It loads 18 tiles from z2 to z11, where Mercator only loads one z11 tile.

NathanMOlson (commented):
At this high zoom level, I would expect globe and mercator to load identical (or almost identical) tiles.

src/geo/projection/covering_tiles.test.ts
msbarry: Also if you add a test for lat~=60 lng ~=-60 it loads 11 tiles up to z1 instead of 1 tile at z11:

```ts
        test.only('zoom = 11, mid lat, mid lng', () => {
            const transform = new GlobeTransform();
            transform.resize(128, 128);
            transform.setZoom(11);
            transform.setPitch(0);
            transform.setCenter(new LngLat(-59, 60.02));
    
            const tiles = coveringTiles(transform, {
                tileSize: 512,
                minzoom: 0,
                maxzoom: 15,
                reparseOverscaled: true
            });
    
            expect(tiles).toEqual([
                new OverscaledTileID(11, 0, 11, 688, 594)
            ]);
        });
```

msbarry:
@NathanMOlson do you think you'd be able to fix this issue?  From what I can tell there are 2 things going on:

1. `createCalculateTileZoomFunction` appears to be setting `thisTileDesiredZ` to low values for some tiles at mid-to-high latitudes but no pitch, so a lot of low zoom tiles get pulled in
2. the aabb/frustum intersection test for whether to include a tile in globe view is too broad, causing tiles that are not visible at a given zoom level to get included

I don't have a great handle on exactly what either of those are doing... it looks like for (2) an [oriented bounding box](https://github.com/Esri/i3s-spec/blob/master/docs/1.7/obb.cmn.md) might be a better fit?

NathanMOlson:
Thanks @msbarry. The job of `calculateTileZoom()` is not the determine whether a tile should be displayed, but rather "Given the distance to this candidate tile, what it the appropriate zoom level to display it at?".

I seems to me that `calculateTileZoom()` is working as desired, and the problem is in `isTileVisible()`. I think if we solve your (2), the problem will be resolved.

src/geo/projection/vertical_perspective_transform.ts
```diff
@@ -448,7 +448,9 @@ export class VerticalPerspectiveTransform implements ITransform {
         const globeMatrixUncorrected = createMat4f64();
         if (this._helper.autoCalculateNearFarZ) {
             this._helper._nearZ = 0.5;
-            this._helper._farZ = this.cameraToCenterDistance + globeRadiusPixels * 2.0; // just set the far plane far enough - we will calculate our own z in the vertex shader anyway
+            const cameraAltitudePixels = this.cameraToCenterDistance*Math.cos(this.pitchInRadians);
+            // Set the farZ at the horizon
+            this._helper._farZ = this.cameraToCenterDistance / cameraAltitudePixels * (globeRadiusPixels + cameraAltitudePixels - globeRadiusPixels * globeRadiusPixels / (globeRadiusPixels + cameraAltitudePixels));
```
NathanMOlson, line 453: @msbarry This change improves things, 2 of the 4 failing unit tests are passing, and the other 2 load fewer excess tiles.

NathanMOlson: Note this change breaks render tests related to extrusions. Maybe this could be applied to only the frustum used by `isTileVisible()`.

NathanMOlson:
I think a good almost-solution would be to rewrite `isTileVisible()` to look like this:

```
export function isTileVisible(frustum: Frustum, aabb: Aabb, plane?: vec4): IntersectionResult {
    const clippedFrustum = clipFrustum(frustum, plane);
    return aabb.intersectsFrustum(clippedFrustum);
}
```

Just needs `clipFrustum()` to be written...

msbarry:
> The job of `calculateTileZoom()` is not the determine whether a tile should be displayed, but rather "Given the distance to this candidate tile, what it the appropriate zoom level to display it at?".
> 
> I seems to me that `calculateTileZoom()` is working as desired, and the problem is in `isTileVisible()`. I think if we solve your (2), the problem will be resolved.

OK got it, so calculateTileZoom() is just returning weird results for tiles it never should have been called for because isTileVisible was returning true too often?

I'm just wondering if globe view should even be using `calculateTileZoom` when there's no pitch?  I see `MercatorCoveringTilesDetailsProvider` does:

https://github.com/maplibre/maplibre-gl-js/blob/9c74cefe6a5bd9e58d1238018d9033767abb1ce5/src/geo/projection/mercator_covering_tiles_details_provider.ts#L42-L46

But `GlobeCoveringTilesDetailsProvider` does:

https://github.com/maplibre/maplibre-gl-js/blob/9c74cefe6a5bd9e58d1238018d9033767abb1ce5/src/geo/projection/globe_covering_tiles_details_provider.ts#L96-L98

msbarry:
> I think a good almost-solution would be to rewrite `isTileVisible()` to look like this:
> 
> ```
> export function isTileVisible(frustum: Frustum, aabb: Aabb, plane?: vec4): IntersectionResult {
>     const clippedFrustum = clipFrustum(frustum, plane);
>     return aabb.intersectsFrustum(clippedFrustum);
> }
> ```
> 
> Just needs `clipFrustum()` to be written...

Would that just be computing a new frustum with farZ overridden to the horizon? 

Another idea I had was unprojecting corners of the tile to screen coordinates to get a screen x/y bounding box and see if it intersects the visible viewport...not sure how that compares to this for performance or accuracy though.

NathanMOlson:
> OK got it, so calculateTileZoom() is just returning weird results for tiles it never should have been called for because isTileVisible was returning true too often?

Correct.

> I'm wondering if globe view should even be using `calculateTileZoom` when there's no pitch? I see `MercatorCoveringTilesDetailsProvider` does:
> 
> https://github.com/maplibre/maplibre-gl-js/blob/9c74cefe6a5bd9e58d1238018d9033767abb1ce5/src/geo/projection/mercator_covering_tiles_details_provider.ts#L42-L46
> 
> But `GlobeCoveringTilesDetailsProvider` does:
> 
> https://github.com/maplibre/maplibre-gl-js/blob/9c74cefe6a5bd9e58d1238018d9033767abb1ce5/src/geo/projection/globe_covering_tiles_details_provider.ts#L96-L98
> 
> with no pitch check?

Not allowing variable tiles in globe with low pitch is definitely possible, and as a non-globe-user I don't have an opinion on it. The reason the pitch check exists in Mercator was to preserve legacy behavior, while globe was unencumbered by historical precedent :) But the issue would still exist when the map is pitched.

NathanMOlson:
> Would that just be computing a new frustum with farZ overridden to the horizon like you did in this PR but just isolated for this use-case?

I over-simplified, the intersection of a frustum and a plane is not another frustum in general. It could be nothing, or a 4, 5, 6, or 7-sided polyhedron. :( This would make the implementation a little tricky. Maybe there's a way to do the intersection without explicitly calculating this polyhedron. But in general, yes, just cut the frustum at the horizon and throw away whatever part of it is behind the horizon.

> Another idea I had was unprojecting corners of the tile to screen coordinates to get a screen x/y bounding box and see if it intersects the visible viewport...not sure how that compares to this for performance or accuracy though.

Sounds like a good idea. I'm also not sure how it would compare in performance/accuracy.

msbarry:
> > Another idea I had was unprojecting corners of the tile to screen coordinates to get a screen x/y bounding box and see if it intersects the visible viewport...not sure how that compares to this for performance or accuracy though.
> 
> Sounds like a good idea. I'm also not sure how it would compare in performance/accuracy.

I tried testing 9 points of the tile (each corner, edge middles, and center) and it does appear to work and reduce the set of ideal tiles pretty aggressively, but it is more expensive.  On my m1 macbook `coveringTiles` was taking about 0.05ms (it's called a lot so it adds up) but with that check after `isTileVisible` it went up 5x to 0.25ms.  Another alternative is putting it as a post-filter before adding a tile ID to the result, that only increased by 2x to 0.1ms. The more robust tile/frustum check seems like the better approach here although I may use this as a quick fix for this until we can get this more robust check is in place.

NathanMOlson:
> Ah so you're saying that if you do a full clip of the frustum instead of just pulling the far-z plane back then you think it would reduce the matching tiles that were still present in those 2 failing test cases?

All the failing unit tests I added have pitch = 0. In this case, doing the full clip is equivalent to just pulling the far z-plane back.

msbarry:
I tried an approach that creates a tighter bounding polyhedron for each tile then does a full SAT intersection test between the camera frustum and the tile polyhedron.   This fixes those remaining test and tightens up a few others to produce even smaller ideal tile sets, unfortunately it's pretty expensive, if I replace intersectsFrustum check with this then it takes ~1ms (20x slower).  If I run it at as a last check before emiitting the ideal tile then it's only 6x slower (0.3ms)

<summary>

<details>

```ts
    getTileFrustum(tileID: { x: number; y: number; z: number }, wrap: number, elevation: number, options: CoveringTilesOptions) {
        if (tileID.z <= 1) {
            return Frustum.fromAabb(this.getTileAABB(tileID, wrap, elevation, options));
        } else {
            // Compute frustum using the 4 corners and center
            const center = projectTileCoordinatesToSphere(EXTENT / 2, EXTENT / 2, tileID.x, tileID.y, tileID.z);
            const near = [
                projectTileCoordinatesToSphere(0, 0, tileID.x+1, tileID.y, tileID.z),
                projectTileCoordinatesToSphere(0, 0, tileID.x, tileID.y, tileID.z),
                projectTileCoordinatesToSphere(0, 0, tileID.x, tileID.y+1, tileID.z),
                projectTileCoordinatesToSphere(0, 0, tileID.x+1, tileID.y+1, tileID.z),
            ];
            const farScale = 1 / Math.min(...near.map(point => vec3.dot(point, center)));
            const far = near.map(point => vec3.scale(vec3.create(), point, farScale));
            // not really a frustum, just using frustum to hold the 6 sides and 8 vertices
            return Frustum.fromPoints([
                ...near.map(vec => vec4.fromValues(vec[0], vec[1], vec[2], 1)),
                ...far.map(vec => vec4.fromValues(vec[0], vec[1], vec[2], 1))
            ]);
        }
    }

    intersects(other: Frustum): IntersectionResult {
        if (this.aabb.disjoint(other.aabb)) return IntersectionResult.None;
        const axes: vec3[] = [...this._planeNormals, ...other._planeNormals];
        Frustum.addEdgeCrossProducts(this, other, axes);
        const pointsA = this.points.map(p => vec3.fromValues(p[0], p[1], p[2]));
        const pointsB = other.points.map(p => vec3.fromValues(p[0], p[1], p[2]));
        let fullyInside = true;
        for (let i = 0; i < axes.length; i++) {
            const axis = axes[i];
            let minA = Infinity;
            let maxA = -Infinity;
            for (let i = 0; i < pointsA.length; i++) {
                const point = pointsA[i];
                const projected = vec3.dot(point, axis);
                minA = Math.min(projected, minA);
                maxA = Math.max(projected, maxA);
            }
            let minB = Infinity;
            let maxB = -Infinity;
            for (let i = 0; i < pointsB.length; i++) {
                const point = pointsB[i];
                const projected = vec3.dot(point, axis);
                minB = Math.min(projected, minB);
                maxB = Math.max(projected, maxB);
            }
            if (maxB < minA || maxA < minB) {
                return IntersectionResult.None;
            }
            if (fullyInside && (minB < minA || maxB > maxA)) {
                fullyInside = false;
            }
        }

        return fullyInside ? IntersectionResult.Full : IntersectionResult.Partial;
    }

    private static addEdgeCrossProducts(a: Frustum, b: Frustum, axes: vec3[]) {
        for (const edgeA of a._edgeVectors) {
            for (const edgeB of b._edgeVectors) {
                const crossed = vec3.cross(vec3.create(), edgeA, edgeB);
                const mag = vec3.dot(crossed, crossed);
                if (mag > 1e-12) {
                    vec3.normalize(crossed, crossed);
                    axes.push(crossed);
                }
            }
        }
    }
```
</details>
</summary>

Also I noticed that the aabb intersectsFrustum check is too loose in some cases...  in `coveringTiles > globe > pitched+rotated` test the frustum/aabb look like this, but that method returns "partial overlap"

image

and for `coveringTiles > globe > zoom = 11, mid lat, mid lng` it returns partial overlap, but the shapes look like this:

image

kubapelc:
Hi, I knew about AABBs being a poor bounding volume for globe tiles, I haven't gotten around to fixing it yet, sadly. I've experimented with computing the exact intersection, but it ended up being too slow and I didn't get the math quite right.

Another thing I wanted to try eventually is just using oriented bounding boxes. The intersection itself should still be very fast. Computing fitting OBBs would be more expensive than AABBs, but it could be done once per tile and then cached - a similar cache is already in place. OBBs would still be a poor fit at very low zooms, but the higher the zoom, the flatter the tile's actual shape would be and the better it would be approximated with an OBB. There would still be some false positives, but there should be a lot fewer of them.

I'll try to implement this in a few days and get back to you.

msbarry:
Thanks! Also do you know if there's a straightforward way to clip the frustum  based on the horizon/horizon plane just for `coveringTiles`? https://github.com/maplibre/maplibre-gl-js/pull/5835#issuecomment-2848408757

The other thing about an OBB is that the side of a tile closer to the pole is shorter than the side closer to the equator so it ends up bleeding into the adjacent tile a bit at lower zoom/higher latitudes 🤔 

My patch that checks if the projected corners of the tile intersect the visible screen brought the p90 time that workers spend processing each from 1.5s down to 1s (was 900ms for mercator tiles) so that fixes my issue in the short term, but I think that only works because my site doesn't support pitch/rotate so this would be a much better fix.

kubapelc:
I've uploaded the debug view / OBB demo:

https://kvaleya.gitlab.io/maplibre/obb-demo/

It is just a custom layer example modified to access the bounding boxes and draw them. It is a terrible hack. One draw call per drawn line, since I was too lazy to use a proper vertex buffer 😅

> I think that with the OBB improvement we wouldn't need to worry about a unclipped frustum intersecting the back-side of bounding boxes for adjacent tiles when pitch=0 like is happening with AABBs now?

Yes, that should be much less of an issue. Adjacent tile that is slightly beyond screen edge should now have its OBB detected to lie entirely on the wrong side of the plane frustum for that given screen edge, and thus be rejected. Some will likely still get drawn if they are very near the screen edge, since the OBBs still have some depth, but should be much less that with AABBs I hope.

kubapelc:
@msbarry

I think I now understand your suggestion with moving the frustum's far plane according to the horizon clipping plane. I found a case where it is clearly visible how it could be helpful.

This is what the camera sees:

image

These are the tiles that are considered visible (using OBBs and the original frustum):

image

Note the 3 tiles on the opposite side of the world. They really do intersect the frustum, but are hidden in reality.
 
This is what is considered visible with your suggestion implemented. The frustum's far plane is moved as close to the camera as possible while still being entirely behind the horizon plane. 

image

msbarry:
Awesome! I think there's also the case when the camera is pitched and the sky is visible that the frustum can extend above the horizon and intersect a corner of the OBB way outside of the unit sphere

NathanMOlson:
Resolved by #5865

Issue #5829, opened 2025-04-30
opacityWhenCovered does not work when both globe and terrain are enabled


acalcutt:
I am creating a marker with something like this code, which includes opacityWhenCovered
```
    const marker = new maplibregl.Marker({
            opacityWhenCovered: 0.2
        })
        .setLngLat([137.9150899566626, 36.25956997955441])
        .addTo(map);
```
This option works if I have just globe enabled, but if I also enable terrain the transparency no longer works

For example,

with just the globe enabled it works
![Image](https://github.com/user-attachments/assets/fba17cbd-31db-4042-ba5a-c94b36bb1807)

with terrain also enable it does not work
![Image](https://github.com/user-attachments/assets/8b5f7e95-233d-4d71-9732-05d621614a71)

I will work on making an example page later today

acalcutt:
Here is an example. It is basically just the globe example with terrain and a marker added
https://stackblitz.com/edit/web-platform-6xsbznqb?file=index.html

By default terrain is on, so it you spin the globe so the marker should go behind you can see the opacity does not get applied.  If you click the terrain button to turn off terrain, the opacity works

HarelM:
Well... Globe and terrain aren't good friends.
Feel free to open a PR to solve this.

acalcutt:
Well that is something that would be good to improve. Maybe a candidate for some terrain bounty (since i know someone was asking in slack)

HarelM:
I have replied to anyone who looked for bounties, unfortunately, there was never a follow up...

PR #5838, opened 2025-05-03
fix: opacityWhenCovered not working to hide the marker behind the globe if terrain is enabled

acalcutt:
This PR fixes https://github.com/maplibre/maplibre-gl-js/issues/5829

Looking into the issue I found when terrain is enabled, _opacityWhenCovered is only applied when terrain is blocking and not if occluded (which if i understand correctly is when the tile is not visible)

This reworks the opacity so it always first checks if the marker is 'occluded ' first. If it is, the _opacityWhenCovered  value is always applied. If it is not, it does the existing checks to see if it behind terrain.

## Example (rotate the map so the marker goes behind the globe with terrain enabled)
Example with v5.4.0
https://stackblitz.com/edit/web-platform-6xsbznqb?file=index.html
image

Example with this PR code
https://stackblitz.com/edit/web-platform-mrytnhfv?file=index.html 
image

## Other Issues Found (not fixed by this PR)

When globe is enabled and terrain is enabled, it does not apply _opacityWhenCovered  when the marker is behind terrain. this is not working in the current release or in this PR. I noticed centerIsInvisible was returning false. I also notice the marker tended to float and not be in a consistent location when globe+terrain is enabled

opacity_terrain-globe

## Launch Checklist

 - [X] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -316,6 +316,7 @@
 - Fixed a performance regression related to symbol placement ([#4599](https://github.com/maplibre/maplibre-gl-js/pull/4599))
 - Fix a bug where cloning a Transform instance didn't include the `lngRange`. This caused a bug where
 using `transformCameraUpdate` caused the `maxBounds` to stop working just for east/west bounds. ([#4625](https://github.com/maplibre/maplibre-gl-js/pull/4625))
+- Fix opacityWhenCovered not working to hide the marker behind the globe if terrain is enabled. ([#5838](https://github.com/maplibre/maplibre-gl-js/pull/5838))
 
 ## 4.6.0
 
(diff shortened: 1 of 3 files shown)


HarelM:
Please add a test to make sure the scenario that you have fixes is covered by a test, and also check out the failing tests in the CI run.
I think there's also an if that end in a return, so "else" is redundant there, right?

acalcutt:
@HarelM  The fix is now just a few lines, basically moving the 'occluded' variable up a few lines and adding to the if statement.

I added a test which seems like it works now as I would expect, and failed when i tried to revert the code change from this PR.

HarelM:
Thanks! This looks pretty straight forward, can you please add a changelog entry?

acalcutt:
Changelog has been added.

I am mainly looking into the globe + terrain right now because we recently upgraded tileserver-gl to use maplibre-gl-js v5 so we could fix issues with the RTL plugin. When we updated to v5 I asked that globe buttons be added so we could show off that new feature.  So now the globe button is shown on this new [terrain preview page](https://tiles.wifidb.net/data/preview/jaxa_terrainrgb/#1.53/41.9/-31.3) which allows you to click the terrain and set a marker to get the elevation of that spot. Because there terrain is enabled by default, I at first was trying to set opacityWhenCovered but realized that by default it should be 0.2 and just wasn't working.

HarelM (approved):

PR #5856, opened 2025-05-06
Fix `queryRenderedFeatures` bounding box crossing antimeridian in globe view

### LINKED ISSUE #5381: map.queryRenderedFeatures with bbox around anti-meridian returns strange results
**maplibre-gl-js version**: 5.0.0

**browser**:

### Steps to Trigger Behavior

 1. Add random polygons to map
 2. Call map.queryRenderedFeatures with bbox based on e.point and some offset
 3. Click on polygons near anti-meridian

### Link to Demonstration
https://jsbin.com/dogozitaqe/1/edit?html,output

### Expected Behavior
Features are correctly returned from this function

### Actual Behavior
Some features are returned correctly, some are simply ignored. When clicking on antimeridian all features around globe on given latitude are returned.

I see 2 issues:
1. Polygon at the bottom of the screen is not returned by queryRenderedFeatures
2. When I click on the anti-meridian, query is performed for features around the globe instead on bbox near anti-meridian.

https://github.com/user-attachments/assets/57979a67-6801-4c76-8a99-3cd8a8728ddb

msbarry:
Fix #5381 by checking when a `queryRenderedFeatures` bounding box crosses the antimeridian and splitting it into 2 bounding boxes that overhang the left and right sides of the tile bordering the antimeridian.

This also adds a new utility `Bounds` that holds an x/y bounding box. This PR only uses it in a single spot, but there appear to be a lot of places in the code that do this manually and could benefit from combining into a single utility in subsequent PRs.

Behavior before (queryRenderedFeatures on mousemove within the red box):

https://github.com/user-attachments/assets/47bd4246-29ac-4777-a413-52f23dbe5c8a

Behavior after:

https://github.com/user-attachments/assets/67e53ff1-4c6f-4c10-bcc5-606d7681cbef

Behavior after (pitched and rotated):

https://github.com/user-attachments/assets/37cf6413-ee1e-4962-8c9f-57f021533725

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


msbarry:
Noticed an issue while gathering some screencasts when pitched/rotated, will reopen after that's fixed.

src/geo/bounds.ts
```diff
@@ -0,0 +1,38 @@
+import Point from '@mapbox/point-geometry';
+
+type PointLike = {x: number; y: number};
+
+export default class Bounds {
```
HarelM, line 5: Use name imports/export instead of default.

src/geo/bounds.ts
```diff
@@ -0,0 +1,38 @@
+import Point from '@mapbox/point-geometry';
+
+type PointLike = {x: number; y: number};
```
HarelM, line 3: Isn't this type defined somewhere already? `Point2D` maybe?

src/geo/bounds.ts
HarelM: Is this the same as `GeoJSON.BBox` in terms of the data it holds? Maybe we can use that internally or simply supply some utility methods to BBox?

src/source/source_cache.test.ts
```diff
@@ -1577,7 +1579,104 @@ describe('SourceCache#tilesIn', () => {
...
+                [new OverscaledTileID(1, 0, 1, 1, 0).key, [
+                    new Point(7219, 7261),
+                    new Point(8937, 7267),
+                    new Point(8913, 8895),
+                    new Point(7251, 8899),
+                    new Point(7219, 7261),
+                ]]
+            ]);
+
+        transform.setCenter(new LngLat(-179.9, 0.1));
```
HarelM, line 1637: Should this part be in a separate test? It feels odd that there's a huge `expect` in the middle of the test...

src/source/source_cache.ts
```diff
@@ -24,6 +24,7 @@ import type {MapSourceDataEvent} from '../ui/events';
 import type {Terrain} from '../render/terrain';
 import type {CanvasSourceSpecification} from './canvas_source';
 import {coveringTiles, coveringZoomLevel} from '../geo/projection/covering_tiles';
+import Bounds from '../geo/bounds';
```
HarelM, line 27: Use named import.

src/source/source_cache.ts
```diff
@@ -979,62 +980,82 @@ export class SourceCache extends Evented {
...
-            ];
+            for (const tileID of tileIDs) {
+
+                const tileSpaceBounds = [
+                    tileID.getTilePoint(new MercatorCoordinate(cameraBounds.minX, cameraBounds.minY)),
+                    tileID.getTilePoint(new MercatorCoordinate(cameraBounds.maxX, cameraBounds.maxY))
+                ];
 
-            if (tileSpaceBounds[0].x - queryPadding < EXTENT && tileSpaceBounds[0].y - queryPadding < EXTENT &&
+                if (tileSpaceBounds[0].x - queryPadding < EXTENT && tileSpaceBounds[0].y - queryPadding < EXTENT &&
```
HarelM, line 1014: Can this if be negated to reduce some indentation?

src/source/source_cache.ts
```diff
@@ -979,62 +980,82 @@ export class SourceCache extends Evented {
...
 
         for (let i = 0; i < ids.length; i++) {
             const tile = this._tiles[ids[i]];
             if (tile.holdingForFade()) {
                 // Tiles held for fading are covered by tiles that are closer to ideal
                 continue;
             }
-            const tileID = transform.getCoveringTilesDetailsProvider().allowWorldCopies() ? tile.tileID : tile.tileID.unwrapTo(0);
+            // if the projection does not render world copies then we need to explicitly check for the bounding box crossing the antimeridian in each direction
+            const tileIDs = allowWorldCopies ? [tile.tileID] : [tile.tileID.unwrapTo(-1), tile.tileID.unwrapTo(0), tile.tileID.unwrapTo(1)];
```
HarelM, line 1003: The previous fix you did was here also, right?

HarelM:
Thanks!
Is there a `queryRenderedFeatures` tests that needs to be added (those with the json files) as well?

src/geo/bounds.ts
msbarry: `GeoJson.BBox` is:

```ts
export type BBox = [number, number, number, number] | [number, number, number, number, number, number];
```

That's a little less clear to work with in code and that type doesn't seem to be referenced in maplibre code at all. 
 If that representation were very widespread in maplibre's code I might use that, but I have seen a lot of places where they bounding boxes get passed around as {minX, minY, maxX, maxY}

src/source/source_cache.test.ts
```diff
@@ -1588,7 +1687,9 @@ describe('SourceCache#tilesIn', () => {
             tile.state = 'loaded';
         };
 
-        await waitForEvent(sourceCache, 'data', e => e.sourceDataType === 'metadata');
+        const dataPromise = waitForEvent(sourceCache, 'data', e => e.sourceDataType === 'metadata');
+        sourceCache.onAdd(undefined);
```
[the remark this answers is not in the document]
msbarry: the test wasn't actually doing anything since it started with `test('globe wrap', () => async () => {` - but when I fixed that it hung forever on `waitForEvent`. I think that sourceCache.onAdd  was needed to trigger the event

src/source/source_cache.ts
msbarry: I can do better! I can use the new `Bounds` utilities to do `.expandBy(-queryPadding).intersects()` test

msbarry:
> Thanks! Is there a `queryRenderedFeatures` tests that needs to be added (those with the json files) as well?

I was trying to figure out where I might put a test like that...
- style.test.ts mocks queryRenderedFeatures on the tile returned from tilesIn
- drag_pan.test.ts, map_events.test.ts, map_wuery_rendered_features.test.ts also mock queryRenderedFeatures
- query_features.test.ts mocks the result of SourceCache.tilesIn

so it doesn't look like there's an existing end to end place to add this to... any suggestions?

src/data/extent.ts
```diff
@@ -11,3 +14,8 @@
  * * This leaves us with 2^13 = 8192
  */
 export const EXTENT = 8192;
+
+/**
+ * The bounding box covering the entire extent.
+ */
+export const EXTENT_BOUNDS = Bounds.fromPoints([new Point(0, 0), new Point(EXTENT, EXTENT)]);
```
msbarry, line 21: Not sure if this is the best place since it's mutable. Maybe it should just be a constant in the one file it's used from (`source_cache.ts`)

src/geo/bounds.ts
HarelM: Fair enough.

HarelM:
I meant the tests here: 
https://github.com/maplibre/maplibre-gl-js/tree/19d33ec82ee33e095ccbb5c4453ad27ad6bea6be/test/integration/query

It might be worth adding a test there if it makes sense, your call.

src/geo/bounds.ts
```diff
@@ -0,0 +1,141 @@
+import Point from '@mapbox/point-geometry';
+
+type Point2D = {x: number; y: number};
```
HarelM, line 3: I meant to say that this type is already defined elsewhere, isn't it? Maybe as part of the style-spec library?

msbarry: Got it! It's `import {type Point2D} from '@maplibre/maplibre-gl-style-spec';`

msbarry:
> I meant the tests here: https://github.com/maplibre/maplibre-gl-js/tree/19d33ec82ee33e095ccbb5c4453ad27ad6bea6be/test/integration/query
> 
> It might be worth adding a test there if it makes sense, your call.

Got it, added 2 tests to reproduce the manual tests I ran for those screencasts in the description.

src/source/source_cache.ts
```diff
@@ -979,62 +982,77 @@ export class SourceCache extends Evented {
...
     }
 
+    private transformBbox(geom: Point[], project: (point: Point) => MercatorCoordinate, checkWrap: boolean): MercatorCoordinate[] {
+        let transformed = geom.map(project);
+        if (checkWrap) {
+            // If the projection does not allow world copies, then a bounding box may span the antimeridian and
+            // instead of a bounding box going from 179°E to 179°W, it goes from 179°W to 179°E and covers the entire
+            // planet except for what should be inside it.
+            const bounds = Bounds.fromPoints(geom);
+            bounds.expandBy(-Math.min(bounds.width(), bounds.height()) / 1e3);
```
HarelM, line 1041: 1e3? What does it stands for?

HarelM (approved):
THANKS!!!

src/data/extent.ts
HarelM: I think it's OK here...

src/source/source_cache.ts
msbarry: 1000... I needed some way to get a bounding box that's just a tiny bit smaller than the original one, so this shrinks it by 1/1000th of the minimum width or height.  `* 0.01` or `* 0.001` might be more clear?

HarelM: `expandBy` is confusing here I guess, although you are passing a negative number.
Maybe move this method to the bounds class to it will be more meaningful there (i.e. better name etc)?

HarelM: Or a comment, which I don't like in general, but might help here...

src/data/extent.ts
HarelM, line 21: Maybe this should be in a different file (of its own) to avoid including this for places where it's not needed? IDK...

src/source/source_cache.ts
msbarry: Like `shrinkBy(amount)` instead of `expandBy(-amount)` ?

HarelM: That might work, yes...

src/data/extent.ts
msbarry: OK, moved it to `extend_bounds.ts` and made a new `ReadOnlyBounds` interface that declares the non-mutating fields/methods of `Bounds` to make it harder for any place in the code to mess with the shared instance.

src/source/source_cache.ts
msbarry: Sounds good, added `shrinkBy()`

src/util/find_pole_of_inaccessibility.ts
```diff
@@ -17,25 +18,15 @@ export function findPoleOfInaccessibility(
...
-        if (!i || p.y < minY) minY = p.y;
-        if (!i || p.x > maxX) maxX = p.x;
-        if (!i || p.y > maxY) maxY = p.y;
-    }
+    const bounds = Bounds.fromPoints(polygonRings[0]);
 
-    const width = maxX - minX;
-    const height = maxY - minY;
-    const cellSize = Math.min(width, height);
+    const cellSize = Math.min(bounds.width(), bounds.height());
```
msbarry, line 23: This triggered the bundle size warning. It should eventually reduce bundle size though when we move more repeated code that computes a bounding box to the Bounds utility, so I just did a few of the refactors now instead of adjusting the bundle size limit.

HarelM: Nice! Thanks!

src/util/find_pole_of_inaccessibility.ts
```diff
@@ -17,25 +18,15 @@ export function findPoleOfInaccessibility(
...
-    let minX = Infinity, minY = Infinity, maxX = -Infinity, maxY = -Infinity;
-    const outerRing = polygonRings[0];
-    for (let i = 0; i < outerRing.length; i++) {
-        const p = outerRing[i];
-        if (!i || p.x < minX) minX = p.x;
-        if (!i || p.y < minY) minY = p.y;
-        if (!i || p.x > maxX) maxX = p.x;
-        if (!i || p.y > maxY) maxY = p.y;
-    }
+    const bounds = Bounds.fromPoints(polygonRings[0]);
```
HarelM, line 21: The code that was replaced here had some strange `!i` stuff, are you sure you kept the same logic?

src/util/find_pole_of_inaccessibility.ts
HarelM: Should we worry about performance degradation since these are no longer "inlined"? are these changes in a hot path that there's a need to worry about?

HarelM (approved):
Looks good otherwise, let me know what you think of my last remaining comments.

src/util/find_pole_of_inaccessibility.ts
msbarry: It looks like that code forces point 0 to always set minX = maxX = outerRing[0].x and minY = maxY = outerRing[0].y but that should already be happening since they are initialized as +/- Infinity.

src/util/find_pole_of_inaccessibility.ts
msbarry: I did a quick vitest benchmark of old vs. new code:

```ts
for (const npoints of [5, 10, 100]) {
    const points: Point[] = [];
    for (let i = 0; i < npoints; i++) {
        points.push(new Point(Math.random(), Math.random()));
    }

    bench(`a ${npoints}`, () => {
        let minX = Infinity;
        let minY = Infinity;
        let maxX = -Infinity;
        let maxY = -Infinity;
        for (const p of points) {
            minX = Math.min(minX, p.x);
            minY = Math.min(minY, p.y);
            maxX = Math.max(maxX, p.x);
            maxY = Math.max(maxY, p.y);
        }
        sum += (maxX - minX);
        sum += (maxY - minY);
    });

    bench(`b ${npoints}`, () => {
        const bounds = Bounds.fromPoints(points);
        sum += bounds.width();
        sum += bounds.height();
    });
}
```

and it printed out:

```
     name              hz     min     max    mean     p75     p99    p995    p999     rme   samples
   · a 5    23,955,438.04  0.0000  0.0870  0.0000  0.0000  0.0000  0.0001  0.0001  ±0.09%  11977720   fastest
   · b 5    19,908,041.24  0.0000  0.0265  0.0001  0.0000  0.0001  0.0001  0.0001  ±0.05%   9954021
   · a 10   21,024,862.40  0.0000  0.0264  0.0000  0.0000  0.0001  0.0001  0.0001  ±0.05%  10512432
   · b 10   17,949,599.10  0.0000  0.0327  0.0001  0.0001  0.0001  0.0001  0.0001  ±0.05%   8974800
   · a 100   7,893,015.18  0.0000  0.0334  0.0001  0.0001  0.0002  0.0002  0.0002  ±0.04%   3946508
   · b 100   7,484,713.22  0.0000  0.0310  0.0001  0.0001  0.0002  0.0002  0.0002  ±0.06%   3742357   slowest
```

So the `Bounds` class is about 20% slower for 5 points, 17% slower for 10, and 5% slower for 100 - both are cheap enough that I don't _think_ either are likely to be a bottleneck relative to other operations. If there's a specific operation you're concerned about I could benchmark that whole operation before/after this change.

msbarry: If there are use-case that need to call `extend` or `contains` with just an x,y then it might make sense to add versions of those APIs like `contains(x: number, y: number)` to save from them needing to allocate new point objects just to do that test. I didn't hit any of those in these couple of refactors though.

HarelM:
I'm good with merging it as is, let me know if your want me to wait or not...

msbarry:
I'm fine merging as-is unless there's any other tests you wanted me to run.

PR #5865, opened 2025-05-07
Improved tile frustum culling for globe

kubapelc:
Implements improved tile bounding volumes and frustum culling for globe projection. Refactors coveringTiles-related code to allow for usage of generic bounding volumes. Should reduce the amount of loaded and rendered tiles that are not visible.

More discussion: https://github.com/maplibre/maplibre-gl-js/pull/5835

Should fix or significantly improve: https://github.com/maplibre/maplibre-gl-js/issues/5834

Demo showing a camera frustum in green lines (pitched, looking north) and bounding volumes of all visible tiles for that frustum:

Pitched camera

Looking straight down, but slightly rotated:

Rotated camera

The new bounding volumes are similar to oriented bounding boxes, but the east and west edge planes of the volume are not perpendicular to the other planes and bound the tile perfectly (north, south, top and bottom planes only approximately bound the tile and are perpendicular to each other).

This is best visible for the edge case of pole-neighbouring tiles, which even use only 5 planes and are wedge-shaped. Two rows of tiles are visible in the next screenshot: y=0 tiles with 5 planes, and y=1 tiles with all 6 planes.

Pole tiles

Furthermore the frustum used for culling is shortened in comparison to the "full" frustum from nearZ to farZ - the frustum is cut short to not extend past the horizon plane more than needed.

Benchmark results (roughly 15% slower in exchange for much improved culling). This benchmark only measures `coveringTiles` calls.

image

## Possible future improvement

The convex volume vs frustum test is still an approximation that can sometimes yield false positives. Separating axis theorem -based culling could be implemented to get 100% accuracy, but likely at some more performance cost.

## Launch Checklist

 - [x] Link to related issues.
 - [x] Post benchmark scores.


kubapelc:
I'm still investigating some problems with false negatives near the poles.

kubapelc:
@NathanMOlson Can I include your unit tests from #5835 in this PR? Most of your tests should pass now, and those that do not are failing because they are culling even more tiles that the test assumes.

(the 3 tests that fail now are also culling more tiles than the test assumes)

NathanMOlson:
> @NathanMOlson Can I include your unit tests from #5835 in this PR?

Yes, please do.

kubapelc:
All your tests are actually passing as-is, only the old tests had extra tiles.

src/geo/projection/covering_tiles_details_provider.ts
```diff
@@ -40,5 +40,29 @@ export interface CoveringTilesDetailsProvider {
...
      */
-    recalculateCache(): void;
+    prepareNextFrame(): void;
+}
+
+/**
+ * A generic variant of the {@link CoveringTilesDetailsProvider} indended for use by actual implementations.
+ * Provides additional type safety to functions working with the given type of bounding volume.
+ */
+export interface CoveringTilesDetailsProviderImplementation<T extends IBoundingVolume> extends CoveringTilesDetailsProvider {
```
HarelM, line 50: Why is another interface needed?

src/util/primitives/convex_volume.ts
HarelM, line 1: Does this class have tests covering its logic?

HarelM:
I've added mostly minor comments. 
Does this affect only globe or does it show down mercator as well? 
I haven't reviewed the math itself too deeply, @NathanMOlson feel free to have a look.

src/geo/projection/covering_tiles_details_provider.ts
```diff
@@ -40,5 +40,29 @@ export interface CoveringTilesDetailsProvider {
     /**
      * Prepare cache for the next frame.
      */
-    recalculateCache(): void;
+    prepareNextFrame(): void;
```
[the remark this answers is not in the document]
kubapelc: I felt that the old name was misleading, "recalculate" implies a heavy operation, which it is not, and it is not even calculating anything. The new name just signals when the function should be called. Maybe something like "swapBuffers" is another alternative?

src/util/primitives/convex_volume.ts
kubapelc: Yes, they are in `primitives.test.ts`. Before globe there was a `primitives.ts` file which contained both the Frustum and Aabb class. I think it makes sense to also break up `primitives.test.ts` into multiple test files?

kubapelc:
This should not affect mercator, but I would like to create & run a benchmark for this to make sure. Please do not merge before I post the results.

src/util/primitives/convex_volume.ts
kubapelc: I've broken the test file up.

kubapelc:
I've found no performance difference between this PR and main, both score close to 10 ms on the new mercator covering tiles benchmark.

src/geo/projection/covering_tiles_details_provider.ts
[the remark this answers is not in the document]
HarelM: swap buffers sounds better.

src/util/primitives/convex_volume.ts
HarelM: Yes, single class or file is better... Both for tests and also regular code.

src/geo/projection/covering_tiles_details_provider.ts
kubapelc: I've removed the extra interface. If typescript ever starts enforcing this, it should be no problem to add it back in.

src/geo/projection/covering_tiles_details_provider.ts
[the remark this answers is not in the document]
HarelM: Was this forgotten by mistake?

src/geo/projection/covering_tiles.ts
```diff
@@ -79,13 +79,13 @@ export type CalculateTileZoomFunction = (requestedCenterZoom: number,
...
-export function isTileVisible(frustum: Frustum, aabb: Aabb, plane?: vec4): IntersectionResult {
+export function isTileVisible(frustum: Frustum, tileBoundingVolume: IBoundingVolume, plane?: vec4): IntersectionResult {
 
-    const frustumTest = aabb.intersectsFrustum(frustum);
+    const frustumTest = tileBoundingVolume.intersectsFrustum(frustum);
     if (!plane) {
         return frustumTest;
     }
-    const planeTest = aabb.intersectsPlane(plane);
+    const planeTest = tileBoundingVolume.intersectsPlane(plane);
```
NathanMOlson, line 88: If `frustumTest` is `None`, we can skip this test and return early. This might improve performance a bit. You could also try doing `intersectsPlane()` first, and early return if `None`. I'm not sure which order would be more performant.

src/util/primitives/frustum.ts
```diff
@@ -26,6 +27,41 @@ export class Frustum {
             return vec4.mul(v as any, v as any, [s, s, 1.0 / v[3], s] as vec4);
         });
 
+        if (plane) {
```
NathanMOlson, line 30: This far Z adjustment looks good to me.

src/util/util.ts
```diff
@@ -85,6 +85,42 @@ export function pointPlaneSignedDistance(
     return plane[0] * point[0] + plane[1] * point[1] + plane[2] * point[2] + plane[3];
 }
 
+/**
+ * Finds an intersection points of three planes. Returns null if no such point exists.
+ * The planes *must* be in Hessian normal form - their xyz components must form a unit vector.
+ */
+export function threePlaneIntersection(plane0: vec4, plane1: vec4, plane2: vec4): vec3 {
```
NathanMOlson, line 92: Looks like this could use a unit test.

src/util/util.ts
```diff
@@ -85,6 +85,42 @@ export function pointPlaneSignedDistance(
...
+
+/**
+ * Returns a parameter `t` such that the point obtained by
+ * `origin + direction * t` lies on the given plane.
+ * Direction does not need to be normalized.
+ */
+export function rayPlaneIntersection(origin: vec3, direction: vec3, plane: vec4): number {
+    const dotOriginPlane = origin[0] * plane[0] + origin[1] * plane[1] + origin[2] * plane[2];
+    const dotDirectionPlane = direction[0] * plane[0] + direction[1] * plane[1] + direction[2] * plane[2];
+    return (-dotOriginPlane -plane[3]) / dotDirectionPlane;
```
NathanMOlson, line 121: possible divide by zero

src/util/primitives/convex_volume.ts
```diff
@@ -0,0 +1,175 @@
...
+                if (plane[0] * point[0] + plane[1] * point[1] + plane[2] * point[2] + plane[3] >= 0) {
+                    boxPointsPassed++;
+                }
+            }
+
+            if (boxPointsPassed === 0) {
+                return IntersectionResult.None;
+            }
+            if (boxPointsPassed < boxPointCount) {
+                fullyInside = false;
```
NathanMOlson, line 128: Could `break;` here to improve performance?

NathanMOlson (commented):
Thanks, @kubapelc!

src/geo/projection/covering_tiles_details_provider.ts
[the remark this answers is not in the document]
kubapelc: Yes, I've been looking at the function of the same name in `bounding_volume_cache.ts`, I didn't realize it was also present in this interface.

I would prefer to keep `prepareNextFrame` here, because it is inside an interface. The name signals how the function should be used, whereas `swapBuffers` describes a specific implementation (and one that is only used for globe).

src/geo/projection/covering_tiles.ts
kubapelc: I've added the break. I've measured no significant difference between the two orders.

src/util/util.ts
kubapelc: I've added some tests. It seems that this way of computing plane intersections is not very numerically stable, if the plane directions are similar, the computed point can be very inaccurate. It should not be an issue though, since the function is used for tile border planes which will be close to perpendicular to each other.

src/util/util.ts
kubapelc: Good point. Divide by zero here means the ray is parallel to the plane. It can also return negative values if the ray is pointing away from the plane. The frustum code did not account for these cases, I've also fixed that now.

kubapelc:
I've modified the frustum creation logic: I've added another way of computing the optimal far plane position so that it touches the horizon exactly. It is mostly used as a fallback if the old ray-based computation fails due to some frustum corner rays pointing away from the horizon.

This lead to two covering tiles tests returning a few more tiles, but I believe the new behaviour is correct. I've checked by vizualizing the frustum and returned tiles for those two tests, and both look reasonable (apart from the tiles having too low zooms, but didn't modify that logic, and these seem like edge cases).

image

image

I've also noticed that globe's matrix actually has a different coordinate system from mercator, which leads to some ugliness in the frustum code. Mercator has Y down (towards south), globe has Y up (towards north pole). Fixing this would probably be a lot of changes, so I'd prefer to do it later in a separate PR.

src/geo/projection/globe_covering_tiles_details_provider.ts
```diff
@@ -101,90 +104,208 @@ export class GlobeCoveringTilesDetailsProvider implements CoveringTilesDetailsPr
...
+
+            // Axes for the east and west edge of our bounding volume.
+            // These axes are NOT opposites of each other, they differ!
+            // They are also not orthogonal to the up/down and north/south axes.
+            const axisEast = vec3.cross([] as any, corners[2], corners[1]);
+            vec3.normalize(axisEast, axisEast);
+            const axisWest = vec3.cross([] as any, corners[0], corners[3]);
+            vec3.normalize(axisWest, axisWest);
+
+            // Now we will expand the extremes point set for bounding volume creation.
```
HarelM, line 206: This should be in a separate method. It's too long.
Consider splitting it even further.

src/util/primitives/frustum.ts
```diff
@@ -35,6 +42,86 @@ export class Frustum {
             [0, 4, 5]   // top
         ];
 
+        const scale = Math.pow(2, zoom);
+
+        // Transform frustum corner points from clip space to tile space, Z to meters
+        const frustumCoords = clipSpaceCorners.map(v => unprojectClipSpacePoint(v, invProj, worldSize, scale));
+
+        if (horizonPlane) {
```
HarelM, line 50: Please split the code into more methods and add these great comments as part of the method declaration docs.

src/util/primitives/frustum.ts
```diff
@@ -35,6 +42,86 @@ export class Frustum {
...
+            // For each of the 4 edges from near to far plane,
+            // we find at which distance these edges intersect the given clipping plane,
+            // select the maximal value from these distances and then we move
+            // the frustum's far plane so that it is at most as far away from the near plane
+            // as this maximal distance.
+
+            let maxDist = 0;
+            const cornerRayLengths: number[] = [];
+            const cornerRayNormalizedDirections: vec3[] = [];
+            for (let i = 0; i < 4; i++) {
```
HarelM, line 61: 4 Here is appearing a lot of times, it is possible to use some constant for it maybe?

src/geo/projection/globe_covering_tiles_details_provider.ts
kubapelc: I've split the function up.

src/util/primitives/frustum.ts
kubapelc: This is now also a separate function.

src/util/primitives/frustum.ts
kubapelc: It makes sense for some 4s to remain hardcoded ("4 sides of a square"), I've converted the rest to a variable ("offset of near plane in the array").

HarelM (approved):
LGTM, @NathanMOlson let me know if you think I can merge this.

NathanMOlson (approved):

HarelM:
@kubapelc can you please resolve the conflict and I'll merge this PR. THANKS!

Issue #5839, opened 2025-05-04
When globe + terrain are enabled, dragging toward the arctic circle  or antarctica zooms in


acalcutt:
When globe + terrain are enabled and you are at a low zoom like 2, grabbing and dragging toawrd the arctic circle or antartica zooms in. it does not do this when only globe is enabled.

![Image](https://github.com/user-attachments/assets/2d9a4b21-218b-4f81-913d-bb7722dfcac7)

**maplibre-gl-js version**: 5.4.0

**browser**: chrome

### Steps to Trigger Behavior / Link to Demonstration

 1. Open https://stackblitz.com/edit/web-platform-21c7kfvt?file=index.html
 2. Grab the map toward the arctic circle and drag downward
 3. while dragging you should see the map zooms in. If you click the terrain button first, it does not do that.

### Expected Behavior

The globe should rotate without zooming it when terrain is enabled, like it does when it is turned off

### Actual Behavior

Map zooms towards the poles when terrain is enabled.

HarelM:
Thanks for taking the time to report this issue.
This was the initial behavior of dragging when globe was first released, and this was solved for non-terrain, so there should be an easy fix somewhere in the code I believe...

Caerbannog:
It also crashes if you go too far.

https://github.com/user-attachments/assets/3d6991b7-9f4e-42f1-ba37-18fdbe21535d

PR #5868, opened 2025-05-08
Zoom Adjustment for Drag Panning towards Poles in Globe Terrain (#5839)

kang-matthew:
## Launch Checklist

**Fixes:** #5839.

**Bug:** When globe + terrain are enabled and panning toward the arctic circle or Antarctica enlarges the globe. Happens only on terrain, and it does not do this when only globe is enabled. (Related: [Zoom Behavior](https://github.com/maplibre/maplibre-gl-js/blob/main/developer-guides/globe.md#zoom-behavior))

**Fix:** The fix is to change one of the handler functions to use `handleMapControlsPan` used in non-terrain, which adds the appropriate zoom adjustment with respect to the deltas. Before it uses `setCenter` directly without factoring zoom adjustment.

**Note:** This does not fix the issue for Key Pan (by up or down arrow keys). The globe is enlarged without zoom adjustment for both terrain and non-terrain when panning towards the poles using arrow keys.

**After Fix Visuals:**

https://github.com/user-attachments/assets/9e6be9c7-9c64-4a0f-804e-efe33dfecf44

 - [X] Link to related issues.
 - [ ] Post benchmark scores.


acalcutt:
Does this still work when you are at a high zoom where you can see terrain? Is seems like the code removed was mapping a point on the screen which I am guessing is supposed to account for the terrain when grabbing.

HarelM:
I'm surprised no test failed. 
Yes, this code should help terrain panning, you can't simply remove it.
You might want to add the current projection state to the if.
But also to where terrain movement is initialized.

kang-matthew:
@acalcutt @HarelM Thank you to both of you! It's my first attempt at an issue and I wasn't familiar with the panning implementations. I'm not sure if I perfectly understand it, but here's what I thought:

My understanding is that terrain mode should drag the map by the pan delta, instead of moving the picked point itself. And in the panning implementations (`handleMapControlSpan`) in different projections - Vertical Perspective and Mercator - vertical perspective (globe projection) should drag the map by the pan delta, and the mercator projection should move the picked point.

Hence for the case when we have mercator projection *and* terrain mode, we should not call `handleMapControlSpan`, as it would move the point around instead of panning by delta. And that's why we should add the current projection state to the if?

Thanks again!

HarelM:
Terrain should move the picked point as in heavy terrain the picked point could be a mountain and you want to move it and not by "simple" delta as the case for no terrain (mercator).
Globe and terrain were both added and I'm not sure the solution to panning both is well defined.
But the idea is that pen delta is not a good fit for terrain due to the above use case.

acalcutt:
Just curious looking at the changes and because I don't know. When in globe mode, does it switch to mercator projection at some zoom level? probably not right?

Seems to me like maybe zoom level could play part in when to change the dragging action, but I think (harel?) may have mentioned before in another PR something like that is a bit hacky.

HarelM:
The definition of globe is vertical persoective until zoom 11 then gradually switch to mercator when in zoom 12 it's fully mercator.

shreeharshshinde:
Hi @HarelM, quick question — in the **terrain+drag block**, we use `tr.setCenter(...)` instead of `cameraHelper.handleMapControlsPan(...) `like in other cases. Was there a specific reason for avoiding the helper there? Just wondering if using the helper uniformly could simplify things or if setCenter is needed for terrain accuracy. Thanks!

HarelM:
Yes, I believe this is intentional as I mentioned earlier where you can grab a mountain when terrain is on, which is not the case for non-terrain map.

clement-igonet:
I tried to check whether this fix is still needed on current main (v6.4.1 at the time of measurement).

Dragging over the pole with terrain on and terrain off gives identical zoom trajectories, with a maximum difference of 0.143 between latitude 45 and 85: https://clement-igonet.github.io/maplibre-gl-js/demo-1551/test-5868.html

As far as I can tell, the setCenter fallback guarded here is no longer reachable with the globe projection.

It looks like #8067 and #8154 addressed the underlying problem, so this one may be closable.

5868-check-main

clement-igonet:
Re-checked on the released v6.6.0.

Same journey ends at the same zoom with terrain on and off (max difference 0.103 during the drag).

https://clement-igonet.github.io/maplibre-gl-js/demo-1551/test-5868-v660.html

HarelM:
This PR is a bit stale.
There were a lot of changes recently in globe panning and terrain which might fix this issue.
I'll close this one to reduce clutter in the PR queue.
If this is still relevant feel free to open reopen this.
Also the issue that is linked and should be fixed was fixed by another PR.

PR #5875, opened 2025-05-09
Fix vector tile reload

NathanMOlson:
Fixes #5806 to get `refreshTiles()` to work for an entire vector layer (before this change, `refreshTiles()` worked on vector layers only when provided a list of tiles to refresh).

Before, `SourceCache::_reloadTile()` was called with `state = 'reloading'`. This caused `VectorTileSource::reloadTile()` to be called, instead of the desired `VectorTileSource::loadTile()`. According to the docs, `WorkerSource::reloadTile()` "re-parses a tile that has already been loaded", which is why `loadTile()` must be called instead.

After this change, `SourceCache::_reloadTile()` is called with `state = 'expired'`. This results in the tile being re-requested, just as it already was when a list of tile IDs was provided to `refreshTiles()`.

 - [X] Link to related issues.

### DIFF
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -2230,8 +2230,7 @@ export class Map extends Camera {
             throw new Error(`There is no source cache with ID "${sourceId}", cannot refresh tile`);
         }
         if (tileIds === undefined) {
-            const sourceDataChanged = true;
-            sourceCache.reload(sourceDataChanged);
+            sourceCache.reload(true);
         } else {
             sourceCache.refreshTiles(tileIds.map((tileId) => {return new CanonicalTileID(tileId.z, tileId.x, tileId.y);}));
         }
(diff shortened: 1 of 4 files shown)


src/ui/map.ts
```diff
@@ -2230,7 +2230,8 @@ export class Map extends Camera {
             throw new Error(`There is no source cache with ID "${sourceId}", cannot refresh tile`);
         }
         if (tileIds === undefined) {
-            sourceCache.reload();
+            const sourceDataChanged = true;
```
HarelM, line 2233: Why not simply pass true? Is there a reason to create this variable here?

NathanMOlson: Avoidance of magic number. Shall I change it to simply pass true?

HarelM: Yes please.

HarelM (approved):
THANKS!

PR #5913, opened 2025-05-20
Hypsometric Tint from terrain-RGB tiles

### LINKED ISSUE #5666: Hypsometric Tint from terrain-RGB tiles

NathanMOlson:
Resolves #5666 (Hypsometric Tint from terrain-RGB tiles).

Demo here: https://nathanmolson.github.io/color_relief

The shader uses a binary search to find the adjacent color stops from the color ramp, which means that larger color palettes take longer to render (O(log(n))).

While #5742 passes the color ramp to the shader using uniforms, this PR passes the color ramp to the shader using textures. This ensures that at least 64 color stops will be supported on all hardware. You can see the difference between these two approaches [here](https://github.com/NathanMOlson/maplibre-gl-js/pull/5).

The elevation stops are packed into an RGBA value using the raster DEM encoding scheme. This means that the minimum distance between stops is set by the quantization of the DEM data (0.1m for `mapbox` encoding, 1/256 m for `terrarium`).

Examples: 

cr_linz
cr_geo
cr_iso
cr_rainbow
cr_cat

 - [X] Link to related issues.
 - [x] Post benchmark scores.


larsmaxfield:
Wow. Nice work.

Do you have an example available that shows the performance when modifying the ramp? I have users who wish to set the high and low breakpoints when they inspect zoomed-in areas.

NathanMOlson:
> Do you have an example available that shows the performance when modifying the ramp? I have users who wish to set the high and low breakpoints when they inspect zoomed-in areas.

@larsmaxfield I haven't added any support for dynamically modifying the ramp. Can you tell me more about how you would expect to use that? Would you be using `map.setStyle()` to modify the ramp? or some other mechanism?

larsmaxfield:
I would expect to modify the ramp similar to how I can change existing layer paint properties. For example, to tighten the range of a monochrome colorramp via `map.setPaintProperty()`:

```js
const monochrome = [
  "interpolate",
  ["linear"],
  ["elevation"],
  0.0, 'black',
  0.1, 'black',
  3000, 'white',
  3001, 'white'
];

...

monochrome[5] = 2000;  // Tighten range from 0.1–3000 to 2000–2100
monochrome[7] = 2100;
map.setPaintProperty('hyspometric-layer', 'color-relief-color': monochrome);
```

Or is this already possible?

larsmaxfield:
User-wise the essential idea is:

image

My users specifically are inspecting painting heightmaps. They want to see height difference in local areas where the change is very small compared to the full data range.

For mapping, I can imagine a dynamic colorramp could help give an impression of flood level.

NathanMOlson:
@larsmaxfield I've added the capability to modify the colorramp dynamically using `map.setPaintProperty()`. In my testing, it happens "instantaneously". Thanks for this suggestion, I think it has really made this feature better.

My [demo](https://nathanmolson.github.io/color_relief) now uses `map.setPaintProperty()` to change between colorramps (previously it was removing the layer and adding a new one). I think this PR will now 100% support your use case. Check it out and let me know what you think.

larsmaxfield:
Excellent. And indeed very fast. Thank you!

src/style/style_layer/color_relief_style_layer.ts
```diff
@@ -0,0 +1,114 @@
...
+     * so that the remapping is only performed once.
+     *
+     * @param maxLength - the maximum number of stops in the color ramp
+     *
+     * @return a `ColorRamp` object with no more than `maxLength` stops.
+     *
+     */
+    getColorRamp(maxLength: number) : ColorRamp {
+        this.colorRamp = this._createColorRamp();
+        if (this.colorRamp.elevationStops.length > maxLength) {
```
HarelM, line 71: Why not encapsulate this logic into the _createColorRamp method?

src/style/style_layer/color_relief_style_layer.ts
```diff
@@ -0,0 +1,114 @@
...
+     * Get the color ramp, enforcing a maximum length for the vectors. This modifies the internal color ramp,
+     * so that the remapping is only performed once.
+     *
+     * @param maxLength - the maximum number of stops in the color ramp
+     *
+     * @return a `ColorRamp` object with no more than `maxLength` stops.
+     *
+     */
+    getColorRamp(maxLength: number) : ColorRamp {
+        this.colorRamp = this._createColorRamp();
```
HarelM, line 70: Is `colorRamp` variable really needed? Feels like it's not cached and gets calculated every time, and I'm not sure it's mandatory for hasOffscreen pass. Can it be removed?

src/style/style_layer/color_relief_style_layer.ts
```diff
@@ -0,0 +1,114 @@
...
+                colorRamp.colorStops.push(this.colorRamp.colorStops[Math.round(i)]);
+            }
+            warnOnce(`Too many colors in specification of ${this.id} color-relief layer, may not render properly.`);
+            this.colorRamp = colorRamp;
+        }
+        return this.colorRamp;
+    }
+    
+    getColorRampTextures(context: Context, maxLength: number, unpackVector: number[]): ColorRampTextures {
+        if (!this.colorRampTextures || this._colorRampChanged()) {
```
HarelM, line 86: Nit-picking: Can this if be inverted to reduce indentation?

src/style/style_layer/color_relief_style_layer.ts
```diff
@@ -0,0 +1,114 @@
...
+            const elevationImage = new RGBAImage({width: colorRamp.colorStops.length, height: 1});
+            for (let i = 0; i < colorRamp.elevationStops.length; i++) {
+                const elevationPacked = packDEMData(colorRamp.elevationStops[i], unpackVector);
+                elevationImage.data[4*i + 0] = elevationPacked.r;
+                elevationImage.data[4*i + 1] = elevationPacked.g;
+                elevationImage.data[4*i + 2] = elevationPacked.b;
+                elevationImage.data[4*i + 3] = 255;
+
+                const pxColor = colorRamp.colorStops[i];
+                colorImage.data[4*i + 0] = Math.round(pxColor.r * 255 / pxColor.a);
```
HarelM, line 98: This is a bit specific to how DEM data works, isn't it? Can we move this into the DEMData class somehow?

Same with the above `4*i +...` logic...?

src/style/style_layer/color_relief_style_layer.ts
NathanMOlson: Done.

HarelM:
Added a few comments:
@mwilsnd feel free to review as this is your suggestion about using texture I believe.

src/shaders/color_relief.fragment.glsl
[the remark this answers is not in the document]
HarelM: Oh, right, I missed that, thanks!

src/style/style_layer/color_relief_style_layer.ts
NathanMOlson: Done.

mwilsnd:
Looking good, thanks for taking the time to make these changes! I'm hopeful our days having to consider WebGL 1 are numbered, I encourage you to join the discussion about this at our next Web meeting.

I'll defer final approval to @HarelM.

src/style/style_layer/color_relief_style_layer.ts
NathanMOlson: This is not specific to DEM data. It is just filling in the image `Uint8Array` with the specified colors. The only part that is DEM specific is `packDEMData()`, which lives in `dem_data.ts` and is well-encapsulated in my opinion.

Note that this is similar to what is done in `renderColorRamp()`, but that function cannot be used here because it requires uniform spacing of the interpolant. Here we allow elevation stop to be nonuniformly spaces to support sharp color breaks.

NathanMOlson: I've added `RGBAImage::setPixel()` to clarify and de-duplicate.

src/util/color_ramp.test.ts
HarelM, line 1: Why did tests here change?

HarelM (approved):

Issue #5929, opened 2025-05-22
API docs for `clusterMaxZoom`  need to specify it needs to be an integer


igal1c0de4n:
**maplibre-gl-js version**:
v5.4.0 (+earlier versions)
**browser**:
Chrome (but seems irrelevant)

### Steps to Trigger Behavior

 1. Create a cluster layer as in example
 https://maplibre.org/maplibre-gl-js/docs/examples/cluster/
 2. Specify a non-integer number for `clusterMaxZoom` e.g. 14.1

### Expected Behavior
- There should be validation  on the `clusterMaxZoom` field: maybe the error should be more descriptive.
- API docs should specify an integer number must be provided to `clusterMaxZoom`

### Actual Behavior
- The clusters disappear and maplibre errors "Invalid array length". 
It makes it difficult to understand the root of the issue.

I didn't bother with providing a codepen because issue is obvious enough. lmk otherwise.

For reference, my chatgpt sesssion: https://chatgpt.com/share/682fa3e8-99d8-800c-817e-7d56b6c28ad3

HarelM:
Thanks for taking the time to open this issue! 
Feel free to submit a PR, the API docs are here I believe (unless you read it somewhere else):
https://github.com/maplibre/maplibre-gl-js/blob/bd885b5ed5fa19b9fbd6f6ac4ecd12a095868395/src/source/geojson_source.ts#L47

Regarding validation, I'm not sure if it should be here or in supercluster code, but feel free to add it as well.

igal1c0de4n:
Ok, see associated PR 🙏🏼  
I've tested it locally, seems to work as expected

shreeharshshinde:
While working with the clusterMaxZoom option, I noticed that if someone passes a non-integer (like a float), the code currently throws an error because Supercluster expects an integer value.

I’d like to clarify the expected behavior here:

Should we:

1. **Strictly enforce an integer** and throw a clear error if a float or non-integer is passed?
2. **Gracefully handle it by rounding** (e.g., using `Math.floor`) and apply the adjusted integer value with a `console.warn()` to inform the user?

Let me know what approach you’d prefer — happy to implement either based on your guidance.

shreeharshshinde:
I checked the PR by @igal1c0de4n and noticed that a separate private function was defined to round non-integer values, which seems like a logically sound approach. It ensures that invalid maxZoom values (like floats) are handled gracefully without immediately throwing an error.

[...]

PR #5933, opened 2025-05-24
Fix 5929: error handling and docs for 'clusterMaxZoom' integer requirement 

igal1c0de4n:
Addresses issue: https://github.com/maplibre/maplibre-gl-js/issues/5929.

This change:
1. Adds a clarification for the API about `clusterMaxZoom` needing to be an integer.
2. When a non-integer value is provided for `clusterMaxZoom`:
  -  The value is rounded - preventing clusters from breaking.
  -  A warning is shown.


src/source/geojson_source.ts
```diff
@@ -178,7 +179,8 @@ export class GeoJSONSource extends Evented implements Source {
                 generateId: options.generateId || false
             },
             superclusterOptions: {
-                maxZoom: options.clusterMaxZoom !== undefined ? options.clusterMaxZoom : this.maxzoom - 1,
+                maxZoom: options.clusterMaxZoom !== undefined ?
```
HarelM, line 182: Why not incorporate the entire logic in this method (taking care of undefined and setting it to max zoom -1 incase it's not defined?

src/source/geojson_source.ts
```diff
@@ -44,7 +44,8 @@ export type SetClusterOptions = {
      */
     cluster?: boolean;
     /**
-     * The cluster's max zoom
+     * The cluster's max zoom.
+     * Must be an integer due to internal implementation.
```
HarelM, line 48: I would consider changing this to be more explicit about internal implementation.
Since it is rounded, then it's not really mandatory, isn't it?

HarelM:
Any chance you can add a test to cover the relevant issue/solution?
Otherwise, I've added a few nit-picking comments.

src/source/geojson_source.ts
```diff
@@ -199,6 +201,14 @@ export class GeoJSONSource extends Evented implements Source {
         return pixelValue * (EXTENT / this.tileSize);
     }
 
+    private _getClusterMaxZoom(clusterMaxZoom: number): number {
+        const effectiveClusterMaxZoom = Math.round(clusterMaxZoom);
+        if (effectiveClusterMaxZoom !== clusterMaxZoom) {
+            warnOnce(`clusterMaxZoom is expected to be an integer, "${clusterMaxZoom}" provided.`);
```
HarelM, line 207: Add which value is used compared to the one provided for better DX.

igal1c0de4n: you mean something like:
```js
warnOnce(`provided value for "clusterMaxZoom" is ${clusterMaxZoom}", converted to ${effectiveClusterMaxZoom} (integer expected)`);
```
If not, pls provide example

src/source/geojson_source.ts
igal1c0de4n: do you mean:
```js
    private _getClusterMaxZoom(clusterMaxZoom: number): number {
        const effectiveClusterMaxZoom = clusterMaxZoom ?
           Math.round(clusterMaxZoom) : this.maxzoom - 1;
        if (effectiveClusterMaxZoom !== clusterMaxZoom) {
            warnOnce(`clusterMaxZoom is expected to be an integer, "${clusterMaxZoom}" provided.`);
        }
        return effectiveClusterMaxZoom;
    }
```
then, here:
```js
 maxZoom: this._getClusterMaxZoom(options.clusterMaxZoom)

File: src/source/geojson_source.ts (hunk 2 above)
Comment (igal1c0de4n) [reply src/source/geojson_source.ts:48]: Do you mean to expand about the histogram-like internal mechanism which "buckets" the clusters into discrete (integer) zoom levels? Might be TMI, but also I didn't bother to dive too deep into the way it actually works, so not sure I'm the right guy to provide the more explicit reasoning 😅  

The change indeed allows `clusterMaxZoom` to be non-integer. 
When that's the case, cluster no longer breaks. 
That said, only integer values are expected - asserting non-integers are supported without an api docs warning may be misleading to the developer (since they are rounded). 

I made it so non-integer value "work" out of consideration to not break maps for users who are sending incompatible values (hence the devtools warning). 
Subjectively, I'd advise against considering non-integers as a valid/legal input.

Anyway - If you believe a different comment should go here, pls provide it and I'll update the PR

File: src/source/geojson_source.ts (hunk 1 above)
Comment (HarelM) [reply src/source/geojson_source.ts:182]: Yup.

File: src/source/geojson_source.ts (hunk 3 above)
Comment (HarelM) [reply src/source/geojson_source.ts:207]: Yes please :-)

File: src/source/geojson_source.ts (hunk 1 above)
Comment (igal1c0de4n) [reply src/source/geojson_source.ts]: Addressed. 
Note I modified the warning display condition to prevent tigger when `clusterMaxZoom === undefined`

File: src/source/geojson_source.ts (hunk 2 above)
Comment (igal1c0de4n) [reply src/source/geojson_source.ts]: Updated. 
@HarelM Hopefully you won't mind I've "optimized" the explanation (it's on the shorter side now)

File: yarn.lock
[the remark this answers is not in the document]
Comment (igal1c0de4n) [reply yarn.lock:1]: Funny thing is I made a conscious effort to not include it yet somehow it slipped into the PR anyway 🤦🏼

**Comment by HarelM** [comment]
Visual tests are a bit of an overkill for this change I believe, a unit test would probably be enough, check out other tests here: 
https://github.com/maplibre/maplibre-gl-js/blob/2abf7930d0d44e51c72da1a4e30ed842df16407d/src/source/geojson_source.test.ts

**Comment by igal1c0de4n** [comment]
@HarelM I've added a test to `geojson_source.ts` it passes. 
Note some tests are failing locally. It seems unrelated to this change. 
Could it be a misconfiguration in my local environment?
```
 1/20: failed tests/text-local-ideographs/cjk-symbols 0.00029635416666666666
 2/20: failed tests/terrain/pitched-world 0.000949859619140625
 3/20: failed tests/terrain/fog-sky-blend-globe 0.001739501953125
 4/20: failed tests/sky/roll45-with-text 0.000274658203125
 5/20: failed tests/regressions/mapbox-gl-js#5631 0.00048828125
 6/20: failed tests/projection/globe/terrain/raster-planet 0.0002899169921875
 7/20: failed tests/projection/globe/terrain/fill-planet-tiles 0.000385284423828125
 8/20: failed tests/projection/globe/terrain/collision-text-variable-anchor/rotated 0.0002593994140625
 9/20: failed tests/projection/globe/terrain/collision-text-point-pole-to-pole 0.0004119873046875
 10/20: failed tests/projection/globe/terrain/collision-text-pitched-rotated 0.000396728515625
 11/20: failed tests/projection/globe/terrain/atmosphere/base 0.000301361083984375
 12/20: failed tests/projection/globe/icon-text-translate-viewport 0.000335693359375
 13/20: failed tests/projection/globe/collision-text-point-pole-to-pole 0.0002593994140625
 14/20: failed tests/projection/globe/collision-text-pitched-rotated 0.00042724609375
 15/20: failed tests/projection/globe/collision-icon-text-translate-map 0.00030517578125
 16/20: failed tests/icon-text-fit/textFit-grid-long 0.0008487654320987654
 17/20: failed tests/high-pitch/pitch95 0.001964569091796875
```

HarelM (approved):
Looks good thanks! Please update me about the coverage and I'll merge this.

igal1c0de4n:
ok, looks like the test is included and exec is successful: 
```shell
$ npm run test-unit-ci

> maplibre-gl@5.5.0 test-unit-ci
> vitest run --config vitest.config.unit.ts --coverage

 RUN  v3.1.3 /Users/igal/work/external-resources/maplibregl/maplibre-gl-js.igal
      Coverage enabled with v8
...
 ✓  unit  src/source/geojson_source.test.ts (31 tests) 31ms
...
 Test Files  176 passed (176)
      Tests  2347 passed (2347)
   Start at  14:24:24
   Duration  33.78s (transform 4.76s, setup 63.29s, collect 50.99s, tests 26.16s, environment 71.06s, prepare 13.92s)

igal1c0de4n:
I'm not 100% I know how to do that. Can you guide me?

I see the`geojson_source.test` file is executed, however I don't see anything which references my specific test `warn if clusterMaxZoom non-integer` (also, I don't see anything in the html which references other specific tests).

Screenshot 2025-05-25 at 2 43 17 PM
Screenshot 2025-05-25 at 2 45 46 PM

Also some numbers next to the new function:

Screenshot 2025-05-25 at 2 44 05 PM

HarelM:
1x means the code is passed one time, which is probably the test you added. 
If this was red or yellow, it would mean it is partially or not covered. 
Thanks!!

Issue #5970, opened 2025-06-03
Could not compile fragment shader: ERROR: 0:27: 'switch' : Illegal use of reserved word


Plantain:
We are seeing errors like below affecting ~4% of users since migrating from 4.7.0 to 5.4.0. I initially thought this was related to the new 5.5.0 hillshade changes, or webgl1, or old browsers, but they seem to have been red herrings as it still occurs in 5.4.0, and we have since observed it on modern browsers (Chrome/Edge 136).

I have very limited information, only coming through Sentry or from non-technical users presently and no live case to debug further.

I still suspect it is hillshade related as I have had some of the traces point to this line:
>m=e.useProgram("hillshade",null,!1,p),f=!e.options.moving;for(const p of r){const r=t {snip}

Error looks like:

Uncaught Error: Could not compile fragment shader: ERROR: 0:27: 'switch' : Illegal use of reserved word
ERROR: 0:27: 'switch' : syntax error
 
    at new aa (maplibregl-BIaNX82f.js:4:702740)
    at Mo.useProgram (maplibregl-BIaNX82f.js:4:781300)
    at To (maplibregl-BIaNX82f.js:4:748619)
    at maplibregl-BIaNX82f.js:4:777856
    at Mo.renderLayer (maplibregl-BIaNX82f.js:4:777880)
    at Mo.render (maplibregl-BIaNX82f.js:4:763256)
    at z.Map._render (maplibregl-BIaNX82f.js:4:913599)
    at maplibregl-BIaNX82f.js:4:915661
    at maplibregl-BIaNX82f.js:4:404297

HarelM:
I don't think 5.4 has a switch statement in shaders but I know 5.5 has.
Are you sure you are not seeing failures related to some caching issue and are not using **your** latest version with maplibre 5.4?

mwilsnd:
Looking at the [specification](https://www.khronos.org/files/opengles_shading_language.pdf) for the version of GLSL used by WebGL 1, it looks like `switch` is not a supported keyword. Sounds like these issues are from old hardware only supporting WebGL 1? We should probably address this in a minor patch release.

If you have a user with this problem, see if they can share what the following page has to say about their device: https://webglreport.com/?v=2.

While we haven't yet officially ended support for WebGL 1, I would anticipate the library ending such support some time in the near future. If these issues are from devices that don't support WebGL 2, I'd be interested to hear more about what kind of hardware your userbase is on.

[...]

PR #2599, opened 2023-05-26
Migrate Shaders to GLSL 3.0

birkskyum:
Closes #2562 

Migrate all shaders to GLSL 3.0

Cleanup after conditional GLSL 3.0 support #2576 

Adds a single macos exception to overlapping-zoom/expected_macos.png because a mac font offset the text a tiny bit.

## Launch Checklist
 - [X] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [X] Briefly describe the changes in this PR.
 - [X] Link to related issues.
 - [X] Write tests for all new functionality.
 - [X] Post benchmark scores.
 - [X] Add an entry to `CHANGELOG.md` under the `## main` section.

**Benchmarks** - In order to do proper profiling here, the variance, p-value etc. would be beneficial. This change should only have positive impact on performance, if any.

Screenshot 2023-05-26 at 14 56 12

PR #5972, opened 2025-06-03
fix: Don't use incompatible GLSL `switch` statement

mwilsnd:
## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

Addresses the use of a WebGL 2.0 reserved keyword as reported in #5970


HarelM:
Is there a way for us to test webgl compatibility somehow while we still support it? Is it worth the effort?

mwilsnd:
I wrote a validator using `glslang` that runs before the render tests, which led me to discover some other issues with WebGL 1.0 non-compliance. This PR seems to be the biggest problem with full compliance: #2599 but it's been in the library since [v3.0.1](https://github.com/maplibre/maplibre-gl-js/releases/tag/v3.0.1). This is confusing as the change from `varying/gl_FragCoord/texture2D` to `in/out/texture` hasn't generated any reports of compile failures. If we can assume the browser is doing some level of patching, it isn't able to handle things like `switch` apparently.

No valid version of GLSL both supports `in/out` qualifiers and not `switch`. If we want to use a validator with the real GLSL compiler frontend, we'll have to revert #2599.

HarelM:
There are some text manipulations to downlevel the shaders code here: 
https://github.com/maplibre/maplibre-gl-js/blob/270b8342ef71b9bc41d0102a248d2fa170400798/src/shaders/shaders.ts#L221

Which is probably the differences your saw...

mwilsnd:
That makes much more sense! Let me revise the validation tool to use this extra logic and we can try adding it to CI.

PR #5973, opened 2025-06-04
Fix #2764: Nix long janks in mouse events.

### LINKED ISSUE #2764: [Performance] Mouse events slow with terrain
If a map has terrain data, mouse events are slow  (~3ms) compared to the typical time of a render call (~8ms). (note these times can vary significantly)

The `mousemove` event is dominated by `readPixels` which synchronously downloads texture data from the GPU, a slow process.

image

Some possible ways to make this faster:
1. Use async data readback [per mdn](https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices#use_non-blocking_async_data_readback)
2. Provide the coordinate as the *ray* under the mouse, rather than the point where that ray intersects the ground (this especially makes sense if the pointer is above the horizon). I especially like this since it can use the same logic whether or not terrain is enabled.
3. Specify the coordinates as non-altitude-corrected values or values corrected by approximate altitude. This is potentially confusing.

alex-kowalczyk:
Addresses https://github.com/maplibre/maplibre-gl-js/issues/2764 

* Implemented async readPixels for WebGL2 (with sync fallback if not supported).
* Implemented async variants of upstream functions: `screenPointToMercatorCoordinateAsync`, `tilesIn` and `queryRenderedFeaturesAsync` .   Extracted common code to private "pre"/"post" functions to avoid duplication.
* Modified mouse events handlers to be async and call `queryRenderedFeaturesAsync`.
* Extended tests to cover async variants.
The async variants do not fully replace original async approach to avoid impacting the public API. They are currently hooked only for mouse events.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 - Fix `refreshTiles()` for vector tiles ([#5875](https://github.com/maplibre/maplibre-gl-js/pull/5875))
 - Revert changes to polygon intersection detection ([#5590](https://github.com/maplibre/maplibre-gl-js/pull/5590) caused issue [5864](https://github.com/maplibre/maplibre-gl-js/issues/5864))
 - Fix breaking clusters when non-integer value provided for `clusterMaxZoom` (warning is shown) ([#5929](https://github.com/maplibre/maplibre-gl-js/issues/5929)) + clarify API docs
+- Improve performance of mouse events with terrain by using WebGL2 async data readback ([2764](https://github.com/maplibre/maplibre-gl-js/issues/2764))
```
HarelM, line 24: Can you push this upwards please?

HarelM:
This seems like a very invasive change to the codebase, is there a different way to solve the issue?

alex-kowalczyk:
> This seems like a very invasive change to the codebase, is there a different way to solve the issue?

A less invasive idea that comes into my mind is to bypass `terrain.pointCoordinate` when `pitch == 0` in `screenPointToMercatorCoordinate`. 
There is already similar fallback there for non-terrain. AFIU, the (faster) fall-back, `screenPointToMercatorCoordinateAtZ`, would be good enough for non-pitched map, even if terrain is enabled, wouldn't it? 
Maybe elevation (Z) could be still obtained for that case (not sure if that is necessary).
This way the performance hit of `WebGL.readPixels` would be only for users who view the tilted 3D map, while still allowing other benefits of terrain, like shading or slopes.

```javascript
    screenPointToMercatorCoordinate(p: Point, terrain?: Terrain): MercatorCoordinate {
        let z = 0;
        // get point-coordinate from terrain coordinates framebuffer
        if (terrain) {
            const pitch = ...;  // can we access it here?
            const coordinate = pitch ? terrain.pointCoordinate(p) : null ;
            if (coordinate != null) {
                return coordinate;
            }
            // Can we determine tileId w/o querying WebGL?
            z = this.getElevation(tileId, x, y)
        }
        return this.screenPointToMercatorCoordinateAtZ(p, z);
    }
```

HarelM:
I think that looking at a 3D map with pitch = 0 is not very interesting, a work around would be to simply enable terrain only when tilting the map (which I think is a good workaround BTW).

alex-kowalczyk:
Yeah, the "disable terrain" workaround is good enough. Closing the PR.

PR #5980, opened 2025-06-05
fix: `textureSize` is not available in WebGL 1.0

mwilsnd:
## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

Replaces the calls to `textureSize` with a uniform vector in the color relief layer. This GLSL method was not added until WebGL 2.0. This is similar to #5972 which was reported in #5970.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@
 - Revert changes to polygon intersection detection ([#5590](https://github.com/maplibre/maplibre-gl-js/pull/5590) caused issue [5864](https://github.com/maplibre/maplibre-gl-js/issues/5864))
 - Fix breaking clusters when non-integer value provided for `clusterMaxZoom` (warning is shown) ([#5929](https://github.com/maplibre/maplibre-gl-js/issues/5929)) + clarify API docs
 - Fix use of reserved GLSL `switch` statement in hillshade shader ([#5972](https://github.com/maplibre/maplibre-gl-js/pull/5972))
+- Fix use of `textureSize` call in color relief shader ([#5980](https://github.com/maplibre/maplibre-gl-js/pull/5980))
 
 
 ## 5.5.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
+- Fix use of `textureSize` call in color relief shader ([#5980](https://github.com/maplibre/maplibre-gl-js/pull/5980))
+
 - _...Add new stuff here..._
 
 ## 5.6.0
@@ -22,7 +24,6 @@
 - Revert changes to polygon intersection detection ([#5590](https://github.com/maplibre/maplibre-gl-js/pull/5590) caused issue [5864](https://github.com/maplibre/maplibre-gl-js/issues/5864))
 - Fix breaking clusters when non-integer value provided for `clusterMaxZoom` (warning is shown) ([#5929](https://github.com/maplibre/maplibre-gl-js/issues/5929)) + clarify API docs
 - Fix use of reserved GLSL `switch` statement in hillshade shader ([#5972](https://github.com/maplibre/maplibre-gl-js/pull/5972))
-- Fix use of `textureSize` call in color relief shader ([#5980](https://github.com/maplibre/maplibre-gl-js/pull/5980))
 
 
 ## 5.5.0
--- a/src/render/draw_color_relief.ts
+++ b/src/render/draw_color_relief.ts
@@ -56,6 +56,7 @@ function renderColorRelief(
     const align = !painter.options.moving;
 
     let firstTile = true;
+    const textureSizes: Array<Number> = [0, 0];
 
     for (const coord of coords) {
         const tile = sourceCache.getTile(coord);
@@ -68,6 +69,8 @@ function renderColorRelief(
             context.activeTexture.set(gl.TEXTURE4);
             colorTexture.bind(gl.LINEAR, gl.CLAMP_TO_EDGE);
             firstTile = false;
+            textureSizes[0] = elevationTexture.size[0];
+            textureSizes[1] = colorTexture.size[0];
         }
 
         if (!dem || !dem.data) {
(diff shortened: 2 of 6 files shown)


mwilsnd:
cc @NathanMOlson While I don't think these changes should cause any issues I'd appreciate any feedback - both the changes here and in #5972 can be reverted once we officially drop WebGL 1.

src/render/uniform_binding.ts
```diff
@@ -41,6 +41,20 @@ class Uniform1i extends Uniform<number> {
     }
 }
 
+class Uniform2i extends Uniform<vec2> {
```
NathanMOlson: A unit test should be added for this new class.

NathanMOlson (approved):

src/render/draw_color_relief.ts
```diff
@@ -56,6 +56,7 @@ function renderColorRelief(
     const align = !painter.options.moving;
 
     let firstTile = true;
+    const textureSizes: Array<Number> = [0, 0];
```
HarelM: [number, number] type instead? or can this hold more than 2 numbers?

src/render/draw_color_relief.ts
```diff
@@ -68,6 +69,8 @@ function renderColorRelief(
             context.activeTexture.set(gl.TEXTURE4);
             colorTexture.bind(gl.LINEAR, gl.CLAMP_TO_EDGE);
             firstTile = false;
+            textureSizes[0] = elevationTexture.size[0];
+            textureSizes[1] = colorTexture.size[0];
```
HarelM: Size only has a single dimension? Is this intentional or a copy-paste mistake?

src/render/uniform_binding.test.ts
```diff
@@ -51,6 +52,24 @@ describe('Uniform Binding', () => {
         u.set(2);
     });
 
+    test('Uniform2i', () => {
+        expect.assertions(4);
```
HarelM: This line is usually not needed and also causes more maintenance for the tests, consider removing it.

src/render/draw_color_relief.ts
mwilsnd: They're both 1D textures, see [here](https://github.com/maplibre/maplibre-gl-js/blob/main/src/style/style_layer/color_relief_style_layer.ts#L86) where they get created. Just sending the width component of each is all that is required and saves uniform space in the shader.

src/render/draw_color_relief.ts
mwilsnd: Only two, I've updated the type.

src/render/uniform_binding.test.ts
mwilsnd: I'm just going off the existing tests for the other uniform wrappers. Would you like me to remove the call assertion check for all of them?

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@
 - Revert changes to polygon intersection detection ([#5590](https://github.com/maplibre/maplibre-gl-js/pull/5590) caused issue [5864](https://github.com/maplibre/maplibre-gl-js/issues/5864))
 - Fix breaking clusters when non-integer value provided for `clusterMaxZoom` (warning is shown) ([#5929](https://github.com/maplibre/maplibre-gl-js/issues/5929)) + clarify API docs
 - Fix use of reserved GLSL `switch` statement in hillshade shader ([#5972](https://github.com/maplibre/maplibre-gl-js/pull/5972))
+- Fix use of `textureSize` call in color relief shader ([#5980](https://github.com/maplibre/maplibre-gl-js/pull/5980))
```
HarelM: This should be under main for next version instead of version 5.6

src/render/uniform_binding.test.ts
HarelM: As far as I can understand this test, there are two assertions happening "async".
So in this case there's a need to create a spy to make sure uniform2i is called twice I guess.
Generally speaking, I prefer the assertions to happen at the end of the test, and if there's a need to make sure something is called then use a spy.
It would be better to fix all this file if it doesn't follow this principle, but it can be a separate PR, your call.

src/shaders/color_relief.fragment.glsl
```diff
@@ -6,6 +6,7 @@ uniform sampler2D u_image;
 uniform vec4 u_unpack;
 uniform sampler2D u_elevation_stops;
 uniform sampler2D u_color_stops;
+uniform ivec2 u_texture_size;
```
HarelM: is "texture_size" a good definition of what is passed to this shader? I find it a bit confusing...?

src/render/program/color_relief_program.ts
```diff
@@ -25,12 +27,14 @@ const colorReliefUniforms = (context: Context, locations: UniformLocations): Col
...
     'u_color_stops': new Uniform1i(context, locations.u_color_stops),
+    'u_texture_size': new Uniform2i(context, locations.u_texture_size),
     'u_opacity': new Uniform1f(context, locations.u_opacity)
 });
 
 const colorReliefUniformValues = (
     layer: ColorReliefStyleLayer,
-    dem: DEMData
+    dem: DEMData,
+    textureSizes?: [number, number]
```
HarelM: Is it less performant to pass two numbers separately here, one for each size of the array?
Or just a single number of the length of the array should be equal?

src/shaders/color_relief.fragment.glsl
mwilsnd: Width technically, the name of the uniform was chosen as a direct replacement for the functionality offered by the built-in `textureSize` method of later GLSL versions.

src/render/program/color_relief_program.ts
mwilsnd: I was mostly focused on mirroring the behavior of `textureSize()`, but you're correct that both textures should be the same width. Just sending one value means we won't need the `Uniform2i` wrapper, either. I'll make some changes.

CHANGELOG.md
mwilsnd: Oops, moved to correct spot.

src/render/uniform_binding.test.ts
mwilsnd: I've simplified this PR to the point I no longer need this additional uniform type, so I've removed it. Let's address the other unit tests in a fresh PR.

src/shaders/color_relief.fragment.glsl
mwilsnd: Renamed to `u_color_ramp_size`

Issue #4088, opened 2024-05-07
Image source is clipped to -180,180 when terrain is enabled.


Hooterr:
**maplibre-gl-js version**: 4.2.0

**browser**: chrome (not important)

### Steps to Trigger Behavior

 1. Add a style with terrain
 2. Add an image source that's outside -180,180 longitude
 4. Enable terrain 

### Link to Demonstration
https://jsbin.com/fiyisil/edit?html,output

```
const urlWms = "https://placehold.co/600x600/png";
const sourceName = "sourceName";
const wmsLayerName = "wms";
const imageCoordinates = [
    [-20, 60],
    [300, 60],
    [300, -60],
    [-20, -60]
];

map.addSource(sourceName, {
        'type': 'image',
        'url': urlWms,
        'coordinates': imageCoordinates
    });

map.addLayer({
    'id': wmsLayerName,
    'type': 'raster',
    'source': sourceName,
    "paint": {
        'raster-opacity': 0.33,
        'raster-fade-duration': 0
    }
});
```

Terrain on
![image](https://github.com/maplibre/maplibre-gl-js/assets/28758721/be4b8342-364b-43a7-8ff0-af287512c4ca)

Terrain off
![image](https://github.com/maplibre/maplibre-gl-js/assets/28758721/afa2b3f8-5673-49e6-87da-bc9158c26390)

### Expected Behavior
Image wraps correctly when terrain is enabled.

### Actual Behavior
Image seems to be clipped to -180,180 longitude
Also the opacity seems to be changing when terrain is enabled/disabled

HarelM:
Is it possible that this is a duplicate of:
- #1559

Hooterr:
Similar but different. I don't see problems with latitude. In the mentioned issue the problem seems to be with any combination of lat/lng, here the problem seems to be related to some sort of normalizing/clipping to a fixed range of -180,180.
Could be the same root cause though.

Panthro22:
@Hooterr 
When the terrain is enabled, it is clipped to a specific tile. The image corner coordinates are then converted to that local tile's coordinate system so it will not go past the Tile edge as it is being cut off to fit that tile, even if it's near the border. If you zoom in to a street-level view, would the image disappear partially or entirely for you? I am asking as there is an issue of zooming in for an image's set near tile z-level of 1.

Hooterr:
The image stays partially visible no matter what the zoom level is.

nbald:
Same here

No terrain:
[...]

PR #5989, opened 2025-06-06
Fix image sources being clipped at -180 and 180 longitude when terrain is enabled

pstaszek:
Fixed issue #4088.

### TL;DR
Lack for wrapping support for terrain render to texture tiles.

### Lond description.

Terrain tiles are rendered to the texture, which then is rendered to the tile (images as well) so content can't overlap between tiles. There was no support for wrapping, so for every wrap image was rendered on the same position relative to the tile.

See the diagram:
wrapping_issue_illustration

The image on tile with wrap = 0 is rendered at desired position, however on tile with wrap = 1 it is rendered on the same position in relation to a tile as for previous one. It should be actually rendered with shift.

### Before:
before

### After:
after

I also added render test.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


src/source/terrain_source_cache.ts
```diff
@@ -294,10 +297,14 @@ export class TerrainSourceCache extends Evented {
...
         canonicalTileRanges: {[zoom: string]: CanonicalTileRange}
     ): boolean {
-        return canonicalTileRanges[tileID.canonical.z] &&
-            tileID.canonical.x >= canonicalTileRanges[tileID.canonical.z].minTileX &&
-            tileID.canonical.x <= canonicalTileRanges[tileID.canonical.z].maxTileX &&
-            tileID.canonical.y >= canonicalTileRanges[tileID.canonical.z].minTileY &&
-            tileID.canonical.y <= canonicalTileRanges[tileID.canonical.z].maxTileY;
+        const range = canonicalTileRanges[tileID.canonical.z];
+
+        return range && (
```
HarelM: This will return `null` in case range is `null` instead of `false`.

HarelM:
Can you attach a screen shot of how the render test would look without the fix to make sure the render test covers this scenario?

HarelM:
Is the code coverage report correct by highlighting that the changed code is not covered in tests?
image

pstaszek:
Render test screenshots
On `main` branch:
actual

On current branch (correct)
expected

src/source/terrain_source_cache.ts
pstaszek: Nice catch! Fixed.

pstaszek:
For the coverage report, this is a bit more complex.

I assume the coverage report is based on unit tests. Unit tests check only logic responsible for adding and omitting tiles. They do not test matrix computation. This is because I believe tests should be easy to verify, it should be clear if they are correct, while in such tests we would compare some magic numbers in 3D matrices.

This logic is however tested in render test - but only a part of it. For the matrix there are 3 different scenarios, depending on a image and terrain tile zoom level (might be higher, lower or the same). I tested only the case which was relevant for the reported issue. My assumption was that I shouldn't add too many render tests, I didn't want to make them too granular covering very narrow edge cases.

Let me know if and how I can improve that. Should I add unit or render tests? Or current state is ok?

HarelM:
Both render tests and unit tests are added to the coverage report, I think you might be able to find it in codecov site, unfortunately it doesn't get printed here due to a bug on their side. 
The requirement in general is that new code that was added will be covered (hopefully fully) with tests.

HarelM:
As far as I can tell from the code coverage report, the last else was not covered before this PR and seems like it is still not covered after this PR.
I'm not sure I know how to get to this code path, but since you have changed it, I think it would be wise to try and get to it using either a render test or unit test.
https://github.com/pstaszek/maplibre-gl-js/blob/4884e54f27a509b3b6535f0b7623fee6023d586e/src/source/terrain_source_cache.ts#L239

Let me know if there's anything else I can help with.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 
 - Fix setting visibility on custom layer ([#6883](https://github.com/maplibre/maplibre-gl-js/issues/6883) (by [melitele](https://github.com/melitele))
 - Hide leading and trailing control characters in `format` expressions ([#6907](https://github.com/maplibre/maplibre-gl-js/pull/6907) by [1ec5](https://github.com/1ec5))
+- Fix image sources being clipped at -180 and 180 longitude when terrain is enabled ([#4088](https://github.com/maplibre/maplibre-gl-js/issues/4088))
```
HarelM: Add "by" link please

HarelM:
Looks good overall, does this fix needed for canvas source as well?

pstaszek:
I checked canvas source as well - it is also fixed. `CanvasSource` inherits most of the logic from `ImageSource`.
I also added "by" link.

PR #6016, opened 2025-06-12
Fix/issue 3504 geolocate control lock loss on window resize

tderflinger:
## Launch Checklist

Issue: https://github.com/maplibre/maplibre-gl-js/issues/3504

This fix is only for the window resize issue. The zoom part will be done in a separate PR. 
I found that the problem with the resize event was that no originalEvent property was attached to it and therefore the code in geolocate_control.ts (lines 584-593) did not work as intended.

Did manual testing to confirm the fix works. On window resize now the lock does not disappear.

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


src/ui/map.ts
```diff
@@ -933,13 +933,13 @@ export class Map extends Camera {
         const fireMoving = !this._moving;
         if (fireMoving) {
             this.stop();
-            this.fire(new Event('movestart', eventData))
-                .fire(new Event('move', eventData));
+            this.fire(new Event('movestart', {originalEvent: eventData}))
```
HarelM: This kinda looks like a breaking change in terms of public API, isn't it?

tderflinger: Ok, I will only change code in the geolocation control, as you mentioned in the issue ticket. Doing an update to the fix. That is what you are aiming for, right?

HarelM: Yes, if this change is important we can do it as part of a breaking change version, but it should be well communicated...

tderflinger: Ok, I see. I will try to find a fix restricted to the geolocation control.

HarelM:
Any updates on this?

tderflinger:
@HarelM, thanks for your reminder. I was busy with other projects, but today I will continue with the issue. Keeping you updated. :)

HarelM:
Ping me when it's ready for review.
Thanks!

tderflinger:
@HarelM Yes, please review. :)

HarelM:
This is the outcome of changing to use resize observer instead of window resize event basically?
- i.e this PR: #2157

HarelM:
I see that there is a test to cover this scenario, however it looks like it uses the old style `resize` event, doesn't it?
https://github.com/maplibre/maplibre-gl-js/blob/6e52b8419592d3f39f4aad39aa26043c56fae078/src/ui/control/geolocate_control.test.ts#L497

I would recommend making sure the test is simulating what is actually happening in the code.

tderflinger:
@HarelM I have now added an integration test for the issue. It runs successfully on my local machine. The test breaks here on GitHub for the Ubuntu machine but curiously works for the Windows machine. I have already set the timeout value high, but to no avail. Is the Ubuntu Linux machine here on GitHub very slow?

HarelM:
I don't think it's very slow, but it's probably a machine with limited HW, so it's not strong...
Is it possible to avoid the integration tests and use unit tests only?
These integration tests are not as stable as I would want them to be, so I'm using them only as a last resort...

tderflinger:
I have now removed the integration and enabled a modified unit test.

test/integration/browser/browser.test.ts
```diff
@@ -138,7 +138,6 @@ describe('Browser tests', () => {
     });
 
     test('Resize viewport (page)', {retry: 3, timeout: 20000}, async () => {
-
```
HarelM: Any chance to revert these line removal changes to avoid having this for as part of this PR?

src/ui/control/geolocate_control.test.ts
```diff
@@ -24,6 +24,37 @@ function lngLatAsFixed(lngLat: LngLat, digits: number): {lat: string; lng: strin
...
+            this.contentRect = {
+                x: 0,
+                y: 0,
+                width: 0,
+                height: 0,
+                top: 0,
+                right: 0,
+                bottom: 0,
+                left: 0,
+                toJSON: () => ({}) // Mock toJSON method
```
HarelM: Is this toJSON method needed for the tests?

HarelM:
I've added a few minor comments, I'll merge it once addressed. 
THANKS!!

src/ui/control/geolocate_control.test.ts
tderflinger: Not really, so I removed it.

test/integration/browser/browser.test.ts
tderflinger: Done.

tderflinger:
Great, thanks. I made the changes.

tderflinger:
Please note, that this fix is only part of the original issue. There is the zoom missing which I will do next.

HarelM (approved):

Issue #4649, opened 2024-09-03
type of `_controlPositions` property of the Map class


JabSYsEmb:
I just noticed that the vscode intelljsense does not autocomplate for `_controlPositions` of `Map` class and found out it is due to that it is defined as `Record<string, HTMLElement>`, is that intentionally? Or typed incorrectly since `_controlPositions` cannot be positioned other than what the ControlPosition type is

https://github.com/maplibre/maplibre-gl-js/blob/160b8f9c3feb0c44c7ab233d4cedaac569236e0b/src/ui/map.ts#L448

I fix the type locally and get the nitelljsense working fine as in the following screenshot:

![image](https://github.com/user-attachments/assets/07f8b4ae-8efb-43c3-9aac-5570fe7ea5ba)

HarelM:
Probably an incorrect type, feel free to open a PR.

shreeharshshinde:
While exploring the code, I noticed that the `_controlPositions` field in [`map.ts`](https://github.com/maplibre/maplibre-gl-js/blob/main/src/ui/map.ts) is currently typed as:

```ts
_controlPositions: Record<string, HTMLElement>;
```
However, based on `[ControlPosition]`(https://github.com/maplibre/maplibre-gl-js/blob/main/src/ui/control/control.ts) , the only valid keys are:

```ts
'top-left' | 'top-right' | 'bottom-left' | 'bottom-right'
```

It might make sense to narrow the type to,
```ts
_controlPositions: Record<ControlPosition, HTMLElement>;
```
Changing the type to `Record<ControlPosition, HTMLElement>` improves type safety and aligns with the defined enum of valid control positions. It also enables proper IntelliSense/autocompletion support in editors like VSCode, enhancing developer experience.

HarelM:
See my comment above...

PR #6077, opened 2025-06-28
Fix incorrect type for `_controlPositions` in `Map.ts` controlposition type

shreeharshshinde:
This PR updates the type definition of the `_controlPositions` field in the `Map.ts` class from:

```ts
Record<string, HTMLElement>
```

to:

```ts
Record<ControlPosition, HTMLElement>
```
which is already defined in `src/ui/control/control.ts` as 
```ts
export const ControlPosition = 'top-left' | 'top-right' | 'bottom-left' | 'bottom-right';
```

Fixes the type Issue #4649


.vscode/settings.json
```diff
@@ -14,5 +14,6 @@
   },
   "[typescriptreact]": {
     "editor.defaultFormatter": "dbaeumer.vscode-eslint"
-  }
+  },
+  "liveServer.settings.port": 5501
```
HarelM: Can this be removed?

HarelM:
Looks good, thanks!

shreeharshshinde:
I’ve cleaned up the PR by removing the unrelated changes only `map.ts` is affected now as intended. Apologies for the number of commits earlier; I ran into some unexpected issues with branch syncing and resolving conflicts. All sorted now. Let me know if anything else needs adjustment

shreeharshshinde:
I used `Partial` so I can start `_controlPositions` as an empty object without TypeScript throwing an error. It still keeps the type safety for the known control positions and helps IntelliSense work properly while coding.

HarelM (approved):
THANKS!

PR #6100, opened 2025-07-02
update maplibre-cog-protocol to latest version (actually v0.5)

geraldo:


HarelM:
Are you sure that always using the latest version is safe enough from a documentation point of view where things always have to work?

geraldo:
I would assume that once a new version is published at unpkg.com it should work in the documentation example. But maybe @oscarfonts has a different opinion on that and we should better use a versioned source like https://unpkg.com/@geomatico/maplibre-cog-protocol@0.5.0/dist/index.js?

HarelM:
I think using a version number is safer for this project.

Issue #5978, opened 2025-06-04
Introduce padding for Popup


maciejmatu:
## User Story
 As a user I can pass `popupPadding` config to `Popup` so that I can limit it's positioning with custom bounds

## Rationale
There was a [similar feature request](https://github.com/mapbox/mapbox-gl-js/pull/12708) made in mapbox in the past, and I require such a feature in my application. Implementing something like that using current maplibre popups requires overriding default behaviour, and using internal functions. I managed to make it somehow work, but the code feels hacky and relies on copying the existing maplibre popup position calculations, so adding it internally would be much nicer. 

Here is an example of how this functionality would work:
Current behaviour:

https://private-user-images.githubusercontent.com/34219675/237748834-020d8de5-4794-4d66-a707-a75e024cf822.mov?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NDkwNzY2MDAsIm5iZiI6MTc0OTA3NjMwMCwicGF0aCI6Ii8zNDIxOTY3NS8yMzc3NDg4MzQtMDIwZDhkZTUtNDc5NC00ZDY2LWE3MDctYTc1ZTAyNGNmODIyLm1vdj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNTA2MDQlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjUwNjA0VDIyMzE0MFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTViMzA0MGYyYjMzMDEzYzk1ZmM5YzZkNjc2OTBkMzU2OTkxZTVmNGZkN2EyMzA4YjczYzQxZjcxNTMzODliOTAmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0In0.ar7NNsUQE51LFkHCV6ARxmZObDW7Uy2_Tk0b4-R7T3c

Desired behaviour:

[...]

PR #6052, opened 2025-06-23
Add popupPadding option to Popup class

yuiseki:
## Launch Checklist

  - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
  - [x] Briefly describe the changes in this PR.
  - [x] Link to related issues.
  - [x] Include before/after visuals or gifs if this PR includes visual changes.
  - [x] Write tests for all new functionality.
  - [x] Document any changes to public APIs.
  - [x] Post benchmark scores.
  - [x] Add an entry to `CHANGELOG.md` under the `## main` section.

  ## Summary

  This PR implements the `popupPadding` feature requested in issue #5978, allowing developers to define buffer zones around map container edges where popups should avoid being positioned. This is
  particularly useful for preventing popups from being hidden behind UI overlays like headers, sidebars, or toolbars.

  **Closes #5978**

  ## Changes Made

  **New API Features:**
  - Added `popupPadding` option to `PopupOptions` interface supporting multiple input formats:
    - Number: Equal padding on all sides
    - PointLike/Array: Horizontal and vertical padding
    - 4-element array: CSS-style [top, right, bottom, left] padding
    - Object: Explicit {top, right, bottom, left} properties
  - Added `setPopupPadding(padding?: PopupPadding)` method for dynamic updates
  - Added `PopupPadding` type definition with comprehensive format support

  **Implementation Details:**
  - Enhanced automatic anchor selection logic in `_update()` method to respect padding constraints
  - Added `normalizePopupPadding()` function with robust input validation and NaN/Infinity handling
  - Maintained complete backward compatibility - existing code unaffected
  - Manual anchor selection bypasses padding calculations (preserves existing behavior)
  - Padding constraints work seamlessly with existing offset system

  ## Testing

  **Comprehensive test coverage with 90 test cases including:**

  **Core Functionality:**
  - Multiple padding input formats (number, object, array)
  - Dynamic padding updates via `setPopupPadding()`
  - Interaction with existing offset system

  **Critical Compatibility Tests:**
  - Backward compatibility verification across 9 different map positions
  - Manual anchor behavior preservation (all 9 anchor positions tested)
  - TrackPointer functionality remains unchanged
  - Performance impact assessment showing no degradation

  **Edge Case Handling:**
  - NaN, Infinity, negative values, and extreme numbers
  - Mixed problematic values in object/array formats
  - Zero padding vs undefined padding equivalence

  **Integration Tests:**
[...]

PR #6106, opened 2025-07-05
feat(popup): introduce configurable popupPadding to clamp popup within map viewport (#5978)

shreeharshshinde:
### Popup Padding Support

This PR introduces a new `popupPadding` option to the `Popup` class to address [issue #5978](https://github.com/maplibre/maplibre-gl-js/issues/5978).  
The padding ensures that the popup remains within the visible area of the map by clamping its position based on user-defined or default padding values.

---

###  Key Changes

- Added `popupPadding` to `PopupOptions`:

  ```ts
  popupPadding?: {
    top: number;
    bottom: number;
    left: number;
    right: number;
  };

- Default padding values are set to : 
  ```ts
    {
      top: 0,
      bottom: 0,
      left: 0,
      right: 0
    }
  ```
- Example Usage
   ```ts
   const popup = new maplibregl.Popup({
    popupPadding: {
      top: 40,
      right: 60,
      bottom: 40,
      left: 60
    }
  })
  .setLngLat([lng, lat])
  .setHTML("<h1>Hello World</h1>")
  .addTo(map);
  ```


HarelM:
How is this PR different than #6052?

Also note that creating a default that is different than current implantation is a breaking change and should not be made without proper planning of a breaking change version. 
As a stop gap I would advise to keep it to 0 and suggest a breaking change feature when there are plans to create a breaking change version (also I don't think this breaking change is really needed)

shreeharshshinde:
Thanks for the feedback! 
I’ve updated the default `popupPadding` to `{top: 0, right: 0, bottom: 0, left: 0}` so there’s no breaking change. Users can opt in by setting a custom value.
Let me know if there's anything else you'd like me to adjust!

Also, I personally prefer this approach as it stays minimal and intuitive without introducing additional API surface. Let me know if you'd like to move forward with this version or prefer closing in favor of #6052.

HarelM:
I think that allowing to `setPadding` is needed to update the popup padding dynamically after the popup was created.
We can introduce this option first and then the other one if it is required, but I think it's a fuller solution.
In any case, tests are needed, also the CSS type is probably better, as I mentioned in the other PR.

HarelM:
@shreeharshshinde what's the status of this PR?

HarelM:
How is this PR different than #6052?

(Besides the fact that the other PR is "stale")

shreeharshshinde:
The previous implementation mentioned in the feedback included:

- A `normalizePopupPadding` function with robust input validation
- Support for multiple input formats (number, PointLike, array, object)

However, the current implementation in the codebase uses a simpler approach:

- Only object format with string values for padding
- A `setPadding` method for dynamic updates
- Direct application of CSS padding styles to the popup container

src/ui/popup.ts
```diff
@@ -95,6 +101,22 @@ export type PopupOptions = {
...
+     * CSS-style padding applied to the popup's positioning logic.
+     * This ensures the popup stays within the map viewport and doesn't overlap padded edges.
+     * Useful for UI where parts of the map are visually obstructed (e.g. sidebars, headers).
+     *
+     * @example
+     * ```
+     * { top: '20px', right: '30px', bottom: '40px', left: '10px' }
+     * ```
+     */
+    popupPadding?: {
```
HarelM: CSS style can have 1, 2 or 4 values, which is not the case here, I would recommend updating the docs or the type.

src/ui/popup.ts
```diff
@@ -166,6 +188,19 @@ const focusQuerySelector = [
  * **Event** `close` of type {@link Event} will be fired when the popup is closed manually or programmatically.
  */
 export class Popup extends Evented {
+    /**
+     * Dynamically set the popup's CSS padding after creation.
```
HarelM, line 206:
```suggestion
     * Dynamically set the popup's CSS padding.
```

src/ui/popup.ts
```diff
@@ -615,6 +650,14 @@ export class Popup extends Evented {
         if (this.options.maxWidth && this._container.style.maxWidth !== this.options.maxWidth) {
             this._container.style.maxWidth = this.options.maxWidth;
         }
+        // Apply popupPadding as CSS
+        const padding = this.options.popupPadding;
```
HarelM: This part looks duplicated.

shreeharshshinde:
@HarelM I am not getting why still those build test are failing ? What is the reason behind it ?

HarelM:
You need to update the build size to accommodate for the code changes.

PR #6131, opened 2025-07-14
Update pbf, vt-pbf, point-geometry, vector-tile

mwilsnd:
## Launch Checklist

Updates vt-pbf, unblocking updates to pbf, point-geometry and vector-tile. Older versions of vt-pbf depended on incompatible pbf versions, resulting in duplicate dependencies in the final bundle. vt-pbf also contained a JS version of `GeoJSONWrapper`, resulting in a little extra bloat. This PR removes `GeoJSONWrapper` and depends on the new typescript version from `@maplibre/vt-pbf` (file was relocated to that dependency).

Closes #6037
Closes #4393
Closes #4378

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


src/source/geojson_worker_source.ts
```diff
@@ -81,11 +81,11 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
             return null;
         }
 
-        const geojsonWrapper = new GeoJSONWrapper(geoJSONTile.features);
+        const geojsonWrapper = new GeoJSONWrapper(geoJSONTile.features, {version: 2, extent: EXTENT});
         // Encode the geojson-vt tile into binary vector tile form.  This
         // is a convenience that allows `FeatureIndex` to operate the same way
         // across `VectorTileSource` and `GeoJSONSource` data.
-        let pbf = vtpbf(geojsonWrapper);
+        let pbf: Uint8Array = fromVectorTileJs(geojsonWrapper);
```
HarelM: Does this mean the return value of this method is not defined well?

mwilsnd: The return was defined explicitly as Uint8Array, the explicit storage type there is redundant and I've gone ahead and removed it.

0xFA11:
just came here to repeat our early finding in our project, this specific commit breaks GeoJSON source data for us. we're still investigating the issue and will try to come back with a minimal repro soon. we wanted to come here to raise a flag sooner than later.

wayofthefuture:
@0xFA11 I think we are having the same issue. The error occurs only from 5.6.1 -> 5.6.2 and here is the stack:

```
Uncaught Error: unknown feature value
    at readValueMessage (maplibre-gl-dev.js:29027:1)
    at readLayer (maplibre-gl-dev.js:29005:1)
    at Pbf.readFields (maplibre-gl-dev.js:30343:1)
    at new VectorTileLayer (maplibre-gl-dev.js:28973:1)
    at readTile (maplibre-gl-dev.js:29051:1)
    at Pbf.readFields (maplibre-gl-dev.js:30343:1)
    at new VectorTile (maplibre-gl-dev.js:29040:1)
    at FeatureIndex.loadVTLayers (maplibre-gl-dev.js:34272:1)
    at FeatureIndex.query (maplibre-gl-dev.js:34279:1)
    at Tile.queryRenderedFeatures (maplibre-gl-dev.js:42840:1)
```

for this block of code:
```typescript
function readValueMessage(pbf) {
    let value = null;
    const end = pbf.readVarint() + pbf.pos;

    while (pbf.pos < end) {
        const tag = pbf.readVarint() >> 3;

        value = tag === 1 ? pbf.readString() :
            tag === 2 ? pbf.readFloat() :
            tag === 3 ? pbf.readDouble() :
            tag === 4 ? pbf.readVarint64() :
            tag === 5 ? pbf.readVarint() :
            tag === 6 ? pbf.readSVarint() :
            tag === 7 ? pbf.readBoolean() : null;
    }
    if (value == null) {
        throw new Error('unknown feature value');
    }

    return value;
}
```

mwilsnd:
Are you able to share a reproduction? I can take a look at this as soon as I have a way to trigger the issue.

wayofthefuture:
@0xFA11 I was unable to get a codepen to reproduce this issue, so I can't quite figure it out. Any idea how to replicate this in a jsbin or codepen?

0xFA11:
I think I found not 1 but 2 problematic areas — first of all, here's my code snippet:

```
<!doctype html>
<html>
    <head>
        <meta charset="utf-8" />
        <title>MapLibre Repros</title>
        <meta name="viewport" content="width=device-width, initial-scale=1" />
        <script src="https://unpkg.com/maplibre-gl@5.6.2/dist/maplibre-gl.js"></script>
        <link href="https://unpkg.com/maplibre-gl@5.6.2/dist/maplibre-gl.css" rel="stylesheet" />
        <style>
            body {
                margin: 0;
                padding: 0;
            }
            html,
            body,
            #map {
                height: 100%;
            }
        </style>
    </head>
    <body>
        <div id="map"></div>
        <script>
                container: "map", // container id
                style: "https://demotiles.maplibre.org/style.json", // style URL
                zoom: 1, // starting zoom
            });

            map.on("load", () => {
                {
                    const redData = {
                        type: "FeatureCollection",
                        features: [],
                    };

                    redData.features.push({
                        id: "foo",
                        type: "Feature",
                        geometry: {
                            type: "Point",
                            coordinates: [0.1276, 51.5072],
                        },
                        properties: {},
                    });

                    redData.features.push({
                        id: "bar",
                        type: "Feature",
                        geometry: {
                            type: "Point",
                            coordinates: [2.3514, 48.8575],
                        },
                        properties: {},
                    });

                    redData.features.push({
                        id: "fizz",
                        type: "Feature",
                        geometry: {
                            type: "Point",
                            coordinates: [-74.006, 40.7128],
                        },
                        properties: {},
                    });

                    redData.features.push({
                        id: "buzz",
                        type: "Feature",
                        geometry: {
                            type: "Point",
                            coordinates: [28.9784, 41.0082],
                        },
                        properties: {},
                    });

                    map.addSource("red-source", {
                        type: "geojson",
                        data: redData,
                    });

                    map.addLayer({
                        id: "red-layer",
                        type: "circle",
                        source: "red-source",
                        paint: {
                            "circle-radius": 5.0,
                            "circle-opacity": 1.0,
                            "circle-color": "red",
                        },
                    });

                    setTimeout(() => {
                        console.log("query red source features:", map.querySourceFeatures("red-source"));
                        console.log("internal red source features:", map.getSource("red-source")._data.features);
                    }, 1000);
                }

                {
                    const blueData = {
                        type: "FeatureCollection",
                        features: [],
                    };

                    blueData.features.push({
                        id: 10,
                        type: "Feature",
                        geometry: {
                            type: "Point",
                            coordinates: [0.1276, 52.5072],
                        },
                        properties: {},
                    });

                    blueData.features.push({
                        id: 11,
                        type: "Feature",
                        geometry: {
                            type: "Point",
                            coordinates: [2.3514, 49.8575],
                        },
                        properties: {},
                    });

                    blueData.features.push({
                        id: 12,
                        type: "Feature",
                        geometry: {
                            type: "Point",
                            coordinates: [-74.006, 41.7128],
                        },
                        properties: {},
                    });

                    blueData.features.push({
                        id: 13,
                        type: "Feature",
                        geometry: {
                            type: "Point",
                            coordinates: [28.9784, 42.0082],
                        },
                        properties: {},
                    });

                    map.addSource("blue-source", {
                        type: "geojson",
                        data: blueData,
                    });

                    map.addLayer({
                        id: "blue-layer",
                        type: "circle",
                        source: "blue-source",
                        paint: {
                            "circle-radius": 5.0,
                            "circle-opacity": 1.0,
                            "circle-color": "blue",
                        },
                    });

                    setTimeout(() => {
                        console.log("query blue source features:", map.querySourceFeatures("blue-source"));
                        console.log("internal blue source features:", map.getSource("blue-source")._data.features);
                    }, 1000);
                }
            });
        </script>
    </body>
</html>
```

console output with 5.6.2
image

console output with 5.6.1
image

first problem is, with 5.6.2 feature ids are being coerced to number 0, in 5.6.1 they were undefined. I think this was the problem in our codebase, we have number of custom layers relying on feature id numbers.

second problem is, `querySourceFeatures()` API doesn't work for string ids. in both 5.6.1 and 5.6.2 outputs, features in red source with string ids come back either undefined (5.6.1) or 0 (5.6.2). however, if you access source data directly with _internal API (`map.getSource("red-source")._data.features`), you can see that string ids are there — so, maybe there's another bug in `querySourceFeatures()` implementation.

as a workaround, we can use `map.getSource("red-source")._data.features` for now in our codebase but I think both of these are legitimate bugs need to be addressed.

mwilsnd:
Thanks for this information! Sounds like feature IDs are a problem with vt-pbf or one of the related dependencies. I'll take a look at `querySourceFeatures` as well.

HarelM:
I don't think string IDs for geojson were ever supported as geojson source is using MVT under the hood, which doesn't supports string IDs as far as I remember...

wayofthefuture:
Is this relevant?

https://github.com/maplibre/maplibre-gl-js/blob/50da15ce18fc8b68396f76d43ef512694a3d195d/src/source/geojson_source_diff.ts#L4

https://github.com/maplibre/maplibre-gl-js/blob/50da15ce18fc8b68396f76d43ef512694a3d195d/src/source/geojson_source_diff.ts#L56-L58

HarelM:
This needs to be defined using promoteId, it's not the default behavior.

wayofthefuture:
So I guess I'm lost. I can't reproduce the error using string or number ids on a GeoJSON Source. We heavily use string ids for features so not sure where this leaves us. The suspect line of code in our project is let features = map.queryRenderedFeatures(e.point); but if I put it in a codepen it doesn't throw. @0xFA11 Any chance you can get the issue to reproduce using this codepen? https://codepen.io/wayofthefuture/pen/YPyegad

mwilsnd:
`readValueMessage` should be reading tile property values. I do see property values are being converted to strings in vt-pbf which isn't quite right, but shouldn't break parsing the protobuf, as seems to be happening in your case. The tag being read indicates what type the property value following the tag in the buffer should be, so that tag is either out of range or the associated `read<Type>` function is returning null. Either way, I would think the protobuf itself has to be malformed to cause that.

I would focus on any properties you might be including in your GeoJSON data and what types they are, the specification only supports strings, numbers and bools.

wayofthefuture:
Ok I was able to replicate it in https://codepen.io/wayofthefuture/pen/YPyegad all you have to do is click the feature (and view browser console not codepen console):
```javascript
      properties: {
        color: randomColor(),
        // testString: 'test',
        // testNumber: 123,
        // obj: {a:1},
        undef: undefined
      }
```
@mwilsnd Yes it occurs if I set a property to "undefined". So I guess I should just remove undefined props. It does not seem to throw if we used "null" however.

wayofthefuture:
@mwilsnd We resolved our issue with your advice. Apologies for the use of your time for an issue that was outside the scope of MapLibre!

@0xFA11 Hope this helps we just put this inline when creating any features:

```javascript
        geometry: {
            type: 'Point',
            coordinates: coordinate
        },
        properties: copyPropertiesForGeoJson(properties, {multiType})
```

```javascript
function copyPropertiesForGeoJson(properties = {}, addProps = {}) {
    const newProps = {};

    for (const key of Object.keys(properties)) {
        let value = properties[key];
        let type = typeof value;
        if (type === 'string' || type === 'number' || type === 'boolean') {
            newProps[key] = value;
        }
    }
    for (const key of Object.keys(addProps)) {
        let value = addProps[key];
        let type = typeof value;
        if (type === 'string' || type === 'number' || type === 'boolean') {
            newProps[key] = value;
        }
    }

    return newProps;
}
```

0xFA11:
I think I could also move id into properties but that would mean their ids will now be undefined, null or 0 — and I wouldn't be able to use id-based APIs and style expressions.

I'm curious if potential fix would be to wire GeoJSON feature's id field into features inside map's internal data. we might say this is a feature request rather than a bug, I'm not sure but I expected the feature with string id being pushed to be added to map data features with its identical string id (and then later to be used in queries, expressions etc.)

image

lucaswoj:
> I'm curious if potential fix would be to wire GeoJSON feature's id field into features inside map's internal data. 

How would this be different from `promoteId`?

Issue #5499, opened 2025-02-13
gulp-sass plugin error with latest version 5.1.0


BusinessClick:
When the gulp-sass plugin process the latest versione 5.1.0 returns the following error:

Error in plugin 'sass'
Message:
    temp/maplibre-gl.css
Error: Function rgb is missing argument $green.
        on line 1 of temp/maplibre-gl.css
>> tion:relative;-webkit-tap-highlight-color:rgb(0 0 0/0)}.maplibregl-canvas{le

This not happens with 5.0.1 version.

hannes011:
I get another error, with the same line of CSS, when I try to import it into my other LESS files:
Dimension is not a number.
      Error in .../node_modules/maplibre-gl/dist/maplibre-gl.css (line 1, column 107)

shreeharshshinde:
Updated the CSS to use the compatible rgb(0, 0, 0, 0) format instead of rgb(0 0 0/0) to ensure compatibility with older gulp-sass versions and prevent the missing argument $green error.

Let me know if any adjustments are needed!

HarelM:
Updated where? I don't see any PR related to this...

shreeharshshinde:
Just a minute I'll release PR

PR #6161, opened 2025-07-18
Fix: Replace incompatible rgb() CSS syntax to support older gulp-sass versions (#5499)

shreeharshshinde:
This PR resolves [#5499](https://github.com/maplibre/maplibre-gl-js/issues/5499) by updating the CSS color syntax in maplibre-gl.css.

### Changes:
Replaced `rgb(0 0 0 / 0)` with the compatible `rgb(0, 0, 0, 0)` format.
This change ensures compatibility with gulp-sass v5.0.1

### Tested:
Verified that the build completes successfully with both gulp-sass v5.0.1 and the latest version.
Confirmed that the rendered CSS behaves as expected visually.


HarelM:
Is there a lint rule you can add to make sure this stays ok in the future?

shreeharshshinde:
Added a rule to block the modern rgb() format with spaces to avoid build errors and only allow the old comma-separated version like rgb(0, 0, 0).

HarelM:
Did you check that the previous version of the css fails compiling with this rule?

HarelM (approved):

shreeharshshinde:
Confirmed that the previous CSS using rgba() fails the linter due to the color-function-alias-notation rule. The lint config is correctly catching legacy formats now.
image

Issue #6154, opened 2025-07-17
Memory leak when updating a GeoJSON source faster than can be processed


andrewda:
When providing rapid updates to a GeoJSON source (e.g., in a `requestAnimationFrame` or fast `setInterval` loop), it is possible to update the source faster than these updates can be applied to the map. This is particularly noticeable with geometries that are slower to render, like large polygons or linestrings.

If updates are provided using `GeoJSONSource.updateData`, the `tasks` within the worker's `Actor` grows indefinitely, until the worker runs out of memory. If updates are provided using `GeoJSONSource.setData`, the memory usage of both the worker and main threads grows indefinitely, primarily with the `JSON.stringify`'d representations of the data. 

This may, to an extent, be the expected behavior for some use-cases. For example, if I have many features and am updating all of them, there is no way to skip any of these feature updates without losing data. On the other hand, if I have a small number of features and am very rapidly updating the geometry and/or specific properties, I only need the latest update to take effect -- there's no need to queue and apply all the now-outdated updates previously supplied.

My naive recommendation (as someone still pretty new to the maplibre codebase) would be to allow specifying a replaceable update id, which could override the current randomly generated `message.id` in `Actor.sendAsync`. A possible API could look like:

```ts
GeoJSONSource.setData(data: GeoJSON | string, replaceableUpdateId?: number);
GeoJSONSource.updateData(diff: GeoJSONSourceDiff, replaceableUpdateId?: number);
```

With the `replaceableUpdateId` being passed into `GeoJSONSource._updateWorkerData` and subsequently `Actor.sendAsync`. If no `replaceableUpdateId` is provided, then the behavior is the same as today, with `Actor.sendAsync` randomly generating an id. If the provided id already exists in `Actor.tasks`, its data could be replaced with the new value, without changing its position in the `Actor.taskQueue`.

I'm sure smarter solutions than this exist (e.g., intelligently merging data/diffs when multiple exist in the task queue), but my goal was to provide minimal changes to existing behavior while resolving this edge condition for the rare times that it comes up. If this seems like a reasonable solution/workaround, I'm happy to submit a PR -- otherwise, would love to discuss better approaches before implementing.

(Related issue: https://github.com/maplibre/maplibre-gl-js/issues/4650)

**maplibre-gl-js version**: 5.6.1

**browser**: Chrome 138

### Link to Demonstration

https://jsbin.com/weroxucaqu/edit?html,output

[...]

PR #6163, opened 2025-07-22
Fix performance and memory leak in GeoJSONSource

andrewda:
## Summary

This PR addresses the issues discussed in https://github.com/maplibre/maplibre-gl-js/issues/6154, where rapid calls to `GeoJSONSource.setData` and `GeoJSONSource.updateData` result in substantial memory usage, which can lead to webpage crashing.

To resolve this issue, this PR refactors how data is sent from the GeoJSONSource to the worker. Instead of firing off a message to the worker every time `setData` or `updateData` is called, we now only send one message at a time. If `setData` or `updateData` are called while a message to the worker is pending, that change is stored pending the response of the previous message. If `updateData` is called multiple times while waiting for a response from the worker, the diffs are combined. If `setData` is called, any previously pending data and data diffs are replaced.

## TODO

I've added some basic tests, but would still like to add more thorough testing. I am putting this PR up preemptively to get feedback on the approach.

## Launch Checklist

 - [x] Link to related issues.


src/source/geojson_source.ts
```diff
@@ -244,8 +247,25 @@ export class GeoJSONSource extends Evented implements Source {
      * @param diff - The changes that need to be applied.
      */
     updateData(diff: GeoJSONSourceDiff): this {
-        this._updateWorkerData(diff);
+        if (!this._pendingDataDiff) {
+            this._pendingDataDiff = {};
+        }
 
+        // Merge pending data diff with the new diff
```
HarelM: I think there's a whole file that takes care of the diff, I think it would be wise to move this logic there.

src/source/geojson_source.ts
```diff
@@ -375,49 +395,71 @@ export class GeoJSONSource extends Evented implements Source {
      * Responsible for invoking WorkerSource's geojson.loadData target, which
      * handles loading the geojson data and preparing to serve it up as tiles,
      * using geojson-vt or supercluster as appropriate.
-     * @param diff - the diff object
      */
-    async _updateWorkerData(diff?: GeoJSONSourceDiff) {
+    async _updateWorkerData(): Promise<void> {
+        if (this._pendingLoads > 0) return;
```
HarelM: Should this still be a counter and not a boolean like "isUpdatingWoker" or something that says that only one can happen at a time?

src/source/geojson_source.ts
```diff
@@ -128,6 +128,9 @@ export class GeoJSONSource extends Evented implements Source {
     _collectResourceTiming: boolean;
     _removed: boolean;
 
+    _pendingData: GeoJSON.GeoJSON | string | undefined;
+    _pendingDataDiff: GeoJSONSourceDiff | undefined;
```
HarelM: Can these two be joined to a single variable somehow? Seems like they are connected...

src/source/geojson_source.ts
```diff
@@ -149,7 +152,7 @@ export class GeoJSONSource extends Evented implements Source {
         this.actor = dispatcher.getActor();
         this.setEventedParent(eventedParent);
 
-        this._data = (options.data as any);
+        this._data = this._pendingData = (options.data as any);
```
HarelM: I find this confusing... Consider splitting this into two lines.
But overall data pendingData and pendingDataDiff is complicated, see if you can find a way to either document what every field is used for and combine them to a single struct that hold different fields maybe, IDK...

src/source/geojson_source.test.ts
```diff
@@ -487,6 +487,29 @@ describe('GeoJSONSource.getData', () => {
     });
 });
 
+describe('GeoJSONSource.updateData', () => {
+    test('it combines multiple diffs when data is loading', async () => {
```
HarelM: This looks good, I think more tests are expected here if I fully understand the change:
- updateData shortly after setData
- setData shortly after updateData
- setData shortly after setData (? not sure if it's really needed)

Any other scenarios you might think that are worth testing.

src/source/geojson_source.test.ts
```diff
@@ -487,6 +487,29 @@ describe('GeoJSONSource.getData', () => {
...
+        const source = new GeoJSONSource('id', {data: {}} as GeoJSONSourceOptions, wrapDispatcher({}), undefined);
+        source._pendingLoads = 1; // Simulate a loading state
```
HarelM: If there's a way to test this without using "internal" fields it would be great.

HarelM:
This looks like a good improvement.

src/source/geojson_source.ts
andrewda: Agreed that the current pendingData/pendingDataDiff implementation is pretty complicated. The reason these are currently structured as two separate variables is for the scenario in which a change is currently pending on the worker, and then `setData` is called, followed by an `updateData`. We want the `setData` to take effect first, then have the `updateData` diff apply on top of that new data.

If you'd rather these get combined into one variable, I could see this being implemented as `_pendingWorkerUpdate: { data?: GeoJSON; diff?: GeoJSONSourceDiff }`

src/source/geojson_source.ts
andrewda: Yep, agreed that a boolean makes more sense here.

src/source/geojson_source.ts
```diff
@@ -375,54 +378,75 @@ export class GeoJSONSource extends Evented implements Source {
...
+
+            this._pendingWorkerUpdate.data = undefined;
+        } else if (diff) {
             options.dataDiff = diff;
-        } else if (typeof this._data === 'string') {
-            options.request = this.map._requestManager.transformRequest(browser.resolveURL(this._data as string), ResourceType.Source);
-            options.request.collectResourceTiming = this._collectResourceTiming;
-        } else {
-            options.data = JSON.stringify(this._data);
+            this._pendingWorkerUpdate.diff = undefined;
```
wayofthefuture, line 404: delete instead? this._pendingWorkerUpdate.diff

wayofthefuture:
Good job Andrew this is interesting and I look forward to testing it under high volume source updates. Do you think in updateData when merging the diffs it will retain the order of updates so that newer feature updates of the same id will occur last? I see in the merge the prior array is first so that is likely enough I believe...

src/source/geojson_source.test.ts
HarelM: The sequence of setData then updateData then setData, and other flows need to written as tests here as well as there is logic around which method was invoked in which order...

wayofthefuture:
This PR matters to us because updateData is at the core of our implementation. After playing with this, thinking and testing, I am having trouble understanding the need for this PR. In the [original issue](https://github.com/maplibre/maplibre-gl-js/issues/6154#issuecomment-3085449744), the memory leak was relating to the accumulation of pending updates (_pendingLoads) in updateData:

https://github.com/maplibre/maplibre-gl-js/blob/d97316776dca4f120ec691af5f6a98144586453c/src/source/geojson_source.ts#L390-L394

This PR deprecates _pendingLoads in favor of much more complex logic. However, _pendingLoads is a quasi queue count wrapped around the promise returned from actor.sendAsync. Hence, you could just simply return the promise in the calling function (async updateData / await) to solve the issue of the memory leak:

```javascript
    //in a breaking way (updateData currently does not return a promise)
    async updateData(diff: GeoJSONSourceDiff): Promise<this> {
        await this._updateWorkerData(diff);
        return this;
    }

    //or in a non-breaking way
    async awaitUpdateData(diff: GeoJSONSourceDiff): Promise<this> {
        await this._updateWorkerData(diff);
        return this;
    }
```

In the Codepen below, returning the promise and awaiting updateData completely solved the memory leak for me. So I'm not sure if the additional complexity this PR provides is necessary.

https://codepen.io/wayofthefuture/pen/empZLwZ?editors=0010

Please let me know if this solves the leak for you as well, or what I am missing.

andrewda:
@wayofthefuture I agree that having an async version of `setData`/`updateData` would be handy for giving us more control over the exact update timing. That said, I'd really like the default/synchronous behavior to also prevent performance issues internally (what this PR attempts to accomplish). That way, one can use `setData`/`updateData` as fire-and-forget functions without any internal knowledge about maplibre's worker model (it isn't immediately intuitive to a new user that calling `setData` repeatedly could lead to memory and performance issues). 

Having `setDataAsync`/`updateDataAsync` doesn't seem mutually exclusive with the changes in this PR -- in fact, if I believe there would be exactly no difference in your codepen between `maplibre-gl-js@main` and this PR, given that in both cases a response from the worker is awaited before any subsequent change is made.

I'm curious if you have any specific concerns about this implementation, rather than just the general increase in complexity (which I don't actually think is too bad, all things considered)?

> Do you think in updateData when merging the diffs it will retain the order of updates so that newer feature updates of the same id will occur last? I see in the merge the prior array is first so that is likely enough I believe...

This is something I'm planning to add a test for, just to be certain!

wayofthefuture:
Well said. Truthfully, I am not of the expertise of both you and Harel so I'll leave it to you guys. High frequency updates with hundreds of features is essential to our software which depends on updateData. My understanding of Maplibre is limited.

src/source/geojson_source.ts
```diff
@@ -375,54 +378,75 @@ export class GeoJSONSource extends Evented implements Source {
...
+        if (data) {
+            if (typeof data === 'string') {
+                options.request = this.map._requestManager.transformRequest(browser.resolveURL(data as string), ResourceType.Source);
+                options.request.collectResourceTiming = this._collectResourceTiming;
+            } else {
+                options.data = JSON.stringify(data);
+            }
+
+            this._pendingWorkerUpdate.data = undefined;
+        } else if (diff) {
```
HarelM, line 402: What happens if I call `setData` and shortly after `updateData` while the worker is busy?
It looks like me last update data will be lost, wouldn't it?

src/source/geojson_source_diff.ts
```diff
@@ -170,3 +170,59 @@ export function applySourceDiff(updateable: Map<GeoJSONFeatureId, GeoJSON.Featur
         }
     }
 }
+
+export function mergeSourceDiffs(
```
HarelM, line 174: Can you check that this method is fully covered by the tests in terms of code coverage?

src/source/geojson_source.ts
andrewda: I will add a test for this, but the expected behavior is:
1. Start with idle worker.
2. `setData` updates `this._pendingWorkerUpdate.data`, and immediately kicks off `_updateWorkerData`.
3. `updateData` updates `this._pendingWorkerUpdate.diff`, and since the worker is currently being updated, `_updateWorkerData` does nothing.
4. When the `setData` worker update completes, `_updateWorkerData` checks for whether any pending updates are set in `this._pendingWorkerUpdate`, and if so, kicks off another call to `_updateWorkerData`. In this case, since we executed `updateData`, `this._pendingWorkerUpdate.diff` is set and `_updateWorkerData` is called again to process the diff. https://github.com/andrewda/maplibre-gl-js/blob/f04cd956db9b511f498c57234a24e871d95230d4/src/source/geojson_source.ts#L441-L444

andrewda:
@HarelM here's the unit test coverage file. All new code is covered.

[coverage.zip](https://github.com/user-attachments/files/21391709/coverage.zip)

CleanShot 2025-07-23 at 09 29 45@2x

CleanShot 2025-07-23 at 09 30 10@2x

andrewda:
I'm having a bit of trouble debugging the failed render tests, since they don't appear to be failing locally. Any advice on where to start there?

src/source/geojson_source.test.ts
andrewda: These conditions should now be covered

src/source/geojson_source.test.ts
andrewda: Done

src/source/geojson_source_diff.ts
andrewda: See my latest comment for coverage stats

HarelM:
Did you run the full build before running the render tests? 
npm run build dist

Also if this still not reproduced locally, try using spaces.
You can run only the relevant tests, you don't need to run the entire suite...

src/source/geojson_source.ts
HarelM: I mean the scenario when calling setData while already updating the worker and then updateData while setData hasn't been sent to the worker still. 
If this is coded as expected, then great.

src/source/geojson_source.test.ts
```diff
@@ -27,6 +28,16 @@ const mockDispatcher = wrapDispatcher({
     sendAsync() { return Promise.resolve({}); }
 });
 
+const waitForLoaded = (source: GeoJSONSource) => {
```
HarelM: I think there's a wait for event test utility somewhere, isn't there?

src/source/geojson_source.test.ts
```diff
@@ -462,7 +473,7 @@ describe('GeoJSONSource.update', () => {
         source.load();
 
         await sleep(10);
-        expect(loadDataCalls).toBeGreaterThan(2);
+        expect(spy.mock.calls.length).toBeGreaterThan(2);
```
HarelM, line 466: toHaveBeenCalled...

src/source/geojson_source.test.ts
```diff
@@ -487,6 +498,149 @@ describe('GeoJSONSource.getData', () => {
...
+        } satisfies GeoJSONSourceDiff;
+        const update2 = {
+            remove: ['4'],
+            add: [{id: '5', type: 'Feature', properties: {}, geometry: {type: 'LineString', coordinates: []}}],
+            update: [{id: '6', addOrUpdateProperties: [], newGeometry: {type: 'LineString', coordinates: []}}]
+        } satisfies GeoJSONSourceDiff;
+        source.updateData(update1);
+        source.updateData(update2);
+
+        expect(source._pendingWorkerUpdate).toEqual({
```
HarelM: Is it possible not to check the content of internal field, but instead check what is being sent to the worker?

src/source/geojson_source.test.ts
```diff
@@ -487,6 +498,149 @@ describe('GeoJSONSource.getData', () => {
...
+                add: [{id: '5', type: 'Feature', properties: {}, geometry: {type: 'LineString', coordinates: []}}],
+                update: [{id: '6', addOrUpdateProperties: [], newGeometry: {type: 'LineString', coordinates: []}}]
+            },
+        });
+
+        await waitForLoaded(source);
+        await waitForLoaded(source);
+
+        expect(source._pendingWorkerUpdate).toEqual({data: undefined, diff: undefined});
+        expect(spy).toHaveBeenCalledTimes(3);
```
HarelM: Can we clear the spy before the two updates or is the first invocation interesting?

src/source/geojson_source.test.ts
```diff
@@ -487,6 +498,149 @@ describe('GeoJSONSource.getData', () => {
...
+                return new Promise((resolve) => {
+                    setTimeout(() => resolve({}), 0);
+                });
+            }
+        });
+
+        const source = new GeoJSONSource('id', {data: {}} as GeoJSONSourceOptions, mockDispatcher, undefined);
+
+        // Wait for initial data to be loaded
+        await source.load();
```
HarelM: This is a different way to wait for the source to load, it's a bit confusing...

src/source/geojson_source.ts
HarelM: I've looked at the tests and haven't seen this scenario - i.e. setData and then updateDatab are both queued and then applied one after the other (in the right order).

src/source/geojson_source.test.ts
r2dassonville: `toHaveBeenCalled`  loses the meaning of the previous test (which is specifically checking that it's been called at least twice). There is `toHaveBeenCalledTimes`, but this expects an exact value and it doesn't seem like there's a "greater than" equivalent (which is needed since this test seems to be inconsistent in the number of calls to `sendAsync`).

src/source/geojson_source.test.ts
andrewda: Yep good point, I'll switch this to `waitForEvent`

wayofthefuture:
Just an fyi I've done some local testing of your PR and I am seeing a nice performance boost versus my method above and am also experiencing no memory leak.

src/source/geojson_source.test.ts
andrewda: Good idea -- done in my latest commit.

src/source/geojson_source.test.ts
andrewda: Yep, I'm now only checking the spied sendAsync value 👍

src/source/geojson_source.test.ts
andrewda: `toHaveBeenCalled`  loses the meaning of the previous test (which is specifically checking that it's been called at least twice). There is `toHaveBeenCalledTimes`, but this expects an exact value and it doesn't seem like there's a "greater than" equivalent (which is needed since this test seems to be inconsistent in the number of calls to `sendAsync`).

src/source/geojson_source.test.ts
andrewda: Yeah good call -- switched to `waitForEvent`

src/source/geojson_source.ts
andrewda: Ah, I understand what you mean now. I've added this test in my latest commit: https://github.com/maplibre/maplibre-gl-js/blob/60ede4a524f3277d9115ccf1692c83c4b54a59b6/src/source/geojson_source.test.ts#L616

andrewda:
The failing unit test seems unrelated to my changes in this PR -- possibly a flake?

src/source/geojson_source.test.ts
HarelM: Ok, fair enough.

HarelM (approved):

0xFA11:
FWIW, after updating from 5.6.1 to 5.6.2, a bug started to happen in our addSource/addLayer code where myGeoJSON.features.push(...) code no longer works. I couldn't create a minimal repro yet but I just wanted to comment here ahead of time to warn that this PR is potentially introducing a bug. I will (or somebody from our team will) hopefully come up with a bug report very soon.

andrewda:
@0xFA11 Interesting, I can't assist much without seeing some example code or a reproduction. Please reply back when you have that available and I can take a deeper dive.

0xFA11:
actually, we bisected every individual commit one by one and turns out this is not the commit that broke it, it's #6131 — so, this is a false flag under this PR. I'll continue conversation over on that other PR instead. (cc @andrewda)

PR #6180, opened 2025-07-25
Bugfix/6179 set transform request types

### LINKED ISSUE #6179: [Typescript] `transformRequest` is wrongly typed when using strict mode
**maplibre-gl-js version**: 5.6.1

**browser**: NA

### Explanation

`Map.setTransformRequest` requires a `transformRequest` parameter of type `RequestTransformFunction`. When using strict mode, TypeScript doesn't allow to provide `null` or `undefined` as parameter (as it is not a `RequestTransformFunction`), which I think is the intended way to remove the `transformRequest` function.

Nullish values are already taken into consideration when calling the function in the `RequestManager` : 

https://github.com/maplibre/maplibre-gl-js/blob/93da1d36f23d0d95fcc42f696b5afbd4198c8734/src/util/request_manager.ts#L30-L36

`null` is also the default value of the map's `transformRequest` option : 

https://github.com/maplibre/maplibre-gl-js/blob/93da1d36f23d0d95fcc42f696b5afbd4198c8734/src/ui/map.ts#L430

### Steps to Trigger Behavior

 1. Set `"strict": true` in your `tsconfig.json`
 2. Create a `Map` instance
 3. Call map's `setTransformRequest` with `null` or `undefined` : `map.setTransformRequest(null)`
 4. TypeScript throws an error

### Link to Demonstration

https://codesandbox.io/p/sandbox/gsj7xx?file=%2Fsrc%2Findex.ts

### Expected Behavior

Calling `Map.setTransformRequest` with `null` or `undefined` should not throw a TypeScript error.

### Actual Behavior

Calling `Map.setTransformRequest` with `null` or `undefined` throws a TypeScript error.

### Workarounds

This problem has workarounds : 
- Disable TypeScript for the line at fault (with `@ts-ignore` or `@ts-expect-error`)
- Use a function that does nothing instead of `null` or `undefined` : `map.setTransformRequest(() => undefined)`

madoci:
## Summary

Fixes #6179 

This PR fixes the typing of the `Map.setTransformRequest` parameter to include `null` (which allows to remove the previous `transformRequest` function).

## Changes

- Adding `null` to `transformRequest` type in `Map.setTransformRequest` and in `RequestManager`.
- Setting `RequestManager._transformRequestFn` to `null` if the constructor parameter is `undefined`. This is not required, but it ensures two things : 
  - The type definition is strict, meaning it is valid even with TypeScript strict mode on.
  - The value is consitent : it's either a function or `null`, but not `undefined` sometimes and `null` some other times.
    > Note : the value can still be `undefined` with `setTransformRequest(undefined)`, but this would only be allowed if strict mode is off.
- Adding a test to check that `transformRequest` is removed when set to null.
- Other changes are automatic removal of trailing spaces.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/ui/map_tests/map_basic.test.ts
+++ b/src/ui/map_tests/map_basic.test.ts
@@ -69,6 +69,15 @@ describe('Map', () => {
             map.setTransformRequest(transformRequest);
             map.setTransformRequest(transformRequest);
         });
+
+        test('removes function when called with null', () => {
+            const map = createMap();
+
+            const transformRequest = (() => {}) as any as RequestTransformFunction;
+            map.setTransformRequest(transformRequest);
+            map.setTransformRequest(null);
+            expect(map._requestManager._transformRequestFn).not.toBe(transformRequest);
+        });
     });
 
     describe('is_Loaded', () => {
@@ -117,7 +126,7 @@ describe('Map', () => {
 
             expect(map.isStyleLoaded()).toBe(false);
             await map.once('load');
-            expect(map.isStyleLoaded()).toBe(true);  
+            expect(map.isStyleLoaded()).toBe(true);
         });
 
         test('Map.areTilesLoaded', async () => {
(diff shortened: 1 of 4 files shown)


src/ui/map_tests/map_basic.test.ts
```diff
@@ -69,6 +69,15 @@ describe('Map', () => {
             map.setTransformRequest(transformRequest);
             map.setTransformRequest(transformRequest);
         });
+
+        test('removes function when called with null', () => {
+            const map = createMap();
+
+            const transformRequest = (() => {}) as any as RequestTransformFunction;
```
HarelM: Use a spy and check that it hasn't been called after removing instead of using internal field.

HarelM:
Thanks! Looks good, I've added some minor comments.

PR #6190, opened 2025-07-28
improve GeolocateControl accuracy circle radius computation when movi…

geekdenz:
…ng map

Using the center position to calculate the pixel scale, and updating calculations at each move/rotate/zoom of the map, as those change the position of the accuracy circle.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/ui/control/geolocate_control.ts
+++ b/src/ui/control/geolocate_control.ts
@@ -303,6 +303,8 @@ export class GeolocateControl extends Evented implements IControl {
         }
 
         DOM.remove(this._container);
+        this._map.off('move', this._onMove);
+        this._map.off('rotate', this._onRotate);
         this._map.off('zoom', this._onZoom);
         this._map = undefined;
         numberOfWatches = 0;
@@ -374,6 +376,11 @@ export class GeolocateControl extends Evented implements IControl {
             this._finish();
 
             return;
+        }
+
+        if (this.options.showAccuracyCircle) {
+            // this is needed by _updateCircleRadius()
+            this._lastKnownPosition = position;
         }
 
         if (this.options.trackUserLocation) {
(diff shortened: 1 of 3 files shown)


.github/workflows/test-all.yml
```diff
@@ -51,6 +51,8 @@ jobs:
       - name: Start display server
         run: nohup Xvfb &
           echo "DISPLAY=:0" >> $GITHUB_ENV
+      - name: Install dependencies for turf
```
HarelM: Why is this needed?

package.json
```diff
@@ -56,6 +53,8 @@
     "@types/diff": "^8.0.0",
     "@types/earcut": "^3.0.0",
     "@types/eslint": "^9.6.1",
+    "@types/geojson": "^7946.0.16",
```
HarelM: That's not the right place for typescript users, it's deliberately in dependencies and not devdeoendencies...

src/ui/control/geolocate_control.ts
```diff
@@ -460,16 +467,30 @@ export class GeolocateControl extends Evented implements IControl {
...
+        const delta_x_pixels = 10; // arbitrary small offset, in x direction because there is no map tilting in the x direction
+        const delta = this._map.unproject([point.x + delta_x_pixels, point.y]);
+        const distance = delta.distanceTo(center);
+        const pixelScale = distance / delta_x_pixels;
+        const circleDiameter = 2 * (this._accuracy / pixelScale);
         this._circleElement.style.width = `${circleDiameter}px`;
         this._circleElement.style.height = `${circleDiameter}px`;
     }
 
+    _onMove = () => {
```
HarelM, line 482: These looks the same, can't we simply use the same method?

HarelM:
I've added some small comments, looks good otherwise. Can you make sure the added code is covered by the tests?
Codecov should've done it, but it's not working for some reason...

src/ui/control/geolocate_control.ts
geekdenz: Hm, this made me think and read the code again. It had been quite some months since I looked into it.
This seems to be the major difference here and if you are correct, @HarelM , which I believe you are then the changes are not needed.

I retested and tried to reproduce the issue. I did this on multiple checkouts of various commits and could not reproduce the issue using the http://localhost:9966/test/examples/locate-user.html example.

Before I commit more work into this, I would suggest we reproduce the issue reliably, otherwise it might be a non-issue.

What do you think?

HarelM: I'm not sure this particular issue was resolved, but feel free to try and reproduce this, and if it can't be reproduced then it means there's no need for a fix, right? :-)

geekdenz:
Cannot reproduce original issue.

HarelM:
I can reproduce the original issue using the following steps:
1. Go to https://maplibre.org/maplibre-gl-js/docs/examples/locate-the-user/
2. Start location tracking
3. Pitch the map to max
4. Zoom so that the circle is big
5. Move the map forward and backwards and see that the size of the circle is increasing and decreasing in the wrong direction.

Let me know if you are unable to reproduce this...

geekdenz:
@HarelM  or anyone reading:

I did not see it in this video I made. Could you make a video reproducing?

I believe that the issue is there but I just don't know how to reproduce it. Could you give me the location and accuracy maybe?

I used 

Lat: 52.520007
Lon: 13.404954
Accuracy: 150

You can do this under DevTools > Three vertical dots ... > More Tools > Sensors

https://youtu.be/5JcEXGjk6Mg

HarelM:
See the following, it happens when panning the map up and down - the circle's size is changing "wrong".


https://github.com/user-attachments/assets/2df650c1-37e6-4f3f-b0d1-a8029f5a1495

geekdenz:
Hi @HarelM 

This is still on my to-do list. I won't forget. 

Could you please let me know the steps to reproduce? I see it in the video but it would be good if I can reproduce it reliably to know the fix works. 

Thanks!

HarelM:
Same as before, 
1. go to the example page
2. Activate geolocation
3. Zoom in so that the circle is big
4. Pitch to max 
5. Pan the map up and down.

geekdenz:
Thanks for your patience, @HarelM ! On my Desktop I could reproduce it just now.

I am hoping to make some time between now and tomorrow to look into this again.

geekdenz:
@HarelM 
I have a little understanding of the problem now:

1. On pitch and pan the circle diameter does not update.
2. The calculation for the diameter is incorrect for when there is a bearing.

So that is why it jumps on zoom and now that I have locally ensured that the `_updateCircleRadius` method is called, the diameter changes smoothly, which looks a little better but is not quite correct.

We need to find the diameter, in pixels considering the view geometries. With the approach using the Turf library works because that is dealt with correctly in the related code for dealing with GeoJSON.

I am getting close to solving it without increasing the bundle size, but I thought I'd let you know my findings so far.

EDIT: Made some significant progress. Should be done soon. Just need to ensure there are unit tests and everything works.

PR #5613, opened 2025-03-11
global-state expression support in filter and paint properties

zbigniewmatysek-tomtom:
This PR adds a new map methods: `setGlobalStateProperty()` and `getGlobalState()` leveraging the new `global-state` expression.

Out of the scope: `global-expression` in layout properties (`setGlobalStateProperty()` will have no effect on those).

See the [style spec update](https://github.com/maplibre/maplibre-style-spec/pull/1044) (needs to be merged first) and related issues: 
https://github.com/maplibre/maplibre-style-spec/issues/886
https://github.com/maplibre/maplibre-gl-js/issues/4964

See examples to see how it works in practice:
* test/examples/filter-symbols-using-global-state.html
* test/examples/global-state.html

## Launch Checklist

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.
 - [x] Link to related issues.
 - [ ] Include before/after visuals or gifs if this PR includes visual changes.
 - [x] Write tests for all new functionality.
 - [x] Document any changes to public APIs.
 - [ ] Post benchmark scores.
 - [x] Add an entry to `CHANGELOG.md` under the `## main` section.

PR #6215, opened 2025-08-03
Fix global state for paint properties

### LINKED ISSUE #6048: `global-state` subexpression is always null in feature-dependent expression
When a `global-state` subexpression is used in a feature-dependent expression (i.e. the overall expression also contains a subexpression like `feature-state`, `has`, `get`, `within`, `distance`, or `id`), the `global-state` subexpression seems to always be treated as null.

I initially thought this was a style spec issue, but I was unable to replicate this behaviour using the style spec evaluator.
I have tried several tests like [this](https://github.com/maplibre/maplibre-style-spec/compare/main...sruenwg:maplibre-style-spec:global-state-with-feature-state-bug), all of which returned the expected result, hence I am thinking it may be a gl-js issue instead.

---

**browser**: Chrome

### Steps to Trigger Behavior

 1. Create map with style where a layer has a paint property containing both `global-state` and `feature-state` subexpressions.

### Link to Demonstration

https://jsbin.com/jitomapiho/edit?html,output

Press 'g' in the demo to toggle `globalState.theme` between 'light' and 'dark';
press 'f' in the demo to toggle feature 0's `featureState.highlighted` between true and false.

Comment out these lines in `circle-color` to observe that `global-state` works as expected when `feature-state` is not present in the same expression:
```
// ['boolean', ['feature-state', 'highlighted'], false],
// 'cyan',
```

Also see https://github.com/maplibre/maplibre-gl-js/commit/e10f8eb51162097f7a209db3a2bac4a6b4ac5dc8 for a render test which shows this unexpected behaviour.

### Expected Behavior

(Ignoring `featureState.highlighted`) the circles should be magenta when `globalState.theme` is set to 'dark'.
In other words, the `global-state` subexpression should always fetch and return the global state value.

### Actual Behavior

The circles remain black regardless of `globalState.theme`.
In other words, the `global-state` subexpression is behaving differently (returning null) due to the presence of a `feature-state` subexpression in the same overall expression.

### Notes

I was also able to observe this behaviour when:
- using `coalesce` or `+` instead of `case` for the overall expression, and when
- using `has` or `get` instead of `feature-state`,

so my guess would be that this is caused either by how global state is handled, or by how a feature's state/properties are handled.

melitele:
- Fixes #6048
## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


sruenwg:
Hey, thanks for looking into this.

One question I have with this approach overall:
It looks like the meat of this fix lies in the `Style._setPaintProperty` called by both `setGlobalState` and `setGlobalStateProperty`.
However, in https://github.com/maplibre/maplibre-gl-js/issues/6048#issuecomment-3123977447 I noticed that feature state updates cause expression re-evaluation to occur via `AttributeBinder.updatePaintArray`. I believe `Style._setPaintProperty` does not trigger a call to this method, so I'm wondering how this solution handles expression evaluation when global state and feature state updates are made in various orders.

For instance, I locally tried adding a `removeFeatureState` after the `setFeatureState` in the `global-state/with-feature-state` test:
```diff
"operations": [
  [
    "setGlobalStateProperty",
    "color",
    "cyan"
  ],
  [
    "setFeatureState",
    {
      "source": "geojson",
      "id": 0
    },
    {
      "color": "magenta"
    }
  ],
+  [
+    "removeFeatureState",
+    {
+      "source": "geojson",
+      "id": 0
+    }
+  ],
  [
    "wait"
  ]
]
```
My understanding is that both circles should be cyan since `globalState.color === 'cyan'` and the circle-color property is set to use `globalState.color` when `featureState.color` is unavailable:
```json
"circle-color": [
  "coalesce",
  [
    "feature-state",
    "color"
  ],
  [
    "global-state",
    "color"
  ],
  "yellow"
]
```
However, I am seeing that feature 0 is yellow in the result, presumably due to `AttributeBinder.updatePaintArray` evaluating the expression on `removeFeatureState` without any knowledge of the global state:
image

I'm not sure how the proposed solution can be adapted to handle this issue, since I assume we don't want to rework how feature state updates are performed, meaning feature state updates will continue to perform expression re-evaluation in `AttributeBinder.updatePaintArray`.
My best guess at a solution would thus still be to somehow pass the global state to `AttribuetBinder.updatePaintArray` (see https://github.com/maplibre/maplibre-gl-js/issues/6048#issuecomment-3123977447), though I'm no expert in MLGL so perhaps there are other viable solutions…

src/data/program_configuration.ts
```diff
@@ -73,7 +73,8 @@ interface AttributeBinder {
         feature: Feature,
         imagePositions: {[_: string]: ImagePosition},
         canonical?: CanonicalTileID,
-        formattedSection?: FormattedSection
+        formattedSection?: FormattedSection,
+        options?: {globalState?: Record<string, any>}
```
HarelM: This is a bit too generic. either only add "globalState" or change the entire method to receive an options object instead of this long list of parameters (I prefer the second option if you'd ask me).

src/style/style.ts
```diff
@@ -1297,13 +1303,17 @@ export class Style extends Evented {
 
         if (deepEqual(layer.getPaintProperty(name), value)) return;
 
+        this._setPaintProperty(layer, name, value, options);
+    }
+
+    _setPaintProperty(layer: StyleLayer, name: string, value: any, options: StyleSetterOptions = {}) {
```
HarelM: Would be great if this had a meaningful name instead of the same name with "_", but I understand if this is the best name you could think of.

src/style/style_layer.ts
```diff
@@ -21,7 +21,7 @@ import type {CrossfadeParameters} from './evaluation_parameters';
 
 import type {IReadonlyTransform} from '../geo/transform_interface';
 import type {CustomLayerInterface} from './style_layer/custom_style_layer';
-import type {Map} from '../ui/map';
+import type {Map as MapClass} from '../ui/map';
```
HarelM: These caused issues in the past, so I think it would be better to remove this change and simply use `global.Map<...>`.

src/style/style_layer.ts
```diff
@@ -193,6 +193,29 @@ export abstract class StyleLayer extends Evented {
         return globalStateRefs;
     }
 
+    /**
+     * Get list of global state references that are used within paint properties.
+     * This is used to determine if layer needs to be repainted when global state property changes.
+     *
+     */
+    getPaintAffectingGlobalStateRefs() {
```
HarelM: Add explicit return value, just to make sure this doesn't get changed accidentally in the future...

HarelM:
This looks very good, thanks!
Any change you could add a few unit tests around the setGlobalState with this as there are only render tests, and I'd like some of the functionality change here to be covered with unit tests, especially the render test with map invocation methods which I'm not sure are fully supported by native.
@louwers please note that some render tests are added here for global-state which I think is not supported in native, so those may need to be excluded from the parity check.

melitele:
@sruenwg addressed passing global state to `AttribuetBinder.updatePaintArray` if you like to try it out again
@HarelM made changes that you suggested plus added some unit tests

src/data/bucket/circle_bucket.ts
```diff
@@ -144,7 +144,10 @@ export class CircleBucket<Layer extends CircleStyleLayer | HeatmapStyleLayer> im
 
     update(states: FeatureStates, vtLayer: VectorTileLayer, imagePositions: {[_: string]: ImagePosition}) {
         if (!this.stateDependentLayers.length) return;
-        this.programConfigurations.updatePaintArrays(states, vtLayer, this.stateDependentLayers, imagePositions);
+        this.programConfigurations.updatePaintArrays(states, vtLayer, this.stateDependentLayers, {
+            imagePositions,
+            globalState: this.globalState
```
sruenwg, line 149: Thanks for making this change.

It looks like `Bucket.globalState` is never reassigned; does this still produce the expected value when a feature state update follows multiple global state updates?
(~~I suspect it may work fine since global state updates trigger full source reloads, which in turn trigger bucket re-instantiation, but I am not sure.~~ Forgot that paint properties don't trigger a source reload)

src/data/program_configuration.ts
```diff
@@ -71,16 +71,22 @@ interface AttributeBinder {
...
         feature: Feature,
-        imagePositions: {[_: string]: ImagePosition},
-        canonical?: CanonicalTileID,
-        formattedSection?: FormattedSection
+        options: {
+            imagePositions: {[_: string]: ImagePosition};
+            canonical?: CanonicalTileID;
+            formattedSection?: FormattedSection;
+            globalState?: Record<string, any>;
+        }
```
sruenwg: Are any of these fields aside from `formattedSection` actually optional? (same for the `updatePaintArray` options)
I haven't taken a super close look, but it at least seems `globalState` is always passed.

src/style/style.ts
```diff
@@ -369,11 +369,17 @@ export class Style extends Evented {
...
             const layoutAffectingGlobalStateRefs = layer.getLayoutAffectingGlobalStateRefs();
+            const paintAffectingGlobalStateRefs = layer.getPaintAffectingGlobalStateRefs();
 
             for (const ref of globalStateRefs) {
                 if (layoutAffectingGlobalStateRefs.has(ref)) {
                     sourceIdsToReload.add(layer.source);
                 }
+                if (paintAffectingGlobalStateRefs.has(ref)) {
+                    for (const {name, value} of paintAffectingGlobalStateRefs.get(ref)) {
+                        this._updatePaintProperty(layer, name, value);
```
sruenwg, line 369: This might be a nitpick, but I would prefer if this call could moved elsewhere since I would not expect a method named `_find*` to cause a destructive change via `_updatePaintProperty`.
Would it be possible to refactor this so the `_updatePaintProperty` call is performed outside of `_findGlobalStateAffectedSources`?

sruenwg:
Thanks for the change; I've left some additional questions & comments but it's looking great overall 👍 
I'll check this out locally some time today/tomorrow to try to answer my own questions.

src/data/bucket/circle_bucket.ts
melitele: `Bucket.globalState` is a reference to the global state object.

sruenwg: My understanding is that buckets are instantiated on the worker thread and passed to the main thread to handle subsequent operations like `updatePaintArray`, so I was concerned the serialize/deserialize step when transferring the global state data may cause the bucket on the main thread to potentially have a stale global state, if that makes sense.
Please correct me if I'm wrong…

src/data/program_configuration.ts
```diff
@@ -185,15 +191,15 @@ class SourceExpressionBinder implements AttributeBinder {
         this.paintVertexArray = new PaintVertexArray();
     }
 
-    populatePaintArray(newLength: number, feature: Feature, imagePositions: {[_: string]: ImagePosition}, canonical?: CanonicalTileID, formattedSection?: FormattedSection) {
+    populatePaintArray(newLength: number, feature: Feature, options: {imagePositions: {[_: string]: ImagePosition}; canonical?: CanonicalTileID; formattedSection?: FormattedSection; globalState?: Record<string, any>}) {
```
HarelM: Extract the type maybe?

sruenwg:
Tried some tests locally and AFAICT the visual results look good.
I'm however seeing that when the expression is feature-dependent, every global state update triggers a source reload. This doesn't seem ideal to me, but is this intentional/necessary?

- In `Style._updatePaintProperty`, `requiresRelayout` is always true since the expression is data-driven:
https://github.com/melitele/maplibre-gl-js/blob/93ddb1a5b88ef8738d12833a71200f7ce4934623/src/style/style_layer.ts#L269

- This means `Style._updateLayer` is called, which sets `Style._updatedSources[layer.source] = 'reload'`:
https://github.com/melitele/maplibre-gl-js/blob/93ddb1a5b88ef8738d12833a71200f7ce4934623/src/style/style.ts#L1441

- This triggers a source reload in the next `Style.update` call:
https://github.com/melitele/maplibre-gl-js/blob/93ddb1a5b88ef8738d12833a71200f7ce4934623/src/style/style.ts#L686-L690

src/data/bucket/circle_bucket.ts
sruenwg: Tested this a bit… It seems like the source reload (described here https://github.com/maplibre/maplibre-gl-js/pull/6215#issuecomment-3153923882) causes all relevant buckets to be re-instantiated in the worker on each global state update, hence the correct visual results are produced regardless of whether or not the old bucket's `globalState` property is updated.
This is thus not currently an issue in itself, but it is something to be aware of when figuring out a solution to https://github.com/maplibre/maplibre-gl-js/pull/6215#issuecomment-3153923882.

melitele:
> I'm however seeing that when the expression is feature-dependent, every global state update triggers a source reload. This doesn't seem ideal to me, but is this intentional/necessary?

For context, my assumption for global state property changes is that it happens in response to user action, i.e. rarely (light vs. dark theme, terrain vs. regular map, user's vs. locality language, etc.)

That said, I'm all for a less resource-consuming way of triggering recalculation of paint properties, if it's implemented already. Suggestions?

sruenwg:
> my assumption for global state property changes is that it happens in response to user action, i.e. rarely (light vs. dark theme, terrain vs. regular map, user's vs. locality language, etc.)

I have the same understanding.
Following the spirit of [this similar discussion](https://github.com/maplibre/maplibre-gl-js/pull/5613#discussion_r2063447652), if there is no reasonable alternative then I think I would find this current solution just fine.

> I'm all for a less resource-consuming way of triggering recalculation of paint properties, if it's implemented already

The only alternative I have in mind is to adopt how feature state updates trigger re-evaluation and repaint (I touched on this here https://github.com/maplibre/maplibre-gl-js/issues/6048#issuecomment-3123977447), though perhaps you've already considered this before raising this PR.

melitele:
> The only alternative I have in mind is to adopt how feature state updates trigger re-evaluation and repaint

Can you implement a function that does that which I can call instead of `_updatePaintProperty`?

sruenwg:
It would be considerably more complicated than a single function, but sure I could give it a shot

HarelM:
I'm a bit lost in terms of the status of this PR.
Do let me know when it's ready for review and I'll be happy to review it.
Make sure all the tests pass first :-)

sruenwg:
@HarelM FYI the back-and-forth above was mostly me requesting a change in how this fix triggers repaints on global state updates, since I felt this current design performs some unnecessary source reloads.
If you (or some contributor who is knowledgeable in this part of the codebase) do not find this to be a concern, then I wouldn't mind leaving this optimization for a future PR.

HarelM:
I think your comment are valuable!
I was only asking to know when to step in basically... :-)

melitele:
@HarelM ready for review.

src/style/global_state.ts
HarelM, line 1: Was this file added to prevent circular dependecies? Otherwise I would have consider moving it to the class that exports methods with these types...

src/style/global_state.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * List of global state references that are used within layout or filter properties.
+ * Collects the names of global state properties that affect layout or filter expressions.
+ * This is used to determine if layer source needs to be reloaded when global state property changes.
+ */
+export class LayoutAffectingGlobalRefs extends Set<string> {}
```
HarelM: Do we really need to export a type of `Set<string>`? Seems trivial enough to just write it where needed...?

src/style/global_state.ts
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * List of global state references that are used within paint properties.
+ * Maps the name of the global state property to an array of property name / value pairs.
+ * This is used to determine if layer needs to be repainted when global state property changes.
+ */
+export class PaintAffectingGlobalStateRefs extends Map<string, Array<{name: string; value: any}>> {}
```
HarelM: Looking at where this is used I find it more confusing than helping, I would recommend removing this file completely and use the set with full type where needed.

HarelM:
I've added some minor comment, sorry for the delay in the review.
Looks good otherwise. Thanks!

src/style/global_state.ts
melitele: All comes down to conflict between standard JS Map and MapLibre Map. Those two can't be used in the same file without some remedy. The options are:

- separate files (this PR)
- `global` prefix for JS Map (so far used solely in tests, will it even work in browser?)
- import MapLibre Map as something else in this file

sruenwg: The existing pattern for similar situations seems to be to use a plain indexed object, but is that not possible here? e.g. `{ [globalStatePropertyKey: string]: { name: string, value: any }[] }` (or `Record` which I also see in some places)
Following the pattern of `FeatureStates`/`LayerFeatureStates` in `source_state.ts` might make sense to me.

I believe defining a class for an object with no custom data or behaviour deviates from existing patterns, so I would definitely prefer a type alias here.
For `LayoutAffectingGlobalRefs`, I agree `Set<string>` is simple enough that it doesn't even need an alias, but I also would not mind having a type alias if you prefer having alignment with `PaintAffectingGlobalRefs`.

src/style/global_state.ts
melitele: This file was added to avoid prefixing `Map` with `global`.

src/style/global_state.ts
melitele: Seems consistent to keep in one place all types for collections of global references affecting layer properties.

src/style/global_state.ts
melitele: @HarelM You made suggestions to keep reference to JS `Map` in the same file prefixing it with `global`, so I know your preferred option. I  only see `global` in tests so far. I would like your confirmation that it will work and that we are OK setting this precedent.

HarelM: I think you can easily test it, I don't see a reason it won't work...
I'm ok setting this precedence.

sruenwg:
I'd like to give an update on this:

> The only alternative I have in mind is to adopt how feature state updates trigger re-evaluation and repaint

I tried implementing what I had in mind here: https://github.com/melitele/maplibre-gl-js/compare/fix-global-state-for-paint-properties...sruenwg:maplibre-gl-js:avoid-source-reload-global-state-paint

However, something I didn't realize earlier is that feature state is currently reapplied to each tile when retrieving the tile from the tile cache. Without this line, when a cached tile is retrieved and rendered, its state may be stale if the state changed while the tile was in the cache:
https://github.com/maplibre/maplibre-gl-js/blob/2ab8b2eab0c286fae530789827bbb30d42c8b87d/src/source/source_cache.ts#L857-L862

This design may make sense for feature state, where it is not commonly expected for many features to have a feature state set; if following this pattern for global state though, it seems that **on every tile load and reload from cache, the paint properties for all features in that tile with a global state reference would have to be re-evaluated**.

My understanding is that tiles are loaded from the cache on very common operations such as panning and zooming. Rather than hurting the performance of these common operations in any way, I believe @melitele's solution of simply reloading the source on `setGlobalState` is more preferable as a user.

src/style/style.test.ts
```diff
@@ -1420,6 +1419,35 @@ describe('Style.setGlobalState', () => {
         expect(style.sourceCaches['line-source-id'].reload).toHaveBeenCalled();
     });
 
+    test('reloads sources when a new state property is used in a paint property that affects layout', async () => {
```
sruenwg, line 1422: Could you please make this more accurate? This test doesn't seem to involve layout.
(Same for the corresponding `setGlobalStateProperty` test)

src/style/style_layer.test.ts
```diff
@@ -281,7 +281,51 @@ describe('StyleLayer.getLayoutAffectingGlobalStateRefs', () => {
...
+        expect(layer.getPaintAffectingGlobalStateRefs()).toEqual(new Map<string, Array<{name: string; value: any}>>());
+    });
+
+    test('returns global-state references from paint properties', () => {
+        const layer = createStyleLayer({
+            'id': 'symbol',
+            'type': 'symbol',
+            source: 'source',
+            'paint': {
+                //@ts-ignore
```
HarelM: Is there a way to solve this without ts-ignore? Maybe by fixing the types in the style spec repo?

HarelM:
I've added a minor comment that can be addressed in a different PR of needed. 
Otherwise looks good. 
Can you check in codecov or locally that all the added/changed lines of code are covered by the tests?

src/style/style.test.ts
melitele: The point of this test is to check that a change to a paint property that requires re-layout results in reloading source. It does that.

sruenwg: I see, would it be possible to make that intent more transparent/explicit?
I got confused by the 'paint property that affects layout' since I interpreted this as global state usage in a paint property actually altering the layout in some way, and I could see future devs making the same misunderstanding.

IMO it would be nice if the condition for the source reload were stated in terms of the test input, so something like 'a feature-dependent paint property which references the new state property'.

src/style/style_layer.test.ts
melitele: not needed anymore, the necessary change in the style spec repo is already in

melitele:
> Can you check in codecov or locally that all the added/changed lines of code are covered by the tests?

does now

HarelM (approved):

HarelM:
Looks great, thanks!!

sruenwg:
Thank you both 👍

HarelM:
I'm pretty sure this wasn't introduced by you, but I'm trying to remove all console log writing from the unit-tests, and some of the tests that are related to global-state are adding some console logging.
I'm not sure why, but it appears to be different than regular console logging done with the evented class.
@sruenwg @melitele any chance you can check this?

```
stderr | src/style/style.test.ts > Style.setGlobalState > does not reload sources when new state property is used in paint property
Could not parse color from value 'null'

stderr | src/style/style.test.ts > Style.setGlobalState > does not reload sources when a new state property is used in a paint property while state property used in filter is unchanged
Could not parse color from value 'null'

stderr | src/style/style.test.ts > Style.setGlobalStateProperty > does not reload sources when state property is only used in paint properties
Could not parse color from value 'null'

stderr | src/style/style.test.ts > Style.setGlobalStateProperty > does not reload sources when state property is used in paint property while a different state property used in filter is unchanged
Could not parse color from value 'null'
```

melitele:
~~I'll fix it.~~ PR https://github.com/maplibre/maplibre-gl-js/pull/6272

lhapaipai:
Hi and thank you for this PR.
Unfortunately, it seems there are still issues when global-state is used together with get, for example in the following configuration:

```ts
// feature
{
  type: "Feature",
  geometry: {
    type: "Polygon",
    coordinates: [
      [
        [2, 0],
        [2.5, 0],
        [2.5, -0.5],
        [2, -0.5],
        [2, 0],
      ],
    ],
  },
  properties: { name: "rectangle-4" },
},

// pseudo code used
const fallbackFillColor = "#999999";
map.setGlobalStateProperty("globalFillColor", "#00ff00");

// layer specification
{
  id: "rectangle-4",
  type: "fill",
  source: "geojson-source",
  filter: ["==", ["get", "name"], "rectangle-4"],
  paint: {
    "fill-pattern": [
      "concat",
      "fill_color_",
      [
        "coalesce",
        ["get", "fillColor"], // not set
        ["global-state", "globalFillColor"], // ERROR not used
        fallbackFillColor, // fallback is used
      ],
    ],
  },
}
```
I've prepared a reproduction example to illustrate this issue.

https://jsbin.com/waxujediku/edit?html,console,output

I’m not sure if my issue is related to this PR, feel free to let me know if I should open a separate issue!!

melitele:
Good catch. Global state is not propagated to calculation of pattern properties. Draft PR #6301

PR #6223, opened 2025-08-05
Pass language to TinySDF to render Simplified and Traditional Chinese characters

mapmeld:
When a MapLibre layer was added to the OSM homepage, I noticed an issue in MapLibre's examples with rendering Simplified or Traditional Chinese variants:
Two labels should have <span lang="zh">门</span> and <span lang="zh-Hant">北</span>:
taipeimap

**Background**: a Unicode codepoint such as 化 may appear differently, such as <span lang="zh-Hans">化</span>, <span lang="zh-Hant">化</span>, or <span lang="ja">化</span> depending on the `lang` tag on a page or element.
**In Chrome** the issue appears to be that the page language isn't applied to TinySDF [until we set ctx.lang](https://github.com/mapbox/tiny-sdf/pull/58) directly on its canvas element.

**Current state of MapLibre**: there is a `localIdeographFontFamily` option with `sans-serif` as default. Users can change it to a single-language font that they've bundled with the page, but fonts often support multiple variants, and it's going to remain an issue for OSM and other MapLibre users.
**This PR** would upgrade TinySDF, and let users set the `language` which is passed to TinySDF to affect which versions of characters that they see. I also updated the rich text demo to `zh-HK`.

### Alternatives / Notes
- we could redesign the `localIdeographFontFamily` option (which is already a little confusing, with false / null values)
- since `document.documentElement.lang` is the expected behavior, we could use that as the default, or avoid adding a user-settable option
- Firefox's behavior remains confusing to me. It looks good on the OSM homepage, but not this demo?
- if it seems like it's on the right track, then I can look into tests, documentation, and changelog

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 - Fix geolocate control lock loss on window resize ([#3504](https://github.com/maplibre/maplibre-gl-js/issues/3504))
 - Fix a memory leak in `GeoJSONSource` when rapidly updating data ([#6163](https://github.com/maplibre/maplibre-gl-js/pull/6163))
 - Fix `Map.setTransformRequest` parameter type to include `null` ([#6179](https://github.com/maplibre/maplibre-gl-js/issues/6179))
+- Pass document's `lang` to Tiny-SDF to render Simplified and Traditional Chinese characters ([#6223](https://github.com/maplibre/maplibre-gl-js/issues/6223))
 
 - _...Add new stuff here..._
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,19 @@
 ## main
-- Fix typo `_rotatePitchHandler` in the `navigation_control.ts` file ([#6207](https://github.com/maplibre/maplibre-gl-js/issues/6207))
 
 ### ✨ Features and improvements
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
+- _...Add new stuff here..._
+
+## 5.6.2
+
+### 🐞 Bug fixes
 - Fix white artifacts when using non-zero elevation ([#6032](https://github.com/maplibre/maplibre-gl-js/pull/6032))
-- Fix geolocate control lock loss on window resize ([#3504](https://github.com/maplibre/maplibre-gl-js/issues/3504))
+- Fix geolocate control lock loss on window resize and zoom ([#3504](https://github.com/maplibre/maplibre-gl-js/issues/3504))
 - Fix a memory leak in `GeoJSONSource` when rapidly updating data ([#6163](https://github.com/maplibre/maplibre-gl-js/pull/6163))
 - Fix `Map.setTransformRequest` parameter type to include `null` ([#6179](https://github.com/maplibre/maplibre-gl-js/issues/6179))
+- Fix typo `_rotatePitchHandler` in the `navigation_control.ts` file ([#6207](https://github.com/maplibre/maplibre-gl-js/issues/6207))
 - Pass document's `lang` to Tiny-SDF to render Simplified and Traditional Chinese characters ([#6223](https://github.com/maplibre/maplibre-gl-js/issues/6223))
 
 - _...Add new stuff here..._
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,11 @@
 ## main
 
 ### ✨ Features and improvements
+- Pass document's `lang` to Tiny-SDF to render Simplified and Traditional Chinese characters ([#6223](https://github.com/maplibre/maplibre-gl-js/issues/6223))
+
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
-- Pass document's `lang` to Tiny-SDF to render Simplified and Traditional Chinese characters ([#6223](https://github.com/maplibre/maplibre-gl-js/issues/6223))
-
 - _...Add new stuff here..._
 
 ## 5.6.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
-- _...Add new stuff here..._
+- Prevent original input style JSON from being mutated by `Style.set*` methods ([#6216](https://github.com/maplibre/maplibre-gl-js/pull/6216))
 
 ## 5.6.2
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
+- Pass document's `lang` to Tiny-SDF to render Simplified and Traditional Chinese characters ([#6223](https://github.com/maplibre/maplibre-gl-js/issues/6223))
+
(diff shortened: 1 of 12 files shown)


HarelM:
Can we relay on `document.documentElement.lang` to begin with instead of adding another parameter at this point in time?
CC: @1ec5
We can always add a parameter later on, right?

mapmeld:
@HarelM that makes sense. With this smaller change I don't see an effect on the tests or docs. Added a line to the CHANGELOG

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
+- Pass document's `lang` to Tiny-SDF to render Simplified and Traditional Chinese characters ([#6223](https://github.com/maplibre/maplibre-gl-js/issues/6223))
```
HarelM, line 4: Thanks! can you pass that to features and improvements section?

src/render/glyph_manager.ts
```diff
@@ -169,6 +169,7 @@ export class GlyphManager {
                 buffer: 3 * textureScale,
                 radius: 8 * textureScale,
                 cutoff: 0.25,
+                lang: typeof document !== 'undefined' ? document.documentElement?.lang ?? '' : '',
```
HarelM: Should this be `undefined` instead?

mapmeld: That makes sense. I just pushed a commit which should send `undefined`

HarelM:
How about adding a unit test to cover this addition? Would it make sense?

src/render/glyph_manager.ts
```diff
@@ -169,6 +169,7 @@ export class GlyphManager {
                 buffer: 3 * textureScale,
                 radius: 8 * textureScale,
                 cutoff: 0.25,
+                lang: typeof document !== 'undefined' ? document.documentElement?.lang : undefined,
```
1ec5: > Can we relay on `document.documentElement.lang` to begin with instead of adding another parameter at this point in time?

This is OK, though ideally, it would be scoped to the `maplibregl-canvas-container` or `maplibregl-canvas`. That would allow, for instance, [an interactive comparison](https://github.com/maplibre/maplibre-gl-compare/) between different CJK locales. But we’d need to make sure it would inherit from the body’s `lang` reliably in each browser.

1ec5:
> This PR would upgrade TinySDF

By the way, #6224 upgraded to TinySDF v2.0.7, which contains mapbox/tiny-sdf#58, so this PR doesn’t need to upgrade TinySDF after all.

mapmeld:
I've added a unit test
Thanks @1ec5 , with rebase there are no changes to dependencies / package.json

Probably the best way to compare two maps with this setup would be to set `localIdeographFontFamily` to two different language-specific fonts. To pass `lang` into the GlyphManager, it would need the style to pass the map canvas or its lang attribute.

1ec5:
> Probably the best way to compare two maps with this setup would be to set `localIdeographFontFamily` to two different language-specific fonts. To pass `lang` into the GlyphManager, it would need the style to pass the map canvas or its lang attribute.

I think it would still be a good idea to have `GlyphManager` accept the language as a parameter. This would give us more control over its environment. You wouldn’t need to set and unset `document.documentElement.lang` around every unit test involving `GlyphManager`.

This doesn’t necessarily mean we’d need to add a language parameter to other classes. When `Style` initializes the glyph manager, I think we can pass in `options.container.lang` or something like that:

https://github.com/maplibre/maplibre-gl-js/blob/7887f2c899dcc7f7bfa8a05f5a98c92bf1a5bf5a/src/style/style.ts#L252

HarelM:
That a good idea. Thanks @1ec5 for looking into this.

src/render/glyph_manager.test.ts
```diff
@@ -23,8 +23,8 @@ describe('GlyphManager', () => {
         });
     };
 
-    const createGlyphManager = (font?) => {
-        const manager = new GlyphManager(identityTransform, font);
+    const createGlyphManager = (font?, language?) => {
```
HarelM: Can you add types here while your are at it? Thanks!

mapmeld:
Newest commit:
- sends the `map._container?.lang` to Style to take precedence over document language
- expands the new unit test
- updates where `lang="zh-HK"` is set in the `test/examples`

src/render/glyph_manager.test.ts
```diff
@@ -151,4 +152,22 @@ describe('GlyphManager', () => {
         await manager.getGlyphs({'Arial Unicode MS': [0x30c6]});
         expect(drawSpy).toHaveBeenCalledTimes(1);
     });
+
+    test('GlyphManager passes container language to TinySDF', async () => {
+        const langSpy = GlyphManager.TinySDF = vi.fn().mockImplementation(() => new TinySDF({}));
```
HarelM: If the idea is to use the real implantation here I would remove the mocking part of the test.

src/render/glyph_manager.ts
```diff
@@ -169,6 +171,7 @@ export class GlyphManager {
                 buffer: 3 * textureScale,
                 radius: 8 * textureScale,
                 cutoff: 0.25,
+                lang: this.lang || (typeof document !== 'undefined' ? document.documentElement?.lang : undefined),
```
HarelM: If lang is injected I would expect this logic to be outside this class. i.e. use fallback document when initializing glyphmanager in the style class.

test/examples/display-and-style-rich-text-labels.html
```diff
@@ -13,7 +13,7 @@
     </style>
 </head>
 <body>
-<div id="map"></div>
+<div id="map" lang="zh-HK"></div>
```
HarelM, line 17: I think it would be good to see that seting it on a body level will work too, or maybe just add a comment about it.

src/render/glyph_manager.test.ts
```diff
@@ -151,4 +152,22 @@ describe('GlyphManager', () => {
...
+        const manager = createGlyphManager('sans-serif');
+        await manager.getGlyphs({'Arial Unicode MS': [0x30c6]});
+        expect(langSpy).toHaveBeenCalledWith(expect.objectContaining({fontFamily: 'sans-serif', lang: ''}));
+
+        // read document language
```
HarelM: Judging from the comments and the expect calls this probably needs to be split into several tests instead of one.

src/render/glyph_manager.test.ts
mapmeld: I'll work on the other comments. In this section I only want to spy on the constructor of TinySDF, but after creating it the same function calls `tinySDF.draw(str)`. I was thinking to let TinySDF create a glyph rather than mocking a function to return a fake glyph. I could pass the real options if that works

HarelM: If you can avoid mocking tinySDF and it will work without a browser, then it's better, otherwise mock away.

mapmeld: ok I think I found a good solution spying and returning a glyph from earlier tests

HarelM (approved):
LGTM, thanks!

PR #6240, opened 2025-08-12
Fix accuracy circle behavior in GeolocateControl (#5432)

geekdenz:
Add precise projection-aware sampling and handle style updates smoothly.

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


geekdenz:
Hi @HarelM 

This is another approach of fixing the problem not using TurfJS. Note that it is still in Draft, but you can do a quick test locally if you like.

src/ui/control/geolocate_control.ts
```diff
@@ -303,7 +305,11 @@ export class GeolocateControl extends Evented implements IControl {
         }
 
         DOM.remove(this._container);
-        this._map.off('zoom', this._onZoom);
+        this._map.off('zoom', this._onUpdate);
+        this._map.off('move', this._onUpdate);
+        this._map.off('pitch', this._onUpdate);
+        this._map.off('rotate', this._onUpdate);
+        this._map.off('render', this._onUpdate);
```
geekdenz: @HarelM Not sure we need all of these. But it seems like the safest option.

src/util/geolocation_support.ts
```diff
@@ -1,5 +1,9 @@
-let supportsGeolocation;
```
geekdenz: All this might be better to do in a different abstraction. Maybe we should have a Circle object that can be added to the map, similar to how GeoJSON is added to the map with TurfJS. However, I only thought of this now. If we do this later, this could be an overall improvement and circles could be rendered generically also for other purpouses than the accuracy.

However, since this is an improvement, I am leaving it as is for now.

geekdenz:
- [x]  Post benchmark scores.

[MapLibre GL JS Benchmarks.pdf](https://github.com/user-attachments/files/21748332/FireShot.Capture.001.-.MapLibre.GL.JS.Benchmarks.-.pdf)

geekdenz:
Ran the unit tests locally with no issues:

image

```bash
Test Files  175 passed (175)
      Tests  2395 passed (2395)
   Start at  18:46:35
   Duration  17.95s (transform 4.14s, setup 66.27s, collect 57.01s, tests 27.26s, environment 73.78s, prepare 16.33s)
```

src/util/geolocation_support.ts
HarelM: You can add a circle object to the map using a circle layer, it also can be tilted with the map.
The only problem with it is that it supports pixels size and not size in meters, but if you calculate it, you can update the size of the circle. It might be a simpler solution without all the extra code that was added and without a geojson source...

geekdenz: That could be an option. However, it works with the above and it is going to be some work to change it now. Should we add an issue and consider merging this for now?

EDIT:  I tried a circle layer with

```javascript
this._map.addLayer({
    id: layerId,
    type: 'circle',
    source: sourceId,
    paint: {
        'circle-radius': 60,
        'circle-color': '#ff0000', // red
        'circle-stroke-color': '#ffffff',
        'circle-stroke-width': 1.5,
        'circle-opacity': 0.9
    }
});
```

but that does not flatten the circle when the map's bearing changes.

So,  I think at this stage the code in this PR is the best solution I can come up with without using GeoJSON and TurfJS or similar.

I wonder if it would be good to download 3rd party libraries like Turf or parts of it on demand, when a feature, such as the GeolocateControl are added.

Just did a rough calculation:

```python
>>> 940120-938754 # without optimised TurfJS code
1366
>>> 941386-938754 # with optimised TurfJS code
2632
```

Adding the circle function pulled out of TurfJS we gain about 1.3KB, but it enables MapLibre to render circles. So, this is a tradeoff.

EDIT:
Just managed to squeeze out some more by removing no longer required code because of the TurfJS inline addition:

```python
>>> 939478-938754
724
```

EDIT:

```python
>>> 939435-938754
681
```

I think an increase in 681 bytes is the best I can do for now.

681 bytes increase is achieved with extracting the TurfJS code for just circle and adding the circle function as a unit.

I will create another PR tomorrow. So, maybe you could consider not merging this one and waiting.

HarelM: I think you can use circle-pitch-alignment: map to achieve what you need, don't you?

HarelM:
I believe the following is a far simpler solution and I believe it solves the problem...
I tested with a goejson source, with a point by updating the circleRadius, but it less smooth.
```ts
        const userLocation = this._userLocationDotMarker.getLngLat();
        if (!this.options.showUserLocation || !this.options.showAccuracyCircle || !this._accuracy ||!userLocation) {
            return;
        }
        const screenPosition = this._map.project(userLocation);
        const userLocationWith100Px = this._map.unproject([screenPosition.x + 100, screenPosition.y]);
        const pixelsToMeters = userLocation.distanceTo(userLocationWith100Px) / 100;
        const circleDiameter = 2 * this._accuracy / pixelsToMeters;
```

src/ui/control/geolocate_control.ts
HarelM, line 312:
```suggestion
```

HarelM (approved):

src/util/geolocation_support.ts
HarelM: I tested it using a geojson source and a circle layer, but it wasn't smooth enough...

PR #6254, opened 2025-08-13
Migrate to be ESM-Only

### LINKED ISSUE #1595: ECMAScript

birkskyum:
- Add ESM output
- ESM doesn't need separate CSP modules - it comply with csp already.
- Docs section for ESM is added, as well as how to use it from various popular bundlers.
- Small test projects for popular bundlers added
- Remove UMD bundle

Closes #1595 
- #1595 

https://github.com/user-attachments/assets/d9613710-9fd6-4dd6-a856-c7c1e6d1532c

I've converted below examlple to use the new ESM bundle
http://localhost:9966/test/examples/filter-layer-symbols-using-global-state.html

## Launch Checklist


HarelM:
I think this setup assumes a bit too much about where the worker code file is.
I think a middle ground would be to immitate the csp idea in terms of setup and have that support both esm and cjs.
It means the user would need to specify where the worker code is, instead of the current assumption that it is next to the main thread code. I think it will also make sure that it will be supported with all kinds of bundlers.
This is similar to RTL plugin and CSP and I think it's a smaller change, and had better chances to work correctly.

birkskyum:
@HarelM , i've split it out so that the worker url is set explicitly

https://github.com/maplibre/maplibre-gl-js/pull/6254/files#diff-538d7808b52996feefd0ff81d905bfb0dbcc1351484e2d255261769ebdb1f362R43-R48

I've also added build tests to make sure the esm build is actually esm (they don't go deep though, because i.e. murmurhash.js does have `module.export` still, but the outermost layer is esm)

package.json
```diff
@@ -3,7 +3,21 @@
...
   "main": "dist/maplibre-gl.js",
+  "module": "dist/maplibre-gl.mjs",
   "style": "dist/maplibre-gl.css",
+  "exports": {
+    ".": {
+      "types": "./dist/maplibre-gl.d.ts",
+      "import": "./dist/maplibre-gl.mjs",
+      "require": "./dist/maplibre-gl.js"
+    },
+    "./dist/maplibre-gl": {
```
HarelM: Is this "dist" section of the exports really needed?

rollup.config.ts
```diff
@@ -8,6 +8,7 @@ const {BUILD} = process.env;
 
 const production = BUILD === 'production';
 const outputFile = production ? 'dist/maplibre-gl.js' : 'dist/maplibre-gl-dev.js';
+const esmOutputFile = production ? 'dist/maplibre-gl.mjs' : 'dist/maplibre-gl-dev.mjs';
```
HarelM: There's a lint warning here about it being unused...?

rollup.config.ts
```diff
@@ -64,4 +65,37 @@ const config: RollupOptions[] = [{
     ],
 }];
 
+// ESM builds (following CSP pattern - separate main and worker files)
+// Users must explicitly call setWorkerUrl() when using these builds
+const esmConfig: RollupOptions[] = [
```
HarelM: Isn't this a duplication of the csp.config.ts definition or am I missing anything?

test/examples/filter-layer-symbols-using-global-state.html
```diff
@@ -41,7 +40,12 @@
     </fieldset>
 </body>
 
-<script>
+<script type="module">
```
HarelM, line 45: I would recommend using a different example to specifically highlight how to use the esm module.
I would also look for the docs about csp and elaborate there about the usage of esm similarly to csp.

HarelM:
This looks great! Thanks!

HarelM:
I would recommend also to use `npm run pack` and check the output in another project with bundling etc.
The fact that this now has esm might be the default for some bundlers and might cause them to pack only the esm part of the main thread instead of the cjs and cause app to stop working, so this needs to be tested as well.
You can use npx-maplibre-gl as an angular project to test and maybe Maputnic as react one.

rollup.config.ts
birkskyum: Oh, you're right. In that case the esm and the csp+esm will result in the exact same build...

test/examples/filter-layer-symbols-using-global-state.html
birkskyum: Good idea. I've added a new version of display-a-map, called display-a-map-with-esm

rollup.config.ts
birkskyum: I now reused the config from the csp

birkskyum:
I've tried npm pack, and loading it into maputnik node_modules, which appear to work fine. I believe it's still just using the .js version, rather than the new .mjs

birkskyum:
ngx-maplibre-gl appear to pick up the esm module in the showcase, so it'll only work if the showcase/main.ts file has the setWorkerUrl  added - otherwise the map will just be the light blue background without data.

```ts
import { bootstrapApplication } from '@angular/platform-browser';
import { appConfig } from './app/app.config';
import { AppComponent } from './app/app.component';
import * as maplibregl from 'maplibre-gl';

// Set the worker URL for MapLibre GL when using ESM version
maplibregl.setWorkerUrl('/maplibre-gl/maplibre-gl-worker.mjs');

bootstrapApplication(AppComponent, appConfig).catch((err) =>
  console.error(err)
);
```

birkskyum:
Either we can use the auto-worker url like there was logic for previously, or this might have to come as a breaking change if the default will be an explicit setting of the worker url.

Or both, having the auto-url logic now, and requiring explicit url from next major.

HarelM:
I don't think the auto-worker url will work for angular (webpack) unfortunately. 
So I'm not sure there's a way around this a breaking change, which is ok as we haven't released a breaking change version this year, and we might need to consider this.
But if we are doing a breaking change version we need to collect breaking changes requirements like we did for v5.

birkskyum:
I think going ESM-only in v6 is the right call here.

test/examples/display-a-map-with-esm.html
```diff
@@ -0,0 +1,33 @@
...
+</head>
+    <style>
+        body { margin: 0; padding: 0; }
+        html, body, #map { height: 100%; }
+    </style>
+</head>
+<body>
+<div id="map"></div>
+<script type="module">
+    import * as maplibregl from '../../dist/maplibre-gl.mjs';
```
HarelM: I'm not sure this will work in the docs' examples.
Can you check if a change is needed around here:
https://github.com/maplibre/maplibre-gl-js/blob/05067bd0b1b393cb7dce4083cb945f3f8182cf5c/build/generate-docs.ts#L101

rollup.config.ts
```diff
@@ -3,11 +3,13 @@ import sourcemaps from 'rollup-plugin-sourcemaps2';
 import {plugins, watchStagingPlugin} from './build/rollup_plugins';
 import banner from './build/banner';
 import {type RollupOptions} from 'rollup';
+import {config as cspConfig} from './rollup.config.csp';
```
HarelM: a better name to the file and method is probably needed here.

test/build/esm.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+describe('ESM build', () => {
+    test('ESM main bundle exists and exports expected API', () => {
+        const esmPath = path.join(process.cwd(), 'dist/maplibre-gl-dev.mjs');
+        expect(fs.existsSync(esmPath)).toBe(true);
+        
+        const content = fs.readFileSync(esmPath, 'utf8');
+        
+        // Check for ES module exports at the end of the file
+        expect(content).toMatch(/export\s+\{[^}]+\};\s*$/m);
```
HarelM: I'm not super thrilled with checking the string content of the file, is there a better way to check this?

HarelM:
I would expect to see more documentation on how to use this new ESM build, one of the places is probably here:
https://github.com/maplibre/maplibre-gl-js/blob/main/docs/index.md

But I think there might be a need for more descriptive information, maybe some info on packagers such as webpack and rollup...

HarelM:
- Relevant tracking issue: #6427

robertgzr (approved):

robertgzr:
@louwers told me so!

HarelM:
This is on the roadmap to version 6, which is the next breaking change version, it's just not clear when we would create a breaking change version...

chrneumann:
I'm experimenting to use the ESM modules with Vite/React Router v7. If you need any feedback, just contact me.

Works so far like this:

```ts
import {  Map, setWorkerURL } from "maplibre-gl";
import webWorkerURL from "maplibre-gl/dist/maplibre-gl-worker.mjs?url";
setWorkerUrl(webWorkerURL);
const map = new Map(...);
```

Vite loads the CJS module when using SSR (as far as I understood because NodeJS resolution is used on the server side per default). To prevent that, one can set it up like this in `vite.config.ts` to use Vites resolution (which uses ESM per default):

```js
export default defineConfig({
  ...,
  ssr: {
    noExternal: ['maplibre-gl'],
  },
});
```

Not sure if this is the best solution. "noExternal" might decrease module resolution performance on the server side.

louwers:
@chrneumann The `?url` search parameter is a non-standard Vite feature. This issue is about distributing an ESM build that can be used directly by browsers without a bundler.

chrneumann:
But it will also be used with bundlers? And this needs documentation (https://github.com/maplibre/maplibre-gl-js/pull/6254#issuecomment-3302105185)? Anyways, maybe it's helpful for someone who wants to use the ESM build with Vite as I do :)

HarelM:
@birkskyum version 6 first pre-releases are now published, let me know if you want to push this forward by making the PR pass or if you need help with this.

test/examples/display-a-map-with-esm.html
birkskyum: Yes, i've added .mjs

```
htmlContent = htmlContent.replace(/-dev\.js/g, '.js');
htmlContent = htmlContent.replace(/-dev\.mjs/g, '.mjs');
```

birkskyum: Screenshot 2026-04-27 at 23 57 21

rollup.config.ts
birkskyum: I've  extracted it to a shared build/rollup_bundle_config.ts

andrewda:
Very interested in this landing in v6. Happy to support with code/docs/testing as needed!

birkskyum:
@HarelM , it's functional now. I've added some docs for installation with all major bundlers / no-bundler etc.

This PR it's strictly additive and non-breaking, but it might be worth considering in a follow-up PR to go ESM-only. That would make build times much faster:
- Removing 2/3 of build targets ( UMD, UMD CSP, ESM )
- Allowing us to replace rollup with rolldown.

Arguably the webgl2 browser support will be the limiting factor anyway, if caniuse.com is to believe:
- https://caniuse.com/webgl2
- https://caniuse.com/es6-module

HarelM:
I think that's a good idea. If this is a major issue we can always add it back, but I would like version 6 to start out with ESM only, and then see if this is still an issue nowadays.

docs/index.md
```diff
@@ -64,6 +64,145 @@ const map = new maplibregl.Map({
 });
 ```
 
+## ESM
```
HarelM, line 73: You did a great work with the examples of the different bundlers, I suggest to link from the docs here to the readme with all the bundler options so people that really needs this will go there and look at the minimal examples.

HarelM:
I think the following script is no longer needed, right?
https://github.com/birkskyum/maplibre-gl-js/blob/esm-build/build/generate-dist-package.js
This creates a minimal package.json file in the dist folder for commonjs, but I think it's OK to remove this now since we can say ESM is now the default, which I'm fine with, also considering the direction of removing commonjs.
This can be made in the other PR if you prefer.
Can we also remove the CSP build then given that ESM is basically the same in terms of setup etc?

The above are not mandatory in order to merge this I think, but are a good cleanup.

HarelM:
Did you plan to add the examples bundlers to the CI?
If we won't add them to the CI they would go stale and might break.
So I advise to add a CI step/job to take care of them, probably part of the build tests.

birkskyum:
> Did you plan to add the examples bundlers to the CI? If we won't add them to the CI they would go stale and might break. So I advise to add a CI step/job to take care of them, probably part of the build tests.

I've added them to the integration suite, because they reuse the same harness as the browser tests. The tests build and check that the map loads in the browser without logging anything concerning to the browser console

.github/workflows/test-all.yml
```diff
@@ -89,7 +89,7 @@ jobs:
         if: '!cancelled()'
         with:
           name: build-dist-${{ matrix.os }}
-          path: dist/maplibre-gl.js
+          path: dist/maplibre-gl.mjs
```
HarelM: You need to upload both this file and the worker, or simply upload the content of the dist folder.

docs/index.md
```diff
@@ -53,36 +57,181 @@ You can then import the MapLibre GL JS module in your project.
...
+See the [ESM](#esm) section below for setting up the worker URL with your bundler.
 
-As a mitigation for Cross-Site Scripting and other types of web security vulnerabilities, you may use a [Content Security Policy (CSP)](https://developer.mozilla.org/en-US/docs/Web/Security/CSP) to specify security policies for your website. If you do, MapLibre GL JS requires the following CSP directives:
+## ESM
+
+MapLibre GL JS v6 ships as ES modules only (`maplibre-gl.mjs`). The `"module"` field in `package.json` points at the ESM bundle, so bundlers pick it up automatically. Every consumer needs to point MapLibre at the worker URL with [`setWorkerUrl()`](./API/functions/setWorkerUrl.md); each environment below shows how to wire it up.
+
+For minimal runnable apps per bundler (Vite, webpack, esbuild, Rollup), see [`test/integration/bundler/`](https://github.com/maplibre/maplibre-gl-js/tree/main/test/integration/bundler).
 
+### Migrating from v5
```
HarelM: This is important, but it won't be needed in the far future, let's move this migration guide next to the other migration guides and keep this intro a bit cleaner.

There might be other stuff to add there, so it makes sense to have a dedicated page for it.

docs/index.md
```diff
@@ -53,36 +57,181 @@ You can then import the MapLibre GL JS module in your project.
...
+<!-- after -->
+<script type="module">
+    import * as maplibregl from 'https://unpkg.com/maplibre-gl@^6.0.0/dist/maplibre-gl.mjs';
+</script>
 ```
 
-For strict CSP environments without `worker-src blob: ; child-src blob:` enabled, there's a separate MapLibre GL JS bundle (`maplibre-gl-csp.js` and `maplibre-gl-csp-worker.js`) which requires setting the path to the worker manually:
+The ESM build also requires a one-time `setWorkerUrl()` call (shown in the subsections below).
+
+### In the browser, without a bundler
```
HarelM: I think that zensical support tabs, so every bundler can have its own tab, I think it will make this part shorter.
Also I would move the browser without bundler to last as I don't think a lot of people are using it.

docs/index.md
```diff
@@ -53,36 +57,181 @@ You can then import the MapLibre GL JS module in your project.
...
+## CSP Directives
+
+As a mitigation for Cross-Site Scripting and other types of web security vulnerabilities, you may use a [Content Security Policy (CSP)](https://developer.mozilla.org/en-US/docs/Web/Security/CSP) to specify security policies for your website. If you do, MapLibre GL JS requires the following CSP directives:
+
+```
+worker-src 'self' ;
+img-src data: blob: 'self' ;
+```
+
+The ESM build loads the worker from a real URL (set via [`setWorkerUrl()`](./API/functions/setWorkerUrl.md)), so no `blob:` worker source is required. The previous CSP-specific bundle from v5 is no longer needed.
```
HarelM: Consider moving this to the v5-v6 migration guide.

test/build/es2020-compat.test.ts
```diff
@@ -2,21 +2,27 @@ import {describe, test, expect} from 'vitest';
 import fs from 'fs';
 import {transform} from 'esbuild';
 
+// Original context: in the v5 UMD build the worker source was inlined as a blob
```
HarelM: I'm not sure the "history lesson" is interesting here.
I would reduce the comment verbosity to just say what the test is actually testing and why, even better if you could simply change the test name to indicate that instead of the comment.

test/build/sourcemaps.test.ts
```diff
@@ -79,7 +86,9 @@ describe('main sourcemap', () => {
 
         const s1 = setMinus(actualEntriesInSourcemapJSON, expectedEntriesInSourcemapJSON);
         expect(s1.length).toBeLessThan(5);
+        // The remaining gap is pure type/interface files, which rollup tree-shakes
+        // because they have no runtime exports.
         const s2 = setMinus(expectedEntriesInSourcemapJSON, actualEntriesInSourcemapJSON);
-        expect(s2.length).toBeLessThan(17);
+        expect(s2.length).toBeLessThan(28);
```
HarelM: What is this actually testing? I've seen this updated when needed, but I'm not sure I understand the reason behind this test. Can you explain?

test/examples/add-a-3d-model-to-globe-using-threejs.html
```diff
@@ -53,6 +53,9 @@
     <br />
     <button id="project">Toggle projection</button>
     <script type="module">
+    import * as maplibregl from '../../dist/maplibre-gl-dev.mjs';
```
HarelM: Is the indentation here correct? Is it worth running something on all the examples to make sure they are properly aligned?

test/integration/browser/browser.test.ts
```diff
@@ -426,18 +427,20 @@ describe('Browser tests', () => {
...
+            }
         });
 
-        // exact message looks like
-        // Failed to execute 'importScripts' on 'WorkerGlobalScope': The script at 'http://localhost:52015/test/integration/browser/fixtures/badURL' failed to load.
-        const regex = new RegExp('Failed to execute \'importScripts\'.*');
-
-        await expect(rtlPromise).rejects.toThrow(regex);
+        // In a module worker, calling importScripts throws this specific error.
+        // Classic-worker (UMD) used to throw "Failed to execute 'importScripts'... failed to load.".
```
HarelM: Not sure it's interesting to comment about what was in the past. consider removing the second line of the comment, or both.

tsconfig.json
```diff
@@ -52,6 +52,11 @@
     "test/bench/rollup",
     "test/bench/run-benchmarks.ts",
     "test/bench/gl-stats.ts"  ,
-    "test/bench/**/*generated*"
+    "test/bench/**/*generated*",
+    "test/integration/bundler/vite-rollup-esbuild/**",
```
HarelM, line 57: `"test/integration/bundler/**"` to simplify?

rollup.config.ts
```diff
@@ -1,67 +1,9 @@
...
-        // Ingest the sourcemaps produced in the first step of the build.
-        // This is the only reason we use Rollup for this second pass
-        sourcemaps(),
-        // When running in development watch mode, tell rollup explicitly to watch
-        // for changes to the staging chunks built by the previous step. Otherwise
-        // only they get built, but not the merged dev build js
-        ...production ? [] : [watchStagingPlugin]
-    ],
-}];
+const config: RollupOptions[] = [
```
HarelM, line 23: This is so nice. This might also pave the way for rolldown, but not just yet, let's release this first to see if there's a pushback about commonjs before we use rolldown.

tsconfig.json
birkskyum: that would ignore the test suite at test/integration/bundler/bundlers.test.ts

CONTRIBUTING.md
```diff
@@ -167,14 +167,14 @@ copy node_modules/headless-gl/deps/windows/dll/x64/*.dll c:\windows\system32
...
-A standalone build allows you to turn the contents of this repository into `maplibre-gl.js` and `maplibre-gl.css` files that can be included on an html page.
+A standalone build allows you to turn the contents of this repository into `maplibre-gl.mjs`, `maplibre-gl-worker.mjs` and `maplibre-gl.css` files that can be included on an html page.
 
 To create a standalone build, run
 ```bash
 npm run build-prod
 npm run build-css
 ```
-Once those commands finish, you will have a standalone build at `dist/maplibre-gl.js` and `dist/maplibre-gl.css`
+Once those commands finish, you will have a standalone build at `dist/maplibre-gl.mjs`, `dist/maplibre-gl-worker.mjs` and `dist/maplibre-gl.css`. Load it via `<script type="module">` and call `setWorkerUrl()` to point at the worker file.
```
HarelM: The above should be `npm run build-dist` I think, it's simpler.

test/integration/render/run_render_tests.ts
```diff
@@ -13,7 +13,8 @@ import type {Page, Browser} from 'puppeteer';
 import {ensureError} from '../../../src/util/util';
 import {localizeURLs} from '../lib/localize-urls';
 import {launchPuppeteer} from '../lib/puppeteer_config';
-import type {default as MapLibreGL, Map as MaplibreMap, CanvasSource, PointLike, StyleSpecification} from '../../../dist/maplibre-gl';
+import type {Map as MaplibreMap, CanvasSource, PointLike, StyleSpecification} from '../../../dist/maplibre-gl';
```
HarelM: I think that there's already a MaplibreMap exported that was recently added, maybe it can be used here instead of the `as MaplibreMap`.

test/integration/render/run_render_tests.ts
```diff
@@ -939,8 +951,8 @@ async function closePageAndFinish(page: Page, reportCoverage: boolean) {
             source: it.text,
             ...it.rawScriptCoverage
         };
-        if (entry.url.endsWith('maplibre-gl-dev.js')) {
-            entry.sourceMap = JSON.parse(fs.readFileSync('dist/maplibre-gl-dev.js.map').toString('utf-8'));
+        if (entry.url.endsWith('maplibre-gl-dev.mjs')) {
```
HarelM: Maybe it's possible to add the worker code as well here for coverage? In the past this was not possible I think because the worker url was created using `createURLObject` which created a different temporary address.
Maybe now we can leverage the tests here to show coverage for the worker code as well.

test/integration/query/query.test.ts
```diff
@@ -99,27 +99,37 @@ describe('query tests', () => {
...
-        server = http.createServer(
-            st({
-                path: 'test/integration/assets',
-                cors: true,
-            })
-        );
+        const assetsMount = st({path: 'test/integration/assets', cors: true, passthrough: true});
+        const distMount = st({path: 'dist', url: '/dist', cors: true, passthrough: true});
+        server = http.createServer((req, res) => {
+            if (req.url === '/blank.html') {
```
HarelM: I think that both query and render are using a page that should import the scripts and has a map div. Wouldn't it be better to simply use that (a small html placed in the assets folder of the tests) instead of all this code with blank, import script, etc?

HarelM:
I believe the failing tests are related to RTL plugin loading inside the worker.
From what I read the only change needed there is to use dynamic import in stead of `importScripts`.
Let me know if you need me to help. We can transfer the branch here so we'll be able to collaborate on it if you want.

src/source/worker.ts
```diff
@@ -26,6 +26,43 @@ import {
...
+    if (/^[ \t]*(import|export)\s/m.test(code)) {
+        const blobUrl = URL.createObjectURL(new Blob([code], {type: 'text/javascript'}));
+        try {
+            await import(/* @vite-ignore */ blobUrl);
+        } finally {
+            URL.revokeObjectURL(blobUrl);
+        }
+        return;
+    }
+    (0, eval)(code);
```
HarelM: What does this `0` do?

src/source/rtl_text_plugin_worker.ts
```diff
@@ -48,7 +48,7 @@ class RTLWorkerPlugin implements RTLTextPlugin {
         return this.pluginStatus;
     }
 
-    public async syncState(incomingState: PluginState, importScripts: (url: string) => void): Promise<PluginState> {
+    public async syncState(incomingState: PluginState, loadScript: (url: string) => Promise<void>): Promise<PluginState> {
```
HarelM, line 51: I'm guessing there's no good reason to pass `loadScript` as a parameter, is it? Or is this for testing capabilities?

src/source/worker.ts
```diff
@@ -26,6 +26,43 @@ import {
...
+ *
+ * - Other URLs: fetch the source and sniff for ESM syntax (top-level `import`
+ *   or `export`). If ESM is detected, run it through a blob-URL dynamic
+ *   `import()` so the browser parses it as a module; this requires
+ *   `script-src blob:` in the worker CSP. Otherwise treat it as UMD/IIFE and
+ *   run it via indirect `eval`, which requires `script-src 'unsafe-eval'`.
+ */
+async function loadScript(url: string): Promise<void> {
+    if (url.endsWith('.mjs')) {
+        await import(/* @vite-ignore */ url);
```
HarelM, line 44: Does dynamic import fails on commonjs code that all this method complexity is needed?

birkskyum: dynamic import() does work for ESM, but it breaks the existing UMD bundle we currently load.  The first line of mapbox-gl-rtl-text@0.3.0 is: var A,B;A=this,B=function(){... 

So if the mapbox rtl plugin would support ESM, then all this trickery would go away. There is fortunately hope that there will be a new version coming with esm on the way - https://github.com/mapbox/mapbox-gl-rtl-text/releases , otherwise we'd have to vendor it

src/source/rtl_text_plugin_worker.ts
birkskyum: Yes, it's for testing convenience, so that we can mock it. The rtl_text_plugin_worker.test.ts passes a fake loadScript (a vi.fn() mock) so the tests don't have to deal with real fetches and script evaluation. Production code just passes the real loadScript from worker.ts. This was also the case for the importScripts parameter.

src/source/worker.ts
birkskyum: It's a shorthand for  `globalThis.eval(code);`, which I've update the code to use because it's more self explanatory.

test/build/sourcemaps.test.ts
birkskyum: The test is a sanity check on what ends up in the bundle. It compares two sets:

1. Source files referenced in the sourcemaps (what rollup actually bundled)
2. Files under src/ that we expect to be bundled

Then it checks both directions:

* s1 = bundled but not in src/. Should be near zero. Catches accidentally-pulled-in node_modules code or generated files.
* s2 = in src/ but not bundled. Should stay small. The only legitimate reason a src/ file is missing is rollup tree-shaking it because it has no runtime exports (pure type/interface files).

HarelM:
I've seen some docs related issues, so I opened the following PR:
- #7540

This might conflict with a change here, but I think it's better to have it in a separate PR.

test/integration/render/tests/filter/modulo/style.json
```diff
@@ -1,6 +1,11 @@
 {
         "version": 8,
         "name": "error_example",
+        "metadata": {
+          "test": {
+            "timeout": 120000
```
HarelM: Are you sure this is needed? Seems like a very long time...

test/integration/render/run_render_tests.ts
```diff
@@ -917,34 +918,67 @@ async function runTests(page: Page, testStyles: StyleWithTestData[], directory:
     }
 }
 
-async function createPageAndStart(browser: Browser, testStyles: StyleWithTestData[], directory: string, options: RenderOptions) {
+async function createPageAndStart(browser: Browser, testStyles: StyleWithTestData[], directory: string, options: RenderOptions, serverPort: number) {
     const page = await browser.newPage();
+    const workers: WebWorker[] = [];
+    page.on('workercreated', async (worker: WebWorker) => {
```
HarelM: Consider moving some of this code to a shared file like pupperteer_config by exporting these methods to reduce code duplication.

mkdocs.yml
```diff
@@ -79,3 +79,8 @@ validation:
   absolute_links: warn
   unrecognized_links: warn
   anchors: warn
+
+# README.md exists for GitHub readers (it documents how to run the docs locally).
+# Without this exclude, zensical/mkdocs-material renders it at `/` instead of index.md.
+exclude_docs: |
```
HarelM, line 83: - I've addressed this change in #7540

HarelM:
This looks great now! Thanks for all the work on this!
I've added some last comments. I'll merge this once I see that a new pre-release is ready to be released with just this change.

test/integration/render/tests/filter/modulo/style.json
birkskyum: fixed with prewarm, so the timeout change is reverted

HarelM (approved):

pakastin:
Vite is skipping the maplibre-gl-shared.mjs on production build. How to fix that?
In other words, the Vite example only works on dev build and not production build, because of maplibre-gl-shared.mjs import failing. Would we also need setSharedUrl(sharedUrl) like the one for worker?

HarelM:
I'm not a vite expert, feel free to open a PR to fix the example/test in this repo to match your issue so that we can make sure this doesn't break in the future.

HarelM:
When using this code in Angular I'm getting the following warning:
main-4536KWGN.js (2:241742): Use of eval in "main-4536KWGN.js" is strongly discouraged as it poses security risks and may cause issues with minification.
This is due to this line:
https://github.com/maplibre/maplibre-gl-js/blame/74590c62d98f825d6966dead9e7ae0abe9e74315/src/source/worker.ts#L67

Is there a different way to solve this?

PR #6263, opened 2025-08-16
Feature/6226 covering tiles method

### LINKED ISSUE #6226: Add a public method for getting tile coordinates covering the current view
## User Story

`map.coveringTiles(options)` returns tile IDs that cover the current view. 

## Rationale

In Maplibre v4, you could use an internal method to get a list of tiles covering the current view — `(map.transform.coveringTiles({tileSize: 512})`. This can be helpful in a number of cases, but one common one is fetching some kind of tiled data from a server for an overlay, e.g. a tiled GeoJSON.

In Maplibre v5, there's no way to get this information, so I'd propose exposing a public method (e.g. `map.coveringTiles(options)` which would simply be an alias to `coveringTiles(this.transform, options)`.

## Impact

Not seeing a way to do this otherwise in v5, other than explicitly importing `covering_tiles.ts` from the source, but this seems very fragile.

shreeharshshinde:
This PR introduces a new public method `map.coveringTiles(options)` to allow users to get a list of tile coordinates that cover the current viewport.

**Motivation**

As described in issue #6226, this was a useful capability in MapLibre v4 that is currently missing in v5. A common use case is fetching data for a custom tiled overlay from a server.

**Changes**

- Added the `coveringTiles(options)` method to the `Map` class in `src/ui/map.ts`.
- Exported a new `CoveringTilesOptions` type for use with the method.
- Added comprehensive JSDoc documentation and usage examples.
- Created a new unit test file with tests to verify functionality and option handling.
- Added an entry to `CHANGELOG.md`.

Closes #6226


shreeharshshinde:
Hi @HarelM ,

The implementation, unit tests, and linting for this feature (map.coveringTiles) are now complete. The PR should be ready for review, but I've run into a persistent warning during the `npm run generate-docs `step that I haven't been able to solve on my own, and I would appreciate your guidance.


When I run `npm run generate-docs`, the process completes but consistently shows the following warning:
```
[warning] CoveringTilesOptions, defined in maplibre-gl/src/ui/map.ts, is referenced by Map.coveringTiles.options but not included in the documentation
```

HarelM:
You'll need to add to the list of exported types in index.ts the covering tiles options object. 

I've looked in the converting tiles file and it seems that there is a type there very similar to the public API you added, the main difference I saw was the terrain object, which obviously can't be exported, but I wonder how it changes the behavior when terrain is on, if at alll.

I would recommend trying to unite the types somehow to avoid the need to duplicate it for the public API.
Otherwise, this PR looks good, thanks!

shreeharshshinde:
Ah, I can't believe I forgot to update index.ts! 🥲 That's a lesson I won't forget. Thank you so much, @HarelM - for your help and guidance!

shreeharshshinde:
> Thanks for taking the time to open this PR! You'll need to add to the list of exported types in index.ts the covering tiles options object.
> 
> I've looked in the converting tiles file and it seems that there is a type there very similar to the public API you added, the main difference I saw was the terrain object, which obviously can't be exported, but I wonder how it changes the behavior when terrain is on, if at alll.
> 
> I would recommend trying to unite the types somehow to avoid the need to duplicate it for the public API. Otherwise, this PR looks good, thanks!

@HarelM are you suggesting importing `CoveringTilesOptions` under a different name, like `InternalCoveringTilesOptions`, and then creating an exported type in the form of a key–value pair inside `Pick<>`?

HarelM:
I find `Pick` fragile, but something like creating a public type and then extend it with relevant properties for internal usage.

shreeharshshinde:
Is this PR planned to be merged as is, or would you prefer me to work on that change within this PR?

HarelM:
I think it would be better to work on the fixes I mentioned as part of this PR, I don't think they are too intrusive...

shreeharshshinde:
@HarelM Please review the updated changes as I have made Public type for covering tiles which is exported in both `map.ts` and `index.ts` !!!

src/geo/projection/covering_tiles.ts
```diff
@@ -34,15 +34,24 @@ export type CoveringZoomOptions = {
...
 };
 
-export type CoveringTilesOptions = CoveringZoomOptions & {
-    /**
-     * Smallest allowed tile zoom.
-     */
+/**
+ * Public options for calculating covering tiles, used in {@link Map.coveringTiles}.
+ */
+export type PublicCoveringTilesOptions = {
```
HarelM: I would go the other way around with the names as these appear in the docs.
i.e. call this `CoveringTilesOptions` and call the other one `CoveringTilesOptionsInternal`.

src/ui/map_tests/map_covering_tiles.test.ts
```diff
@@ -0,0 +1,45 @@
...
+            center: [0, 0],
+            zoom: 2,
+            style: {
+                version: 8,
+                sources: {},
+                layers: []
+            }
+        });
+
+        map.on('load', () => {
```
HarelM: use `await map.once('load')` instead.

src/ui/map_tests/map_covering_tiles.test.ts
```diff
@@ -0,0 +1,45 @@
+import {describe, vi, expect, test} from 'vitest';
+import {createMap} from '../../util/test/util';
+
+vi.stubGlobal('ResizeObserver', vi.fn(() => ({
```
HarelM, line 4: Is this really needed as part of these tests? Seems a bit odd to me...

src/ui/map_tests/map_covering_tiles.test.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    test('respects the maxzoom parameter', () => {
+        const map = createMap({
+            zoom: 5,
+            style: {version: 8, sources: {}, layers: []}
+        });
+
+        map.on('load', () => {
+            const tiles = map.coveringTiles({tileSize: 512, maxzoom: 4});
+            tiles.forEach(tile => {
```
HarelM: Nit picking: Use regular `for` instead

src/ui/map_tests/map_covering_tiles.test.ts
```diff
@@ -0,0 +1,45 @@
...
+                layers: []
+            }
+        });
+
+        map.on('load', () => {
+            const tiles = map.coveringTiles({tileSize: 512});
+            expect(Array.isArray(tiles)).toBe(true);
+            expect(tiles.length).toBeGreaterThan(0);
+            expect(tiles[0]).toHaveProperty('canonical');
+            map.remove();
```
HarelM: Consider using beforeEach to initialize the map, and afterEach to remove the map, and only change the zoom as part of the test.

src/ui/map_tests/map_covering_tiles.test.ts
shreeharshshinde: I found that the tests were failing with a '''ReferenceError: ResizeObserver is not defined''' without this mock.

I agree it probably belongs in a more central location. Is there a global test setup file where I should move this instead?

src/ui/map_tests/map_covering_tiles.test.ts
shreeharshshinde: Done

src/ui/map_tests/map_covering_tiles.test.ts
shreeharshshinde: Done

src/ui/map_tests/map_covering_tiles.test.ts
shreeharshshinde: Done

src/geo/projection/covering_tiles.ts
shreeharshshinde: Done

src/ui/map_tests/map_covering_tiles.test.ts
shreeharshshinde: Was there a specific reason to do that @HarelM ? , would like to know about it

HarelM: When using forEach you create a lambda which is harder to debug in terms of stack trace etc, without gaining anything, so the coding convention of this code it to use for...

shreeharshshinde: Ahh, I get it

src/ui/map.ts
```diff
@@ -838,6 +840,23 @@ export class Map extends Camera {
...
+    *
+    * @param options - Options for calculating the covering tiles.
+    * @returns An array of `OverscaledTileID` objects.
+    * @example
+    * ```ts
+    * // Get the tiles to cover the view for a 512x512px tile source
+    * const tiles = map.coveringTiles({tileSize: 512});
+    * ```
+    */
+    coveringTiles(options: CoveringTilesOptionsInternal): OverscaledTileID[] {
```
HarelM, line 856: This is the new public API.
It needs to declare a method which uses CoveringTilesOptions and not use something with the name "internal"

src/ui/map_tests/map_covering_tiles.test.ts
HarelM: Checkout other tests, there's a setup method that can be used I belive.

src/ui/map.ts
shreeharshshinde: I did that before but whole pipeline was crashed due to terrain

src/ui/map_tests/map_covering_tiles.test.ts
shreeharshshinde: `
beforeEach(() => {
    beforeMapTest();
});` is used in along all test cases in map_tests and unit test is passed as I expected so should I push that code replacing the `vi.stubGlobal()`

src/ui/map.ts
HarelM: You probably need to rename the current one to internal and then create a new one with only the relevant options for the public API, and then try and unite them...

HarelM:
Closing in favor of:
- #6292

Thanks for the work on this, I took most of it, just did a few minor changes as I want to release a new minor version this month and I wanted this to be included...

shreeharshshinde:
Thanks for your guidance @HarelM, catch you in next PR !!

PR #6270, opened 2025-08-18
Add bundle statistics to maplibre-gl-js website

wayofthefuture:
There seemed to be interest on the call about being able to analyze bundle composition and dependency sizes... I posted this on Slack and didn't seem to have any objections. This generates four separate html files and places them in the docs/assets/bundle-stats directory, which are then accessible through the docs website. I'm not certain of how this would affect the build so definitely need a pro to give it a look before merging.

https://github.com/user-attachments/assets/2726fe3b-dbe8-4f20-a565-fdc0868abfd1


HarelM:
Yeah, saw it in slack and forgot to reply, sorry.
While I understand there's a question about bundle size and what's being used etc, I'm not sure the docs site is the right place to put it.
In the past, there was a bundle size analysis as part of every PR. I removed it since no one was looking at it.
I would consider adding this to the instructions of a readme as I believe this is more for developers and less for someone who decides to use this library...
There's this folder with useful developers tips, maybe the instructions on how to generate this report can be added there?
https://github.com/maplibre/maplibre-gl-js/tree/main/developer-guides
What do you think?

wayofthefuture:
Yeah maybe it could be in https://github.com/maplibre/maplibre-gl-js/blob/main/developer-guides/release-process.md to check build stats and size... would have to be different a package.json script I think

HarelM:
You might be able to use the bump PR version github action to add it to the generated PR, but since I'm mostly doing these releases and I'm not sure how helpful this would be when releasing a version I'm not sure I fully follow.
I guess what I'm saying is that I'm missing a problem statement:
Who are the users who need this?
What will it solve for them?
And from those definitions, we can better define the process and where to integrate the tool you have showed.
Please know that I'm not trying to discourage you, or being hard, I just want to better understand the motivation etc...

wayofthefuture:
Oh, my motivation was solely based on @nyurik's statements on the 7/9/25 call, but I could have misunderstood! I believe we were talking about the vt-pbf dependency at the time. Nevertheless I thought it was wanted and looked cool, could help the devs, and a flashy pitch on the public side 😄. I put it below the CDN build as it felt like hey here's the build you can take a look if you'd like, Maplibre's got it together! Anyway, whatever you guys think 🧠 🌀

HarelM:
I would advise to add the instructions here: 
https://github.com/maplibre/maplibre-gl-js/blob/main/CONTRIBUTING.md

wayofthefuture:
Roger!

wayofthefuture:
Done. (maybe 😄)

HarelM:
Is an entry in gitignore missing? Where is the output of this?

CONTRIBUTING.md
```diff
@@ -165,9 +165,15 @@ To create a standalone build, run
...
 npm run build-css
 ```
-
 Once those commands finish, you will have a standalone build at `dist/maplibre-gl.js` and `dist/maplibre-gl.css`
 
+## Analyze your Standalone Build
+
+MapLibre seeks to keep production bundle sizes small. If your changes involve substantial additions or changes to dependencies, the effects on bundle size can be visualized by running
+```bash
+npm run bundle-stats
```
HarelM, line 174: What do I do after that to see the different outputs?

wayofthefuture: The parameter "open: true" in the rollup plugin options will automatically open the default browser to all 4 visualizations.

wayofthefuture:
> Is an entry in gitignore missing? Where is the output of this?

The files are placed into the staging directory next to the staging build which is already included in .gitignore.

HarelM (approved):

PR #6292, opened 2025-08-25
Expose covering tiles method in map

### LINKED ISSUE #6226: Add a public method for getting tile coordinates covering the current view

HarelM:
## Launch Checklist

- Replaces #6263
- Closes #6226

This adds the ability to get the covering tiles from the map object.

 - [x] Link to related issues.

### DIFF
--- a/src/geo/projection/covering_tiles.test.ts
+++ b/src/geo/projection/covering_tiles.test.ts
@@ -1,7 +1,7 @@
 import {beforeEach, describe, expect, test} from 'vitest';
 import {GlobeTransform} from './globe_transform';
 import {LngLat} from '../lng_lat';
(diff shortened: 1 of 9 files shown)


birkskyum:
I can see why that can be useful - is there a feature request ticket, or some background anywhere?

PR #6301, opened 2025-08-26
Fix `global-state` in pattern properties

melitele:
Fix recalculating paint properties `...-pattern` after change in `global-state` (issue described [here](https://github.com/maplibre/maplibre-gl-js/pull/6215#issuecomment-3221373866))

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
Thanks for looking into this!!
Can you check if other places are missing the global state somehow? Maybe make it non optional somehow to see where it might be needed?

dpeter99:
Hi! I have been trying to use the global state to store the ID of a selected location and than draw a different marker based on that. (Not sure if there is any better way to do this, given that we have too many locations to store them as HTML markers)
The following icon layoud doesn't seam to work:
```js
'icon-image': [
        'case',
        // ['global-state', 'selectedId']
        ['==', ['get', 'id'], ['global-state', 'selectedId']], `${icon}-selected`,
        icon,
      ],
```
nor does this:
```js
'text-field': [
      'coalesce',
      ['global-state', 'selectedId'],
      'null',
    ],
```
(I only get null)

HarelM:
I would advise to remove and add a layer instead of using global state for that use case...

dpeter99:
> I would advise to remove and add a layer instead of using global state for that use case...

The use case for me is that we have map pins for each location and than when clicked on one, I want to show a different pin for that single location.

Obviously I want the least amount of re-draw and flickering for this. The data is in GeoJSON and each feature point has the ID as property.
How would addig/removing a separate layer solve this. Would the default pin not render under the selected layer's pin? And I would still need to filter the GeoJSON containing all the locations?

HarelM:
In that case a feature-state is probably a better solution, although feature state has a lot of bugs as well...

melitele:
> 
> ```js
> 'icon-image': [
>         'case',
>         // ['global-state', 'selectedId']
>         ['==', ['get', 'id'], ['global-state', 'selectedId']], `${icon}-selected`,
>         icon,
>       ],
> ```
> 
> ```js
> 'text-field': [
>       'coalesce',
>       ['global-state', 'selectedId'],
>       'null',
>     ],
> ```

@dpeter99 Both work in my tests. Can you create failing render tests or [jsbin](https://jsbin.com)?

src/data/bucket/line_bucket.ts
```diff
@@ -120,6 +120,7 @@ export class LineBucket implements Bucket {
     programConfigurations: ProgramConfigurationSet<LineStyleLayer>;
     segments: SegmentVector;
     uploaded: boolean;
+    features: any;
```
HarelM: Is it possible to type this better?

src/data/bucket/line_bucket.test.ts
```diff
@@ -9,8 +9,11 @@ import {LineBucket} from './line_bucket';
 import {LineStyleLayer} from '../../style/style_layer/line_style_layer';
 import {type LayerSpecification} from '@maplibre/maplibre-gl-style-spec';
 import {type EvaluationParameters} from '../../style/evaluation_parameters';
-import {type BucketFeature, type BucketParameters} from '../bucket';
+import {type BucketFeature, type BucketParameters, type IndexedFeature, type PopulateParameters} from '../bucket';
 import {SubdivisionGranularitySetting} from '../../render/subdivision_granularity_settings';
+import {OverscaledTileID} from '../../source/tile_id';
+import {FeatureIndex} from '../feature_index';
+import {GeoJSONWrapper} from '@maplibre/vt-pbf';
```
HarelM: Why is this used here? Is it possible to use something else instead?

src/data/bucket/line_bucket.ts
melitele: copilot ;-)

HarelM:
Can you make sure the changed lines of code are covered with tests?
Codecov is not working for me for a while now, I probably need to look for alternatives... Sorry for the extra work...

melitele:
> Can you make sure the changed lines of code are covered with tests?

[100%](https://app.codecov.io/gh/maplibre/maplibre-gl-js/commit/655d84ad720a0a8341dfd1a95d0881e8f6e17f20)

src/data/bucket/fill_extrusion_bucket.test.ts
```diff
@@ -0,0 +1,46 @@
...
+import {FillExtrusionBucket} from './fill_extrusion_bucket';
+import {FillExtrusionStyleLayer} from '../../style/style_layer/fill_extrusion_style_layer';
+import {type LayerSpecification} from '@maplibre/maplibre-gl-style-spec';
+import {type EvaluationParameters} from '../../style/evaluation_parameters';
+import {type BucketParameters, type IndexedFeature, type PopulateParameters} from '../bucket';
+import {OverscaledTileID} from '../../source/tile_id';
+import {FeatureIndex} from '../feature_index';
+
+// Load a fill feature from fixture tile.
+const vt = new VectorTile(new Protobuf(fs.readFileSync(path.resolve(__dirname, '../../../test/unit/assets/mbsv5-6-18-23.vector.pbf'))));
```
HarelM, line 15: Would it make sense to move this logic to a test util file? I've seen these used a few times in these bucket test. Maybe even move it to a before each method or something more "vitest-ty"?

HarelM:
This looks great! Thanks! I've added last nit-picking comment and after addressing it this can be merged.

src/data/bucket/fill_extrusion_bucket.test.ts
melitele: separate PR to refactor *_bucket tests

HarelM: Fair enough. Thanks!

PR #6317, opened 2025-08-29
chore(ci): pin actions to shas

CommanderStorm:
## Launch Checklist

Our CI uses a few actions.
For these actions, we currently just use the mutable GitHub tag.

Since we use Dependabot to update the versions, we should use SHAs.
This makes sure that we are not affected by a certain class of supply chain vulnerability where attackers re-publish bad tags.

Using SHAs matches GitHub recommendations and is a part of the OpenSSFs Scorecard.

       ^--- not sure if you want this. Other maintenance actions don't show up as well.


HarelM:
Will dependabot update those? Will it use sha when updating? 
I generally think this is a good practice, but I don't want to add maintenance overhead for that.

CommanderStorm:
> Will dependabot update those? Will it use sha when updating?

Yes, the current sheme (sha, and version in comment) is updated correctly by dependabot.
Blog post from 2022: https://github.blog/changelog/2022-10-31-dependabot-now-updates-comments-in-github-actions-workflows-referencing-action-versions/

I don't think this adds mantenance overhead, at least it does not for martin.

Issue #6308, opened 2025-08-27
Global state doesn't work with icon-size/text-size layout property of symbol layer


Hooterr:
**maplibre-gl-js version**: 5.7.0

**browser**: Chrome

### Steps to Trigger Behavior

 1. Set global state in style `size: { default: 5.0 }`
 3. Use it in symbol layer `'icon-size': ['global-state', 'size']`
4. Expression evaluates to null as per console warning `Expected value to be of type number, but found null instead.`

This appears to happen with both `icon-size` and `text-size`.

### Link to Demonstration
https://jsbin.com/lasutonuzi/1/edit?html,output
(There are 2 buttons in the top left that call `map.setGlobalStateProperty` but the map still doesn't react to this change)

melitele:
`EvaluationParameters` missing reference to `globalState` in [symbol/symbol_layout.ts](https://github.com/maplibre/maplibre-gl-js/blob/3547c2c45d3696111c232fcda6ad264f04b4f350/src/symbol/symbol_layout.ts#L98) and [symbol/symbol_size.ts](https://github.com/maplibre/maplibre-gl-js/blob/3547c2c45d3696111c232fcda6ad264f04b4f350/src/symbol/symbol_size.ts#L44)

melitele:
@Hooterr PR #6321 fixes this issue if you like to test it.

Hooterr:
Thanks a lot!

PR #6321, opened 2025-08-31
Fix for #6308 Global state doesn't work with icon-size/text-size layout property of symbol layer

melitele:
Fixes the issue [#6308] of `global-state` ignoring `icon-size`, `text-size` and any other layout property that has a special evaluation point that was missed when adding `globalState` to `EvaluationParameters`.

The change ignores all the references to `globalState` passed via `EvaluationParameters` for layout properties and instead "binds" the `globalState` reference - that has been passed from `Style` to `Layer` to `_unevaluatedLayout` - to the `expression.evaluate` method.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


src/style/properties.ts
```diff
@@ -315,10 +328,12 @@ export class Transitioning<Props> {
 export class Layout<Props> {
     _properties: Properties<Props>;
     _values: {[K in keyof Props]: PropertyValue<any, PossiblyEvaluatedPropertyValue<any>>};
+    #globalState: Record<string, any>; // reference to global state
```
HarelM: Why the #?

src/style/style_layer.ts
```diff
@@ -306,6 +311,13 @@ export abstract class StyleLayer extends Evented {
         (this as any).paint = this._transitioningPaint.possiblyEvaluate(parameters, undefined, availableImages);
     }
 
+    set globalState(globalState: Record<string, any>) {
```
HarelM: If this is not a trivial setter use a method instead.

HarelM:
Thanks for taking the time to open this PR and fix this issue! 

Can you elaborate a bit more on the design around this fix? I'm not sure I fully understand what you wrote in the initial comment of this PR.

melitele:
> Can you elaborate a bit more on the design around this fix? I'm not sure I fully understand what you wrote in the initial comment of this PR.

The change is based on the `global-state` being a singleton that is already properly propagated to workers. Instead of handing it to `expression.evaluate` as a field of `EvaluationParameters` (and making sure we don't forget to reference it in each and every place we call `expression.evaluate`), we pass the reference to `global-state` from `Style` to `Layer` to `_unevaluatedLayout` to layout expression. The expression holds the reference to `global-state` and uses it whenever `expression.evaluate` is called.

This PR only addresses layout properties. It'll get some unit tests before it's ready. The follow up PRs will do the same for paint properties and filters. The final step is to clean the code from references to `global-state` that were passed around with the only purpose of being added to the instance of `EvaluationParameters`.

src/style/properties.ts
melitele: `#` makes it runtime `private` as opposed to gentleman's agreement  `private` of underscore `_` prefix and compile-time `private` of TypeScript
[MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements)

HarelM: Thanks! I was not familiar with it. 
But I prefer to keep the current possibilities and not introduce a new one, so either use _ or typescript's private.

HarelM:
Ahh, great! This sounds like the right approach.

src/style/style.test.ts
```diff
@@ -469,6 +469,38 @@ describe('Style.loadJSON', () => {
...
+                        type: 'fill',
+                        source: 'fill-source-id',
+                        filter: ['global-state', 'showFill']
+                    }
+                ]
+            })
+        );
+        await style.once('style.load');
+        const globalState = style.getGlobalState();
+        //@ts-ignore
```
HarelM: Is it possible to check public properties instead somehow?

src/style/properties.ts
```diff
@@ -93,6 +96,16 @@ export class PropertyValue<T, R> {
     ): R {
         return this.property.possiblyEvaluate(this, parameters, canonical, availableImages);
     }
+
+    setGlobalState(globalState: Record<string, any>) {
+        //@ts-ignore
+        this.expression.evaluate = (globals: GlobalProperties, feature?: Feature, featureState?: FeatureState, canonical?: ICanonicalTileID, availableImages?: Array<string>, formattedSection?: FormattedSection) => {
```
HarelM, line 102: This is a temporary fix until the change in style spec will be incorporated, right?

melitele: right

src/source/worker_tile.test.ts
```diff
@@ -286,4 +286,33 @@ describe('worker tile', () => {
...
+                paint: {
+                    'fill-extrusion-height': ['global-state', 'height']
+                }
+            }
+        ]);
+
+        const globalState = {join: 'bevel', height: 1};
+        const tile = createWorkerTile({globalState});
+        await tile.parse(createLineWrapper(), layerIndex, [], {} as any, SubdivisionGranularitySetting.noSubdivision);
+        //@ts-ignore
```
HarelM: Is it possible to check public properties instead somehow?

src/style/style.test.ts
```diff
@@ -2739,7 +2771,7 @@ describe('Style.setLayerZoomRange', () => {
         vi.spyOn(style, '_reloadSource');
 
         style.setLayerZoomRange('raster', 5, 12);
-        style.update(0 as any as EvaluationParameters);
+        style.update({zoom: 0} as EvaluationParameters);
```
HarelM, line 2775: This is great! Thanks for improving the types!

HarelM:
You want this merged before changing the style-spec package, right?

src/style/style.test.ts
melitele: Do you mean that there should be a public property on the StyleLayer that exposes the reference to `global-state`?

melitele:
yes, it'll take a while to do the change in style-spec package and refactor the `global-state` implementation in `maplibre-gl-js` and this fixes the bug now

src/style/style.test.ts
HarelM: No, I think that it should be kept private, but you should be able to test something that is public which global state affects, maybe the results of evaluate or something?

HarelM: If this is impossible in this scope (the scope of style.ts file) then maybe this test should be placed elsewhere, or removed, I'm not sure...

melitele: The intension of the test is to check that it is reference to the same object, checking result of evaluate wouldn't do that as it would pass if it was a copy.

melitele: `global-state` is a singleton (with allowances for workers) and the test tries to check that it is treated as such

HarelM: If this is a singleton, then it mean you should be able to change it from the "outsite" and see that the changes are reflected, wouldn't you? i.e. set global state, then evaluate, then change global state then evaluate again and see that the change is correct.
I mean I think the test should be around **what** it does and less about **how** it does it.

melitele: yes!!! will do

HarelM (approved):
THANKS!!!

src/style/style.test.ts
```diff
@@ -469,6 +471,37 @@ describe('Style.loadJSON', () => {
...
+                    }
+                ]
+            })
+        );
+        await style.once('style.load');
+        // tests that reference to globalState is propagated to layers
+        // by changing globalState property and checking if the changed value
+        // was used when evaluating the layer
+        const globalState = style.getGlobalState();
+        globalState.size = 12;
```
HarelM, line 499: I must admit this feels a bit weird from an API perspective where you get a reference to an object, change it and it changes stuff inside the library.
From an API perspective I would expect the code to be:
```js
const globalState = style.getGlobalState();
globalState.size = 12;
style.setGlobalState(globalState);
```
And changing the `globalstate` should not affect the internal state of the library.

Having said that, I'm guessing this would change soon so let's focus on solving the bug as this is a minor issue.

Here's an issue that was opened recently on a similar case where an object is returned and manipulated and causes unexpected behavior from the developer point of view:
- https://github.com/maplibre/maplibre-gl-js/issues/6184

PR #6335, opened 2025-09-02
Fix 6111

### LINKED ISSUE #6111: Drag/touch pan moves in the wrong direction when map is pitched.
**maplibre-gl-js version**: 5.6.0 and 5.6.1

**browser**: Chrome/Edge

If the map is pitched and you attempt to drag pan upwards quickly, the map moves in the wrong (opposite) direction. Note that this also occurs on touch screens where a normal finger swipe easily triggers the behavior.

### Steps to Trigger Behavior

 1. Load a pitched map. (60 degrees of pitch is enough)
 2. Quickly drag pan the map upwards. (similar to a finger swipe)

### Link to Demonstration
https://codepen.io/wayofthefuture/pen/xbGvoJa?editors=0011

There are two buttons on this demo. The first button behaves normally, the second button causes the issue to occur. Note that each of these buttons use the same underlying logic, except the drag Y distance is increased.

NathanMOlson:
This PR fixes #6111.

The fundamental issue is that the pan inertia is calculated in screen pixels. At high pitch angles, the calculated pan motion sometimes crosses the horizon. When it does the calculated map point is behind the camera instead of in front, causing the motion to be backwards.

This PR fixes the issue by reducing the pan offset so that it never crosses the horizon. The exact logic I used is:

- Clip the pan offset to half the distance to horizon.

There is nothing special about this exact choice, the important thing is that the pan offset does not cross the horizon.

 - [X] Link to related issues.

### DIFF
--- a/src/geo/projection/mercator_camera_helper.ts
+++ b/src/geo/projection/mercator_camera_helper.ts
@@ -3,7 +3,7 @@ import {LngLat, type LngLatLike} from '../lng_lat';
 import {cameraForBoxAndBearing, type CameraForBoxAndBearingHandlerResult, type EaseToHandlerResult, type EaseToHandlerOptions, type FlyToHandlerResult, type FlyToHandlerOptions, type ICameraHelper, type MapControlsDeltas, updateRotation, type UpdateRotationArgs} from './camera_helper';
 import {normalizeCenter} from '../transform_helper';
 import {rollPitchBearingEqual, scaleZoom, zoomScale} from '../../util/util';
-import {projectToWorldCoordinates, unprojectFromWorldCoordinates} from './mercator_utils';
+import {getMercatorHorizon, projectToWorldCoordinates, unprojectFromWorldCoordinates} from './mercator_utils';
 import {interpolates} from '@maplibre/maplibre-gl-style-spec';
 
 import type {IReadonlyTransform, ITransform} from '../transform_interface';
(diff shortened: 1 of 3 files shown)


wayofthefuture:
Good job. Does the panning / touch finger swipe feel natural? Hopefully I'll have some time tomorrow to pull down and give it a try.

NathanMOlson:
> Good job. Does the panning / touch finger swipe feel natural? Hopefully I'll have some time tomorrow to pull down and give it a try.

It's definitely better. To me it still feels asymmetrical if I "throw it" up versus down at high pitch angles. This is because the panning is done in screen coordinates, and the relationship between "screen distance" and "map distance" is changing very rapidly near the horizon.

src/geo/projection/mercator_camera_helper.ts
```diff
@@ -79,7 +79,15 @@ export class MercatorCameraHelper implements ICameraHelper {
 
         const zoom = optionsZoom ? +options.zoom : tr.zoom;
 
-        let pointAtOffset = tr.centerPoint.add(options.offsetAsPoint);
+        // Reduce the offset so that it never goes past the horizon. If it goes past
+        // the horizon, the pan direction is opposite of the intended direction.
+        const offsetLength = options.offsetAsPoint.mag();
```
HarelM, line 84: Does it make sense to move this logic to where the inetria is calculated, or does is this issue related to easeTo in general and not just the inetria?

HarelM:
Thanks for looking into this! 
Should there be a fix for globe as well?

src/geo/projection/mercator_camera_helper.ts
NathanMOlson: I think it makes more sense to do it here, in case other callers of `easeTo()` requests offsets that cross the horizon.

NathanMOlson:
> Should there be a fix for globe as well?

I added identical unit test for globe and they were already passing.

src/geo/projection/mercator_camera_helper.ts
HarelM: I think that in globe it is handled in panInertiaHandler:
https://github.com/maplibre/maplibre-gl-js/blob/4ad1e69f7c26ec385b7df47df1b38110492bc8b5/src/geo/projection/vertical_perspective_camera_helper.ts#L26

So I would recommend fixing it the same way in both, and I tend to think the inertia handler is the right place.

NathanMOlson: OK, see #6345.

NathanMOlson:
Closed in favor of #6345

PR #6338, opened 2025-09-03
Merge coverage reports and add coverage annotation

HarelM:
## Launch Checklist

Codecov seems to ignore my issue and I'm having a hard time understanding if the coverage of tests for a specific PR is good enough or not.
This PR does the following:
1. Uploads coverage report for type of test (unit, integration, render).
2. Combines the coverage report to a single html report and uploads it as an artifact.
3. Uploads only one coverage report to codecov (instead of sending all the files), which means that the coverage report on a PR when it is successful won't be partial until the tests complete, as it is today.
4. Annotate on the PR for changes with missing coverage.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -194,14 +194,7 @@ function setWorkerUrl(value: string) { config.WORKER_URL = value; }
  */
 function importScriptInWorkers(workerUrl: string) { return getGlobalDispatcher().broadcast(MessageType.importScript, workerUrl); }
 
-function uncoveredMethodForTestingThatNeedsToBeDeleted() {
-    let i = 0;
-    i++;
-    console.log(i);
-}
-
 export {
-    uncoveredMethodForTestingThatNeedsToBeDeleted,
     Map,
     NavigationControl,
     GeolocateControl,


.github/workflows/test-all.yml
```diff
@@ -167,3 +183,48 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: [unit-tests, integration-tests, render-tests]
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+        with:
+          node-version-file: '.nvmrc'
+
+      - name: Download coverage reports
+        uses: actions/download-artifact@abefc31eafcfbdf6c5336127c1346fdae79ff41c
```
louwers, line 198: Is this a real commit?

We recently switched to SHA based actions. I know it is best practice but I feel like there should be some workflow that auto fixes it if you just specify a version.

Pretty hard to author workflows otherwise.

HarelM: Yeah, it's not a great developer experience...
It's a real commit.

PR #6342, opened 2025-09-03
Fix regression in mergeSourceDiffs: handle add/remove/removeAll 

### LINKED ISSUE #6339: Regression with mergeSourceDiffs: add/remove/update inconsistencies
By merging sourceDiffs, PR #6163 fixed some performance issues.
Unfortunately, this change also introduced regressions.

The resulting data after a merge can differ from what is expected.

**maplibre-gl-js version**: 5.7.0

**browser**: Firefox / Chrome

### Steps to Trigger Behavior

Here is a minimal example:

```ts
const point: GeoJSON.Feature = {
    type: 'Feature',
    id: 'point',
    geometry: {
        type: 'Point',
        coordinates: [0, 0]
    },
    properties: {},
};

source.updateData({add: [point]});
source.updateData({removeAll: true});
```

Without merging, the point is correctly removed.
With merging, the point is still visible.

The following [test](https://github.com/maplibre/maplibre-gl-js/blob/09f1516380487f931b5be59eb98b3de2cc142f26/src/source/geojson_source_diff.test.ts#L412) demonstrates the issue:
`merged.add` is still present after merging:

```ts
test('merges two diffs with removeAll', () => {
    const diff1 = {
        add: [{type: 'Feature', id: 'feature2', geometry: {type: 'Point', coordinates: [1, 1]}, properties: {}}],
    } satisfies GeoJSONSourceDiff;

    const diff2 = {
        removeAll: true,
    } satisfies GeoJSONSourceDiff;

    const merged = mergeSourceDiffs(diff1, diff2);
    expect(merged.add).toHaveLength(1);
    expect(merged.removeAll).toBe(true);
});
```

However, in [applySourceDiff](https://github.com/maplibre/maplibre-gl-js/blob/09f1516380487f931b5be59eb98b3de2cc142f26/src/source/geojson_source_diff.ts#L110) , `removeAll` is applied first.
As a result, the point remains present after applying the merged diff:

```ts
export function applySourceDiff(updateable: Map<GeoJSONFeatureId, GeoJSON.Feature>, diff: GeoJSONSourceDiff, promoteId?: string): void {
    if (diff.removeAll) {
        updateable.clear();
    }

    if (diff.remove) {
        for (const id of diff.remove) {
            updateable.delete(id);
        }
    }

    if (diff.add) {
        for (const feature of diff.add) {
            const id = getFeatureId(feature, promoteId);

            if (id != null) {
                updateable.set(id, feature);
            }
        }
    }

    // ...
}
```

Other problematic cases include:

- Adding a feature, then removing it → results in removing and then adding it back so the point is still present
- updating properties in `existingDiff` that don't need to be updated because they are removed in the `newDiff`.

### Link to Demonstration

I’ve prepared a reproduction for the first case.
The issue is revealed in tests 3 and 6, which introduce a heavy workload for the worker and therefore force updates to be merged.

click the reset button between each test.

- Tests 1, 2 and 6 work as expected by **removing all the features from the source** (the operation can take more or less time depending on the timers set up)
- Tests 3 and 6 highlight the issue by **displaying the red rectangle**

https://jsbin.com/ribelevexo/edit?html,output

I can open a PR to address the issue for tests 3 and 6.

Unfortunately, test 5 reveals a separate issue that I cannot explain: only the first time, and only with Symbol text, the updateData is not respected. (with jsbin : refresh the browser and click the test 5 button) On Firefox, a vertical green line also remains visible.

https://github.com/user-attachments/assets/538e120d-3637-422e-b7f7-05f1763d87b7

just in case I tested with a transition duration at 0 and it didn't change anything

lhapaipai:
## Launch Checklist

PR #6163 improved performance by merging source diffs, but it also introduced regressions where the merged result could diverge from the expected behavior.

This PR fixes the inconsistencies in `mergeSourceDiffs` by ensuring that:

- removeAll clears previous add and update operations
- remove cancels out corresponding add and update operations
- redundant add/remove sequences are properly neutralized

- Fixes #6339

## Changes in this PR

Fixed `mergeSourceDiffs` logic to correctly handle edge cases with add, remove, and removeAll

Added unit tests to cover the problematic cases:
- add → removeAll
- removeAll → add
- update → remove
- add → remove
- remove → add → remove

## Impact

This aligns mergeSourceDiffs with the semantics of applySourceDiff, ensuring consistency and preventing features from incorrectly persisting after removal.


HarelM:
This looks great!
I'm currently working on making coverage work again, so please give me a day or two to complete this and then I want to check it with the PR.

HarelM (approved):

PR #6345, opened 2025-09-03
Fixes wrong movement when inertia crosses the horizon

### LINKED ISSUE #6111: Drag/touch pan moves in the wrong direction when map is pitched.

NathanMOlson:
This PR fixes #6111.

The fundamental issue is that the pan inertia is calculated in screen pixels. At high pitch angles, the calculated pan motion sometimes crosses the horizon. When it does the calculated map point is behind the camera instead of in front, causing the motion to be backwards.

This PR fixes the issue by reducing the pan offset so that it never crosses the horizon. The exact logic I used is:

- Clip the pan offset to half the distance to horizon.

There is nothing special about this exact choice, the important thing is that the pan offset does not cross the horizon.

 - [X] Link to related issues.


src/ui/camera.test.ts
```diff
@@ -15,6 +15,8 @@ import {MercatorCameraHelper} from '../geo/projection/mercator_camera_helper';
 
 import type {GlobeProjection} from '../geo/projection/globe_projection';
 import type {Terrain} from '../render/terrain';
+import {getMercatorHorizon} from '../geo/projection/mercator_utils';
```
HarelM: Sorry for nit picking, I should probably find a lint rule for this, but can you move these imports up to be before the type imports?

HarelM:
Thanks! This looks great!
Also thanks for adding the globe tests as well!
Can I close the other PR?

src/ui/camera.test.ts
NathanMOlson: Done.

HarelM:
Thanks!
Let me see if I can integrate the changes I made to the coverage report and I'll test them out on this branch.
I'll merge this right afterwards.
@wayofthefuture this is completed from my point of view.

wayofthefuture:
I will be sitting down in about 15 min to give it a look thanks

wayofthefuture:
Ok so I have been testing it on an iPad Pro 13" and the first thing I notice is that sideways panning on pitch is sluggish. I also compared it to the inertia of kinetic pinch zoom as I felt they should be similar. So I increased the factor applied by @NathanMOlson which seemed to make the inertia more natural (similar to pinch zoom), would this be viable?

```javascript
const offsetAsPoint = pan.mult(Math.min(pixelsToHorizon * 0.8 / offsetLength, 1.0));
```

NathanMOlson:
> So I increased the factor applied by @NathanMOlson which seemed to make the inertia more natural (similar to pinch zoom), would this be viable?

Any factor less than 1.0 should prevent the direction reversal.

wayofthefuture:
This seems to feel natural on both desktop and tablet:

```javascript
        const offsetAsPoint = pan.mult(Math.min(pixelsToHorizon / 3 / offsetLength, 1.0));
```

wayofthefuture:
I keep getting tripped up because it totally depends on the amount of pitch, which is skewing my judgment. I do think the higher factor might be too fast for desktop however, so your original 50% might be the safest solution. So I'm good with the commit as is @HarelM

HarelM:
OK, Please ping me when tests are completed and you are happy with the results and I'll merge this.

NathanMOlson:
This is ready to go from my perspective.

wayofthefuture:
@HarelM I just spent about an hour playing with this because I was concerned with tablet touch and pitch. I frequently get the feeling when panning that 50% of the horizon distance is just a little too damped, and that 75% of this distance feels more responsive and inline with other gestures especially at 0 pitch.

NathanMOlson:
> @NathanMOlson Please let me know your thoughts.

I am indifferent.

wayofthefuture:
You understand it much better than me! I think you can assess my theory about swipes above as from here it's just a theory I'm still maplibre green! 📗 🍏 💚

NathanMOlson:
@wayofthefuture The logic I added limits how far inertia can move the map based on where the horizon is. It only comes into play with big moves near the horizon, it leaves the behavior unchanged for small moves or moves at low pitch angle.

Changing the 0.5 to 0.75 just changes the definition of 'big' and 'near the horizon'. It makes it require a bigger move to hit the limit.  There's no right answer, it's just a "feel" thing. I'm indifferent because I don't use a mouse or trackpad with highly-pitched maps, so I happy to defer to someone who does.

wayofthefuture:
@NathanMOlson Ok that makes sense, if you are okay with changing the factor in your code to 0.75 then I'm happy with that.

NathanMOlson:
> @NathanMOlson Ok that makes sense, if you are okay with changing the factor in your code to 0.75 then I'm happy with that.

Done!

HarelM:
@wayofthefuture let me know if this can be merged. Looks good from my point of view.

wayofthefuture:
Looks great, thanks!

HarelM:
Great, thanks! I'll release a new version shortly.

PR #6363, opened 2025-09-06
Fix #6361 - `global-state` is not evaluated for layers added after loading style

### LINKED ISSUE #6361: Global state not evaluated correctly in layers outside map style.
**maplibre-gl-js version**: 5.7.1

**browser**: Any

### Steps to Trigger Behavior

 1. Add global style in style `color: { default: "red" },`
 2. Use global state in a layer outside map style 
 3. `global-state` expression evaluates to `null`

Layers that are part of map style evaluate global state correctly.

Same behavior if you remove global state from style definition and set it before `addLayer` call. Layers that are part of the style will evaluate it correctly but others won't.

### Link to Demonstration
https://jsbin.com/qanihureqe/edit?html,output

melitele:
- Fixes #6361 

`global-state` was evaluated for layers added after loading style using `Style.addLayer()` prior to PR #6321. That PR was based on the premise that layout properties will be evaluated using `global-state` help by the layer while all other properties will continue to be evaluated using `global-state` passed as parameter to `evaluate()` function. More generally all invocations of the `evaluate()` function that didn't have the reference to `global-state` in expression, would continue to use `global-state` passed as parameter. PR #6321 didn't fully implement this assumption leading to regression issue #6361.

This PR ensures that reference to `global-state` is taken from layer private property if set and from parameter to `evaluate()` if not set on the layer.

## Launch Checklist

 - [x] Link to related issues.


src/style/style.ts
```diff
@@ -1069,6 +1069,8 @@ export class Style extends Evented {
             layer.setEventedParent(this, {layer: {id}});
         }
 
+        layer.setGlobalState(this._globalState);
```
melitele, line 1072: Having missed this case in #6321 makes me think that passing `global-state` [via constructor](https://github.com/maplibre/maplibre-style-spec/pull/1267#discussion_r2315842125) is more robust than a setter. This would be a [substantial change](https://github.com/maplibre/maplibre-style-spec/pull/1267#discussion_r2317948916) not suitable for this PR but to be done in the refactoring of `global-state` evaluation #6366 that I'm working on.

HarelM:
Is there anything else needed before I merge this?

src/style/properties.ts
```diff
@@ -100,9 +100,10 @@ export class PropertyValue<T, R> {
     setGlobalState(globalState: Record<string, any>) {
         //@ts-ignore
         this.expression.evaluate = (globals: GlobalProperties, feature?: Feature, featureState?: FeatureState, canonical?: ICanonicalTileID, availableImages?: Array<string>, formattedSection?: FormattedSection) => {
-            // force the use of global state object when evaluating any expression
+            // use `global-state` reference if not provided as parameter to evaluate()
+            // passing an empty object as evaluate() for `global-state` relies on it existing
             //@ts-ignore
-            globals.globalState = globalState;
+            globals.globalState ??= globalState ?? {};
```
HarelM, line 106: Nit picking: in most other places in the code we use ||

melitele: I think we should move toward using ?? instead of || for initializations as it doesn't confuse falsy with uninitialized [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing)

melitele:
> Is there anything else needed before I merge this?

It's ready.

src/style/properties.ts
HarelM: Yeah, the problem is that the rest of the code is not using it.
But I generally agree.
Maybe a lint rule can help solving those throughout the codebase...

melitele: The rest of the code should migrate to using it, starting with places that are being changed. There are other modern constructs that we started using without changing it everywhere all at once. For instance `Set` and `Map` while we still have objects as hashmaps elsewhere...
That's my interpretation of [Code Conventions](https://github.com/maplibre/maplibre-gl-js/blob/fdc61e0b75f7c054e1e86c9ac496531c9be44033/CONTRIBUTING.md?plain=1#L196)

HarelM: Fair enough.

PR #6366, opened 2025-09-07
Refactor passing `global-state` object from `Style` to expressions

melitele:
The reference to `global-state` has been passed around between classes and functions that directly don't do anything related to it, only to make it available when expression's `evaluate` function is called. It resulted in a number of issues (#6301, #6308, #6048), all caused by missing to provide the reference to `global-state` to `evaluate`.

This PR addresses the problem of making sure that all invocations of the `evaluate` function have reference to the `global-state` object by passing it to expression's constructor instead and letting the expression hold on to it and use when needed.

Specifically the reference to the `global-state` from Style to layers' constructors to expressions' constructors. To make it available in workers, it is also passed together with serialized layers in `setLayers` and `updateLayers` called from main thread to worker (and removed from `WorkerTile` parameters).

This PR relies on PR [#1267](https://github.com/maplibre/maplibre-style-spec/pull/1267) and has to be merged in tandem.

The only 2 top commits are relevant here:

- [pass global-state from Style to layers to expressions as parameter to constructor](https://github.com/maplibre/maplibre-gl-js/commit/6048e22845b8abf5573db0ce204817a7ce4151f9)

- [remove global-state passed around in order to pass it to the evaluate function](https://github.com/maplibre/maplibre-gl-js/commit/fc98cf172a21d52b56583408b051ef3f28e8399b)

This PR _sits_ on PR #6363 plus some code necessary for the checks to pass before `maplibre-style-spec` with PR [#1267](https://github.com/maplibre/maplibre-style-spec/pull/1267) is released. The commit history will be cleaned prior to graduating from Draft.

## Launch Checklist

 - [x] Link to related issues.


HarelM:
Don't worry about commit history as PR are squashed anyway...

HarelM:
Thanks for digging into this!!

melitele:
There are 2 issues that need to be fixed to complete this PR:

- function [EvaluationParameters.isSupportedScript](https://github.com/maplibre/maplibre-gl-js/blob/64789be735d37a71f700f4a40f48e75c3efe99ff/src/style/evaluation_parameters.ts#L43) is not passed when evaluating expressions (that's the reason [is-supported-script/layout](https://github.com/maplibre/maplibre-gl-js/tree/main/test/integration/render/tests/is-supported-script/layout) fails) - to be fixed here

- filters based on `global-state` don't work - requires change in `maplibre-style-spec` PR [#1279](https://github.com/maplibre/maplibre-style-spec/pull/1279)

HarelM:
Thanks for the update! Does this mean more render tests are needed or otherwise other tests to cover the filter part or is this taken care of in style-spec repo?

melitele:
> Does this mean more render tests are needed or otherwise other tests to cover the filter part ...

Yes. I'm adding test(s) for filters with 'global-state' here.

package.json
```diff
@@ -24,7 +24,7 @@
     "@mapbox/unitbezier": "^0.0.1",
     "@mapbox/vector-tile": "^2.0.4",
     "@mapbox/whoots-js": "^3.1.0",
-    "@maplibre/maplibre-gl-style-spec": "^23.3.0",
+    "@maplibre/maplibre-gl-style-spec": "melitele/maplibre-style-spec.git#global-state-property-dist",
```
HarelM: Don't forget to revert this :-)

test/bench/benchmarks/filter_evaluate.ts
```diff
@@ -27,7 +27,7 @@ export default class FilterEvaluate extends Benchmark {
             const layerFilters = [];
             for (const filter of filters) {
                 if (filter.layer === name) {
-                    layerFilters.push(createFilter(filter.filter));
+                    layerFilters.push(createFilter(filter.filter as any));
```
HarelM: Why was this added? Is there a way to avoid this "as any"?

test/bench/benchmarks/filter_create.ts
```diff
@@ -6,7 +6,7 @@ import filters from '../data/filters.json' with {type: 'json'};
 export default class FilterCreate extends Benchmark {
     bench() {
         for (const filter of filters) {
-            createFilter(filter.filter);
+            createFilter(filter.filter as any);
```
HarelM: Same here

src/style/style_layer.ts
```diff
@@ -155,7 +156,7 @@ export abstract class StyleLayer extends Evented {
 
     setFilter(filter: FilterSpecification | void) {
         this.filter = filter;
-        this._featureFilter = featureFilter(filter);
+        this._featureFilter = featureFilter(filter, this._globalState);
```
HarelM, line 159: Just to make sure - when global state changes these places in the code where global state was passed get update, right?

src/style/evaluation_parameters.ts
```diff
@@ -62,3 +57,7 @@ export class EvaluationParameters {
             {fromScale: 0.5, toScale: 1, t: 1 - (1 - t) * fraction};
     }
 }
+
+function isSupportedScript(str: string): boolean {
```
HarelM, line 66: Why was this moved out of the class?

src/source/query_features.ts
```diff
@@ -40,6 +40,7 @@ export type QueryRenderedFeaturesOptions = {
 
 export type QueryRenderedFeaturesOptionsStrict = Omit<QueryRenderedFeaturesOptions, 'layers'> & {
     layers: Set<string> | null;
+    globalState?: Record<string, any>;
```
HarelM, line 47: This is not part of the public API, right? If I understand correctly, this is needed in order to pass global state into the method, right?

src/source/query_features.ts
```diff
@@ -199,20 +203,21 @@ export function queryRenderedSymbols(styleLayers: {[_: string]: StyleLayer},
...
 
-export function querySourceFeatures(sourceCache: SourceCache, params: QuerySourceFeatureOptions | undefined): GeoJSONFeature[] {
+export function querySourceFeatures(sourceCache: SourceCache, params: QuerySourceFeatureOptions | undefined, globalState?: Record<string, any>): GeoJSONFeature[] {
     const tiles = sourceCache.getRenderableIds().map((id) => {
         return sourceCache.getTileByID(id);
     });
 
     const result: GeoJSONFeature[] = [];
 
+    const qsfParams = globalState ? (params ? {...params, globalState} : {globalState}) : params;
```
HarelM: Why not pass an empty object by default here and reduce some of the complexity of this line?
Maybe `globalState` can be mandatory in `QueryRenderedFeaturesOptionsStrict`?

src/source/tile.test.ts
```diff
@@ -47,9 +47,17 @@ describe('querySourceFeatures', () => {
         tile.querySourceFeatures(result, {sourceLayer: undefined, filter: ['!=', 'oneway', true]});
         expect(result).toHaveLength(0);
         result = [];
-        const polygon = {type: 'Polygon',  coordinates: [[[-91, -1], [-89, -1], [-89, 1], [-91, 1], [-91, -1]]]};
+        const polygon = {type: 'Polygon',  coordinates: [[[-91, -1], [-89, -1], [-89, 1], [-91, 1], [-91, -1]]]} as GeoJSON.GeoJSON;
         tile.querySourceFeatures(result, {sourceLayer: undefined, filter: ['within', polygon]});
         expect(result).toHaveLength(1);
+
+        // filter with global-state
```
HarelM: This comment probably indicates that this needs to be a separate test, shouldn't it?

HarelM:
I've added a few minor comments but otherwise, this looks great.
Also the fact that the tests were not changed a lot and still cover the original flow is a good indication that the tests previously written were well written to capture the flow and not the implementation details.

The coverage report annotation that is failing is due to a missing coverage by tests of one of the lines in the code that changed I believe, it's still new so I'm sorry for testing it out on you.
Do let me know if it makes sense or not, and I'll think how to improve the DX around it.
It used to be codecov in the past, but it stopped working, so this is still a work in progress...

src/style/style_layer.ts
melitele: it does now :-) see this [jsbin](https://jsbin.com/divagan/edit?html,js,output)
good catch, forgot to update `globalState` in `worker`

src/style/evaluation_parameters.ts
melitele: TLDR: because [`spread operator (...)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) only works on own enumerable properties

My read on the definition of the [`GlobalProperties`](https://github.com/maplibre/maplibre-style-spec/blob/458cff9ead2e9f030c77068b1fb81de5bde32784/src/expression/index.ts#L59) is a simple, read only object. Hence my assumption that I can use operations that work on own enumerable properties (`spread operator (...)`, `Object.keys()` etc). And since it's read only, it has to be [copied prior to modification](https://github.com/maplibre/maplibre-style-spec/blob/458cff9ead2e9f030c77068b1fb81de5bde32784/src/expression/index.ts#L117) and spread operator is what is usually used.

`EvaluationParameters` is passed as `GlobalProperties` - this change makes the relevant fields into own enumerable properties. The `isSupportedScript` implementation has no use of `this` so I moved it out.

The alternative to this change is to modify `maplibre-style-spec` so it doesn't assume the fields defined by `GlobalProperties` are own and enumerable.

HarelM: Does moving the method out what allowed the render test to pass?
I not sure I'm following the value of moving the method out of the class, can you elaborate on that?

melitele: Adding `isSupportedScript` as a property:
```
isSupportedScript: (_: string) => boolean = isSupportedScript;
```
allowed the render test to pass because the property became own and enumerable, i.e. visible to spread operator that I used to make a copy of the object passed as a parameter to 'evaluate` [here](https://github.com/maplibre/maplibre-style-spec/blob/458cff9ead2e9f030c77068b1fb81de5bde32784/src/expression/index.ts#L117).

Moving the method out of the class dealt with the duplicate identifier problem that arisen when I added the property.

src/source/query_features.ts
melitele: `QueryRenderedFeaturesOptions` is public, `QueryRenderedFeaturesOptionsStrict` is not. The latter is the definition of parameters passed to the internal functions [`queryRenderedFeatures`](https://github.com/maplibre/maplibre-gl-js/blob/c4347b81a6f3a742118035cd30e91ebc767e3f67/src/source/query_features.ts#L104) & [`queryRenderedSymbols`](https://github.com/maplibre/maplibre-gl-js/blob/c4347b81a6f3a742118035cd30e91ebc767e3f67/src/source/query_features.ts#L144) that now need `globalState` because they indirectly instantiate filters (specifically call functions that do).

src/style/evaluation_parameters.ts
HarelM: This seems a bit weird to me. While all other properties are data, and this object is DTO (Data transfer object), this property kind of break it.
It is possible to remove this property and calculate this part when it's needed?
What is the trade of if we remove `isSupportedScript` from `GlobalProperties`? Or it's not possible because the evaluation logic is in style-spec and the RTL and other stuff is here?

src/source/query_features.ts
melitele: This is `params: QuerySourceFeatureOptions | undefined`, `QuerySourceFeatureOptions` being public. I'll see what I can do, I don't like this line either.

src/style/evaluation_parameters.ts
melitele: It's possible to move some or all of the [script detection](https://github.com/maplibre/maplibre-gl-js/blob/main/src/util/script_detection.ts) code to `maplibe-style-spec`. It may even be a better _home_ for it. Then the only parameter that needs to be passed is a boolean that the RTL plugin is loaded.

Beyond the scope of this PR.

HarelM: Definitely beyond the score of this PR.
I was thinking of moving the evaluation logic into this library and keeping only the valuation logic there. 
But it's a big task either way.

src/source/worker.ts
```diff
@@ -170,14 +173,30 @@ export default class Worker {
...
 
         this.actor.registerMessageHandler(MessageType.updateLayers, async (mapId: string, params: UpdateLayersParameters) => {
-            this._getLayerIndex(mapId).update(params.layers, params.removedIds);
+            this._getLayerIndex(mapId).update(params.layers, params.removedIds, this._getGlobalState(mapId));
+        });
+
+        this.actor.registerMessageHandler(MessageType.updateGlobalState, async (mapId: string, params: Record<string, any>) => {
+            const globalState = this._getGlobalState(mapId);
+            for (const key in params) {
+                globalState[key] = params[key];
```
HarelM, line 183: What about deleting "removed" keys from global state?

src/source/worker.test.ts
```diff
@@ -140,4 +140,13 @@ describe('Worker generic testing', () => {
         worker.actor.messageHandlers[MessageType.removeMap]('0', undefined);
         expect(worker.layerIndexes['0']).toBeUndefined();
     });
+
+    test('propagates global state', () => {
+        const globalState = {key: 'value'};
+        worker.actor.messageHandlers[MessageType.updateGlobalState]('0', globalState);
+        //@ts-ignore
```
HarelM: Is there a better way to test this without using `ts-ignore`?

src/source/worker.ts
melitele: `globalState` in worker is an exact copy of the `globalState` in main thread. There is no deleting "removed" keys there.

HarelM: This for look like like it a "sync" between the worker and the main thread, but it only adds keys and not removing keys, or am I misreading this?

melitele: this is what [Style.setGlobalStateProperty](https://github.com/maplibre/maplibre-gl-js/blob/c437b82ea8a4f93b5cb9c3fce4f486ddb717951f/src/style/style.ts#L308) does:
```
const newValue = value === null ? this.stylesheet.state?.[name]?.default ?? null : value;
this._globalState[name] = newValue; 
```
and this is what [Style.setGlobalState](https://github.com/maplibre/maplibre-gl-js/blob/c437b82ea8a4f93b5cb9c3fce4f486ddb717951f/src/style/style.ts#L328) does:
```
for (const propertyName in newStylesheetState) {
    this._globalState[propertyName] = newStylesheetState[propertyName].default;
}
```

HarelM: If params (main thread global state) has prop1 and prop2 but global state used to have prop3 on the worker, where is prop3 removed?

melitele: How do you remove `prop3` from `global-state` on the main thread?

HarelM: I set a new global state that doesn't have prop3...?

melitele: won't work, the [Style.setGlobalState](https://github.com/maplibre/maplibre-gl-js/blob/c437b82ea8a4f93b5cb9c3fce4f486ddb717951f/src/style/style.ts#L328) iterates over properties of the new object, so the `prop3` stays:
```
for (const propertyName in newStylesheetState) {
    this._globalState[propertyName] = newStylesheetState[propertyName].default;
}
```

HarelM: Is this the intended behavior? Seems a bit strange to me that I can't delete properties from global state, although probably not a very common use case, but still...

melitele: This is how it was originally implemented #5613
We can discuss it. Preferably outside this PR.

HarelM: Fair enough, I can respect that.

HarelM:
I did a review of the current changes up util now, they look good.
Ping me when this is ready for review and I'll do another sweep.
Thanks again for all the hard work on this feature!

melitele:
@HarelM Ready for another review. I appreciate your thoroughness. Thank you.

PR #6385, opened 2025-09-10
Automatically update the reference size of the minified bundle

melitele:
There is a test [build/min.test.ts](https://github.com/maplibre/maplibre-gl-js/blob/main/test/build/min.test.ts) to verify that the PR doesn't increase the minified bundle by more than 1024 bytes (unless intended).

The test fails even if the PR contributes less 1024 bytes if the last merged PR was below 1024 bytes but didn't update the reference size hardcoded in the test [build/min.test.ts](https://github.com/maplibre/maplibre-gl-js/blob/main/test/build/min.test.ts).

The test wouldn't fail in that case if the reference size were an actual size of the minified bundle produced by the last merged PR i.e. the current top of the `main` branch.

This PR introduces a separate file [build/bundle_size.json](https://github.com/melitele/maplibre-gl-js/blob/auto-dist-size-maplibre/test/build/bundle_size.json) that is automatically updated with the size of the minified bundle whenever there is commit to the `main` branch.

The value in that file will still have to be manually updated by the person creating a new PR but only when it exceeds 1024 bytes.

## Launch Checklist

### DIFF
--- a/.github/workflows/test-all.yml
+++ b/.github/workflows/test-all.yml
@@ -105,6 +105,18 @@ jobs:
             ${{ github.workspace }}/upload/coverage-final-integration.json
             ${{ github.workspace }}/upload/coverage-final-query.json
           retention-days: 1
+      - name: Get bundle size
+        if: ${{ !cancelled() && runner.os == 'Linux' && startsWith(github.ref, 'refs/heads/main') }}
+        run: |
+          du -b dist/maplibre-gl.js | cut -f1 > test/build/bundle_size.json
+      - name: Commit bundle size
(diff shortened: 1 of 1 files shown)


HarelM:
While I definitely relate to your point, I'm good with slowly increasing it when needed, as I do want to manage this more closely. 
I think updating the size on main will not achieve this unfortunately. 
But I'm open to discussion 😀

melitele:
If you have an idea how to manage bundle size while avoiding manual updates, let me know. I'll be happy to help to implement it.

HarelM:
I think manual updates are OK. Maybe better ergonomics around updating this value from a developer perspective are needed.
Something like running a script locally that updates the size so you can run it and commit the change easily?

PR #6395, opened 2025-09-12
Fix #5430 - when TileJSON fails to load ignore source but trigger load and idle events

### LINKED ISSUE #5430: load and idle event never triggered when  a source loading fails
It seems that when source fails to load that the `load` and `idle` event are never triggered. 
I suppose  that if tiles.json requests fails  the map should just ignore them and triggers the events properly.

**maplibre-gl-js version**: 5

**browser**: all

### Steps to Trigger Behavior

 1. Open https://codepen.io/oterral/pen/YPKgyGq
 2. An alert box should appear for the 'load' and 'idle' event but notthing happens

### Link to Demonstration

Open https://codepen.io/oterral/pen/YPKgyGq

or try this:

```
// Initialise the map
  container: 'map',
  style: "https://vectortiles.geo.admin.ch/styles/ch.swisstopo.leichte-basiskarte_world.vt/style.json",
  center: [7, 47],
  zoom: 10
});

map.on('load',()=> alert("loadevent"));
map.on('idle',()=> alert("idleevent"));
```

melitele:
- Fixes #5430

When source TileJSON fails to load it is already ignored. However the source is treated as not loaded yet which causes events `load` and `idle` to be withheld.

The `load` event is used to signal that the map is ready. Specifically it means that map style can be modified. This condition is met after TileJSON fails to load, so the `load` event can be triggered in that case. Similarly the `idle` event means that there are no outstanding repaints, which again is true after TileJSON fails to load and map is fully repainted.

This PR marks sources with TileJSON that failed to load as loaded, enabling `load` and `idle` events to trigger. 

## Launch Checklist

 - [x] Link to related issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 ### 🐞 Bug fixes
 - Fix `_updateRetainedTiles` checking for children when children length is 1 overscaled tile "child" ([#6388](https://github.com/maplibre/maplibre-gl-js/pull/6388))
 - Fix evaluating `global-state` for layers added after loading style ([#6361](https://github.com/maplibre/maplibre-gl-js/issues/6361))
+- Fix triggering `load` and `idle` events when source TileJSON fails to load ([#5430](https://github.com/maplibre/maplibre-gl-js/issues/5430))
 - _...Add new stuff here..._
 
 ## 5.7.1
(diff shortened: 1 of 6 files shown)


HarelM:
LGTM, let me know when this is ready to merge.

melitele:
@HarelM it's ready

Issue #6242, opened 2025-08-12
Handling of context lost/restored doesn't work?


elo58:
**maplibre-gl-js version**: tested with 4.7.1 and 5.6.2 

**browser**: chrome and firefox

### Steps to Trigger Behavior

Create a simple map, force a lose/restore context with ctx.getExtension("WEBGL_lose_context").loseContext() and ctx.getExtension("WEBGL_lose_context").restoreContext() 

The map is not correctly displayed
Lot of warnings in console about webgl errors.

### Link to Demonstration
https://jsbin.com/gesoyacivu/1/edit?html,js,console,output

Use 'l' key to lose context and 'r' key to restore the context.

### Expected Behavior
Tiles are correctly displayed after  the restore event.

### Actual Behavior
Tiles are not correctly displayed (some texts are displayed but some elements are missing)

HarelM:
Thanks for taking the time to report this issue 
I'm guessing no one really tested this so far and thus this broken behavior. 
Feel free to dig into this...

HarelM:
I think this is now fixes by #6398.
If this is not the case please open request to reopen this issue.

PR #6398, opened 2025-09-12
fix: contextRestored event not working

ToHold:
## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.
 
## Problem description:
The map is not properly restored after a contextRestored event. Some buffers and resources from the previous context are still being used, which causes the new context to crash and triggers WebGL warnings.

## Changes: 
Ensure that the style is properly cleared when a contextLost event occurs, so that the contextRestored event can work as expected.
On contextRestored, create a new Style object based on the existing one to fully reinitialize the style state.

## Pull request note:
I'm not entirely confident about all the changes in this PR, as this area is new to me. A review with advice and suggestions would be greatly appreciated.
I also need help with testing, since it requires a real WebGL API. Could you provide some guidance or share a sample test setup?
 
## Related issue :
https://github.com/maplibre/maplibre-gl-js/issues/6242

### DIFF
--- a/test/build/min.test.ts
+++ b/test/build/min.test.ts
@@ -38,7 +38,7 @@ describe('test min build', () => {
         const decreaseQuota = 4096;
 
         // feel free to update this value after you've checked that it has changed on purpose :-)
-        const expectedBytes = 958463;
+        const expectedBytes = 963443;
 
         expect(actualBytes).toBeLessThan(expectedBytes + increaseQuota);
         expect(actualBytes).toBeGreaterThan(expectedBytes - decreaseQuota);


src/style/style.ts
```diff
@@ -1967,4 +1967,90 @@ export class Style extends Evented {
             }
         }
     }
+
+    /**
+     * Détruit toutes les ressources internes du style (sources, images, layers, etc.)
```
HarelM: Comments should be in English...

HarelM:
There's a webgl mocking library that is being used for unit tests. There are integration tests if you need a real browser. 
There are readme files scattered around the project in strategic places.
If there's more specific help you might need let me know. 
Generally speaking, it seems like clearing everything is risky, but if you can test and and make sure it works as expected then I'm ok with this change. 
I believe the convention around resource clearing is done in most cases with a method called "remove" if I'm not mistaken, but do validate this...

ToHold:
Thanks for your response.
There is indeed a Puppeteer test environment that allows testing in a real WebGL context. I’ve set up two tests so far:

- one checks that the map is not empty (i.e., it renders pixels other than just white),

- the other ensures that there are no more warnings related to the context restore issue.

Regarding the naming convention for destruction methods, I noticed that several classes already use destroy, so it seemed consistent to follow the same pattern.

For clearing everything, I tested that the “polygons” and “custom” layers render correctly again afterward. Specifically, I copy the Style when the context lost event occurs, then recreate it on context restore. My thinking was that destroying everything first and then rebuilding it would help avoid memory leaks. However, I’m not completely sure if I’m destroying everything properly.
Is there someone familiar with the Style handling who could review this part?

Finally, do you think I should add more tests, such as rendering tests, to verify that the map looks identical after the restore?

ToHold:
I added a commit introducing an operation called loseAndRestoreContext for test-render.
Should I add this operation to all test-render tests, or only to some of them?

HarelM:
I don't see an added value in adding this to the render tests... Render test are more static in most cases, they also should be paired to native, which I'm not sure context loss exists there...

ToHold:
Render tests compare the rendered output as image data against the expected image data. They use Puppeteer, which runs a real headless browser (usually Chromium or Chrome). This allows the tests to use WebGL in the browser to render layers, styles, etc.

Since this PR destroys and recreates the Style, I added an operation to check some render tests. Right now, I’m running it locally with all the style.json (I added LoseAndRestoreContext operation to them). Most tests are passing, but a few specific ones are timing out. I’m investigating the cause and working on fixing it.

ToHold:
I was just asking whether you want me to add the LoseAndRestoreContext operation to all existing tests, since it could slow them down.

ToHold:
In my opinion, if all render tests pass with the LoseAndRestoreContext operation enabled, the PR can be considered fully safe.

HarelM:
As a feature test, that's a great tests, but I don't want to run this on every PR, the integration test you added is good enough for keeping the functionality working for the long run...

ToHold:
So I’ll remove the last commit with the LoseAndRestoreContext operation in the render tests?
I’ll still investigate why some of them were failing and fix those issues.

HarelM:
Yes please

ToHold:
I fixed an issue where images added via map.addImage were not being restored after a context restore.

Also, I should mention that custom layers need to be re-added after a context restore. In test/examples/add-a-3d-model-to-globe-using-threejs.html, this works well because the example adds the custom layer inside map.on('style.load', ...), so it automatically gets added again when the context is restored.

src/render/image_manager.ts
```diff
@@ -334,4 +348,16 @@ export class ImageManager extends Evented {
             }
         }
     }
+
+    _cloneImages() {
+        const clonedImages: {[_: string]: StyleImage} = {};
```
HarelM: Can `Record` be used here instead?

test/integration/browser/browser.test.ts
```diff
@@ -442,4 +442,81 @@ describe('Browser tests', () => {
...
+            const x = Math.floor(width / 2);
+            const y = Math.floor(height / 2);
+            const readY = height - y - 1;
+            const rgba = new Uint8Array(4);
+            gl.readPixels(x, readY, 1, 1, gl.RGBA, gl.UNSIGNED_BYTE, rgba);
+
+            return {ok: true, pixel: Array.from(rgba), x, readY, width, height, dpr};
+        });
+
+        if (!evalResult.ok) {
```
HarelM: I'm not sure I'm following all this logic at the end of the test, why not simply write the relevant "expect"s?

src/render/image_manager.ts
```diff
@@ -334,4 +348,16 @@ export class ImageManager extends Evented {
             }
         }
     }
+
+    _cloneImages() {
```
HarelM: Why is clone images needed? won't they be created when the context is restored?

ToHold: Some images added via map.addImage are not stored in the Style, so they weren’t being restored from this._lostContextStyle in map.ts. I decided the simplest solution was to clone the images before the context is restored.

test/integration/browser/browser.test.ts
ToHold: You can’t use expect inside page.evaluate; you need to test the resolved data afterward. That’s why I use expect later with:
``` TS
const px = evalResult.pixel as number[];
expect(px[3]).toBeGreaterThan(0);
```
But I could just check the buffer directly instead of using an object with "ok" property. I’ll update this commit accordingly.

src/render/image_manager.ts
ToHold: Alright I'll update this part

src/render/image_manager.ts
HarelM: Yeah... hmm... IDK, feels a bit weird to store those alongside the style.
Or maybe use a single object with two fields so that if we find other stuff in the future that needs to be added they won't pile up on the map object?

HarelM: BTW, since it is used outside this class, it probably shouldn't be prefixed with "_"...

HarelM:
BTW, what about customer layers? I don't think they withstand context lost, do they?

src/render/image_manager.ts
ToHold: I’ll take a look at this and try to come up with another solution.

src/style/style.ts
```diff
@@ -1973,4 +1973,90 @@ export class Style extends Evented {
...
+        // Destroy glyphManager 
+        if (this.glyphManager) {
+            this.glyphManager.destroy();
+        }
+
+        // Remove layers
+        for (const layerId in this._layers) {
+            const layer = this._layers[layerId];
+            layer.setEventedParent(null);
+            if (layer.onRemove) layer.onRemove(this.map);
```
HarelM, line 2075: use `?.` maybe?

HarelM: Oh, wait, I'm not sure it's possible... :-/

ToHold:
> BTW, what about customer layers? I don't think they withstand context lost, do they?

They need to be restored either from the contextrestore event fired by the map or by listening to the style.load event, which is triggered when the style is restored.
Would you like me to propose another solution?

src/style/style.ts
```diff
@@ -1973,4 +1973,90 @@ export class Style extends Evented {
...
+        this._updatedPaintProps = {};
+        this._layerOrderChanged = false;
+        this.crossTileSymbolIndex = new (this.crossTileSymbolIndex?.constructor || Object)();
+        this.pauseablePlacement = undefined;
+        this.placement = undefined;
+        this.z = 0;
+
+        // Remove event listeners
+        this.setEventedParent(null);
+        this.dispatcher.remove();
```
HarelM: What about unregistering to events from workers? Also what about cleaning workers data etc?

HarelM:
I've added a few comments.

src/render/image_manager.ts
ToHold: > BTW, since it is used outside this class, it probably shouldn't be prefixed with "_"...

Oh, I thought it was meant to indicate that it’s for internal use only and shouldn’t be used by end users.
I'll update it too.

src/style/style.ts
ToHold: I’m not very familiar with this part of the code yet; I need to study it more to fully understand what you’re suggesting.

ToHold: Were you thinking about something like this ?

```TS
// Clear workers 
this.dispatcher.workerPool.release(this.map._mapId);
this.dispatcher.actors.forEach((actor) => {
    actor.remove();
});
this.dispatcher.workerPool.workers.forEach((worker) => {
    worker.terminate();
});
```

src/render/image_manager.ts
ToHold: Maybe it would be cleaner to pause rendering (e.g., this._repaint) on context loss, copy what we need, then destroy and fully recreate everything on restore. That way we wouldn’t need to store extra properties on the map, and this._lostContextStyle and this._lostContextImages wouldn’t be necessary.

HarelM: I'm not sure it's possible given the event nature of this, but if this is possible then it can be a better approach.

ToHold: I can't delete painter and style from the context lost event without GL warnings because they belong to the old context, but I tried a cleaner way to restore them.

src/style/style.ts
ToHold: I tried `layer?.onRemove(this.map);` but yeah it didn't work

src/style/style.ts
HarelM: The dispatcher is registering events in style.ts file:
https://github.com/maplibre/maplibre-gl-js/blob/5e0f8f749eab9bb5e0ea7e0a34a14112b52f0a66/src/style/style.ts#L246

I'm guessing these need to be unregistered to avoid receiving messages for a destroyed style.

HarelM: Regarding workers, these are created when the app starts, but I'm not sure how stateless they are, so I would recommend looking at what is stored in worker.ts file and see if clean up is needed there as well:
https://github.com/maplibre/maplibre-gl-js/blob/main/src/source/worker.ts

src/style/style.ts
```diff
@@ -1986,4 +1986,96 @@ export class Style extends Evented {
...
+            const layer = this._layers[layerId];
+            layer.setEventedParent(null);
+            if (layer.onRemove) layer.onRemove(this.map);
+        }
+        this._layers = {};
+        this._order = [];
+        this._serializedLayers = {};
+
+        // Reset other internal state
+        this.stylesheet = null;
```
HarelM: Consider moving all this initialization logic to a method and maybe use the same method in the constructor.

src/ui/map.ts
```diff
@@ -445,6 +445,14 @@ const defaultOptions: Readonly<Partial<MapOptions>> = {
     centerClampedToGround: true
 };
 
+let oldContextStyle: {
```
HarelM: I don't think we should define a global variable... I'm guessing that this won't work when there are multiple maps in the same page...

HarelM:
> They need to be restored either from the contextrestore event fired by the map or by listening to the style.load event, which is triggered when the style is restored.
Would you like me to propose another solution?

If we this is not handled automatically at least a `warnOnce` should be used to let a developer seeing this to know that they need to handle this problem.
If there's a way to actually solve this without any interaction from the developer it might be a better DX.

In what scenario this happens that all this code is needed? I'm a bit lost around the motivation of this PR, that slowly becomes bigger and bigger...

src/style/style.ts
ToHold: I added an unregisterMessageHandler method to the dispatcher, and also refactored that part since what I was doing was already handled when passing true to dispatcher.remove.

In all my tests, the workers didn’t seem to cause any issues.

src/ui/map.ts
ToHold: Honestly, I was pretty tired the day I added that, but now I don’t see any other option than adding a property to the Map class.

ToHold:
> > They need to be restored either from the contextrestore event fired by the map or by listening to the style.load event, which is triggered when the style is restored.
> > Would you like me to propose another solution?
> 
> If we this is not handled automatically at least a `warnOnce` should be used to let a developer seeing this to know that they need to handle this problem. If there's a way to actually solve this without any interaction from the developer it might be a better DX.
> 
> In what scenario this happens that all this code is needed? I'm a bit lost around the motivation of this PR, that slowly becomes bigger and bigger...

Sorry for the delay in getting back to this PR. I was actually working on the restore part of the app I’m developing.

This issue happens quite frequently in the app. As soon as the user has multiple tabs open, comes back to the page after some time, or when multiple WebGL contexts are used, the browser tends to “pause” the context, which leads to a context loss, and then restores it once the context becomes visible again.

To give an example, this issue occurs on both mobile and desktop, but it’s more frequent on mobile devices since the limit for simultaneous WebGL canvases is lower. As soon as the map stops rendering for any reason, the browser pauses the context.

In the app, I only create one canvas for the map, but some libraries may internally create additional ones. When users open multiple tabs containing the map, the problem is almost guaranteed to occur.

For reference, around 3,500 users encounter this error every month. Since the map is used in a professional context and is the central element of our application, this issue is quite problematic.

For a bit more context: I’m working on a 3D generation app that renders buildings, trees, water, etc. I’ve been focusing lately on handling the restore logic in Three.js, and this part is now ok.

Honestly, even though this PR has grown quite a bit, there’s currently nothing functional in place to handle the restore.
In my opinion, even if it’s not perfect yet, it’s still better than nothing; right now the page just goes blank with no error message or feedback at all.

I’d really appreciate it if this PR could be accepted, it would be a huge help and take a big weight off my shoulders.
Unfortunately, I can’t fully fix this issue from my app’s side alone without making changes in MapLibre itself.

ToHold:
> > They need to be restored either from the contextrestore event fired by the map or by listening to the style.load event, which is triggered when the style is restored.
> > Would you like me to propose another solution?
> 
> If we this is not handled automatically at least a `warnOnce` should be used to let a developer seeing this to know that they need to handle this problem. If there's a way to actually solve this without any interaction from the developer it might be a better DX.
> 
> In what scenario this happens that all this code is needed? I'm a bit lost around the motivation of this PR, that slowly becomes bigger and bigger...

I can either add another method instead of using serialize to keep the layers, that way I could handle the custom layers properly, or I can simply add a warning if you’d prefer to avoid making the PR any bigger.

I’m fine with either approach, whichever you prefer.

ToHold:
Please don’t hesitate to share any concerns or be as picky as needed. I’ll do whatever it takes to address this issue, as it’s been quite problematic for me.

HarelM:
I see, thanks for the context, that makes a lot of sense. 
Can you verify that this change will actually solve your issue? 
You can use patch-package to verify this if need be. 

A warning about the custom layer is enough for now, if someone would need a more complicated solution they are welcome to contribute it.

I'll do a final review in the next few days, sorry for the delay in my response...

src/ui/map.ts
```diff
@@ -545,6 +545,18 @@ export class Map extends Camera {
      */
     _imageQueueHandle: number;
 
+    /**
+     * @internal
+     * Used to store the previous style and images when a context loss occurs, so they can be restored.
+     */
+    _oldContextStyle: {
+        style: StyleSpecification | null;
```
HarelM: nit picking: you might want to define a type for this as is being repeated a bit.

HarelM:
Two PRs were merged today with some minor changes, mainly to sourceCache -> tileManager rename, and a feature in the glyph manager, it should be an easy conflict resolution, I hope.

ToHold:
Oops I failed my git rebase on main, i'm fixing it

ToHold:
Everything should be OK, I'll just test a bit and tell you when I'm sure everything is ok

ToHold:
Everything seems to be OK ! :smile:

src/style/style.ts
```diff
@@ -261,15 +261,7 @@ export class Style extends Evented {
...
-        this._spritesImagesIds = {};
-        this._layers = {};
-
-        this._order = [];
-        this.tileManagers = {};
-        this.zoomHistory = new ZoomHistory();
-        this._loaded = false;
-        this._availableImages = [];
-        this._globalState = {};
+        this._setInitialValues();
```
HarelM, line 264: The content of this method is a lot longer than what was here before, can you elaborate on why?

src/ui/map.ts
```diff
@@ -545,6 +550,15 @@ export class Map extends Camera {
      */
     _imageQueueHandle: number;
 
+    /**
+     * @internal
+     * Used to store the previous style and images when a context loss occurs, so they can be restored.
+     */
+    _oldContextStyle: OldContextStyle = {
```
HarelM: nit-picking: `_previous...` maybe instead of `old`?
or `_lostContextStyle` so that it will tell explicitly what it is used for?

src/ui/map.ts
```diff
@@ -387,6 +387,11 @@ type DelegatedListener = {
 
 type Delegate<E extends Event = Event> = (e: E) => void;
 
+type OldContextStyle = {
```
HarelM: Same as the other comment about naming.

src/style/style.ts
```diff
@@ -1997,4 +2019,72 @@ export class Style extends Evented {
...
+
+        // Destroy glyphManager
+        if (this.glyphManager) {
+            this.glyphManager.destroy();
+        }
+
+        // Remove layers
+        for (const layerId in this._layers) {
+            const layer = this._layers[layerId];
+            layer.setEventedParent(null);
```
HarelM, line 2074: Does this mean that if I have a listener on a layer (like layer click listener) when the context is lost, this listener will also be lost, right?
Maybe we should add a warning about that as well?

src/style/style.ts
```diff
@@ -1997,4 +2019,72 @@ export class Style extends Evented {
...
+            if (layer.onRemove) layer.onRemove(this.map);
+        }
+
+        // reset internal state
+        this._setInitialValues();
+
+        // Remove event listeners
+        this.setEventedParent(null);
```
HarelM, line 2082: Same as above for event listeners for style...?

HarelM:
It's hard to think about all the edge cases of this, but everything you did here is probably a lot better than how it is now.
I've added a few last comments before merging this.

src/style/style.ts
ToHold: It was to make sure these properties aren’t set when a context loss occurs, basically a way for me to fully reset everything back to the initial state.

src/style/style.ts
ToHold: You’re right, for now I’ve chosen to go with a warning.
I can revisit this and work on a more complete solution in a follow-up PR.

src/style/style.ts
melitele: `layer.onRemove?.(this.map)`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining#syntax

HarelM: Feel free to submbit a PR to fix this, I'm pretty sure there are other places that is doing this type of check, but we can gradually remove them as we go.

ToHold:
I found a case where it can still render a blank map, it's more specific than before. If a resize canvas event is fired during / before the restore event, it'll render a blank app. It happens in my app, less than before, but still. I'll find  the cause and propose a fix as soon as possible.

ToHold:
> I found a case where it can still render a blank map, it's more specific than before. If a resize canvas event is fired during / before the restore event, it'll render a blank app. It happens in my app, less than before, but still. I'll find the cause and propose a fix as soon as possible.

I found a great fix, I just need to create a test and I'll create a PR for it.

PR #6399, opened 2025-09-12
Fixes case where retain loaded children does not retain uppermost loaded children

### LINKED ISSUE #5969: Raster tiles flicker issue when pitching above 60 degrees
**maplibre-gl-js version**: 5.5.0

**browser**: Google Chrome

### Steps to Trigger Behavior

 1. Create a map with `maxPitch` set to something higher than `60`
 2. Pitch above 60 degrees and rotate the map
 3. Observe flickering and missing tiles

### Link to Demonstration

https://jsbin.com/hayisov/edit?html,output

Video demonstrates issue, where white empty spaces appear in the raster tileset for a short amount of time.

https://github.com/user-attachments/assets/0b16dcac-76da-44cf-894b-e2d076f3d33b

wayofthefuture:
This PR fixes #5969.

The newer version of _retainLoadedChildren does not retain all loaded children (nor does the prior accepted version). This causes flickering of the map when zooming out especially at higher pitches.

This change updates _retainLoadedChildren to correctly handle pitched maps. Instead of using a fixed maxCoveringZoom (which can skip uppermost descendants in distant, high-pitch areas), we compute a maxCoveringZoom relative to each ideal tile’s overscaledZ. This makes the max-covering plane parallel to the ideal-tile plane, ensuring we always retain the correct uppermost loaded descendants across the pitched view.

 - [X] Link to related issues.

Before:

https://github.com/user-attachments/assets/33106e34-3cb1-4f35-b0cc-985a86d682de

After:

https://github.com/user-attachments/assets/8f0d4891-7f96-4d36-94c5-fad96af8099e


wayofthefuture:
@HarelM Which method do you prefer for enumerating dictionaries and filtering arrays? Readability in Object.values() / Object.entries() Array.filter() vs for (const key in dict) for (const item of arr)... The helpers decrease code length and improve readability but I've heard they are theoretically less performant...

HarelM:
I favor readability usually. Can you link to a point where you were not sure which is preferred?

wayofthefuture:
Not quite there yet. I have solved the field of view issue but the fading logic is highly conflicting and the tests won't allow proper retainment... from my understanding fading doesn't work currently...

src/source/tile_id.test.ts
```diff
@@ -98,12 +103,42 @@ describe('OverscaledTileID', () => {
...
+        // not a child at a different wrap
         expect(
             new OverscaledTileID(2, 0, 2, 0, 0).isChildOf(new OverscaledTileID(0, 1, 0, 0, 0))
         ).toBeFalsy();
     });
 
+    test('.isChildOf - root tile should not be child of itself', () => {
+        const root = new OverscaledTileID(0, 0, 0, 0, 0);
+        expect(root.isChildOf(root)).toBe(false);
+    });
```
wayofthefuture, line 119: @HarelM 😕

wayofthefuture:
@HarelM I uploaded before/after videos above. The isChildOf bug was quite discouraging and returned true for the root tile, which caused tests to continually fail. The field of view test had a race condition with tile fading logic so I had to modify that test setup slightly. Also, the tile fading logic has some major issues and needs to be improved (but separately as a different PR). I may work on that next if I can find some time. I have added additional tests for pitch, retain loaded children, and is child of. The visual difference is quite drastic and I believe this to be a good step towards full support for pitch up to 89.

src/util/test/util.ts
```diff
@@ -278,4 +279,17 @@ export function createTestCameraFrustum(fovy: number, aspectRatio: number, zNear
     mat4.invert(invProj, proj);
 
     return Frustum.fromInvProjectionMatrix(invProj, 1.0, 0.0);
-};
+}
+
+// export function logTiles(tileIDs: OverscaledTileID[]) {
```
HarelM: This is probably a leftover...?

src/util/test/util.ts
```diff
@@ -10,6 +10,7 @@ import {RequestManager} from '../request_manager';
 import {type IReadonlyTransform, type ITransform} from '../../geo/transform_interface';
 import {type Style} from '../../style/style';
 import {type Terrain} from '../../render/terrain';
+import {type OverscaledTileID} from '../../source/tile_id';
```
HarelM: This has a lint warning...

src/source/tile_id.test.ts
```diff
@@ -98,12 +103,42 @@ describe('OverscaledTileID', () => {
     });
 
     test('.isChildOf', () => {
+        // simple child of root tile
```
HarelM: Move the comments to the test description and split the tests if needed.

src/source/tile_id.test.ts
```diff
@@ -68,6 +68,11 @@ describe('OverscaledTileID', () => {
             /*eslint no-new: 0*/
             new OverscaledTileID(7, 0, 8, 0, 0);
         }).toThrow();
+        expect(() => {
+            //tile with deeper canonicalZ than overscaledZ
```
HarelM: Consider moving this comment to be a different test.

src/source/tile_id.test.ts
```diff
@@ -98,12 +103,42 @@ describe('OverscaledTileID', () => {
...
+        expect(child.isChildOf(parent)).toBe(false);
+    });
+
+    test('.isChildOf - descendant is child of ancestor', () => {
+        const parent = new OverscaledTileID(1, 0, 1, 0, 0);
+        const child  = new OverscaledTileID(4, 0, 4, 3, 5);
+        expect(child.isChildOf(parent)).toBe(true);
+    });
+
+    test('.isChildOf - descendant with different coordinates is not child of ancestor', () => {
```
HarelM, line 129: These tests are great! Good job of adding those!

src/source/source_cache.ts
```diff
@@ -729,7 +750,7 @@ export class SourceCache extends Evented {
         }
 
         // retain any loaded children of ideal tiles up to maxCoveringZoom
-        this._retainLoadedChildren(missingTiles, zoom, maxCoveringZoom, retain);
+        this._retainLoadedChildren(missingTiles, retain);
```
HarelM, line 759: Looks like the method parameters have changed, does the comment above it still relevant?
I would consider moving the comment to be a tsdoc comment of the method instead of adding a comment above its usage...

src/source/source_cache.ts
```diff
@@ -358,50 +358,75 @@ export class SourceCache extends Evented {
...
+     *
+     * To solve this, we calculate the max covering zoom for each ideal tile separately using its
+     * `overscaledZ`. This effectively makes the "max covering zoom plane" parallel to the
+     * "ideal tile plane," ensuring that we correctly capture the uppermost children
+     * of each ideal tile across the pitched view.
+     *
+     * Analogy: imagine two sheets of paper in 3D space:
+     *   - one sheet = ideal tiles at varying overscaledZ
+     *   - the second sheet = maxCoveringZoom
+     * With the old approach, the sheets intersected. With the below method, they remain parallel.
```
HarelM: I don't think there's a need to mention the old approach here, it would make sense to talk about it in the PR, but this comment should only reflect what the method is currently doing.
BTW, this is a great summary of the problem and the implementation!

src/source/source_cache.ts
```diff
@@ -358,50 +358,75 @@ export class SourceCache extends Evented {
...
-
-                if (tile && tile.hasData()) {
-                    topmostLoadedID = parentID;
+        //convert target tiles dictionary to array for code readability and quicker performance below
+        const targetTileIDs = Object.values(targetTiles);
+
+        // create dictionary for the qualified loaded descendents of each target tile
+        const loadedDescendents: { [_: string]: Tile[] } = {};
+
+        // enumerate tiles currently in this source and find the loaded descendents of each target tile
```
HarelM: This method is well written with good comments, but I would like to see less comments and maybe more small methods with the relevant comment as a method name? Or maybe names that are descriptive enough to avoid some of the comments here, as comments test to get stale over time.

src/source/source_cache.test.ts
```diff
@@ -938,31 +939,128 @@ describe('SourceCache._updateRetainedTiles', () => {
...
             idealTile
-        ].concat(loadedChildren).map(t => t.key).sort());
+        ];
+
+        const retained = sourceCache._updateRetainedTiles([idealTile], 3);
+        expect(Object.keys(retained).sort()).toEqual(expectedTiles.map(t => t.key).sort());
+    });
+
+    test('retains loaded children for pitched maps', async () => {
+        testPitch(0);
```
HarelM: You can create an array and run the same test several times, this will allow the test harness to report the right failure in case one of these tests faile, i.e.

```ts
for (let pitch of [0, 20, 40, ...]) {
     test('retains loaded children for pitch: ' + pitch, () => {
       ...
```

src/source/source_cache.test.ts
```diff
@@ -938,31 +939,128 @@ describe('SourceCache._updateRetainedTiles', () => {
...
+            new OverscaledTileID(4, 0, 4, 4, 4),  //discard
+            new OverscaledTileID(5, 0, 5, 8, 8),  //discard
+        ];
+        for (const child of children) {
+            const tile = new Tile(child, undefined);
+            tile.state = 'loaded';
+            sourceCache._tiles[child.key] = tile;
+        }
+
+        // create retainment dictionary to pass by reference to _retainLoadedChildren for modification
```
HarelM: I'm not sure this comment is useful, it's part of the parameters of the method, you can use the tsdoc to write this down I believe.

src/source/source_cache.test.ts
```diff
@@ -938,31 +939,128 @@ describe('SourceCache._updateRetainedTiles', () => {
...
+            }
+
+            // create retainment dictionary to pass by reference to _retainLoadedChildren for modification
+            const retain: {[key: string]: OverscaledTileID} = {};
+
+            // create missing dictionary for ideal tiles for passing into _retainLoadedChildren
+            const missingTiles: {[key: string]: OverscaledTileID} = {};
+
+            // mark all ideal tiles as retained and also as missing with no data for child retainment
+            idealTileIDs.forEach(idealID => {
```
HarelM: nit picking: can we use regular `for (const a of b)`?

src/source/source_cache.test.ts
```diff
@@ -938,31 +939,128 @@ describe('SourceCache._updateRetainedTiles', () => {
...
+                retain[idealID.key] = idealID;
+                missingTiles[idealID.key] = idealID;
+            });
+
+            // retain loaded children for the missing ideal tiles
+            sourceCache._retainLoadedChildren(missingTiles, retain);
+
+            expect(Object.keys(retain).sort()).toEqual(
```
HarelM: This there a collocation assert that can be used instead of sorting etc every time? If not, this is good enough...

src/source/source_cache.test.ts
```diff
@@ -938,31 +939,128 @@ describe('SourceCache._updateRetainedTiles', () => {
...
+    test('retains only uppermost zoom children when multiple zoom levels are loaded', () => {
+        const sourceCache = createSourceCache();
+        sourceCache._source.loadTile = async (tile) => {
+            tile.state = 'errored';
+        };
+
+        const idealTileID = new OverscaledTileID(2, 0, 2, 1, 1);
+        const idealTiles: {[key: string]: OverscaledTileID} = {[idealTileID.key]: idealTileID};
+
+        // add children at z=3, z=4, and z=5
```
HarelM: This method is redundant as the zoom can be seen in the below list.

HarelM:
I think that you did a great job here! Thanks!
Also it's evident that all the lines you changed are covered with test, which is not an easy task.
I've added some minor comments, mostly code style related comments.
Can you run `npm run lint -- --fix` to solve the lint issues?

src/util/test/util.ts
wayofthefuture: I wrote it as it was very helpful for creating tests. The linter complained about it being unused. Guess I will remove...

src/source/source_cache.ts
wayofthefuture: Lol and proving your point that comments become stale over time 😄

src/source/source_cache.test.ts
wayofthefuture: All the other tests were using so I don't think so...

wayofthefuture:
Ok hopefully that did it...

HarelM:
Forgot to ask, are all the render tests really needed? feels a bit out of scope for render tests as this is something that happens when you zoom in and the render tests are more of a static kind of test, at least this is how I see it.

HarelM (approved):
Looks great! Thanks.
Let me know what you decide about the render tests.

HarelM:
Another point worth mentioning is that we have some future concept around making the render tests have similar results for both native and gl-js, so that's another thing to take into consideration regarding render tests...

wayofthefuture:
Right, I see what you mean. I could remove all except for maybe pitch 65... or just remove all... pitch 65 provides some protection for the covering plane logic.

HarelM:
I'm good with removing them and making sure the unit tests cover the logic changes, which I think they do.

wayofthefuture:
Done!

HarelM:
Let me know how urgent this is for you, I usually release a version at least once a month.

wayofthefuture:
Not urgent...

PR #6421, opened 2025-09-16
move from npm to pnpm

birkskyum:
As a measure to avoid supply chain attacks, we can move to pnpm and only allow installation of packages that are more than 24h old, with exception of our whitelist

https://pnpm.io/supply-chain-security

We see increased attempts at tricking our systems with supply chain attacks, and pnpm allow us more granularly to opt into dangerous postinstall scripts too

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


pnpm-workspace.yaml
```diff
@@ -0,0 +1,11 @@
+minimumReleaseAge: 1440
```
CommanderStorm: Should we give ourselves a bit more time (i.e. a weekend => 2.5 days)

Something to be decided by @HarelM

birkskyum: Sure, I've updated it to 2.5 days, we can always whitelist things we'd like to expedite

HarelM:
Sorry, I don't think that's a good enough reason

We can configure the auto merge to wait a day if needed instead of complicating developers setup.

louwers:
https://github.blog/changelog/2025-07-01-dependabot-supports-configuration-of-a-minimum-package-age/

birkskyum:
That might do the trick as well. I can´t wrap my head around if it gives the same safety guarantee (I think it won´t be as safe as configuring the package manager), but i think just having one step in the pipeline that delay updates just a bit would be good.

CommanderStorm:
But dependabot is not the only way to change the lockfile. PRs can also change the lockfile.

Ideally, having both would likely be best if we realy want to be safe.
If we are ok with this vector, that is fine too. We just have to be aware of it.

louwers:
Maybe npm will implement this as well with the supply chain attacks left and right.  This issue has some new activity. https://github.com/npm/rfcs/issues/646

@CommanderStorm How often do PRs add new dependencies? 

I still think it's not addressing the root cause of the problem though.

birkskyum:
in some cases where we help dependabot, like when it can´t figure how to update two packages at the same time, or handle changes associated to major bumps, a manual intervention to the pr is made - but those wont be as fast as the autoomerge anyway

louwers:
I think it makes sense to vet any change in dependencies, either manually or more realistically, automate it.

Maybe try out socket.dev https://socket.dev/compare/socket-vs-dependabot It does static analysis on dependencies and I think it is even free for open source.

CommanderStorm:
Seems worth a try.
I can install it if @HarelM agrees that we should try it.

HarelM:
I don't mind trying it out, if it creates too much noise we can always turn it off. 
Does it replace dependabot or just helps it, as I would like to keep dependabot since it's a mature and battle tested tool.

louwers:
I don't think it replaces dependabot. It just scans PRs with dependency changes.

CommanderStorm:
Ok done, we will see what pops up

PR #6436, opened 2025-09-18
Switch `MapEventType` from type to interface to allow declaration merging

lhapaipai:
## Summary

This PR changes `MapEventType` from a type alias to an interface, which enables TypeScript declaration merging

## Motivation

Currently, `MapEventType` is defined as a type. This prevents developers from extending the event map.

By switching it to an interface, developers can easily add their own custom events using module augmentation:

```ts
// ./app/env/maplibre.d.ts
import 'maplibre-gl';
import type { MyCustomEvent } from './types';

declare module 'maplibre-gl' {
  export interface MapEventType {
    'my-lib:custom': MyCustomEvent;
  }
}
```

## Benefits

- Enables type-safe custom events for projects integrating with MapLibre.
- No breaking change for existing users.

## Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
What's preventing from merging a type to a type instead of an interface?
In general, I wanted to make sure that interfaces of the public API are those that you need to inherit in order to implement and use the public API, I'm not sure this is the case for the event types...?
I'll need to better understand your use case I guess... Please elaborate on what can't be done with current public API.

lhapaipai:
Actually the `on` method is declared like this:

- when listening for an event whose name is a key of `MapEventType`, the listener is strongly typed: `(ev: MapEventType[T] & Object) => void` 


event-typing

- When listening for a custom event **not** defined in `MapEventType` the listener falls back to the generic type `Listener`: : `(a: any) => any`

custom-event-any


```ts
export class Map extends Camera {
    /**
     * Overload of the `on` method that allows to listen to events without specifying a layer.
     * @event
     * @param type - The type of the event.
     * @param listener - The listener callback.
     */
    on<T extends keyof MapEventType>(type: T, listener: (ev: MapEventType[T] & Object) => void): Subscription;
    /**
     * Overload of the `on` method that allows to listen to events without specifying a layer.
     * @event
     * @param type - The type of the event.
     * @param listener - The listener callback.
     */
    on(type: keyof MapEventType | string, listener: Listener): Subscription;
    on(type: keyof MapEventType | string, layerIdsOrListener: string | string[] | Listener, listener?: Listener): Subscription {
      // ...
    }
}
```

For MapLibre plugins that rely on `map.fire` to trigger library-specific events, it would be useful to allow strongly typed listeners for these custom events.

Exemple with [maplibre-geoman](https://geoman.io/docs/maplibre/events#event-listening-methods).

```ts
map.on('gm:create', (event) => {
  // Actually event is typed as `any`
  console.log('Feature created:', event);
});
```

Strong typing here would improve type safety and provide better guidance for new users.

Unfortunately, this is not possible today because `MapEventType` is declared as a **type alias** instead of an **interface**. Unlike interfaces, type aliases cannot be extended or merged. See [Differences between type Alises and Interfaces](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces).

Currently, `MapEventType` is defined once in `src/ui/events.ts` and cannot be re-opened.

By changing `MapEventType` from a **type** to an **interface**, plugin developers and end-users would be able to extend it seamlessly.


Exemple of Source code of MapLibre plugin like `maplibre-geoman` 

```ts
// ./env.d.ts
import 'maplibre-gl';
import type { FeatureCreatedFwdEvent } from './types';

declare module 'maplibre-gl' {
  export interface MapEventType {
    'gm:create': (e: FeatureCreatedFwdEvent) => void;
  }
}
```

End-users would simply reference this file in their project (a file like `src/env.d.ts`)

```ts
/// <reference types="@geoman-io/maplibre-geoman-free/env" />
```

With that, they automatically benefit from proper type checking when using the plugin.

event-typing-gm



This approach is not limited to plugin authors — it could also help developers who emit their own custom events on the map.

## Note

This pattern is already used in other libraries. A common example is the DOM event system:

```ts
// defined in lib.dom.d.ts
interface WindowEventMap extends GlobalEventHandlersEventMap, WindowEventHandlersEventMap {
    "DOMContentLoaded": Event;
    "devicemotion": DeviceMotionEvent;
    "deviceorientation": DeviceOrientationEvent;
    "deviceorientationabsolute": DeviceOrientationEvent;
    "gamepadconnected": GamepadEvent;
    "gamepaddisconnected": GamepadEvent;
    "orientationchange": Event;
}
```

```ts
// can be extended in local project
declare global {
  interface WindowEventMap {
    "my-custom-event": MyCustomEvent;
  }
}


window.addEventListener("my-custom-event", (event) => {
  event //  is correctly typed as MyCustomEvent
})

HarelM:
Thanks for the clarification! 
Can you add a chancelog entry for this?

lhapaipai:
Thanks ! I've updated the changelog.

HarelM (approved):

PR #6437, opened 2025-09-18
chore(security): remove unused permissions

CommanderStorm:
We are currently not using these permissions.
Lets remove them.

THis also seems to be a [best practice](https://docs.zizmor.sh/audits/#artipacked), but that is secondary to this PR.

When removing them, I was conservative, there are other occurrences where we might be able to remove them. This is the larger part that we can just remove without deeply looking at.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -40,7 +40,6 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
       with:
-        fetch-depth: 0
         persist-credentials: false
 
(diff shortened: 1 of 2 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -39,6 +39,9 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      with:
+        fetch-depth: 0
```
HarelM: Didn't @louwers just removed those?

CommanderStorm: Yes, sorry. That was a mistake

HarelM (approved):

PR #6442, opened 2025-09-19
Fix issue where parent tiles are retained when deeper descendant tiles already cover the missing ideal tile.

wayofthefuture:
**Summary**
This PR fixes an issue where parent tiles were being retained/requested even when deeper descendant tiles already covered the missing ideal tile.

**Background**
Previously, _updateRetainedTiles only checked one level of children before falling back to parent tiles. Meanwhile, _retainLoadedChildren correctly seeks descendants up to maxCoveringZoom (e.g. 4, 16, 64 tiles for successive generations). This mismatch causes unnecessary parent retainment.

**Change**
Instead of duplicating multi-generation logic in _updateRetainedTiles, the function now relies on _retainLoadedChildren to report ideal tiles with incomplete descendants. This keeps the logic consistent and prevents redundant parent retainment/requests.

### DIFF
--- a/src/source/source_cache.test.ts
+++ b/src/source/source_cache.test.ts
@@ -971,7 +971,7 @@ describe('SourceCache._updateRetainedTiles', () => {
         expect(Object.keys(retained).sort()).toEqual(expectedTiles.map(t => t.key).sort());
     });
 
-    test('_updateRetainedTiles incorrectly retains/requests parents when 2nd generation children are loaded', () => {
+    test('_updateRetainedTiles does not retain parents when 2nd generation children are loaded', () => {
         const sourceCache = createSourceCache();
         sourceCache._source.loadTile = async (tile) => {
             tile.state = 'errored';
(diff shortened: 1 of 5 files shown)


src/source/source_cache.ts
```diff
@@ -434,6 +440,23 @@ export class SourceCache extends Evented {
         return loadedDescendents;
     }
 
+    /**
+     * Determine if tile ids fully cover the current generation.
+     * - 1st generation: need 4 children or 1 overscaled child
+     * - 2nd generation: need 16 children or 1 overscaled child
+     */
+    _areDescendentsComplete(generationIDs: OverscaledTileID[], generationZ: number, ancestorZ: number) {
+        if (generationZ <= ancestorZ) throw new Error('provided generation requires higher Z than ancestor');
```
HarelM, line 449: Do we really want to throw here? What about warnOnce or something?

src/source/source_cache.test.ts
```diff
@@ -971,6 +971,31 @@ describe('SourceCache._updateRetainedTiles', () => {
         expect(Object.keys(retained).sort()).toEqual(expectedTiles.map(t => t.key).sort());
     });
 
+    test('_updateRetainedTiles incorrectly retains/requests parents when 2nd generation children are loaded', () => {
```
HarelM, line 974: Is this test correct?

wayofthefuture: I wrote it as a failing test it should probably be titled something like "...does not retain parents if 2nd generation children are loaded"

src/source/source_cache.ts
wayofthefuture: It could warn and return false but the false would not necessarily mean anything (could cause weird bugs?). Something would be very wrong with the code if it were to be called in this manner. So I don't know...

HarelM: If this is a safe guard against something that we don't know how to reproduce or believe can truly happen I suggest to remove this code...

src/source/source_cache.test.ts
HarelM: Can you change the test name then, please?

HarelM:
Looks good otherwise, thanks again for digging into this!

wayofthefuture:
I forgot to remove the throw test then mispelled the commit! Sorry for the force pushes. Otherwise, all done!

PR #6445, opened 2025-09-20
chore(security): add release attestation for `dist.zip`

CommanderStorm:
This PR adds release attestation for the `dist.zip` we attach to our releases

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,6 +35,10 @@ jobs:
     needs: release-check
     if: ${{ needs.release-check.outputs.publish == 'true' }}
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      id-token: write
+      attestations: write
     defaults:
       run:
         shell: bash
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -35,6 +35,10 @@ jobs:
     needs: release-check
     if: ${{ needs.release-check.outputs.publish == 'true' }}
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
```
CommanderStorm, line 39: Nessesary to be `write` because we push to the wiki further down among others.

.github/workflows/release.yml
```diff
@@ -111,7 +119,7 @@ jobs:
           bodyFile: ${{ steps.release_notes.outputs.release_notes }}
           artifacts: "dist.zip"
           artifactContentType: "application/zip"
-          allowUpdates: true
+          allowUpdates: false
```
CommanderStorm, line 122: This will fail if we enable immutable releases for this repo.

CommanderStorm:
Attestations are listed here https://github.com/maplibre/maplibre-gl-js/attestations

The attestation I added is basically "I have dist.zip, is it the one from maplibre?".
There are different kinds, for example we might want to add one for our SBOM. (there is apparently tooling which works better with this in place)

PR #6446, opened 2025-09-21
Fix raster flickering when using terrain 3D and optimize terrain logic.

wayofthefuture:
This PR fixes flickering in the raster basemap when used with 3D terrain.

Previously, tile-fade logic was applied to raster rendering. The 3D terrain implementation, which relies on a raster basemap, then attempted to undo this fading since terrain does not currently support it.

Because we can already determine whether terrain is present after creating ideal tiles for a raster basemap, there’s no need to apply fading in the first place. This change avoids applying the fading logic entirely for terrain, eliminating the need to later reverse it.

Before:
https://github.com/user-attachments/assets/1b41c21f-bab3-4f8f-805e-2fc8ab7581d7

After:
https://github.com/user-attachments/assets/c3f22429-ca46-410f-9b80-0224bc84a419

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
+- Fix raster flickering when using terrain 3D and optimize terrain logic.
 - Fix issue where parent tiles are retained when deeper descendant tiles already cover the missing ideal tile. ([#6442](https://github.com/maplibre/maplibre-gl-js/pull/6442))
 
 ## 5.7.3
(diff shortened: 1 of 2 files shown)


wayofthefuture:
I mainly refactored by shifting existing logic and removing an unnecessary part imo. There were also obvious problems in the logic I removed. Other than that, I tried not to change any core logic, but needed to abstract the fading logic. I'm currently working on a complete fix for fading which I can't do without abstracting it first. That'll be my next PR, if I can fix it hopefully, but then after that I have to get back to the real world so it should be my last for a while!

Issue #6316, opened 2025-08-28
defaultProjectionData.fallbackMatrix and .mainMatrix do not reflect the correct types


bigmistqke:
I am currently in the process of making the MapLibre-project at my job functional in earlier browser versions. In our project we have a CustomLayer, in which I am passing the `defaultProjectionData` from its render-function to a glsl shader.

The `fallbackMatrix` and `mainMatrix` are typed as `mat4`, implying they are `Float32Array`, but in reality they are `Float64Array` being cast to `any` to circumvent the types of `gl-matrix` (see https://github.com/maplibre/maplibre-gl-js/issues/326#issuecomment-924249326).

Currently `gl.uniformMatrix4fv` accepts any `Iterable` returning numbers, but chrome <68 only accepted `Float32Array` or `Array<number>`, causing the CustomLayer to fail when trying to set the uniform.

as https://github.com/maplibre/maplibre-gl-js/issues/326#issuecomment-924249326 mentioned:

> It might be good to consider making a 64bit version/extend glmatrix to play better with Typescript for things like this..

which I agree with. It would be cool if `gl-matrix` had `mat<Float64Array>`! Else we could maybe add type annotations in the jsdoc?

The `fallbackMatrix` and `mainMatrix` are only 2 examples of this cast, it happens at multiple occasions throughout the codebase.

HarelM:
Feel free to improve typings.
Since gl-matrix doesn't support this genetics, as far as I know, we use variable names to indicate if this is a 32 or 64.
You could also see if it's possible to open an issue for gl-matrix repo to add support for that...

bigmistqke:
Hi HareIM

Finally had some time today to have a look at it.
I made an initial draft PR: https://github.com/maplibre/maplibre-gl-js/pull/6460

HarelM:
Thanks @bigmistqke 
I added a few comments. 
I'd also be interested in seeing what the gl-matrix maintainers has to say, maybe just presenting them with the problem statement instead of the proposed solution ti get an unbiased opinion.

bigmistqke:
> I'd also be interested in seeing what the gl-matrix maintainers has to say, maybe just presenting them with the problem statement instead of the proposed solution ti get an unbiased opinion.

Good idea, I will make an issue at their repo.

PR #6460, opened 2025-09-25
feat: enhance gl-matrix types

bigmistqke:
Continuing from issue https://github.com/maplibre/maplibre-gl-js/issues/6316 in which I proposed for `MapLibre`'s `gl-matrix` types to reflect more their actual types.  The original usecase for this type enhancement was a bug in an earlier version of chrome where a `Float64Array` from `MapLibre` that was typed as a `Float32Array` was passed to WebGL, which caused an error as earlier WebGL versions only accepted a `Float32Array`.

## Introduction

Earlier versions of `gl-matrix` had as type `Tuple | Float32Array`. 
`MapLibre` uses `Float64Array` in several spots, but these were cast to `Float32Array` to appease the squiggly typelords.
The most recent version of `gl-matrix` fixed this issue slightly by widening its types to `Tuple | IndexedCollection<number>` where `IndexedCollection extends Iterable<number>`.

This removed the need for casting to `Float32Array`, but there is still room for improvement:

- `gl-matrix` does not make use of any generics, this means that all of its methods return `vec3` and the like, widening the types of its input.
- `.create()` and a handful of other methods create a default array-type which can be set with `import {glMatrix} from 'gl-matrix'; glMatrix.ARRAY_TYPE = ...`. They also return `vec3` and the like.
- The types use `Tuple | ...` because it wants to communicate that only certain indices are valid and typescript has no alternative for a tuple for `TypedArrays`.

## Changes

I patched `gl-matrix` in the following ways (see https://github.com/bigmistqke/gl-matrix/tree/feat-improve-typings):

1. Added a type-utility `TypedArrayTuple` to simulate tuple-like typings for `TypedArrays` and other `Iterables`, see [ts playground](https://www.typescriptlang.org/play/?#code/PTAEAkFMBsAdIE6gC4E94oPagOaQHaICGyko+ArgLaICWAxqAkfnqAAxagByAtAIwAoNBm7U69AEos8AHm6hIAD1L4AJgGdy1AEaIANKACC9RstWbQFfAGt8mAO74A2gF1QAXlBuAfJ8GgxqbOAOTQBDjIABYh7uYEltwBgaAA-EH0zpRUegiuyYEAXDziCAzSrJDyhs4AdPUm9IaNoeGs0bGuPgDcgoIgoABCmNZqkGqgAGLQmCQAzABMRgjMqCjoZNEkoJj40GtE0DMOWgBuh7QTtOoMkBrCG6AAKhtqy6tPFLDhsgDS12pDAAZCLRRQqBJabK5PweZIAeSotGQfwBhmhiFAAB9QGFQTE-AAyUAAbwK3l+oGuJRoZSkMiqIPaUR8rmKGIQvUCAF9QMSSUxIEQ1Lt9qA2pEosUmZLuqBub1kgMAKJKIhUb5kCgaIh4QoPDBGTzedj6fj6Bb5A1kABqkHoc3eRDWXhe8DeK2dn01smms2QiydqGarXxsR6fQGTyitC0DhG0Am8YQNn19F2GmQoFO9uKdodQeNhAcUxm8yWntQAAo5gBKUBELT5x2V3o5zLsVxy0ADQDI5KB4b9BO3nPwu4E+wPKaBh-bnJbu5PB5GwNHY6B4xRE7haDmG+sMIgEJgEPqR3Nxz2wIB0cmejyPJ9AvFAyvCNHwWdoGugDGRYqijYNvguIsKgIQHpAs6ZPw7CdnKAy3m6ZAPkgz6vpA76ft+v7IP+gEsCB+BgRBQA)
2. methods use generics and return `Vec3Result<T>` and the like, `Vec3Result` is
    - `Vec3Tuple` if `T` is an array and 
    - `TypedArrayTuple<Kind, Vec3Tuple['length']>` if it's not
3. Have the default array type be overwritable via module augmentation: 
    - by default it is `Float32Array` (although this is not 100% correct as it falls back to `Array` if `Float32Array` is not available in the runtime the code is executed)
    - but if you do `declare module 'gl-matrix' { interface Overrides { arrayType: Float64Array; }}` it is typed as `Float64Array`

This allowed me to remove a bunch of type-casts and highlight methods/properties that have multiple types throughout their lifetimes.

## Conclusion

I am not completely sure of how I refactored the base types from 

```tsx
declare type vec3 = IndexedCollection | Vec3Tuple;
```

to 

```tsx
declare type vec3<T extends IndexedCollection | Vec3Tuple = IndexedCollection | Vec3Tuple> = T;
```

which isn't super useful as we do lose some type-information when doing `position: vec3<Float64Array> = new Float64Array(3)` as this just becomes `Float64Array`. I could either 
- narrow `vec3<Float64Array>` to `TypedArrayTuple<Float64Array, 3>`, but this would cause a bunch of type-errors in existing code and the only solution would be to cast as typescript natively does not capture its length
- revert the base type refactor and don't wrap the property-types with `vec3` and the like: `position: Float64Array = new Float64Array(3)`

Ideally I want to upstream this to `gl-matrix`, but I first wanted to get some feedback from you guys.


src/geo/projection/globe_transform.ts
```diff
@@ -234,7 +234,7 @@ export class GlobeTransform implements ITransform {
         this._mercatorTransform.getCoveringTilesDetailsProvider().prepareNextFrame();
     }
 
-    private get currentTransform(): ITransform {
+    private get currentTransform(): VerticalPerspectiveTransform | MercatorTransform {
```
HarelM, line 237: Why was this changed?

src/geo/projection/globe_transform.ts
```diff
@@ -270,13 +270,13 @@ export class GlobeTransform implements ITransform {
...
+    public get projectionMatrix(): mat4<Float64Array> | mat4<Float32Array> { return this.currentTransform.projectionMatrix; }
 
-    public get modelViewProjectionMatrix(): mat4 { return this.currentTransform.modelViewProjectionMatrix; }
+    public get modelViewProjectionMatrix(): mat4<Float64Array> { return this.currentTransform.modelViewProjectionMatrix; }
 
-    public get inverseProjectionMatrix(): mat4 { return this.currentTransform.inverseProjectionMatrix; }
+    public get inverseProjectionMatrix(): mat4<Float64Array> { return this.currentTransform.inverseProjectionMatrix; }
 
-    public get cameraPosition(): vec3 { return this.currentTransform.cameraPosition; }
+    public get cameraPosition(): vec3<Float64Array | Vec3Tuple>  { return this.currentTransform.cameraPosition; }
```
HarelM: This seems like an odd typescript definition...

src/geo/projection/globe_transform.ts
```diff
@@ -339,7 +339,7 @@ export class GlobeTransform implements ITransform {
         this._helper._farZ = this._mercatorTransform.farZ;
     }
 
-    calculateFogMatrix(unwrappedTileID: UnwrappedTileID): mat4 {
+    calculateFogMatrix(unwrappedTileID: UnwrappedTileID): mat4<Float64Array> | mat4<Float32Array> {
```
HarelM: This looks a bad return value... Is it really the case here?

src/geo/projection/mercator_transform.ts
```diff
@@ -286,7 +286,7 @@ export class MercatorTransform implements ITransform {
     getCameraFrustum(): Frustum {
         return Frustum.fromInvProjectionMatrix(this._invViewProjMatrix, this.worldSize);
     }
-    getClippingPlane(): vec4 | null {
+    getClippingPlane(): null {
```
HarelM, line 289: I'm guessing the return value here was meant for the interface implementation...?

HarelM:
Is be more interested to see the typescript charges in gl-matrix that solves all these issues.
It might be that simply adding some overloads will solve some of the issues here, wouldn't it? 
I also find the word tuple confusing, coming from a c# backgrond where tuple is a pair with two items in it...

src/geo/projection/globe_transform.ts
bigmistqke: Yes, I agree, this looks bug prone. All the 'ugly' types are values that switch type during their lifetime, which the previous typings had hidden away from us.

bigmistqke:
> Is be more interested to see the typescript charges in gl-matrix that solves all these issues.

The patch is generated from a build of the last commit of https://github.com/bigmistqke/gl-matrix/tree/feat-improve-typings

> It might be that simply adding some overloads will solve some of the issues here, wouldn't it?

I am not completely following here.

> I also find the word tuple confusing, coming from a c# backgrond where tuple is a pair with two items in it...

I agree, it's a horrible name, but sadly it is what `typescript` officially calls it: https://www.typescriptlang.org/docs/handbook/2/objects.html#tuple-types

src/geo/projection/mercator_transform.ts
bigmistqke: I will revert this!

src/geo/projection/globe_transform.ts
bigmistqke: I will revert this. It's a more narrow type, but it's also a bit irrelevant to this PR.

HarelM:
> It might be that simply adding some overloads will solve some of the issues here, wouldn't it?

This is in regards to the possible solution of the gl-matrix types. Maybe allowing the matrix operators to receive more general types by adding an overload to the methods definition can solve some of the issues here.

birkskyum:
> I also find the word tuple confusing, coming from a c# backgrond where tuple is a pair with two items in it...

I think C# also can hold n-tuples - not just 2-tuples? https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/value-tuples

"You can define tuples with an arbitrary large number of elements:"

HarelM:
True, but I think the original implementation was only `item1` and `item2` fields, but I might remember it wrong as it was a long time ago.
In any case, this is a bit besides the point...

HarelM:
Seems like gl-matrix is a bit stale, not sure what we should do with that information... :-/

CommanderStorm:
Given the number of merge conflicts this PR is unlikely to be mergable soon-ish.
Better types are always good though. Just needs a new attepmt, maybe a more focused one.

CommanderStorm:
or actually, the conflicts might be mergable, let me attempt that

CommanderStorm:
ah, PR does not allow edits by maintainers, so I cannot fix it

PR #6469, opened 2025-09-27
Improve fading - dynamic bi-directional raster cross-fading

wayofthefuture:
Draft - in progress...

Before:

https://github.com/user-attachments/assets/83fb7d9a-b60d-4a91-a04e-3da9bdceadd8

After:

https://github.com/user-attachments/assets/617ec005-db93-4d54-b137-5e2970f2c4db

Before (browser cache disabled):

https://github.com/user-attachments/assets/5651a670-6c17-4692-84a6-7a9c1c71b418

After (browser cache disabled):

https://github.com/user-attachments/assets/19d6a68f-31c5-4d8c-bfbd-6d9304c2c8d2

### DIFF
--- a/src/source/source_cache.test.ts
+++ b/src/source/source_cache.test.ts
@@ -586,6 +586,29 @@ describe('SourceCache.update', () => {
         expect(sourceCache.getIds()).toEqual([new OverscaledTileID(0, 0, 0, 0, 0).key]);
     });
 
+    test('adds ideal (covering) tiles only once for zoom level on raster maps', async () => {
+        const transform = new MercatorTransform();
+        transform.resize(512, 512);
+        transform.setZoom(1);
+
+        const sourceCache = createSourceCache({});
+        sourceCache._source.loadTile = async (tile) => {
+            tile.state = 'loaded';
+        };
+
+        // trigger raster fading logic to begin seeking fading tiles - see _updateRetainedTiles
+        (sourceCache._source as any).type = 'raster';
+
+        const addSpy = vi.spyOn(sourceCache, '_addTile');
+        const dataPromise = waitForEvent(sourceCache, 'data', e => e.sourceDataType === 'metadata');
+        sourceCache.onAdd(undefined);
+        await dataPromise;
+
+        // on update at zoom 1 there should be 4 ideal tiles added through _addTiles
+        sourceCache.update(transform);
+        expect(addSpy).toHaveBeenCalledTimes(4);
+    });
+
     test('respects Source.hasTile method if it is present', async () => {
         const transform = new MercatorTransform();
         transform.resize(511, 511);


wayofthefuture:
Looking for opinions on the before/after videos above...

HarelM:
The videos look nice! 
I think that the fact that fading logic is part of something called "cache" is also problematic. 
@msbarry also reported this as a problem when it comes to implementing the contour source. 
a cache should be cache and loading and evicting it should be handled elsewhere. 
It's a big change, and I'm not expecting you to do it as part of this PR, but I thought it should be mentioned here as well...
In theory, we can also use an existing cache  library like lru-cache to remove the cache handling code here and leave only the logic related to which tiles to keep and which to evict.

wayofthefuture:
@HarelM In this PR I’ve already consolidated caching into the existing SourceCache._cache (backed by tile_cache.ts). I think this is the correct cache to use, which has the behavior you seem to be advocating for.

One possible next step could be a logic-neutral rename of SourceCache in a follow-up PR. From my perspective, SourceCache acts more like a controller than a pure cache (see style/style.ts). If we renamed it to something like SourceController, it might better reflect its role while managing an abstracted SourceController._cache (TileCache) internally.

Then elsewhere such as in render -> painter.ts you would see sourceController.getVisibleCoordinates() and in style.ts sourceControllers[id].hasTransition(), which makes more sense IMO... then the fading logic is then handled by a controller vs cache. That could be a clean solution without changing too much logic.

HarelM:
I don't see a lot of code changes as part of this PR, but SourceController sounds like a good name for something that controls a source.

wayofthefuture:
Once I'm finished writing tests I'll drop the new code once I know it works properly

msbarry:
TileCache sounds good to me, my only concern with "SourceController" is that it isn't exactly controlling the input to sources, it's controlling the input to each different kind of usage of that source. For example the set of tiles from a DEM source needed for 3d terrain for a given viewport are different than the set of tiles from that same source needed for hillshading, and the set needed for rendering contours will be different from both of those.

For raster and vector sources, all layers that use them have the same rules which is probably why they didn't distinguish between source and usages of that source.

Alternatively we could call it SourceController and just say that contour tiles will need a separate instance of it like terrain_source_cache

wayofthefuture:
@msbarry I was thinking that since terrain_source_cache.ts also implements its own custom internal cache (_sourceTileCache) that it should also be renamed possibly to TerrainSourceController. It seems to me that x_source_cache sits right under map render/painter, why i thought controller name was good (thinking of mvc though)... Another possible name could be SourceManager? But I think you have much more expertise than me on this so I'd definitely pivot to your opinion.

wayofthefuture:
@HarelM Code uploaded and new "after" videos uploaded above...

CHANGELOG.md
```diff
@@ -6,7 +6,8 @@
 - Implement data-driven styling support for `line-dasharray` ([#5812](https://github.com/maplibre/maplibre-gl-js/pull/5812))
 
 ### 🐞 Bug fixes
-- Fix raster flickering when using terrain 3D and optimize terrain logic.
+- Improve fading - dynamic bi-directional raster cross-fading and self fading ([#6469](https://github.com/maplibre/maplibre-gl-js/pull/6469))
```
HarelM: More of a feature than a bug, isn't it?

src/style/style.ts
```diff
@@ -480,6 +481,12 @@ export class Style extends Evented {
             const styledLayer = createStyleLayer(layer, this._globalState);
             styledLayer.setEventedParent(this, {layer: {id: layer.id}});
             this._layers[layer.id] = styledLayer;
+
+            // for raster types let the source know fade duration for fade logic
+            if (isRasterStyleLayer(styledLayer) && this.sourceCaches[styledLayer.source]) {
```
HarelM, line 485: This feels like the wrong place to put this logic for some reason I can't define.
But this method looks like a style generation method, what about when updating the style? shouldn't this be done there too?

src/source/tile.ts
```diff
@@ -68,6 +73,10 @@ export class Tile {
     state: TileState;
     timeAdded: number = 0;
     fadeEndTime: number = 0;
+    fadeOpacity: number = 1;
+    fadingBaseRole: FadingRoles;
+    fadingParent: OverscaledTileID;
```
HarelM: What does parent mean in respect to fading? Can't we fade from parent to child and also from child to parent?

src/source/source_cache.ts
```diff
@@ -244,16 +248,23 @@ export class SourceCache extends Evented {
...
     }
 
     _isIdRenderable(id: string, symbolLayer?: boolean) {
-        return this._tiles[id] && this._tiles[id].hasData() &&
-            !this._coveredTiles[id] && (symbolLayer || !this._tiles[id].holdingForFade());
+        const tile = this._tiles[id];
+        return (
+            tile?.hasData() &&
+            (!tile.fadeEndTime || tile.fadeOpacity > 0) &&  // raster fading
+            (symbolLayer || !tile.holdingForSymbolFade())   // symbol fading
```
HarelM: Is it possible to unite the fading logic between symbol and tile somehow? Maybe encapsulate this logic inside the tile class somehow?

src/source/source_cache.ts
```diff
@@ -456,42 +467,14 @@ export class SourceCache extends Evented {
...
      */
-    findLoadedSibling(tileID: OverscaledTileID): Tile {
-        // If a tile with this ID already exists, return it
-        return this._getLoadedTile(tileID);
-    }
-
     _getLoadedTile(tileID: OverscaledTileID): Tile {
         const tile = this._tiles[tileID.key];
-        if (tile && tile.hasData()) {
+        if (tile?.hasData()) {
```
HarelM, line 465: Can you change the signature of this method to be more explicit about returning undefined?
Maybe even add `return undefined` or `return null` at the end to make it more readable?

src/source/source_cache.ts
```diff
@@ -75,17 +82,15 @@ export class SourceCache extends Evented {
...
     terrain: Terrain;
     used: boolean;
     usedForTerrain: boolean;
     tileSize: number;
     _state: SourceFeatureState;
-    _loadedParentTiles: {[_: string]: Tile};
-    _loadedSiblingTiles: {[_: string]: Tile};
     _didEmitContent: boolean;
     _updated: boolean;
+    _rasterFadeDuration: number;
```
HarelM, line 86: This feels a bit too specific to raster, which makes it a bit awkward IMHO.
Can this be placed elsewhere?

src/source/source_cache.ts
```diff
@@ -619,33 +594,36 @@ export class SourceCache extends Evented {
...
         // Remove the tiles we don't need anymore.
         const remove = keysDifference(this._tiles, retain);
-        for (const tileID of remove) {
-            const tile = this._tiles[tileID];
-            if (tile.hasSymbolBuckets && !tile.holdingForFade()) {
-                tile.setHoldDuration(this.map._fadeDuration);
-            } else if (!tile.hasSymbolBuckets || tile.symbolFadeFinished()) {
-                this._removeTile(tileID);
+        for (const key of remove) {
+            if (isRaster) {
```
HarelM: Consider moving all this logic inside the `_removeTile` method and maybe pass the `isRaster` flag (although I'm not sure you'd need to pass it as it can be calculated inside the method).

src/source/source_cache.ts
```diff
@@ -736,103 +714,162 @@ export class SourceCache extends Evented {
...
+     *
+     * Tiles that are candidates for fading out must be loaded and rendered tiles, as loading a tile to then
+     * fade it out would not appear smoothly. The first source of truth for tile fading always starts at the
+     * ideal tile, which continually changes on map adjustment. The state of the previously rendered ideal
+     * tile plane indicates which direction to fade each part of the newer ideal plane (with varying z).
+     *
+     * For a pitched map, the back of the map can have decreasing zooms while the front can have increasing zooms.
+     * Fade logic must therefore adapt dynamically based on the previously rendered ideal tile set.
+     */
+    _updateFadingTiles(idealTileIDs: OverscaledTileID[], retain: {[_: string]: OverscaledTileID}) {
```
HarelM: `Recoed<string, OverscaledTileID>`?

src/source/source_cache.ts
```diff
@@ -736,103 +714,162 @@ export class SourceCache extends Evented {
...
-            // when fadeEndTime is 0, the tile is created but registerFadeDuration
-            // has not been called, therefore must be kept in fadingTiles dictionary
-            // for next round of rendering
-            if (!tile || (tile.fadeEndTime !== 0 && tile.fadeEndTime <= now)) {
-                continue;
+        for (const idealID of idealTileIDs) {
+            const idealTile = this._tiles[idealID.key];
+            const parentIsFader = this._updateFadingParent(idealTile, retain, now);
+            const childIsFader = this._updateFadingChildren(idealTile, retain, now);
+            if (!parentIsFader && !childIsFader) {
```
HarelM: I think this if can be revered to both reduce indentation and make it a bit more readable.

src/source/source_cache.ts
```diff
@@ -736,103 +714,162 @@ export class SourceCache extends Evented {
...
-
-            const tile = this._tiles[id];
 
-            // when fadeEndTime is 0, the tile is created but registerFadeDuration
-            // has not been called, therefore must be kept in fadingTiles dictionary
-            // for next round of rendering
-            if (!tile || (tile.fadeEndTime !== 0 && tile.fadeEndTime <= now)) {
-                continue;
+        for (const idealID of idealTileIDs) {
+            const idealTile = this._tiles[idealID.key];
```
HarelM, line 749: Can idealTile be null?

src/source/source_cache.ts
```diff
@@ -736,103 +714,162 @@ export class SourceCache extends Evented {
...
+     *                                                         ■
+     *                                   ┌──────────────┬──────┴───────┬──────────────┐
+     * Ideal tiles - fading in           ■              ■              ■              ■          -- Fading Parent
+     *                             ┌───┬─┴─┬───┐  ┌───┬─┴─┬───┐  ┌───┬─┴─┬───┐  ┌───┬─┴─┬───┐
+     * Child tiles - fading out    ■   ■   ■   ■  ■   ■   ■   ■  ■   ■   ■   ■  ■   ■   ■   ■    -- Base Role = Departing
+     *
+     * Try direct children first. If none found, try grandchildren. Stops as soon as a generation provides a fader.
+     */
+    _updateFadingChildren(idealTile: Tile, retain: {[_: string]: OverscaledTileID}, now: number): boolean {
+        const parseChildren = (tileID: OverscaledTileID): boolean => {
```
HarelM: Please move this to a different method, no need to have that inside another method.

src/source/source_cache.ts
```diff
@@ -736,103 +714,162 @@ export class SourceCache extends Evented {
...
+            if (tileID.overscaledZ >= this._source.maxzoom) return false;
+            let foundFader = false;
+
+            // find loaded child tiles to fade with the ideal tile
+            for (const childID of tileID.children(this._source.maxzoom)) {
+                const childTile = this._getLoadedTile(childID);
+                if (childTile) {
```
HarelM: Can indentation be reduced here by inverting this if?

src/source/source_cache.ts
```diff
@@ -736,103 +714,162 @@ export class SourceCache extends Evented {
...
+                    const {fadingBaseRole, fadingParent} = childTile;
+
+                    if (fadingBaseRole === FadingRoles.Departing && fadingParent) {
+                        // child is already fading - do nothing
```
HarelM: While I understand the idea behind this comment, I still find this empty if weird, please invert it.

src/source/source_cache.ts
```diff
@@ -736,103 +714,162 @@ export class SourceCache extends Evented {
...
+                if (parseChildren(childID)) {
+                    hasFader = true;
+                }
             }
         }
+
+        return hasFader;
+    }
+
+    _setCrossFadeLogic({baseTile, baseFadingRole, parentTile, now}: CrossFadeArgs) {
```
HarelM: `baseTile` and `targetTile` instead of `parentTile` maybe?
or maybe `sourceTile` and `targetTile`? IDK...

src/source/source_cache.ts
```diff
@@ -736,103 +714,162 @@ export class SourceCache extends Evented {
...
+        baseTile.resetFadeLogic();
+        baseTile.fadingBaseRole = baseFadingRole;
```
HarelM: I think it would be better to move this logic into the tile class and hopefully expose less fields.

src/source/source_cache.ts
```diff
@@ -736,103 +714,162 @@ export class SourceCache extends Evented {
...
     }
 
     /**
-     * Update the cache of loaded sibling tiles
-     *
-     * Sibling tiles are tiles that share the same zoom level and
-     * x/y position but have different wrap values
-     * Maintaining sibling tile cache allows fading from old to new tiles
-     * of the same position and zoom level
+     * Self fading for unloaded tiles. When loading tiles over gaps it feels more natural for them to fade in, however
```
HarelM: While this is nice I think it adds a layer of complexity that I'm not sure worth it.
Parent to child and child to parent fading is symmetric and makes sense (and obviously worth the maintenance effort), but I'm not sure about self fading.
Your call though.

src/render/draw_raster.ts
```diff
@@ -83,40 +81,55 @@ function drawTiles(
...
         context.activeTexture.set(gl.TEXTURE1);
 
-        if (parentTile) {
-            parentTile.texture.bind(textureFilter, gl.CLAMP_TO_EDGE, gl.LINEAR_MIPMAP_NEAREST);
-            parentScaleBy = Math.pow(2, parentTile.tileID.overscaledZ - tile.tileID.overscaledZ);
-            parentTL = [tile.tileID.canonical.x * parentScaleBy % 1, tile.tileID.canonical.y * parentScaleBy % 1];
-        } else {
-            tile.texture.bind(textureFilter, gl.CLAMP_TO_EDGE, gl.LINEAR_MIPMAP_NEAREST);
+        // calculate tile fading values
+        let parentTile = null, fadeValues = null, parentScaleBy = null, parentTopLeft = null;
```
HarelM: Please use separate lines for initialization.

HarelM: Better yet, extract this part to a method and return these values.

src/render/draw_raster.ts
```diff
@@ -139,45 +151,42 @@ function drawTiles(
...
+        parentTile.fadeOpacity = doFadeIn ? opacities[1] : opacities[0];
+
+        // when crossfading, this tile is drawn fully opaque, and mix controls partner visibility
         return {
             opacity: 1,
+            mix: 1 - tile.fadeOpacity
+        };
+    }
+    // simple fade-in for tile without partner
+    else {
```
HarelM: nit picking: can the else be moved up and the comment moved inside the else?
Or better yet if there's a "definite" return, simply remove the `else` statement.

src/render/draw_raster.ts
```diff
@@ -139,45 +151,42 @@ function drawTiles(
     }
 }
 
-function getFadeValues(tile: Tile, parentTile: Tile, sourceCache: SourceCache, layer: RasterStyleLayer, transform: IReadonlyTransform, terrain: Terrain) {
-    const fadeDuration = layer.paint.get('raster-fade-duration');
+function getFadeValues(tile: Tile, parentTile: Tile, fadeDuration: number): {opacity: number; mix: number} {
```
HarelM: The signature here is really good! I love that the references to source cache and transform were removed!

HarelM: This method seems to update some fields in the tile object, I'm not sure `getFadeValues` is the right name for it...

src/render/draw_raster.ts
```diff
@@ -139,45 +151,42 @@ function drawTiles(
...
+    const now = browser.now();
+    const timeSinceTile = (now - tile.timeAdded) / fadeDuration;
+
+    // crossfading for parent tile
+    if (parentTile) {
+        const timeSinceParent = (now - parentTile.timeAdded) / fadeDuration;
+
+        // set fading opacity based on current fade direction
+        const doFadeIn = (tile.fadingBaseRole === FadingRoles.Incoming);
+        const opacities = [
```
HarelM: Can this be split into two variables instead of an array?

src/source/source_cache.test.ts
```diff
@@ -105,6 +101,13 @@ afterEach(() => {
     vi.clearAllMocks();
 });
 
+class ResizeObserverStub {
```
HarelM: Why is this class needed? I think it is handled well in `beforeMapTest`...?

src/source/source_cache.test.ts
```diff
@@ -586,6 +589,60 @@ describe('SourceCache.update', () => {
...
+        transform.resize(512, 512);
+        transform.setZoom(1);
+
+        const sourceCache = createSourceCache({});
+        sourceCache._source.loadTile = async (tile) => {
+            tile.state = 'loaded';
+        };
+
+        // trigger raster fading logic to begin seeking fading tiles - see _updateRetainedTiles
+        (sourceCache._source as any).type = 'raster';
```
HarelM: Can't the `createSrouceCache` handle this part? I'd like to avoid `as any` as much as possible.

src/source/source_cache.test.ts
```diff
@@ -586,6 +589,60 @@ describe('SourceCache.update', () => {
...
+        sourceCache.onAdd(undefined);
+        await dataPromise;
+
+        // on update at zoom 1 there should be 4 ideal tiles added through _addTiles
+        sourceCache.update(transform);
+        expect(addSpy).toHaveBeenCalledTimes(4);
+    });
+
+    test('bypasses fading logic when raster fading is disabled', async () => {
+        const map = new Map({
```
HarelM: there's a `createMap` method for this, isn't there?

src/source/source_cache.test.ts
```diff
@@ -586,6 +589,60 @@ describe('SourceCache.update', () => {
...
+                sources: {rasterSource: {type: 'raster', tiles: [], tileSize: 256}},
+                layers: [{id: 'rasterLayer', type: 'raster', source: 'rasterSource',
+                    paint: {'raster-fade-duration': 0}
+                }]
+            },
+            interactive: false
+        });
+        await map.once('styledata');
+
+        const style = (map as any).style;
```
HarelM: `getStyle`?

src/source/source_cache.test.ts
```diff
@@ -910,6 +842,163 @@ describe('SourceCache.update', () => {
...
+        sourceCache._source.loadTile = async (tile) => {
+            tile.state = 'loaded';
+        };
+        sourceCache.on('data', (e) => {
+            if (e.dataType === 'source' && e.sourceDataType === 'metadata') {
+                sourceCache.update(transform);
+            }
+        });
+        (sourceCache._source as any).type = 'raster';
+        sourceCache._rasterFadeDuration = 300;
```
HarelM: Is there a public API that can be used here instead of `_rasterFadeDuration`?

src/source/source_cache.test.ts
```diff
@@ -910,6 +842,163 @@ describe('SourceCache.update', () => {
...
+        sourceCache.onAdd(undefined);
+
+        // get default zoom ideal tiles at zoom specified above
+        await Promise.resolve();
```
HarelM: `await sleep(0)`?

src/source/source_cache.test.ts
```diff
@@ -910,6 +842,163 @@ describe('SourceCache.update', () => {
...
+        // ideal tiles will become fading grandparent when zooming in
+        const grandParents: Tile[] = Object.values(sourceCache._tiles);
```
HarelM: It would be great if there was another method to extract the current file instead of using `_tiles`, but I don't know if there is...

src/source/source_cache.test.ts
```diff
@@ -910,6 +842,163 @@ describe('SourceCache.update', () => {
...
+        const children: Tile[] = Object.values(sourceCache._tiles);
+
+        // zoom out 1 level - ideal tiles (new children) should fade out
+        transform.setZoom(9);
+        sourceCache.update(transform);
+        await Promise.resolve();
+
+        // ensure that the loaded child was retained and fading logic was applied
+        for (const child of children) {
+            expect(sourceCache._tiles).toHaveProperty(child.tileID.key);
```
HarelM: These are a bit too specific to the current implementation.
Can this be somehow improved by collecting the loaded tiles from the `loadTile` method above and check their state when the test finishes?

HarelM:
Thanks a lot for all the work on this!
I've added some comment, some are minor, some might require a bit more changes.
This looks good overall, thanks for looking into this!

src/source/tile.ts
wayofthefuture: No, that is the core issue. The shader won't allow it and I don't understand shaders enough to know if it can or even attempt to touch those. Hence my mention of many-to-one in the diagrams showing many must fade with one - hence the child -> parent relationship, and naming base -> parent. I had to re-write my entire modification due to this twist, and why the previous implementation likely didn't work...

src/style/style.ts
wayofthefuture: If we don't want fade on page load then it can be deleted.

src/source/source_cache.ts
wayofthefuture: parent or ancestor imo as target cannot be a descendent, see tile diagrams above fade logic

src/source/source_cache.test.ts
wayofthefuture: Source cache is inaccessible from serialized result. I can use (map as Map).style however...

src/source/source_cache.ts
wayofthefuture: I wish!

src/source/source_cache.ts
wayofthefuture: I think this would risk fading problems with vector symbols which I made extensive efforts not to touch... much relies on the behavior of removeTile regarding symbols...

src/source/source_cache.ts
wayofthefuture: No because _updateRetainedTiles which is called before it this calls _addTile on every ideal tile.

wayofthefuture: Line 664

src/style/style.ts
wayofthefuture: I went ahead and remove on page load so the map comes right up. If otherwise it becomes too much I'll revert it...

HarelM: Did you commit the latest changes? I don't see a change in the code...

src/source/source_cache.ts
HarelM: That's not great :-/ Maybe just add a `removeTileWithRaster` method to call `_removeTile` internally?
It feels like this should be better addressed somehow, definitely a code smell... :-/

src/source/source_cache.test.ts
HarelM: map should have a `Map` type, why is casting needed?

src/source/source_cache.ts
HarelM: Ahh, I see now it's alway in the same "direction".

src/style/style.ts
wayofthefuture: Trying to get it ready now!

src/source/tile.ts
HarelM: Is there a way we can help with that somehow?
@mwilsnd knows his way around shader code and might be able to help a little too if we could understand the problem you faced...?

wayofthefuture:
I reduced self-fading just to the edges, I believe this is a nice touch. Otherwise there are more tests in there for you to look at for the edge calculation, and I did my best to address your code reviews.

src/source/tile.ts
wayofthefuture: Maybe... @mwilsnd If I attempt to cross-fade 1 tile with 4 in draw_raster, either the first or the last of the 4 fades. I.e. you would see 1 parent fading and only 1 of 4 children fade. However, if you cross-fade 4 tiles with 1, all 4 tiles fade, and the only 1 parent tile fades. It looked like significant work at the shader level to expand parentTL, and seemed a heavy lift. So I had to work within the constraints of the shader to get it working.

src/source/source_cache.ts
wayofthefuture: I think it is a much large issue as raster and vector are mixed in this same source_cache and they should really be separate... i.e. another PR lol

wayofthefuture:
Updated video:

https://github.com/user-attachments/assets/cc1b1b82-3e1d-414a-911f-b0233f50003c

src/source/source_cache.ts
wayofthefuture: I extracted the removal logic and flattened it... I think you'll like it much better now

wayofthefuture: I further improved it I think it's more inline with what you were looking for

src/source/tile.ts
mwilsnd: The raster shader only has two texture samplers. When you fade from parent to child, the 4 new child tiles each sample the same parent tile texture during the fade (as I understand things). Going the other way then, a parent tile would have to be able to sample all 4 child textures, meaning you would need a shader variant that samples and mixes a total of 5 textures. You might be able to take the 4 children and stitch their textures together and feed that into the single fade texture. You'd have a larger than normal texture with the 4 children packed into it, but it would probably work.

HarelM: Given this feature and the code written here, does it make sense to change the shader code or the solution here is good enough?

mwilsnd: This is probably good enough. If the stitching trick works I'd prefer that over changing the shaders, texture blending is one of the more (relatively) costly operations you can do in a fragment shader.

wayofthefuture:
@HarelM Just a reminder that there is additional code (regarding tile edge detection) for you to review that was not in the original code sent for review.

src/source/source_cache.ts
```diff
@@ -736,103 +726,193 @@ export class SourceCache extends Evented {
...
             }
+        }
 
-            // Cache the result of this traversal to all newly visited tiles
-            for (const key of path) {
-                this._loadedParentTiles[key] = parentTile;
+        return hasFader;
+    }
+
+    _parseFadingChildren(idealTile: Tile, childIDs: OverscaledTileID[], retain: Record<string, OverscaledTileID>, now: number): boolean {
```
HarelM: This method name doesn't explain what it does, and why it returns a boolean, any chance you could improve the method's name?

src/source/source_cache.ts
```diff
@@ -736,103 +726,193 @@ export class SourceCache extends Evented {
...
+     *
+     * Try direct children first. If none found, try grandchildren. Stops as soon as a generation provides a fader.
+     */
+    _updateFadingChildren(idealTile: Tile, retain: Record<string, OverscaledTileID>, now: number): boolean {
+        // search first level of descendents (4 tiles)
+        const idealChildren = idealTile.tileID.children(this._source.maxzoom);
+        let hasFader = this._parseFadingChildren(idealTile, idealChildren, retain, now);
+
+        // if none found search second level of descendents (16 tiles)
+        if (!hasFader) {
```
HarelM: Consider inverting this if to reduce indentation.

src/source/source_cache.ts
```diff
@@ -736,103 +726,193 @@ export class SourceCache extends Evented {
...
+            return true;
+        }
 
-                // Do we have a cached result from previous traversals?
-                if (currentId.key in this._loadedParentTiles) {
-                    parentTile = this._loadedParentTiles[currentId.key];
-                    break;
-                }
+        // find a loaded parent tile to fade with the ideal tile
+        const maxAncestorLevels = 5;
```
HarelM: Should this be configured somewhere? This number seems a bit arbitrary (Although I tend to understand why you chose it).

src/source/source_cache.ts
```diff
@@ -736,103 +726,193 @@ export class SourceCache extends Evented {
...
-        this._retainLoadedChildren(fadingTiles, retain);
-
-        for (const id in tilesForFading) {
-            if (!retain[id]) {
-                // If a tile is only needed for fading, mark it as covered so that it isn't rendered on it's own.
-                this._coveredTiles[id] = true;
-                retain[id] = tilesForFading[id];
-            }
+        // fallback to looking for loading edge tiles for the non-faders
+        nonFaders = this._updateFadingEdges(nonFaders, now);
```
HarelM: Am I misreading this or `nonFaders` content it always overwritten...?

HarelM: Oh, wait, it is passed into the method too...
If this is passed into the method and the method name is updateFadingEdges maybe it doesn't need to return anything and simply update the given object?
IDK... it's a bit confusing I guess.
Probably nit-picking more than anything else.

src/source/source_cache.ts
```diff
@@ -736,103 +726,193 @@ export class SourceCache extends Evented {
...
-     * x/y position but have different wrap values
-     * Maintaining sibling tile cache allows fading from old to new tiles
-     * of the same position and zoom level
+     * One-to-one self fading for unloaded edge tiles (for panning sideways on map). for loading tiles over gaps it feels
+     * more natural for them to fade in, however if they are already loaded/cached then there is no need to fade as map will
+     * look cohesive with no gaps. Note that draw_raster determines fade priority, as many-to-one fade supersedes edge fading.
      */
-    _updateLoadedSiblingTileCache() {
-        this._loadedSiblingTiles = {};
+    _updateFadingEdges(idealTiles: Tile[], now: number): Tile[] {
```
HarelM: I'm not sure this extra logic is worth the UX change, I think my intuition is that panning left and right should simply appear without fading in.
Also I'm not sure this will work as expected if the map is rotated...?

src/render/draw_raster.ts
```diff
@@ -139,46 +135,68 @@ function drawTiles(
...
-                opacity: 1,
-                mix: 1 - childOpacity
-            };
-        } else {
-            return {
-                opacity: childOpacity,
-                mix: 0
-            };
+function getTileFadeProps(tile: Tile, sourceCache: SourceCache, fadeDuration: number, isTerrain: boolean): TileFadeProps {
+    let parentTile = null;
```
HarelM: These need to by typed as there's no way to understand what their type is with this type of initialization.

src/render/draw_raster.ts
```diff
@@ -139,46 +135,68 @@ function drawTiles(
...
-            return {
-                opacity: 1,
-                mix: 1 - childOpacity
-            };
-        } else {
-            return {
-                opacity: childOpacity,
-                mix: 0
-            };
+function getTileFadeProps(tile: Tile, sourceCache: SourceCache, fadeDuration: number, isTerrain: boolean): TileFadeProps {
```
HarelM: `getTileFadeProperties`?

HarelM: This still updates the tile's fade properties, right? It's still not clear from the method name.

src/render/draw_raster.ts
```diff
@@ -139,46 +135,68 @@ function drawTiles(
...
+            parentScaleBy = 1;
+            parentTopLeft = [0, 0];
         }
-    } else {
+    }
+
+    return {parentTile, fadeValues, parentScaleBy, parentTopLeft};
+}
+
+function doFadeValues(tile: Tile, parentTile: Tile, fadeDuration: number): {opacity: number; mix: number} {
```
HarelM: Not sure what `doFadeValues` is really doing. Is it updating the tile's fade values? If so it should be called `updateTileFadeValues` or something similar.

I would try and split between update logic, calculation logic and fetching logic as much as possible to make these method as readable as possible.

src/render/draw_raster.ts
```diff
@@ -139,46 +135,68 @@ function drawTiles(
...
-                mix: 0
-            };
+function getTileFadeProps(tile: Tile, sourceCache: SourceCache, fadeDuration: number, isTerrain: boolean): TileFadeProps {
+    let parentTile = null;
+    let fadeValues = null;
+    let parentScaleBy = null;
+    let parentTopLeft = null;
+
+    // first looking to cross-fade with a departing/incoming parent, otherwise looking for self fading edge tiles
+    if (fadeDuration > 0 && !isTerrain) {
```
HarelM: Consider inverting these ifs to reduce indentation and code complexity.

src/source/source_cache.test.ts
```diff
@@ -586,6 +603,55 @@ describe('SourceCache.update', () => {
...
+                version: 8,
+                sources: {rasterSource: {type: 'raster', tiles: [], tileSize: 256}},
+                layers: [{id: 'rasterLayer', type: 'raster', source: 'rasterSource',
+                    paint: {'raster-fade-duration': 0}
+                }]
+            }
+        });
+        await map.once('styledata');
+
+        const style = (map as Map).style;
```
HarelM: Why is this casting needed? isn't map of type Map?

src/source/source_cache.test.ts
```diff
@@ -586,6 +603,55 @@ describe('SourceCache.update', () => {
...
+        const sourceCache = style.sourceCaches['rasterSource'];
+        const spy = vi.spyOn(sourceCache, '_updateFadingTiles');
```
HarelM, line 640: Can we spy on things that are not part of the class we are testing? or since we are overriding the loadTile method maybe we can leverage this somehow?

src/source/source_cache.test.ts
```diff
@@ -910,6 +851,167 @@ describe('SourceCache.update', () => {
...
 
+    test('retains fading children and applies fading logic when zooming out', async () => {
+        const transform = new MercatorTransform();
+        transform.resize(1024, 1024);
+        transform.setZoom(10);
+
+        const sourceCache = createSourceCache({raster: true});
+        const loadedTiles: {[_: string]: Tile} = {};
+        sourceCache._source.loadTile = async (tile) => {
+            loadedTiles[tile.tileID.key] = tile;
```
HarelM, line 862: GREAT!

src/source/source_cache.test.ts
```diff
@@ -910,6 +851,167 @@ describe('SourceCache.update', () => {
         expect(sourceCache.getTile(wrappedTileID)).toBe(tile);
     });
 
+    test('retains fading children and applies fading logic when zooming out', async () => {
+        const transform = new MercatorTransform();
+        transform.resize(1024, 1024);
+        transform.setZoom(10);
+
+        const sourceCache = createSourceCache({raster: true});
+        const loadedTiles: {[_: string]: Tile} = {};
```
HarelM: Consider using `Record`.

HarelM:
I've added some more comments, I finding it hard to see what are the changes between reviews, so I'm probably repeating myself and also finding new things to comment about...
Sorry...

wayofthefuture:
No worries! My primary concern is whether the fading edge logic could be too heavy ie all the array ops. Other popular libraries fade the edges, but the compute expense may not be worth it. It's also only parallel / perpendicular to lat/lon so on rotation it would look like a fading corner when panning...

I wonder if a small over-render buffer around the map could be better but I'm sure this has been exhaustively discussed before...

src/source/source_cache.ts
wayofthefuture: Agreed

src/source/source_cache.test.ts
wayofthefuture: it specifically targets the calling "if" statement in update() that previously did not have a check for fade logic, causing fade logic to adulterate the field-of-view-150 test

wayofthefuture:
I significantly improved and optimized the edge fading logic and it is looking quite nice. Your suggestions greatly improved the readability of the fading logic. Other than that I did my best to hopefully address the issues...

src/source/source_cache.ts
wayofthefuture: I checked and it looks good with rotation as well...

wayofthefuture:
Please don't merge yet I found a slight issue...

src/render/draw_raster.ts
```diff
@@ -139,46 +148,67 @@ function drawTiles(
...
+    const parentTileOpacity = doFadeIn ? opacity2 : opacity1;
+    const fadeMix = {
+        opacity: 1,
+        mix: 1 - tileOpacity
+    };
+
+    return {tileOpacity, parentTileOpacity, fadeMix};
+}
+
+// simple fade-in for tile without parent (i.e. edge tiles)
```
HarelM: Use tsdocs style comments for a method as it allow IDEs to show this comment.

src/render/draw_raster.ts
```diff
@@ -139,46 +148,67 @@ function drawTiles(
...
+        parentTile = sourceCache._getLoadedTile(tile.fadingParentID);
+        if (parentTile) {
+            parentScaleBy = Math.pow(2, parentTile.tileID.overscaledZ - tile.tileID.overscaledZ);
+            parentTopLeft = [
+                (tile.tileID.canonical.x * parentScaleBy) % 1,
+                (tile.tileID.canonical.y * parentScaleBy) % 1
+            ];
+            fadeValues = getCrossFadeValues(tile, parentTile, fadeDuration);
+        }
+    } else if (tile.selfFading) {
```
HarelM: nit picking: This method can be indented even less:
```ts
if (tile.selfFading) {
    ...
    return ...
}
if (!tile.fadingParentID) {
    return ...
}
parentTile = sourceCache._getLoadedTile(tile.fadingParentID);
if (!parentTile) {
    return ...
}
...
```

src/render/draw_raster.ts
```diff
@@ -83,38 +94,37 @@ function drawTiles(
 
     const colorMode = painter.colorModeForRenderPass();
     const align = !painter.options.moving;
+    const rasterOpacity = layer.paint.get('raster-opacity');
+    const rasterResampling = layer.paint.get('raster-resampling');
+    const fadeDuration = layer.paint.get('raster-fade-duration');
+    const isTerrain = !!painter.style.map.terrain;
+
+    // update raster fade duration in source cache with the newest layer paint property
+    sourceCache.setRasterFadeDuration(fadeDuration);
```
HarelM: Are we sure we can't move it to somewhere else? this feels so weird to call it as part of a drawing logic...
Can we set it when a layers is added or changed maybe?

src/render/draw_raster.ts
```diff
@@ -83,38 +94,37 @@ function drawTiles(
...
+        // create and bind first texture
         context.activeTexture.set(gl.TEXTURE0);
         tile.texture.bind(textureFilter, gl.CLAMP_TO_EDGE, gl.LINEAR_MIPMAP_NEAREST);
 
+        // create second texture - use either the current tile or fade tile to bind second texture below
         context.activeTexture.set(gl.TEXTURE1);
-
+        const {parentTile, parentScaleBy, parentTopLeft, fadeValues} = getFadeProperties(tile, sourceCache, fadeDuration, isTerrain);
         if (parentTile) {
+            tile.fadeOpacity = fadeValues.tileOpacity;
```
HarelM: Isn't this line called regardless of the value of the if condition and can be moved upwards?

src/source/source_cache.test.ts
HarelM: Thanks, but this does not answer my question regarding if a spy can be placed on a method that is not part of this class, like a tile method or property...

src/source/source_cache.test.ts
```diff
@@ -2619,5 +2542,84 @@ describe('SourceCache::refreshTiles', () => {
         expect(spy.mock.calls[2][1]).toBe('expired');
         expect(spy.mock.calls[3][1]).toBe('expired');
     });
+});
+
+describe('SourceCache._getEdgeTiles', () => {
```
HarelM: I think that this method doesn't use any of the sourceCache fields.
Can we move it out of this class and export it somewhere (in some utils files maybe) and move these tests as well?
The fact that we are testing it like so is a good indication that it might not belong to the source cache file and class.

src/source/tile.ts
```diff
@@ -124,16 +143,41 @@ export class Tile {
...
+            this.hasData() &&
+            (!this.fadeEndTime || this.fadeOpacity > 0) &&  // raster fading
+            (symbolLayer || !this.holdingForSymbolFade())   // symbol fading
+        );
+    }
 
-        if (fadeEndTime < this.fadeEndTime) {
-            return;
-        }
+    // many-to-one crossfade between a base tile and parent (ancestor) tile
```
HarelM: Use tsdoc style comment.

src/render/draw_raster.ts
wayofthefuture: I can do it the right way but have to change two other files

src/render/draw_raster.ts
wayofthefuture: Yes I did for readability alone

src/render/draw_raster.ts
HarelM: Let's try that then. If you think it would not look great you can open a PR against this branch so I could see how it looks.

src/render/draw_raster.ts
HarelM: I think it makes sense to keep only the part about the parent inside the if, it makes the if statement stand out. 
But it's not picking mostly...

src/render/draw_raster.ts
wayofthefuture: See next commit - I think this i the best we could do considering... source setting needs to be updated on either layer creation or manually changing it...

wayofthefuture:
@HarelM Do you have any idea why the browser test is failing? "requestAnimationFrame is not defined"

HarelM:
Flaky tests that I don't understand how it can fail. 
Ignore it, sorry...

wayofthefuture:
New Demo:

https://github.com/user-attachments/assets/30f99d49-b34d-49c0-a273-9ce784195401

wayofthefuture:
New Demo:

https://github.com/user-attachments/assets/3dde117d-4c83-4f63-84f9-b383f33a4e01

HarelM (approved):

HarelM:
Thanks!!

HarelM:
Thanks again for all the hard work on this! Truly appreciated!

hiddewie:
I am getting reports from users about the (background) map not being loaded in https://github.com/hiddewie/OpenRailwayMap-vector/issues/644.

The map is created with
```js
const backgroundMap = new maplibregl.Map({
  container: 'background-map',
  style: buildBackgroundMapStyle(),
  attributionControl: false,
  interactive: false,
  renderWorldCopies: false,
  zoom: 1,
});
```
and a style like
```js
{
  name: 'Background map',
  version: 8,
  layers: [
    {
      id: "background-map",
      type: "raster",
      source: "background_map",
    },
  ],
  sources: {
    background_map: {
      type: 'raster',
      tiles: [url],
      tileSize: 256,
    },
  },
}
```

Since Maplibre GL JS 5.9.0, the initial map does not show all tiles, and some tiles stay (partially) "faded out". A single interaction like a pan or zoom instantly loads all tiles. All following map interaction works as normal. There are no console errors.

Reading the release notes led me to this pull request, which seems to be related.
 
With Maplibre GL JS 5.9.0:
image

With Maplibre GL JS 5.8.0:
image

Is this a known problem, and is there a workaround? Or should I create a separate issue with this bug report?

HarelM:
I think a separate bug report is the right direction, preferably with something that can be reproduced.

wayofthefuture:
I am currently taking a look at this...

hiddewie:
I created issue #6649 with a reproducing HTML file.

Issue #5026, opened 2024-11-09
getZoomAdjustment() doesn't scale correct for globe with terrain


birkskyum:
The globe is supposed to be rendered at the same size when changing the center, which require ongoing adjustment of the zoom level (a fixed zoom level, say zoom=2, will be much closer to earth surface at the pole than at equator).

This continuous scaling doesn't work when terrain is enabled.

https://github.com/user-attachments/assets/6e5c337a-4608-44bd-ab83-1cc11f2099a1

**Repro:**
for vector map - the hybrid-terrain.html example has this on by default.
for raster map - set `projection: {type: "globe" } (or `vertical-perspective`)  in the 3d-terrain.html examlpe

PR #6470, opened 2025-09-28
Fix zoom issue with globe + terrain when dragging to poles

### LINKED ISSUE #5839: When globe + terrain are enabled, dragging toward the arctic circle  or antarctica zooms in

birkskyum:
Closes #5839
- #5839

Closes https://github.com/maplibre/maplibre-gl-js/issues/5026
- https://github.com/maplibre/maplibre-gl-js/issues/5026

Rotating a globe with terrain enabled used to “zoom toward the poles” because the mercator-specific terrain pan logic bypassed the globe helper’s zoom compensation.

This PR redirects all globe+terrain drags through the globe pan helper so the apparent radius stays constant, while keeping the old pixel-delta path for mercator terrain users.

Before

https://github.com/user-attachments/assets/5611978c-8eec-43bb-af5b-5d0578c37d06

After

https://github.com/user-attachments/assets/5dfbbe04-29d5-4b51-a7f5-46fcc98fa89e

## Launch Checklist

 - [X] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/ui/handler_manager.test.ts
+++ b/src/ui/handler_manager.test.ts
@@ -1,0 +1,203 @@
+import {describe, it, expect, vi} from 'vitest';
+import Point from '@mapbox/point-geometry';
+
+import {HandlerManager} from './handler_manager';
+import type {TerrainScenarioOptions, EventsInProgress} from './handler_manager';
+import type {Map} from './map';
+import {LngLat} from '../geo/lng_lat';
+import type {ICameraHelper, MapControlsDeltas} from '../geo/projection/camera_helper';
+import type {Terrain} from '../render/terrain';
+import type {ITransform} from '../geo/transform_interface';
+
+type ManagerWithInternals = HandlerManager & Record<string, unknown>;
+
+type CameraHelperStub = Pick<ICameraHelper, 'handleMapControlsRollPitchBearingZoom' | 'handleMapControlsPan'> & {
+    useGlobeControls: boolean;
+};
+
+const createManager = (useGlobeControls = false) => {
+    const manager = Object.create(HandlerManager.prototype) as ManagerWithInternals;
+    const handleZoom = vi.fn<(deltas: MapControlsDeltas, tr: ITransform) => void>();
+    const handlePan = vi.fn<(deltas: MapControlsDeltas, tr: ITransform, preZoomAroundLoc: LngLat) => void>();
+
+    const cameraHelper: CameraHelperStub = {
+        handleMapControlsRollPitchBearingZoom: handleZoom,
+        handleMapControlsPan: handlePan,
+        useGlobeControls,
+    };
+
+    const mapStub: Pick<Map, 'cameraHelper' | '_elevationFreeze'> = {
+        cameraHelper: cameraHelper as unknown as ICameraHelper,
+        _elevationFreeze: false,
+    };
+
+    manager._map = mapStub as Map;
+    manager._terrainMovement = false;
+
+    return {manager, handleZoom, handlePan};
+};
+
+type TerrainOptionsParams = {
+    terrain?: Terrain | null;
+    combinedEventsInProgress?: EventsInProgress;
+    panDelta?: Point;
+    centerPoint?: Point;
+    screenPointToLocationResult?: LngLat;
+};
+
+const createTerrainOptions = (params: TerrainOptionsParams = {}): {
+    options: TerrainScenarioOptions;
+    screenPointToLocationMock: ReturnType<typeof vi.fn>;
+    setCenterMock: ReturnType<typeof vi.fn>;
+} => {
+    const panDelta = params.panDelta;
+    const terrain = Object.prototype.hasOwnProperty.call(params, 'terrain') ? params.terrain : ({} as Terrain);
+    const combinedEventsInProgress = Object.prototype.hasOwnProperty.call(params, 'combinedEventsInProgress') ?
+        params.combinedEventsInProgress : {};
+    const screenPointToLocationMock = vi.fn<(point: Point) => LngLat>(() => params.screenPointToLocationResult ?? new LngLat(1, 1));
+    const setCenterMock = vi.fn<(center: LngLat) => void>();
+
+    const transformStub = {
+        centerPoint: params.centerPoint ?? new Point(0, 0),
+        center: new LngLat(0, 0),
+        screenPointToLocation: screenPointToLocationMock,
+        setCenter: setCenterMock,
+    } satisfies Pick<ITransform, 'centerPoint' | 'center' | 'screenPointToLocation' | 'setCenter'>;
+
+    const deltasForHelper: MapControlsDeltas = {
+        panDelta: panDelta ?? new Point(0, 0),
+        zoomDelta: 0,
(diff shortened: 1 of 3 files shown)


src/ui/handler_manager.ts
```diff
@@ -537,19 +537,30 @@ export class HandlerManager {
             tr.center :
             tr.screenPointToLocation(panDelta ? around.sub(panDelta) : around);
 
+        // Two terrain modes share the same zoom/pitch handling; panning differs depending on projection.
```
HarelM: Due to all the ifs here I think this should be placed in a method of its own and explicitly detail the condition of each if here, i.e:
```ts
if (!terrain) {
    doSomething()
    return;
}
if (terrain && this._map.cameraHelper.useGlobeControls && (!this._terrainMovement && (combinedEventsInProgress.drag || combinedEventsInProgress.zoom)) {
   doSeomthingElse();
   return;
}
```
This way is will be easier to understand which scenario is being handled in each if, which in its current state is very hard to understand.

It might not require a method of its own, but I tend to think it will allow more readable code this way...

HarelM:
While not obvious from the failing CI name, it looks like the code that was changed was not fully covered with tests, if possible please try and cover it with tests.
I still need to understand what I need to fix in order for the coverage annotations to work (instead of codecov which stopped working a while ago)...

src/ui/handler_manager.ts
birkskyum: Good idea - It is broken into three scenarios now, for no-terrain, mercator-terrain, globe-terrain - and made more type safe, so it should be easier to reason about now.

```ts
    _applyTerrainScenario(options: TerrainScenarioOptions) {
        if (!options.terrain) {
            this._applyNoTerrainScenario(options);
            return;
        }

        if (this._map.cameraHelper.useGlobeControls) {
            this._applyGlobeTerrainScenario(options);
            return;
        }

        this._applyMercatorTerrainScenario(options);
    }
```

src/ui/handler_manager.ts
```diff
@@ -572,6 +566,54 @@ export class HandlerManager {
 
     }
 
+    _applyTerrainScenario(options: TerrainScenarioOptions) {
```
HarelM: `_handleMapControls` maybe as this is applied to non terrain scenario as well...?

HarelM: `_handleMapMovement`?
`_applyMovementScenario`?
IDK...

HarelM: BTW, this method looks great now! Thanks!

src/ui/handler_manager.test.ts
```diff
@@ -0,0 +1,218 @@
...
+        handlerName: name,
+        originalEvent: new MapEvent(`${String(name)}start`),
+    };
+}
+
+describe('HandlerManager terrain scenarios', () => {
+    describe('_applyTerrainScenario', () => {
+        it('delegates to no-terrain scenario when terrain is missing', () => {
+            const {manager} = createManager();
+            const noTerrain = vi.spyOn(manager, '_applyNoTerrainScenario');
```
HarelM: Is it possible to spy on the transform methods instead of private methods in the class that is under test?

src/ui/handler_manager.test.ts
```diff
@@ -0,0 +1,218 @@
...
+            expect(globeTerrain).not.toHaveBeenCalled();
+        });
+    });
+
+    describe('_applyGlobeTerrainScenario', () => {
+        it('enables terrain movement and freezes elevation on new drag or zoom', () => {
+            const {manager, handleZoom, handlePan} = createManager(true);
+            const {options} = createTerrainOptions({combinedEventsInProgress: {drag: true}});
+
+            manager._applyGlobeTerrainScenario(options);
```
HarelM: I would advise to keep the check only on the `_handleMapControls` method and not dive into more "private" methods.
Ideally I would even look for a more "public" method (those without the "_" prefix), but I'm not sure this class has those or it's easy enough to configure those, so I'm guessing testing `_handleMapControls` would do in this case, but feel free to explore testing a more "public API" for this class.

HarelM:
Overall, this looks very good, thanks!!

src/ui/handler_manager.test.ts
birkskyum: it's just _handleMapControls now

src/ui/handler_manager.test.ts
birkskyum: I haven't been able to find a way to do so

HarelM:
I've rewritten the first test to use some existing test infrastructure, here it how it looks:
I think the `createTerrainOptions` so not handle the logic of all the tests, but instead in each test declare only what you need to mock in order for the test to pass, this way the logic in the test files is kept to a minimum thus making sure we are testing the actual code and not the test file code as much as possible.
```ts
describe('HandlerManager terrain scenarios', () => {
    let map: Map;
    beforeEach(() => {
        beforeMapTest();
        map = createMap();
    });

    it('_handleMapControls keeps terrain movement disabled when terrain is not enabled', () => {
        const manager = new HandlerManager(map, {} as any as CompleteMapOptions);
        const {options, setCenterMock} = createTerrainOptions({
            terrain: null,
            combinedEventsInProgress: {drag: true},
        });

        manager._terrainMovement = false;
        map._elevationFreeze = false;
        map.cameraHelper = {
            handleMapControlsRollPitchBearingZoom: vi.fn(),
            handleMapControlsPan: vi.fn(),
            useGlobeControls: false,
        } as unknown as ICameraHelper;
        manager._handleMapControls(options);

        expect(map.cameraHelper.handleMapControlsRollPitchBearingZoom).toHaveBeenCalledWith(options.deltasForHelper, options.tr);
        expect(map.cameraHelper.handleMapControlsPan).toHaveBeenCalledWith(options.deltasForHelper, options.tr, options.preZoomAroundLoc);
        expect(map._elevationFreeze).toBe(false);
        expect(manager._terrainMovement).toBe(false);
        expect(setCenterMock).not.toHaveBeenCalled();
    });
```

src/ui/handler_manager.test.ts
birkskyum: They are using the beforeMapTest, createMap now, and checking values like map._elevationFreeze

birkskyum:
Thanks - I have updated to use the beforeMapTest, createMap instead of mocking

HarelM (approved):
Thanks!!

PR #6476, opened 2025-09-29
Fix negative index due to integer overflow in symbol placement

### LINKED ISSUE #3538: Uncaught TypeError: Cannot read properties of undefined (reading '_mult')
I sometimes get this error when displayed the three.js model in mapbox using maplibre-gl and zooming the model. 
Could you please check this issue?
I think you prefer to add the conditional code before _mult.

HarelM:
## Launch Checklist

- Fixes #3538

This checks in the case of integer overflow to avoid continue placing symbols, this should avoid a crash in iOS and an error in chrome.

 - [x] Link to related issues.


HarelM:
@ChrisLoer I know you wrote about this issue in the past, and while I know this is not a full solution as you described I think it would still make sense just in order to prevent rendering issues or crashes.
What do you think?

birkskyum (approved):

HarelM:
I've added a unit test since the render test simply times out, and the actual image is a blank one, so I'm not sure if I should keep this render test.
My gut feeling is that I should remove it, but I'll be happy to know what you think.

HarelM:
I want to release this as part of a version today, so do let me know if there are fixes later on and I'll do them as part of a different PR.

ChrisLoer:
>@ChrisLoer I know you wrote about this issue in the past, and while I know this is not a full solution as you described I think it would still make sense just in order to prevent rendering issues or crashes.
What do you think?

Yeah I think a small defensive change improves the situation a bit here. 👍

Issue #6192, opened 2025-07-29
Bad performance with many overlapping symbols


rubengees:
**maplibre-gl-js version**: 5.6.1

**browser**: Any (tested on Firefox and Chromium)

### Steps to Trigger Behavior

 1. Add a layer with `icon-allow-overlap` and add many features to it.
 2. Zoom out so they overlap.

### Link to Demonstration

https://jsbin.com/cixafubugo/edit?html,output

### Expected Behavior

Zooming is lag-free.

### Actual Behavior

Zooming when many features overlap has heavy lag. The further zoomed out the worse it is. Moving the map or rotating it works good.

Here is a screenshot from the Firefox profiler:

<img width="2687" height="720" alt="Image" src="https://github.com/user-attachments/assets/de3919d8-1735-4a8d-b2ad-73332284e55a" />

Almost all time is spent in `findMatches`. Is optimization possible there?

tonysmid:
We are experiencing the same issue. Indeed in the `findMatches` method, returning the symbols from a KDBush index is slow in this case. A lot of the same items at the same place is makes the index inefficient probably. Any chance updating the KDBush node-size would help?

<img width="781" height="714" alt="profiler results" src="https://github.com/user-attachments/assets/2ed68725-8509-4c07-bf11-113ecace44da" />

HarelM:
Thanks for taking the time to open this issue.
This is not new to recent versions, right?

rubengees:
> Thanks for taking the time to open this issue. This is not new to recent versions, right?

Thanks for answering! Yes, right, I can also reproduce on older versions like 4.7.1 that we are using.

HarelM:
Thanks for looking into previous version. Since this is not a regression it probably means it was always like that. Feel free to take a stab at fixing it. Let me know if you need any help with setting up the project.

HatulaPro:
I expermimented with it a bit. The `findMatches` function calls `KDBush.range`, which returns an array with *all* the matches. But we only need the first one that fits (i think?), so we're probably doing some extra work there. 

I wrote a dirty range function that works with yield and it seemed to make a significant imrpovement. 

Opened an issue in KDBush to see what we can do:
https://github.com/mourner/kdbush/issues/49

HarelM:
I would encourage to open a PR and see if it gets picked up. 
You can also try and use patch-package on this repo and run the benchmarks to see if it creates overhead in a different place and also see if it solves the problem. 
[...]

PR #6492, opened 2025-10-03
Patch KDBush to make symbol matching faster

HatulaPro:
Hopefully solves https://github.com/maplibre/maplibre-gl-js/issues/6192

As explained in the issue, KDBush currently has no API option to return the "first good match" (or something similar). This means that when finding matches for symbols, we're finding _all possible matches_, and only then filtering for really only the one that fits, so there's a lot of extra work there we can skip. I tried opening a PR for KDBush to implement a solution (https://github.com/mourner/kdbush/pull/50), but in the meantime, I'm wondering if my approach even works. 

I made a small POC with the changes here, using a `rangeSome` function (as opposed to the old `range`). On my machine it seems to make the `findMatches` function significantly faster:

New:
image

Old: 
image

In both cases I just opened many-overlapping-symbols.html and zoomed out. I tried running the benchmark itself, but it keep crushing for some reason.

## Launch Checklist

 - [X] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/patches/kdbush+4.0.2.patch
+++ b/patches/kdbush+4.0.2.patch
@@ -1,0 +1,76 @@
+diff --git a/node_modules/kdbush/index.d.ts b/node_modules/kdbush/index.d.ts
+index b874ea6..f9142d9 100644
+--- a/node_modules/kdbush/index.d.ts
++++ b/node_modules/kdbush/index.d.ts
+@@ -41,6 +41,7 @@ export default class KDBush {
+      * @returns {number[]} An array of indices correponding to the found items.
+      */
+     range(minX: number, minY: number, maxX: number, maxY: number): number[];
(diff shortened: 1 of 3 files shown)


HatulaPro:
Small note: my KDBush PR is not as tailor-made to maplibre's needs, so not sure if we'd get the same performance benefits. Still better than better than none probably

bradymadden97:
Note: worth comparing against https://github.com/maplibre/maplibre-gl-js/issues/6192#issuecomment-3373993549

HarelM:
Is this PR replaced by:
- #6641

Can I close this one?

HatulaPro:
> Is this PR replaced by:
> 
> * [Further optimize cross tile symbol index `findMatches` #6641](https://github.com/maplibre/maplibre-gl-js/pull/6641)
> 
> Can I close this one?

Glad to see it's being fixed :)

PR #6510, opened 2025-10-06
Refactor source data handler

wayofthefuture:
Flatten, add comments, and move data handler...

### DIFF
--- a/src/source/source_cache.ts
+++ b/src/source/source_cache.ts
@@ -157,6 +157,30 @@ export class SourceCache extends Evented {
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


src/source/source_cache.ts
```diff
@@ -157,6 +157,30 @@ export class SourceCache extends Evented {
         return true;
     }
 
+    /** @internal
```
HarelM: Can this method be moved to the original position to reduce "unwanted" changes? It will allow easier review to see that the content of the method hasn't changed.

PR #6515, opened 2025-10-07
Emit sourceDataChanged boolean and optimize geojson events

wayofthefuture:
SourceCache currently relies on the sourceDataChanged boolean for reloading tiles:

https://github.com/maplibre/maplibre-gl-js/blob/d08fc5a10210b36c9beb9a2548e74866cb1c57fa/src/source/source_cache.ts#L1028

However, when changing GeoJSON source data via updateData, the boolean is not emitted (see failing test). This PR properly emits this boolean.

Additionally, updateData double-emits a data event (which causes map.update() to double-execute for re-render) on minor data updates. This was done in order to mark a GeoJSON source as 'loaded' so it can request tiles:

https://github.com/maplibre/maplibre-gl-js/blob/d08fc5a10210b36c9beb9a2548e74866cb1c57fa/src/source/geojson_source.ts#L429-L431

https://github.com/maplibre/maplibre-gl-js/blob/d08fc5a10210b36c9beb9a2548e74866cb1c57fa/src/source/source_cache.ts#L1019-L1022

Without the metadata event, the source will not allow the loading of tiles because it's never marked loaded. Since an empty GeoJSON source is ready for loading of tiles upon its load (even though it is empty at this point), it can be called here and does not need to be called on every single small diff data change. This PR moves this metadata event to the load() portion of geojson_source.

### DIFF
--- a/src/source/geojson_source.test.ts
+++ b/src/source/geojson_source.test.ts
@@ -1,3 +1,4 @@
+import type {StyleSpecification} from '@maplibre/maplibre-gl-style-spec';
 import {describe, test, expect, vi} from 'vitest';
 import {Tile} from './tile';
 import {OverscaledTileID} from './tile_id';
@@ -12,9 +13,11 @@ import {SubdivisionGranularitySetting} from '../render/subdivision_granularity_s
 import {type ActorMessage, MessageType} from '../util/actor_messages';
 import {type Actor} from '../util/actor';
 import {MercatorTransform} from '../geo/projection/mercator_transform';
-import {sleep, waitForEvent} from '../util/test/util';
+import {sleep, waitForEvent, beforeMapTest, createMap as globalCreateMap} from '../util/test/util';
 import {type MapSourceDataEvent} from '../ui/events';
+import {type Source} from './source';
 import {type GeoJSONSourceDiff} from './geojson_source_diff';
+import type {MapDataEvent} from '../ui/events';
 
 const wrapDispatcher = (dispatcher) => {
     return {
@@ -61,6 +64,18 @@ const hawkHill = {
         }
     }]
 } as GeoJSON.GeoJSON;
+
+type MapOptions = {
+    style: StyleSpecification;
+};
+
+function createMap(options: MapOptions) {
+    const container = window.document.createElement('div');
+    window.document.body.appendChild(container);
+    Object.defineProperty(container, 'clientWidth', {value: 512});
+    Object.defineProperty(container, 'clientHeight', {value: 512});
+    return globalCreateMap({container, ...options});
+}
 
 describe('GeoJSONSource.constructor', () => {
     const mapStub = {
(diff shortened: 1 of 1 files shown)


src/source/geojson_source.test.ts
```diff
@@ -62,6 +65,18 @@ const hawkHill = {
     }]
 } as GeoJSON.GeoJSON;
 
+type MapOptions = {
+    style: StyleSpecification;
+};
+
+function createMap(options: MapOptions) {
```
HarelM, line 72: Can't we use the regular method? Or more this one to utils as I have seen it copied from time to time...

src/source/geojson_source.test.ts
```diff
@@ -489,6 +504,58 @@ describe('GeoJSONSource.getData', () => {
...
+                version: 8,
+                sources: {geojson: {type: 'geojson', data: {type: 'FeatureCollection', features: []}}},
+                layers: [{id: 'geojson', type: 'fill', source: 'geojson'}]
+            }
+        });
+        await map.once('load');
+        const source: GeoJSONSource = map.getSource('geojson');
+
+        let sourceDataChanged = false;
+        (source as Source).fire = (event: MapSourceDataEvent) => {
```
HarelM, line 520: Can we use a spy here and avoid the casting?

src/source/geojson_source.test.ts
```diff
@@ -489,6 +504,58 @@ describe('GeoJSONSource.getData', () => {
...
+                version: 8,
+                sources: {geojson: {type: 'geojson', data: {type: 'FeatureCollection', features: []}}},
+                layers: [{id: 'geojson', type: 'fill', source: 'geojson'}]
+            }
+        });
+        await map.once('load');
+        const source: GeoJSONSource = map.getSource('geojson');
+
+        let fireCount = 0;
+        (source as Source).fire = (event: MapDataEvent) => {
```
HarelM, line 546: Use a spy

src/source/geojson_source.test.ts
```diff
@@ -563,8 +630,8 @@ describe('GeoJSONSource.updateData', () => {
         source.updateData(update2);
 
         // Wait for the setData and updateData calls to be performed
-        await waitForEvent(source, 'data', (e: MapSourceDataEvent) => e.sourceDataType === 'metadata');
-        await waitForEvent(source, 'data', (e: MapSourceDataEvent) => e.sourceDataType === 'metadata');
+        await waitForEvent(source, 'data', (e: MapSourceDataEvent) => e.sourceDataType === 'content');
+        await waitForEvent(source, 'data', (e: MapSourceDataEvent) => e.sourceDataType === 'content');
```
HarelM, line 634: Are all these changes considered a breaking change due to how the events are fired now?

src/source/geojson_source.ts
```diff
@@ -212,6 +212,8 @@ export class GeoJSONSource extends Evented implements Source {
     }
 
     async load() {
+        // let SourceCache know it's ok to start requesting tiles
+        this.fire(new Event('data', {dataType: 'source', sourceDataType: 'metadata'}));
```
HarelM, line 216: Should this be after the updating the worker with initial data?

src/source/geojson_source.test.ts
wayofthefuture: How to spy on the actual event data to see what it contains?

src/source/geojson_source.test.ts
wayofthefuture: I don't believe so...

src/source/geojson_source.ts
wayofthefuture: Maybe... I'll think about it

src/source/geojson_source.test.ts
HarelM: Something like const spy = vi.spyOn(source, "fire").
And later on check what spy invocation happened.

wayofthefuture: I have to get to the object that was provided as a parameter somehow

HarelM: Which object?

wayofthefuture: The event data object to see its properties

HarelM: Are we mixing two conversations together?
The initial question was about the usage of `createMap` and why it was added and if it is really needed or we can use the "regular" method to create a map.

Regarding the spy object you can use spy.mock.calls (IIRC) to get the call that were used and filter those according to parameters or what not...

wayofthefuture: You were seeking the regular method which I departed due to needing access to the object... hence the removal of createMap if I can gain access to the object in the event. Anyway I'll look more into this I'll see what I can do with spy.mock.calls...

wayofthefuture:
Sorry accidentally deleted it and had to restore...

HarelM:
Was this intentional this time?

wayofthefuture:
Yes it was! It is kind of stale now

PR #6517, opened 2025-10-08
Adding role to marker to satisfy aria-label accessibility issue (fix: #6435)

### LINKED ISSUE #356: [accessibility] Interactive markers should be `role="button"`
Interactive markers (e.g. markers that bind to popups) are focusable and actionable elements, and should have appropriate semantics (i.e. [`role="button"`](https://www.w3.org/TR/wai-aria/#button)).

Related WCAG 2.1 Success Criterion: [4.1.2 Name, Role, Value](https://www.w3.org/TR/WCAG21/#name-role-value) (this issue is to address the <q>Role</q> part of the Success Criterion for interactive markers).

<hr>

(The same issue was also reported with MapBox: https://github.com/mapbox/mapbox-gl-js/issues/11031.)

### LINKED ISSUE #6435: aria-label incorrectly set on markers with no role attribute causing accessibility issues
Map markers use `div` elements with an `aria-label="Map marker"` and no role which is not recommended from an accessibility standpoint. Markers should either not use `div`, a proper `role` attribute should be assigned or no `aria-label` should be set.

**maplibre-gl-js version**: 5.7.2

**browser**: All

### Steps to Trigger Behavior

 1. Render a marker on the map
 2. Inspect marker

### Link to Demonstration

https://maplibre.org/maplibre-gl-js/docs/examples/add-a-default-marker/

The issue can be clearly seen on the maplibre example by just inspecting the marker on the map. 

Image

### Expected Behavior

`div` with `role` attribute and `aria-label` OR element that accepts `aria-label` OR no `aria-label` supplied

### Actual Behavior

`div` with no `role` and default `aria-label` of "Map marker"

### LINKED ISSUE #7627: [Accessibility] aria-prohibited-attr — Marker div uses aria-label without a valid ARIA role
maplibre-gl-js version:  "maplibre-gl": "^4.5.2",

browser: All (violation is in the DOM output, not browser-specific)

Steps to Trigger Behavior
Create a map with a default Marker (e.g. new maplibregl.Marker().setLngLat([0,0]).addTo(map))
Run an axe-core ≥ 4.11 accessibility audit (via Cypress-axe, Lighthouse, or browser DevTools)
Observe aria-prohibited-attr violation on .maplibregl-marker

Expected Behavior
The marker element should use a valid ARIA role that permits aria-label, allowing the accessible name to be communicated to assistive technologies without violating WCAG 2.1 SC 4.1.2 (Name, Role, Value).

Actual Behavior
The marker renders as:
```
<div class="maplibregl-marker" aria-label="Map marker" style="transform: translate(...)">
```
A plain `<div>` has an implicit role of generic, which [prohibits naming via aria-label](https://www.w3.org/TR/wai-aria-1.2/#generic). axe-core reports:

[SERIOUS] aria-prohibited-attr
  WCAG tags: wcag2a, wcag412
  Nodes: .maplibregl-marker
  Fix: aria-label attribute cannot be used on a div with no valid role attribute.
More info: https://dequeuniversity.com/rules/axe/4.11/aria-prohibited-attr?application=axeAPI

Related Issues
#356 — Interactive markers should be role="button"
#362 — Consider conveying only one of aria-label and title in controls

cmburcus:
## Launch Checklist

In March of 2025, `aria-label` was added by default to markers even if not set when implementing. Since the marker is a `div` which has no role by default, this causes accessibility audits to pick it up as an issue.

Instead of removing the `aria-label` since it should likely be there, I'm adding a role of button to the marker.

Related issue: https://github.com/maplibre/maplibre-gl-js/issues/6435

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/ui/marker.ts
+++ b/src/ui/marker.ts
@@ -320,6 +320,10 @@ export class Marker extends Evented {
             this._element.setAttribute('aria-label', map._getUIString('Marker.Title'));
         }
 
+        // aria-label is set either by user or above default, so set role
+        // since div is interactive and cannot have aria-label without a role
+        this._element.setAttribute('role', 'button');
+
         map.getCanvasContainer().appendChild(this._element);
         map.on('move', this._update);
         map.on('moveend', this._update);
--- a/src/ui/marker.ts
+++ b/src/ui/marker.ts
@@ -322,7 +322,9 @@ export class Marker extends Evented {
 
         // aria-label is set either by user or above default, so set role
         // since div is interactive and cannot have aria-label without a role
-        this._element.setAttribute('role', 'button');
+        if (!this._element.hasAttribute('role')) {
+            this._element.setAttribute('role', 'button');
+        }
 
         map.getCanvasContainer().appendChild(this._element);
         map.on('move', this._update);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 - Improve fading - dynamic bi-directional raster cross-fading and self fading ([#6469](https://github.com/maplibre/maplibre-gl-js/pull/6469))
 
 ### 🐞 Bug fixes
-- _...Add new stuff here..._
+- Added `button` role to marker div to fix accessibility issues with the `aria-label` ([#6435](https://github.com/maplibre/maplibre-gl-js/issues/6435))
 
 ## 5.8.0
 
--- a/src/ui/marker.test.ts
+++ b/src/ui/marker.test.ts
@@ -290,6 +290,13 @@ describe('marker', () => {
         expect(markerWithHtmlElement.getElement().getAttribute('aria-label')).toBe('custom aria label');
     });
 
+    test('Marker should have a role attribute to satisfy accessibility requirements for the aria-label', () => {
+        const map = createMap({locale: {'Marker.Title': 'alt title'}});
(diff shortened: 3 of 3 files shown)


src/ui/marker.ts
```diff
@@ -320,6 +320,10 @@ export class Marker extends Evented {
             this._element.setAttribute('aria-label', map._getUIString('Marker.Title'));
         }
 
+        // aria-label is set either by user or above default, so set role
+        // since div is interactive and cannot have aria-label without a role
+        this._element.setAttribute('role', 'button');
```
HarelM: Can't this override a role set by the user? Maybe only set it inside the if?

cmburcus: Good point. Updated

HarelM (approved):
Thanks!!

skorasaurus:
Thank you for implementing this :heart:

andrewharvey:
Hang on, this sets `role=button` even for non interactive visual only markers. `role=button` is for 

> An input that allows for user-triggered actions when clicked or pressed.

source: https://www.w3.org/TR/wai-aria-1.2/#button

If the marker is not interactive or clickable (as is the default if you do `new Marker().setLngLat([0, 0]).addTo(map);` then it should not have a `role=button.`

I came here after looking at #7627.

Issue #6427, opened 2025-09-17
v6 breaking changes features


HarelM:
This is a placeholder issue for things that we might consider for v6 as a breaking changes.

- [ ] #7941 - In continue to the style spec changes
- [x] https://github.com/maplibre/maplibre-style-spec/pull/1740 - Improve expression evaluation warnings: 
- [x] #7902 - change styleimagemissing to be an event only
- [x] #7796 - this is needed to support shared part when loading a module.
- [x] Replace map - camera inheritance with composition? or simply remove the camera class. #7800 
- [x] #7789 - Change events to use classes, related to render and viewporttiles event PRs that were recently open and then closed.
- [x] https://github.com/maplibre/maplibre-gl-js/pull/6811#issuecomment-3615666795 - from the comment here: I'd love to see `GeoJSONSource#serialize` and `Map#getStyle` become async and pull this giant dataset from the worker only when it's requested - I'm not 100% sure about this to be honest...
- [x] https://github.com/maplibre/maplibre-style-spec/pull/1545 - deprecated expression validation
- [x] #7537 - Change the default to be 4 instead of not using this feature, has a bug that needs to be solved.
- [x] #6872 - I don't think there's a need to do that
- [x] #6254 - Add ESM build, this will break Angular default behavior since it uses ESM by default and there's a need to load the worker code separately there.
- [x] #7538 - remove wait for completion and return a promise, maybe change other return values to avoid returning "this".
- [x] ~#6972 - Change the signature of `source.loadTile` method to return a boolean~ tired to do it, not worth the effort I beleive.
- [x] #6992 - Changes how geojson is returned from query render features - merging this will require a breaking change version
- [x] #7073 - change hash to use urlsearchpatams
- [x] #7258
- [x] bumping from ES2016 to reduce the bundle size #7404
- [x] Drop Webgl1 support #7453
  Because this allows the luma.gl migration to go forward and some performance fixes for example in terrain3d

Issue #5618, opened 2025-03-12
Support variable depth pyramids


manuel-hegner:
## User Story

As a developer I can use variable depth pyramid tilesets, so that I can improve performance and reduce network traffic.

## Rationale

- Variable depth pmtiles datasets are much smaller than normal ones and make fewer requests to the server.
- Tippecanoe recently introduced `--generate-variable-depth-tile-pyramid` which is gracefully supported by pmtiles. Currently the pmtiles Protocol implementation returns an `data:Uint8Array` of length zero for nonexistent tiles which maplibre interprets as empty water instead of the intended "this tile does not exist but this is no error".
- As a workaround the pmtiles protocol can be wrapped to to return a rejected promise instead. This currently triggers a lot of error events in maplibre, but otherwise works mostly. Maplibre will not try to use larger tiles aggressively enough (often only checking 3 zoom levels).
- Maplibre should add to its Protocol API a way to gracefully return "this tile does not exist as expected, please use overzoom instead". Ideally this would even support a returned hint of what zoom level does exist.

spatialillusions:
I ran into similar problems when implementing a custom protocol for the VTPK format that uses a vector tile index to minimize the number of tile requests. 

This shows an example of how the tile coverage for a VTPK file can look.
![Image](https://github.com/user-attachments/assets/08a019ff-bcc3-4bdd-801c-e625ca0dbb67)
(From https://pro.arcgis.com/en/pro-app/3.3/tool-reference/data-management/create-vector-tile-index.htm) 

I found out in the pmtiles protocol that it is possible to get tiles from another zoom level to be reused, if you throw an error instead of returning ```data:Uint8Array```, and that this can be done by the pmtiles protocol if you set the ```errorOnMissingTile ``` property when you create the protocol:
 ``` 
new pmtiles.Protocoll({
    errorOnMissingTile: true,
});
```

So far so good, and this will give you something like this:
![Image](https://github.com/user-attachments/assets/185f67a8-23fe-416d-8789-8500b1da605e)

Note that the "empty" tiles are not in fact empty, you can see this because there are some labels displayed, and also a tiny body of water. The reason for not more data showing up, is that the style used has a lot of ```minzoom```s set, and if the style has a ```minzoom:16``` set, nothing for these over scaled tiles will be shown since they are zoom level 14/15.

[...]

PR #6521, opened 2025-10-09
Split tiles on-the-fly to improve over scale vector handling

### LINKED ISSUE #2507: Proposal: Better support for overzooming via very large tile extents
MapLibre currently "supports" up to 10-levels of overzooming (e.g. `SourceCache.maxOverzooming = 10;`), but its actual behavior at that level of overzooming is not very good -- it essentially makes gigantic 524,288px square tile and then renders a small portion of it on the viewport. There are I think two main problems with this:

- The WebGL/shader-enforced tile extent of 8192 (or 16384 after https://github.com/maplibre/maplibre-gl-js/pull/1244) means that each tile unit covers up to 64 pixels -- that makes for a very imprecise map!
- Elements line line labels that repeat based on the overzoomed tile scale end up with an enormous number of copies to fill the gigantic tile. This is bad for performance, but also once more than 64k worth of glyphs are generated for a tile, you'll [start seeing garbled labels and potentially even throw exceptions during rendering](https://github.com/mapbox/mapbox-gl-js/issues/2907#issuecomment-1444606857).

Screenshot 2023-05-11 at 12 40 40 PM

## High overzooming is unavoidable in non-giant tilesets
The vector tile spec supports extents going all the way up to max_uint32, but why is it useful to have such large extents? An extent of 8192 works fine for tilesets that go out to z15 or z16. The problem is that making tile sets that go all the way out to z15 is expensive (more than a billion tiles!). Making those giant tile sets is just part of the job for global basemaps like Mapbox Streets or MapTiler Planet, but when you're working on data visualizations, there are lots of datasets that are: (1) big enough to need tiling, and (2) small enough that you should expect to generate a tileset in minutes without needing a compute cluster.

Using a large tile extent allows you to tile your dataset only as far as is necessary to keep maxzoom tile data size within your target limits, and then by increasing the extent of your maxzoom tiles, you can avoid giving up any precision. This is what Felt does with Tippecanoe when we tile user uploads to our pipeline -- unfortunately, when we moved from Protomaps.js to MapLibre, we lost the precision because the extent gets converted down to 8192 for display. Moving to MapLibre was great, but this is the biggest remaining regression for us.

## Proposed solution: reuse GeoJSON-VT
Luckily, MapLibre already pretty much knows how to do good overzooming for high precision tiles, because it does it for GeoJSON sources, which are essentially just a single-tile source with very high precision. The `geojson-vt` library takes a single high-precision tile (the GeoJSON) as an input, and then answers requests for Vector Tiles from MapLibre by chopping up the source tile into smaller geometries on the fly. Points and lines are easy to do (just take the subset within the new tile bounds), but polygons are a bit trickier because you have to make new polygons that fit within the new tile. `geojson-vt` makes this fast in part by maintaining an in-memory tile-pyramid so that each new tile can take advantage of any chopping work that already happened above it in the pyramid.

My proposal is to adapt `geojson-vt` so that instead of just working with GeoJSON, it can be used for client-side overzooming of _any_ high detail tile. Imagine a tileset that's tile to z8, with the tiles at z8 having an extent of 65536:

- From z1 to z8, MapLibre just shows the tiles using the existing logic (the extra precision in the extent is thrown away at z8, but it doesn't matter at this zoom level).
- At z9, when MapLibre requests a tile from the source, it finds the nearest parent (in this case, the z8 tile), loads it, and then uses `geojson-vt` to generate a z9 tile.
- At zN, it can take advantage of the zN-1 tiles that are already in memory to generate a zN tile
- As long as any tiles that are children of the original z8 tile are in the _map's_ tile pyramid, keep the z8 tile in as well.
- Once the z8 tile leaves the _map's_ pyramid, you also clear the in-memory `geojson-vt` pyramid.

I prototyped this locally (without doing the pyramid cleanup parts), and it seemed to work pretty well -- took about a day to get working. I did have to reach into some `geojson-vt` internals a little bit, so we would probably want to put a PR into `geojson-vt` or fork it in order to enable this.

## Variants
There are a few different ways I can imagine this plugging into MapLibre

### Add this behavior directly to Vector Tile Source
My preferred solution here would be to just build this into the existing Vector Tile Source logic. Nothing about the existing style spec would have to change, the only difference would be that at source maxzoom, the Vector Tile Source would watch for high-extent tiles, and if it found them it would deduce "I have enough precision to provide overzooming for X more zoom levels" -- it would tell the SourceCache (so that the "covering tiles" logic knew to keep requesting tiles), and then provide the overzoomed tiles using the `geojson-vt` behavior.

I like this approach because:
- No change to style spec/docs/etc.
- Minimal extra bundle size (most of the code needed is already included in `geojson-vt`)
- No change to behavior for sources that aren't using high extents (which is probably almost all of them, since this hasn't worked well previously!)
- The change to behavior for sources _with_ high extents should be a pure improvement

### Add a new Source type derived from Vector Tile Source
My prototype implemented this behavior as a "HighDetailSource" that extended "VectorTileSource". The source was defined with a max "network" zoom as well as a regular maxzoom (where the "network" zoom determined when it decided to start making tiles locally instead of asking the network). Doing it this way avoids having to touch the SourceCache logic at all or do any reasoning about what the maxzoom should be based on the extent -- but it makes the user specify two zoom levels, and adds a whole new source type to the style spec, so it feels a bit clunky to me.

### Make Sources more "pluggable"
This is kind of a meta-solution, but if we think this functionality is too "niche" to be a core part of MapLibre, you could imagine supporting extra source types as a "plugin". I haven't really sketched out what this would look like, but I think at the very least you want access to all vector-tile->gl-bucket logic that's in VectorTileSource. This is already half-exposed with the `LoadVectorData` callback, but to make the "high detail source" work you would have to expose at least a few more things (like hooking `unloadTile` so you'd know when to do cleanup).

## Rationale / Impact
I think the main workarounds here are:

- Use GeoJSON (but this becomes less feasible as datasources get larger)
- Generate your tilesets to higher zoom if you want more precision (but this is expensive)
- Build overzooming support into your tileserver instead of into the client (this is an option we're looking at in Felt, but I think this is a higher impact improvement if we can do it in maplibre instead of every different user of maplibre having to figure out how to do it in their respective tile servers)

And of course, the status quo has worked for years despite its shortcomings! I _know_ Felt has a use for this, and I think it's "medium size vector tilesets" are a very real use case for lots of data visualization, but I can't really make the case for what other groups would use this off the bat.

cc @wipfli @HarelM @ibesora @bdon @dnomadb

### LINKED ISSUE #6524: Safari crashes when zoomed in and there are a lot of labels
**maplibre-gl-js version**: 5.8.0

**browser**: Safari

### Steps to Trigger Behavior

 1. Surf to the following jsbin:
 2. https://jsbin.com/fewoduguto/edit?html,output or https://jsbin.com/daqoyowica/edit?html,console,output
 3. Wait a few seconds

### Link to Demonstration

Minimal reproduction with a single layer:
https://jsbin.com/daqoyowica/edit?html,console,output
With a full style:
https://jsbin.com/fewoduguto/edit?html,output

### Expected Behavior
It should not crash safari

### Actual Behavior
White screen of death

### Additional Information
I believe this is related to the warning that there are too many glyphs, and the fact that they are sent to the GPU in safari is what causes this crash, but this is only a hunch.
I've opened a bug report for Apple on this, but I'm not expecting Apple to actually solve this in the near future.

- This is also related to #2507

HarelM:
## Launch Checklist

- Resolves #2507
- Fixes #6524

This is an experimental feature allowing to over-scale vector tiles by slicing them instead of over scaling them for a given number of zoom levels.
The flag is called `experimentalZoomLevelsToOverscale`.
This feature will slice vector tiles after the source's max zoom in the worker until it reaches the zoom level of `map's max zoom - experimentalZoomLevelsToOverscale`.
This allow a few last layers to be scaled to improve rendering and reduce repetition of polygon centered labels while the rest are sliced to improve performance.
The performance of this seems to work well and it solves the crash on iOS as far as my tests went when setting it to 4.
The default is undefined to keep rendering backward compatibility.
When setting to a number bigger than 0 it will slice tiles and by extension polygons and may slightly change how the map looks if there are polygons that has a label in their center (because the polygon is now sliced after source's max zoom and the center may change depending on the zoom level and the slice size).

- We should consider setting a different default value as part of #6427, maybe.

CC: @wayofthefuture @ChrisLoer 

 - [x] Link to related issues.

### DIFF
--- a/src/source/vector_tile_worker_source.ts
+++ b/src/source/vector_tile_worker_source.ts
@@ -44,7 +44,7 @@ export class VectorTileWorkerSource implements WorkerSource {
     fetching: {[_: string]: FetchingState };
     loading: {[_: string]: WorkerTile};
     loaded: {[_: string]: WorkerTile};
-    overzoomedTilesCache: BoundedLRUCache<string, LoadVectorTileResult>;
+    overzoomedTileResultCache: BoundedLRUCache<string, LoadVectorTileResult>;
 
     /**
      * @param loadVectorData - Optional method for custom loading of a VectorTile
@@ -59,7 +59,7 @@ export class VectorTileWorkerSource implements WorkerSource {
         this.fetching = {};
         this.loading = {};
         this.loaded = {};
-        this.overzoomedTilesCache = new BoundedLRUCache<string, LoadVectorTileResult>(1000);
+        this.overzoomedTileResultCache = new BoundedLRUCache<string, LoadVectorTileResult>(1000);
     }
 
     /**
@@ -167,7 +167,7 @@ export class VectorTileWorkerSource implements WorkerSource {
         const {maxZoomTileID} = overzoomParameters;
 
         const cacheKey = `${maxZoomTileID.key}_${tileID.key}`;
-        const cachedOverzoomTile = this.overzoomedTilesCache.get(cacheKey);
+        const cachedOverzoomTile = this.overzoomedTileResultCache.get(cacheKey);
         
         if (cachedOverzoomTile) {
             return cachedOverzoomTile;
@@ -188,7 +188,7 @@ export class VectorTileWorkerSource implements WorkerSource {
             }
         }
         const overzoomedVectorTileResult = toVirtualVectorTile(overzoomedVectorTile);
-        this.overzoomedTilesCache.set(cacheKey, overzoomedVectorTileResult);
+        this.overzoomedTileResultCache.set(cacheKey, overzoomedVectorTileResult);
 
         return overzoomedVectorTileResult;
     }


HarelM:
All the failing render test are failing due to placement of labels, which is expected as part of this PR.
I've looks at all the labels diff and I can't say which one is more "correct" between the old and the new, so I tend to say that this is OK.
The query tests are failing too, from what I looked on a single test the query geometry is returned as a sliced geometry instead of the full one, which I think is also expected as part of this change, but can be considered a breaking change for anyone that expects to get the geometry from the tile at max zoom and not a further down sliced geometry.

src/source/overzoomed-geojson-vector-tile.ts
```diff
@@ -0,0 +1,86 @@
+import type geojsonvt from 'geojson-vt';
```
wayofthefuture: filename typo

wayofthefuture:
Like a pro!

HarelM:
I'm thinking about intruducing an experimental flag to the map options and maybe remove it as part if version 6, what do you think?

wayofthefuture:
Yes that's a good idea... but you would have to add back the isSafari check...

HarelM:
Oh, yeah, forgot the main motivation around this feature was the Safari crash...
I don't know what to do then.
Maybe enable this flag by default for Safari then?
Ooof...

wayofthefuture:
Yeah I think so

HarelM:
This is now ready for review.

src/ui/map.ts
```diff
@@ -361,6 +361,15 @@ export type MapOptions = {
      * keep the camera above ground when pitch \> 90 degrees.
      */
     centerClampedToGround?: boolean;
+    /**
+     * Allows overzooming using geojson-vt for vector tile sources. 
+     * If `true`, tiles over the source's maxzoom will be split using geojson-vt into subtiles.
+     * if `false`, tiles will be overzoomed using scaling.
+     * @defaultValue `true` for Safari and false for other browsers due to Safari crash issues.
+     * This may change or be removed in future versions.
+     * @experimental
```
CommanderStorm: Could we include a sentence on pros/cons of this option?
Seems hard four our users to choose otherwise

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## main
 
 ### ✨ Features and improvements
+- Use geojson-vt to improve over scale vector handling ([#6521](https://github.com/maplibre/maplibre-gl-js/pull/6521))
```
CommanderStorm: lets be more concrete what this affects

```suggestion
- Add the `experimentalOverzoomingWithGeojsonVt` option (`true` on safari, `false` elsewere) to improve overzoomed vector tile handling. If set to `false`, we scale instead of partitioning ([#6521](https://github.com/maplibre/maplibre-gl-js/pull/6521))
```

src/tile/tile_cache.ts
```diff
@@ -210,3 +210,38 @@ export class TileCache {
...
+            const oldestKey = this.map.keys().next().value;
+            this.map.delete(oldestKey);
+        }
+        this.map.set(key, value);
+    }
+
+    clear(): void {
+        this.map.clear();
+    }
+}
```
CommanderStorm, line 247: I don't like that this design means that every `get` has to do a `Map.get`, `Map.delete`, `Map.set` => very high churn.

It is fairly simple though and I would need a benchmark to see if optimising this is actually a bottleneck.

src/source/tile_cache.ts
```diff
@@ -210,3 +210,38 @@ export class TileCache {
         }
     }
 }
+
+export class BoundedLRUCache<K, V> {
+    private maxEntries: number;
+    private map: Map<K, V>;
+
+    constructor(maxEntries: number = 1000) {
```
CommanderStorm: All usage instances have this explicitly defined

```suggestion
    constructor(maxEntries: number) {
```

src/source/vector_tile_worker_source.ts
```diff
@@ -144,6 +161,67 @@ export class VectorTileWorkerSource implements WorkerSource {
...
+            geoJSONFeatures.push(feature.toGeoJSON(maxZoomTileID.x, maxZoomTileID.y, maxZoomTileID.z));
+        }
+
+        return geojsonvt({
+            type: 'FeatureCollection',
+            features: geoJSONFeatures
+        }, {
+            extent: EXTENT,
+            buffer: this._pixelsToTileUnits(128, tileSize),
+            maxZoom: maxOverzoom,
```
wayofthefuture: If a user calls map.setMaxZoom() to a higher number than current map's maxZoom, then you would have geojsonvt in the cache that won't return tiles for the greater Z. So you can add the maxzoom to the cache key or just hard-code maxZoom: 24 as geojsonvt generates tiles "on the fly" up to maxzoom. Then you wouldn't need to pass it at all through overzoom parameters...

src/tile/tile_cache.ts
wayofthefuture: I believe those are all O(1) operations... but nevertheless there is also this npm package which slams at 240 million weekly downloads: https://www.npmjs.com/package/lru-cache

wayofthefuture: I did a little more research and I like our simple implementation. All ops are O(1). I also found a gotcha in our current source_cache which carries an O(n) lookup in all layers and all tiles, which has great opportunity for performance increase. We currently retrieve tiles from the cache using O(n) Line 922: tile = this._cache.getAndRemove(tileID); We may want to consider also switching source_cache to the new lru cache. (@HarelM ?) Here is a ChatGPT summary of the difference between the current tilecache and the new lru cache:

image

HarelM: This can be done in another PR, I generally prefer libraries over code here, but this is a very small implementation so we might as well keep it, I'm just not sure map implementation really tracks the order...

wayofthefuture: Map implementation tracks the order of the keys, which are kept "in insertion order"... 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/keys.
https://medium.com/sparkles-blog/a-simple-lru-cache-in-typescript-cba0d9807c40

Delete/set moves the key to the end of keys() and next() gets the oldest key...

HarelM: Cool, didn't know that.

src/source/overzoomed-geojson-vector-tile.ts
HarelM: I'm not sure there is a typo, what am I missing?

HarelM:
> I don't fully understand all the code (src/source/overzoomed-goejson-vector-tile.ts), but that is too be expected.

The code part of this is less interesting, the interesting part is the fact that the client size is doing overzooming by splitting the original tiles, I think this feature can also be used in Martin if it's not already there - i.e. allow generating more zoom levels on the fly and cache them.

> Before we merge, I would like
> - a testcase that covers the new experimentalOverzoomingWithGeojsonVt option, or is this not the norm here

I've tested this with automatically enabling this flag (before it existed) and made sure all the tests are behaving as expected, I don't want to introduce tests that will be removed later on as I would like to avoid keeping this experimental flag for long. So my plan is either give it a go for a few versions and then remove it or not introduce this flag at all, I want to get some feedback from the community on this as I'm not 100% sure what's the right thing to do here. 

> - a bit more guideance in the doc strings when this is benfitial/ why to activate/deactivate this

Let me know if the changelog and tsdoc comment covers this request.

src/source/vector_tile_worker_source.ts
HarelM: Ah, interesting observation, I did not though about this scenario.

HarelM: But this would mean that there's a need to clear the cache if the max zoom had changed, doesn't it?

HarelM: I've added it to the key, but I don't think it's bullet proof...

src/ui/map.ts
```diff
@@ -370,10 +370,12 @@ export type MapOptions = {
...
     /**
-     * Allows overzooming using geojson-vt for vector tile sources. 
-     * If `true`, tiles over the source's maxzoom will be split using geojson-vt into subtiles.
+     * Allows overzooming using geojson-vt for vector tile sources.
+     * When set to `true` this seems to have better performance at high zoom levels and prevents Safari from crashing.
+     * The default is `false` for most cases as it changes query render features results in high zoom levels due to tile splitting.
+     * If `true`, tiles over the source's maxzoom will be split using geojson-vt into subtiles (partitioning).
      * if `false`, tiles will be overzoomed using scaling.
-     * @defaultValue `true` for Safari and false for other browsers due to Safari crash issues.
+     * @defaultValue `true` for Safari and `false` for other browsers due to Safari crash issues.
```
CommanderStorm, line 273:
```suggestion
     * @defaultValue `true` for Safari to prevent crashes and `false` for other browsers
```

src/source/overzoomed-geojson-vector-tile.ts
CommanderStorm: goejson should be spelled geojson, right?

HarelM:
I have such a blind spot for letter switching, amazing... Thanks!

src/source/overzoomed-geojson-vector-tile.ts
HarelM: It should be... yes...

src/source/vector_tile_worker_source.ts
wayofthefuture: Since it is LRU old maxzooms will be removed eventually. A new maxzoom wouldn't reuse if it's in the cache key so I think that's pretty solid.

HarelM: Right...

HarelM:
@ChrisLoer let me know what you think about this implementation. I saw in your implementation that there was some conversion between tile coordinates and mercator coordinates copied from geojson vt package which is not used here, I'm not sure if this was an optimization of some sort or is the code here is good enough as is.

HarelM:
I've posted a feedback request here:
https://osmus.slack.com/archives/C03TFH5NE83/p1760394977393069

HarelM:
Unfortunately, while this feature is very cool, it still doesn't solve the crash in iOS :-(((

wayofthefuture:
iOS version?

HarelM:
iOS version 18.7.1, iPhone 12.
I tried disabling this experimental feature and setting overscaling in symbol layer to 64 max, but I was still able to make it crash.
I'm not sure if this is an extreme edge case or that the style is "too heavy" or something.
IDK...
I'll continue my investigation tomorrow I guess...

wayofthefuture:
No runtime for 18.7 but iOS 18.6 on iPhone 12 working... very odd, see both working with and without tile borders:


https://github.com/user-attachments/assets/875881ef-65f6-4835-ba86-e20c238a8e67

HarelM:
The crash doesn't happen right away, it happens after panning and zooming around...

wayofthefuture:
Have you tried reducing lru cache size to maybe 100 just to rule that out? My time is limited for the new few days but then I can look more into it...

HarelM:
I haven't, but memory consumption is low, so I don't think it's memory related. 
I need to try reproduce this with a simpler style and see if I can find out if there's a specific layer that causes this issue, and then maybe try and think what can cause this issue and how to solve it...

wayofthefuture:
Ok. I do think that it's maybe another issue than the one we have fixed here however. Not sure if you plan on merging this or waiting...

HarelM:
I'll wait a bit with this, nothing urgent about this PR.

HarelM:
Here's my investigation summary so far:
1. When using only labels layer and disabling the experimental geojson-vt the reproduction below causes white screen.
2. When enabling it and only drawing labels it doesn't crash.
3. When using only contour lines and enabling it, it crashes.
4. When using only contour lines and disabling it, it doesn't crash.

So lines are crashing when using this feature and label are crashing when not using this feature.
I'll need to see what's the cause of this... 

```js
new maplibregl.Map({
  container: 'map',
  style: {
    version: 8,
    center: [-140.42294782037015, 60.61745451249607],
    zoom: 12.941097799870885,
    sources: {
      Contour: {
        type: 'vector',
        url: 'https://global.israelhikingmap.workers.dev/JAXA_AW3D30_2024_contour_z5-Z12_vector.json',
        tiles: [
          'https://global.israelhikingmap.workers.dev/JAXA_AW3D30_2024_contour_z5-Z12_vector/{z}/{x}/{y}.mvt',
        ],
        minzoom: 5,
        maxzoom: 12,
        scheme: 'xyz',
        attribution:
          '<a href=https://earth.jaxa.jp/en/data/policy/>AW3D30 (JAXA)</a>',
      },
    },
    layers: [ // *** use one of the below labels according to the instruction above ***
      {
        id: 'contour-major-casing',
        type: 'line',
        metadata: { 'IHM:overlay': true },
        source: 'Contour',
        'source-layer': 'contours',
        minzoom: 9,
        paint: { 'line-color': 'black', 'line-width': 3, 'line-blur': 2 },
      },
      {
      "id": "contour-label",
      "type": "symbol",
      "metadata": {"IHM:overlay": true},
      "source": "Contour",
      "source-layer": "contours",
      "filter": [
        "let",
        "ele",
        ["to-number", ["get", "ele"]],
        [
          "case",
          [">=", ["zoom"], 14],
          ["==", 0, ["%", ["var", "ele"], 50]],
          ["==", ["zoom"], 13],
          ["==", 0, ["%", ["var", "ele"], 100]],
          ["==", ["zoom"], 12],
          ["==", 0, ["%", ["var", "ele"], 250]],
          ["==", ["zoom"], 11],
          ["==", 0, ["%", ["var", "ele"], 500]],
          ["==", ["zoom"], 10],
          ["==", 0, ["%", ["var", "ele"], 1000]],
          ["==", ["zoom"], 9],
          ["==", 0, ["%", ["var", "ele"], 2500]],
          false
        ]
      ],
      "layout": {
        "text-field": "{ele}",
        "symbol-placement": "line",
        "text-font": ["Open Sans Cond Bold"],
        "symbol-avoid-edges": true,
        "text-max-angle": 45,
        "text-letter-spacing": 0.15,
        "text-keep-upright": false
      },
      "paint": {
        "text-color": "rgb(161, 127, 86)",
        "text-halo-color": "rgba(255, 255, 255,0.8)",
        "text-halo-width": 3,
        "text-halo-blur": 1.5
      }
    },
    ],
  },
});

map.once('load', () => reproduceCrash());

async function reproduceCrash() {
        const locations = `-140.42294782037015, 60.61745451249607, 12.941097799870885
-140.4202786266386, 60.61981958907671, 14.540055784675747
-140.4250208580154, 60.61696090026271, 14.540055784675747
-140.4245078872823, 60.61636524073424, 16.534685060726996
-140.43415660737026, 60.618717383995005, 12.19047050945691
-140.45273597661077, 60.62482825663426, 12.19047050945691
-140.44708315097012, 60.62298385597779, 14.158788133918403
-140.4522547378101, 60.624405627363956, 14.158788133918403
-140.450700101564, 60.62405646917759, 16.387482547275887
-140.450700101564, 60.62405646917759, 22
-140.44708315097012, 60.62298385597779, 22`;

        const lines = locations.split('\n');
        let initialized = false;
        for (const line of lines) {
            const parts = line.trim().split(',').map(Number);
            const coords = [parts[0], parts[1]] as [number, number];
            const zoom = parts[2];
            if (!initialized) {
                this.mapComponent().mapInstance.setZoom(zoom);
                this.mapComponent().mapInstance.setCenter(coords);
                initialized = true;
            } else {
                this.mapComponent().mapInstance.easeTo({ center: coords, zoom });
                await new Promise(resolve => setTimeout(resolve, 1000));
            }
        }
```

HarelM:
It seems that removing the cache prevents iOS from crashing when using goejson-vt on the contour lines, unfortunately it also degrades performance.
I'm not sure what should be a good middle ground here... I'll continue thinking about...

wayofthefuture:
But it's not related to the size of the cache? I'm about to open a PR on geojson-vt if it's a small issue I could attempt to address it there. Have you looked at the error through developer portal on Mac OS safari (while using a physical phone)?

HarelM:
There's no error message, just white screen of death, unfortunately. 
The cache size is 4 in terms of items, but it might be that the items are large. Although I don't think they can be huge given that they index like 4 tiles...
So I'm not sure what's my next step, but if you change the geojsonvt code make sure you test it on iOS...
I'm not sure the problem is there to be honest, but I'm not sure how the internals there look like...
I'll continue my trial and error I guess... So frustrating... 😕

wayofthefuture:
I'll have to look more into it. It's a dynamically generated full tile stack in each item of the cache. So if you zoom to 22 and pan all over the place it's exponentially growing in the cache.

https://github.com/mapbox/geojson-vt/blob/5e880a3954c994d4471f1c1a2afd4ab01050066b/src/index.js#L66

wayofthefuture:
Also may want to try removing the tolerance: 0 option...

Your source max 14 to map max 22 is approaching 4^8 tiles... so maybe cache size should be a function of source -> map maxzoom delta

HarelM:
Actually, the source max zoom is 12, but I see that it is crashing when in zoom 14-16, so I'm having a hard time explaining why it crashes.
When I set the worker count to 1 (which probably means less memory usage) it doesn't crash, but I'm guessing it's a matter of how many indexes are being used so if the number of cached indexes will grow it will crash eventually.
I'm wondering if there's a fast way to slice just the needed part of the vector tile without all the indexing and memory optimization, maybe some feature in geojson-vt where I can give it a list of features and a bbox (tile) and get back just the slice, fast. IDK...
The memory is a guess as when I debug it with xcode I see the app's memory usage at about 15 Mb, which doesn't sound too much, but it might be that it's not counting it correctly since it's using capacitor and safari.
In any case, if you can think of stuff I can check and try let me know.
The good news is that it's easy for me to reproduce the crash, so I can iterate on different solutions fast.

HarelM:
I've enabled debug in geojson-vt and I see that it creates indexes from z0 to z14 where most of the data is duplicated, and there's a lot of points there (although nothing it shouldn't handle I believe).
I'll try and see if there's a way to maybe avoid the toGeoJSON part of the code and back to vector tiles and maybe keep everything in tile coordinates somehow...

HarelM:
So I wrote the following code to "simply" split a tile and take the relevant geometry out of it.
It seems to prevent the crash on iOS. 
I copied some of the code from geojson-vt related to slicing/clipping, so @wayofthefuture if you plan to update the code there it might be worth considering exposing some of the clipping methods, maybe.
For some reason it's not perfect and I'm not sure why, I see both contour labels in the wrong place and some line are not scaled correctly (see image below bottom left corner - orange lines), I'm not sure what I did wrong and I'm not sure it's worth the debugging, so I'm leaving it for now.

image



```ts
import {type VectorTile, VectorTileFeature, VectorTileLayer} from '@mapbox/vector-tile';
import Protobuf from 'pbf';
import {fromVectorTileJs} from '@maplibre/vt-pbf';
import {type LoadVectorTileResult} from './vector_tile_worker_source';
import { CanonicalTileID } from './tile_id';

class OverzoomedFeatureWrapper extends VectorTileFeature {
    myGeometry: Point[][];

    constructor(type: 0 | 1 | 2 | 3, geometry: Point[][], properties: any, id: number, extent: number) {
        super(new Protobuf(), 0, extent, [], []);
        this.type = type;
        this.properties = properties ? properties : {};
        this.extent = extent;
        this.myGeometry = geometry;
        this.id = id;
    }

    loadGeometry() {
        return structuredClone(this.myGeometry);
    }
}

class OverzoomedTileLayer extends VectorTileLayer {
    private _myFeatures: OverzoomedFeatureWrapper[];
    name: string;
    extent: number;
    version: number = 2;
    length: number;

    constructor(features: OverzoomedFeatureWrapper[], layerName: string, extent: number) {
        super(new Protobuf());
        this._myFeatures = features;
        this.name = layerName;
        this.length = features.length;
        this.extent = extent;
    }

    feature(i: number): VectorTileFeature {
        return this._myFeatures[i];
    }
}

export class OverzoomedVectorTile implements VectorTile {
    layers: Record<string, VectorTileLayer> = {};

    addLayer(layer: OverzoomedTileLayer) {
        this.layers[layer.name] = layer;
    }
}

/**
 * Encodes the virtual tile into binary vector tile form. 
 * This is a convenience that allows `FeatureIndex` to operate the same way across `VectorTileSource` and `GeoJSONSource` data.
 * @param virtualVectorTile - a VectorTile created from GeoJSON data using geojson-vt
 * @returns 
 */
export function toVirtualVectorTile(virtualVectorTile: VectorTile): LoadVectorTileResult {
    
    let pbf: Uint8Array = fromVectorTileJs(virtualVectorTile);
    if (pbf.byteOffset !== 0 || pbf.byteLength !== pbf.buffer.byteLength) {
        pbf = new Uint8Array(pbf);  // Compatibility with node Buffer (https://github.com/mapbox/pbf/issues/35)
    }
    
    return {
        vectorTile: virtualVectorTile,
        rawData: pbf.buffer
    };
}

export function sliceTileLayer(sourceLayer: VectorTileLayer, maxZoomTileID: CanonicalTileID, targetTileID: CanonicalTileID): OverzoomedTileLayer {
    const { extent } = sourceLayer;
    if (extent !== 4096) {
        console.log(`Warning: overzooming only tested with extent=4096, got extent=${extent}`);
    }
    const dz = targetTileID.z - maxZoomTileID.z;
    const scale = Math.pow(2, dz);
    const startXRelative = targetTileID.x / scale - maxZoomTileID.x;
    const startYRelative = targetTileID.y / scale - maxZoomTileID.y;

    const clipMinX = Math.round(startXRelative * extent);
    const clipMinY = Math.round(startYRelative * extent);
    const clipMaxX = Math.round((startXRelative + (1 / scale)) * extent);
    const clipMaxY = Math.round((startYRelative + (1 / scale)) * extent);

    // The clipping box coordinates within the source tile (0-extent)
    const clippingBox = {
        minX: clipMinX - 128,
        minY: clipMinY - 128,
        maxX: clipMaxX + 128,
        maxY: clipMaxY + 128
    };

    const featureWrappers: OverzoomedFeatureWrapper[] = [];
    for (let index = 0; index < sourceLayer.length; index++) {
        const feature: VectorTileFeature = sourceLayer.feature(index);
        let geometry = feature.loadGeometry();
        geometry = clip(geometry, feature.type, clippingBox.minX, clippingBox.maxX, AxisType.X);
        geometry = clip(geometry, feature.type, clippingBox.minY, clippingBox.maxY, AxisType.Y);
        // scale coordinates to the new tile extent
        for (const ring of geometry) {
            for (const point of ring) {
                point.x = (point.x - clipMinX) * scale;
                point.y = (point.y - clipMinY) * scale;
            }
        }
        if (geometry.length === 0) {
            continue;
        }
        featureWrappers.push(new OverzoomedFeatureWrapper(
            feature.type,
            geometry,
            feature.properties,
            feature.id,
            extent
        ));
    }
    return new OverzoomedTileLayer(featureWrappers, sourceLayer.name, extent);
}

const enum AxisType {
    X = 0,
    Y = 1
}

/* clip features between two vertical or horizontal axis-parallel lines:
 *     |        |
 *  ___|___     |     /
 * /   |   \____|____/
 *     |        |
 *
 * k1 and k2 are the line coordinates
 * axis: 0 for x, 1 for y
 * minAll and maxAll: minimum and maximum coordinate value for all features
 */
export default function clip(geometry: Point[][], type: number, start: number, end: number, axis: AxisType): Point[][] {
    //const min = axis === AxisType.X ? geometry.map(ring => ring.reduce((min, p) => Math.min(min, p.x), Infinity)).reduce((a, b) => Math.min(a, b), Infinity) : geometry.map(ring => ring.reduce((min, p) => Math.min(min, p.y), Infinity)).reduce((a, b) => Math.min(a, b), Infinity);
    //const max = axis === AxisType.Y ? geometry.map(ring => ring.reduce((max, p) => Math.max(max, p.y), -Infinity)).reduce((a, b) => Math.max(a, b), -Infinity) : geometry.map(ring => ring.reduce((max, p) => Math.max(max, p.x), -Infinity)).reduce((a, b) => Math.max(a, b), -Infinity);

    //if (max < start || min >= end) { // trivial reject
    //    return [];
    //}
    switch (type) {
        case 1: // POINT
            return clipPoints(geometry, start, end, axis);
        case 2: // LINESTRING
            return clipLines(geometry, start, end, axis, false);
        case 3: // POLYGON
            return clipLines(geometry, start, end, axis, true);
    }
    
    return null;
}

function clipPoints(geometry: Point[][], start: number, end: number, axis: AxisType): Point[][] {
    const newGeometry: Point[][] = [];
    for (const ring of geometry) {
        for (const point of ring) {
            const a = axis === AxisType.X ? point.x : point.y;
            if (a >= start && a <= end) {
                newGeometry.push([point]);
            }
        }
    }
    return newGeometry;
}

function clipLine(line: Point[], start: number, end: number, axis: AxisType, isPolygon: boolean): Point[][] {
    const intersect = axis === AxisType.X ? intersectX : intersectY;

    let slice: Point[] = [];
    const newLine: Point[][] = [];
    for (let i = 0; i < line.length - 1; i++) {
        const p1 = line[i];
        const p2 = line[i + 1];
        const a = axis === AxisType.X ? p1.x : p1.y;
        const b = axis === AxisType.X ? p2.x : p2.y;
        let exited = false;

        if (a < start) {
            // ---|-->  | (line enters the clip region from the left)
            if (b > start) {
                slice.push(intersect(p1, p2, start));
            }
        } else if (a > end) {
            // |  <--|--- (line enters the clip region from the right)
            if (b < end) {
                slice.push(intersect(p1, p2, end));
            }
        } else {
            slice.push(p1);
        }
        if (b < start && a >= start) {
            // <--|---  | or <--|-----|--- (line exits the clip region on the left)
            slice.push(intersect(p1, p2, start));
            exited = true;
        }
        if (b > end && a <= end) {
            // |  ---|--> or ---|-----|--> (line exits the clip region on the right)
            slice.push(intersect(p1, p2, end));
            exited = true;
        }

        if (!isPolygon && exited) {
            newLine.push(slice);
            slice = [];
        }
    }

    // add the last point
    let last = line.length - 1;
    const a = axis === AxisType.X ? line[last].x : line[last].y;
    if (a >= start && a <= end) {
        slice.push(line[last]);
    }

    // close the polygon if its endpoints are not the same after clipping
    if (isPolygon && slice.length > 0 && !slice[0].equals(slice[slice.length - 1])) {
        slice.push(new Point(slice[0].x, slice[0].y));
    }
    if (slice.length > 0) {
        newLine.push(slice);
    }
    return newLine;
}


function clipLines(geometry: Point[][], start: number, end: number, axis: AxisType, isPolygon: boolean): Point[][] {
    const newGeometry: Point[][] = [];
    for (const line of geometry) {
        const clippedLines = clipLine(line, start, end, axis, isPolygon);
        if (clippedLines.length > 0) {
            newGeometry.push(...clippedLines);
        }
    }
    return newGeometry;
}

function intersectX(p1: Point, p2: Point, x: number): Point {
    const t = (x - p1.x) / (p2.x - p1.x);
    return new Point(x, p1.y + (p2.y - p1.y) * t);
}

function intersectY(p1: Point, p2: Point, y: number): Point {
    const t = (y - p1.y) / (p2.y - p1.y);
    return new Point(p1.x + (p2.x - p1.x) * t, y);
}
```

wayofthefuture:
[vector_tile_worker_source.ts](https://github.com/user-attachments/files/23026919/vector_tile_worker_source.ts)

[overzoomed-geojson-vector-tile.ts](https://github.com/user-attachments/files/23026926/overzoomed-geojson-vector-tile.ts)

HarelM:
Nice! Thanks for this! So this was about when to do the coordinates conversion basically?
Are you sure about the need to cache the calculated tiles? I'm considering removing the cache all together...?

wayofthefuture:
Yes... I didn't get a chance to calculate performance but processing could be heavy and it is a nice o(1) lru cache that we can also use in source cache eventually. It's simple and I don't see how it wouldn't improve performance.

HarelM:
I think the problem here is that iOS might kill things that use too much memory, this is what I think is causing the crash with the geojson index, although it might be a magnitude bigger...

wayofthefuture:
Many magnitudes larger especially with tolerance: 0. I was able to monitor memory with the worker, and the previous version was growing well over 10GB in memory and quickly. The current version with 10,000 (notice I added a 0!) in cache didn't exceed 2GB with heavy automated movement. With 0 cache the page was about 1GB in total. You had 1 pyramid per layer per max zoom tile, that's like n^n^n or something! The current cache is simply flat vector tiles so I think it will work nicely.

HarelM:
I found this when searching for something related to clipping in the code base as I assumed there should already be something:
https://github.com/maplibre/maplibre-gl-js/blob/main/src/symbol/clip_line.ts

It's not exactly the same, but looks very similar, I wonder if I can use it or adopt it to our needs...

HarelM:
I moved the clipping code to the same file as the other method.
They are very similar, but they differ a bit with the rounding of the intersection points, so I'll keep them side by side at the moment.
I also changed the name of the experimental variable to match the current implementation.

src/source/vector_tile_overzoomed.ts
wayofthefuture, line 1: maybe rename to vector_tile_overzoom.ts

src/source/vector_tile_worker_source.ts
```diff
@@ -64,7 +59,7 @@ export class VectorTileWorkerSource implements WorkerSource {
         this.fetching = {};
         this.loading = {};
         this.loaded = {};
-        this.overzoomedTilesCache = new BoundedLRUCache<string, GeoJSONVT>(1000);
+        this.overzoomedTilesCache = new BoundedLRUCache<string, LoadVectorTileResult>(1000);
```
wayofthefuture: extra extra safe 😉 

image

"Page" in Safari Timeline Memory pane recording is where you find the memory build-up crash (you'll notice 9-10GB before this commit)

HarelM: I've searched a lot and couldn't find this gauge. They really tried to hide it.
Lucking you posted a screen shot, so I know not to give up...

src/source/vector_tile_overzoomed.ts
HarelM: Should I change the names in that file as well?
The file name does seem better as it will be placed next to the other file...

wayofthefuture: Nope just the filename

src/source/vector_tile_worker_source.ts
wayofthefuture: image

Timeline, click the red button, click the green "memory" box

andrewharvey:
Thanks for your work on this @HarelM.

> It is only used by default for Safari because it changes a bit the results of query render features geometry, since the tiles are no longer capped at max zoom, so the geometry is sliced as well in high zoom tiles, thus creating a "spliced" geometries, while this might be expected up to max-zoom, this is might not be expected in higher zoom levels.
The performance of this seems to work well and it solves the crash on iOS as far as my tests went.

I'm not a fan of having the library function differently on different browsers like this. We should set a clear expectation to users about what `queryRenderedFeatures` returns. We already say in the API docs 

> Because features come from tiled vector data or GeoJSON data that is converted to tiles internally, **feature geometries may be split or duplicated across tile boundaries** and, as a result, features may appear multiple times in query results. For example, suppose there is a highway running through the bounding rectangle of a query. The results of the query will be those parts of the highway that lie within the map tiles covering the bounding rectangle, even if the highway extends into other tiles, and the portion of the highway within each map tile will be returned as a separate feature. Similarly, a point feature near a tile boundary may appear in multiple tiles due to tile buffering.

I think we should just accept that geometries returned from queryRenderedFeatures at an overzoom will have further clipping applied beyond the sources maxZoom. That should still represent the geometries visible on the map and therefore works the same as any other zoom level under the maxZoom of the source.

HarelM:
> I'm not a fan of having the library function differently on different browsers like this

Neither am I but currently Safari crashes while Chrome doesn't, so the difference is already there to some extent. 

The main difference I saw is for areas that has a label placed in the middle, this label is now repeated in every sliced tile, it becomes apparent in high zoom levels (20+) since this is most of what you see in the screen, since there's not a lot of data in one screen.

I suggest to try the branch out with a style your are familiar with and see if you can spot the changes.

HarelM:
Ok, So I wanted to polish this a bit more and I think I came up with an idea to set the number of zoom level to overscale instead of doing it for all zoom levels.
This would mean that you can specify how many zoom levels to overscale and the rest would be using the split technique.
I'm currently testing a default of 4 - meaning, the last 4 zoom levels will be overscaled and the rest will be sliced.
This gives more control on when to overscale and when to slice, which I think is a good middle ground.

wayofthefuture:
So if you have map tile boundaries enabled would you see it switch from 1 tile to 256 subdivided tiles at maxzoom + 4 -> maxzoom + 5? Not sure how such a switch would look and it could have drastic style changes maybe?

HarelM:
I'll push it to a different branch so you could play with it if you want.
Let's use number to illustrate it better:

Source max zoom is 10.
Map max zoom is 20.
We are over-scaling for only 3 zoom levels.

In that case the slicing will happen for zoom 11 to 17, and after that, a level 17 "sliced tile" will be used as the base for over scaling, similar to what happens now with the tiles at max zoom, only in this implementation we first slice the tile and later on, if needed scale it.

It gives more control on when to stop slicing and start scaling.
A private case would be the behavior before this PR which is to always scale.
It also removes the "bypass covering tiles" part by using the maxzoom parameter.

HarelM:
See the following PR:
- https://github.com/maplibre/maplibre-gl-js/pull/6617

wayofthefuture:
Got it I was thinking of it backwards...

wayofthefuture:
Sorry I accidentally deleted I have to be careful when cleaning up in Webstorm!

HarelM:
With great power...

HarelM:
@wayofthefuture this is now ready to be reviewed and hopefully merged.
I ended up not changing the default behavior of this library and only allowed to use this experimental features.
I've seen some reports on iOS 18.7.2 breaking webgl completely, so I hope the white screen of death will be solved by Apple instead of here.
Never the less, this adds some possible performance optimization in high zoom levels.
I want to get some feedback on this before I break stuff, so the default here might change in a breaking changes version (6) whenever we decide to release it.

wayofthefuture:
@HarelM As I was reviewing I realized that the problem is not only iOS. 'Too many glyphs being rendered in a tile' also occurs on Chrome. Even though Chrome doesn't crash, it's nevertheless hitting the limit. I've been messing with it and may have found an alternate solution. After I have some more time to test it I'll let you know if it's viable and you can decide the path. Maybe it would be nice to solve this without having to add this extra logic. I don't know.

src/source/vector_tile_worker_source.ts
```diff
@@ -144,6 +156,43 @@ export class VectorTileWorkerSource implements WorkerSource {
...
+     * @param params - the worker tile parameters
+     * @param maxZoomVectorTile - the original vector tile at the source's max available canonical zoom
+     * @returns the overzoomed tile and its raw data
+     */
+    private _getOverzoomTile(params: WorkerTileParameters, maxZoomVectorTile: VectorTile): LoadVectorTileResult {
+        const {tileID, source, overzoomParameters} = params;
+        const {maxZoomTileID} = overzoomParameters;
+
+        const cacheKey = `${maxZoomTileID.key}_${tileID.key}`;
+        const cachedOverzoomTile = this.overzoomedTilesCache.get(cacheKey);
```
wayofthefuture: cachedOverzoomTileResult?

src/source/vector_tile_worker_source.ts
```diff
@@ -58,6 +59,7 @@ export class VectorTileWorkerSource implements WorkerSource {
         this.fetching = {};
         this.loading = {};
         this.loaded = {};
+        this.overzoomedTilesCache = new BoundedLRUCache<string, LoadVectorTileResult>(1000);
```
wayofthefuture: overzoomedTileResultCache?

src/tile/tile_manager.ts
```diff
@@ -568,11 +568,13 @@ export class TileManager extends Evented {
             idealTileIDs = coveringTiles(transform, {
                 tileSize: this.usedForTerrain ? this.tileSize : this._source.tileSize,
                 minzoom: this._source.minzoom,
-                maxzoom: this._source.maxzoom,
+                maxzoom: this._source.type === 'vector' && this.map._zoomLevelsToOverscale !== undefined
+                    ? transform.maxZoom - this.map._zoomLevelsToOverscale
```
wayofthefuture: this seems odd to me. why is it not this._source.maxzoom + this.map._zoomLevelsToOverscale.

It is also strange that coveringTiles returns overscaled tiles except when zoomLevelsToOverscale is defined.

src/ui/map.ts
```diff
@@ -370,6 +370,18 @@ export type MapOptions = {
...
+     * Defines the number of zoom level that will overscale from map's max zoom and below.
+     * For example if the map's max zoom is 20 and this is set to 3, the zoom levels of 20, 19 and 18 will be overscaled 
+     * and the rest will be split.
+     * When undefined, all zoom levels after source's max zoom will be overscaled.
+     * This can help in reducing the size of the overscaling and improve performance in high zoom levels.
+     * The drawback is that it changes rendering for polygon centered labels and changes the results of query rendered features.
+     * @defaultValue undefined
+     * @experimental
+     */
+    experimentalZoomLevelsToOverscale?: number;
```
wayofthefuture, line 384: Without this setting, tiles are still overscaled. So the name is confusing. Maybe invert this and rename to experimentalMaxSplittingZoomLevels

- enables splitting of vector tiles after the source's max zoom
- after max splitting zooms, reverts to standard vector tile overscaling

src/tile/tile_manager.ts
wayofthefuture, line 572: so coveringTiles returns overscaled tiles unless zoomLevelsToOverscale is defined. odd right?

maybe something like this.source._maxzoom + _maxSplittingZoomLevels

wayofthefuture (commented):
First and foremost "dependabug" littering the file changes makes it impossible to give a solid review.

Without the experimental option `experimentalZoomLevelsToOverscale` enabled, vector tiles are still overscaled. So it is very confusing with this name.

I think invert the option and call it `maxSplittingZoomLevels` or something like that, because after that point the default overscaling behavior kicks in right?

HarelM:
I get how the current name can be confusion, but the idea is to limit the number of overscaled zoom levels, so if there are 2 sources with different max zoom they will still have the same number of overscaled zoom levels (counting from the end).
If you have a better idea how to solve this let me know...

src/ui/map.ts
HarelM: Are the docs above this parameter sufficient?

wayofthefuture:
Can the option be moved to source options? Then you can invert it and different sources can have different behaviors.

HarelM:
This would need a spec change which I would like to avoid for this feature, especially since it's experimental and I'm not sure it's a technical detail style authors should be dealing with.

wayofthefuture:
Ok... but keep in mind the feature being enabled is tile splitting which is kind of obscured

src/ui/map.ts
wayofthefuture: I guess although highly confusing.

HarelM: I agree, but I don't have a good idea how to make it more clear to be honest.
In any case, this is experimental for now. I hope it can be useful as is.

HarelM:
I'll merge this and see if I get feedback on it (which I guess I won't) and then see what can be improved.
The docs of the flag does highlight the splitting part as best as it can, so let's hope it would be useful.

spatialillusions:
Sorry for jumping in here, but would it be possible to expose the ability to expose this functionality to split tiles on the fly so it can be used in plugins? That would make it possible to solve the problem with formats that have sparse tile matrices such as in #5618.

HarelM:
You can checkout the code I wrote, it doesn't have a lot of lines, you can copy this to any plugin you might need.
We are looking into improving geojson-vt, and for that there might be a way to expose this functionality as part of that package (mainly the clipping method, which is copied from there for now and modified a bit to be more readable).
I don't think it should be exposed from this package though.

etienneJr:
Hello @HarelM, has this PR resolved (or should it have resolved) the issue raised in the very first comment on #2507, namely the fact that street names are all mixed up?
image
Thanks

HarelM:
I think there's a different bug according to the following issue: 
- #6720

Nevertheless, this issue does resolve the too many glyph warning and prevents iOS crashes in dense labeled areas.

PR #6529, opened 2025-10-09
Contextmenu events triggered during scrolling

### LINKED ISSUE #5683: contextmenu events not firing before scrolling "finishes"
**maplibre-gl-js version**: 5.3.0

**browser**: Firefox 136

### Steps to Trigger Behavior

 1. Add an event handler for `contextmenu`: `map.on('contextmenu', () => { console.log('contextmenu fired'); };`
 2. Scroll the map using the mouse wheel
 3. Directly after scrolling, right click on the map

### Expected Behavior

The `contextmenu` event should fire and `contextmenu fired` should be logged to the console.

### Actual Behavior

Nothing. The expected behaviour only works after some time (half a second? seems to be after the render finishes?).

### LINKED ISSUE #6302: "mousemove" events not firing while scroll-zooming
**maplibre-gl-js version**: 5.6.2

**browser**: Chrome 139 on MacOS

### Steps to Trigger Behavior

 1. Add a `mousemove` listener to the map, e.g., `map.on('mousemove', (e) => console.log(e))`
 2. Move pointer on the map 👉🏻 Notice logs
 3. Scroll and move pointer 👉🏻 Map zooms in/out and logs stop until scrolling has completed

### Link to Demonstration

Notice how the LngLat text isn't updating while you scroll with the scroll wheel.

https://jsbin.com/worolupeki/edit?html,output

https://github.com/user-attachments/assets/f72830af-ab84-47f8-b848-94ea2e7d09e1

### Expected Behavior

`mousemove` is fired when the pointer moves, or at least as soon as it's "safe" to do so (i.e., right when the scroll action ends)

### Actual Behavior

`mousemove` isn't fired at all while scrolling, not even at the end of the scroll action/animation. This results in a mismatch between the pointer's actual and last recorded positions

This also impacts popup that are set up to track the pointer: they stay behind until the mouse is moved again __outside__ of a scroll gesture

### Notes

Related to #5683 

Mapbox updated this logic back in 2020 https://github.com/mapbox/mapbox-gl-js/issues/10155. I have also verified that the fix they applied (see https://github.com/mapbox/mapbox-gl-js/commit/20b953937ac54e3743aed06066b0bbe0092f5c9a) can work in maplibregl too. As such, I'd be happy to open a PR that mirrors those changes

mmc1718:
- Fixes #5683 
- Fixes #6302

This PR updates the contextmenu event handler to allow contextmenu events to be not blocked by scrolling.

 - [x] Link to related issues.

### DIFF
--- a/test/integration/browser/browser.test.ts
+++ b/test/integration/browser/browser.test.ts
@@ -79,10 +79,10 @@ describe('Browser tests', () => {
     test('Mousemove events are fired during scrollzoom', {retry: 3, timeout: 20000}, async () => {
         const mouseMoveFired = await page.evaluate(() => {
             return new Promise<Array<number>>((resolve, _reject) => {
-                let mouseMoveCount = 0
-                let wheelCount = 0
+                let mouseMoveCount = 0;
+                let wheelCount = 0;
                 map.on('mousemove', () => {mouseMoveCount++;});
-                map.on('wheel', () => {wheelCount++})
+                map.on('wheel', () => {wheelCount++;});
                 map.getCanvas().dispatchEvent(new WheelEvent('wheel', {deltaY: 120, bubbles: true}));
                 map.getCanvas().dispatchEvent(new WheelEvent('wheel', {deltaY: 120, bubbles: true}));
                 map.getCanvas().dispatchEvent(new MouseEvent('mousemove', {bubbles: true}));
@@ -91,9 +91,9 @@ describe('Browser tests', () => {
                 map.getCanvas().dispatchEvent(new MouseEvent('mousemove', {bubbles: true}));
                 map.getCanvas().dispatchEvent(new WheelEvent('wheel', {deltaY: 120, bubbles: true}));
                 map.getCanvas().dispatchEvent(new MouseEvent('mousemove', {bubbles: true}));
-                resolve([mouseMoveCount, wheelCount])
-            })   
+                resolve([mouseMoveCount, wheelCount]);
             });
+        });
         expect(mouseMoveFired[0]).toBe(4);
         expect(mouseMoveFired[1]).toBe(4);
     });


src/ui/handler_manager.ts
```diff
@@ -307,7 +307,7 @@ export class HandlerManager {
             map.keyboard.enable();
         }
 
-        this._add('blockableMapEvent', new BlockableMapEventHandler(map));
+        this._add('blockableMapEvent', new BlockableMapEventHandler(map), ['scrollZoom']);
```
HarelM: Can we simply move this up a bit instead? I think it will solve the problem better...?

HarelM:
If we can't simulate this in a unit test let's remove it and keep only the browser test...

src/ui/handler_manager.ts
mmc1718: Done :)

HarelM:
Does this change also fixes #6302?

mmc1718:
@HarelM this does seem to fix #6302 . I also wrote a test to check https://github.com/maplibre/maplibre-gl-js/commit/feb0cc1e2ac0bac3d1dff4d662479373693eace8

Do you think the test is worth adding? If so I could add it to this PR or I could open a new PR (that way the test should fail before this PR is merged)

HarelM:
If you wrote a test to make sure a bug doesn't return, then definitely add it! :-)

mmc1718:
Thanks @HarelM ! I've added the test in a [separate PR](https://github.com/maplibre/maplibre-gl-js/pull/6546) to demonstrate that the test fails. Once this one is merged, I will rebase the other branch on main and the test should pass.

HarelM:
Since this PR fixes both issues I think it's better to have both tests here.
I see that the test is failing on main, which is good, now you can have the test here and we can merge it.
What do you think?

mmc1718:
Sounds good, I'll cherry pick the commits from the other branch

PR #6532, opened 2025-10-09
Vector source overzoom improvements on PR-6521

wayofthefuture:
My first PR against a PR so yeah

### DIFF
--- a/src/source/vector_tile_worker_source.ts
+++ b/src/source/vector_tile_worker_source.ts
@@ -195,7 +195,7 @@ export class VectorTileWorkerSource implements WorkerSource {
 
             // if we are seeking a tile deeper than the sources max available canonical tile, get the overzoomed tile
             if (overzoomParameters) {
-                const overzoomTile = this.getOverzoomTile(params, response.vectorTile);
+                const overzoomTile = this._getOverzoomTile(params, response.vectorTile);
                 response.rawData = overzoomTile.rawData;
                 response.vectorTile = overzoomTile.vectorTile;
             }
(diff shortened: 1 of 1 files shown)


HarelM:
This looks great! Thanks for the source cache bypass, I'll continue looking to see which issues I'm finding as currently there's an issue with areas and tiles boundaries that needs to be adjusted, but this PR looks good, I just need to make sure it is working before merging it.
THANKS!!

src/source/vector_tile_worker_source.ts
```diff
@@ -265,6 +276,25 @@ export class VectorTileWorkerSource implements WorkerSource {
...
+            geoJSONFeatures.push(feature.toGeoJSON(maxZoomTileID.x, maxZoomTileID.y, maxZoomTileID.z));
+        }
+
+        return geojsonvt({
+            type: 'FeatureCollection',
+            features: geoJSONFeatures
+        }, {
+            extent: EXTENT,
+            maxZoom: maxOverzoom,
+            buffer: Math.round(256 / 512 * tileSize),  // fix tile seams (especially when zooming out)
```
HarelM: Can we use the tile's buffer by any chance?

wayofthefuture: Not sure... I think 50% buffer is a temporary hack... it seems to be a style some sort the bigger issue as it matches legitimate style:
image

wayofthefuture: The gradient line looks just like border line:
image

wayofthefuture:
Might want to add indexMaxZoom: 0 to geojson_source huge gain in performance 1ms vs 10ms

src/source/vector_tile_worker_source.ts
wayofthefuture: I copied the geojson_source buffer implementation and it works! not a hack anymore!

HarelM:
Thanks for all the fixes! I'll check what can be done to either fix the query render results or update the expected values depending on my findings.

PR #6540, opened 2025-10-10
fix: console in warnOnce might be undefined after minifying

romanpickl:
## Launch Checklist

We ran into a situation were terser minified this function, sometimes renaming the reference to console (e.g. to 'n'). This then broke in production (e.g. could not parse tiles with geometries exceeding the allowed content rather then only showing the log message: "Geometry exceeds allowed extent, reduce your vector tile buffer size") as 'n' is undefined.
maplibre_minified

Maybe there's a better solution. So feel free to comment.

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
This sounds like a problem in a different package, I advise to get this big fixed there or use a different minifier...
I would like to avoid fixing it here if possible...

Issue #6543, opened 2025-10-11
Feature: Time control API for deterministic rendering


bjperson:
## User Story

As a **MapLibre GL developer**, I can **control the internal clock deterministically**, so that
I can **export frame-perfect videos and create reproducible tests for time-based features**.

### Why is this feature needed?

MapLibre GL JS currently lacks the ability to control its internal time source, making it
impossible to achieve deterministic rendering. This creates significant limitations for
developers who need:

1. **Frame-by-frame video export**: Creating smooth, high-quality videos from map animations
2. **Deterministic testing**: Writing reliable tests for animations, transitions, and time-based
 behaviors
3. **Screenshot consistency**: Capturing identical frames across multiple runs
4. **Animation debugging**: Stepping through animations frame-by-frame

### Example use cases

**Video Export:**
```javascript
// Record a 30-second animation at exactly 60fps
let virtualTime = 0;
maplibregl.setNow(virtualTime);

for (let frame = 0; frame < 1800; frame++) {
    virtualTime += 1000 / 60;  // Advance 16.67ms per frame
    maplibregl.setNow(virtualTime);
    captureFrame();  // Always captures consistent state
}

maplibregl.restoreNow();

Testing:
test('flyTo animation completes in 2 seconds', () => {
    maplibregl.setNow(0);
    map.flyTo({center: [0, 0], duration: 2000});

    maplibregl.setNow(1000);  // Halfway
    expect(map.getZoom()).toBeCloseTo(expectedZoom);

    maplibregl.setNow(2000);  // Complete
    expect(map.getCenter()).toEqual([0, 0]);
});
```

Documentation/Demos:
Creating consistent animated GIFs or videos for documentation without timing variations.

Can it be done today?

No workaround exists. The internal time source (browser.now()) cannot be controlled from outside
 MapLibre. Monkey-patching performance.now affects the entire browser and breaks other
libraries.

Attempted workarounds:
- ❌ Mock performance.now globally → breaks other code
- ❌ Adjust animation durations → imprecise, doesn't work for complex sequences
- ❌ Use real-time recording → inconsistent frame timing, dropped frames

Impact

If we do nothing:

Developers lose significant capabilities:

1. Video export market: Growing demand for map video content (marketing, presentations, social
media) remains unserved by MapLibre
2. Testing quality: Time-based features remain difficult to test reliably, leading to flaky
tests
[...]

PR #6544, opened 2025-10-11
Feat/time control API for deterministic rendering

bjperson:
- Add setNow(), restoreNow() and isTimeFrozen() methods for controlling internal clock
- Enable frame-by-frame video export with deterministic timing
- Support deterministic testing of time-based features
- Clean-room implementation to avoid copyright concerns
- Comprehensive test coverage with 17 test cases

This feature has been requested since the MapLibre/Mapbox fork in 2020.
Mapbox added similar functionality right after the fork in [commit 0063cbd](https://github.com/mapbox/mapbox-gl-js/commit/0063cbd10a97218fb6a0f64c99bf18609b918f4c). This functionality never made it to MapLibre, leaving a gap for developers who need deterministic rendering. This implementation uses a different architecture
(class-based TimeManager) to ensure it's a clean-room implementation.

Related issue : [6543](https://github.com/maplibre/maplibre-gl-js/issues/6543)

## Launch Checklist

 - [x] Link to related issues.
 - [x]  Demo of a [work in progress plugin](https://www.ideeslibres.org/maplibre-gl-video-export-demo/) using this patch.

### DIFF
--- a/src/util/time_control.test.ts
+++ b/src/util/time_control.test.ts
@@ -24,17 +24,20 @@ describe('time_control', () => {
     });
 
     describe('now()', () => {
-        test('returns current time when not frozen', () => {
-            const time1 = now();
-            expect(typeof time1).toBe('number');
-            expect(time1).toBeGreaterThanOrEqual(0);
+        test('returns a valid number when not frozen', () => {
+            const currentTime = now();
+            expect(typeof currentTime).toBe('number');
+            expect(currentTime).toBeGreaterThanOrEqual(0);
+        });
 
-            // Small delay to ensure time advances
-            waitForRealTime(1);
-            const time2 = now();
+        test('advances over time when not frozen', () => {
+            const timeBeforeDelay = now();
 
-            // Time should advance (or at least not go backwards)
-            expect(time2).toBeGreaterThanOrEqual(time1);
+            const minimalDelayMs = 1;
+            waitForRealTime(minimalDelayMs);
+
+            const timeAfterDelay = now();
+            expect(timeAfterDelay).toBeGreaterThanOrEqual(timeBeforeDelay);
         });
 
         test('returns frozen time when time is set', () => {
@@ -44,10 +47,9 @@ describe('time_control', () => {
             const time1 = now();
             expect(time1).toBe(frozenTime);
 
-            // Wait a bit to ensure real time advances
-            waitForRealTime(10);
+            const realTimeDelayMs = 10;
+            waitForRealTime(realTimeDelayMs);
 
-            // Time should still be frozen
             const time2 = now();
             expect(time2).toBe(frozenTime);
         });
@@ -85,13 +87,13 @@ describe('time_control', () => {
             expect(isTimeFrozen()).toBe(true);
         });
 
-        test('handles edge case values', () => {
-            // Can freeze at 0
+        test('freezes time at zero', () => {
             setNow(0);
             expect(now()).toBe(0);
             expect(isTimeFrozen()).toBe(true);
+        });
 
-            // Can freeze at negative values
+        test('freezes time at negative values', () => {
             const negativeTime = -123.456;
             setNow(negativeTime);
             expect(now()).toBe(negativeTime);
@@ -138,16 +140,16 @@ describe('time_control', () => {
 
     describe('isTimeFrozen()', () => {
         test('correctly tracks frozen state', () => {
-            // Initially not frozen
-            expect(isTimeFrozen()).toBe(false);
+            const initialFrozenState = isTimeFrozen();
+            expect(initialFrozenState).toBe(false);
 
-            // Becomes true when time is set
             setNow(12345);
-            expect(isTimeFrozen()).toBe(true);
+            const frozenStateAfterSet = isTimeFrozen();
+            expect(frozenStateAfterSet).toBe(true);
 
-            // Becomes false after restore
             restoreNow();
-            expect(isTimeFrozen()).toBe(false);
+            const frozenStateAfterRestore = isTimeFrozen();
+            expect(frozenStateAfterRestore).toBe(false);
         });
     });
 


bjperson:
Added a link to a demo of a plugin currently in development that uses this patch.

src/util/browser.ts
```diff
@@ -17,6 +14,18 @@ export const browser = {
      */
     now,
 
+    /**
```
HarelM: Does it makes sense to keep these methods here? What's the point of wrapping them in another object?

src/index.ts
```diff
@@ -370,5 +371,8 @@ export {
     removeProtocol,
     addSourceType,
     importScriptInWorkers,
-    createTileMesh
+    createTileMesh,
+    setNow,
```
HarelM, line 375: Should these be "global" or part of the maps's API?

src/util/time_control.ts
```diff
@@ -0,0 +1,96 @@
...
+ * // ... render frame ...
+ * setNow(16.67);       // Second frame
+ * // ... render frame ...
+ * setNow(33.34);       // Third frame
+ * // ... done ...
+ * restoreNow();        // Resume normal time
+ * ```
+ */
+export function setNow(timestamp: number): void {
+    timeManager.freezeAt(timestamp);
```
HarelM: Why not use the same terminology?

src/util/time_control.ts
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/**
+ * Restores normal time flow after freezing with setNow().
+ * Call this after finishing deterministic rendering operations.
+ */
```
HarelM, line 87: Add examples here as well, even if it's obvious ones.

src/util/time_control.ts
```diff
@@ -0,0 +1,96 @@
+/**
+ * Time control utilities for MapLibre GL JS.
```
HarelM: Not sure there's a need to say that it's for maplibre gl js, it goes without saying, I think...

src/util/time_control.ts
```diff
@@ -0,0 +1,96 @@
...
+ * Provides deterministic time control for use cases like video export
+ * where consistent frame-by-frame rendering is required.
+ */
```
HarelM: Why is this comment split?

src/util/time_control.test.ts
```diff
@@ -0,0 +1,277 @@
...
+        Date.now.bind(Date);
+
+    beforeEach(() => {
+        // Ensure we start each test with unfrozen time
+        restoreNow();
+    });
+
+    afterEach(() => {
+        // Clean up after each test
+        restoreNow();
```
HarelM, line 23: Not sure I see a reason to clear both before and after...?

src/util/time_control.test.ts
```diff
@@ -0,0 +1,277 @@
...
+            expect(time1).toBeGreaterThanOrEqual(0);
+
+            // Small delay to ensure time advances
+            const startReal = realPerformanceNow();
+            while (realPerformanceNow() - startReal < 1) {
+                // busy wait
+            }
+            const time2 = now();
+
+            // Time should advance (or at least not go backwards)
```
HarelM: This comment usually indicate another test is in place here...

src/util/time_control.test.ts
HarelM, line 1: Nit picking: I think the tests are great, but it feels like the code that was written can be covered by less tests.
I would also avoid logic in the tests add much as possible (for loops etc) and maybe move the busy-wait to a method in the test class..

HarelM:
I've added some nit picking comments, thanks for contributing this code!!

src/index.ts
bjperson: I believe these should remain global because they control time for the entire library, not per-map instance. When freezing time for deterministic rendering (e.g., video export), we want all Map instances to share the same frozen time state to ensure consistency. This is similar to how performance.now() works as a global time source.

HarelM: Ok, fair enough.

src/util/browser.ts
bjperson: Good point! I've removed setNow and restoreNow from the browser object since they're not used via browser.* anywhere in the codebase. They're only exported through the main API in index.ts. (commit: 55845e8b6)

src/util/time_control.ts
bjperson: Good point! I've aligned the internal method names with the public API for consistency (commit: 5bff70ea4).

src/util/time_control.test.ts
bjperson: Thanks for the thorough review! I've addressed all the points including simplifying the tests. Let me know if anything else needs adjustment.

src/util/browser.ts
```diff
@@ -1,10 +1,7 @@
 import {createAbortError} from './abort_error';
+import {now} from './time_control';
```
HarelM, line 2: Can `now` be removed from here as well and simply used straight from the other file?

src/util/time_control.test.ts
```diff
@@ -0,0 +1,202 @@
...
+            const frozenTime = 123456.789;
+            setNow(frozenTime);
+
+            const time1 = now();
+            expect(time1).toBe(frozenTime);
+
+            // Wait a bit to ensure real time advances
+            waitForRealTime(10);
+
+            // Time should still be frozen
```
HarelM: Use better variable names instead of comments...

src/util/time_control.test.ts
```diff
@@ -0,0 +1,202 @@
...
+            expect(isTimeFrozen()).toBe(true);
+        });
+
+        test('handles edge case values', () => {
+            // Can freeze at 0
+            setNow(0);
+            expect(now()).toBe(0);
+            expect(isTimeFrozen()).toBe(true);
+
+            // Can freeze at negative values
```
HarelM: Should be a different test instead of two asserts in one test...

src/util/time_control.test.ts
```diff
@@ -0,0 +1,202 @@
...
+    describe('isTimeFrozen()', () => {
+        test('correctly tracks frozen state', () => {
+            // Initially not frozen
+            expect(isTimeFrozen()).toBe(false);
+
+            // Becomes true when time is set
+            setNow(12345);
+            expect(isTimeFrozen()).toBe(true);
+
+            // Becomes false after restore
```
HarelM: The assertion code should indicate that instead of a comment...

HarelM:
Thanks! Added some last comments, looks good otherwise!

src/util/browser.ts
bjperson: Good question! There's an important difference:

- setNow/restoreNow are new functions added in this PR, exported directly as global API functions via index.ts (never added to the browser object).
- now already exists and has 33 usages via browser.now() across the codebase (animations, camera, rendering, style, etc.).

Two options:

1. Do it in this PR: Replace all 33 browser.now() calls with direct now() imports and remove it from the browser object. This touches many files but keeps everything consistent.
2. Separate PR: Keep this PR focused on adding the time control API, then do a follow-up refactoring PR to migrate browser.now() → now() everywhere.

I'd lean towards option 2 to avoid mixing the new feature with a larger refactoring, but I'm happy to do either! What do you prefer?

HarelM: I'm good with a separate PR.

bjperson:
Thanks for the detailed review! I've addressed all the feedback. Ready for another look when you have time! 🙂

bjperson:
Thanks for merging! 🎉

HarelM:
Thank YOU!

wayofthefuture:
@bjperson It would be nice to have more of an explanation at the top of the file in the code, or an example of some sort. This looks interesting to me however I have no idea what I can also be used for. Do you know if it can be used to reduce or throttle map rendering?

bjperson:
Hi @wayofthefuture!


I actually added documentation with examples in the `now()` function docstring in `time_control.ts` as part of this PR (#6644).

You can see it here: https://github.com/maplibre/maplibre-gl-js/pull/6644/files#diff-d98bf7d878c6fff9bef92e0988a1b894e1d66a82fb764ec68873df193511b2b6

Does that address your concern, or would you like to see more examples/use cases in the code comments?

The timeControl API is primarily designed for deterministic rendering - situations where you need MapLibre to render frames at controlled, predictable times rather than real-time.

Main Use Cases:

1. Frame-by-frame video export (my use case):

```ts
// Export 60fps video
const fps = 60;
const frameDuration = 1000 / fps; // 16.67ms

for (let frame = 0; frame < totalFrames; frame++) {
    maplibregl.setNow(frame * frameDuration);  // Freeze time at exact frame time
    map.triggerRepaint();                      // Force render
    await captureFrame();                      // Grab the frame
}
maplibregl.restoreNow(); // Resume normal time
```

2. Automated testing - Jump to exact points in animations without waiting:

```ts
// Test an animation at its midpoint without waiting 5 seconds

maplibregl.setNow(0);
map.flyTo({center: [lng, lat], duration: 10000});  // 10 second animation

maplibregl.setNow(5000);  // Jump to 5 seconds
map.triggerRepaint();     // Force MapLibre to recalculate at t=5000
expect(map.getCenter()).toBeCloseTo(expectedMidpoint);
```

Note: This should work for MapLibre's native animations (flyTo, easeTo, style transitions) since they use now() internally. Custom JavaScript animations would need to use maplibregl.now() instead of Date.now() to benefit from this.

Can it throttle/reduce rendering?

Not directly - setNow() doesn't change the frequency of rendering, it controls the timestamp used during rendering. MapLibre will still render as fast as it can.

However, maybe you could combine it with manual repaint triggers to control rendering frequency:

```ts
map.stop();  // Stop automatic repaints
setInterval(() => {
    maplibregl.setNow(performance.now());
    map.triggerRepaint();
}, 100);  // Render only every 100ms instead of ~16ms
```

Hope that helps! Let me know if you have other questions.

bjperson:
@wayofthefuture @HarelM Do you think it would be valuable to add a documentation page and/or interactive example for the timeControl API?

I'd be happy to contribute that, though I'd need to study how your examples system works first (looking at files in `test/examples/` and the docs structure).

Let me know if this would be useful - I don't want to create work that might not fit your documentation strategy!

Some ideas:
- Simple timeControl demonstration (freeze/resume time, controlled animations)
- Animation testing example

What do you think?

wayofthefuture:
Wow that was really cool! I love the ability to throttle rendering, we have significant battery concerns so this is helpful.

Have you looked at browser.frame somewhere around map.ts render? I know that already throttles to the monitor/browser frame rate.

I'm open to anything that helps for easier understanding and implementation.

Thanks!

HarelM:
The example "system" is pretty straight forward, you create an HTML page in the examples and run the script to generate an image for it.
If you can think of an example that is small enough to pass the message of what the new API can do without tons of code, go ahead. but if you think that it will take a lot of code to show minimal value than I think the docs might be enough.
I'm good either way.

bjperson:
@wayofthefuture @HarelM

@HarelM You were right, the example system is indeed pretty straightforward! I was able to put one together: #6648

Let me know if you think any changes would make it clearer or more useful!

PR #6555, opened 2025-10-14
Deprecate unused cache timers

wayofthefuture:
Deprecate unused cache timers that serve no purpose. There are no other occurrences in the project that I could find.


HarelM:
I don't think it's worth a chancelog entry, looks good otherwise, thanks!

HarelM:
Dead code then, right?

wayofthefuture:
Yes

PR #6561, opened 2025-10-14
Abstract tile reload timers

wayofthefuture:
Abstracting tile reload timer logic to simplify source_cache.ts. Not sure if this needs a changelog...

### DIFF
--- a/src/ui/handler/scroll_zoom.test.ts
+++ b/src/ui/handler/scroll_zoom.test.ts
@@ -341,6 +341,7 @@ describe('ScrollZoomHandler', () => {
 
         expect(map.getZoom() - startZoom).toBe(0.0);
 
+        map.remove();
     });
 
     test('Gracefully handle wheel events that cancel each other out before the first scroll frame', () => {
@@ -364,6 +365,7 @@ describe('ScrollZoomHandler', () => {
         browserNow.mockReturnValue(now);
         map._renderTaskQueue.run();
 
+        map.remove();
     });
 
     test('does not zoom if preventDefault is called on the wheel event', () => {
@@ -433,6 +435,7 @@ describe('ScrollZoomHandler', () => {
         expect(startCount).toBe(1);
         expect(endCount).toBe(1);
 
+        map.remove();
     });
 
     test('emits one zoomstart event and one zoomend event while zooming', async () => {
@@ -481,6 +484,7 @@ describe('ScrollZoomHandler', () => {
         expect(startCount).toBe(1);
         expect(endCount).toBe(1);
 
+        map.remove();
     });
 
     test('Zooms for single mouse wheel tick while in the center of the map, should zoom to center', () => {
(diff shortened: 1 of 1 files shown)


wayofthefuture:
damn you fluke tests LOL! 😑

src/source/source_cache.ts
```diff
@@ -65,9 +65,7 @@ export class SourceCache extends Evented {
     _tiles: Record<string, Tile>;
     _prevLng: number;
     _cache: TileCache;
-    _timers: {
-        [_ in any]: ReturnType<typeof setTimeout>;
-    };
+    _reloadTimers: Record<string, ReturnType<typeof setTimeout>>;
```
HarelM: Can we keep the old variable name? While this is private, it may be used outside and if we can keep it, lets try that?

src/source/source_cache.ts
```diff
@@ -943,18 +933,42 @@ export class SourceCache extends Evented {
...
 
+    /**
+     * Set a timeout to reload the tile after it expires
+     */
     _setTileReloadTimer(id: string, tile: Tile) {
-        if (id in this._timers) {
-            clearTimeout(this._timers[id]);
-            delete this._timers[id];
-        }
+        if (!this.map?._refreshExpiredTiles) return;
```
HarelM: Is this new logic?

wayofthefuture: It's the same protection here moved to the lower-most point (i.e. the test was running expiry logic regardless of setting): https://github.com/maplibre/maplibre-gl-js/blob/bd3b89fc78394376099d97ca03bb53c96aacc9ce/src/source/vector_tile_source.ts#L244

HarelM: If this was added here, why this wasn't removed from there?
I'm probably missing a very basic thing here...

wayofthefuture: Redundancy and afraid to modify multiple files

HarelM: If this is only in order to refactor this part of the code, let's not change logic at all at this point, what do you think?

wayofthefuture: The problem is update() gets called on every frame which calls handleWrapJump which then causes a full O(n(tiles)) traversal on each frame

wayofthefuture: Nevermind I'm mistaken there is a check for that wrap jump... I'll remove it.

src/source/source_cache.test.ts
```diff
@@ -2323,6 +2323,8 @@ describe('SourceCache reloads expiring tiles', () => {
         const expiryDate = new Date();
         expiryDate.setMilliseconds(expiryDate.getMilliseconds() + 50);
         const sourceCache = createSourceCache({expires: expiryDate});
+        const map = {_refreshExpiredTiles: true} as Map;
```
HarelM: This can probably be reverted as well for now?

Issue #4364, opened 2024-07-05
Performance issue on large FeatureCollection GeoJSON source updates


zxwild:
## User Story
Greetings, I'm building a geojson drawing tool with maplibre and encounter a performance issue when I try to move/update a GeoJSON shape when there are a lot of shapes on the map.

## Rationale
I'm trying to offer a high performance for end users, so I want to be able to handle a few thousands of GeoJSON shapes and update them with my tool.

On the screenshot you can see the example how my map could look (it contains 2278 GeoJSON shapes).

For the first I had very slow performance when I used a source for each of these shapes. This case I needed to have "a source + 1-2 layers" for a shape, now I switched to a FeatureCollection source + limited number of sources (I have 5 overall).

Now performance is OK, it's possible to move, zoom a map without issues.

BUT, when I try to update a source, for example to move a shape, maplibre handles it very slow. It seems the whole canvas is rerendered or something like this (I use updateData to update source partially).

The question/offer: is it possible to have another canvas?

This case I'll be able:
 - to remove a target shape from the original canvas and keep its state static
 - add another temporary transparent canvas on top
 - perform my operations on the new canvas in the fast way
 - when operations are finished I can remove the new canvas and draw the updated feature on the original canvas

![2278-shapes](https://github.com/maplibre/maplibre-gl-js/assets/1629312/66d60ea6-886a-4c67-b8d3-82ef2cb5d458)

## Impact
Maplibre performance is OK when shapes are static, but shapes updating is a huge trouble.

HarelM:
Updating a geojson source rebuilds the tile pyramid every update, improving the performance will need to dive into the geojson-vt package I believe in order to solve this. 
I'm not sure I can think of a better solution, but different sources might be the right solution. 
I would recommend sharing a minimal reproduction, otherwise it's hard to really understand the problem.

zxwild:
Ok, I'll implement a demo, it would be something like 3000 GeoJSON features bundled in one FeatureCollection source, with a few layers.

I think dragging should be enougth: `mousedown` to select a Feature, `mousemove` to move it, `mouseup` to release.

zxwild:
Here is a demo, try to zoom, move map, and then drag any polygon (I adjusted feature amount to m1-max mac performance).

github repository:
https://github.com/zxwild/maplibre-performance

codepen:
https://codepen.io/zxwild/pen/WNqeaZM

[...]

PR #6562, opened 2025-10-14
Speed up GeoJSONSource#updateData

lucaswoj:
This PR speeds up `GeoJSONSource#updateData` when called with small diffs. Now we only reload tiles affected by the diff, rather than reloading all tiles. This significantly improves performance when making small changes to large GeoJSON datasets.

### Performance Impact

- 50% faster against my benchmark updating 1 point in a 100k point dataset
- Some use cases may benefit or more less, depending on specifics
- No performance regression expected for any use case

### Implementation

When `updateData` is called with a diff, the source now determines which tiles need reloading by:

- Checking if tiles contain features being updated or removed
- Testing if added/updated geometries intersect with tile bounds
- Falls back to full reload for large diffs (1000+ changes)

### Additional Changes

- Add `LngLatBounds.intersects()` method for bounds overlap testing
- Add `CanonicalTileID.toLngLatBounds()` to convert tile coordinates to geographic bounds
- Add benchmark to track `updateData` performance

Related to https://github.com/maplibre/maplibre-gl-js/issues/4364

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/test/examples/display-a-map.html
+++ b/test/examples/display-a-map.html
@@ -10,84 +10,18 @@
     <style>
         body { margin: 0; padding: 0; }
         html, body, #map { height: 100%; }
-        #time { position: absolute; top: 0; left: 0; }
     </style>
 </head>
 <body>
 <div id="map"></div>
-<div id="time"></div>
 <script>
-    const COUNT = 100_000;
-
     const map = new maplibregl.Map({
-        container: 'map',
-        style: 'https://demotiles.maplibre.org/style.json',
-        center: [0, 0],
-        zoom: 1,
-        maplibreLogo: true,
-        fadeDuration: 0
+        container: 'map', // container id
+        style: 'https://demotiles.maplibre.org/style.json', // style URL
+        center: [0, 0], // starting position [lng, lat]
+        zoom: 1, // starting zoom
+        maplibreLogo: true
     });
-
-    map.showTileBoundaries = true;
-
-    function getRandomPoint() {
-        return
-    }
-
-    map.on('load', () => {
-        map.addSource('test', {
-            type: 'geojson',
-            data: {
-                type: "FeatureCollection",
-                features: new Array(COUNT).fill(0).map((_, i) => ({
-                    type: "Feature",
-                    id: i,
-                    geometry: {
-                        type: "Point",
-                        coordinates: [
-                            -180 + Math.random() * 360,
-                            -85 + Math.random() * 170
-                        ]
-                    },
-                }))
-            }
-        });
-
-
-        map.addLayer({
-            id: 'test',
-            type: 'circle',
-            source: 'test',
-            paint: {
-                'circle-radius': ['case', ['==', ['id'], 0], 4, 1],
-                'circle-color': ['case', ['==', ['id'], 0], 'red', 'blue'],
-                'circle-opacity': ['case', ['==', ['id'], 0], 1, 0.2]
-            }
-        });
-
-        map.once('idle', () => {
-            test();
-        });
-    });
-
-    let i = 0;
-
-    function test() {
-        const start = performance.now()
-
-        map.getSource('test').updateData({
-            update: [{
-                id: 0,
-                newGeometry: {type: 'Point', coordinates: i++ % 2 ? [-185, 10] : [-175, 10] }
-            }]
-        });
-
-        map.once('idle', () => {
-            console.log('------ IDLE -----')
-            document.getElementById('time').innerText = Math.round(performance.now() - start) + "\n" + document.getElementById('time').innerText
-            test();
-        })
-    }
 </script>
 </body>
 </html>


test/bench/benchmarks/geojson_diff.ts
```diff
@@ -0,0 +1,85 @@
+import Benchmark from '../lib/benchmark';
+import createMap from '../lib/create_map';
+import type {Map} from '../../../src/ui/map';
+import type {GeoJSONSource} from '../../../src/source/geojson_source';
+
+let pointCount = 50_000;
```
lucaswoj: This bench takes about 45s to run. The large `pointCount` is useful for my purposes but we may want to dramatically decrease it before merging this into `main` to ensure the bench suite can finish in a reasonable timeframe.

src/geo/lng_lat_bounds.ts
```diff
@@ -286,6 +286,17 @@ export class LngLatBounds {
...
+    intersects(other: LngLatBoundsLike): boolean {
+        other = LngLatBounds.convert(other);
+
+        return !(
+            other.getWest() > this.getEast() ||
+            other.getEast() < this.getWest() ||
+            other.getNorth() < this.getSouth() ||
+            other.getSouth() > this.getNorth()
+        );
+    }
```
lucaswoj, line 336: This needs unit tests

src/source/geojson_source.ts
```diff
@@ -445,6 +449,47 @@ export class GeoJSONSource extends Evented implements Source {
         }
     }
 
+    _shoudReloadTile(diff: GeoJSONSourceDiff, tile: Tile) {
```
lucaswoj: This needs unit tests

test/examples/display-a-map.html
```diff
@@ -10,18 +10,84 @@
     <style>
         body { margin: 0; padding: 0; }
         html, body, #map { height: 100%; }
+        #time { position: absolute; top: 0; left: 0; }
```
lucaswoj: Revert this before merging

src/source/geojson_source.ts
lucaswoj: Add a bailout option if the diff is really big

src/source/geojson_source.ts
```diff
@@ -445,6 +448,49 @@ export class GeoJSONSource extends Evented implements Source {
...
+            if (ids.has(feature.id)) {
+                return true;
+            }
+        }
+
+        // Update all tiles that WILL contain an updated feature.
+        const geometries = [
+            ...update.map(f => f.newGeometry),
+            ...add.map(f => f.geometry),
+        ];
```
wayofthefuture: diff geometries are reconstructed for every tile request, maybe pass it as parameter (affectedGeometries?)

src/source/source_cache.ts
```diff
@@ -264,7 +267,9 @@ export class SourceCache extends Evented {
         this._cache.reset();
 
         for (const i in this._tiles) {
-            if (sourceDataChanged) {
+            if (shouldReloadTile && !shouldReloadTile(this._tiles[i])) {
```
wayofthefuture: confusing... typeof shouldReloadTile === 'function' ?

src/source/tile_id.ts
```diff
@@ -65,6 +66,19 @@ export class CanonicalTileID implements ICanonicalTileID {
...
     }
+
+    toLngLatBounds(extent: number = EXTENT, extentBuffer: number = 0) {
+        const buffer = extentBuffer / extent;
+        const width = ((extent - 1) / extent);
+
+        const lngMin = lngFromMercatorX((this.x - buffer) / Math.pow(2, this.z));
+        const latMin = latFromMercatorY((this.y + width + buffer) / Math.pow(2, this.z));
+
+        const lngMax = lngFromMercatorX((this.x + width + buffer) / Math.pow(2, this.z));
```
wayofthefuture: since you have `extent - 1` in width above, could this miss a small strip at the right edge of a tile?

src/source/geojson_source.ts
```diff
@@ -445,6 +448,49 @@ export class GeoJSONSource extends Evented implements Source {
...
+
+        const tileBounds = tile.tileID.canonical.toLngLatBounds(
+            this.workerOptions.geojsonVtOptions.extent,
+            this.workerOptions.geojsonVtOptions.buffer
+        );
+
+        for (const geometry of geometries) {
+            if (tileBounds.intersects(this._getGeoJSONBounds(geometry))) {
```
wayofthefuture: lots of geometry processing being reconstructed for each tile O(n^2), probably better to calculate geometry bounds and pass as parameter above... then you don't even need affectedGeometries above

wayofthefuture (commented):
Fine job Lucas I like the way this is headed... hopefully it does well with medium size updates i.e. 100 features?

src/source/geojson_source.test.ts
```diff
@@ -834,3 +834,114 @@ describe('GeoJSONSource.load', () => {
...
+        expect(result).toBe(false);
+    });
+
+    test('returns false when diff is empty', () => {
+        const result = shared(
+            new OverscaledTileID(0, 0, 0, 0, 0),
+            [],
+            {}
+        );
+        expect(result).toBe(false);
```
lucaswoj, line 961: TODO: Add a test for features crossing the international date line

lucaswoj:
> Fine job Lucas I like the way this is headed... hopefully it does well with medium size updates i.e. 100 features?

Roughly speaking, the performance gain will be a function how many features are _outside_ of the tiles that need to be updated. So, if there are 1000 features in the viewport, and 500 of them are in tiles that don't need to be updated, you will see about a 50% performance gain.

src/source/source_cache.ts
lucaswoj: What do you find confusing about this expression?

src/source/tile_id.ts
lucaswoj: Without `extent - 1`, this would include an extra strip at the right edge of the tile.

src/source/geojson_source.ts
```diff
@@ -445,6 +448,49 @@ export class GeoJSONSource extends Evented implements Source {
...
+
+        const {add = [], update = [], remove = []} = diff;
+
+        // For large diffs it's more efficient to reload all tiles.
+        const count = add.length + update.length + remove.length;
+        if (count > 1000) {
+            return true;
+        }
+
+        const ids = new Set([...update.map(u => u.id), ...remove]);
```
lucaswoj: Passing in the whole `diff` object as a parameter  provides better encapsulation for this method: it hides the implementation details of this method from its callers. This is a win for testability and maintainability.

This performance optimization has no measurable impact on our benchmarks. This method accounts for less than 0.1% of the runtime of the `GeoJSONSource#updateData` method.

So this is a case where I've chosen more maintainable code over "optimized" code

src/source/geojson_source.ts
lucaswoj: Same as https://github.com/maplibre/maplibre-gl-js/pull/6562/files/7cd8f1ee9f3a1ee17f3a907671bd9886f1d9982d#r2449602676

src/source/geojson_source.ts
lucaswoj: https://github.com/maplibre/maplibre-gl-js/pull/6562/files/7cd8f1ee9f3a1ee17f3a907671bd9886f1d9982d#r2449602676

src/source/geojson_source.ts
wayofthefuture: You can still pass in the whole diff object but use a closure to close over the diff and return a function, then the logic only executes once. Do the benchmarks account for larger feature updates with more complex geometries? If so then it may not be necessary, but I think that both optimization and readability can be maintained...

src/source/source_cache.ts
wayofthefuture: from a reading eye it looks like a boolean...

lucaswoj: This is a normal way to check for falsey values in JavaScript.

src/source/geojson_source.ts
lucaswoj: Thanks for the suggestion! You're right that we could use a closure here.

I ran benchmarks with up to 100,000 coordinates. The overhead of this method remained unmeasurable in those tests.

I prefer keeping the logic inline for now since it's clearer and the performance cost is negligible. If we see this become a bottleneck in profiling with real-world data, we can revisit.

Premature optimization is the root of all evil.

wayofthefuture: The code I'm referring doesn't refer to the base it refers to the number of features being modified, which you have capped at 1,000. If you are benchmarking against a single feature change then the test may be overfitted.

wayofthefuture: If the update is already capped then it does not appear to be "premature"

src/source/source_cache.ts
wayofthefuture: I was only speaking to readability... they can trace back to find out what it is... anyway, doesn't matter

src/source/geojson_source.ts
lucaswoj: I've pushed a new commit that ensures we construct `ids` and geometry `bounds` once per `GeoJSONSource#updateData` operation. When tested against a single feature with 100k coordinates, this yields a 0.008% speedup.

wayofthefuture: I know this will help the implementations where they are pushing single feature updates with thousand of points in one update geometry to append to a line, flight tracking, trails, etc. thanks.

Fyi typo in function name

lucaswoj:
@HarelM This is ready for review.

src/geo/lng_lat_bounds.ts
```diff
@@ -286,6 +287,53 @@ export class LngLatBounds {
         return containsLatitude && containsLongitude;
     }
 
+    /**
+     * Checks if this bounding box intersects with another bounding box.
+     *
+     * This method properly handles cases where either or both bounding boxes cross
+     * the antimeridian (date line).
+     */
+    intersects(other: LngLatBoundsLike): boolean {
```
lucaswoj, line 296: Question for @HarelM -- does this logic already exist somewhere in the codebase? Hard to believe we don't have something like this already.

src/source/geojson_source.ts
```diff
@@ -445,6 +448,46 @@ export class GeoJSONSource extends Evented implements Source {
...
     }
 
+    _getShouldReloadTile(diff?: GeoJSONSourceDiff): (tile: Tile) => boolean {
+        if (!diff || diff.removeAll) return undefined;
+
+        const {add = [], update = [], remove = []} = (diff || {});
+
+        const ids = new Set([...update.map(u => u.id), ...remove]);
+
+        if (ids.size + add.length > 1000) return undefined;
```
andrewda: We should probably document why 1000 items was chosen as the cutoff. Was this based on performance testing? What is the trade-off if we increase or decrease this threshold (more geojson bounds and intersection calculations?)

src/source/source_cache.ts
```diff
@@ -255,7 +255,10 @@ export class SourceCache extends Evented {
      * Reload tiles in this source. If source data has changed, reload all tiles using a state of 'expired',
      * otherwise reload only non-errored tiles using state of 'reloading'.
      */
-    reload(sourceDataChanged?: boolean) {
+    reload(
+        sourceDataChanged?: boolean,
+        shouldReloadTile?: (tile: Tile) => boolean
+    ) {
```
andrewda, line 262: Probably worth updating the function doc to mention how `shouldReloadTile` now plays into the reload logic.

andrewda (commented):
Overall looking good -- a couple minor documentation suggestions.

src/geo/lng_lat_bounds.ts
HarelM: I don't think so, or at least I haven't seen something similar...
There's polygon intersection, but it's not what you need here...

src/ui/events.ts
```diff
@@ -471,6 +471,13 @@ export type MapSourceDataEvent = MapLibreEvent & {
      * the event is related to loading of a tile.
      */
     tile: any;
+
+    /**
+     * Function that can be called to determine whether a tile should be reloaded
+     * after an source diff.
+     * @private
+     */
+    shouldReloadTile?: (tile: any) => boolean;
```
HarelM: Is there a different way to do this? This is an event that is fired, I'd hate to have a method attached to it, it should be DTO mostly - i.e. one way...

src/source/tile_id.ts
```diff
@@ -65,6 +66,18 @@ export class CanonicalTileID implements ICanonicalTileID {
     toString() {
         return `${this.z}/${this.x}/${this.y}`;
     }
+
+    toLngLatBounds(extent: number = EXTENT, extentBuffer: number = 0) {
```
HarelM: Is this the right place for this method? seems a bit odd... I would like to avoid linking between tiles and bounds if possible to remove circular references in the future, maybe, I don't know...

src/source/geojson_source.ts
lucaswoj: Great catch. As of the latest version of this PR, I don't think we need this heuristic anymore. Even if the user updates 100% of features in the source, this new logic doesn't add any significant overhead to `GeoJSON#updateData`. Removed in the latest commit.

src/source/source_cache.ts
lucaswoj: Done in latest commit (https://github.com/maplibre/maplibre-gl-js/pull/6562/commits/2564678f52fee47094cf0bfd99fb8aaf680f3a57)

src/ui/events.ts
lucaswoj: This event is the only mechanism that "connects" the `GeoJSONSource` to the `SourceCache` so somehow we need to encode this information on this object.

One different design I considered is attaching the `diff` object to the event. The downside of that design is that the `SourceCache` would need to add quite a bit of logic related to the `GeoJSONFeatureDiff` format, which is currently encapsulated in the `GeoJSONSource`

For what it's worth, this event already includes a `tile: Tile` property, which is a class with many methods. If this object were to get serialized, the `tile` property and `shouldReloadTile` properties would probably be omitted, and that's okay.

src/source/tile_id.ts
lucaswoj: I moved this to a standalone file in the latest commit (16a5eb828). Does that feel right to you?

src/ui/events.ts
HarelM: How about just sending a list of tiles ids that needs reload?

lucaswoj: Many events also include a `preventDefault` method and an instance of `Map`

src/source/tile_id_to_lng_lat_bounds.ts
```diff
@@ -0,0 +1,20 @@
+import {EXTENT} from '../data/extent';
+import {latFromMercatorY, lngFromMercatorX} from '../geo/mercator_coordinate';
+import {LngLatBounds} from '../geo/lng_lat_bounds';
+import type {CanonicalTileID} from './tile_id';
+
+export function tileIdToLngLatBounds(
+    {x,y,z}: CanonicalTileID,
+    extent: number = EXTENT,
+    extentBuffer: number = 0
```
HarelM: Can we use a single input here instead of two? It would also be better to keep both inputs in the same coordinates system...
I think it's good to have it in a separate file.

src/ui/events.ts
HarelM: True, but this method creates strong coupling between geojson source and source cache which I would like to break if possible.

src/source/tile_id_to_lng_lat_bounds.ts
lucaswoj: Done in cfb29924b

src/geo/lng_lat_bounds.ts
```diff
@@ -286,6 +287,50 @@ export class LngLatBounds {
...
+        const otherEast = wrap(other.getEast(), -180, 180);
+
+        // Check if either bounds wraps around the antimeridian
+        const thisWraps = thisWest > thisEast;
+        const otherWraps = otherWest > otherEast;
+
+        if (thisWraps && otherWraps) {
+            // Both wrap: they always intersect
+            return true;
+        } else if (thisWraps) {
```
HarelM: nit picking: I think some of the `else`s are not needed due to the return statements...?

src/ui/events.ts
lucaswoj: > How about just sending a list of tiles ids that needs reload?

`GeoJSONSource` doesn't know what tiles are loaded and doesn't have a reference to `SourceCache`. Even if it did, it's possible the set of loaded tiles could change between when we create this event object, and when the event is processed by `SourceCache`.

src/source/geojson_source.ts
```diff
@@ -445,6 +449,45 @@ export class GeoJSONSource extends Evented implements Source {
...
+        const ids = new Set([...update.map(u => u.id), ...remove]);
+
+        const boundsArray = [
+            ...update.map(f => f.newGeometry),
+            ...add.map(f => f.geometry)
+        ].map(g => this._getGeoJSONBounds(g));
+
+        return (tile: Tile) => {
+            // Update the tile if it PREVIOUSLY contained an updated feature.
+            const layers = tile.latestFeatureIndex.loadVTLayers();
```
HarelM: This is happening on the main thread, which is not great, can we move this logic to the worker thread somehow?

src/ui/events.ts
lucaswoj: > True, but this method creates strong coupling between `GeoJSONSource` and source cache which I would like to break if possible.

I respectfully disagree. The `shouldReloadTile` interface is generic. It exposes no `GeoJSONSource` implementation details to `SourceCache`. It's easy to imagine how other source types could use the same interface (e.g. we could eliminate the `Map#_refreshExpiredTiles` method)

src/geo/lng_lat_bounds.ts
lucaswoj: Each `else` statement has a comment nothing the case it handles. Is there a specific `else` statement you'd like me to explain?

src/ui/events.ts
HarelM: This back and forth between the geojson and the source cache is not a great pattern.
It's not coupling in terms of who is familiar with what, but more of a ping-pong between the two on a fired event.
It's a "code smell", and I'd like to avoid it if possible.
Can we send the list of "affected bounds" maybe and have the source cache (poor name BTW, it should be name source tile controller or manager) handle the logic of deciding if a tile should be reloaded, similar to what it does now, but with more logic?

HarelM: Or have the worker calculate the affected tiles and send that back to the main thread after the update has been made?

src/geo/lng_lat_bounds.ts
HarelM: No, they are well explained, but you can simply write `if` instead of `else if` since you are returning in each `if` statement.

src/source/geojson_source.test.ts
```diff
@@ -834,3 +834,117 @@ describe('GeoJSONSource.load', () => {
         expect(warnSpy).toHaveBeenCalledWith('No data or diff provided to GeoJSONSource id.');
     });
 });
+
+describe('GeoJSONSource._shoudReloadTile', () => {
+    function shared(id: CanonicalTileID, tileFeatures: Array<{id: string | number}>, diff: GeoJSONSourceDiff) {
```
HarelM: Why is this function called `shared`? I'm a bit confused by its name...

src/source/geojson_source.test.ts
```diff
@@ -834,3 +834,117 @@ describe('GeoJSONSource.load', () => {
...
+                get: (featureIndex: number) => ({featureIndex})
+            },
+            loadVTLayers: () => ({
+                _geojsonTileLayer: {
+                    feature: (i: number) => tileFeatures[i] || {}
+                }
+            })
+        } as any;
+
+        return (source._getShouldReloadTile(diff) || (() => true))(tile);
```
HarelM: I would move this part out to highlight what the test is actually testing I think...

src/source/geojson_source.test.ts
lucaswoj: Renamed to `shouldReloadTile` in 88c13d665

src/source/geojson_source.test.ts
lucaswoj: I hope that 88c13d665 helps address this comment. This `shared` function is meant to be a helper function that abstracts over all the setup for `getShouldReloadTile` in the individual test cases. As of https://github.com/maplibre/maplibre-gl-js/commit/88c13d6652e8932b62c75370e9ee8a49647da397 the helper method is called `shouldReloadTile`, improving readability of the test cases.

HarelM: I would use something like `beforeEach` to indicate that this is a setup that happens for all the tests and have the relevant logic in each test.
Right now, while keeping the DRY principle, the tests are not detailed enough to understand what is tested in each test.

I would advice to refactor it a bit to the AAA pattern - Arrange (probably in the before each), Act - run the shouldReloadTile method, Assert - check the return value.

src/geo/lng_lat_bounds.ts
lucaswoj: Done in 76565ca54

src/ui/events.ts
lucaswoj: > Can we send the list of "affected bounds" maybe and have the source cache ... handle the logic of deciding if a tile should be reloaded

I did not architect the code that way because 

 - it add adds more complexity to SourceCache, which is already 1200 LOC and has a great many responsibilities already 
 - it will leak implementation details from GeoJSONSource to SourceCache, such as how to read vtLayers and iterate over its feature ids 
 - it makes this PR longer and more complex 
 - it adds friction to improving this feature in the future (more lines of code need to change to accommodate additional performance improvements) 
 - it reduces the reusability of this feature, no other source types need to refresh tiles by bbox AFIK. Other source types do need to refresh other tiles by other criteria.

lucaswoj: > Or have the worker calculate the affected tiles and send that back to the main thread after the update has been made?

Let's consolidate the discussion about offloading more logic to the workers at https://github.com/maplibre/maplibre-gl-js/pull/6562#discussion_r2450574761

src/source/geojson_source.ts
lucaswoj: There are several reasons why I chose to run this operation on the main thread rather than in a worker: 

1. We must avoid race conditions in `SourceCache`. If we sent a list of currently loaded tiles to the worker, and then had to await a response about which to refresh, it's possible some other tiles would load in the interim and not be properly handled. We effectively need a lock on the `TileCache` while deciding what to refresh. 

2. This operation runs so fast that the overhead of sending messages to the worker would probably outweigh the runtime of the logic itself. A diff with 100k coordinates is processed in about a millisecond. 

3. Some of this data only lives on the main thread. Sending it back to the worker is extremely expensive and/or potentially destructive.

src/ui/events.ts
wayofthefuture: I think that the diff should be passed in the event as sourceDataDiff.

```ts
this.fire(new Event('data', {...eventData, sourceDataType: 'content', sourceDataDiff: diff}));
```

Source_cache is the primary handler of tile reloading, is essentially and may be renamed to source_controller in the future, and is the place for the logic.

A bigger problem is that source_cache(controller) may need to be extended in the future (or merged somehow).

Nevertheless I agree with Harel it should be DTO one direction.

HarelM: I would like to keep events in the library as simple as possible, and would like to avoid sending lambdas/methods on the event data.
The reason for this is from an architecture perspective.
I would like to keep the events in one of the following options to make sure we follow this architecture:
```mermaid
sequenceDiagram
SourceCache->>(geojson)Source: loadTile
(geojson)Source-->>SourceCache: sourcedata
SourceCache->>SourceCache: reload relevant tiles
```
Or
```mermaid
sequenceDiagram
SourceCache->>(geojson)Source: loadTile
(geojson)Source-->>SourceCache: sourcedata
SourceCache->>(geojson)Source: reload relevant tiles
```

The idea behind this is to make sure we keep the events as a simple object that only passes data, which allows at a later state to be able to use message bus, worker communication or anything else that you would be able to pass data only and not functions.
Function should be used as a last resort, and I don't think this case calls for it.
While I agree it's not ideal, and we do pass a tile object on the event (which I would prefer if we didn't), I don't want to continue in that direction.
It also keep the hierarchy well defined and easy to follow - methods calls from one side and events from the other - it keeps the definition of who "knows" who better.
In this case SourceCache is familiar with a source, but a source does not know anything about the SourceCache, nor does it passes methods to it to run.

If there's no way to do it this way, then let's discuss what alternatives there are.

src/source/geojson_source.ts
HarelM: The reason I asked about utilizing the worker is mainly because the call to the worker is happening anyway, and this part is done only after the worker call returns.
So I thought about utilizing the worker for this.
Besides the keeping the main thread free it also feel like the worker is doing some of the relevant calculations in order to update the the geojson index so I though it might be a good place to know which tiles changed due to the diff request and return those.
It felt more natural to me to do these calculations in the worker instead of parsing the results that the worker sent back after they were encoded, and do the decoding on the main thread.
While it might improve performance, which is good, I'm more interested in trying to avoid decoding and encoding to reduce code complexity and keeping the responsibility of the geojson data in the worker.

If you have a specific race condition in mind that is relevant to this example I'd love to better understand it and maybe think about a solution together.

While I don't have strong feelings about moving this logic to the worker, it does feel like a better place for it, mainly because there a need to decode back the data here, which seems a bit odd.

HarelM:
Summary of current status:
I've added my input to the two main discussions: one about the event callback and the other about moving some of the logic to the worker.
I'd be happy to hear your input on those.
Once we settle those, the test code should be updated to be a bit more readable and less "DRY" to allow for easier understanding when reading each test separately.

Thanks a lot for all your hard work on this! It's truly appreciated.

src/source/geojson_source.test.ts
lucaswoj: I have rewritten the tests in b096625e306c199faaa448bb344aa52fa27ddb8d. Do you fund them more readable now?

src/source/geojson_source.ts
lucaswoj: Are you blocking the PR on this comment? Or do you not feel strongly?

src/ui/events.ts
lucaswoj: > The idea behind this is to make sure we keep the events as a simple object that only passes data, which allows at a later state to be able to use message bus, worker communication or anything else that you would be able to pass data only and not functions.

I understand the goal of keeping events simple for future flexibility. However, this requirement doesn't align with the reality of our codebase or any particular plans for its future.

80% (10 out of 13) of the event types defined in `events.ts` contain functions or objects with functions.

`VectorSource` already handles a similar problem with tile expiration. It directly mutates `Tile` objects owned by `SourceCache` and then calls a private method on `Map`. If we're concerned about architecture, that pattern is more problematic than passing a function on an event.

> While I agree it's not ideal, and we do pass a tile object on the event (which I would prefer if we didn't), I don't want to continue in that direction.

I agree that the way `SourceCache` and `Source` communicate with each other is confusing and complex. This PR isn't the place to address that.

> If there's no way to do it this way, then let's discuss what alternatives there are.

I don't think there's a way to fundamentally eliminate the need for `SourceCache` to execute some logic defined by `GeoJSONSource`. All alternatives I can think of have significant downsides:

**Adding `refreshTileIds` based on currently loaded tiles:** `GeoJSONSource` would need to know what tiles `SourceCache` has loaded. That's a private implementation detail not currently shared with any other part of the system. Additionally, because this operation goes through the event system, which is asynchronous, `SourceCache` may have loaded new tiles between when the event is fired and when it's processed. This would leave the user stuck seeing stale data that should've been updated in the `updateData` operation.

**Adding `refreshTileIds` with every possible tile ID:** This could produce millions of tile IDs that would need to be stored in memory and iterated over by `SourceCache`.

**Adding `nextBounds`, `prevIds`, `extent`, and `buffer` to the event:** This creates worse coupling. `SourceCache` would need be told about the specific settings we're passing `geojson-vt`, how to calculate tile coordinates for GeoJSON shapes, and contain unit tests to exercise this logic.

**Having SourceCache call `GeoJSON#_shouldUpdateTile` directly:** This increases race condition risk. We need to associate the expired tiles list with this specific update operation, not with the source itself. If the user sends many rapid `updateData` events, there could be a mismatch between `GeoJSONSource`'s internal state and the actual update operation that needs to be performed.

Is there a design I'm missing or do you think one of these alternatives is actually better than the current approach?

HarelM: > Adding refreshTileIds with every possible tile ID: This could produce millions of tile IDs

What about the "affected bounds" instead of a list of tiles?

> Adding nextBounds, prevIds, extent, and buffer to the event

Is it possible to create a PR against this PR to show how this would look? 
If you pass in the event a "bag of parameters" that later on the source cache will send back to the geojson source it might allow removing the method and pass only data, which is what I would like to achieve.

lucaswoj: > If you pass in the event a "bag of parameters" that later on the source cache will send back to the geojson source it might allow removing the method and pass only data, which is what I would like to achieve.

Done in 45860bce3

wayofthefuture:
I had to stop working mid-commit but I wanted to share how I believe it should be done... it's not the best solution but I think it is a good alternative:

https://github.com/maplibre/maplibre-gl-js/commit/44356887ba58d747d0cedf7198b97b71f2f91aca

src/source/geojson_source.test.ts
HarelM: Yes, thanks!

src/source/geojson_source.ts
```diff
@@ -279,8 +295,11 @@ export class GeoJSONSource extends Evented implements Source {
      * @returns a promise which resolves to the source's boundaries
      */
     async getBounds(): Promise<LngLatBounds> {
+        return this._getGeoJSONBounds(await this.getData());
+    }
+
+    _getGeoJSONBounds(data: GeoJSON.GeoJSON): LngLatBounds {
```
HarelM: This method can be moved to a file of its own, or added to one of the util files and doesn't have to be a part of the geojson source class as it is a "pure" function.

HarelM:
> I had to stop working mid-commit but I wanted to share how I believe it should be done... it's not the best solution but I think it is a good alternative

I think it's a good alternative.
I think both options are good and follow the architecture I highlighted before, the current one (not the one you "mid-commit" one) feels a bit more OOP since there's no mention of the specific geojson source in the source cache as it only knows about the source interface which has an optional `shouldReloadTile` method, so it's more generic in that sense.
But if you think the second option is better that's fine too, I'm not sure I have strong feeling either way.

I can't help but wonder if there's no part of the logic that we can do in the worker? doesn't have to be all the work, but maybe just part of it, as it has the geometry data already parsed.
To your previous question about the worker, no, it's not blocking, but it would be great if you could take a look and see if there can be an easy win by doing some part of the work there beforehand.

src/source/geojson_source.ts
HarelM: See the current last comment in PR main thread.

wayofthefuture:
My vision is of SourceController, GeoJSONSourceController, and RasterSourceController, which extend and have their own custom methods with a separate reload function for reloading GeoJSON tiles. The encompassing source cache is ripe for extension?

Also I think shouldReloadTile should be renamed and placed inside Tile class as tile.isAffectedBy(featureIds, featureBounds) or something like that.

src/source/geojson_source.ts
lucaswoj: Done in 1a0514024

lucaswoj:
> not the one you "mid-commit" one

This was posted by @wayofthefuture, not me.

> I can't help but wonder if there's no part of the logic that we can do in the worker? doesn't have to be all the work, but maybe just part of it, as it has the geometry data already parsed.

In eb692742c I moved the calculation of `nextBounds` to the worker. 

This does not make the code measurably faster in any benchmark I have run. It does make the code harder to read and harder to test.

> it has the geometry data already parsed. 

There is no "parsed geometry data" on the worker that we can reuse. For the purposes of this feature, the worker has exactly the same data as the main thread.

wayofthefuture:
Oh well, now I'm just confused. In a perfect world geojson source would always control source but it doesn't. That's our reality. Source cache does the controlling. reloadTile, loadTile, addTile lives in source cache. The decision to reload is currently in source cache, splitting that decision and creating types to transport a function is confusing. The hodgepodge that is source cache is a problem for another day. I'm not a fan of shouldReloadTile in the current form... and I'm definitely not a fan of shouldReloadTileOptions so that a function can be passed... respectfully... my opinion. Sorry

HarelM:
> This does not make the code measurably faster in any benchmark I have run. It does make the code harder to read and harder to test.

I tend to agree, sorry for pushing in the wrong direction, I think this can be reverted if it's not a big ask. 
Otherwise, this looks good. 
Thanks!

HarelM:
I meant the last change with the worker, the rest is ok.
Source cache is a mess, we'll eventually need to confront with, but not today...

lucaswoj:
> I meant the last change with the worker, the rest is ok. Source cache is a mess, we'll eventually need to confront with, but not today...

Reverted in cea0e71f9

src/source/geojson_source.ts
```diff
@@ -36,6 +38,21 @@ export type GeoJSONSourceInternalOptions = {
     generateId?: boolean;
 };
 
+/**
+ * @internal
+ */
+export type GeJSONSourceShouldReloadTileOptions = {
```
HarelM: `geojson` - typo...

src/source/geojson_source.ts
```diff
@@ -453,6 +443,55 @@ export class GeoJSONSource extends Evented implements Source {
...
+
+        return {
+            nextBounds,
+            prevIds
+        };
+    }
+
+    /**
+     * Determine whether a tile should be reloaded based on a set of options associated with a
+     * `MapSourceDataChangedEvent`.
```
HarelM: you can use `{@link MapSourceDataChangedEvent}` I think...?
If this fails the docs generation than you can ignore this comment.

src/source/source.ts
```diff
@@ -121,6 +121,12 @@ export interface Source {
      * Optional function to redefine how tiles are loaded at high pitch angles.
      */
     calculateTileZoom?: CalculateTileZoomFunction;
+    /**
+     * Optional function to determine whether a tile should be reloaded, given a
+     * set of options associated with a `MapSourceDataChangedEvent`.
+     * @internal
+     */
+    shouldReloadTile?(tile: Tile, options: any): boolean;
```
HarelM: Can we have the options with the current options type? I prefer not to have `any` and since this is only one we have at the moment, let's use that for now.

src/source/geojson_source.ts
HarelM: Maybe even move it to the source interface definition and call it `ShouldReloadTileOptions`?

HarelM:
Added a few last final comments, and I'll merge this afterwards.
Thanks for all the hard work on this!

src/source/geojson_source.ts
lucaswoj: Fixed typo in c3f8b0378. I don't think we should call this `ShouldReloadTileOptions` because it'd be a mistake to use with any source type other than `GeoJSONSource`.

src/source/geojson_source.ts
lucaswoj: Done in 4dcc783c

src/source/source.ts
lucaswoj: Done in b828bdaa3

test/examples/display-a-map.html
HarelM: Please revert...

lucaswoj: Done in c77ab5b1e71452a33ba85c38651bd498a3ba9dd8

wayofthefuture:
@lucaswoj I'm not sure if this PR is the cause, but see the behavior difference between 5.10.0 and 5.11.0 where this commit was merged. Notice how the extruded boxes are sticking around and not being removed...

Before:

https://github.com/user-attachments/assets/5cc1a01d-7b76-48fa-8891-881ca0538ec1

After:

https://github.com/user-attachments/assets/8f723ad5-4537-4a08-8577-0a43b9cde14b

All you have to do is change the version number in this codepen to reproduce:
https://codepen.io/wayofthefuture/pen/JoGRwKE?editors=1000

PR #6570, opened 2025-10-15
Add support for MLT

### LINKED ISSUE #6258: Support MapLibre Vector Tiles (MLT)
## User Story

As an integrator, I can render vector data from an MLT source so that storage, decoding, and rendering efficiency are improved, and I can take advantage of additional capabilities like M-values (elevation, etc.) and direct-to-GPU vertex data.

## Rationale

MLT is designed to provide several benefits over MVT, see [github.com/maplibre/maplibre-tile-spec](https://github.com/maplibre/maplibre-tile-spec) for details.

HarelM:
## Launch Checklist

- Resolves #6258 

This PR will introduce support for MLT format.

 - [x] Link to related issues.


HarelM:
@Salkin975 I've opened the following branch for integration, and this PR to see that we keep this branch working and up-to-date with the main branch.:
https://github.com/maplibre/maplibre-gl-js/tree/mlt

I tried to copy over the code from your branch but I'm not sure this is the right approach to solve it.
Most of the code is duplicated between mlt and vector source when there's no real need to copy it.

I would recommend as an initial step to only try and make mlt work in maplibre without optimizing the buckets and other things.
Once we have it working with some basic tests for lines, fill, symbols etc, we can start implementing the optimizations that are needed.
What do you think?

package.json
```diff
@@ -25,6 +25,7 @@
     "@mapbox/vector-tile": "^2.0.4",
     "@mapbox/whoots-js": "^3.1.0",
     "@maplibre/maplibre-gl-style-spec": "^24.3.0",
+    "@maplibre/mlt": "^0.0.1-alpha.12",
```
HarelM: This need to be a "production" release before this branch can be merged.

HarelM:
Unfortunately, coverage report does not work on the worker as is, and there might be other configuration needed in order to make it work, but the code is fully covered.

src/source/vector_tile_source.ts
```diff
@@ -61,6 +61,7 @@ export class VectorTileSource extends Evented implements Source {
     maxzoom: number;
     url: string;
     scheme: string;
+    encoding: string;
```
acalcutt, line 64: Does this encoding conflict with 'encoding' used for a terrain source?  maybe not because there are not vector terrain sources right now, but it wonder if it should this be more like 'format', which usually tells what the tile format is in metadata. (like (pbf, mvt, image type in 'format' metadata tag)

HarelM: I would recommend posting this message on the design proposal in the style spec repo instead of here, this is merely the implementation of the agreed upon design.
Although at this point it might be a bit late...

HarelM:
For future reference, I've used the following page to check the geometry rendering and conversion,
I'll see if I can add some test to cover this functionality.
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <title>Display a map</title>
    <meta charset='utf-8'>
    <style>
    </style>
</head>
<body>
<div id="map"></div>
<script>
        style: 'https://harelm.github.io/demotiles/tiles-mlt/plain.json', // style URL
        zoom: 1, // starting zoom
        maplibreLogo: true
    });

    map.once('load', () => {
        map.addSource('gj', {
            type: 'geojson',
            data: []
        });
        map.addLayer({
            id: 'filll',
            type: 'fill',
            source: 'gj',
            filter: ["==", "$type", "Polygon"],
            paint: {
                'fill-color': 'red'
            }
        });

        map.addLayer({
            id: 'lineee',
            type: 'line',
            source: 'gj',
            filter: ["==", "$type", "LineString"],
            paint: {
                'line-color': 'blue'
            }
        });

        map.addLayer({
            id: 'pointtt',
            type: 'circle',
            source: 'gj',
            filter: ["==", "$type", "Point"],
            paint: {
                'circle-color': 'green'
            }
        });

        map.on("mousemove", (e) => {
            const features = map.queryRenderedFeatures(e.point);
            if (features.length > 0) {
                console.log(features.map(f => f.geometry));
                map.getSource('gj').setData({ 
                    type: 'FeatureCollection',
                    features: features.filter(f => f.source !== 'gj')
                })
            }
        })
    })
</script>
</body>
</html>
```

mwilsnd (approved):

PR #6573, opened 2025-10-15
refactor: use timeControl.now() instead of browser.now()

bjperson:
This refactoring removes the  export from  and updates all usages to directly import from . This change:

- Removes  wrapper (33 occurrences across codebase)
- Updates source files to import  from './util/time_control'
- Renames local variables from  to  to avoid shadowing
- Updates test files to mock instead of
- Removes the  test from browser.test.ts (already covered in time_control.test.ts)

All tests passed successfully.

Related issue : [6543](https://github.com/maplibre/maplibre-gl-js/issues/6543)

## Launch Checklist

 - [x] Link to related issues.
 - [x] Updates test files.


src/ui/handler/cooperative_gestures.test.ts
```diff
@@ -1,5 +1,5 @@
 import {describe, beforeEach, test, expect, vi} from 'vitest';
-import {browser} from '../../util/browser';
+import * as timeControl from '../../util/time_control';
```
HarelM, line 2: Is it possible to do it without this type of import? I'm trying to avoid using it...

HarelM:
THANKS!
I left a nit-picking comment, looks great otherwise.

src/ui/handler/cooperative_gestures.test.ts
bjperson: I also try to avoid using import * as when possible, but in this case it's necessary for vi.spyOn() to work properly - it requires an object containing the function to spy on, while import {now} gives us the function directly.

I noticed this pattern is already used in MapLibre's test suite (e.g., src/render/draw_symbol.test.ts line 11: import * as symbolProjection from '../symbol/projection' for the same mocking purpose).

That said, if you prefer using vi.mock() instead, I'm happy to refactor the tests! It would be more verbose but would avoid the namespace imports. Let me know your preference! 😊

HarelM: Whatever you think is best and worth the effort. I thought this was the reason, but I wanted to be sure...
Your call.

HarelM (approved):

src/ui/handler/cooperative_gestures.test.ts
bjperson: Thanks for letting me decide! I'll keep it as is since the pattern is already used in the codebase (symbolProjection, ajax, etc.) and vi.spyOn() is cleaner than vi.mock() for this use case. Appreciate you checking! 😊

bjperson:
Thanks for the quick review and merge, Harel! Appreciate your time and feedback on this refactoring. 🙂

Issue #6441, opened 2025-09-19
source.setClusterOptions() stopped updating cluster options


plixplox:
**maplibre-gl-js version**: >=5.6.2

**browser**: chrome 140

### Steps to Trigger Behavior

 1. Add data via addSource
 2. Toggle clustering on/off using the dropdown with source.setClusterOptions
 3. Nothing happens, but the console shows the warning: "No data or diff provided to GeoJSONSource earthquakes"

### Link to Demonstration

https://jsbin.com/jediqeb/edit?html,output

### Expected Behavior

Switching should update the source without needing to recreate the source/layers (as in 5.6.1).

### Actual Behavior

In 5.6.2, changes are not applied, and clustering remains in the previous state.

HarelM:
- I believe this regression was introduced by #6163

@andrewda can you take a look at this please?
It looks like the following line is getting hit since the geojson data did not change as part of this call...
https://github.com/maplibre/maplibre-gl-js/blob/1bb2d2473d06e8aa2d5f81f627b0d6337ad4b287/src/source/geojson_source.ts#L387-L390

andrewda:
@HarelM I'll take some time to dig into this week 👀

HarelM:
@andrewda any updates on this?

andrewda:
@plixplox @HarelM #6603 resolves this issue!

PR #6603, opened 2025-10-22
Fix cluster options with no pending data changes

andrewda:
Fixes https://github.com/maplibre/maplibre-gl-js/issues/6441

Prior to this change, `setClusterOptions` would not trigger a worker update when no data changes were pending. This PR adds a check to trigger an empty update if there is no pending `data` or `dataDiff`.

## Launch Checklist

 - [x] Link to related issues.
 - [x] Post benchmark scores.

### DIFF
--- a/src/source/geojson_worker_source.test.ts
+++ b/src/source/geojson_worker_source.test.ts
@@ -344,18 +344,13 @@ describe('loadData', () => {
     });
 
     test('loadData should process cluster change with no data', async () => {
-        const spy = vi.fn();
-        const mockCreateGeoJSONIndex: typeof createGeoJSONIndex = (data, params) => {
-            spy(params);
-            return createGeoJSONIndex(data, params);
-        };
-
+        const mockCreateGeoJSONIndex = vi.fn(createGeoJSONIndex);
         const worker = new GeoJSONWorkerSource(actor, layerIndex, [], mockCreateGeoJSONIndex);
 
         await worker.loadData({source: 'source1', data: JSON.stringify(updateableFeatureCollection), cluster: false} as LoadGeoJSONParameters);
-        expect(spy.mock.calls[0][0].cluster).toBe(false);
+        expect(mockCreateGeoJSONIndex.mock.calls[0][1].cluster).toBe(false);
         await expect(worker.loadData({cluster: true} as LoadGeoJSONParameters)).resolves.toBeDefined();
-        expect(spy.mock.calls[1][0].cluster).toBe(true);
+        expect(mockCreateGeoJSONIndex.mock.calls[1][1].cluster).toBe(true);
     });
 });
 


src/source/geojson_source.test.ts
```diff
@@ -303,17 +303,15 @@ describe('GeoJSONSource.update', () => {
...
         const mockDispatcher = wrapDispatcher({
             sendAsync(message) {
-                expect(message.type).toBe(MessageType.loadData);
-                expect(message.data.cluster).toBe(true);
-                expect(message.data.superclusterOptions.radius).toBe(80 * EXTENT / source.tileSize);
-                expect(message.data.superclusterOptions.maxZoom).toBe(16);
-                spy();
-                return Promise.resolve({});
+                spy(message);
+                return new Promise((resolve) => {
```
HarelM: I think there's a sleep utility method for this...?

src/source/geojson_source.test.ts
```diff
@@ -303,17 +303,15 @@ describe('GeoJSONSource.update', () => {
         expect(spy).toHaveBeenCalled();
     });
 
-    test('modifying cluster properties after adding a source', () => {
+    test('modifying cluster properties after adding a source', async () => {
```
HarelM, line 306: Should there be 2 tests? One with pending data and the other without?

src/source/geojson_source.test.ts
andrewda: Looks like the timeout wasn't actually necessary in this test -- I reverted back to `return Promise.resolve({});`

HarelM:
Thanks, did you take a look at #6562 to make sure the changes there won't break this fix? Just to be in the safe side...

src/source/geojson_source.test.ts
andrewda: Good call -- added a second test for modifying cluster properties with a pending set.

andrewda:
> Thanks, did you take a look at https://github.com/maplibre/maplibre-gl-js/pull/6562 to make sure the changes there won't break this fix? Just to be in the safe side...

@HarelM Yep, these changes should be independent from one another.

src/source/geojson_source.ts
```diff
@@ -309,12 +309,18 @@ export class GeoJSONSource extends Evented implements Source {
...
-            }
+        if (options.clusterRadius !== undefined) {
+            this.workerOptions.superclusterOptions.radius = this._pixelsToTileUnits(options.clusterRadius);
+        }
+        if (options.clusterMaxZoom !== undefined) {
+            this.workerOptions.superclusterOptions.maxZoom = this._getClusterMaxZoom(options.clusterMaxZoom);
         }
+
+        // If no pending updates, create an empty diff to trigger a recluster
+        if (this._pendingWorkerUpdate.data === undefined && this._pendingWorkerUpdate.diff === undefined) {
```
HarelM: This is probably checked elsewhere, maybe a small method should be used instead?

HarelM: The comment above it was an indication a method is probably missing 😀

HarelM:
Looks good, added a last nit-picking comment. I'll merge this afterwards.

src/source/geojson_source.ts
andrewda: Added a new `_hasPendingWorkerUpdate` utility method in the latest commit 👍

andrewda:
Standby... Found one more issue that I'm working to resolve.

src/source/geojson_source.ts
```diff
@@ -309,12 +313,18 @@ export class GeoJSONSource extends Evented implements Source {
...
-                this.workerOptions.superclusterOptions.maxZoom = this._getClusterMaxZoom(options.clusterMaxZoom);
-            }
+        if (options.clusterRadius !== undefined) {
+            this.workerOptions.superclusterOptions.radius = this._pixelsToTileUnits(options.clusterRadius);
+        }
+        if (options.clusterMaxZoom !== undefined) {
+            this.workerOptions.superclusterOptions.maxZoom = this._getClusterMaxZoom(options.clusterMaxZoom);
         }
+
+        // If no pending updates, create an empty diff to trigger a recluster
```
HarelM: I would place the comment inside the if to explain why the empty object is created, the other stuff is now self explanatory since the method name is good.

HarelM:
Ok, let me know when you are done as in terms of code change, this can be merged I believe.

andrewda:
@HarelM Code changes are done, and ready for one more pass. The previous implementation did not work for non-updateable GeoJSON objects, so I've tried to make it more generic.

src/source/geojson_worker_source.ts
```diff
@@ -248,6 +255,10 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
         }
     }
 
+    isClustered(): boolean {
```
HarelM: Is this needed only for the tests? If so I would recommend finding a different way to test this instead of adding a public method.

src/source/geojson_worker_source.ts
```diff
@@ -117,7 +120,11 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
 
         this._pendingRequest = new AbortController();
         try {
-            this._pendingData = this.loadAndProcessGeoJSON(params, this._pendingRequest);
+            // Load and process data if no data has been loaded previously, or if there is
+            // a new request, data, or dataDiff to process.
```
HarelM, line 132: Can you elaborate why was this added?
I was under the impression the previous change was only in the geojson_source and not the worker, right?

andrewda: Prior to https://github.com/maplibre/maplibre-gl-js/pull/6163, `setClusterOptions` would update the worker using the last `_data` stored in the `GeoJSONSource`. This effectively meant that using `setClusterOptions` was incompatible with sending a diff to the worker, since the diff would be reset to the last `_data` state as soon as `setClusterOptions` was called. See the following test:

<details>
<summary>Test showing that data would get reset after sending a diff</summary>

```js
test('modifying cluster properties after sending a diff', async () => {
    const spy = vi.fn();
    const mockDispatcher = wrapDispatcher({
        sendAsync(message) {
            spy(message);
            return Promise.resolve({});
        }
    });
    const geoJsonData = {
        'type': 'FeatureCollection',
        'features': [
            {
                'type': 'Feature',
                'id': 1,
                'properties': {},
                'geometry': {
                    'type': 'Point',
                    'coordinates': [-122.48369693756104, 37.83381888486939]
                }
            }
        ]
    } as GeoJSON.GeoJSON;

    const source = new GeoJSONSource('id', {
        type: 'geojson',
        data: geoJsonData,
        cluster: false,
        clusterMaxZoom: 8,
        clusterRadius: 100,
        clusterMinPoints: 3,
        generateId: true
    }, mockDispatcher, undefined);

    // Wait for initial data to be loaded
    source.load();
    await waitForEvent(source, 'data', (e: MapSourceDataEvent) => e.sourceDataType === 'metadata');

    spy.mockReset();

    const diff = {remove: [1]};
    source.updateData(diff);
    source.setClusterOptions({cluster: true, clusterRadius: 80, clusterMaxZoom: 16});

    await waitForEvent(source, 'data', (e: MapSourceDataEvent) => e.sourceDataType === 'metadata');

    // Performs an initial diff to remove a feature
    expect(spy.mock.calls[0][0].data.cluster).toBe(false);
    expect(spy.mock.calls[0][0].data.dataDiff).toEqual(diff);
    expect(spy.mock.calls[1][0].data.cluster).toEqual(true);

    // FAILS prior to #6163 -- data was set back to the original geoJsonData
    expect(spy.mock.calls[1][0].data.data).not.toBeDefined();

    // FAILS after #6163 -- `setClusterOptions` does not trigger a request to the worker with no pending data
    expect(spy).toHaveBeenCalledTimes(2);
});
```

</details>

The `optionsChanged` flag in this PR allows this test to pass, but will then cause errors in the worker because no `data` or `dataDiff` was passed in the message. With that context in mind, we need a way to send a request to the worker to perform a reload with no new data, only changes to the options. This change makes that possible by not performing a GeoJSON load if no data or diff is provided -- only updating the `_geoJSONIndex` based on the previous data and new options.

src/source/geojson_worker_source.ts
andrewda: Correct, this is only needed for testing. The only other way I can think to check whether the options were correctly processed would be to test `worker._geoJSONIndex` directly, but that seems more fragile than having a simple public method.

src/source/geojson_worker_source.ts
HarelM: Ahh, so this is another issue your are fixing here. Makes sense.

src/source/geojson_worker_source.ts
HarelM: Yeah, this is a dependency injection issue basically as the class is being created inside instead of a factory.
I'm wondering if this can be improved a bit by adding a factory method to the constructor of this class? What do you think?
I honestly don't like adding public methods only for test or checking the content of internal fields...

andrewda: @HarelM sorry for the delay -- took a stab at improving this in 7f7092cf7d7793cf0c1be3a85d2719e8f99d6e31. Is that close to what you had in mind?

HarelM (approved):
THANKS!!

PR #6617, opened 2025-10-26
Define how much to slice and how much to scale

HarelM:
## Launch Checklist

This is a PR to showcase a different approach to overzoom with more control on when to slice and when to scale.


HarelM:
CC: @wayofthefuture

Issue #6607, opened 2025-10-23
Canvas layer renders as a black square if the source canvas is square with a size of 64*2^n (two's power)


tavimori:
**maplibre-gl-js version**: 5.9.0

**browser**: Chrome (and others?)

### Steps to Trigger Behavior

1. Create a canvas layer with `canvas.width==canvas.height==64*2^n`, where `n=0,1,2,...`. 
2. The map will render the layer as a black square, regardless of its contents.

### Link to Demonstration

https://jsbin.com/godiyil/edit?html,output

### Expected Behavior

The two squares should both be blue.

### Actual Behavior

Only the right one is blue; the left one is black due to its canvas size.

Omkarthipparthi:
Interesting! I can reproduce this - power-of-2 canvas dimensions render black while non-power-of-2 work fine. 
I'm taking a look at this one. Thank you for reporting.

HarelM:
Thanks for taking the time to report this issue!
This is indeed an interesting bug...

obeyeur:
Hello, I still have this issue when using an ImageSource (not CanvasSource), with the setting "raster-resampling: 'nearest'". Tested with the latest 5.22 version

HarelM:
@obeyeur did you try applying the same fix to image source? I think the PR only fixes canvas source...

PR #6624, opened 2025-10-28
Fix canvas sources with power-of-2 dimensions rendering as black squares

Omkarthipparthi:
## Description
Fixes a bug where canvas sources with power-of-2 dimensions (64×64, 128×128, 256×256, 512×512, etc.) render as black squares instead of displaying their actual content. Non-power-of-2 dimensions were rendering correctly.

Closes [#6607](https://github.com/maplibre/maplibre-gl-js/issues/6607)

## Problem
When using canvas sources with dimensions that are powers of 2 (where `width == height == 64 * 2^n` for `n = 0, 1, 2, ...`), the canvas renders as a solid black square regardless of its actual content. Non-power-of-2 dimensions (e.g., 129×128) render correctly.

### Root Cause
The canvas source implementation in `src/source/canvas_source.ts` was missing a critical `bind()` call when creating WebGL textures. Without explicit texture parameter configuration:

1. **WebGL defaults** set `MIN_FILTER` to `NEAREST_MIPMAP_LINEAR`, which requires mipmaps
2. **Power-of-2 textures** are treated specially by WebGL - they can use mipmaps and repeat wrapping
3. When WebGL expects mipmaps but none exist, the texture becomes **incomplete**
4. **Incomplete textures render as solid black** (0, 0, 0, 1) in WebGL
5. **Non-power-of-2 textures** automatically fall back to `LINEAR` filtering in most browsers, which is why they worked

The bug existed because `CanvasSource` didn't call `bind()` to set proper texture parameters, unlike `ImageSource` and `VideoSource` which both correctly include this call.

## Solution
Added the missing `texture.bind(gl.LINEAR, gl.CLAMP_TO_EDGE)` call in `src/source/canvas_source.ts` (line 181), matching the pattern used in `ImageSource` and `VideoSource`.

### Changes
**File: `src/source/canvas_source.ts`**

```typescript
// Before (missing bind):
if (!this.texture) {
    this.texture = new Texture(context, this.canvas, gl.RGBA, {premultiply: true});
}

// After (with bind):
if (!this.texture) {
    this.texture = new Texture(context, this.canvas, gl.RGBA, {premultiply: true});
    this.texture.bind(gl.LINEAR, gl.CLAMP_TO_EDGE);  // ← Added
}
```

## Testing
- Linting passes with no errors
- **Added new render test**: `test/integration/render/tests/canvas/power-of-two`
  - Creates clean map with 64x64 (power-of-2) blue canvas source
  - Uses `addFakeCanvas` to simulate the canvas source
  - Verifies the output matches the expected blue square
  - Test passes with the fix and failed without the fix
- All existing canvas render tests pass:
  - `canvas/default` ✓
  - `canvas/update` ✓
- Manually tested in local, works well now

## Checklist
- [x] Confirmed no backports from Mapbox projects
- [x] Changes described above
- [x] Linked to issue [#6607](https://github.com/maplibre/maplibre-gl-js/issues/6607)
- [x] No visual changes to existing functionality (bug fix only)
- [x] Linting passes
- [x] Integration tests pass for canvas sources
- [x] No public API changes
- [x] CHANGELOG.md updated


HarelM:
Thanks a lot for providing this fix! 
Can you please add a render test (preferably) or a browser test to cover this scenario?

HarelM:
I don't see the blue square in the render test, are you sure the tests falls without your fix?

Omkarthipparthi:
Let me make sure that and get back to you with a more solid test

Omkarthipparthi:
I went ahead with a browser test approach and this is much clear and straight forward now.

HarelM:
Browser tests are used to check interactive behavior mostly, if it's possible to do it in a render test I would prefer that, since this is a "static" scenario, and render tests are optimal for this.

Omkarthipparthi:
@HarelM Apologies for the delay. Here you go!
I have written the render test. Please review and let me know. Thank you.

HarelM (approved):

HarelM:
Looks good! Thanks! 
Just a quick question before I merge this, can you make sure the test fails without your fix?

HarelM:
I saw you updated the initial post with the answer. Thanks!

PR #6625, opened 2025-10-28
offsetLine to handle linestrings with duplicate points

### LINKED ISSUE #5431: Mouse events are not triggering for layers with line-offset at a high zoom level
**maplibre-gl-js version**: 5.1.0

**browser**:  Chrome 131.0.6778.205 

### Steps to Trigger Behavior

 1. Zoom in to a very high zoom level
 2. Hover with the mouse over one of the line features
 3. Zoom in at a high zoom level, hover again. At a certain zoom level the mousemove event is not triggered any more. Note: this only seems to happend for lineString features with more than 2 coordinates. In the demo there wil be one feature with 2 coordinates, the bug will not occur for that feature.

### Link to Demonstration

[JS Bin](https://jsbin.com/kulupin/edit?html,output)

### Expected Behavior
The mousemove event is triggered.

### Actual Behavior
The mousemove event is not triggered.

mmc1718:
These changes update the `offsetLine` function in `query_utils` used to offset line geometries as part of querying whether a given linestring intersects a point or bounding box/polygon.

In cases of linestrings containing duplicate consecutive points:

- offsetLine removes duplicate consecutive values from the line before performing the offset
- In the case of no change across multiple points, points are skipped

This ensures offsetLines do not contain null values.

Fixes #5431 

## Launch Checklist

 - [x] Link to related issues.

### DIFF
--- a/src/style/query_utils.ts
+++ b/src/style/query_utils.ts
@@ -77,26 +77,14 @@ export function offsetLine(rings: Array<Array<Point>>, offset: number) {
     for (let ringIndex = 0; ringIndex < rings.length; ringIndex++) {
         const ring = _stripDuplicates(rings[ringIndex]);
         const newRing: Array<Point> = [];
-        const zeroPoint = new Point(0, 0);
         for (let index = 0; index < ring.length; index++) {
+            const point = ring[index];
+            const prevPoint = ring[index - 1];
+            const nextPoint = ring[index + 1];
             // perpendicular unit vectors (outward unit normal vector):
             // these indicate which direction the segments should be offset in
-            let unitNormalAB: Point;
-            let unitNormalBC: Point;
-            const point = ring[index];
-            if (index === 0) {
-                unitNormalAB = new Point(0, 0);
-            } else {
-                const prevPoint: Point = ring[index - 1];
-                unitNormalAB = point.sub(prevPoint)._unit()._perp();
-            }
-            if (index === ring.length - 1) {
-                // final iteration; no next point
-                unitNormalBC = new Point(0, 0);
-            } else {
-                const nextPoint: Point = ring[index +1];
-                unitNormalBC = nextPoint.sub(point)._unit()._perp();
-            }
+            const unitNormalAB: Point = index === 0 ? new Point(0, 0) : point.sub(prevPoint)._unit()._perp();
+            const unitNormalBC: Point = index === ring.length - 1 ? new Point(0, 0) : nextPoint.sub(point)._unit()._perp();
             // unit bisector direction
             const bisectorDir = unitNormalAB._add(unitNormalBC)._unit();
             const cosHalfAngle = bisectorDir.x * unitNormalBC.x + bisectorDir.y * unitNormalBC.y;


src/style/style_layer/line_style_layer.test.ts
```diff
@@ -48,4 +52,75 @@ describe('LineStyleLayer', () => {
         lineLayer.setPaintProperty('line-gradient', null);
         expect(lineLayer.gradientVersion).toBeGreaterThan(gradientVersion);
     });
+
+    describe('queryIntersectsFeature', () => {
+        afterEach(() => {
```
HarelM: why `afterEach` and not `beforeEach`?

test/integration/query/tests/line-offset/overzoom-precision/style.json
```diff
@@ -0,0 +1,326 @@
...
+            "type": "geojson",
+            "maxzoom": 10,
+            "data": {
+                "type": "Feature",
+                "properties": {
+                    "internalId": 1
+                },
+                "geometry": {
+                    "type": "LineString",
+                    "coordinates": [
```
HarelM: Can we use a smaller example in terms of number of points?

src/style/query_utils.ts
```diff
@@ -62,21 +62,47 @@ export function offsetLine(rings: Array<Array<Point>>, offset: number) {
...
-            const c = ring[index + 1];
-            const aToB = index === 0 ? new Point(0, 0) : b.sub(a)._unit()._perp();
-            const bToC = index === ring.length - 1 ? new Point(0, 0) : c.sub(b)._unit()._perp();
-            const extrude = aToB._add(bToC)._unit();
-
-            const cosHalfAngle = extrude.x * bToC.x + extrude.y * bToC.y;
-            if (cosHalfAngle !== 0) {
-                extrude._mult(1 / cosHalfAngle);
+        let refPoint;
+        ring.forEach((point, index, ring) => {
```
HarelM: Can we use a regular `for` here? I find `forEach` a paint in terms of debug stack etc.

src/style/query_utils.ts
```diff
@@ -62,21 +62,47 @@ export function offsetLine(rings: Array<Array<Point>>, offset: number) {
...
+            let aToB: Point;
+            let bToC: Point;
+            let nextPoint: Point;
+            if (index === 0) {
+                // first iteration, set refPoint as current point
+                refPoint = point;
+                aToB = new Point(0, 0);
+            } else if (refPoint.x === point.x && refPoint.y === point.y) {
```
HarelM: I think point class has `isEqual`?

HarelM:
Since we are touching this code, what about making it more robust by extending the offset lines and trimming and clipping when needed or is it doing it already?
I would like to have something like the following if possible.
https://github.com/maplibre/maplibre-gl-js/pull/2260

src/style/style_layer/line_style_layer.test.ts
mmc1718: No particular reason, I've updated this

test/integration/query/tests/line-offset/overzoom-precision/style.json
mmc1718: Updated.

I've changed the test name too; while doing this I realized there are actually duplicate consecutive points in this input example, so the duplicates in this case aren't caused by high precision of the points. So I've updated the test name to reflect this (don't know how I missed that...)

src/style/query_utils.ts
mmc1718: Updated

src/style/query_utils.ts
mmc1718: Thanks for the tip, there's an `equals` method. Updated this

mmc1718:
@HarelM thanks for the review! I have addressed your points and also rebased changes on main.

Re trimming and clipping: the linked issue seems to be about rendering/the way the offset looks on the map. From what I could see this `offsetLine` function is only used when checking for intersections, not for actually drawing them - is that right? Or do you mean that trimming and clipping here could improve the accuracy when querying?

Let me know if I've missed something.

Also, I wasn't sure if I should run benchmarks tests and post the results here - if you think it's not needed I'll remove that point from the checklist

HarelM:
Currently, lines with offset are drawn badly since they don't take into consideration these considerations. 
This PR is not about rendering, but I wonder how complicated it would be to improve this, and hopefully later on improve the rendering as well...

src/style/query_utils.ts
```diff
@@ -62,20 +62,47 @@ export function offsetLine(rings: Array<Array<Point>>, offset: number) {
...
+                bToC = new Point(0, 0);
+            } else {
+                nextPoint = ring[index +1];
+                if (point.equals(nextPoint)) {
+                    bToC = new Point(0, 0);
+                } else {
+                    bToC = nextPoint.sub(point)._unit()._perp();
+                }
+            }
+            const zeroPoint = new Point(0, 0);
```
HarelM: Better to move this before the loop to avoid creating many points, right?

src/style/query_utils.ts
```diff
@@ -62,20 +62,47 @@ export function offsetLine(rings: Array<Array<Point>>, offset: number) {
...
-            const b = ring[index];
-            const c = ring[index + 1];
-            const aToB = index === 0 ? new Point(0, 0) : b.sub(a)._unit()._perp();
-            const bToC = index === ring.length - 1 ? new Point(0, 0) : c.sub(b)._unit()._perp();
-            const extrude = aToB._add(bToC)._unit();
-
-            const cosHalfAngle = extrude.x * bToC.x + extrude.y * bToC.y;
-            if (cosHalfAngle !== 0) {
-                extrude._mult(1 / cosHalfAngle);
+            let aToB: Point;
```
HarelM: I think this isn't really the aToB vector, but a perpendicular unit vector, it might be worth changing the variable name to match that.
I think this is how it used to be before, but I found the code really hard to read due to these non-accurate variables names.

src/style/query_utils.ts
```diff
@@ -62,20 +62,47 @@ export function offsetLine(rings: Array<Array<Point>>, offset: number) {
     for (let ringIndex = 0; ringIndex < rings.length; ringIndex++) {
         const ring = rings[ringIndex];
         const newRing: Array<Point> = [];
+        let refPoint: Point;
```
HarelM: Wouldn't it be simpler if instead of keeping a refPoint just to make sure that there are no duplications we could simply go over the ring, filter out consecutive duplicate points and then run the rest of the algorithm as it was before?
I think it will make the condition inside simpler and also clearly state what happens in each "phase".
I.e. 
1. clear duplicate points
2. offset the line

src/style/query_utils.ts
```diff
@@ -62,20 +62,47 @@ export function offsetLine(rings: Array<Array<Point>>, offset: number) {
...
+            } else {
+                nextPoint = ring[index +1];
+                if (point.equals(nextPoint)) {
+                    bToC = new Point(0, 0);
+                } else {
+                    bToC = nextPoint.sub(point)._unit()._perp();
+                }
+            }
+            const zeroPoint = new Point(0, 0);
+            if (aToB.equals(zeroPoint) && bToC.equals(zeroPoint)) {
```
HarelM: I think the previous version of this algorithm was handling the case of bToC equals 0,0 differently, didn't it?

src/style/query_utils.ts
mmc1718: Updated

src/style/query_utils.ts
mmc1718: Thanks, yes I found it quite confusing. The math is a bit new to me but I've tried updating the variable names, hopefully it's clearer

src/style/query_utils.ts
mmc1718: I thought about this, I agree it would be more readable and easier to reason about.

It would mean looping over the ring twice - once to remove duplicates/create a copy with duplicates removed, and then to perform the offset - which I thought could slightly impact performance, since currently there is only one loop.

Only consecutive duplicates should be removed, so I can't think of a quick way to filter them out without looping (e.g. converting to a Set to make points unique wouldn't be a good option in this case).

Do you think looping twice is worth it?

src/style/query_utils.ts
mmc1718: bToC equalling 0 wasn't explicitly handled, neither was aToB equalling 0

If both aToB and bToC are zero, the value of `extrude` ends up being a point with NaN values, and these NaN values end up in the final output. This check is added to ensure results with NaN values are skipped

mmc1718:
@HarelM Re improving the way offset lines are drawn: I initially assumed that since there is no call to `offsetLine` in e.g. `src/render/draw_line` there must be a different piece of code responsible for offsetting lines when drawing.

However, I have since realised that the value of `geometry` is actually updated in `queryIntersectsFeature` : 

https://github.com/maplibre/maplibre-gl-js/blob/a59f8ffeb810a2800402d429cb730fcf8bf442c8/src/style/style_layer/line_style_layer.ts#L111

So I guess this must be the actual geometry, and this function runs at loading so the geometry is already offset by the time the line is drawn? I thought I read something about when `queryIntersectsFeature` is run but I can't find it anymore.

I'll have a look at trimming/clipping, I can add some helper functions for `offsetLine`

src/style/query_utils.ts
HarelM: Yes, looping twice will not impact performance due to how loop performance work...

HarelM:
Draw line is using buckets for rendering, the geometry here is used only for querying.
I don't know how offset is actually calculated, but I'm guessing it is done in the worker thread and not here. 
But it's out of scope for this PR.

src/style/query_utils.ts
HarelM: Oh, great!

src/style/query_utils.ts
mmc1718: I've updated this so that duplicate points are clearer before the line is offset

src/style/query_utils.ts
```diff
@@ -57,25 +57,57 @@ export function translate(queryGeometry: Array<Point>,
...
+            if (index === ring.length - 1) {
+                // final iteration; no next point
+                unitNormalBC = new Point(0, 0);
+            } else {
+                const nextPoint: Point = ring[index +1];
+                unitNormalBC = nextPoint.sub(point)._unit()._perp();
+            }
+            if (unitNormalAB.equals(zeroPoint) && unitNormalBC.equals(zeroPoint)) {
+                // no change means final point will be a duplicate of the previous; can be excluded
+                continue;
```
HarelM: Given that there are no duplicate consecutive points, does this `continue` can ever be hit?
From the code coverage, it seems like it's not getting hit, so I'm wondering if this can actually happen...?

src/style/query_utils.ts
```diff
@@ -57,25 +57,57 @@ export function translate(queryGeometry: Array<Point>,
...
-            const extrude = aToB._add(bToC)._unit();
-
-            const cosHalfAngle = extrude.x * bToC.x + extrude.y * bToC.y;
+            // perpendicular unit vectors (outward unit normal vector):
+            // these indicate which direction the segments should be offset in
+            let unitNormalAB: Point;
+            let unitNormalBC: Point;
+            const point = ring[index];
+            if (index === 0) {
+                unitNormalAB = new Point(0, 0);
```
HarelM: I kinda like the previous code where it's clear that the reason for these ifs is to calculate the value of `unitNormalAB` and `unitNormalBC`, so I would recommend writing something like it was before, but with better names:
```ts
const prevPoint = ring[index - 1];
const point = ring[index];
const nextPoint = ring[index + 1];
const unitNormalAB = index === 0 ? new Point(0, 0) : point.sub(prevPoint)._unit()._perp();
const unitNormalBC = index === ring.length - 1 ? new Point(0, 0) : nextPoint.sub(point)._unit()._perp();
```

HarelM:
I've updated the code a bit, left a few last code review comments which afterwards this can be merged.

src/style/query_utils.ts
mmc1718: Good point, I've removed this. Doesn't seem like it can be hit

src/style/query_utils.ts
mmc1718: I've updated this

mmc1718:
@HarelM thanks for the feedback, I've updated changes

I pulled your commits and then rebased locally on main to update the branch but it seems like something has gone wrong here as this PR now includes all those changes... I'm looking into it, could try reverting the new commits. If you can see what happened or help let me know

(this sounds like a similar situation: https://stackoverflow.com/questions/79441392/changes-from-the-main-branch-appearing-in-pr-after-rebase-pulling-after-a-rebas so I'm reading those replies)

HarelM:
Consider updating the main branch on your fork, create a new branch from it and add your changes to it instead of fighting with git. my 2 cents.

mmc1718:
Thanks for the tip. I've created https://github.com/maplibre/maplibre-gl-js/pull/6745 to replace this one, it includes all the relevant commits.

PR #6635, opened 2025-10-31
Refactor SourceCache to TileManager

wayofthefuture:
### Summary
This PR refactors the `SourceCache` class to `TileManager` to better reflect its actual responsibilities within the rendering pipeline.

### Rationale

The name `SourceCache` has been misleading and it significantly understates what this class actually does. After considering alternatives like `SourceController` and `SourceManager`, I believe `TileManager` to be the most accurate name.

#### What This Class Actually Does

**Beyond simple caching:**
- **Tile Lifecycle Management** - Creates, loads, unloads, and reloads tiles; manages state transitions (loading → loaded → reloading → expired → errored)
- **Viewport-Driven Selection** - Calculates ideal tiles for the current viewport, determines retention strategies for parent/child tiles, handles world wrap-around
- **Direct Render Pipeline Integration** - Called by `Painter` during render passes, provides renderable tile IDs to drawing functions
- **Advanced Rendering** - Implements raster and symbol fading, DEM backfilling for terrain, feature state updates across tiles

#### Why "TileManager"?

- **Tiles are the primary concern** - The class manages the lifecycle of `Tile` objects
- **Manager pattern** - It orchestrates multiple subsystems (source, cache, painter, terrain) around tile operations
- **Pipeline integration** - Deeply embedded in the rendering flow, not just a cache layer
- **Broader scope** - Makes high-level decisions about tile existence, loading, rendering, retainment, and disposal

### Implementation & Notes

- Class renamed from `SourceCache` to `TileManager`
- All references updated throughout the codebase
- Variable names updated (e.g., `sourceCache` → `tileManager`)
- Filename changed to `tile_manager.ts`
- New directory called `tile` that includes `tile_manager` as well as the adjacent `tile_` files.
- Create abstraction between `source` directory and `tiles` directory.
- Tees up the extension of this tile manager class to additional child classes based on source types.


wayofthefuture:
💥

src/render/draw_fill_extrusion.ts
```diff
@@ -8,15 +8,15 @@ import {
...
 import type {FillExtrusionStyleLayer} from '../style/style_layer/fill_extrusion_style_layer';
 import type {FillExtrusionBucket} from '../data/bucket/fill_extrusion_bucket';
-import type {OverscaledTileID} from '../source/tile_id';
+import type {OverscaledTileID} from '../tile/tile_id';
 
 import {updatePatternPositionsInProgram} from './update_pattern_positions_in_program';
 import {translatePosition} from '../util/util';
 
-export function drawFillExtrusion(painter: Painter, source: SourceCache, layer: FillExtrusionStyleLayer, coords: Array<OverscaledTileID>, renderOptions: RenderOptions) {
+export function drawFillExtrusion(painter: Painter, manager: TileManager, layer: FillExtrusionStyleLayer, coords: Array<OverscaledTileID>, renderOptions: RenderOptions) {
```
HarelM: nit-picking: This can probably use the full name (`tileManager`) like in other files.

src/ui/map.ts
```diff
@@ -2107,12 +2107,12 @@ export class Map extends Camera {
...
-        if (source === undefined) {
-            this.fire(new ErrorEvent(new Error(`There is no source with ID '${id}'`)));
+        const manager = this.style && this.style.tileManagers[id];
+        if (manager === undefined) {
+            this.fire(new ErrorEvent(new Error(`There is no tile manager with ID '${id}'`)));
             return;
         }
-        return source.loaded();
+        return manager.loaded();
     }
```
wayofthefuture, line 2116: I don't like this but attempting to refactor only...

src/render/draw_fill_extrusion.ts
wayofthefuture: I knew you were going to say that!

src/render/draw_raster.ts
```diff
@@ -36,13 +36,13 @@ const cornerCoords = [
...
 
-export function drawRaster(painter: Painter, sourceCache: SourceCache, layer: RasterStyleLayer, tileIDs: Array<OverscaledTileID>, renderOptions: RenderOptions) {
+export function drawRaster(painter: Painter, tileManager: TileManager, layer: RasterStyleLayer, tileIDs: Array<OverscaledTileID>, renderOptions: RenderOptions) {
     if (painter.renderPass !== 'translucent') return;
     if (layer.paint.get('raster-opacity') === 0) return;
     if (!tileIDs.length) return;
 
     const {isRenderingToTexture} = renderOptions;
-    const source = sourceCache.getSource();
+    const source = tileManager.getSource();
```
HarelM, line 45: [Very optional] maybe `getSource` should be renamed to `getTileSource`? I'm not sure about it, but reading `tileManager.getSource()` was weird...

src/render/draw_raster.ts
```diff
@@ -159,7 +159,7 @@ function getFadeProperties(tile: Tile, sourceCache: SourceCache, fadeDuration: n
 
     // cross-fade with parent first if available
     if (tile.fadingParentID) {
-        const parentTile = sourceCache._getLoadedTile(tile.fadingParentID);
+        const parentTile = tileManager._getLoadedTile(tile.fadingParentID);
```
HarelM: Maybe we should remove the `_` from `_getLoadedTile` if it is being used from outside the class? just a thought.

src/render/terrain.ts
```diff
@@ -233,7 +233,7 @@ export class Terrain {
         // create matrix for lookup in dem data
         const matrixKey = sourceTile && (sourceTile + sourceTile.tileID.key) + tileID.key;
         if (matrixKey && !this._demMatrixCache[matrixKey]) {
-            const maxzoom = this.sourceCache.sourceCache._source.maxzoom;
+            const maxzoom = this.tileManager.tileManager._source.maxzoom;
```
HarelM: Can we use `getSource()` here instead of `_source`?
This is not related to this PR, but since I'm reviewing it I thought we might just improve the code if possible...

src/render/draw_raster.ts
wayofthefuture: Source is so ubiquitous I think keep it

src/source/query_features.ts
```diff
@@ -136,27 +136,27 @@ export function queryRenderedFeatures(
             queryResults: tileIn.tile.queryRenderedFeatures(
                 styleLayers,
                 serializedLayers,
-                sourceCache._state,
+                tileManager._state,
```
HarelM: same here for `_state`, we should think about if this should be kept with `_` if it is used from "outside".

src/source/query_features.ts
```diff
@@ -136,27 +136,27 @@ export function queryRenderedFeatures(
...
                 tileIn.queryGeometry,
                 tileIn.cameraQueryGeometry,
                 tileIn.scale,
                 params,
                 transform,
                 maxPitchScaleFactor,
-                getPixelPosMatrix(sourceCache.transform, tileIn.tileID),
+                getPixelPosMatrix(tileManager.transform, tileIn.tileID),
```
HarelM: Not that I think it is easy to change, but is the tileManager's transform different from the one that is passed here as part of the parameters of this method?
Probably a different PR to address this though...
I'm mainly highlighting that this is strange...

src/source/query_features.ts
wayofthefuture: Maybe getState() to match getSource()... corresponding to this._state, this._source

src/tile/tile_cache.ts
```diff
@@ -7,7 +7,7 @@ import type {Tile} from './tile';
  * with hash lookup made possible by keeping a list of keys in parallel to
  * an array of dictionary of values
  *
- * source_cache offloads currently unused tiles to this cache, and when a tile gets used again,
+ * tile_manager offloads currently unused tiles to this cache, and when a tile gets used again,
```
HarelM: `TileManager` instead?

src/source/query_features.ts
wayofthefuture: It's the same transform, both tracing back to map.transform. I'll fix.

src/ui/map.ts
```diff
@@ -2107,12 +2107,12 @@ export class Map extends Camera {
      * ```
      */
     isSourceLoaded(id: string): boolean {
-        const source = this.style && this.style.sourceCaches[id];
-        if (source === undefined) {
-            this.fire(new ErrorEvent(new Error(`There is no source with ID '${id}'`)));
+        const manager = this.style && this.style.tileManagers[id];
```
HarelM: `tileManager`

src/tile/tile_manager.ts
```diff
@@ -1210,8 +1216,8 @@ export class SourceCache extends Evented {
     }
 }
 
-SourceCache.maxOverzooming = 10;
-SourceCache.maxUnderzooming = 3;
+TileManager.maxOverzooming = 10;
```
HarelM, line 1223: Why are these defined here? are these static members?

wayofthefuture: A long running question I have had. The variables are also named backwards fyi.

HarelM:
I've added mostly minor comments, thanks for making this happen!

src/source/query_features.ts
HarelM: Sounds good.

HarelM:
I had a radical idea to make everything that is not used outside the class "private".
But it would probably break all kind of plugins not to mention the tile manager tests that relay heavily on private members.
But probably outside the scope of this PR...

src/tile/tile_manager.ts
HarelM: Can we move them inside the class and add a `static` modifier to them? I think this is a old javascript way of doing static variables...

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 ### ✨ Features and improvements
 - Improve performance of `GeoJSONSource#updateData` when called on small diffs ([#6562](https://github.com/maplibre/maplibre-gl-js/pull/6562))
-- _...Add new stuff here..._
+- Refactor SourceCache to TileManager ([#6635](https://github.com/maplibre/maplibre-gl-js/pull/6635))
```
```suggestion
- ⚠️ Refactor SourceCache to TileManager ([#6635](https://github.com/maplibre/maplibre-gl-js/pull/6635)) - this is not a breaking change since SourceCache is not part of the public API, but if you have a plugin that uses internal stuff, it might break...
```

HarelM (approved):

HarelM:
Nice work! Thanks!

src/tile/tile_manager.ts
HarelM: I saw that you merged the PR  so consider doing this in a different PR then...

wayofthefuture: Sorry I didn't see these above! I'm about to open another PR in the next 5 min though... it should be draft a while so we can do it there... same file

wayofthefuture:
BOOM?

HarelM:
BOOM!

PR #6638, opened 2025-10-31
Tile Manager Composition Classes, Vector & Raster Tile Strategies, Tile Reload Strategies

wayofthefuture:
**Overview**

This PR refactors the tile management system by converting the `TileManager` class to a composition class with `Raster` and `Vector` manager strategies. The new strategy classes are responsible for the rendering behavior specific to their source types.

**Key Changes**

This change removes a significant amount of source-type–specific branching from the base manager. Code paths such as raster fade animations and vector symbol fading are now encapsulated within their respective classes instead of being handled through conditional logic in a unified class.

**Additional Improvements**

In addition, this PR introduces the boiler-plate for separate `TileReloadStrategy` interfaces to allow different source types to define how their tiles are refreshed. This enables custom reload behavior (e.g., specialized handling for GeoJSON sources) without requiring changes to the shared tile management code.

**Benefits**

Overall, this refactor reduces complexity, improves separation of concerns, and makes it easier to introduce new tile source types or behaviors in the future.

### DIFF
--- a/test/build/min.test.ts
+++ b/test/build/min.test.ts
@@ -38,7 +38,7 @@ describe('test min build', () => {
         const decreaseQuota = 4096;
 
         // feel free to update this value after you've checked that it has changed on purpose :-)
-        const expectedBytes = 1012983;
+        const expectedBytes = 1015572;
 
         expect(actualBytes, `Consider changing expectedBytes to: ${actualBytes}`).toBeLessThan(expectedBytes + increaseQuota);
         expect(actualBytes).toBeGreaterThan(expectedBytes - decreaseQuota);


src/style/style.ts
```diff
@@ -982,7 +985,14 @@ export class Style extends Evented {
         const shouldValidate = builtIns.indexOf(source.type) >= 0;
         if (shouldValidate && this._validate(validateStyle.source, `sources.${id}`, source, null, options)) return;
         if (this.map && this.map._collectResourceTiming) (source as any).collectResourceTiming = true;
-        const tileManager = this.tileManagers[id] = new TileManager(id, source, this.dispatcher);
+
+        if (isRasterType(source.type)) {
```
HarelM: Probably better to move this into a factory method, maybe even inject this method to allow testing to mock this part...

src/tile/vector_tile_manager.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @internal
+ * `VectorTileManager` extends `TileManager` with vector-specific functionality,
+ * primarily handling symbol fade logic for smooth label transitions.
+ *
+ * Responsibilities specific to vector tiles:
+ *  - Managing symbol fade hold duration to prevent label flicker
+ *  - Retaining tiles with symbols for gradual fade out
+ *  - Coordinating symbol placement across tile boundaries
+ */
+export class VectorTileManager extends TileManager {
```
HarelM: Is it possible to use composition instead of inheritance for this PR? I know it's a big shift, but I prefer composition if possible, make the readability of the code easier...

HarelM:
I'm not sure if this should be part of this PR or not, but there's also the terrain tile manager, or something similar, which I think differs in the loading tiles strategy mainly, and there's also the contour source that needs ability to load different tiles, I think, I'm not sure how all of this is related, but if we are considering redesigning this class, it's worth taking a broader look on what parts can me modularized.

src/tile/vector_tile_manager.ts
wayofthefuture: I'll look into it. I thought you may have an architecture change so I wanted to throw it up as draft. Out of time gtg.

wayofthefuture:
> I'm not sure if this should be part of this PR or not, but there's also the terrain tile manager, or something similar, which I think differs in the loading tiles strategy mainly, and there's also the contour source that needs ability to load different tiles, I think, I'm not sure how all of this is related, but if we are considering redesigning this class, it's worth taking a broader look on what parts can me modularized.

Yes it is totally different. I'm afraid to touch it LOL.

src/tile/vector_tile_manager.ts
HarelM: Early feedback is a good idea for these kind of changes, thanks!

wayofthefuture: Composition draft --- better? not sure...

HarelM: There's now both composition and inheritance? so I think it's confusing...
But I like the fact that the strategy part is the composition, makes a lot of sense...

wayofthefuture: So I've been reading and I think that composition might not work well in the case where the shared component has lots of public methods to expose. In this case, Tile Manager would have 20+ public methods to expose, meaning that Vector/Raster classes would carry those 20+ delegates to the tile manager. See:

https://en.wikipedia.org/wiki/Composition_over_inheritance?#Drawbacks

On another note however, the abstract class and abstract function keyword in the previous implementation requires the extending classes to have those functions (acting like an interface). So it may be better to revert back to the previous method, then see how we can make it easier to understand from there. Thoughts?

HarelM: Are you sure about the public API of tile manager? I think that I came across like 5-9 methods. 

Generally speaking, thinking about terrain and in theory also contour, the changes should be around tile loading strategy, so I think it makes sense to have that as a concept, and it feels like raster and vector have different loading strategies, I'm not sure how terrain fits here as it can be both vector or raster, so I would try and take a moment to think about a design before rushing to implemention.

In any case, you can do it on a different branch and see how it looks, even before all the tests pass, this easy you can just throw it away if it sucks...

I think we did hit that problem in the transform class composition we did, too bad some didn't show me this before that happened...

wayofthefuture: Public Methods Used Outside TileManager
onAdd(map: Map) - Used by the Map class to initialize the tile manager
onRemove(map: Map) - Used by the Map class to clean up when removing
loaded(): boolean - Used by Map and Style classes to check if tiles are loaded
getSource(): Source - Used throughout the codebase to access the underlying source
getState(): SourceFeatureState - Used to access feature state
pause() - Used by Map to pause tile loading
resume() - Used by Map to resume tile loading
serialize() - Used to serialize the source configuration
prepare(context: Context) - Used by Painter to prepare tiles for rendering
getIds(): Array<string> - Used to get all tile IDs
getRenderableIds(symbolLayer?: boolean): Array<string> - Used extensively by rendering pipeline
hasRenderableParent(tileID: OverscaledTileID) - Used by rendering logic
reload(sourceDataChanged?: boolean, tileReloadStrategy?: TileReloadStrategy) - Used to reload tiles
getTile(tileID: OverscaledTileID): Tile - Used by rendering to get specific tiles
getTileByID(id: string): Tile - Used by rendering to get specific tiles by ID
updateCacheSize(transform: IReadonlyTransform) - Used by Map to update cache size
handleWrapJump(lng: number) - Used by Map to handle world wrapping
update(transform: ITransform, terrain?: Terrain) - Used by Map/Style to update tiles for current viewport
tilesIn(pointQueryGeometry, maxPitchScaleFactor, has3DLayer): TileResult[] - Used for querying features at a point
getVisibleCoordinates(symbolLayer?: boolean): Array<OverscaledTileID> - Used by rendering pipeline
setFeatureState(sourceLayer, featureId, state) - Used by Map for feature state management
removeFeatureState(sourceLayer?, featureId?, key?) - Used by Map for feature state management
getFeatureState(sourceLayer, featureId) - Used by Map for feature state management
setDependencies(tileKey, namespace, dependencies) - Used by worker sources
reloadTilesForDependencies(namespaces, keys) - Used to reload tiles when dependencies change
compareTileId(a: OverscaledTileID, b: OverscaledTileID): number - Used for sorting tile IDs
hasTransition(): boolean - Abstract method used by rendering to check for transitions
clearTiles() - Used to remove all tiles
refreshTiles(tileIds: Array<ICanonicalTileID>) - Used to refresh specific tiles

wayofthefuture: After looking at terrain it seems clear that it carries different functionality and it requires two sets of tiles to be maintained. So just a simple file rename may be the solution here, as looking at the content, it's more about rendering: terrain_render_manager.ts or terrain_manager.ts.

HarelM: Interesting. 
But bear in mind that not all public methods need an equivalent in composition. 
I think that it should be similar to the number of abstract methods you defined. 
I'm not sure, I haven't tried it myself, this is more of a hunch...

wayofthefuture: I was able to simplify it by reversing the direction so that TileManager is instantiates (vs Vector/RasterTileManager) elsewheres in the code using a factory function. Give this a look even though it doesn't work yet.

Catches here: setRasterFadeDuration / clearSymbolFadeHold is part of TileManager. Will have to expose some more methods in tileManager such as getTiles() in place of this.tileManager._tiles, etc...

wayofthefuture: And now there's no more inheritance... might want to look at the new files bare without changes...

wayofthefuture: I see how it reads better now

wayofthefuture: Added some method exposure

src/tile/tile_manager.ts
```diff
@@ -84,14 +132,17 @@ export class TileManager extends Evented {
...
     static maxUnderzooming: number;
     static maxOverzooming: number;
 
-    constructor(id: string, options: SourceSpecification | CanvasSourceSpecification, dispatcher: Dispatcher) {
+    /**
+     * Overridable function for extending classes to process the tile when retrieved from cache
+     */
+    constructor(id: string, options: SourceSpecification | CanvasSourceSpecification, dispatcher: Dispatcher, strategy: TileManagerStrategy) {
         super();
+        strategy.setManager(this);
```
HarelM: This is problematic as it creates circular reference between strategy and manager.
I would advise against this pattern.

wayofthefuture: Ok I tried another way... not sure how else to do it.

src/style/style.ts
```diff
@@ -982,7 +982,14 @@ export class Style extends Evented {
         const shouldValidate = builtIns.indexOf(source.type) >= 0;
         if (shouldValidate && this._validate(validateStyle.source, `sources.${id}`, source, null, options)) return;
         if (this.map && this.map._collectResourceTiming) (source as any).collectResourceTiming = true;
-        const tileManager = this.tileManagers[id] = new TileManager(id, source, this.dispatcher);
+
+        if (source.type === 'raster') {
```
wayofthefuture: Pass source.type as parameter and delete lines

src/style/style.ts
```diff
@@ -1326,7 +1333,7 @@ export class Style extends Evented {
             this._updateLayer(layer);
         }
 
-        if (isRasterStyleLayer(layer) && name === 'raster-fade-duration') {
+        if (name === 'raster-fade-duration') {
```
wayofthefuture: delete - not needed

src/tile/vector_tile_manager.ts
```diff
@@ -0,0 +1,164 @@
...
+        const cameraPointQueryGeometry = has3DLayer ?
+            transform.getCameraQueryGeometry(pointQueryGeometry) :
+            pointQueryGeometry;
+
+        const project = (point: Point) => transform.screenPointToMercatorCoordinate(point, this.tileManager.terrain);
+        const queryGeometry = this._transformBbox(pointQueryGeometry, project, !allowWorldCopies);
+        const cameraQueryGeometry = this._transformBbox(cameraPointQueryGeometry, project, !allowWorldCopies);
+        const cameraBounds = Bounds.fromPoints(cameraQueryGeometry);
+
+        const ids = this.tileManager.getIds();
```
wayofthefuture: getSortedIds

wayofthefuture:
@HarelM Exported these versions for a full view of inheritance vs composition, hopefully this helps decide which path to proceed. There are 3 files in each html separate by //======= comments lines.

Inheritance:
[inheritance.ts.html](https://github.com/user-attachments/files/23290765/inheritance.ts.html)

Composition:
[composition.ts.html](https://github.com/user-attachments/files/23290768/composition.ts.html)

HarelM:
Thanks! I'll take a look later today or early tomorrow.

HarelM:
I think both are not great...
Both have the same problem basically - you call a method on tile manager for example, which calls the derived class or the strategy, and this method calls back to the tile manger. This circular handling makes it very hard to follow which part is written where and makes the code harder to read and recon with.
A possible way to break this apart more is understanding what part of the tile manager is needed in every method in the strategy and maybe pass the data instead of the whole tile manager object, for example passing the sourceMaxZoom instead of using that of the tile manager.
One thing that stands out is getting tiles and getting a loaded tile, this is not a simple parameter value, but something more complicated. which might need to break more parts off the tile manager, something like TileStorage, where you can store which tiles are loaded and be able to get-a-loaded-tile. Maybe even pass on the something similar to a tile cache object if this satisfy the storage part.
Narrowing what each class is able to use is the key to breaking stuff apart I believe.
These are my 2 cents.
I think that for the composition to succeed you'd have to remove the `setStrategy` method and see what really the composition needs in order to be able to run the relevant methods which are source specific.
Maybe, but it might be that I'm too optimistic.

HarelM:
@msbarry - we are looking to refactor source cache class (AKA tile manager now).
Any input from your end would be useful as you have looked at the beast (and ran away :-P)

wayofthefuture:
> @msbarry - we are looking to refactor source cache class (AKA tile manager now). Any input from your end would be useful as you have looked at the beast (and ran away :-P)

Me:

image

src/tile/tile_manager.ts
```diff
@@ -84,13 +113,21 @@ export class TileManager extends Evented {
...
+    static maxOverzooming: number = 3;
+    /**
+     * Maximum number of zoom levels to seek ancestor substitute tiles for unloaded ideal tiles.
+     */
+    static maxUnderzooming: number = 10;
 
-    constructor(id: string, options: SourceSpecification | CanvasSourceSpecification, dispatcher: Dispatcher) {
+    /**
+     * Overridable function for extending classes to process the tile when retrieved from cache.
+     */
```
wayofthefuture: Remove comment

src/tile/vector_tile_manager.ts
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }
+
+    /**
+     * Search through our current tiles and attempt to find the tiles that
+     * cover the given bounds.
+     * @param pointQueryGeometry - coordinates of the corners of bounding rectangle
+     * @returns result items have `{tile, minX, maxX, minY, maxY}`, where min/max bounding values are the given bounds transformed in into the coordinate space of this tile.
+     */
+    tilesIn(pointQueryGeometry: Array<Point>, maxPitchScaleFactor: number, has3DLayer: boolean, transform: ITransform | undefined, terrain: Terrain | undefined): TileResult[] {
```
wayofthefuture: remove undefined

wayofthefuture:
@HarelM I think you'll like this latest commit much better...

src/render/draw_fill.test.ts
```diff
@@ -29,6 +29,13 @@ vi.mock('../data/bucket/symbol_bucket', () => {
 });
 vi.mock('../symbol/projection');
 
+function createMockMap() {
```
HarelM: Probably worth moving this to utils, there's already a createMap there, which is a lot more complicated, but I think they should be side-by-side.

src/render/painter.ts
```diff
@@ -145,7 +145,7 @@ export class Painter {
 
         // Within each layer there are multiple distinct z-planes that can be drawn to.
         // This is implemented using the WebGL depth buffer.
-        this.numSublayers = TileManager.maxUnderzooming + TileManager.maxOverzooming + 1;
+        this.numSublayers = TileManager.maxOverzooming + TileManager.maxUnderzooming + 1;
```
HarelM, line 148: There's no real change here, right? just order, which is obviously transitive...

src/tile/raster_tile_strategy.ts
HarelM, line 1: Should this file be renamed?

src/tile/raster_tile_manager.ts
```diff
@@ -0,0 +1,246 @@
+import {type TileManagerStrategy} from './tile_manager';
+import {type TileStore} from './tile_store';
+import {type Tile, FadingDirections, FadingRoles} from './tile';
+import {getEdgeTiles} from '../util/util';
```
HarelM: nit-picking: `getEdgeTiles` sounds like something I would find in the tile_store file?

src/tile/raster_tile_strategy.ts
```diff
@@ -0,0 +1,246 @@
...
+     *
+     * Tiles that are candidates for fading out must be loaded and rendered tiles, as loading a tile to then
+     * fade it out would not appear smoothly. The first source of truth for tile fading always starts at the
+     * ideal tile, which continually changes on map adjustment. The state of the previously rendered ideal
+     * tile plane indicates which direction to fade each part of the newer ideal plane (with varying z).
+     *
+     * For a pitched map, the back of the map can have decreasing zooms while the front can have increasing zooms.
+     * Fade logic must therefore adapt dynamically based on the previously rendered ideal tile set.
+     */
+    _updateFadingTiles(idealTileIDs: OverscaledTileID[], retain: Record<string, OverscaledTileID>, sourceMinZoom: number, sourceMaxZoom: number) {
```
HarelM, line 87: I think we are now allowed to use `private` modifiers...?

src/tile/tile_manager.ts
```diff
@@ -68,9 +99,7 @@ export class TileManager extends Evented {
     _sourceLoaded: boolean;
 
     _sourceErrored: boolean;
-    _tiles: Record<string, Tile>;
```
HarelM, line 71: I think this is great, but we need some feedback from plugin authors as I think this is probably the most used "private" API.
I'll post a slack message, hopefully it will get some traction...

src/tile/tile_manager.ts
```diff
@@ -84,13 +113,21 @@ export class TileManager extends Evented {
...
+    /**
+     * Maximum number of zoom levels to seek ancestor substitute tiles for unloaded ideal tiles.
+     */
+    static maxUnderzooming: number = 10;
 
-    constructor(id: string, options: SourceSpecification | CanvasSourceSpecification, dispatcher: Dispatcher) {
+    /**
+     * Overridable function for extending classes to process the tile when retrieved from cache.
+     */
+    constructor(id: string, options: SourceSpecification | CanvasSourceSpecification, dispatcher: Dispatcher, type: 'vector' | 'raster' | string) {
```
HarelM: do we really need the type parameter? I think we can use what's provided in the options object, can't we?

src/tile/tile_manager.ts
```diff
@@ -107,14 +144,14 @@ export class TileManager extends Evented {
         });
 
         this._source = createSource(id, options, dispatcher, this);
+        this._store = new TileStore(tile => this._unloadTile(tile));
+        this._strategy = type === 'raster'
```
HarelM: [Optional] Consider moving this few lines outside this file including the interface, this will remove all kind of circular references I believe.
In theory, if you pass the factory method in the constructor you can mock the relevant strategy in the tests.

src/tile/tile_manager.ts
```diff
@@ -358,18 +392,17 @@ export class TileManager extends Evented {
                 tile.neighboringTiles[borderId].backfilled = true;
         }
     }
-    /**
-     * Get a specific tile by TileID
-     */
+
     getTile(tileID: OverscaledTileID): Tile {
-        return this.getTileByID(tileID.key);
+        return this._store.getTile(tileID);
```
HarelM: Previous code was using `getTileByID` here, I'm sure it's similar, but just wanted to point this out.

src/tile/tile_manager.ts
```diff
@@ -410,7 +443,7 @@ export class TileManager extends Evented {
             }
 
             // find descendents within the max covering zoom range
-            const maxCoveringZoom = targetID.overscaledZ + TileManager.maxUnderzooming;
+            const maxCoveringZoom = targetID.overscaledZ + TileManager.maxOverzooming;
```
HarelM, line 466: Is this change intentional?

src/tile/tile_manager.ts
```diff
@@ -84,13 +113,21 @@ export class TileManager extends Evented {
...
-    static maxOverzooming: number;
+    /**
+     * Maximum number of zoom levels to seek descendent substitute tiles for unloaded ideal tiles.
+     * Warning: this value increases tile processing logic exponentially with each additional zoom level.
+     */
+    static maxOverzooming: number = 3;
+    /**
+     * Maximum number of zoom levels to seek ancestor substitute tiles for unloaded ideal tiles.
+     */
+    static maxUnderzooming: number = 10;
```
HarelM, line 132: I now understand that this was an intentional change but I read this the opposite way - under zoom mean for me to go up to lower zoom levels while overzoom means go deeper to higher zoom levels.

So I'm not sure this change should be kept...

src/tile/tile_manager.ts
```diff
@@ -490,7 +512,7 @@ export class TileManager extends Evented {
      * are more likely to be found on devices with more memory and on pages where
      * the map is more important.
      */
-    updateCacheSize(transform: IReadonlyTransform) {
+    _updateCacheSize(transform: IReadonlyTransform) {
```
HarelM, line 535: Feels like some parts of this method should be inside the store.
especially the `_maxTileCacheSize` parameter.
So I would consider adding a `setMaxCacheSize(width, height, tileSize)` method instead.
I think...

src/tile/tile_manager.ts
```diff
@@ -1039,136 +833,54 @@ export class TileManager extends Evented {
...
+    }
 
-        if (isRasterType(this._source.type) && this._rasterFadeDuration > 0) {
-            const currentTime = now();
-            for (const id in this._tiles) {
-                const tile = this._tiles[id];
-                if (tile.fadeEndTime >= currentTime) {
-                    return true;
-                }
+    areTilesLoaded(): boolean {
```
HarelM, line 902: I guess this method can be moved to tilesStore and simply get called here, right?

src/tile/tile_manager_state.ts
HarelM, line 1: Amazing!

src/ui/map_tests/map_basic.test.ts
```diff
@@ -137,9 +137,9 @@ describe('Map', () => {
             await map.once('load');
             const fakeTileId = new OverscaledTileID(0, 0, 0, 0, 0);
             map.addSource('geojson', createStyleSource());
-            map.style.tileManagers.geojson._tiles[fakeTileId.key] = new Tile(fakeTileId, undefined);
+            map.style.tileManagers.geojson._store._tiles[fakeTileId.key] = new Tile(fakeTileId, undefined);
```
HarelM: `getStore().addTile(...)`?

HarelM:
WOW! this looks amazing!
I think that breaking the manager to "store" and "strategy" is a good idea.
The only part in the puzzle that I'm still interested to know is how "terrain source cache" fits into all this...
I would like to better understand before we do all this changes to make sure we are on the right direction.

src/render/painter.ts
wayofthefuture: The variables were named backwards. I reversed while maintaining same values.

src/tile/raster_tile_manager.ts
wayofthefuture: Good idea

src/tile/raster_tile_strategy.ts
wayofthefuture: not sure if this messes with tests

src/tile/tile_manager.ts
wayofthefuture: add a getTiles method?

src/tile/tile_manager.ts
wayofthefuture: i'll check!

src/tile/tile_manager.ts
wayofthefuture: Yup

src/tile/tile_manager.ts
wayofthefuture: "under zoom mean for me to go up to lower zoom levels" - yes, that's how i read it. Before, underzoom meant to higher zoom levels and overzoom meant lower zoom levels.

Now, underzoom means to up to lower zoom levels (lower z), overzoom means go down to higher zoom levels (higher z). Right?!?!

src/tile/tile_manager.ts
wayofthefuture: Unfortunately this design is inherited from previous design and I was planning on fixing it when I convert store to use new lru cache but waiting on the merge of the overzoom pr

src/tile/tile_manager.ts
wayofthefuture: i made the exposures match the internal seemed like the right thing to do as we also have getTileById

src/tile/tile_manager.ts
wayofthefuture: It's a viewport animal. I'd really prefer to work on this when I migrate it to new lru cache after you merge overzoom. Also, I'd like to discuss using a fixed or bigger cache size that doesn't care about viewport.

src/tile/tile_manager.ts
wayofthefuture: All of the tile.state string logic is now consolidated into the manager I think keep it here. and thinking about the overall function of manager it feels right here... imo

src/tile/tile_manager_state.ts
wayofthefuture: Yes it feels great!

src/tile/tile_manager.ts
wayofthefuture: nevermind changing back

wayofthefuture:
It's slightly different than a tile manager, as it's doing other things for render purposes.

```
/**
 * @internal
 * This class is a helper for the Terrain-class, it:
 *
 * - loads raster-dem tiles
 * - manages all renderToTexture tiles.
 * - caches previous rendered tiles.
 * - finds all necessary renderToTexture tiles for a OverscaledTileID area
 * - finds the corresponding raster-dem tile for OverscaledTileID
 */
```

It's part of a different architecture in map.ts:

https://github.com/maplibre/maplibre-gl-js/blob/5e525bb049d0a62adb13d58795d475dc99c1878a/src/ui/map.ts#L3391

And terrain.tileManager is not a TileManager instance:

https://github.com/maplibre/maplibre-gl-js/blob/5e525bb049d0a62adb13d58795d475dc99c1878a/src/ui/map.ts#L2167

I renamed to terrain_render_manager. Hopefully that will work for now. I think that a major restructure of this should probably be part of a different PR.

HarelM:
I think the original code for terrain copied source cache code and changed it to fit the terrain needs for different tiles and caching, so I thought it might have similarities that we can combine somehow, but it might not be the case... IDK...

wayofthefuture:
Ok so I had to deal with an incoming merge conflict which redundantly cleared the cache twice, so I cleaned that up and extracted it into tile manager (style.ts -> destroy()). Otherwise I think this is a great start further improving abstraction. Since you merged the tile splitting I can now apply the new LRU cache which will reduce o(n) operations and clean that up as well. Hopefully this commit doesn't break anything!

src/style/style.ts
```diff
@@ -1004,15 +1004,17 @@ export class Style extends Evented {
         const shouldValidate = builtIns.indexOf(source.type) >= 0;
         if (shouldValidate && this._validate(validateStyle.source, `sources.${id}`, source, null, options)) return;
         if (this.map && this.map._collectResourceTiming) (source as any).collectResourceTiming = true;
-        const tileManager = this.tileManagers[id] = new TileManager(id, source, this.dispatcher);
+
+        const tileManager = new TileManager(id, source as SourceSpecification, this.dispatcher);
```
HarelM: Why is a casting needed here?

src/style/style.ts
```diff
@@ -2038,19 +2042,11 @@ export class Style extends Evented {
...
-
-            if (tileManager._tiles) {
-                for (const tileId in tileManager._tiles) {
-                    const tile = tileManager._tiles[tileId];
-                    tile.unloadVectorData();
-                }
-                tileManager._tiles = {};
-            }
-            tileManager._cache.reset();
+            tileManager.clearVectorData();
```
HarelM: This sound specific, can this be part of the onRemove maybe?

src/tile/terrain_render_manager.ts
```diff
@@ -22,7 +22,7 @@ import {type CanonicalTileRange} from '../source/image_source';
  * - finds all necessary renderToTexture tiles for a OverscaledTileID area
  * - finds the corresponding raster-dem tile for OverscaledTileID
  */
-export class TerrainTileManager extends Evented {
+export class TerrainRenderManager extends Evented {
```
HarelM, line 25: Note that this class has tiles, min max zoom, getSource, getTileById and a tile manager on top of that.
I would argue that it's not all about rendering and there are some similarities that we should think how they fit in the new design.
Might be a different PR but definitely need another pass.

src/tile/tile_manager.test.ts
```diff
@@ -81,24 +81,18 @@ function createSource(id: string, sourceOptions: any, _dispatcher: any, eventedP
...
+function createTileManager(options?, used?): TileManager {
+    options = extend({
         tileSize: 512,
         minzoom: 0,
         maxzoom: 14,
         type: 'mock-source-type'
-    }, options), {} as Dispatcher);
+    }, options);
+
+    const sc: TileManager = new TileManager('id', options, {} as Dispatcher);
```
HarelM: Sc is a lefover of source cache?

src/tile/tile_manager.test.ts
```diff
@@ -227,26 +221,26 @@ describe('TileManager.addTile', () => {
...
         const id = tileID.key;
         expect(tileManager._timers[id]).toBeFalsy();
-        expect(tileManager._cache.has(tileID)).toBeFalsy();
+        expect(tileManager.getStore()._cache.has(tileID)).toBeFalsy();
 
         tileManager._addTile(tileID);
 
         expect(tileManager._timers[id]).toBeTruthy();
-        expect(tileManager._cache.has(tileID)).toBeFalsy();
+        expect(tileManager.getStore()._cache.has(tileID)).toBeFalsy();
```
HarelM: Can these calls use a "public" API?

CHANGELOG.md
```diff
@@ -3,12 +3,12 @@
 ### ✨ Features and improvements
 - Slice vector tiles to improve over scale vector handling ([#6521](https://github.com/maplibre/maplibre-gl-js/pull/6521)). It adds the `experimentalZoomLevelsToOverscale` flag to `MapOptions` to allow controlling how many zoom levels to slice and how many to scale. It seems to have better performance at high zoom levels. It can prevent Safar crashes in some scenarios by setting it to 4 or less. (by [@HarelM](https://github.com/HarelM))
 - Add reduceMotion option to Map Options ([#6661](https://github.com/maplibre/maplibre-gl-js/pull/6661))
+- Tile Manager Composition Classes, Vector & Raster Tile Strategies, Tile Reload Strategies ([#6638](https://github.com/maplibre/maplibre-gl-js/pull/6638))
```
HarelM: Please add PR author

src/ui/map.ts
```diff
@@ -2264,13 +2264,10 @@ export class Map extends Camera {
...
      */
     areTilesLoaded(): boolean {
-        const sources = this.style && this.style.tileManagers;
-        for (const id in sources) {
-            const source = sources[id];
-            const tiles = source._tiles;
-            for (const t in tiles) {
-                const tile = tiles[t];
-                if (!(tile.state === 'loaded' || tile.state === 'errored')) return false;
+        const tileManagers = this.style?.tileManagers || [];
```
HarelM: Fallback to empty object instead of array?

HarelM:
I still need to go over a few more files, but this looks promising. 
Good job! 
I'll continue tomorrow
See if you can spot "_" methods used in the tests you changed and try using more public API, like the cache size for example...

src/style/style.ts
wayofthefuture: will remove

src/style/style.ts
wayofthefuture: This was added 2 days ago from another contributor. I can consolidate it, hopefully it won't have any effects

src/tile/terrain_render_manager.ts
wayofthefuture: Yes it's a complicated one.

src/tile/tile_manager.test.ts
wayofthefuture: Will change

src/tile/tile_manager.test.ts
wayofthefuture: If you want to expose it

src/ui/map.ts
wayofthefuture: Yes!

msbarry:
Hello! Just catching up on this and what I missed from #6635 - there are a lot of changes 😅  - I'm trying to understand the high level organization in the new world... I see that tile_manager still maintains its own tile_store and is responsible for adding and removing tiles from it but I can't quite see if/how this will allow tiles to be shared across different usages of that tile (for example terrain/contour/hillshading) like a read-through cache of the source would?

HarelM:
In theory, I think the store should be responsible for knowing if a tiles is still in use or not (kind of like how shared pointer is managed in c++ where the "last person turns off the light").
So if we make the store shared across tile managers and the tile manager would say to the store - I don't need this tile any more, and the store will decide to remove it if it's no in used by other tile manger, it might solve this aspect, wouldn't it?
It's just a rudimentary idea at this point in time, but I think that breaking things into smaller more defined components will allow us to really go forward.
In any case, a class diagram of who's responsible for what and who hold a reference to who would be a good addition to this PR and the docs in general.

msbarry:
Yes, what you describe is exactly what I was thinking... once tile manager is decoupled from tile_store then later we could pass in a shared tile_store.  Or maybe a tile manager has it's own store, they just share a tile_cache instance?  TBH I don't quite understand the difference between tile_store and tile_cache? The tile_store API surface is pretty large too so it still feels a little bit like it's still coupled to tile_manager?

> In any case, a class diagram of who's responsible for what and who hold a reference to who would be a good addition to this PR and the docs in general.

That would be great!

wayofthefuture:
In other words class composition has failed to provide the remedy that it claims. Maybe the proper path is to do the original class inheritance where tile manager kept its own tiles. I still have that code ready.

HarelM:
I don't think there's a problem with the composition here...

wayofthefuture:
Looking at it with a macro lens, I see two types of tile managers. The fact that a diagram is needed for this leads me to believe that maybe composition might not be correct. But if you think composition is the correct path, then tile store should probably be renamed to tile manager state because it is really the current state of the tile manager. Then functions that would be trying to access tiles from tile manager would be calling getState versus getStore.

HarelM:
A diagram is needed to be able to talk about it regardless of composition or inheritance.

wayofthefuture:
Mermaid?

HarelM:
Yeah, there should be built in support in GitHub for it.

wayofthefuture:
This is constantly changing currently it may be a bit premature to create this diagram only to change it when terrain manager is addressed in a future PR.

@msbarry I don't know if what you seek is possible, I am trying to keep this change to primarily a refactor alone, with the exception of when @HarelM has some small mods. I don't want to break anything in this PR. I prefer class extension over composition here, but Harel is the tie-breaker.

@msbarry I think that you are the person for the job of addressing terrain. It is beyond my understanding. It is important to note that terrain keeps a 2nd tile store, completely independent of the tile managers store, and terrain operates in unique spots in the code architecture (separate from style.ts tile managers dictionary). To me it is not a tile manager in the context of the current tile manager, it is a superset of tile manager. There is also terrain.ts which must be considered in the vision. This would not be a refactor, but an actual logic change of which I am not capable, but I believe you are.

wayofthefuture:
@HarelM:

```mermaid

classDiagram
    TileManager --> TileManagerState
    VectorTileStrategy --> TileManagerState
    RasterTileStrategy --> TileManagerState
    TileManagerState --> TileCache

    class VectorTileStrategy {
        -_state: TileManagerState
    }
    
    class RasterTileStrategy {
        -_state: TileManagerState
    }
    
    class TileManager {
        -_state: TileManagerState
        -_strategy: TileManagerStrategy
        -_source: Source
    }

    class TileManagerState {
        -_tiles: Record~string, Tile~
        -_cache: TileCache
    }

    class TileCache {
        -data: Record~string, Tile~
    }

```

HarelM:
Thanks! 
Is it possible to add some of the public APIs of each class? 
I think it will help with the naming, because a state that holds a cache sounds weird in terms of design. 
Also the fact that both state and cache have a dictionary of tiles is odd...
Looking at the big picture helps identifying these sort of quirks...

wayofthefuture:
@HarelM I extracted tile cache from state.

```mermaid
classDiagram
direction LR
    class TileManager {
	    -_state: TileManagerState
	    -_strategy: TileManagerStrategy
	    -_cache: TileCache
	    -_source: Source
	    +onAdd()
	    +onRemove()
	    +loaded()
	    +getSource()
	    +getState()
	    +getCache()
	    +pause()
	    +resume()
	    +serialize()
	    +prepare()
	    +getRenderableIds()
	    +hasRenderableParent()
	    +reload()
	    +getTile()
	    +getTileByID()
	    +getLoadedTile()
	    +update()
	    +refreshTiles()
	    +clearVectorData()
	    +clearTiles()
	    +removeTile()
	    +getVisibleCoordinates()
	    +hasTransition()
	    +areTilesLoaded()
	    +releaseSymbolFadeTiles()
	    +setRasterFadeDuration()
	    +tilesIn()
	    +getSourceFeatureState()
	    +setFeatureState()
	    +removeFeatureState()
	    +getFeatureState()
	    +setDependencies()
	    +reloadTilesForDependencies()
    }

    class TileManagerState {
	    -_tiles: Record~string, Tile~
	    +addTile()
	    +getTileByID()
	    +getTiles()
	    +getTilesSorted()
	    +getTileKeysSorted()
	    +unwrapTiles()
	    +getLoadedTile()
	    +removeTileByID()
	    +getEdgeTiles()
    }

    class TileCache {
	    +max: number
	    +data: Record~string, Tile~
	    +order: Array~string~
	    +reset()
	    +add()
	    +has()
	    +getAndRemove()
	    +get()
	    +getByKey()
	    +remove()
	    +setMaxSize()
	    +filter()
    }

    class VectorTileStrategy {
	    -_state: TileManagerState
	    +onFinishUpdate()
	    +isTileRenderable()
	    +releaseSymbolFadeTiles()
	    +tilesIn()
    }

    class RasterTileStrategy {
	    -_state: TileManagerState
	    +onTileLoaded()
	    +onTileRetrievedFromCache()
	    +onFinishUpdate()
	    +isTileRenderable()
	    +hasTransition()
	    +setRasterFadeDuration()
    }

    TileManager --> TileManagerState
    TileManager --> TileCache
    VectorTileStrategy --> TileManagerState
    RasterTileStrategy --> TileManagerState
```

HarelM:
Thanks for this! This truly helps focusing the discussion (although I think there's an arrow missing from tile manager to the strategies).
If there are public methods for `getState`, `getSource` (`getStrategy`?), and `getCache` you might be able to reduce the public API of tile manager by delegating the relevant calls to the relevant modules.
Another point worth considering is the relation between the name and the public API.
For example the TileManagerState doesn't seem to have method related to "get current state", "get next state", "can move to next state", which are usually expected when talking about state.
Don't get me wrong, naming is always the hardest part :-)

wayofthefuture:
Can you think of a better name?

HarelM:
The methods' names indicate that it is some sort of tile storage, but it's different than the cache, so it's confusing... IDK, I'll probably need to do the same exercise you did in order to get better understand how everything is connected, which I'm trying to avoid 🤪 but I guess I'll need to do it at some point...

wayofthefuture:
Ok I added diagram and detailed explanation. I hope this suffices!

HarelM:
I wonder if there's an easy way to create the diagram straight from the code, or even add it to the tsdocs comments to avoid the code getting out of sync with the diagrams (which always happens eventually).
In any case, I'll do a similar exercise of breaking things apart in the next few days and I'll see where this leads me.
This way we can compare notes and I can better understand the problems you were facing...

HarelM:
Sorry for dragging this, but I want to make the most of this refactoring...

wayofthefuture:
Not that I know of... but it's 100X better than it was

src/tile/vector_tile_strategy.ts
```diff
@@ -0,0 +1,167 @@
...
+            tile?.hasData() &&
+            (symbolLayer || !tile.holdingForSymbolFade())
+        );
+    }
+
+    /**
+     * Release all tiles that are held for symbol fading.
+     * This is useful when forcing an immediate cleanup without waiting for fade completion.
+     */
+    releaseSymbolFadeTiles() {
```
HarelM, line 80: I'm not sure this is equivalent to the previous logic.
The previous logic was calling removeTile from the tile manager, which does other stuff besides actually removing the tile from the dictionary of tiles.
But please check my understanding as well.

wayofthefuture: You are very right.

wayofthefuture:
I don't believe in spaghetti. Closing in favor of something better.

PR #6639, opened 2025-10-31
Generate Unicode character property data at build time

1ec5:
Added a build script that generates lookups of the Unicode character database. To update the lookup functions, run `npm run generate-unicode-data`. It relies a package that represents [a specific version of the Unicode standard](https://www.npmjs.com/package/@unicode/unicode-17.0.0). The Unicode Consortium publishes a new package with every release of the standard. It also relies on [Regenerate](https://github.com/mathiasbynens/regenerate/) to consolidate the raw Unicode character data into compact regular expressions.

The script is much more verbose than the former functions of script_detection.ts, but the generated code is much more compact and performant, as it relies entirely on simple regular expressions. On the other hand, we no longer need to maintain `unicodeBlockLookup` by hand. When a new version of Unicode comes out, we only need to review the lookup functions to make sure they cover all the cases we care about, but we no longer need to update the raw code block codepoint ranges.

This refactor was requested in https://github.com/maplibre/maplibre-gl-js/pull/4550#discussion_r1730056367. There should be no user-facing changes, other than potentially some very minor theoretical changes from upgrading from Unicode 16 to 17.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/developer-guides/unicode.md
+++ b/developer-guides/unicode.md
@@ -1,0 +1,19 @@
+## Unicode
+
+MapLibre GL&nbsp;JS supports rendering Unicode text with the following caveats:
+
+* Codepoints beyond the Basic Multilingual Plane (up to U+FFFF) are not supported.
+* A single codepoint is incorrectly assumed to correspond to a single glyph in any font, and vice versa.
+* Complex text shaping is not yet implemented.
+* Right-to-left writing systems, such as Arabic and Hebrew, require a separate [mapbox-gl-rtl-text](https://github.com/mapbox/mapbox-gl-rtl-text/) plugin.
+
+### Updating compliance with the Unicode Standard
+
+We use various properties from the Unicode Character Database to determine the behavior of a character in a label, such as whether it stays upright in vertical text. When a new major version of the Unicode Standard is released, follow these steps to ensure up-to-date text layout behavior:
+
+1. Note the version of [the most recent Unicode Standard](https://www.unicode.org/versions/enumeratedversions.html).
+2. Find the [Unicode package](https://www.npmjs.com/org/unicode) on NPM that corresponds to the version from step 1. Note that a separate package is published for each version of the standard, and each package has its own versions. Look for the package whose **name** contains the version from step 1, regardless of the version of any package according to NPM.
+3. In package.json, update the `devDependencies` section to refer to `@unicode/unicode-x.y.z`, where _x.y.z_ is the version from step 1.
+4. In build/generate-unicode-data.ts, update the `unicodeVersion` constant to match the version from step 1.
+5. In build/generate-unicode-data.ts, update the `hasUprightVerticalOrientation()` and `hasNeutralVerticalOrientation()` functions to reflect the scripts, blocks, and individual characters enumerated in the latest VerticalOrientation.txt file of the Unicode Character Database. To determine any necessary changes, open `https://www.unicode.org/Public/x.y.z/ucd/VerticalOrientation.txt`, where _x.y.z_ is the previous value of `unicodeVersion` in step 4, and diff that file against [the latest file](https://www.unicode.org/Public/UCD/latest/ucd/VerticalOrientation.txt). If any new codepoint range is listed as `U` or `Tu`, it may need to be added to one of these functions, depending on whether it must be written upright or is more neutral.
+6. Run `npm run generate-unicode-data` and verify that unit tests and render tests all pass.


src/util/script_detection.ts
1ec5, line 1: More of the functions in this file can be converted to the new Regenerate-based approach, but I kept the flow as close as possible to the previous code so that we can more easily spot any inadvertent changes in logic.

package.json
```diff
@@ -165,7 +168,7 @@
     "test-build-ci": "vitest run --config vitest.config.build.ts --coverage",
     "test-watch-roots": "vitest --config vitest.config.unit.ts --watch",
     "test-render": "node --no-warnings --loader ts-node/esm test/integration/render/run_render_tests.ts",
-    "codegen": "run-p --print-label generate-dist-package generate-style-code generate-struct-arrays generate-shaders && npm run generate-typings",
+    "codegen": "run-p --print-label generate-dist-package generate-style-code generate-unicode-data generate-struct-arrays generate-shaders && npm run generate-typings",
```
1ec5, line 171: Weird, this is [failing on CI](https://github.com/maplibre/maplibre-gl-js/actions/runs/18974734980/job/54191141581?pr=6639) because generate-unicode-data.ts has a .ts extension. I can’t reproduce the issue locally. I’d rather not convert the file back to plain JavaScript.

> TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for /home/runner/work/maplibre-gl-js/maplibre-gl-js/build/generate-unicode-data.ts

HarelM: You are missing some flags, check out the other commands...

1ec5: Unfortunately, Regenerate doesn’t support ESM. I couldn’t find another way to import the package.

src/util/script_detection.test.ts
```diff
@@ -1,37 +1,5 @@
...
-        expect(charAllowsIdeographicBreaking('아'.codePointAt(0))).toBe(false);
-    });
-
-    test('allows ideographic breaking of Japanese text', () => {
-        expect(charAllowsIdeographicBreaking('あ'.codePointAt(0))).toBe(true);
-        expect(charAllowsIdeographicBreaking('カ'.codePointAt(0))).toBe(true);
-    });
-
-    test('allows ideographic breaking of Yi text', () => {
-        expect(charAllowsIdeographicBreaking('ꉆ'.codePointAt(0))).toBe(true);
```
1ec5, line 32: I kept some of these tests around while migrating the functions to generate-unicode-data.ts and Regenerate so I could ensure I didn’t flip any logic by accident. But I didn’t keep them around because they aren’t strictly relevant to script_detection.ts.

package.json
1ec5: Never mind, I was dealing with a very cryptic error, but it turned out to be a bug in the script after all.

tsconfig.json
```diff
@@ -33,7 +33,9 @@
       "types": [
         "node"
       ]
-    }
+    },
+    "logError": true,
+    "pretty": true
```
1ec5, line 38: I was facing some cryptic errors running generate-unicode-data that I couldn’t diagnose until I added these settings per https://github.com/TypeStrong/ts-node/issues/2129#issuecomment-3344287930.

build/generate-unicode-data.ts
```diff
@@ -0,0 +1,307 @@
+import * as fs from 'fs';
+import * as regenerate from 'regenerate';
+
+/**
+ * The heuristics in the functions below are based on this version of the
+ * Unicode Standard. When upgrading to a new version of the standard, consider
```
HarelM: Since this is semi-manual there should be a process written somewhere how and when to update this. Maybe in the contributing.md file?

package.json
```diff
@@ -71,6 +71,7 @@
     "@types/window-or-global": "^1.0.6",
     "@typescript-eslint/eslint-plugin": "^8.46.2",
     "@typescript-eslint/parser": "^8.43.0",
+    "@unicode/unicode-17.0.0": "^1.6.14",
```
HarelM, line 74: What will happen when dependabot updates this?

tsconfig.json
HarelM: I must admit, I don't like ts-node. I moved to use `tsx` and it solved a lot of quicks when running typescript files that are "simple" scripts...
It might be out of scope for this PR though.

build/generate-unicode-data.ts
HarelM: If this relays on the version of the unicode package, maybe we can simply remove the version definition below and have dependabot update this automatically, which will generate new code when the package updates.
If this is well covered with tests, there's no problem having it be fully automated.

HarelM:
Add a few question related to process, this looks good otherwise although still feels very verbose in the add/remove range part...

tsconfig.json
1ec5: OSM Americana is using `tsx` too. It’s much better experience. I agree that it would be out of scope for this PR. Just trying to leave things slightly better than I found it. 🙂

build/generate-unicode-data.ts
1ec5: This constant is about the version of the Unicode Standard, which is confusingly distinct from the `unicode` package. Actually, there are lots of `unicode` packages, one per version of the Unicode Standard. This PR would depend on a particular version of the `@unicode-17.0.0` package, which means this constant needs to be `17.0.0`. I guess they publish a new version of this package any time they spot an error in the package itself, as opposed to an error in the standard. When a new version of the standard comes out, we’ll update both this constant and the dependency in package.json. We could keep these two spots in sync by reading package.json, but it would be weird for package.json to run a script that reads package.json!

I left some comments elsewhere in the file to explain the editorial choices, so that someone could reevaluate those choices based on a newer version of the standard, typically by diffing the cited Unicode Character Database file against the previous version. Overall I’d expect these functions to be pretty stable. The Unicode Technical Committee is very averse to changing character properties (as I recently witnessed in their deliberations over my own character proposal). When I diffed the 16.0.0 and 17.0.0 versions of the vertical orientation property file, for example, I didn’t find anything relevant to change in this file, because the Unicode package already took care of updating the character ranges for upright blocks.

package.json
1ec5: Any new version that Dependabot upgrades to will still be based on version 17.0.0 of the Unicode Standard, so you should be able to merge without any additional considerations. Dependabot doesn’t track new versions of the Unicode Standard, which result in brand-new `@unicode/unicode-*` packages at version 1.0.0. By introducing any manual Unicode script detection, Mapbox signed up to revisit the Unicode version at least annually. Fortunately, it isn’t a big deal if we’re one major version behind: that’s still more current than most operating system vendors. Each new major version of the standard does come with many more CJKV characters, but even without this PR, we should already be covering them automatically by looking for the `Han` script.

HarelM: Ahh, I see, I missed the part where the package name has the unicode version... weird...

1ec5:
> still feels very verbose in the add/remove range part...

Definitely. It’s unavoidable for now, but at least the logic is more readable than before, now that more verbose code here doesn’t bloat the package. In the long run, I think we should push as much as we can upstream to the Unicode package. node-unicode/node-unicode-data#94 would greatly simplify the two longest functions in this file. node-unicode/node-unicode-data#57 would also come in handy once we tackle complex text shaping.

build/generate-unicode-data.ts
HarelM: I see, so a process definition of what needs to be done to update all this should be written somewhere.

1ec5: I’ve added a guide to the developer-guides/ folder. I’m sorry there’s not much more detail I can provide about what to include in each function. In addition to node-unicode/node-unicode-data#94, it would be great if the Unicode Character Database or maybe ICU could track the specific behaviors we’re interested in here, namely, whether a given vertically oriented character would make the surrounding text vertical, similar to how some bidirectional characters are “strong” or “weak” when typesetting right-to-left writing systems.

HarelM:
Does this need a changelog entry?

1ec5:
If I’ve done it right, this should be a purely internal build change without any effect on developer or user experience. The upgrade from Unicode 16 to 17 shouldn’t make any difference until #6640 lands, because Unicode is only adding new scripts beyond the BMP at this point. If you think there should be a changelog entry anyways, I can write one up.

HarelM:
Let's hope you are right 😀

PR #6640, opened 2025-10-31
Render characters beyond the Basic Multilingual Plane

### LINKED ISSUE #2307: Labels omit any character outside of the Basic Multilingual Plane

1ec5:
Generalized glyph management and shaping code to operate on UTF-16 characters instead of code units, no longer splitting surrogate pairs. Enabled glyph rendering for additional code blocks beyond the Basic Multilingual Plane (BMP), including the Supplementary Ideographic Plane (SIP) and Tertiary Ideographic Plane (TIP). Both SDF fonts and TinySDF are supported. This improves character coverage of Chinese, Japanese, Korean, and Vietnamese (CJKV); adds support for various minority languages’ writing systems, such as Osage; and introduces support for historical scripts such as Egyptian Hieroglyphics and Cuneiform that are essential for historical maps such as OpenHistoricalMap.

Enabled local glyph rendering in some additional CJKV code blocks in the SIP and TIP, as well as in Tangut, which is thrash-prone like CJKV. Disabled letter spacing for more cursive scripts.

Before | After
:--:|:--:
[嘎味魠魚焿](https://output.jsbin.com/wifikaq/1#21/24.840493/121.00114)<br>[嘎味魠魚焿](https://www.openstreetmap.org/node/10687926045#map=19/24.840493/121.001140) | 嘎味𩵚魠魚焿<br>[嘎味𩵚魠魚焿](https://www.openstreetmap.org/node/10687926045#map=19/24.840493/121.001140)
[南市新化區𦰡拔及羊林里活動中、拔林清水宮](https://output.jsbin.com/wifikaq/1#19/23.0660088/120.3503736)<br>[南市新化區𦰡拔及羊林里活動中](https://www.openstreetmap.org/node/3991710202)<br>[拔林清水宮](https://www.openstreetmap.org/way/219165374) | 臺南市新化區𦰡拔及羊林里活動中、𦰡拔林清水宮<br>[臺南市新化區𦰡拔及羊林里活動中](https://www.openstreetmap.org/node/3991710202)<br>[𦰡拔林清水宮](https://www.openstreetmap.org/way/219165374)
[Pawhuska](https://output.jsbin.com/wifikaq/1#15/36.66284/-96.3395)<br>[Pawhuska](https://www.openstreetmap.org/node/151597427) | Pawhuska 𐓄𐓘𐓢𐓶𐓮𐓤𐓘<br>[Pawhuska 𐓄𐓘𐓢𐓶𐓮𐓤𐓘](https://www.openstreetmap.org/node/151597427)
[](https://output.jsbin.com/wifikaq/1#20/50.871984/20.628967)<br>[&nbsp;](https://www.openstreetmap.org/node/13205822836) | 𝕷𝖆𝖒𝖊𝖓𝖙 𝕿𝖆𝖙𝖙𝖔𝖔 𝕾𝖍𝖔𝖕<br>[𝕷𝖆𝖒𝖊𝖓𝖙 𝕿𝖆𝖙𝖙𝖔𝖔 𝕾𝖍𝖔𝖕](https://www.openstreetmap.org/node/13205822836)
[La naissance d'un champignon](https://output.jsbin.com/wifikaq/1#21/46.800384/6.99962)<br>[_4 La naissance d'un champignon_](https://www.openstreetmap.org/node/10746302560) | 🍄4 La naissance d'un champignon<br>[_🍄4 La naissance d'un champignon_](https://www.openstreetmap.org/node/10746302560)

See https://github.com/maplibre/maplibre-gl-js/pull/4550#issue-2462790709 for more screenshots and links to a live demonstration. Run [this query](https://qlever.dev/osm-planet/pQnXnD) for about 2,000 examples of affected features in OpenStreetMap. [This query](https://qlever.dev/osm-planet/l82Y9Y) finds over 100 examples in Europe alone.

This is a reprise of #4550 that fixes #2307.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/test/build/min.test.ts
+++ b/test/build/min.test.ts
@@ -38,7 +38,7 @@ describe('test min build', () => {
         const decreaseQuota = 4096;
 
         // feel free to update this value after you've checked that it has changed on purpose :-)
-        const expectedBytes = 1012983;
+        const expectedBytes = 1014309;
 
         expect(actualBytes, `Consider changing expectedBytes to: ${actualBytes}`).toBeLessThan(expectedBytes + increaseQuota);
         expect(actualBytes).toBeGreaterThan(expectedBytes - decreaseQuota);


1ec5:
This is a reprise of #4550. I’ve only rebased it atop #4564, but this does mean that it fully fixes #2307 instead of only getting halfway there. Some of the review feedback in #4550 went into #4564 and #6639. There’s still some outstanding feedback about tests that I haven’t gotten to yet. The rebase conflicted with #5043, so we’ll need to pay some attention to vertical alignment.

src/symbol/shaping.ts
```diff
@@ -222,7 +218,8 @@ class TaggedString {
         this.text += section.text;
         this.sections.push(SectionOptions.forText(section.scale, section.fontStack || defaultFontStack, section.verticalAlign));
         const index = this.sections.length - 1;
-        for (let i = 0; i < section.text.length; ++i) {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
HarelM: Can this be replaced with `_char` maybe?

src/symbol/shaping.ts
```diff
@@ -87,7 +87,7 @@ export type TextJustify = 'left' | 'center' | 'right';
 const PUAbegin = 0xE000;
 const PUAend = 0xF8FF;
 
-class SectionOptions {
+export class SectionOptions {
```
HarelM: If these are exported only for tests, I would consider finding a way to test them without exporting them, even if it means a more complicated setup, so that the tests are actually testing the public API and internal methods can change without breaking the tests.

src/symbol/shaping.ts
```diff
@@ -304,18 +300,21 @@ function shapeText(
...
         const untaggedLines =
-            processBidirectionalText(logicalInput.toString(),
-                determineLineBreaks(logicalInput, spacing, maxWidth, glyphMap, imagePositions, layoutTextSize));
+            processBidirectionalText(logicalInput.toString(), lineBreaks);
         for (const line of untaggedLines) {
             const taggedLine = new TaggedString();
             taggedLine.text = line;
             taggedLine.sections = logicalInput.sections;
-            for (let i = 0; i < line.length; i++) {
+            // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
HarelM: Same as the other comment, if possible.

HarelM: Is this only for adding 0's? Can this be done in a regular "for" loop or is this because of the special char handling?

HarelM:
As you mentioned, it might be worth figuring out what else can be done in terms of coverage.
Note that I recently found out that code that runs only in the worker is not considered covered by the coverage report, and I think a lot of the shaping in the render tests are happening in the worker code, so if you think the coverage report is incorrect let me know and we can avoid trying to hack unit tests for cases that are covered with render tests.

1ec5:
Yes, I think almost everything is running on a worker thread. I do think these unit tests are exercising useful aspects of the shaping code that could break in future refactors. Although I could set up some more render tests that exercise these same cases, it’s going to be difficult to diagnose any failures later on, due to additional variables like the system’s available fonts that have nothing to do with the shaping code.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## main
 
 ### ✨ Features and improvements
+- Text labels can now include relatively uncommon Chinese, Japanese, Korean, and Vietnamese characters, as well as characters from historical writing systems. When using server-side fonts, the map may request glyph PBFs beyond U+FFFF from the server instead of throwing an error as before. ([#6640](https://github.com/maplibre/maplibre-gl-js/pull/6640))
```
HarelM: See new format - add "(by ...)" at the end.

src/symbol/shaping.ts
1ec5: The previous use of `line.length` counts characters, whereas we need to iterate over the string in order to get the number of code units. I’ve replaced these for loops with a slightly more cryptic one-liner that avoids the need for an unused local variable.

src/symbol/shaping.ts
1ec5: I split these classes out as separate modules that can be used in unit tests, as suggested in https://github.com/maplibre/maplibre-gl-js/pull/4550#discussion_r1720292300.

1ec5:
I’ve added more unit and integration tests. Most of the uncovered lines that Codecov is flagging appear to be related to bidirectional text, which requires the RTL plugin.

HarelM:
Can you elaborate on the decision to make "internal" methods and structures "public"?
If this is only for tests (which I believe is the motivation) can't this be done using the existing APIs without exposing new ones?

1ec5:
I split `SectionOptions` and `TaggedString` into separate files based on the suggestion in https://github.com/maplibre/maplibre-gl-js/pull/4550#discussion_r1720292300. It made shaping.ts quite a bit more organized, less top-heavy.

These types are only used beyond Shaping in unit tests. I have pushed new integration tests, but they’re indirect enough that we wouldn’t be able to catch logic problems reliably. Is there a way to split types like `SectionOptions` into separate files without exposing them publicly (to consumers of the package, I presume)? The shaping code has always suffered from a lack of unit testability; making it more of a black box would pose a challenge when fixing bugs.

It would be possible to test `determineLineBreaks` in the symbol-shaping integration test. However, line breaking is a rather different piece of functionality than the rest of what shaping.ts does. It’s only in the same module because it has to happen at a particular moment in the shaping process. If I move it there, I’ll probably need to add a comment to the effect of, “If this fails, look at determineLineBreaks in shaping.ts,” because it wouldn’t be obvious otherwise.

HarelM:
If writing unit tests to shaping is too complicated and so you moved classes and functions to different files to test them separately that's a perfectly good solution, assuming that's the last resort, as it does increase the API surface of shaping so that more internal stuff can be used from outside, but to be honest, I don't really think anyone would really use them.

src/symbol/section_options.ts
HarelM, line 1: Since this is mostly a type and two static methods, used in tagged_string I would consider changing this to a type instead of a class, and define a factory method in tagged_string file to create the relevant data.
I think this is aligned with the fact that it's mainly used in tagged_string and also you didn't add any unit test to it (as it's mainly to create a DTO).
Let me know if that sounds right or if this a wrong interpretation of the changes in the code that I'm reading (I'm mostly looking at the diff and not the entire code itself, so keep me honest here).

src/symbol/tagged_string.test.ts
```diff
@@ -0,0 +1,67 @@
...
+            tagged.text = '茹𦨭';
+            expect(tagged.length()).toBe(2);
+        });
+    });
+
+    describe('trim', () => {
+        test('turns a whitespace-only string into the empty string', () => {
+            const tagged = new TaggedString();
+            tagged.text = '  \t    \v ';
+            tagged.sections = [new SectionOptions()];
```
HarelM: Is it possible to use more "public" API for the tests as well - something like addTextSection or something similar? This looks like it relies heavily on what looks to me more like internal fields than actually something that need to be used outside the class.

HarelM: Actually, I see that the same properties are used in the `breakLines` and `shapeText` methods, so I would suggest in that case to allow passing this data in the class' constructor instead of how it is done today, I think it make encapsulation of the field better and also better explain what the class needs when it is initialized.

src/symbol/shaping.ts
```diff
@@ -536,7 +376,7 @@ function leastBadBreaks(lastLineBreak?: Break | null): Array<number> {
     return leastBadBreaks(lastLineBreak.priorBreak).concat(lastLineBreak.index);
 }
 
-function determineLineBreaks(
+export function determineLineBreaks(
```
HarelM: I would consider moving the following line into the tagged string class to encapsulate the text field better:
```
const hasServerSuggestedBreakpoints = logicalInput.text.indexOf('\u200b') >= 0;
```

src/symbol/shaping.ts
```diff
@@ -558,27 +398,43 @@ function determineLineBreaks(
...
+    let i = 0;
+    const chars = logicalInput.text[Symbol.iterator]();
+    let char = chars.next();
+    const nextChars = logicalInput.text[Symbol.iterator]();
+    nextChars.next();
+    let nextChar = nextChars.next();
+    const nextNextChars = logicalInput.text[Symbol.iterator]();
+    nextNextChars.next();
+    nextNextChars.next();
+    let nextNextChar = nextNextChars.next();
```
HarelM: Is this really needed? I haven't seen where it is being used besides moving it to next char...?

src/symbol/shaping.ts
```diff
@@ -558,27 +398,43 @@ function determineLineBreaks(
 
     let currentX = 0;
 
-    for (let i = 0; i < logicalInput.length(); i++) {
+    let i = 0;
+    const chars = logicalInput.text[Symbol.iterator]();
```
HarelM: Should this method be a "stand alone" method and not part of tagged_string? It feels like it touches the text part too much to not be part of the class, especially if we want not to expose `text` field.
Let me know if I'm reading this wrong...

HarelM:
I've added a few last comments after trying to better understand what every class is doing.
I'm guessing this code won't be touched in the near future, so I'm trying to help move the right methods to the right parts.
Sorry if I'm dragging this for too long...

src/symbol/shaping.ts
1ec5: Good idea, that’ll be more object-oriented and mitigate the feedback about exposing too much of the shaping code.

src/symbol/tagged_string.test.ts
1ec5: Yes, that would be more consistent with [the native implementation](https://github.com/maplibre/maplibre-native/blob/955b063ef7dc754c32d8771fa94284f86a37769f/src/mbgl/text/tagged_string.hpp#L75) and make it harder for these two properties to get out of sync. Maybe we should have setters that keep them in sync too.

1ec5:
Thanks, I appreciate the thorough look, since I’m not up to date on TypeScript architectural best practices.

As it happens, this code will change dramatically again as part of retrofitting complex text support (#50). Throughout the shaping code, iterators over codepoints will need to be replaced by iterators over grapheme strings.

src/symbol/shaping.ts
1ec5: Yes, the only way to know whether we’ve reached the last codepoint is to look ahead by one and check `done`. This loop already needs to look ahead by one codepoint anyways, so we need an iterator that looks two ahead to detect the end of the string.

1ec5: (This code will go away completely once we refactor the method to iterate over words using a proper segmenter.)

src/symbol/section_options.ts
1ec5: This is showing up as a diff because I split it into a separate file, but otherwise I haven’t touched `SectionOptions` at all. That said, `SectionOptions` needs a rewrite. It’s actually two different types conjoined in one with mutually exclusive properties. That no longer makes sense since the port to TypeScript.

HarelM: I know you haven't changed it besides moving it, but since it us part of this PR it's worth fixing it I think.

1ec5: I split `SectionOptions` into distinct `TextSectionOptions` and `ImageSectionOptions` classes that both implement the `SectionOptions` interface. I also replaced quite a few duck tests with more formal type checking. As far as I can tell, this meant I had to keep `TextSectionOptions` and `ImageSectionOptions` as classes instead of converting them to types.

test/build/min.test.ts
```diff
@@ -38,7 +38,7 @@ describe('test min build', () => {
         const decreaseQuota = 4096;
 
         // feel free to update this value after you've checked that it has changed on purpose :-)
-        const expectedBytes = 1012983;
+        const expectedBytes = 1014309;
```
1ec5, line 41: This is in response to the following [assertion failure](https://github.com/maplibre/maplibre-gl-js/actions/runs/19211607524/job/54914721160?pr=6640), which started occurring after a rebase atop main as of v5.12.0:

> AssertionError: Consider changing expectedBytes to: 1014309: expected 1014309 to be less than 1014007

src/symbol/shaping.ts
```diff
@@ -414,13 +247,13 @@ function getGlyphAdvance(
     spacing: number,
     layoutTextSize: number
 ): number {
-    if (!section.imageName) {
-        const positions = glyphMap[section.fontStack];
+    if (section instanceof TextSectionOptions) {
```
HarelM: I think there's a better way to do it - by checking if a property exists: i.e.
```ts
if ("fontStack" in section) {
```
This will also avoid the need to cast later on and avoid the need to use classes, interfaces etc, for these types.

src/symbol/shaping.ts
HarelM: Yeah, I missed where `nextNextChar.done` was used.
I've heard "This code will go away" too many time before, so forgive me for being skeptical :-P
But since this is the proper way of doing it, I guess we'll have to live this it for now.

src/symbol/shaping.ts
```diff
@@ -753,20 +605,20 @@ function shapeLines(shaping: Shaping,
...
+                sectionAttributes = shapeImageSection(section as ImageSectionOptions, vertical, lineMaxScale, lineShapingSize, imagePositions);
                 if (!sectionAttributes) continue;
                 imageOffset = Math.max(imageOffset, sectionAttributes.imageOffset);
             }
 
             const {rect, metrics, baselineOffset} = sectionAttributes;
             positionedGlyphs.push({
                 glyph: codePoint,
-                imageName: section.imageName,
+                imageName: (section as ImageSectionOptions)?.imageName ?? null,
```
HarelM: Consider initializing this object with all of the properties that are possible before the above if-else condition and only set the relevant part inside the if.
This will allow removing the casting of `imageName` and `fontStack`.

HarelM:
I've added a few last comments on the casting stuff, and also there's still an open comment about moving the `determineLineBreaks` method.
Other than that, this looks great, thanks!

src/symbol/shaping.ts
1ec5: That’s essentially what the code used to do, but it led to a lot of defensive coalescing like `|| null` and `|| ''`, and the factory methods you pointed out earlier. Ideally these properties would be optional, but I didn’t go that far because it could require refactoring `PositionedGlyph` and the code that touches it. Text and image sections aren’t really the same thing; they just share a couple properties in common and appear in some of the same places, but I’d expect them to diverge over time.

HarelM: You can still (and should) use two types.
And a joint type (which is similar to the interface you declared).
Make all the properties of the type required and you basically have the same code, you just don't need to case or ask for instance of, but instead use these check, which also narrows down the type to what you want to use in the if block etc.

HarelM: i.e.:
```ts
export type SectionOptions  = TextSectionOptions | ImageSectionOptions;

export type TextSectionOptions = {
    scale: number;
    verticalAlign: VerticalAlign;
    fontStack: string;
}

export type ImageSectionOptions = {
    scale: 1; // this ensure it is always equals to 1, but you can put number here and pass 1 when initializing.
    verticalAlign: VerticalAlign; // explicitly pass here 'bottom' when needed.
    imageName: string;
}
```

1ec5: OK, I’ll need to read up on why this is preferable to more explicit type checks, but it does work with some minor adjustments. By the way, this leads to some spots where we’re checking for `imageName` purely to identify it as an image section or `fontStack` to identify it as a text section. I suspect this will break in the future if it becomes possible to create an image section based on something other than an image name, or if we introduce a concept like alt text to the shaping code.

src/symbol/shaping.ts
1ec5: The code in quads.ts and the shaping integration tests all require these fallbacks, not `undefined`. I think it points to a deeper refactor that should be out of scope for this PR. At least we’d need to split `PositionedGlyph` into two subtypes `PositionedTextGlyph` and `PositionedImageGlyph`.

Instead of casting `section`, I ended up having to circumvent any type checks by getting the properties by name instead of by symbol.

src/symbol/shaping.ts
1ec5: Done. This is going to conflict quite a bit with the complex text stuff I’ve got queued up, but it’s for the best.

1ec5:
I’m suddenly getting the following [assertion failure](https://github.com/maplibre/maplibre-gl-js/actions/runs/19218683857/job/54932312855?pr=6640), no idea why:

```
 FAIL   build  test/build/sourcemaps.test.ts > main sourcemap > should match source files
AssertionError: expected 16 to be less than 16
 ❯ test/build/sourcemaps.test.ts:87:27
```

src/symbol/shaping.ts
HarelM: This is why I left the above comment. 
You can initialize an object before the if and before the push and inside the if (where you know the relevant type) assign the relevant property. 
Initialize both properties before that if to null and empty string.
This way you won't need to use properties by name.

HarelM:
Maybe some bad merge conflict resolution? IDK...
BTW, I saw that you moved some methods from one file the other, is it because of the determineLineBreak movement?

1ec5:
> Maybe some bad merge conflict resolution? IDK...

No merge markers. Everything builds, but this test about sourcemaps just started failing.

>  BTW, I saw that you moved some methods from one file the other, is it because of the determineLineBreak movement?

Yes, `determineLineBreaks` has a lot of helper methods that nothing else uses.

src/symbol/shaping.ts
1ec5: Ah, I understand now. Done.

src/symbol/shaping.ts
```diff
@@ -735,17 +337,24 @@ function shapeLines(shaping: Shaping,
...
 
-        for (let i = 0; i < line.length(); i++) {
+        let i = 0;
+        for (const char of line.text) {
             const section = line.getSection(i);
-            const sectionIndex = line.getSectionIndex(i);
-            const codePoint = line.getCharCode(i);
+            const codePoint = char.codePointAt(0);
             const vertical = isLineVertical(writingMode, allowVerticalPlacement, codePoint);
+            const positionedGlyph = {
```
HarelM:
```suggestion
            const positionedGlyph: PositionedGlyph = {
```

src/symbol/shaping.ts
```diff
@@ -735,17 +337,24 @@ function shapeLines(shaping: Shaping,
...
+                glyph: codePoint,
+                x,
+                vertical,
+                sectionIndex: line.getSectionIndex(i),
```
HarelM, line 352:
```suggestion
                sectionIndex: line.getSectionIndex(i),
                imageName: null,
                fontStack: ''
```

src/symbol/shaping.ts
```diff
@@ -735,17 +337,24 @@ function shapeLines(shaping: Shaping,
...
+            } as PositionedGlyph;
```
HarelM:
```suggestion
            };
```

src/symbol/shaping.ts
```diff
@@ -735,17 +337,24 @@ function shapeLines(shaping: Shaping,
...
 
             let sectionAttributes: ShapingSectionAttributes;
-
-            if (!section.imageName) {
+            if ('fontStack' in section) {
                 sectionAttributes = shapeTextSection(section, codePoint, vertical, lineShapingSize, glyphMap, glyphPositions);
                 if (!sectionAttributes) continue;
+                positionedGlyph.imageName = null;
```
HarelM:
```suggestion
```

src/symbol/shaping.ts
```diff
@@ -756,29 +365,26 @@ function shapeLines(shaping: Shaping,
                 sectionAttributes = shapeImageSection(section, vertical, lineMaxScale, lineShapingSize, imagePositions);
                 if (!sectionAttributes) continue;
                 imageOffset = Math.max(imageOffset, sectionAttributes.imageOffset);
+                positionedGlyph.imageName = section.imageName;
+                positionedGlyph.fontStack = '';
```
HarelM:
```suggestion
```

src/symbol/shaping.ts
1ec5: I had to temporarily add placeholder values for the rest of the properties too.

test/build/sourcemaps.test.ts
```diff
@@ -84,6 +84,6 @@ describe('main sourcemap', () => {
         const s1 = setMinus(actualEntriesInSourcemapJSON, expectedEntriesInSourcemapJSON);
         expect(s1.length).toBeLessThan(5);
         const s2 = setMinus(expectedEntriesInSourcemapJSON, actualEntriesInSourcemapJSON);
-        expect(s2.length).toBeLessThan(16);
+        expect(s2.length).toBeLessThan(17);
```
1ec5: Apparently this needs to happen any time you add a .ts file that only exports types. This maximum limit was last increased in #3233 when src/util/actor_messages.ts was added.

HarelM: I think you can move the section_options to be part of the tagged_string file, I thought about commenting on that but I decided it's OK as is.
If this will solve this change I think it's a better fit instead of creating a file that only has definitions.
Either way, it's not that critical, mostly nit-picking.

src/symbol/tagged_string.ts
```diff
@@ -0,0 +1,396 @@
...
+
+function leastBadBreaks(lastLineBreak?: Break | null): Array<number> {
+    if (!lastLineBreak) {
+        return [];
+    }
+    return leastBadBreaks(lastLineBreak.priorBreak).concat(lastLineBreak.index);
+}
+
+export class TaggedString {
+    text: string;
```
HarelM, line 175: Can we make those field `private`?
[Optinal] (because I don't know how I feel about the readability of this suggestion) You might also define them in the constructor maybe?
```ts
constructor(private text: string = '', private sections: Array<SectionOptions> = [], private sectionIndex: Array<number> = [])
```

HarelM:
Two last comments (I hope).
Looks good otherwise.

src/symbol/tagged_string.ts
1ec5: There’s still a couple dozen spots in shaping.ts and tagged_string.test.ts that access these properties. Accessing these properties is safe; we’re only concerned about the properties getting out of sync when setting one of them directly.

test/build/sourcemaps.test.ts
1ec5: I folded the `SectionOptions` types into tagged_string.ts and reverted this change to the expected number of source maps.

src/symbol/tagged_string.ts
HarelM: Fair enough.

HarelM (approved):

etienneJr:
Thanks @1ec5 for this great feature !
Regarding emoji, how can I get them in colors and not in black ?
for example
image
becomes
image
Do I need to use a specific font ?

1ec5:
Unfortunately, GL&nbsp;JS only supports silhouettes of emoji, not color emoji: #6887.

Issue #3084, opened 2023-09-15
Zoom perfomance with layer type symbol


just-website:
I'm trying to draw 2000 points on a type symbol layer and get big performance dips at every zoom step.  The"text-allow-overlap" parameter does not cause visible changes.
Now I can see that the findMatches method takes a very long time to complete at each zoom step.

Are such performance dips to be expected with the symbol type layer?
What can be done to fix it on large layers symbol type than 30k points (is it possible?)?

![image](https://github.com/maplibre/maplibre-gl-js/assets/20595147/7d2a9d98-87b5-4e54-9de2-bd0adfa35023)

https://github.com/maplibre/maplibre-gl-js/blame/18dcac916ccc0b7e7f1160533e5db5c34c28fffe/src/symbol/cross_tile_symbol_index.ts#L102

HarelM:
Can you link to an example that demonstrates the issue? Jsbin?

just-website:
https://codepen.io/Gleb-Shchukin/pen/QWzOLPN?editors=1010
Aactually the problem was solved recently in mapbox. But I see that it is not possible to copy their solution directly (and that's why I’m not even sure if it makes sense to add a link to the PR here).
I reproduced their PR locally and got *10 performance for the findMatch method.

HarelM:
Yes, wet can't copy their code, you shouldn't link to their PR...

just-website:
Do you have any ideas to fix this?

HarelM:
There's a need to run performance analysis and better understand what's causing this.
I have no idea how to fix this.
If this is something that you need solved, I think your best approach would be to try and solve this yourself and submit a PR.
Make sure you don't look at the solution in mapbox so that you don't infringe their copyright.

bradymadden97:
PR'ed a potential improvement here: https://github.com/maplibre/maplibre-gl-js/pull/6641

PR #1755, opened 2022-10-19
Improve performance of cross_tile_symbol_index by using kdbush

mfedderly:
With a large number of symbols visible, maplibre can spend a lot of time in `TileLayerIndex.findMatches`. These requests are called in a pretty tight loop and we can greatly reduce the number of comparisons we need to make by using an index as a first step.

KDBush was already a transitive dependency via supercluster, so hopefully it doesn't impact bundle size much.

To test this I modified the geojson-updates demo to use symbols instead of circles, which triggers the slow path and then set to work making things faster. To my eye the visual output is the same between the old and new methods.

For a speed comparison I used some Chrome profiles of before and after. On my M1 mac I went from 16% idle to 61% idle when using the index.

## Launch Checklist

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.
 - [x] Post benchmark scores.
 - [x] Manually test the debug page.
 - [x] Add an entry to `CHANGELOG.md` under the `## main` section.

PR #6641, opened 2025-10-31
Further optimize cross tile symbol index `findMatches`

bradymadden97:
## Issue
When zooming in or out over a source that has a large number of symbols with _the same_ label (or no label), maplibre will hit an extremely slow path trying to find and pair symbols across tiles (for fading behavior). This will result in noticable stalls and frame drops while zooming. This is also documented in the linked issues below.

## Demo of issue

https://github.com/user-attachments/assets/e6b73ea0-535a-4c40-8353-ce46d4cf2ca7

(this is using https://jsbin.com/girobohihu/1/edit?html,console,output which is the simplest repro I could create. I've seen much worse stalls in production and the linked issues have some examples as well). The TLDR of the issue is:
- Many symbols near eachother
- All with same label or no label at all
- Zoom in/ out quickly
- See stallling

## Background

Essentially, as I understand it, what `findMatches` is doing is iterating through all the symbols on the current tile, and all the symbols on a future tile (at a different zoom level, depending on if we're zooming in or out) and trying to pair as many symbols from the starting tile and the destination tile as possible. When symbols are "paired", the symbol on the destination tile will not "fade in". If the symbols did always "fade in" it gives a weird experience while zooming. The video below mocks what it would look like if we _never_ paired any symbols while zooming:

https://github.com/user-attachments/assets/cfa4c9c4-e1dc-46cf-8e7e-792823ac4b5f

You can see the symbol flash every time we enter a new zoom level, as it's reloaded and unpaired/transitioned from the previous zoom level.

Compare that to the current state:

https://github.com/user-attachments/assets/2fae16b7-e496-4961-bf7f-f80bd36f4e7c

So I think we established the purpose of the code, now let's talk about the pathologically slow case. The slow case happens when we have a bunch of symbols on the origin tile that need to be matched with a bunch of symbols on the destination tile. This happens when all symbols have the same _label_, as the label is what is used to match symbols. However, unlabeled symbols are also essentially treated as being "the same label" for this matching purpose, so a simple pathologically slow case can occur any time we just have a bunch of symbols with no label in the same tile.

We tried to optimize _this exact case_ in https://github.com/maplibre/maplibre-gl-js/pull/1755 by building a spatial index _per_ unique symbol key (aka label, or no label) ([source](https://github.com/bradymadden97/maplibre-gl-js/blob/c834bd2c6102b24d91c32062b8e879599b5bbc54/src/symbol/cross_tile_symbol_index.ts#L65)). But then we ultimately re-run the `range` query - which is the slow path as highlighted in some of the linked issues - for every single symbol instance (of which there are many) [[source](https://github.com/maplibre/maplibre-gl-js/blob/main/src/symbol/cross_tile_symbol_index.ts#L125)]. 

There was a PR to attempt to _exit early_ from these range queries after we've found 1 result (see [here](https://github.com/mourner/kdbush/pull/50)), but I actually don't think this solution would solve this optimally. You wouldn't necessarily want to only fetch **1** result from your range query, as it's possible that first result would have _already_ been paired with a different symbol instance at the same zoom level with the same key. So then the question becomes: "How many _do_ you fetch?", and there is no answer that can be correct besides "All of them", which is what we're currently doing, and is currently slow.

## Changes in this PR

That brings us to the changes in this PR. The idea here is to find a way to _avoid_ running the slow range query `N` times, where `N` is the number of symbol instances with the same symbol key. We want to run the range query for a given symbol key _once_, and then pair as many of the results with as many symbol instances as possible. 

Effectively the pseduo code changes from:

```
// assuming all symbols have the same key

for symbol in symbols:
  let entries = index.range(symbol.location);
  for entry in entries:
    if entry.isNotAlreadyPaired:
      symbol.pair = entry
      break;
```

to 
```
// assuming all symbols have the same key

for location in locationToSymbols:
  let entries = index.range(location);
  loop until entries or location.symbols is exhausted:
    location.symbols[index] = entries[index];
    index++
```

## Before vs. After

Running performance profiles on this example https://jsbin.com/girobohihu/1/edit?html,console,output

Screenshot 2025-10-31 at 1 19 59 PM

Screenshot 2025-10-31 at 1 26 15 PM

The code to run locally (which is where the screenshots themselves came from) is [here](https://github.com/bradymadden97/maplibre-tools/tree/main/repros/slow-cross-tile-symbol-index):
```bash
pnpm run dev:repro slow-cross-tile-symbol-index
```

## Manual Regression Test

I performed a manual regression test with labels to confirm our matching behavior still works with this new algorithm. You can see `52` does not fade in/out which implies it is being matched across tiles:

https://github.com/user-attachments/assets/6d528e14-0520-4774-acf6-bdf83c48ad2b

## Benchmark

Wrote a small benchmark to demonstrate the speedup. There is a third comparison, to an internal fork we have that had some additional optimizations beyond what is currently in main, but this improvement exceeds the improvement on that fork.

This was ran last month, against 5.7, but there have been no changes to this file since then, besides https://github.com/maplibre/maplibre-gl-js/pull/6635 which was just a rename, so they should still be valid.

```typescript

let X = 10 // 100, 1000, 2000, 3000, 10_000, etc.
export class CrossTileSymbolIndexX extends Benchmark {

    bench() {
        const index = new Index();

        const mainID = new OverscaledTileID(6, 0, 6, 8, 8);
        const childID = new OverscaledTileID(7, 0, 7, 16, 16);

        const mainInstances: any[] = [];
        const childInstances: any[] = [];

        for (let i = 0; i < X; i++) {
            mainInstances.push(makeSymbolInstance(0, 0, ''));
            childInstances.push(makeSymbolInstance(0, 0, ''));
        }
        const mainTile = makeTile(mainID, mainInstances);
        const childTile = makeTile(childID, childInstances);
        index.addLayer(styleLayer as any, [mainTile], 0);
        index.addLayer(styleLayer as any, [childTile], 0);
    }
};
```

| <div style="width:80px">Iterations</div> | Maplibre 5.7 | Palantir Internal Fork | Maplibre 5.7 + PR | Speedup |
| -- | -- | -- | -- | -- | 
| 1k | Screenshot 2025-09-23 at 4 06 32 PM | Screenshot 2025-09-23 at 4 52 58 PM | Screenshot 2025-09-23 at 4 25 17 PM | 5.7: 233x<br /><br />Pal: 20x
| 2k | Screenshot 2025-09-23 at 4 17 36 PM | Screenshot 2025-09-23 at 4 54 35 PM | Not run, see 3,000 | See 3,000
| 3k | Screenshot 2025-09-23 at 4 22 41 PM | Screenshot 2025-09-23 at 4 54 51 PM | Screenshot 2025-09-23 at 4 24 17 PM | 5.7: 661x<br /><br />Pal: 55x |
| 10k | Times out / never finishes | Screenshot 2025-09-23 at 4 57 15 PM | Screenshot 2025-09-23 at 4 25 24 PM | 5.7: n/a <br/><br/> Pal: 111x

## Related issues
* https://github.com/maplibre/maplibre-gl-js/issues/6192
* https://github.com/maplibre/maplibre-gl-js/issues/3084

## Launch Checklist

 - [x] Link to related issues.
 - [x] Post benchmark scores.


HarelM:
Is there a chance to write a test to cover this scenario or this is a "proper" refactoring so no changes to logic besides performance optimization?

src/symbol/cross_tile_symbol_index.ts
```diff
@@ -120,25 +122,21 @@ class TileLayerIndex {
...
-                        // don't let any other symbols at the same zoom level duplicate against
-                        // the same parent (see issue #5993)
-                        zoomCrossTileIDs[crossTileID] = true;
-                        symbolInstance.crossTileID = crossTileID;
-                        break;
+                // Build a map keyed by common symbol instance keys, which are also
+                // used to key indexes. For each symbol index key, build a reverse map
+                // of the scaled coordinates back to a set of SymbolInstance that
+                // resolve to that coordinate.
+                const scaledCoordinateId = scaledSymbolCoord.x << 16 | scaledSymbolCoord.y;
```
HarelM: Do we have to use numbers as keys?

src/symbol/cross_tile_symbol_index.ts
```diff
@@ -159,6 +157,54 @@ class TileLayerIndex {
...
+
+        for (const [symbolKey, coordinateMap] of symbolKeyToScaledCoordinatesToSymbolInstanceMap.entries()) {
+            const entry = this._symbolsByKey[symbolKey];
+            if (!entry) {
+                // Shouldn't happen
+                continue;
+            }
+
+            if (!entry.index){
+                // Shouldn't happen
```
HarelM: Why add this if then? Also it seems like this code might be covered by tests, so it is possible that a code flow runs through here?

HarelM:
The explanation in the PR is just amazing! Thanks for all the work on this, it really helps reviewing this change.

src/symbol/cross_tile_symbol_index.ts
bradymadden97: Couldn't figure out another way to use the coordinates themselves as keys in a relatively efficient way

src/symbol/cross_tile_symbol_index.ts
bradymadden97: Updated the comment to attempt to further explain. Let me know if it still isn't clear, though. 

Mostly this is just making typescript happy. I'm not sure what to say if it says this line is hit by test coverage because I'm actually not sure how that can happen.

bradymadden97:
> Is there a chance to write a test to cover this scenario or this is a "proper" refactoring so no changes to logic besides performance optimization?

Found a small bug in the logic that the updates to the test caught. Fixed in the latest version. Test also passes on `main`. I think this update to the test should cover the refactor logic.

HarelM:
Can you add a changelog entry please?

src/symbol/cross_tile_symbol_index.ts
HarelM: There's the tileID.key that is usually used, can it be used here?

src/symbol/cross_tile_symbol_index.ts
```diff
@@ -160,6 +158,52 @@ class TileLayerIndex {
                 }
             }
         }
+
+        for (const [symbolKey, coordinateMap] of symbolKeyToScaledCoordinatesToSymbolInstanceMap.entries()) {
+            const entry = this._symbolsByKey[symbolKey];
+            if (!entry) {
+                // Shouldn't happen, as we only populate keys in this
```
HarelM: Can we avoid these two if altogether? I'm good with the code breaking in case this assumption changes assuming we have the right tests to cover these code paths...

src/symbol/cross_tile_symbol_index.ts
```diff
@@ -160,6 +158,52 @@ class TileLayerIndex {
                 }
             }
         }
+
+        for (const [symbolKey, coordinateMap] of symbolKeyToScaledCoordinatesToSymbolInstanceMap.entries()) {
```
HarelM: Consider moving this entire block to a different method, this method is too long anyway...

HarelM:
I've added a few minor comments. the map of maps is a bit confusing to understand so I would recommend adding some docs and export some part of the method to smaller methods with good documentation.
Other than that, this looks good!

src/symbol/cross_tile_symbol_index.ts
bradymadden97: This coordinate is within a given tile, so not equivalent to tileID.key I don't think.

bradymadden97:
Ok I think this is ready for final re-review.

src/symbol/cross_tile_symbol_index.ts
```diff
@@ -26,18 +26,26 @@ import type {Tile} from '../tile/tile';
 // Round anchor positions to roughly 4 pixel grid
 const roundingFactor = 512 / EXTENT / 2;
 
-export const KDBUSH_THRESHHOLD = 128;
+const SymbolKindType = {
+    INDEXED: 0,
+    UNINDEXED: 1,
+} as const;
+interface IndexedSymbolKind  {
```
HarelM: Is it possible to use `type` instead of `interface`?
I'm trying to make interfaces something you inherit in order to implement and use types when it is related to data passing around.

src/symbol/cross_tile_symbol_index.ts
HarelM: Please don't resolve a comment without a replying or linking to the relevant commit.
This comment was not addressed...

src/symbol/cross_tile_symbol_index.ts
```diff
@@ -119,28 +129,24 @@ class TileLayerIndex {
...
+                    if (existingSymbolsAtCoordinate) {
+                        existingSymbolsAtCoordinate.push(symbolInstance);
+                    } else {
+                        existingSymbolKey.set(scaledCoordinateId, [symbolInstance]);
                     }
+                } else {
+                    symbolKeyToScaledCoordinatesToSymbolInstanceMap.set(symbolInstance.key, new Map([[scaledCoordinateId, [symbolInstance]]]));
                 }
-            } else if (entry.positions) {
+            } else {
```
HarelM, line 131: This else is too nested and it's not clear from the code when it is hit.
Moveover, there's a break inside it which I'm not sure it fully covered with all different flows here.
I would recommend splitting this method into the code that is responsible for non indexed and after that for the indexed.
This way it's easier to read, and also simplifies the code.
I would also start with non-index logic first as I think it's simpler if I read this correctly.

src/symbol/cross_tile_symbol_index.ts
```diff
@@ -160,10 +166,61 @@ class TileLayerIndex {
                 }
             }
         }
+
+        for (const [symbolKey, coordinateMap] of symbolKeyToScaledCoordinatesToSymbolInstanceMap.entries()) {
```
HarelM: Is this for loop relevant for non-indexed symbols? It's hard to understand right now just from reading the code...

src/symbol/cross_tile_symbol_index.ts
```diff
@@ -190,7 +247,7 @@ class CrossTileSymbolLayerIndex {
     };
     lng: number;
 
-    constructor() {
+    constructor(private readonly indexThreshold: number) {
```
HarelM: I'm not a great fan of passing this down, I'm guessing this is mainly for the tests, can we make the test add enough symbols to pass the threshold maybe?

src/symbol/cross_tile_symbol_index.ts
HarelM: Can we use strings maybe? I'm terrified of number overflows when I see this kind of code? How worse is the performance if this is string instead of number?

src/symbol/cross_tile_symbol_index.test.ts
```diff
@@ -30,10 +30,12 @@ const makeTile = (tileID, symbolInstances): any => {
     };
 };
 
+const INDEX_THRESHOLDS = [1, 128];
+
 describe('CrossTileSymbolIndex.addLayer', () => {
 
-    test('matches ids', () => {
-        const index = new CrossTileSymbolIndex();
+    test.each(INDEX_THRESHOLDS)('matches ids: indexThreshold %s', (indexThreshold) => {
```
HarelM: I'm not sure I fully follow why every test need to be tested with two thresholds, is this to verify the same logic is kept in both index and non indexed flows?
I would recommend removing it and focusing on the relevant logic that is being tested, either the index or non indexed and simply write the relevant tests, it might be more verbose, but I think it's better in unit tests.

HarelM:
Thanks for all the time spent on this!
Sorry for all the annoying comments, but I need to make sure I fully understand this code before pushing it through as I'll probably be the one that needs to maintain it...

src/symbol/cross_tile_symbol_index.ts
bradymadden97: Yeah was mainly for tests, but based on 

> focusing on the relevant logic that is being tested, either the index or non indexed and simply write the relevant tests, it might be more verbose, but I think it's better in unit tests.

I will undo this and instead just write some indexed tests to go along with the non-indexed tests :)

bradymadden97:
Ok I refactored as suggested to try and pull the indexed and non-indexed codepaths apart from each other. Let me know if it's a bit clearer what's going on now.

Will follow up with some tests specifically for indexed path tomorrow as suggested.

src/symbol/cross_tile_symbol_index.ts
```diff
@@ -99,71 +106,136 @@ class TileLayerIndex {
...
+    /**
+     * Find matches for entries with an index built
+     */
+    private findMatchesForIndexedEntry(
+        entry: IndexedSymbolKind,
+        symbolInstances: SymbolInstance[],
+        newTileID: OverscaledTileID,
+        zoomCrossTileIDs: {[crossTileID: number]: boolean}
+    ) {
+        const tolerance = this.tileID.canonical.z < newTileID.canonical.z ? 1 : Math.pow(2, this.tileID.canonical.z - newTileID.canonical.z);
```
HarelM: I think this can be moved outside both methods and passed (it is the same calculation, right?), maybe this will allow removing other inputs of this method in return, not sure...

src/symbol/cross_tile_symbol_index.ts
```diff
@@ -99,71 +106,136 @@ class TileLayerIndex {
...
+
+        // Map instances by their scaled coordinate
+        const instancesByScaledCoordinate = new  Map<number, SymbolInstance[]>();
```
HarelM:
```suggestion
        const instancesByScaledCoordinate = new Map<number, SymbolInstance[]>();
```

src/symbol/cross_tile_symbol_index.ts
```diff
@@ -99,71 +106,136 @@ class TileLayerIndex {
...
             }
         }
     }
 
-    getCrossTileIDsLists() {
-        return Object.values(this._symbolsByKey).map(({crossTileIDs}) => crossTileIDs);
+    /**
+     * Find matches for entries with an index built
+     */
+    private findMatchesForIndexedEntry(
```
HarelM: I'm having a hard time understanding what's the output of this method is.
Is it possible to change that to return something, or mark the relevant variable as output somehow? updating the TSdoc of the method is also expected.

bradymadden97: These methods don't return anything they mutate `symbolInstance` and `zoomCrossTileIDs` in place and return nothing. I added an explicit return type and updated the method names to hopefully make that more clear.

src/symbol/cross_tile_symbol_index.ts
bradymadden97: It won't allow to remove any other parameters but I did pull it up, you're right that it would save re-calculating it a few times.

src/symbol/cross_tile_symbol_index.ts
bradymadden97: Changed to doubly nested map by x/y coordinates if that's better https://github.com/maplibre/maplibre-gl-js/pull/6641/commits/964916cbdc502fe687bf3a0ffa75a4188ac8c810

bradymadden97:
Added a test suite that mimics the non-indexed code paths but should hit the indexed code paths.

CHANGELOG.md
```diff
@@ -11,7 +12,7 @@
 
 ### ✨ Features and improvements
 
-- Add support for MapLibre Tiles (MLT) by using `encoding: 'mlt'` in vector source definition ([#6570](https://github.com/maplibre/maplibre-gl-js/pull/6570)) (by [@Salkin975](https://github.com/Salkin975) and [@HarelM](https://github.com/HArelM)) 
+- Add support for MapLibre Tiles (MLT) by using `encoding: 'mlt'` in vector source definition ([#6570](https://github.com/maplibre/maplibre-gl-js/pull/6570)) (by [@Salkin975](https://github.com/Salkin975) and [@HarelM](https://github.com/HArelM))
```
bradymadden97:
```suggestion
- Add support for MapLibre Tiles (MLT) by using `encoding: 'mlt'` in vector source definition ([#6570](https://github.com/maplibre/maplibre-gl-js/pull/6570)) (by [@Salkin975](https://github.com/Salkin975) and [@HarelM](https://github.com/HArelM)) 
```

src/symbol/cross_tile_symbol_index.test.ts
```diff
@@ -214,6 +214,213 @@ describe('CrossTileSymbolIndex.addLayer', () => {
...
+
+        const mainID = new OverscaledTileID(6, 0, 6, 8, 8);
+        const mainInstances = Array.from({length: INSTANCE_COUNT}, () => makeSymbolInstance(1000, 1000, 'Detroit'));
+        mainInstances.push(makeSymbolInstance(2000, 2000, 'Toronto'));
+        const mainTile = makeTile(mainID, mainInstances);
+
+        index.addLayer(styleLayer, [mainTile], 0);
+        // Assigned new IDs
+        for (let i = 1; i <= INSTANCE_COUNT + 1; i++) {
+            expect(mainInstances.find(j => j.crossTileID === i)).not.toBeUndefined();
```
HarelM: `toBeDefined()`?

src/symbol/cross_tile_symbol_index.test.ts
```diff
@@ -214,6 +214,213 @@ describe('CrossTileSymbolIndex.addLayer', () => {
...
+        index.addLayer(styleLayer, [mainTile, childTile], 0);
+        // matched parent tile for all Detroit
+        const detroitChildren = childInstances.filter(i => i.key === 'Detroit');
+        for (let i = 1; i <= INSTANCE_COUNT; i++) {
+            expect(detroitChildren.find(j => j.crossTileID === i)).not.toBeUndefined();
+        }
+
+        // does not match Windsor because of different key
+        const windsorInstance = childInstances.find(i => i.key === 'Windsor');
+        expect(windsorInstance.crossTileID).toEqual(131);
```
HarelM, line 251: This test has too many `expect` calls, I would recommend splitting this test and use `beforeEach` to arrange all the relevant information if it is similar between the tests.
The comments above each paragraph is a good indication that this needs to be split and the comment should probably be the test name.

src/symbol/cross_tile_symbol_index.ts
```diff
@@ -99,72 +106,154 @@ class TileLayerIndex {
...
+    getCrossTileIDsLists() {
+        return Object.values(this._symbolsByKey).map(({crossTileIDs}) => crossTileIDs);
+    }
+
     findMatches(symbolInstances: SymbolInstanceArray, newTileID: OverscaledTileID, zoomCrossTileIDs: {
         [crossTileID: number]: boolean;
     }) {
-        const tolerance = this.tileID.canonical.z < newTileID.canonical.z ? 1 : Math.pow(2, this.tileID.canonical.z - newTileID.canonical.z);
-
+        // Group symbol instances by key
```
HarelM: This comment sounds like a great method name, can the grouping part be moved to a separate method?

src/symbol/cross_tile_symbol_index.ts
```diff
@@ -99,72 +106,154 @@ class TileLayerIndex {
...
-                        // Once we've marked ourselves duplicate against this parent symbol,
-                        // don't let any other symbols at the same zoom level duplicate against
-                        // the same parent (see issue #5993)
-                        zoomCrossTileIDs[crossTileID] = true;
-                        symbolInstance.crossTileID = crossTileID;
-                        break;
-                    }
+            const xMap = instancesByScaledCoordinate.get(scaledSymbolCoord.x);
+            if (xMap) {
+                const yMap = xMap.get(scaledSymbolCoord.y);
```
HarelM: is `yMap` a map or an array? I don't see method related to `Map` when using `yMap`, am I missing anything?

src/symbol/cross_tile_symbol_index.ts
```diff
@@ -99,72 +106,154 @@ class TileLayerIndex {
...
+                    new Map([[scaledSymbolCoord.y, [symbolInstance]]])
+                );
+            }
+        }
+
+        // For each scaled coordinate, match instances with indexed results
+        // at the same location.
+        for (const [x, yMap] of instancesByScaledCoordinate.entries()) {
+            for (const [y, instances] of yMap.entries()) {
+                const indexes = entry.index.range(
```
HarelM, line 237: I tried to compare this to the previous code and I'm not sure I understand why there will be less calls to `range`, can you elaborate on this?

HarelM:
Sorry for dragging this PR, I just really want to understand what you did here...

src/symbol/cross_tile_symbol_index.test.ts
bradymadden97: Was copying the structure of the section of tests above, which do have interleaved expects and do not use beforeEach. I would suggest the structures remain the same actually, in my opinion. Do you agree / would you prefer me to include `beforeEach` refactor for the existing tests in the other decribe block as well?

HarelM: I think it would be better to make the file similar to other files and also make sure it is consistent within the same file, meaning change to use beforeEach etc.
Having said that, I don't want to force you to rewrite the entire file, so at least change the tests you added, it would be great if you would change the other tests as well, but I would understand if you decide not to.

src/symbol/cross_tile_symbol_index.ts
bradymadden97: Absolutely! So the situation where this is faster is because we're now running a `range` query only _once_ for each coordinate position that has any symbol instances within it. Then we are matching all symbol instances at that position with any results from the range query.


Previously, we would independently run a range query for every single symbol instance, and then match a result from that range query with the symbol instance.

A pathological case would be 10,000 symbol instances at the exact same coordinate. Previously that would be 10,000 range queries. Now that would be 1 range query.

Based on that algorithm, it appears to never be _more_ range queries than before, but in bad cases it can be significantly less.

src/symbol/cross_tile_symbol_index.ts
bradymadden97: Does this rename improve it or not? https://github.com/maplibre/maplibre-gl-js/pull/6641/commits/5c2c7937d5925bba3edc22b0dd3058905fcb693f

It's a doubly nested map (since I need to group instances by their x,y coordinates). Happy to change to other suggestions, or clarify if the reasoning still doesn't make sense

bradymadden97:
> Sorry for dragging this PR, I just really want to understand what you did here...

Not a problem at all, I similarly want to make the code understandable so thank you for the feedback. Happy to keep iterating until it's in a good place.

src/symbol/cross_tile_symbol_index.ts
HarelM: But this is for x and y that are exactly the same, right? If all the x and y are in slightly different places then it would behave as before, right?
I'm just trying to make sure I understand what you mean and where this will improve performance.

src/symbol/cross_tile_symbol_index.ts
bradymadden97: Pulled out in https://github.com/maplibre/maplibre-gl-js/pull/6641/commits/f729a857f9062fdbf3483bbf83af70ef8f205726

src/symbol/cross_tile_symbol_index.ts
HarelM: instancesByScaledCoordinate is a map, xMap is a map, but yMap is an array, am I reading it wrong?

package-lock.json
```diff
@@ -469,8 +469,7 @@
       "resolved": "https://registry.npmjs.org/@cspell/dict-css/-/dict-css-4.0.18.tgz",
       "integrity": "sha512-EF77RqROHL+4LhMGW5NTeKqfUd/e4OOv6EDFQ/UQQiFyWuqkEKyEz0NDILxOFxWUEVdjT2GQ2cC7t12B6pESwg==",
       "dev": true,
-      "license": "MIT",
-      "peer": true
+      "license": "MIT"
```
bradymadden97, line 534: Not sure what's going on with these generated changes.

src/symbol/cross_tile_symbol_index.test.ts
bradymadden97: Resolved in https://github.com/maplibre/maplibre-gl-js/pull/6641/commits/3d218f37d258077ad3ac9a6491271f980c61fe4c

src/symbol/cross_tile_symbol_index.ts
HarelM: I don't think https://github.com/maplibre/maplibre-gl-js/commit/5c2c7937d5925bba3edc22b0dd3058905fcb693f make it more readable unfortunately.
maybe just rename yMap to `instancesArray`? or simply `instances`?
Man this is so confusing...
I think the names should match the below for loops, but I'm not sure they do...

package-lock.json
HarelM: I would advise to simply revet this file...

HarelM: Probably different version of node...

src/symbol/cross_tile_symbol_index.ts
melitele: `interface` in TypeScript provides a syntactical contract while `type` is a versatile way of defining structure of objects. In this case `type` is more appropriate.
However using interfaces solely to specify what class is supposed to implement is too narrow. It can and should also be used to specify parameters and return values of exported functions to emphasize the contract aspect.

src/symbol/cross_tile_symbol_index.ts
bradymadden97: Yes, though these are x,y values output from `getScaledCoordinates`, so they are scaled and down-sampled so there is a higher liklihood of overlaps vs. the original coordinates, as I understand it.

src/symbol/cross_tile_symbol_index.test.ts
bradymadden97: After re-looking at it, most of these tests require a good bit of set up to get into the state we ultimately are testing. I'm not sure it makes sense / is easily possible to split them up, and I don't see a lot of shared setup across tests that can be pulled into a `beforeEach`. 

If you wouldn't mind could you take another look at a few of these and see if you think otherwise?

src/symbol/cross_tile_symbol_index.ts
```diff
@@ -99,72 +106,163 @@ class TileLayerIndex {
...
 
+    getCrossTileIDsLists() {
+        return Object.values(this._symbolsByKey).map(({crossTileIDs}) => crossTileIDs);
+    }
+
     findMatches(symbolInstances: SymbolInstanceArray, newTileID: OverscaledTileID, zoomCrossTileIDs: {
         [crossTileID: number]: boolean;
     }) {
         const tolerance = this.tileID.canonical.z < newTileID.canonical.z ? 1 : Math.pow(2, this.tileID.canonical.z - newTileID.canonical.z);
+        const instancesByKey = this.groupSymbolInstancesByKey(symbolInstances);
```
melitele, line 117: If the map returned by `groupSymbolInstancesByKey` doesn't change for the same `symbolInstances`, the `groupSymbolInstancesByKey` can moved out of `findMatches`

melitele:
The need for optimization stems from the fact that symbols don't have IDs unique across tiles. If they did, then cross tile symbol index would be a simple map. There is a mechanism that assigns IDs to features at least for [GeoJSON source](https://maplibre.org/maplibre-style-spec/sources/#generateid) and symbols are unique per feature/layer combo.

src/symbol/cross_tile_symbol_index.test.ts
HarelM: This test specific has a lot of steps and logic.
I would recommend move the initial 20-30 lines into a `beforeEach` and then test only the part according to each test.
This would mean that you would split this test every time you wrote a comment like:
`matched Detroit children tiles from parent` for example.

This would make the definition of the test more readable and avoid the need to have out of sync comments inside the test.

If this is not possible to split this test up, then a re-write of this test in needed as it just has too many logic and tests "too much".

I hope this is clearer now.

src/symbol/cross_tile_symbol_index.test.ts
```diff
@@ -214,6 +214,213 @@ describe('CrossTileSymbolIndex.addLayer', () => {
...
+        index.addLayer(styleLayer, [childTile], 0);
+        expect(Math.min(...childInstances.map(i => i.crossTileID))).toBe(INSTANCE_COUNT + 1);
+        expect(Math.max(...childInstances.map(i => i.crossTileID))).toBe(2 * INSTANCE_COUNT);
+
+        // Expect all to have a crossTileID
+        expect(mainInstances.some(i => i.crossTileID === 0)).toBeFalsy();
+        expect(childInstances.some(i => i.crossTileID === 0)).toBeFalsy();
+
+        // Overwrites the old id to match the already-added tile
+        index.addLayer(styleLayer, [mainTile, childTile], 0);
```
HarelM, line 318: If this `addLayer` can happen earlier it would be great in order to keep the test in arrange-act-assert (AAA) style.
If this is not possible, consider splitting this test into two tests.

src/symbol/cross_tile_symbol_index.test.ts
```diff
@@ -214,6 +214,213 @@ describe('CrossTileSymbolIndex.addLayer', () => {
...
+        expect(Math.max(...mainInstances.map(i => i.crossTileID))).toBe(INSTANCE_COUNT);
+
+        const layerIndex = index.layerIndexes[styleLayer.id];
+        expect(Object.keys(layerIndex.usedCrossTileIDs[6]).length).toEqual(INSTANCE_COUNT);
+        for (let i = 1; i <= INSTANCE_COUNT; i++) {
+            expect(layerIndex.usedCrossTileIDs[6][String(i)]).toBeDefined();
+        }
+
+        // copies parent ids without duplicate ids in this tile
+        index.addLayer(styleLayer, [childTile], 0);
```
HarelM, line 350: Same as previous comment.

melitele:
> The need for optimization stems from the fact that symbols don't have IDs unique across tiles. If they did, then cross tile symbol index would be a simple map. There is a mechanism that assigns IDs to features at least for [GeoJSON source](https://maplibre.org/maplibre-style-spec/sources/#generateid) and symbols are unique per feature/layer combo.

[Here's the commit](https://github.com/melitele/maplibre-gl-js/commit/5f27b5438878864fc4a043d09006270f57738be8) that makes up cross tile ids for symbols from feature and layer ids. For features without ids it'll fall back to the code this PR is optimizing. For GeoJSON sources with features without `id` field, one can set either [`generateId`](https://maplibre.org/maplibre-style-spec/sources/#generateid) or [`promoteId`](https://maplibre.org/maplibre-style-spec/sources/#promoteid_1). Features in vector tiles generated from OSM tend to have ids. If not, cross tile ids can be generated for feature in vector tiles from geo coordinates and label (I haven't done this). 

I modified the bin from this PR to use the cross tile ids from the commit above and `label` as feature id via `generateId`:  https://jsbin.com/nemadaj/1/edit?html,output

In my ad hoc testing it doesn't stall and I am curious to see the level of actual improvement if @bradymadden97 can run the benchmark with the change from the commit above.

bradymadden97:
Tried to get the benchmarks above running again so I could test out the above suggestion, but running into errors even trying to _build_ the benchmarks on main.

```
[!] (plugin typescript) RollupError: [plugin typescript] @rollup/plugin-typescript: Path of Typescript compiler option 'outDir' must be located inside Rollup 'dir' option.
```

Perhaps benchmarks has been broken?

HarelM:
Are you using a different version of the libraries somehow? I think this error message if from a recent breaking change of the roll-up plugin. 
Benchmarks are running on main every commit I believe: 
https://github.com/maplibre/maplibre-gl-js/actions/workflows/upload-benchmarks.yml

HarelM:
Any updates on this? What do we need in order to push this forward?

NeluQi:
Any news on PR?

HarelM:
Seems a bit abandoned... Do you want to copy the code over to a different PR and continue the work @NeluQi?

bradymadden97:
New implementation using @melitele's great suggestion here: https://github.com/maplibre/maplibre-gl-js/pull/7665

PR #6644, opened 2025-11-01
Export now() function in timeControl API

bjperson:
- Adds now() to public API exports alongside setNow, restoreNow, isTimeFrozen
- Fixes #6643
- Enables external code to read controlled time consistently with MapLibre's internal rendering
- This is a non-breaking change that completes the timeControl API

## Launch Checklist

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.
 - [x] Link to related issues.
 - [x] Document any changes to public APIs.
 - [x] Add an entry to `CHANGELOG.md` under the `## main` section.

PR #6648, opened 2025-11-02
Add timeControl API interactive example

bjperson:
Adds a new example demonstrating the timeControl API (`now()`, `setNow()`, `restoreNow()`, `isTimeFrozen()`)  with an interactive animation that users can freeze and jump to specific timestamps.

## Features:
- Continuous Paris ↔ New York animation loop
- Jump to specific moments (1s, 2s, 10s) with time frozen
- Resume to restart the animation with real time
- Live display of elapsed time and frozen state (`isTimeFrozen()`)

## Demo
The example shows users how to:
- Control animation timing deterministically
- Freeze and resume time during animations
- Jump to specific timestamps in an animation

## Note
Requires `maplibregl.now()` from #6644 (already merged to main, will be available in next release).

## Related
- #6644 - PR that exported `now()` function
- Discussion: https://github.com/maplibre/maplibre-gl-js/pull/6544#issuecomment-2550359146

## Launch Checklist

 - [x] Link to related issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### 🐞 Bug fixes
 - If a required glyph PBF is unavailable or it lacks a glyph for a character in a `text-field`, try to render it locally instead of crashing. ([#4564](https://github.com/maplibre/maplibre-gl-js/pull/4564))
 - Export `now()` function in timeControl API to complete the API and enable external code to read controlled time ([#6644](https://github.com/maplibre/maplibre-gl-js/pull/6644))
+- ScaleControl CSS styling contains `white-space: nowrap` to prevent wrapping ([#6647](https://github.com/maplibre/maplibre-gl-js/pull/6647))
 - _...Add new stuff here..._
 
 ## 5.10.0
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -79,7 +79,6 @@ export default [
             'no-unused-vars': 'off',
             'no-warning-comments': 'error',
             'object-curly-spacing': ['error', 'never'],
-            'prefer-arrow-callback': 'error',
 
             'prefer-const': ['error', {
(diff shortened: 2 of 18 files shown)


HarelM:
Thanks for taking the time to write this example!
Any change you can reduce the amount of code in it? It's like 250 lines now, which is a bit too much to get the gist of what you want to show here, I think.

Regarding merging, there's no need to be alarmed, the examples and other docs are published only when a new version is released.

No need to add it to the changelog, in theory you could've added the example as part of the time API feature, so it's basically there in terms of changelog.

1ec5:
By the way, I think GIS people may stumble upon this example and be confused because it’s about animation timing rather than animating temporal data. This isn’t your fault at all, it’s an inevitable naming collision. Maybe I could separately create an example of using [maplibre-gl-dates](https://github.com/OpenHistoricalMap/maplibre-gl-dates/) to divert people searching for that distinct use case, so that people landing on this example will have a clearer idea of what it’s about?

bjperson:
@1ec5 You're right about the potential confusion with temporal data.

After reflecting on this, I think the TimeControl API is actually too specialized for a general example:
- It's primarily for video recording libraries (maplibre-gl-video-export)
- Used in automated testing scenarios
- The JSDoc is already ok with examples

The API reference docs are sufficient. I don't think we need this example in the main docs.

@HarelM What do you think about just skipping the example entirely?

HarelM:
I'm OK with skipping the example.

bjperson:
Closing per maintainer feedback - JSDoc documentation is sufficient for this API.

1ec5:
> It's primarily for video recording libraries (maplibre-gl-video-export)

Looking forward to [maplibre-gl-video-export](https://github.com/bjperson/maplibre-gl-video-export/)!

PR #6659, opened 2025-11-03
Allow global state expressions in layer visibility

### LINKED ISSUE #6495: Support global state expression for layer visibility
## User Story

I would like to create a layer that is "togglable" to be visible/invisible with a `global-state` expression.

For example:
```js
{
  id: 'historical-stations',
  layout: {
    visible: ['global-state', 'feature-enabled'],
  }
}
```

The `visibility` attribute of the layout properties of a layer is handled specifically, and does not seem to support any expressions. 

See https://maplibre.org/maplibre-style-spec/layers/#visibility and https://github.com/maplibre/maplibre-gl-js/blob/5e0f8f749eab9bb5e0ea7e0a34a14112b52f0a66/src/style/style_layer.ts#L90 and https://github.com/maplibre/maplibre-gl-js/blob/5e0f8f749eab9bb5e0ea7e0a34a14112b52f0a66/src/style/style_layer.ts#L230-L233

It does not make sense for the `visibility` to support expressions related to the properties of a feature, but it does makes sense to support expressions related to `global-state`.

## Rationale

Currently, making layers dynamically visible requires dynamically adding or deleting layers, or alternatively attaching metadata to every layer and dynamically reading the metadata to toggle the layer visibility.

The MapLibre style specification is made to be statically defined, instead of dynamic. After the addition of the global map state through the proposal in https://github.com/maplibre/maplibre-style-spec/issues/886 (https://maplibre.org/maplibre-style-spec/expressions/#global-state), we can now define map features with defaults. Such features can be used to make the map dynamic using a static style, in most properties except for `visibility`.

## Impact

Make it easier for map developers to create MapLibre-powered maps that behave more dynamically based on the user preferences. Such styles will also be more reusable because the map style itself documents the global state parameters, instead of depending on client-side scripting to make manipulate the map layers.

melitele:
- Resolves #6495

- Depends on PR https://github.com/maplibre/maplibre-style-spec/pull/1359

This PR adds support for global state visibility expressions.

 - [x] Link to related issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## main
 
 ### ✨ Features and improvements
+- Support global state expression for layer visibility ([#6495](https://github.com/maplibre/maplibre-gl-js/pull/6495)) (by [@melitele](https://github.com/melitele))
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
@@ -752,7 +753,7 @@
 
 ### ✨ Features and improvements
 
-- ⚠️ Change attribution to be on by default, change `MapOptions.attributionControl` to be the type that the control handles, removed `MapOptions.customAttribution` ([#3618](https://github.com/maplibre/maplibre-gl-js/pull/3618)). Note: showing the logo of MapLibre is not required for using MapLibre. (by [@HarelM](https://github.com/HarelM)) 
+- ⚠️ Change attribution to be on by default, change `MapOptions.attributionControl` to be the type that the control handles, removed `MapOptions.customAttribution` ([#3618](https://github.com/maplibre/maplibre-gl-js/pull/3618)). Note: showing the logo of MapLibre is not required for using MapLibre. (by [@HarelM](https://github.com/HarelM))
 - ⚠️ Changed cooperative gesture config and removed the strings from it in favor of the locale variable ([#3621](https://github.com/maplibre/maplibre-gl-js/pull/3621)) (by [@HarelM](https://github.com/HarelM))
 - ⚠️ Changed the terrain enable disable locale key to match the other keys' styles, updated the typings to allow using locale with more ease ([#3621](https://github.com/maplibre/maplibre-gl-js/pull/3621)) (by [@HarelM](https://github.com/HarelM))
 - Add "opacity" option and "setOpacity" method to Marker ([#3620](https://github.com/maplibre/maplibre-gl-js/pull/3620)) (by [@sbachinin](https://github.com/sbachinin))


melitele:
@HarelM ready for review

src/style/style_layer.ts
```diff
@@ -85,7 +90,9 @@ export abstract class StyleLayer extends Evented {
     minzoom: number;
     maxzoom: number;
     filter: FilterSpecification | void;
-    visibility: 'visible' | 'none' | void;
+    visibility: VisibilitySpecification | void;
+    private _visibility: 'visible' | 'none' | void;
```
HarelM: Is this "evaluatedVisibility"?

src/style/style_layer.ts
```diff
@@ -97,6 +104,8 @@ export abstract class StyleLayer extends Evented {
 
     _featureFilter: FeatureFilter;
 
+    _visibilityExpression: any;
```
HarelM: Can this be better typed?

src/style/style_layer.ts
```diff
@@ -281,10 +306,10 @@ export abstract class StyleLayer extends Evented {
         return false;
     }
 
-    isHidden(zoom: number, roundMinZoom: boolean = false) {
+    isHidden(zoom: number = this.minzoom, roundMinZoom: boolean = false) {
         if (this.minzoom && zoom < (roundMinZoom ? Math.floor(this.minzoom) : this.minzoom)) return true;
         if (this.maxzoom && zoom >= this.maxzoom) return true;
-        return this.visibility === 'none';
+        return this.visibility === 'none' || this._visibility === 'none';
```
HarelM: This seems a bit odd, can we somehow consolidate both fields?

src/style/style_layer.ts
```diff
@@ -323,7 +352,7 @@ export abstract class StyleLayer extends Evented {
 
         if (this.visibility) {
             output.layout = output.layout || {};
-            output.layout.visibility = this.visibility;
+            output.layout.visibility = this.visibility as ExpressionSpecification;
```
HarelM: Why is this needed? I would expect the typings to facilitate for this and not need to use "as ...".

src/style/style.test.ts
```diff
@@ -1871,6 +1871,34 @@ describe('Style.setGlobalStateProperty', () => {
...
+                source: 'circle-source-id',
+                layout: {
+                    'visibility': ['case', ['global-state', 'visibility'], 'visible', 'none']
+                }
+            }]
+        }));
+
+        await style.once('style.load');
+
+        style.tileManagers['circle-source-id'].resume = vi.fn();
```
HarelM, line 1892: This is a good test.
Is it possible to use "getSource" or something a bit more public in the test?

HarelM:
I've added some minor comments, overall this looks good, thanks for pushing this forward!

src/style/style.test.ts
melitele: why in this test?

src/style/style_layer.ts
melitele: yes

src/style/style.test.ts
HarelM: This is a new test...

melitele: there are 19 others in this file just like that...

HarelM: Yeah, so I would like to avoid adding more if possible...

melitele: [Original discussion](https://github.com/maplibre/maplibre-gl-js/pull/5613#discussion_r2082089022) on those tests accepted them in this form. The last change to those 19 tests was in PR #6635 and they were accepted without further discussion. Why is it no longer accepted to write tests in this way?

melitele: Regardless, give me a better method and I'll refactor them all.

HarelM: I was hoping `getSource` would solve this.
If this is not the case this comment can be marked as resolved.
The other PR was a "search and replace" and didn't add logic or tests.
Note that while I'm very tempted to ask to change the other tests as well, I did not ask that :-)
I've seen recently some bad refactoring problems when tests are using "private" fields so I'm more sensitive to it now (to answer your question on what had changed).
The tile manager refactoring was partially the cause of this new revelation/sensitivity...

src/style/style_layer.ts
melitele: [65ba631](https://github.com/maplibre/maplibre-gl-js/pull/6659/commits/65ba6312fa2ae7e0a10fa960c7d22e57ac1d9624)

src/style/style_layer.ts
melitele: [65ba631](https://github.com/maplibre/maplibre-gl-js/pull/6659/commits/65ba6312fa2ae7e0a10fa960c7d22e57ac1d9624)

src/style/style_layer.ts
melitele: [65ba631](https://github.com/maplibre/maplibre-gl-js/pull/6659/commits/65ba6312fa2ae7e0a10fa960c7d22e57ac1d9624)

src/style/style_layer.ts
```diff
@@ -125,6 +134,8 @@ export abstract class StyleLayer extends Evented {
         this.minzoom = layer.minzoom;
         this.maxzoom = layer.maxzoom;
 
+        this._visibilityExpression = createVisibilityExpression(this.visibility as VisibilitySpecification, globalState);
```
HarelM: Isn't `this.visibility` of type `VisibilitySpecification`? I'm not sure I understand why this `as` is needed...

src/style/style.test.ts
melitele: My preference is to refactor those tests in a separate PR as my initial attempt at implementing your suggestion using `getSource` was not successful. I can't tell yet how long it'll take and I don't want to hold up this PR just to make better tests.

HarelM: Fair enough, I can respect that.

src/style/style_layer.ts
melitele: You are `typescript` expert.
Screenshot 2025-12-14 at 4 57 09 AM

src/style/style_layer.ts
```diff
@@ -97,6 +104,8 @@ export abstract class StyleLayer extends Evented {
 
     _featureFilter: FeatureFilter;
 
+    _visibilityExpression: VisibilityExpression;
```
HarelM, line 107: There are now 3 properties which are very similar, and I don't know if I understand why all 3 are needed:
`visibility`, `_visibility`, `_visibilityExpression`.
I would suggest to give them better names to properly explain how each one is different and also try and see if all 3 are needed or is there a way to reduce it to 2 or maybe 1 field.

melitele: I believe they are all needed. If you disagree, tell me how to reduce them. Same with names.

src/style/style_layer.ts
HarelM: Is the definition of `createVisibilityExpression` correct? What would it do if you pass `void`? At the very least I would expect it to be `VisibilitySpecification | undefined` and not `void`. Am I missing anything?

melitele: [d4b1dd2](https://github.com/maplibre/maplibre-gl-js/pull/6659/commits/d4b1dd2b6cac2ca35a1151f0a8693a7fa8400594)

src/style/style_layer.ts
HarelM: Here's my understanding, and I might not understand is thoroughly.
`visibilty` is needed since this is what the user defined in the style, and we need to keep the "original" definition to be able to create expression and eveluate.
`_visibility` should be renamed to `_evaluatedVisibility` (as suggesting in a different comment), I'm guessing it is expensive to evaluate the expression every time there's a check for `isHidden` or otherwise when `_evaluatedVisibility` is needed , but if it's not expensive I guess this can be removed.
`_visibilityExpression` is caching the response of `createVisibilityExpression`, the question here as well is how expensive it is to use `createVisibilityExpression` on demand instead of only when visibility changes, as far as I can tell, once visibility is evaluated there's no need to keep `_visibilityExpression` isn't there? Can't we create the expression, evaluate and throw is away?
If this a performance bottle neck do let me know.

HarelM:
BTW, I'm planning to try out some CI changes I made so if you need to include a change in maplibre-style-spec do let me know.
- https://github.com/maplibre/maplibre-style-spec/pull/1424

src/style/style_layer.ts
melitele: [f91aab3](https://github.com/maplibre/maplibre-gl-js/pull/6659/commits/f91aab31782232373686e3b7c603b4518bdfa639)

> `_visibility` should be renamed to `_evaluatedVisibility` (as suggesting in a different comment)

so that was the reason for this other comment. I missed the intent completely.

> Can't we create the expression, evaluate and throw is away?

no, because of `setValue` which is there to avoid holding `_globalState` in the `StyleLayer` solely for the purpose of recreating `_visibilityExpression`. `FeatureFilter` can be refactored following this pattern and then `_globalState` reference can be removed from `StyleLayer` .

HarelM (approved):
THANKS!!

dannybloe:
There appears to be a problem due to this change when you use custom layers and call setGlobalStateProperty function. It will result in an error: TypeError: Cannot read properties of undefined (reading 'getGlobalStateRefs')

This can easily be reproduced by modifying the example https://maplibre.org/maplibre-gl-js/docs/examples/add-a-custom-style-layer/

And add map.setGlobalStateProperty('zork', 'dork') in the on load callback.

melitele:
> There appears to be a problem due to this change when you use custom layers and call setGlobalStateProperty function. It will result in an error: TypeError: Cannot read properties of undefined (reading 'getGlobalStateRefs')
> 
> This can easily be reproduced by modifying the example https://maplibre.org/maplibre-gl-js/docs/examples/add-a-custom-style-layer/
> 
> And add map.setGlobalStateProperty('zork', 'dork') in the on load callback.

Yes, it has already been fixed - PR #6886.

PR #6661, opened 2025-11-03
Add reduceMotion option to Map Options

wayofthefuture:
This PR introduces a new reduceMotion option to the Map configuration. The goal is to improve battery performance on iPad devices by allowing reduced motion while still preserving the ability to pan the map with inertia.

This change is fully backward-compatible and does not alter existing behavior unless the option is explicitly enabled. If reduceMotion is not set, the library will continue to respect the device’s system preference as before.

### DIFF
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -178,11 +178,6 @@ export type MapOptions = {
      */
     dragPan?: boolean | DragPanOptions;
     /**
-     * If `true`, gesture inertia (such as panning) is disabled. If not provided, gesture inertia defaults to the user's device settings.
-     * @defaultValue undefined
-     */
-    reduceMotion?: boolean | undefined;
-    /**
      * If `true`, keyboard shortcuts are enabled (see {@link KeyboardHandler}).
      * @defaultValue true
      */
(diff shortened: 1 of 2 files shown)


HarelM:
Can't you simply solve it with some CSS?

wayofthefuture:
I cannot be overridden via css unfortunately.

HarelM:
It cannot or it's not possible in your use case?

wayofthefuture:
It cannot.

src/ui/map.ts
```diff
@@ -177,6 +177,11 @@ export type MapOptions = {
      * @defaultValue true
      */
     dragPan?: boolean | DragPanOptions;
+    /**
+     * If `true`, gesture inertia (such as panning) is disabled. If not provided, gesture inertia defaults to the user's device settings.
+     * @defaultValue undefined
```
HarelM: Can you add the default to the list of the "complete" object?

src/util/browser.ts
```diff
@@ -60,4 +62,8 @@ export const browser = {
         }
         return reducedMotionQuery.matches;
     },
+
+    set prefersReducedMotion(value: boolean) {
```
HarelM, line 66: Can you add a unit test for that?
I know it's dumb, but still...

HarelM:
I've added some minor comments, I've read about it a bit and I see that there's no easy way to override that, so I see why this option is needed.

src/util/browser.ts
wayofthefuture: I actually like the idea of a test so it doesn't accidentally get reversed

wayofthefuture:
Test added and default option set. I also moved the setting further down the options list as well.

PR #6666, opened 2025-11-04
Fix symbol label density at extreme zoom overscaling

wayofthefuture:
Possible alternative to: https://github.com/maplibre/maplibre-gl-js/pull/6521.

Fixed "Too many glyphs being rendered in a tile" errors at high overscaling levels by dynamically adjusting symbol spacing based on the overscaling factor. Previously attempted to clamp the overscaling value to 128, which wasn't a complete fix and didn't address all cases. The new approach scales spacing proportionally (overscaling / 64) to maintain even label distribution while preventing glyph count from exceeding the 65,535 limit.

### DIFF
--- a/src/data/bucket/symbol_bucket.ts
+++ b/src/data/bucket/symbol_bucket.ts
@@ -363,7 +363,7 @@ export class SymbolBucket implements Bucket {
     constructor(options: BucketParameters<SymbolStyleLayer>) {
         this.collisionBoxArray = options.collisionBoxArray;
         this.zoom = options.zoom;
-        this.overscaling = isSafari(globalThis) ? Math.min(options.overscaling, 128) : options.overscaling;
+        this.overscaling = options.overscaling;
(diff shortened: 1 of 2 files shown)


HarelM:
This will create a situation where you might pan a long time in order to find a lanel on a line, wouldn't it?

wayofthefuture:
Yes if you are zoomed very far. So you could possible provide an option:

```js
/**
 * The overscaling threshold for applying additional label spacing.
 * When tile overscaling exceeds this value, label spacing is increased
 * proportionally to prevent excessive label density at high zoom levels.
 * @defaultValue 64
 */
labelSpacingOverscaleFactor?: number;
```

Or the on-the-fly tile slicing.

HarelM:
But then there's no solution if you want to zoom in a lot and still see label with high density, right?

wayofthefuture:
Right

Issue #6667, opened 2025-11-04
Optimize GeoJSON Data Transfer - Avoid Sending Full Source Data on Diff Updates


wayofthefuture:
### Problem

Currently, when a GeoJSON source receives a **diff update** (via `updateData()` with `GeoJSONSourceDiff`), the worker thread processes the incremental changes efficiently, but the **entire GeoJSON dataset is sent back** to the main thread as part of the response.

**Workflow:**
1. Main thread → Worker: Small diff update (`add`, `remove`, `update` operations)
2. Worker: Efficiently applies diff to `_dataUpdateable` map
3. Worker → Main thread: **Sends back entire source data** 

This is inefficient for large datasets where diffs may only affect a few features.

### Root Cause

The issue stems from how `loadData()` works in `geojson_worker_source.ts`. The full dataset is included in the response (`GeoJSONWorkerSourceLoadDataResult`), which then gets transferred back to the main thread.

### Complicating Factor: `serialize()` Usage

The main challenge is that `serialize()` is used elsewhere in the codebase and is **synchronous**. This design pattern assumes data is available on the main thread for serialization purposes (e.g., for state persistence, style serialization, etc.).

Making data worker-only would break this existing synchronous API pattern.

#### Option 1: Add `serializable` option to Source Options
Add a source-level option that allows users to opt-out of data being sent back. This would allow non-breaking behavior for users to opt-in to this functionality.

#### Option 2: Remove `data` from the event
Remove the `data` object from the event. This would be breaking, but a user who is listening for the event could simply use getData asynchronously inside the callback for this event.

#### Option 3: Make `serialize` asynchronous
If serialize is made asynchronous, then getData() can be used inside of it and the source data can be consolidated to the worker. This is likely not viable option due to the complexity.

CC: @HarelM @andrewda @lucaswoj @lhapaipai

andrewda:
I think it would be helpful to get some profiling metrics to better understand how much of a performance hit this is causing at various data sizes.

lucaswoj:
The main thread already has both the original GeoJSON data and the diff. We can simply run `applySourceDiff` there and avoid sending the full GeoJSON string back. No breaking change needed.

And the performance opportunity is clear. The cost of transferring large objects from `WebWorkers` to the main thread is well understood to be high. The GeoJSON diff feature was specifically designed to address this problem. 

[...]

PR #6668, opened 2025-11-04
Improve GeoJSON#updateData performance on small diffs by not sending entire "data" back from worker

lucaswoj:
fix https://github.com/maplibre/maplibre-gl-js/issues/6667
ref https://github.com/maplibre/maplibre-gl-js/issues/4364

`GeoJSONSource#updateData` was created specifically to avoid sending entire serialized GeoJSON from main thread to the worker when only a few features changed. But we still send the entire serialized GeoJSON from the worker to main thread. This isn't necessary. We can apply the diff within the main thread to its existing copy of the data.

This results in a 50% speedup on the `GeoJSONDiff` bench. Actual speedup of real world cases will vary but none will be slower.
```
                               v5.12.0                       main  geojson-diff-perf 667f56c 
 GeoJSONDiff               33.3291 ms                 33.3170 ms                 16.6608 ms 
```

## Launch Checklist

 - [x] Link to related issues.
 - [x] Post benchmark scores.

### DIFF
--- a/src/source/geojson_source.ts
+++ b/src/source/geojson_source.ts
@@ -1,5 +1,5 @@
 import {Event, ErrorEvent, Evented} from '../util/evented';
-import {extend, warnOnce} from '../util/util';
+import {extend, warnOnce, type ExactlyOne} from '../util/util';
 import {EXTENT} from '../data/extent';
 import {ResourceType} from '../util/request_manager';
 import {browser} from '../util/browser';
@@ -137,7 +137,11 @@ export class GeoJSONSource extends Evented implements Source {
 
     isTileClipped: boolean;
     reparseOverscaled: boolean;
-    _data: GeoJSON.GeoJSON | string | undefined | globalThis.Map<GeoJSONFeatureId, GeoJSON.Feature>;
+    _data: ExactlyOne<{
+        url: string;
+        geojson: GeoJSON.GeoJSON;
+        updateable: globalThis.Map<GeoJSONFeatureId, GeoJSON.Feature>;
+    }>;
     _options: GeoJSONSourceInternalOptions;
     workerOptions: GeoJSONWorkerOptions;
     map: Map;
@@ -169,7 +173,7 @@ export class GeoJSONSource extends Evented implements Source {
         this.actor = dispatcher.getActor();
         this.setEventedParent(eventedParent);
 
-        this._data = options.data;
+        this._data = typeof options.data === 'string' ? {url: options.data} : {geojson: options.data};
         this._options = extend({}, options);
 
         this._collectResourceTiming = options.collectResourceTiming;
@@ -250,7 +254,7 @@ export class GeoJSONSource extends Evented implements Source {
     setData(data: GeoJSON.GeoJSON | string, waitForCompletion: true): Promise<void>;
     setData(data: GeoJSON.GeoJSON | string, waitForCompletion?: false): this;
     setData(data: GeoJSON.GeoJSON | string, waitForCompletion?: boolean): this | Promise<void> {
-        this._data = data;
+        this._data = typeof data === 'string' ? {url: data} : {geojson: data};
         this._pendingWorkerUpdate = {data};
         const updatePromise = this._updateWorkerData();
         if (waitForCompletion) return updatePromise;
@@ -421,7 +425,7 @@ export class GeoJSONSource extends Evented implements Source {
             }
 
             if (!result.shouldApplyDiff) {
-                this._data = result.data;
+                this._data = {geojson: result.data};
             } else {
                 this._applyDiff(diff);
             }
@@ -460,15 +464,15 @@ export class GeoJSONSource extends Evented implements Source {
 
         // Lazily convert `this._data` to updateable if it's not already
         if (
-            typeof this._data !== 'string' &&
-            !(this._data instanceof globalThis.Map) &&
-            isUpdateableGeoJSON(this._data, promoteId)
+            !this._data.url &&
+            !this._data.updateable &&
+            isUpdateableGeoJSON(this._data.geojson, promoteId)
         ) {
-            this._data = toUpdateable(this._data, promoteId);
+            this._data = {updateable: toUpdateable(this._data.geojson, promoteId)};
         }
 
-        if (diff && this._data instanceof globalThis.Map) {
-            applySourceDiff(this._data, diff, promoteId);
+        if (diff && this._data.updateable) {
+            applySourceDiff(this._data.updateable, diff, promoteId);
         } else {
             // This should never happen because the worker would not set `shouldApplyDiff: true` if the source was not updateable.
             warnOnce('Cannot apply GeoJSONSource#updateData due to internal error');
@@ -588,12 +592,12 @@ export class GeoJSONSource extends Evented implements Source {
     serialize(): GeoJSONSourceSpecification {
         return extend({}, this._options, {
             type: this.type,
-            data: this._data instanceof globalThis.Map ?
+            data: this._data.updateable ?
                 {
                     type: 'FeatureCollection',
-                    features: Array.from(this._data.values())
+                    features: Array.from(this._data.updateable.values())
                 } :
-                this._data
+                this._data.url || this._data.geojson
         });
     }
 
--- a/src/util/util.ts
+++ b/src/util/util.ts
@@ -1086,6 +1086,13 @@ export type Complete<T> = {
  */
 export type RequireAtLeastOne<T> = { [K in keyof T]-?: Required<Pick<T, K>> & Partial<Pick<T, Exclude<keyof T, K>>>; }[keyof T];
 
+/**
+* A helper to allow require exactly one one property
+ */
+export type ExactlyOne<T, Keys extends keyof T = keyof T> = {
+    [K in Keys]: Required<Pick<T, K>> & { [P in Exclude<Keys, K>]?: never }
+}[Keys];
+
 export type TileJSON = {
     tilejson: '2.2.0' | '2.1.0' | '2.0.1' | '2.0.0' | '1.0.0';
     name?: string;
--- a/test/build/min.test.ts
+++ b/test/build/min.test.ts
@@ -38,7 +38,7 @@ describe('test min build', () => {
         const decreaseQuota = 4096;
 
         // feel free to update this value after you've checked that it has changed on purpose :-)
-        const expectedBytes = 1014309;
+        const expectedBytes = 1015376;
 
         expect(actualBytes, `Consider changing expectedBytes to: ${actualBytes}`).toBeLessThan(expectedBytes + increaseQuota);
         expect(actualBytes).toBeGreaterThan(expectedBytes - decreaseQuota);


src/source/geojson_source.ts
```diff
@@ -412,7 +414,16 @@ export class GeoJSONSource extends Evented implements Source {
                 return;
             }
 
-            this._data = result.data;
+            if (result.dataDiff) {
+                if (!this._dataUpdateable) {
+                    this._dataUpdateable = toUpdateable(this._data as any, undefined);
```
lucaswoj: Re: `this._data as any` -- this should always be a `GeoJSON.FeatureCollection` as required, but we need to validate this assumption and better handle this typecast.

lucaswoj: Re: `undefined`, we need to pass the correct `promoteId` parameter here

src/source/geojson_source.ts
```diff
@@ -412,7 +414,16 @@ export class GeoJSONSource extends Evented implements Source {
...
+                }
+
+                applySourceDiff(this._dataUpdateable, result.dataDiff);
+            } else {
+                this._data = result.data;
+                this._dataUpdateable = undefined;
```
lucaswoj: Confirm we need to reset `this._dataUpdateable` here AND in `setData`

wayofthefuture:
Wow, very nice... the only thing that stands out to me is the entire data is stored in _dataUpdatable in the worker, but in the main thread it is stored in both _data and _dataUpdatable. Not sure if if can both be stored in _dataUpdatable like in the worker?

wayofthefuture:
First, thank you again for taking the time to work on this. Our current usage falls into two main cases:

1. Partial feature updates: We have a dataset of roughly 500 features, and we typically update only a subset (maybe a dozen or so) at a time.
2. Full frame-by-frame updates: We also have scenarios where we update all ~500 features every animation frame. Currently, we use setData() for this, since it replaces the entire feature set at once.

Your PR already provides a significant improvement for use case 1, which is great and very appreciated.

However, we’re wondering if there’s a way to further optimize for use case 2 as well. This would really help support our MapLibre usage where we need smooth animations involving many rapidly changing features. Please see:

https://github.com/maplibre/maplibre-gl-js/blob/d5ed620801df04c013b12bddc44f9f1cc725c801/src/style/style.ts#L1009-L1013

The event pipeline has a function that calls TileManager.serialize(), which then calls Source.serialize(), which then return source._data.

https://github.com/maplibre/maplibre-gl-js/blob/d5ed620801df04c013b12bddc44f9f1cc725c801/src/util/evented.ts#L141-L148

https://github.com/maplibre/maplibre-gl-js/blob/d5ed620801df04c013b12bddc44f9f1cc725c801/src/ui/map.ts#L811-L814

https://github.com/maplibre/maplibre-gl-js/blob/d5ed620801df04c013b12bddc44f9f1cc725c801/src/tile/tile_manager.ts#L208-L210

https://github.com/maplibre/maplibre-gl-js/blob/d5ed620801df04c013b12bddc44f9f1cc725c801/src/source/geojson_source.ts#L543-L548

The event data is currently populated using serialize in geojson_source.ts when the map receives the event and calls fire. The serialize method simply returns this._data, which means that as long as the data in geojson_source.ts is kept up to date, there’s no need for the worker to send **any** data back to the main thread.

In other words, if we maintain a single authoritative copy of the data in geojson_source (mirroring what the worker stores), then serialize can always return this._data directly. This eliminates the need to send updated feature collections from the worker back to the main thread.

This would improve performance because large data updates would only need to be transmitted one way (main thread → worker), instead of round-trip. For workloads with frequent full-dataset updates, this could substantially reduce overhead and improve runtime smoothness.

wayofthefuture:
Please see https://github.com/maplibre/maplibre-gl-js/pull/6688.

I have added a 2nd parameter to updateData and setData. Maybe a better alternative to the changes is to convert this to an options object and create an additional option here called `returnData` that defaults to true. If you set it to false it doesn't send any data back from the worker on both setData or updateData. This would be fully backward compatible and only used for people needed high speed updates.

HarelM:
I'm not sure if I should review this since it's in draft.
Do let me know if, when and what to review, as I'm a bit lost with all the PRs related to geojson updates...

lucaswoj:
When this PR is ready for review, I will provide a detailed technical summary of what's changed in the PR description, and mark it as "Ready for Review"

src/source/geojson_source.ts
```diff
@@ -412,7 +414,23 @@ export class GeoJSONSource extends Evented implements Source {
                 return;
             }
 
-            this._data = result.data;
+            if (result.shouldApplyDiff) {
+                const data = this._data;
+                const promoteId = typeof this.promoteId === 'string' ? this.promoteId : undefined;
+
+                // This expression should always return `true`. The worker would not pass `shouldApplyDiff: true`
+                // if a diff was not provided or the data is not updateable.
```
lucaswoj: TODO add an exception or warning if this expression is NOT true for some reason

lucaswoj:
@HarelM 

> I'm a bit lost with all the PRs related to geojson updates...

Totally fair! I've got several PRs in flight and @wayofthefuture has several more. I wrote a high level roadmap and status report at https://github.com/maplibre/maplibre-gl-js/issues/4364#issuecomment-3514275432 for your benefit. I hope that helps.

lucaswoj:
@wayofthefuture Extending this behavior to `GeoJSONSource#setData` is possible and a good idea. However, I'd like to target that in a separate PR because

 - it only provides a 10% speedup (assuming 100k point features)
 - it relates to rather different codepaths
 - it needs to be measured via an entirely different benchmark
 - it may result in some breaking public API changes
 
 Two smaller separate PRs will be easier to write, review, and debug than one mega PR!

src/source/geojson_source.ts
```diff
@@ -412,7 +412,23 @@ export class GeoJSONSource extends Evented implements Source {
...
+
+                if (this._data instanceof globalThis.Map) {
+                    applySourceDiff(this._data, diff, promoteId);
+                } else {
+                    // This should never happen because whe worker would not set `shouldApplyDiff: true` if the source was not updateable.
+                    warnOnce('Cannot apply GeoJSONSource#updateData due to internal error');
+                }
+            } else {
+                this._data = result.data;
+            }
```
wayofthefuture: Please consider matching the flow of the if directly above this block and moving the !== 'string' check to the top of isUpdateableGeoJSON as it should probably be there. I also think the warning should be removed.

```ts
if (data) {
    this._data = result.data;
} else if (diff) {
    const promoteId = typeof this.promoteId === 'string' ? this.promoteId : undefined;
    
    if (!(this._data instanceof globalThis.Map) && isUpdateableGeoJSON(this._data, promoteId)) {
        this._data = toUpdateable(this._data, promoteId);
    }
    
    if (this._data instanceof globalThis.Map) {
        applySourceDiff(this._data, diff, promoteId);
    }
}
```

lucaswoj: In the latest commit, I've moved the `data` case to the top of this if/else statement.

I'd rather leave the `typeof this._data !== 'string'` check here, because it's highly specific to the `GeoJSONSource` where this value could be a URL to a GeoJSON source, and not relevant to any other callosities of this API. I added some additional comments to clarify.

lucaswoj:
The automated `GeoJSONDiff` benchmark reports a 50% speedup 🎉 

```
                               v5.12.0                       main  geojson-diff-perf 667f56c 
 GeoJSONDiff               33.3291 ms                 33.3170 ms                 16.6608 ms 
```

lucaswoj:
This is ready for review

src/source/geojson_source.ts
```diff
@@ -412,7 +412,32 @@ export class GeoJSONSource extends Evented implements Source {
                 return;
             }
 
-            this._data = result.data;
+            if (!result.shouldApplyDiff) {
+                this._data = result.data;
```
wayofthefuture: If result.data then shouldApplyDiff will always be false so it's not needed

src/source/geojson_source.ts
```diff
@@ -412,7 +412,32 @@ export class GeoJSONSource extends Evented implements Source {
...
+            } else {
+                const promoteId = typeof this.promoteId === 'string' ? this.promoteId : undefined;
+
+                // Lazily convert `this._data` to updateable if it's not already
+                if (
+                    // If this._data is a string, it's a still URL.
+                    // This should not happen because the worker must have already loaded and parsed the data before sending this message.
+                    typeof this._data !== 'string' &&
```
wayofthefuture: Previous version was better...

src/source/geojson_source.ts
```diff
@@ -412,7 +412,32 @@ export class GeoJSONSource extends Evented implements Source {
...
+                    isUpdateableGeoJSON(this._data, promoteId)
+                ) {
+                    this._data = toUpdateable(this._data, promoteId);
+                }
+
+                if (this._data instanceof globalThis.Map) {
+                    applySourceDiff(this._data, diff, promoteId);
+                } else {
+                    // This should never happen because whe worker would not set `shouldApplyDiff: true` if the source was not updateable.
+                    warnOnce('Cannot apply GeoJSONSource#updateData due to internal error');
```
wayofthefuture: If it should never happen then remove otherwise a test should cover it

src/source/geojson_source.ts
lucaswoj: That's not necessarily true. `data` could be `null`. I don't want to rely on different semantics for `null` vs `undefined`.

src/source/geojson_source.ts
lucaswoj: As in, the version without the code comments?

src/source/geojson_source.ts
lucaswoj: Are you sure you'd prefer a silent failure instead of a clear error message? AFIK this invariant holds now, but someone else making changes to this part of the codebase could easily encounter this case.

src/source/geojson_source.ts
lucaswoj: Removed the comments in 8cbd0af45

src/source/geojson_source.ts
wayofthefuture: So you mean if (!diff)?

lucaswoj: We need to keep the data in the main thread and the worker in sync. It's possible that `diff` is defined but the worker didn't apply the diff because the request was aborted or errored.

wayofthefuture (approved):

wayofthefuture:
@HarelM I approved but handing off to you for review and final approval / merging...

src/source/geojson_source.ts
```diff
@@ -412,7 +412,27 @@ export class GeoJSONSource extends Evented implements Source {
                 return;
             }
 
-            this._data = result.data;
+            if (!result.shouldApplyDiff) {
```
HarelM, line 427: in the worker it's either `data` or `shouldApplyDiff` I think checking if `data` is defined i.e.`if (result.data)` would be better here, also because data is what is being set, or it is the same comment as below?

lucaswoj: I think this is the same comment as below (https://github.com/maplibre/maplibre-gl-js/pull/6668/files/2c538b6de3a9331373b352e5d8eda51cf8400a13#r2519051566)

src/source/geojson_source.ts
```diff
@@ -137,7 +137,7 @@ export class GeoJSONSource extends Evented implements Source {
 
     isTileClipped: boolean;
     reparseOverscaled: boolean;
-    _data: GeoJSON.GeoJSON | string | undefined;
+    _data: GeoJSON.GeoJSON | string | undefined | globalThis.Map<GeoJSONFeatureId, GeoJSON.Feature>;
```
HarelM: Is there a way to simplify the type of `_data`?
There's a lot of checks to check its type and act accordingly which makes the code fragile, so if we can normalize it once and treat it the same it would make the code simper I believe.

src/source/geojson_source.ts
HarelM: Can this happen if `_data` is a string?

src/source/geojson_source.ts
lucaswoj: I can't think of a simplification that doesn't come with a performance penalty. By my count, there are 2 places that we check the type of `this._data`

src/source/geojson_source.ts
HarelM: Also I think this entire new section should be moved to a method of its own, the updateData method is too long.

src/source/geojson_worker_source.ts
```diff
@@ -125,7 +125,10 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
 
             this.loaded = {};
 
-            const result = {data} as GeoJSONWorkerSourceLoadDataResult;
+            const result: GeoJSONWorkerSourceLoadDataResult = params.dataDiff && isUpdateableGeoJSON(data) ?
```
wayofthefuture, line 128: Here we are parsing the entire dataset on every diff update via isUpdateableGeoJSON. however, in the above lines:
```ts
this._pendingData = this.loadAndProcessGeoJSON(params, this._pendingRequest);
 const data = await this._pendingData;
```
isUpdateableGeoJSON is already being called on that data in the functions... can you just check if it's truthy?

src/source/geojson_source.ts
HarelM: Can you elaborate on the performance penalty?
2 places for a type with 3 possible options sounds like a bug waiting to happen to be honest...

src/source/geojson_source.ts
lucaswoj: I had a comment explaining why each case should never occur but I removed it per https://github.com/maplibre/maplibre-gl-js/pull/6668#discussion_r2519055502

`this._data` cannot be a string at this point. Once the `loadData` response has been received for a diff, we're guaranteed that the data has been loaded and set to an object in `GeoJSONSource`.

HarelM: Comments are not a great solution for this type of issue, this relates to the comment I made about the different types of the same `_data` field.

src/source/geojson_worker_source.ts
wayofthefuture: maybe this is a subject for another pr because I kind of think isUpdateableGeoJson should be completely removed. it's json or its not.

lucaswoj: If we got rid of the `isUpdateableGeoJSON` check, we'd need to do an unsafe `data as UpdateableGeoJSON` expression, which I'd like to avoid unless we're working around a measurable performance bottleneck.

src/source/geojson_source.ts
lucaswoj: Well, we could "normalize" this value into `string | globalThis.Map<GeoJSONFeatureId, GeoJSON.Feature>`. At that point, we'd still need to check it's type at runtime, and we'd have to convert every GeoJSON object into an updatable map, even if the user had no intention of ever using the `updataData` API.

src/source/geojson_worker_source.ts
wayofthefuture: If you have 100,000 features it's a full traversal for each update... of course I could just be over micro-optimizing

src/source/geojson_source.ts
lucaswoj: Do you still have a question or concern about this code? Or can we resolve this thread while continuing the `_data` type conversation in https://github.com/maplibre/maplibre-gl-js/pull/6668#discussion_r2519623541?

src/source/geojson_source.ts
HarelM: I see. Another approach might be to define more obvious names for the same object maybe?
Something like
```ts
_data: {
   url?: string,
   geojson?: GeoJSON.GeoJSON,
   updateable?: globalThis.Map<GeoJSONFeatureId, GeoJSON.Feature>
}
```
This way they should be mutually exclusive and you can better understand the state of the component by checking which one of the has something.
It's very similar to type checking but I think it's more readable.
Let me know what you think, or if you have a better idea how to solve this readability complexity.

src/source/geojson_source.ts
HarelM: Resolving in favor of the other thread.

src/source/geojson_source.ts
lucaswoj: I can rewrite that way if you want but the code will be longer, more complex, and less typesafe. Please confirm if you'd like me to go ahead with that.

HarelM: Do you have a better idea on how to solve this?
If not, maybe a PR against this PR so we can see how the code looks "before" and "after" and can decide together which version looks more complex.

wayofthefuture:
Is it readability or am I just an idiot??? LOL 🤣 🤣 😜

src/source/geojson_worker_source.ts
lucaswoj: This `isUpdateableGeoJSON` call accounts for 0.3% of the overall runtime of the `updateData` operation.

src/source/geojson_source.ts
lucaswoj: IMO performance is a good reason to tolerate slightly more complex code. The code I am putting forward here for review is my best idea. What's great about this type union is that the compiler can guarantee we are handling every case properly.

src/source/geojson_worker_source.ts
wayofthefuture: Roger.

src/source/geojson_source.ts
HarelM: That's not true as can be seen in a previous thread, the union type might work to some extent for the compiler, but less for people reading the code.
Can you please open a PR against this PR with the suggested change to see if it make readability better?

lucaswoj: Here's a PR against this PR https://github.com/lucaswoj/maplibre-gl-js/pull/1. I'd need to do some more testing before merging but it gives an idea.

src/source/geojson_source.ts
lucaswoj: > Also I think this entire new section should be moved to a method of its own, the updateData method is too long.

Done in 0c675fc67

src/source/geojson_source.ts
lucaswoj: > That's not true as can be seen in a previous thread

What are you referring to?

HarelM: Never mind, let's focus on the options at hand.
I find it more readable to see the assignment of the relevant type and the checks.
A few things can be improved in the above PR, see my comments there.

Thanks!

HarelM: I can comment inside the PR if that would be better.

lucaswoj: Please comment in the sub-PR so that we can get separate threads going for each topic

HarelM: Done.

lucaswoj: I'm assuming you want to go ahead with this design so I merged the PR into this branch.

src/source/geojson_source.ts
```diff
@@ -420,7 +424,11 @@ export class GeoJSONSource extends Evented implements Source {
                 return;
             }
 
-            this._data = result.data;
+            if (!result.shouldApplyDiff) {
```
HarelM, line 427: [Optional] Consider inverting this if so `shouldApplyDiff` will call `this._applyDiff`.

lucaswoj: It was that order originally and then @wayofthefuture insisted on reversing the order in https://github.com/maplibre/maplibre-gl-js/pull/6668#discussion_r2518397752. I don't care either way.

lucaswoj: I'd prefer to merge without making this change, if you're open to it.

HarelM: I think the difference then was that this wasn't a single line of code. But it's OK either way.

Issue #4474, opened 2024-07-31
Text animation in cluster map lags behind zoom transition on MapLibre GL JS


m345054:
maplibre-gl-js version: 4.5.0

browser: Chrome, Firefox

**Steps to Trigger Behavior**
Open the cluster example on MapLibre: [Cluster Map Example](https://maplibre.org/maplibre-gl-js/docs/examples/cluster/)
Zoom in and out on a cluster.
Observe the text animation lag.

**Link to Demonstration**
A minimal demonstration can be seen directly on the official MapLibre example page for clustering: [Cluster Map Example](https://maplibre.org/maplibre-gl-js/docs/examples/cluster/)

**Expected Behavior**
The text within the clusters should update smoothly and in sync with the map zoom level without any noticeable delay.

**Actual Behavior**
When zooming in and out on a cluster, there is a noticeable delay where the text animation lags behind. Adjusting the transition duration settings as per the MapLibre style specification does not resolve the issue.

<img width="924" alt="147140947-70b3672a-3eb6-44d8-ba6c-73188d36262d" src="https://github.com/user-attachments/assets/93c229da-6aa5-412c-9e50-21cabbea1424">

<img width="924" alt="147140661-fd6c15e4-0c7e-47a9-a708-ff315ae8f41c" src="https://github.com/user-attachments/assets/c21ebc7a-49d0-494e-aa12-f5d8530bee55">

HarelM:
The circle transition and the text transition seems to have different duration, what did you try that didn't work in terms of changing the transition duration?

Christopher-Hayes:
I ran into this issue as well. I ended up scrapping clusters, so don't have the code, but you can see this issue on the official maplibre example.

## Reproduction

The cluster circle and text are not in sync, the circle instantly jumps to the new location/size, the text fades out and then pops up in its new location

https://github.com/user-attachments/assets/9d686fca-6a13-4f2a-b8ba-9247e9a1153e

_Video from the official cluster example: https://maplibre.org/maplibre-gl-js/docs/examples/create-and-style-clusters/_

This might be a side-effect of [`transition`](https://maplibre.org/maplibre-style-spec/transition/) and [`fadeDuration`](https://maplibre.org/maplibre-gl-js/docs/API/type-aliases/MapOptions/#fadeduration) being on by default, with Transition only affecting the text visibility. I'm not 100% sure, transitions are a bit mysterious to me.

## Solutions

[...]

PR #6679, opened 2025-11-09
Set fade duration to 0 for all text in the map in the cluster example

HarelM:
## Launch Checklist

- Fixes: #4474

I couldn't make the circles fade-in in the example, and for some reason I could make the text not to fade-in when setting the relevant `text-allow-overlap: true` for the layer, so I ended up setting `fadeDuration` to 0 for all the text :-/

I also improved the types a bit and moved a string that was reused to the right place.

 - [x] Link to related issues.


birkskyum (commented):
I see some non-example files changed, with this TRANSITION_SUFFIX, but there's not a changelog

HarelM:
These are small refactoring that I did while looking at the code to see if I can find out how to set transition to text per layer.
Not a real change...

PR #6681, opened 2025-11-10
GeoJSON Source Diff: improve, abstract, refactor, optimize, and fix regressions.

wayofthefuture:
## Summary

This PR refactors and improves `geojson-source-diff`. The previous implementation combined multiple responsibilities within the same function logic (nullish checking, diff merging, feature merging, and map/set construction), which made behavior harder to follow and contributed to regressions, especially when using `promoteId`.

## Key Improvements

- **Clear separation of responsibilities**
  - Explicit `Map`/`Set` based conversions are abstracted from the logic.
  - Idempotency of hashed collections to simplify the nullish checking logic.
  - New `mergeFeatureDiffs` method to separate out feature merging.
  - Improved logic order - establishing merge conflicts - then executing the merge.
  
- **Consistent `promoteId` handling**
  - Feature IDs are normalized before merge and restored afterwards
  - Prevents mismatched identity conflicts during merges using promotedId.

- **Performance / complexity improvements**
  - Eliminates repeated full-array scans during merge and apply operations.
  - Merges now operate at O(n) complexity and removed O(n^2) op.

- **Correct some regressions**
  - Since the underlying source may already contain a feature with a given ID (even if a diff previously “added” one with the same ID), **a subsequent remove in a merged diff must never be dropped**.
  - This restores intended behavior where removes always take precedence unless explicitly overridden by a subsequent add.

## Result

Merging and applying diffs is now more predictable, easier to read and understand, consistent with promoteId usage, and resilient to reuse. The logic is clearer, more maintainable, and reduces the risk of subtle regressions going forward.


src/source/geojson_source_diff.ts
```diff
@@ -1,10 +1,54 @@
+/**
```
HarelM: This comment is very nicely written, but it should be placed somewhere that the IDE or the docs can use it, I would recommend moving it to describe one of the methods below, or split it to improve the docs of each method.
Keeping it here will get stale and probably no one will read it, which is a shame :-)

src/source/geojson_source_diff.ts
```diff
@@ -131,115 +175,225 @@ export function applySourceDiff(updateable: Map<GeoJSONFeatureId, GeoJSON.Featur
...
+        // Temporarily normalize diff.add for features using promoteId
+        promoteFeatureIds(prevDiff.add, promoteId);
+        promoteFeatureIds(nextDiff.add, promoteId);
+    }
+
+    // Hash for o(1) lookups while creating a mutatable copy of the collections
+    const prev = diffToHashed(prevDiff);
+    const next = diffToHashed(nextDiff);
+
+    // Resolve merge conflict - removing all features with added or updated features in previous - and clear no-op removes
```
HarelM: I think the whole part with the `// Resolve merge conflict - ...` can be moved to a separate method to handle merge conflicts, this will reduce the verbosity of the comments at least :-)

src/source/geojson_source_diff.ts
```diff
@@ -131,115 +175,225 @@ export function applySourceDiff(updateable: Map<GeoJSONFeatureId, GeoJSON.Featur
...
-            merged.update = merged.update.filter(f => !newRemovedSet.has(f.id));
+    const merged: GeoJSONSourceDiffHashed = {};
+
+    // Merge the two diffs
+    if (prev.removeAll || next.removeAll) merged.removeAll = true;
+    merged.remove = new Set([...prev.remove , ...next.remove]);
+    merged.add    = new Map([...prev.add    , ...next.add]);
+    merged.update = new Map([...prev.update , ...next.update]);
+
+    // Squash the merge - removing then adding the same feature
```
HarelM, line 224: What will happen in the case of `remove` that has some properties and then `add` with other properties (or geometry)? Is this covered in the tests?

HarelM: It's also the question about what happens when you call add with a feature that already exists but has different properties/geometry.

HarelM:
I've added a few minor comments, and question about a scenario, looks good otherwise.

src/source/geojson_source_diff.ts
wayofthefuture: Good idea

src/source/geojson_source_diff.ts
wayofthefuture: So I think what you are referring to is applySourceDiff and not mergeSourceDiff... which is a map that just overwrites the key on add. Nevertheless a test would be great to cover this so I went ahead and created it.

wayofthefuture:
I think there are many GeoJSON PRs right now lol! I think this is ready 😜

HarelM:
Can you check the coverage report (maybe locally) and make sure there are not missing use cases in the tests?

src/source/geojson_source_diff.ts
HarelM: I think it documents `GeoJSONFeatureId` now...?

HarelM:
Just the two comments above about coverage and docs location. Otherwise this can be merged.

wayofthefuture:
Awesome! I found another regression... Hopefully we have gotten them all now!

src/source/geojson_source_diff.ts
```diff
@@ -1,10 +1,19 @@
+/**
+ * GeoJSON Source Diff:
```
HarelM: This still remains a "hanging" comment, I suggest to either remove it or split it and move the parts into the method's tsdocs comments.

wayofthefuture:
Done!

HarelM:
Oh, wait, I think this needs a changelog entry as it changes the behavior a bit and not a "straight forward" refactoring.

CHANGELOG.md
```diff
@@ -11,7 +12,7 @@
 
 ### ✨ Features and improvements
 
-- Add support for MapLibre Tiles (MLT) by using `encoding: 'mlt'` in vector source definition ([#6570](https://github.com/maplibre/maplibre-gl-js/pull/6570)) (by [@Salkin975](https://github.com/Salkin975) and [@HarelM](https://github.com/HArelM)) 
+- Add support for MapLibre Tiles (MLT) by using `encoding: 'mlt'` in vector source definition ([#6570](https://github.com/maplibre/maplibre-gl-js/pull/6570)) (by [@Salkin975](https://github.com/Salkin975) and [@HarelM](https://github.com/HArelM))
```
```suggestion
- Add support for MapLibre Tiles (MLT) by using `encoding: 'mlt'` in vector source definition ([#6570](https://github.com/maplibre/maplibre-gl-js/pull/6570)) (by [@Salkin975](https://github.com/Salkin975) and [@HarelM](https://github.com/HarelM))
```

PR #6688, opened 2025-11-10
GeoJSONSource setData and updateData waitForCompletion option

wayofthefuture:
Adds an optional `waitForCompletion` parameter to `setData()` and `updateData()` methods on `GeoJSONSource`. When set to `true`, these methods return a `Promise<this>` that resolves after the worker completes processing the data, allowing users to await data operations. When `false` or omitted, the methods return `this` synchronously, maintaining full backwards compatibility with existing code.

This is useful for sequencing data operations, coordinating updates with other map actions, performing animations, and testing code that depends on data being loaded.

**Example:**
```typescript
await source.setData(geojsonData, true);
await source.updateData({add: [feature]}, true);
```

### DIFF
--- a/src/source/geojson_source.ts
+++ b/src/source/geojson_source.ts
@@ -458,10 +458,19 @@ export class GeoJSONSource extends Evented implements Source {
 
         const prevIds = new Set([...update.map(u => u.id), ...remove]);
 
+        for (const id of prevIds.values()) {
+            if (typeof id !== 'number') {
+                warnOnce(`GeoJSONSource "${this.id}": updateData is slower when using string GeoJSON feature IDs (e.g. "${id}"). Consider using numeric IDs for better performance.`);
+                return undefined;
+            }
+        }
+
         const nextBounds = [
             ...update.map(f => f.newGeometry),
             ...add.map(f => f.geometry)
-        ].map(g => getGeoJSONBounds(g));
+        ]
+            .filter(Boolean)
+            .map(g => getGeoJSONBounds(g));
 
         return {
             nextBounds,
@@ -474,7 +483,9 @@ export class GeoJSONSource extends Evented implements Source {
      * @internal
      */
     shouldReloadTile(tile: Tile, {nextBounds, prevIds}: GeoJSONSourceShouldReloadTileOptions) : boolean {
-        if (!tile.latestFeatureIndex) return false;
+        if (!tile.latestFeatureIndex) {
+            return tile.state !== 'unloaded';
+        }
 
         // Update the tile if it PREVIOUSLY contained an updated feature.
         const layers = tile.latestFeatureIndex.loadVTLayers();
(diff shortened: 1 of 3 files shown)


HarelM:
I don't like all the `return this` everywhere.
I've removed it from some places, I should've removed it from here in one of the breaking changes versions.
But I would prefer to add it to the list of stuff to break as part of v6 instead of adding this awkward parameter.
Although this is not a breaking change, so we can add it now and remove it as part of v6 when we decide we have enough things we want to break.
IDK... your call...

wayofthefuture:
When will v6 come out?

HarelM:
There's no actual plans for it at the moment, you can check 
- #6427

Last year we did 2 breaking changes version, this year we still haven't done one.
I don't have a better answer, sorry...

wayofthefuture:
Ok sounds good... we can undo it later it would be pretty easy

src/source/geojson_source.ts
```diff
@@ -266,10 +268,12 @@ export class GeoJSONSource extends Evented implements Source {
      * Updates are applied on a best-effort basis, updating an ID that does not exist will not result in an error.
      *
      * @param diff - The changes that need to be applied.
+     * @param waitForCompletion - If true, the method will return a promise that resolves when the update is complete.
      */
-    updateData(diff: GeoJSONSourceDiff): this {
+    updateData(diff: GeoJSONSourceDiff, waitForCompletion?: boolean): this | Promise<this> {
```
HarelM: Do you think it's possible to improve typings so that when someone sets it to true it will return a promise and otherwise return this?
Without it, while not a breaking change it will cause issue with typescript transpilation.
I think you can use overloads to solve this.

wayofthefuture:
Do overrides look correct?

HarelM:
You should see in the IDE when writing the tests what it interprets.
From a code perspective, I think the overrides are OK if they produce the required results.

HarelM:
I would avoid returning a `Promise<this>` but instead return `Promise<void>`, as this will allow changing this to something more meaningful in the future without a breaking change, and not be stuck with `this`, which is mostly useless.

wayofthefuture:
Good idea!

src/source/geojson_source.ts
```diff
@@ -266,10 +270,14 @@ export class GeoJSONSource extends Evented implements Source {
      * Updates are applied on a best-effort basis, updating an ID that does not exist will not result in an error.
      *
      * @param diff - The changes that need to be applied.
+     * @param waitForCompletion - If true, the method will return a promise that resolves when the update is complete.
      */
-    updateData(diff: GeoJSONSourceDiff): this {
+    updateData(diff: GeoJSONSourceDiff, waitForCompletion: true): Promise<this>;
```
HarelM: `Promise<void>` here as well please.

wayofthefuture: Totally missed that one!

HarelM:
Since this PR is linked to the v6 issue, I hope to remember to remove this flag when releasing v6, but until then, this will stay.

wayofthefuture:
image &nbsp;👀&nbsp;&nbsp;👀&nbsp;&nbsp;👀

Issue #1043, opened 2022-03-03
Feature id can only be positive integer when using feature-state


crooy:
**maplibre-gl-js version**: 2.1.6

**browser**: chrome

### Steps to Trigger Behavior

 1. create a geojson dataset (featureCollection with Features).
 2. The dataset I use is very simple just geopoints in Features with a uuid as id, and that same id in a property 
 3. I then use feature-state to set the "selected" property to true when clicking on the markers, and i reset it when clicking another marker. Using a case statement in the styling from one of the examples i want to change color for the dots i render.
 4. Now while debugging why it doesnt work, i found out the id's were empty. So i read up on the geojson spec, and indeed strings were allowed. I tried integers and it almost worked, only the first marker did not work, the one with 0 as id. 
 5.  I now give all markers positive integers as id and it works perfectly.

### Link to Demonstration

the broken version:
https://stackblitz.com/edit/typescript-kmywnn?file=index.ts

the working version:
https://stackblitz.com/edit/typescript-ese6ag?file=index.ts

note the differences in id string vs int

```
const features = [
 {
      'type': 'Feature',
      'properties': {    
        'posterId': this.id // tried uuids, later integers
      },
      'id': this.id,
      'geometry': this.location
},
...
]
const geodata =  {
    'type': 'FeatureCollection',
    'features': features
  }
...

  mapRef.addSource('my-layer', {
          type: 'geojson',
          data: geodata,
        });
mapRef.on('click', 'my-layer', function (e: MapLayerMouseEvent) {
        const myId = e?.features && e?.features[0]?.properties?.myId; // the uuid i was expecting
        const id = e.features?.[0].id // undefined
})
```

### Expected Behavior
I expected to get the uuid in the id, or the 0-integer.
I also expected the positive integers to work as id-values.

### Actual Behavior
i got undefined for string ids, undefined for 0
But i did get the positive integers back when used as id.

HarelM:
Can you please share a stackblitz/jsbin/code pen of the issue?
The code in this example has typos and missing context...

crooy:
the broken version:
https://stackblitz.com/edit/typescript-kmywnn?file=index.ts

the working version:
https://stackblitz.com/edit/typescript-ese6ag?file=index.ts

note the differences in id string vs int

HarelM:
You set `myId` in the feature and then look for `posterId` in the event? I'm a bit lost here...

crooy:
[...]

PR #6690, opened 2025-11-10
Improve `GeoJSONSource#updateData` support for string feature ids and reduce likelihood of race conditions

lucaswoj:
Fix https://github.com/maplibre/maplibre-gl-js/issues/6694
Ref https://github.com/maplibre/maplibre-gl-js/issues/4364
Ref #6562 

This PR contains two small fixes for `GeoJSONSource#updateData`

 - It casts `string` feature ids to `number`, if possible. Per #1043 support for non-numeric string feature IDs in GeoJSON is limited. However support for_numeric_ string feature IDs was broken for `GeoJSONSource#updateData` primer to this PR. This PR also adds a warning when using `GeoJSONSource#updateData` with non-numeric string feature IDs. 
 - It reduces the likelihood of race conditions when calling `GeoJSONSource#updateData` while the source is in the middle of another update. 

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/source/geojson_source.ts
+++ b/src/source/geojson_source.ts
@@ -452,7 +452,7 @@ export class GeoJSONSource extends Evented implements Source {
 
         for (const id of prevIds.values()) {
             if (typeof id !== 'number') {
-                warnOnce('GeoJSONSource#updateData is slower when using string GeoJSON feature IDs. Consider using numeric IDs for better performance.');
+                warnOnce(`GeoJSONSource "${this.id}": updateData is slower when using string GeoJSON feature IDs (e.g. "${id}"). Consider using numeric IDs for better performance.`);
                 return undefined;
             }
         }


CHANGELOG.md
```diff
@@ -5,14 +5,15 @@
...
+- Fix runtime exception when calling `GeoJSONSource#updateData` while the source is processing another data update ([#6689](https://github.com/maplibre/maplibre-gl-js/pull/6689), [#6690](https://github.com/maplibre/maplibre-gl-js/pull/6690) (by [@lucaswoj](https://github.com/lucaswoj)))
+- Handle numeric string feature IDs passed to `GeoJSONSource#updateData` ([#6690](https://github.com/maplibre/maplibre-gl-js/pull/6690) (by [@lucaswoj](https://github.com/lucaswoj)))
 - _...Add new stuff here..._
 
 ## 5.12.0
 
 ### ✨ Features and improvements
 
-- Add support for MapLibre Tiles (MLT) by using `encoding: 'mlt'` in vector source definition ([#6570](https://github.com/maplibre/maplibre-gl-js/pull/6570)) (by [@Salkin975](https://github.com/Salkin975) and [@HarelM](https://github.com/HArelM)) 
+- Add support for MapLibre Tiles (MLT) by using `encoding: 'mlt'` in vector source definition ([#6570](https://github.com/maplibre/maplibre-gl-js/pull/6570)) (by [@Salkin975](https://github.com/Salkin975) and [@HarelM](https://github.com/HArelM))
```
lucaswoj:
```suggestion
- Add support for MapLibre Tiles (MLT) by using `encoding: 'mlt'` in vector source definition ([#6570](https://github.com/maplibre/maplibre-gl-js/pull/6570)) (by [@Salkin975](https://github.com/Salkin975) and [@HarelM](https://github.com/HArelM)) 
```

src/source/geojson_source.ts
```diff
@@ -448,7 +448,14 @@ export class GeoJSONSource extends Evented implements Source {
...
         const {add = [], update = [], remove = []} = (diff || {});
 
-        const prevIds = new Set([...update.map(u => u.id), ...remove]);
+        const prevIds = new Set([...update.map(u => u.id), ...remove]
+            // Cast all feature ids to numbers. Arbitrary string feature ids are not supported per #1043
+            .map((id) => typeof id === 'number' ? id : parseInt(id, 10))
+        );
+
+        if (prevIds.has(NaN)) {
+            warnOnce('GeoJSONSource#updateData only supports numeric feature IDs or string IDs that can be cast to numbers. See issue #1043 for more details.');
```
HarelM: A link to the issue would be better I believe from a DX perspective.

wayofthefuture:
So we are completely breaking GeoJSON source so that shouldReloadTile works?

HarelM:
@wayofthefuture can you maybe add a test case that you would like to make sure is working after all these fixes/changes?
In theory this test should pass prior to the changes related to geojson optimization, right?

wayofthefuture:
With the shouldReloadTile function, GeoJSON sources will not update all tiles. Polygons will be cut off. So in order for a small performance gain this happens to those of us using strings:


https://github.com/user-attachments/assets/53dddcc6-3690-42af-95d6-ebc046c748a2

HarelM:
I'm guessing this can fallback to previous behavior when IDs are strings, can't it?

lucaswoj:
> I'm guessing this can fallback to previous behavior when IDs are strings, can't it?

Sure. That's possible and a reasonable measure for backwards compatibility. 

Do you still want a warning message in that case? I am wary of officially supporting string IDs here while so many other parts of MapLibre don't support them.

HarelM:
I think the warning should stay, yes.

wayofthefuture:
Would it be possible to use something like https://www.npmjs.com/package/xxhashjs to convert strings to numbers?

```ts
function stringToId(str: string): number {
  return XXH.h32(str, 0xABCD).toNumber();
}
```

Or something like that.

lucaswoj:
> Would it be possible to use something like https://www.npmjs.com/package/xxhashjs to convert strings to numbers?
> 
> 
> 
> ```ts
> 
> function stringToId(str: string): number {
> 
>   return XXH.h32(str, 0xABCD).toNumber();
> 
> }
> 
> ```
> 
> 
> 
> Or something like that.

Unfortunately it's not so easy. String->i32 hashing produces frequent collisions so we'd need to implement a hash collision strategy too. Not impossible, but way beyond the scope of this bugfix.

wayofthefuture:
If hashing isn't possible then we need this:

```ts
shouldReloadTile(tile: Tile, {nextBounds, prevIds}: GeoJSONSourceShouldReloadTileOptions) : boolean {
        if (!tile.latestFeatureIndex) {
            return tile.state !== 'unloaded';
        }
        if (prevIds.has(NaN)) {
            return true;
        }
```

Then nothing should break...

wayofthefuture:
Scratch the prior, this is better (inside of _getShouldReloadTileOptions):

```ts
    if (prevIds.has(NaN)) {
        warnOnce('GeoJSONSource#updateData only supports numeric feature IDs or string IDs that can be cast to numbers. See issue #1043 for more details.');
        return undefined;
    }
```

src/source/geojson_source.ts
lucaswoj: I revised the warning in the latest commit. The linked issue is no longer so relevant so I opted to omit the link entirely.

lucaswoj:
This is ready for another round of review.

 - I have added logic to detect the use of string feature IDs and fall back to the unoptimized behavior. 
 - I looked around and discovered that MapLibre nor `vt-pbf` tries to cast string ids to number ids, so I removed the use of `parseInt` here. Strings are strings and are not supported. 
 - I added one other small bug fix with a regression test: diff updates containing `addOrUpdateProperties` but not `newGeometry` could crash.

src/source/geojson_source.ts
```diff
@@ -450,10 +450,19 @@ export class GeoJSONSource extends Evented implements Source {
 
         const prevIds = new Set([...update.map(u => u.id), ...remove]);
 
+        for (const id of prevIds.values()) {
+            if (typeof id !== 'number') {
+                warnOnce('GeoJSONSource#updateData is slower when using string GeoJSON feature IDs. Consider using numeric IDs for better performance.');
```
HarelM: It might be helpful to add the id to the warning in case someone sees it and is looking to understand which feature might have caused this.

HarelM: If you have information about the source id even better.

HarelM:
I think the warning can be improved a bit to help developers find the problem faster (if they accidentally somehow didn't intent for string IDs).
But otherwise this looks good. 
The above comment is not gating for this PR.

src/source/geojson_source.ts
lucaswoj: Good idea. Done in add286768

0xFA11:
FWIW, I think this might also fix #6131 too but I can't tell for sure

lucaswoj:
> FWIW, I think this might also fix #6131 too but I can't tell for sure

Unfortunately I don't think it will

wayofthefuture:
@0xFA11 possible solutions I am considering:

1. Wait for MLT.
2. Reduce the id string length to about 4-5 characters and hash to numbers on the fly. (Must fit in 32bit)

0xFA11:
@wayofthefuture 
> @0xFA11 possible solutions I am considering:
> 
> 1. Wait for MLT.
> 2. Reduce the id string length to about 4-5 characters and hash to numbers on the fly. (Must fit in 32bit)

I think we're happy to switch from MVT to MLT at some point in the near future, we like it already!

for id string hashing, I saw this comment from @lucaswoj 

> Unfortunately it's not so easy. String->i32 hashing produces frequent collisions so we'd need to implement a hash collision strategy too. Not impossible, but way beyond the scope of this bugfix.

we are actually using https://www.npmjs.com/package/js-xxhash package instead of https://www.npmjs.com/package/xxhashjs and `js-xxhash` is performing a bit better in our case (slightly better performance, better typescript, newer than xxhashjs: 8 years vs 1 year)

having said that, xxhash is a pretty solid non-cryptographic hashing algorithm with a good uniform hash distribution and fast performance (even without SIMD). our workaround for not being able to use string IDs in GeoJson features array was literally to do `data.features.push({ id: xxHash32("some-id-here"), type: "Feature", geometry: { ... }, properties: { ... }, ... });` and we had no issues so far after that. only downside there was to also xxHash32 an ID when we wanted to query or update features based on their IDs, not huge deal.

FWIW, I'd say `js-xxhash` even with strings longer than 4-5 characters wouldn't be too bad and people wouldn't bump into hash collisions left and right. we had ~1 million GeoJson points with unique string IDs (like GUIDs/UUIDs) and xxHash'ing them was perfectly fine, we had no collisions whatsoever.

lucaswoj:
If you hash 1 million string feature IDs to 32-bit integers, you've got about a 100% chance of a collision.

This is a result of the famously counterintuitive birthday paradox. With `n` string feature IDs, the probability of a hash collision on a 32 bit integer can be approximated as:
```
P(collision) ≈ 1 - e^(-n²/(2×2^32))
```

10k features = 1.2% chance
100k features = 69% chance
1 million features = 100% chance

xxHash32 may have excellent distribution, but the math is unavoidable. The collision risk comes from the hash space size, not the hash algorithm quality.

0xFA11:
> If you hash 1 million string feature IDs to 32-bit integers, you've got about a 100% chance of a collision.
> 
> This is a result of the famously counterintuitive birthday paradox. With `n` string feature IDs, the probability of a hash collision on a 32 bit integer can be approximated as:
> 
> ```
> P(collision) ≈ 1 - e^(-n²/(2×2^32))
> ```
> 
> 10k features = 1.2% chance 100k features = 69% chance 1 million features = 100% chance
> 
> xxHash32 may have excellent distribution, but the math is unavoidable. The collision risk comes from the hash space size, not the hash algorithm quality.

after a second thought, I agree that we would (and probably did) actually have hash collisions with xxHash32 over 1 million unique strings. I also don't exactly remember if we had total 1 million features and/or had visible collisions. Now thinking about it, we probably have a bug in our map due to unknown hash collisions, scary! I will go back and dive deep into it and try to fix potential bugs there (funny nobody reported any issues yet, better get them fixed before anyone notices lol)

thank you for correcting me here, I really appreciate it, honestly.

following the same birthday paradox and p(collision) equation, I guess xxHash64 would be safe? a quick napkin math gives me 1/37,000,000 probability of collisions or am I miscalculating it again for xxHash64 for 1 million unique strings?

wayofthefuture:
I guess my ultimate question would be what is the maximum number of lowercase alpha characters that you can uniquely convert to a number that fits in 32bit with no collisions and how to do it?

lucaswoj:
The max feature id allowed by the Mapbox tile spec is only 2^52. So the likelihood of a collision on 1 million features with a 52 bit hash is small but not insignificant:

```
1 - e^(-(1,000,000)^2/(2×2^53)) =0.00555%
```

Careful! If your dataset continues to grow, the probability continues to rise. At 100 million features, the probability is about 43%. 

And careful! This assumes a perfectly random distribution hash values. In reality the values aren't quite randomly distributed, so this number is only a lower bound on the probability. The actual probability could be much higher.

It's up to you if these risks are acceptable for your application.

lucaswoj:
> I guess my ultimate question would be what is the maximum number of lowercase alpha characters that you can uniquely convert to a number that fits in 32bit with no collisions

4

0xFA11:
> The max feature id allowed by the Mapbox tile spec is only 2^52. So the likelihood of a collision on 1 million features with a 52 bit hash is small but not insignificant:
> 
> ```
> 1 - e^(-(1,000,000)^2/(2×2^53)) =0.00555%
> ```
> 
> Careful! If your dataset continues to grow, the probability continues to rise. At 100 million features, the probability is about 43%.
> 
> And careful! This assumes a perfectly random distribution hash values. In reality the values aren't quite randomly distributed, so this number is only a lower bound on the probability. The actual probability could be much higher.
> 
> It's up to you if these risks are acceptable for your application.

we have a top-level 1.2 million cap, it won't grow beyond that but I will do more testing and propose switching from xxhash32 to xxhash64 hash numbers in our specific case. we don't pack data in MVT tiles, we actually set data on map layers/features at runtime progressively while streaming in.

nevertheless, this has been a very good reminder and educational.

PR #6696, opened 2025-11-11
Small fixes in tile manager

HarelM:
## Launch Checklist

Small fixes to code readability, and moved the `_geojsonTileLayer` to a constant.


src/tile/tile_manager.ts
```diff
@@ -687,28 +686,24 @@ export class TileManager extends Evented {
...
-
             if (!idealTile.hasData()) {
-                missingIdealTiles[idealID.key] = idealID;
+                idealTilesWithoutData.add(idealID);
             }
         }
 
-        missingIdealTiles = this._retainLoadedChildren(missingIdealTiles, retain);
+        // retain the tile even if it's not loaded because it's an ideal tile.
+        const retainTileMap: Record<string, OverscaledTileID> = idealTileIDs.reduce((acc, t) => { acc[t.key] = t; return acc;}, {});
```
wayofthefuture, line 700: Maybe move this up, create as empty, and add to it in the previous loop?

```ts
    const idealTilesWithoutData = new Set<OverscaledTileID>();
    const retainTileMap: Record<string, OverscaledTileID> = {};
    for (const idealID of idealTileIDs) {
        const idealTile = this._addTile(idealID);

        if (!idealTile.hasData()) {
            idealTilesWithoutData.add(idealID);
        }

        // retain the tile even if it's not loaded because it's an ideal tile.
        retainTileMap[idealID.key] = idealID;
    }
```

HarelM: I removed it from the loop to keep it separate as I thought I will extract this part of the loop to a method.

HarelM: I did not extract the loop eventually as it's like 4 lines of code, but there's no connection between the two things, so I preferred to keep it "separate".

wayofthefuture: Ok... I like it better the other way but hey! 😜 

You can merge without my approval right?

HarelM: I need someone to approve it. :-)
I also moved the `_geojsonTileLayer` string to a const.

HarelM:
Thanks!

Issue #106, opened 2021-03-19
GeoJSON Source `setData` can take surprisingly long


JannikGM:
When passing a GeoJSON object to `setData`, it uses `JSON.stringify` on the main thread to send data to a worker:

https://github.com/maplibre/maplibre-gl-js/blob/256d1b898f1bd21f829977d4dec602a7f649f5a0/src/source/geojson_source.js#L262-L276

We recently ran into a situation where we generate a GeoJSON with about 200 `LineString` features with about 4500 geometry coordinates each.
Because of a design issue in our software, we use `setData` about 5 times per second.

The `JSON.stringify` for said data will take about 200ms each time. We experimentally did `JSON.parse` on that string (which will happen in the worker) and assume that it also takes about 200ms for said case.
At 5 updates per second, it's impossible for the worker to catch up, and the main thread blocks forever. The website becomes unresponsive.

---

While our software is to blame for the frequent and large updates, it also raises questions about mapbox GeoJSON performance.

Interestingly, generating the GeoJSON is much quicker than the `JSON.stringify`; I don't have exact numbers for this, but eyeballing in Chrome profiler, it's about 1/10th of the time (roughly 20ms I'd assume).

One assumption that we had was that `setData` would be mostly non-blocking and relatively fast. We saw mapbox talking about 100MB GeoJSON sources and dynamic updating of GeoJSON. It's obvious that the combination of this is bad, but we didn't expect it to be this bad - especially not on the main thread.

Here are some thoughts that I had while looking at this:

- Are there any source types which are faster, for locally generated content?
- Could we split data upload into smaller chunks or other ways to mitigate this potential issue on the main thread?
- How could one potentially improve maplibre to allow faster data submission to workers? What about a memory-optimized binary GeoJSON with `Float64Array` coordinates or similar?
- Can we potentially improve GeoJSON parsing in geojson-vt?
- Can we potentially improve cancellation in workers, to cancel an ongoing conversion earlier?
- Should there be a GeoJSON "diff" where one would only update modified features?
- It seems that each source is bound to one specific worker actor; so if that source is busy, it might have effect on other sources that share the same actor?
- Can we round feature location precision to slim the numbers down? We did this naively in a `JSON.stringify`-`replacer`, but it makes stringify even slower (in Chrome at least); we'll try it while generating the features next.

Things we do not want to resort to / things to avoid:

[...]

PR #6702, opened 2025-11-13
GeoJSON Worker Refactor

wayofthefuture:
This PR is intended to be a logic-neutral refactor of the GeoJSON Worker, focusing on readability, clearer structure, and improved long-term maintainability so it's less likely to break over time.

After code reviews, the data object sent from the worker has been changed from a JSON.stringified object to the actual object. This should hopefully provide a performance improvement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 - GeoJSON Source Diff: improve, abstract, refactor, optimize, and fix regressions. ([#6681](https://github.com/maplibre/maplibre-gl-js/pull/6681)) (by [@wayofthefuture](https://github.com/wayofthefuture))
 - GeoJSONSource setData and updateData waitForCompletion option ([#6688](https://github.com/maplibre/maplibre-gl-js/pull/6688)) (by [@wayofthefuture](https://github.com/wayofthefuture))
 - Improve performance of `GeoJSONSource#updateData` ([#6668](https://github.com/maplibre/maplibre-gl-js/pull/6668) (by [@lucaswoj](https://github.com/lucaswoj)))
+- GeoJSON Worker Refactor ([#6702](https://github.com/maplibre/maplibre-gl-js/pull/6702) (by [@wayofthefuture](https://github.com/wayofthefuture))
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes


wayofthefuture:
https://github.com/wayofthefuture/maplibre-gl-js/blob/c157073985836993543385c21d0f7d624facc067/src/source/geojson_worker_source.ts

src/source/geojson_worker_source.ts
```diff
@@ -178,69 +187,108 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
...
+    _filterGeoJSON(data: GeoJSON.GeoJSON, filter: Array<unknown>): GeoJSON.FeatureCollection {
+        const compiled = createExpression(filter, {type: 'boolean', 'property-type': 'data-driven', overridable: false, transition: false} as any);
+
+        if (compiled.result === 'error') {
+            throw new Error(compiled.value.map(err => `${err.key}: ${err.message}`).join(', '));
         }
-        if (!this._dataUpdateable) {
-            throw new Error(`Cannot update existing geojson data in ${params.source}`);
+
+        const features = (data as any).features.filter(feature => compiled.value.evaluate({zoom: 0}, feature));
```
HarelM, line 271: This line assumes data is always FeatureCollection I believe, so I would try and see if this can be typed this way.

src/source/geojson_worker_source.ts
```diff
@@ -267,8 +315,11 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
 }
 
 export function createGeoJSONIndex(data: GeoJSON.GeoJSON, params: LoadGeoJSONParameters): GeoJSONIndex {
-    return params.cluster ? new Supercluster(getSuperclusterOptions(params)).load((data as any).features) :
-        geojsonvt(data, params.geojsonVtOptions);
+    if (params.cluster) {
+        return new Supercluster(getSuperclusterOptions(params)).load((data as any).features);
+    } else {
```
HarelM: else is not needed, otherwise it's not clear what is returned from this method if the else is not reached (from a reader perspective, the compiler knows it can't be reached).

HarelM:
Looks good, I've added some minor comments.

src/source/geojson_worker_source.test.ts
```diff
@@ -111,7 +111,7 @@ describe('resourceTiming', () => {
 
         const layerIndex = new StyleLayerIndex(layers);
         const source = new GeoJSONWorkerSource(actor, layerIndex, []);
-        source.loadGeoJSON = () => Promise.resolve(geoJson);
+        source._loadGeoJSONFromUrl = () => Promise.resolve(geoJson);
```
HarelM: If we can mock `getJSON` instead it would be better, but I'm not sure how easy that would be...

src/source/geojson_worker_source.ts
```diff
@@ -32,20 +32,24 @@ export type GeoJSONWorkerOptions = {
...
  */
 export type LoadGeoJSONParameters = GeoJSONWorkerOptions & {
     type: 'geojson';
+    /**
+     * Request parameters including a URL to fetch GeoJSON data.
+     */
     request?: RequestParameters;
     /**
-     * Literal GeoJSON data. Must be provided if `request.url` is not.
+     * Literal GeoJSON data in string form.
```
HarelM: Why are we passing a string? is it faster?

wayofthefuture: I have no idea.

src/source/geojson_worker_source.ts
```diff
@@ -178,69 +187,108 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
...
      *
      * @param params - the parameters
      * @param abortController - the abort controller that allows aborting this operation
      * @returns a promise that is resolved with the processes GeoJSON
      */
     async loadAndProcessGeoJSON(params: LoadGeoJSONParameters, abortController: AbortController): Promise<GeoJSON.GeoJSON> {
-        let data = await this.loadGeoJSON(params, abortController);
+        let data;
+
+        // Data is loaded from a fetchable URL
```
HarelM: I think moving the comment to be inside the if would be better, as they sit now in a weird place.
Or consider removing them altogether, as they are explicitly specified in the relevant method that is being called, so not sure they are needed.

src/source/geojson_worker_source.ts
wayofthefuture: Hmmm... the linter complains: Argument of type Feature<Geometry, { [name: string]: any; }> is not assignable to parameter of type Feature... when I change the data parameter to GeoJSON.FeatureCollection

src/source/geojson_worker_source.ts
wayofthefuture: lol i learned this from vlodomir in geojson-vt but i'll move

src/source/geojson_worker_source.test.ts
wayofthefuture: I thought it was a refactor?!?!

src/source/geojson_worker_source.ts
HarelM: Proabably the difference between GeoJSON.Feature collection and a list of MaplibreGeoJSONFeature I guess, there are types that are from the vector tiles that are not exactly geojson...
But I think we should fix this `as any` somehow...

src/source/geojson_worker_source.test.ts
HarelM: but this was previously "public" and now "private".
Feel free to keep it as is, although it's not great to override a method of the class you are testing, in terms of what tests should do...

wayofthefuture: Ok let me rename it. I didn't know public rules applied to tests...

HarelM: Just ignore this comment, there's no need to change the name, it will be "the same evil"...

src/source/geojson_worker_source.ts
wayofthefuture: I battled with this, it is a can of worms. I wind up getting pulled into mapbox style spec. You might want to give it a try.. it may be simple for you... you can commit direct to this pr.

src/source/geojson_worker_source.ts
wayofthefuture: If we can pass the diff as an object why can't we pass the data as an object... I think there is an extensive old thread about this somewhere.

wayofthefuture: https://github.com/maplibre/maplibre-gl-js/issues/106

Maybe we could try passing it as an object if you want...

wayofthefuture: Apparently actor is using structured cloning and if we pass the object it is 2x-5x faster for large objects versus the stringify/parse route...

wayofthefuture:
@HarelM Based on your comment I was able to convert the data string to an object! It was much simpler to do after the file was re-factored. Unit tests passed on my computer let's see on the render tests. This would be great news if we could keep it this way!

src/source/geojson_worker_source.ts
HarelM: It might be that the definition of evaluate in the maplibre-style-spec is incorrect as the feature the is being passes to the evaluate method is not a MVT feature, but a geojson feature, and example of a diff is the the type of geojson feature is `Feature` while the type of MVT is related to the geometry type.
So it's a problem in a different place.
You can ignore as part of this PR though.

HarelM: Hu, I think there's a bug here, probably very edge case that no one bothered digging, but definitely a bug.

wayofthefuture: Ok thanks for checking into it

HarelM: I wonder what will happen if you create a geojson of points and lines and then try to filter only the points.
Although this doesn't make a lot of sense, since you can simply filter the geojson prior to using it in a source, I think this will showcase the bug here that the feature passed into the evaluation is not a MVT type of feature but rather geojson feature, and they don't match.

Another probably more edge case is to filter the geojson by distance from something.

In any case, while I haven't tested my theory, this looks like a bug.

HarelM:
Since you changed the geojson from string to object I think it's worth adding a changelog entry.
BTW, do you have a comparison of how much time it took before all the optimizations vs now (just curious)?

HarelM (approved):

wayofthefuture:
No but I definitely plan on doing a comparison!

PR #6712, opened 2025-11-14
Fix queryRenderedFeatures with terrain

### LINKED ISSUE #6563: Terrain can cause queryRenderedFeatures to not return all results at high zoom levels
When zoomed in far enough, the queryRenderedFeatures() function does not return all results if terrain is enabled. 

**maplibre-gl-js version**: 5.9.0

**browser**: Chrome 141.0.7390.67

### Steps to Trigger Behavior

 1. Add circles layer to the map
 2. Zoom in at least to around 14 or so, with points visible
 3. See results of queryRenderedFeatures() for the layer
 4. Turn on terrain
 5. See results of queryRenderedFeatures() for the layer has changed, sometimes drastically, from what was returned from 3 above, even if all the points are still visible

### Link to Demonstration
https://jsbin.com/gehedaluvi/1/edit?html,output

This should already be zoomed and centered to a handful of points. Hit the terrain toggle button to see the rendered feature count change.

### Expected Behavior
The queryRenderedFeatures() function should return all visible results when terrain is enabled

### Actual Behavior
The queryRenderedFeatures() function does not always return all visible results when terrain is enabled

lucaswoj:
fix #6563 

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/style/style_layer/circle_style_layer.ts
+++ b/src/style/style_layer/circle_style_layer.ts
@@ -66,13 +66,17 @@ export class CircleStyleLayer extends StyleLayer {
         let transformedPolygon: Array<Point>;
         let transformedSize: number;
         if (pitchAlignment === 'map') {
+            // TODO I have confirmed that this codepath is correct
             transformedPolygon = translatedPolygon;
             transformedSize = size * pixelsToTileUnits;
         } else {
+            // TODO I suspect this codepath is INCORRECT or BROKEN.
+            // TODO Investigate the "cannot calculate elevation if elevation maxzoom > source.maxzoom" warnings
             transformedPolygon = projectQueryGeometry(translatedPolygon, transform, unwrappedTileID, getElevation);
             transformedSize = size;
         }
 
+        // TODO I have confirmed that circleIntersection is correct.
         return circleIntersection({
             queryGeometry: transformedPolygon,
             size: transformedSize,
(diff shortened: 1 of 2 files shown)


lucaswoj:
Screenshot 2025-11-14 at 3 47 01 PM

I have determined that the `translatedPolygon` (in red above) being passed to `projectQueryGeometry` is correct but the returned value of `transformedPolygon` (in blue above) is incorrect. https://github.com/maplibre/maplibre-gl-js/blob/main/src/style/style_layer/circle_style_layer.ts#L72

Internally, it appears that we're passing some bad values to `getElevation` (e.g. large negative X coordinates) and getting bad values back (e.g. 0 when the elevation should be ~1000). 

I wonder if we're getting tripped up on the mismatch between the maxzoom of the circle tiles vs the terrain tiles?

lucaswoj:
Before https://github.com/maplibre/maplibre-gl-js/pull/5599 we effectively returned every circle feature in every loaded tile, so running `queryRenderedFeatures` on the whole viewport seemed to work. If you had run it on a specific rectangle within the viewport, you would've noticed that it wasn't returning the expected results. That's all to say, the fix goes much deeper than #5599, and touches on some areas where the 3D terrain implementation seems to have never worked properly.

HarelM:
@lucaswoj what's the status of this PR? Do you want to push it forward or close it?

lucaswoj:
Closing

PR #6715, opened 2025-11-15
Consolidate Tile Caches to Lightweight Map-based LRU Cache

wayofthefuture:
Since ES6, Map has made simple in-memory caching far easier. It provides predictable iteration, avoids key pitfalls, and gives fast O(1) access, ideal for MapLibre's tile-cache needs. The previous tile cache implemented viewport based sizing in order to keep O(n) operations to a minimum, however, with a newer O(1) based cache, this is no longer an issue as the cache can grow in size without slowing it's performance. This allows us to keep many more tiles in the cache and will improve user experience.

I reviewed several npm caching libraries, but none fit well:
- Many relied on pre-ES6 object-based patterns with unnecessary complexity.
- Some were fast but lacked strong TypeScript typing.
- Others used Map but bundled far more features than we need.
- Several showed low adoption or minimal maintenance.

This PR introduces a lightweight cache purpose-built for MapLibre, with the following improvements:

## O(N) Comparison Table

| Operation | Complexity (v1 → v2) | Potential Speedup |
|----------|------------------------|--------------------|
| **Get / Has** | O(1) → O(1) | ~1× |
| **Set (not full)** | O(1) → O(1) | ~1× |
| **Set (with LRU eviction)** | **O(N)** → **O(1)** | **Up to N× faster** |
| **Remove (by key)** | **O(N)** → **O(1)** | **Up to N× faster** |
| **Filter** | **O(N²)** → **O(N)** | **Up to N× faster** |
| **Resize (setMaxSize)** | **O(N²)** → **O(N)** | **Up to N× faster** |
| **Reset / Clear** | O(N) → O(N) | ~1× |

## Timers Comparison Table

| Aspect | v1 (Timer-based cache) | v2 (LRU cache, no timers) |
|--------|--------------------------|----------------------------|
| **Timers per tile** | 1 per entry (hundreds–thousands) | 0 |
| **Event-loop overhead** | High (timer scheduling + callbacks) | None |
| **GC pressure** | High (many closures + timer objects) | Low |
| **Callback churn** | Thousands of callbacks | None |
| **Memory per tile** | Large (timeout + closure) | Small (Map entry only) |
| **Latency/jank risk** | High | Low |

### DIFF
--- a/src/tile/bounded_lru_cache.test.ts
+++ b/src/tile/bounded_lru_cache.test.ts
@@ -65,7 +65,7 @@ describe('BoundedLRUCache', () => {
         expect(removeCount).toBe(1);
     });
 
-    test('set of tile id using the same tile retains tile and moves to end without calling remove', () => {
+    test('set of same tile id using the same tile retains tile and moves to end without calling remove', () => {
         let removeCount = 0;
         const cache = new BoundedLRUCache<string, Tile>({
             maxEntries: 10,
@@ -84,7 +84,7 @@ describe('BoundedLRUCache', () => {
         expect(cache.get(idA)).toBe(tileA);
     });
 
-    test('set of same tile id using a new tile remove previous tile reference and calls remove', () => {
+    test('set of same tile id using a new tile removes previous tile reference and calls remove', () => {
         let removeCount = 0;
         const cache = new BoundedLRUCache<string, Tile>({
             maxEntries: 10,
@@ -151,7 +151,7 @@ describe('BoundedLRUCache', () => {
         expect(() => cache.remove(idB)).not.toThrow();
     });
 
-    test('removes the oldest entry', () => {
+    test('removeOldest removes the oldest entry', () => {
         const cache = new BoundedLRUCache<string, Tile>({
             maxEntries: 10,
             onRemove: () => {}
--- a/src/tile/bounded_lru_cache.ts
+++ b/src/tile/bounded_lru_cache.ts
@@ -27,8 +27,9 @@ export class BoundedLRUCache<K, V> {
     }
 
     /**
-     * Adds or updates a value in the cache. If the key already exists, it removes the old entry first.
-     * If adding would exceed the maximum size, it removes the oldest entry before adding.
+     * Adds or updates a value in the cache. If the key already exists, it removes the old entry only
+     * if the new value is a different object reference, otherwise it moves the existing entry to the end.
+     * If adding exceeds the maximum allowed cache size, it removes the oldest entry before adding.
      */
     set(key: K, value: V) {
         const existing = this.map.get(key);
@@ -52,7 +53,7 @@ export class BoundedLRUCache<K, V> {
     }
 
     /**
-     * Moves a key to the end of the cache without removing it (most recently used).
+     * Moves an entry to the end of the cache without removing it (most recently used).
      */
     moveToEnd(key: K, value: V) {
         this.map.delete(key);
@@ -70,7 +71,7 @@ export class BoundedLRUCache<K, V> {
     }
 
     /**
-     * Removes entries from the cache until the number of entries is within maxEntries.
+     * Removes entries from the cache until the number of entries is within the limit.
      */
     trim() {
         while (this.map.size > this.maxEntries) {
@@ -80,7 +81,7 @@ export class BoundedLRUCache<K, V> {
 
     /**
      * Removes the least recently used entry from the cache.
-     * The first key is the least recently used since get/set above places entries at the end.
+     * The first key is the least recently used since map get/set above places entries at the end.
      */
     removeOldest() {
         const iterator = this.map.keys().next();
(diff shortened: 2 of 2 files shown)


HarelM:
Can you clarify how this cache is better than the previous one that was implemented? 
Does it behave differently in any way (I saw the 1000 entries limit and I was wondering if that was the case with previous cache as well)?

wayofthefuture:
> Can you clarify how this cache is better than the previous one that was implemented? Does it behave differently in any way (I saw the 1000 entries limit and I was wondering if that was the case with previous cache as well)?

I added comparisons into the PR text at the top. Map options have been updated so both tile manager and overzooming vector tiles can have custom set max cache numbers. We previously used the viewport to restrict the cache size but I believe that is no longer needed considering we can maintain a much larger cache at high performance using O(1) Map operations. See comparison above...

src/tile/tile_manager.ts
```diff
@@ -923,18 +902,8 @@ export class TileManager extends Evented {
...
-
-            // set timer for the reloading of the tile upon expiration
-            this._setTileReloadTimer(tileID.key, tile);
-
-            // set the tileID because the cached tile could have had a different wrap value
-            tile.tileID = tileID;
-            this._state.initializeTileState(tile, this.map ? this.map.painter : null);
-        }
+        // Check the cache for the tile first
+        tile = this._getTileFromCache(tileID);
```
HarelM, line 906: Will this cause a change in how the events were triggered? I see some logic around event firing if a tile was not in the cache...

src/tile/tile_manager.ts
```diff
@@ -952,6 +921,47 @@ export class TileManager extends Evented {
...
+
+        return tile;
+    }
+
+    /**
+     * Add a tile to the cache using a wrapped tileID (wrap = 0)
+     */
+    _addTileToCache(tile: Tile) {
+        const cacheKey = this._getCacheKey(tile.tileID);
+        this._cache.set(cacheKey, tile);
```
HarelM, line 955: This will have the same pointer to a tile as before, I saw some logic around it just wanted to make sure you took that into consideration.

src/tile/tile_manager.ts
```diff
@@ -1006,18 +1016,17 @@ export class TileManager extends Evented {
...
         this._clearTileReloadTimer(id);
 
-        if (tile.uses > 0)
-            return;
+        if (tile.uses > 0) return;
 
         if (tile.hasData() && tile.state !== 'reloading') {
-            this._cache.add(tile.tileID, tile, tile.getExpiryTimeout());
+            // Cache the removed tile
+            this._addTileToCache(tile);
```
HarelM, line 1029: Should this operation be considered as "usage" in terms of LRU? Should the tile removed from the view here should be at the beggining or end of the LRU stack?

src/ui/map.ts
```diff
@@ -250,15 +250,15 @@ export type MapOptions = {
...
-     * The maximum number of tiles stored in the tile cache for a given source. If omitted, the cache will be dynamically sized based on the current viewport which can be set using `maxTileCacheZoomLevels` constructor options.
-     * @defaultValue null
+     * The maximum number of tiles that can be stored in the tile cache for a given source.
+     * @defaultValue 1000
      */
     maxTileCacheSize?: number | null;
     /**
-     * The maximum number of zoom levels for which to store tiles for a given source. Tile cache dynamic size is calculated by multiplying `maxTileCacheZoomLevels` with the approximate number of tiles in the viewport for a given source.
-     * @defaultValue 5
+     * The maximum number of tiles that can be stored in the overzoomed vector tile cache for a given source past the maximum zoom level of the source.
```
HarelM, line 258: This s looks like a breaking change... I would like to avoid those if possible...

HarelM:
Thanks for the update in the initial post, I'm still struggling to understand if there's a logical change in what is kept in the cache and what is removed, and when, if there's no change in that logic and this is only performance optimization, than great, but I have a feeling this is not the case...

wayofthefuture:
Definitely not the case and still thinking about it... in progress

wayofthefuture:
Previously, when a tile is used from the cache, it is removed from the cache: addTile calls getAndRemove(). When the map is moved, on update, we have this.updateCacheSize(transform), which then trims the cache. So this is constant removal O(n) operations.

But we can't have it both ways. We can't trim with the viewport and have LRU at the same time. So it's more of a viewport cache than it is a tile cache. And I don't know the consequences of switching to a full LRU cache. So maybe the best path forward is to refine this current PR to be less LRU and more viewport-based but with O(1).

So simply remove the LRU portion of this PR and we have an optimized cache. The only major change after that is the removal of timers from the cache, because this can be handled upon retrieval, which I believe to be a much better method.

wayofthefuture:
Please ignore previous comment...

wayofthefuture:
Closing this will re-open with a different variation...

Issue #6716, opened 2025-11-15
Return to north button actually does much more, so needs correct tooltip


jidanni:
https://github.com/osm-americana/openstreetmap-americana/issues/1294#issuecomment-3536079915

People think that this is the simple Google "return to north up" button of Google Maps.

Without a full tooltip, no user will guess its full functionality.

HarelM:
Can you clarify the "steps to reproduce" and better define a problem statement? 
I read the linked issue and did not understand the reported issue completely...

PR #6717, opened 2025-11-15
Update NavigationControl.ResetBearing message

jidanni:
For #6716. Else users will never guess they are able to rotate the map! They would think all they can do is reset rotated maps only.

### DIFF
--- a/src/ui/default_locale.ts
+++ b/src/ui/default_locale.ts
@@ -8,7 +8,7 @@ export const defaultLocale = {
     'LogoControl.Title': 'MapLibre logo',
     'Map.Title': 'Map',
     'Marker.Title': 'Map marker',
-    'NavigationControl.ResetBearing': 'Reset bearing to north',
+    'NavigationControl.ResetBearing': 'Drag to rotate map. Click to reset.',
     'NavigationControl.ZoomIn': 'Zoom in',
     'NavigationControl.ZoomOut': 'Zoom out',
     'Popup.Close': 'Close popup',


HarelM:
I think it's ok that it's not obvious that you can rotate the map by dragging on the button, it's not a great UX anyway due to how small the button is...
If anything the click should be first...

jidanni:
All I know is I saw this map,
https://github.com/osm-americana/openstreetmap-americana/issues/846#issuecomment-1492560260
and like a ****@@@@**** had to ask around how they did it,
making a ****@@@@**** out of myself. Even after checking every button in the interface.

HarelM:
While I understand your pain, as I said, I read the linked issue and still don't understand the whole story, I suggest writing it more elaborately.
Really defining what the problem statement is, and if changing a tool tip text will really solves it...

jidanni:
On Google Maps mobile, there is the same reset to north button.

But that's all it does, resets to north.

So you have the same reset to north button.

And that's what you label it as.

But it does much more.

Therefore the label should say so.

Or,

if the map is already north,

then the label reset to north makes no sense.

Therefore the label needs to become a second label:

rotate map.


But then just clicking it isn't enough to rotate the map.

so it should say drag to rotate map.

HarelM:
You are still in the solution domain, I need you to step back please and describe the problem. 
Comparing to Google maps is not helpful at this point when we are trying to better define the problem statement.

jidanni:
I am saying the button declares the only thing it can do is return to north.

It even says that after it has returned to north, further cementing the case that that is all that it can do, in the eyes of the user.

How is the user supposed to "magically" know it can do more things?

Accidentally tug on it one day?

HarelM:
Again, not a problem definition. You keep ignoring my request...

jidanni:
OK then there is no problem.

PR #6731, opened 2025-11-17
Improve, optimize, and consolidate Tile LRU cache to new ES6 Map

wayofthefuture:
Since ES6, Map has made simple in-memory caching far easier, providing fast average-case O(1) access. Our previous tile cache relied heavily on O(n) array operations. This PR replaces that implementation with the existing BoundedLRUCache and wires it into TileCache. I evaluated several npm caching libraries, but none fit well with the specifics of MapLibre’s tile loading model or offered a good balance of size and simplicity.

The most critical operation used in TileManager, getAndRemove, previously incurred three O(n) operations per call. It is now implemented as take, which performs the same behavior in O(1). Additionally, the old tile cache maintained many concurrent timers for tiles sitting idle in the cache. The new cache no longer tracks per-tile timers; instead, expired tiles are evicted/unloaded lazily on access, avoiding the overhead of a large number of active timers.

## O(N) Comparison Table

| Operation | Complexity (v1 → v2) | Potential Speedup |
|----------|------------------------|--------------------|
| **Get** | O(1) → O(1) | ~1× |
| **Set** | O(1) → O(1) | ~1× |
| **getAndRemove / Take** | **O(N)** → **O(1)** | **Up to N× faster** |
| **Remove (by key)** | **O(N)** → **O(1)** | **Up to N× faster** |
| **Filter** | **O(N²)** → **O(N)** | **Up to N× faster** |
| **Resize (setMaxSize)** | **O(N²)** → **O(N)** | **Up to N× faster** |
| **Reset / Clear** | O(N) → O(N) | ~1× |

## Timers Comparison Table

| Aspect | v1 (Timer-based cache) | v2 (LRU cache, no timers) |
|--------|--------------------------|----------------------------|
| **Timers per tile** | 1 per entry (hundreds–thousands) | 0 |
| **Event-loop overhead** | High (timer scheduling + callbacks) | None |
| **GC pressure** | High (many closures + timer objects) | Low |
| **Callback churn** | Thousands of callbacks | None |
| **Memory per tile** | Large (timeout + closure) | Small (Map entry only) |
| **Latency/jank risk** | High | Low |

### DIFF
--- a/src/tile/bounded_lru_cache.ts
+++ b/src/tile/bounded_lru_cache.ts
@@ -18,7 +18,7 @@ export class BoundedLRUCache<K, V> {
     /**
      * Retrieves a value from the cache and marks it as most recently used by moving it to the end.
      */
-    get(key: K): V | undefined {
+    public get(key: K): V | undefined {
         const value = this.map.get(key);
         if (value === undefined) return undefined;
         this.moveToEnd(key, value);
@@ -28,7 +28,7 @@ export class BoundedLRUCache<K, V> {
     /**
      * Retrieves a value from the cache and removes it without triggering the onRemove callback.
      */
-    take(key: K): V | undefined {
+    public take(key: K): V | undefined {
         const value = this.map.get(key);
         if (value === undefined) return undefined;
         this.map.delete(key);
@@ -40,7 +40,7 @@ export class BoundedLRUCache<K, V> {
      * if the new value is a different object reference, otherwise it moves the existing entry to the end.
      * If adding exceeds the maximum allowed cache size, it removes the oldest entry before adding.
      */
-    set(key: K, value: V) {
+    public set(key: K, value: V) {
         const existing = this.map.get(key);
 
         // Adding a new entry
@@ -73,7 +73,7 @@ export class BoundedLRUCache<K, V> {
      * Updates the maximum number of entries allowed in the cache.
      * If the new size is smaller than the current number of entries, removes oldest entries until within limit.
      */
-    setMaxSize(maxEntries: number) {
+    public setMaxSize(maxEntries: number) {
         if (maxEntries === this.maxEntries) return;
         this.maxEntries = maxEntries;
         this.trim();
@@ -101,7 +101,7 @@ export class BoundedLRUCache<K, V> {
     /**
      * Removes a specific entry from the cache and triggers the onRemove callback if configured.
      */
-    remove(key: K) {
+    public remove(key: K) {
         const value = this.map.get(key);
         if (value === undefined) return;
         this.map.delete(key);
@@ -111,7 +111,7 @@ export class BoundedLRUCache<K, V> {
     /**
      * Removes entries from the cache that don't satisfy the provided filter function.
      */
-    filter(func: (value: V) => boolean) {
+    public filter(func: (value: V) => boolean) {
         for (const [key, value] of this.map.entries()) {
             if (func(value)) continue;
             this.remove(key);
@@ -121,7 +121,7 @@ export class BoundedLRUCache<K, V> {
     /**
      * Removes all entries from the cache. If an onRemove callback is configured, it will be called for each entry.
      */
-    clear() {
+    public clear() {
         // If onRemove callback is not configured, just clear the map directly.
         if (!this.onRemove) {
             this.map.clear();
@@ -137,14 +137,14 @@ export class BoundedLRUCache<K, V> {
     /**
      * Returns an array of all keys currently in the cache in their current order.
      */
-    getKeys(): K[] {
+    public getKeys(): K[] {
         return Array.from(this.map.keys());
     }
 
     /**
      * Returns the maximum number of entries allowed in the cache.
      */
-    getMaxEntries(): number {
+    public getMaxEntries(): number {
         return this.maxEntries;
     }
 }


src/tile/bounded_lru_cache.ts
```diff
@@ -0,0 +1,150 @@
...
+        }
+
+        // Updating exact same tile reference - move to the end (most recently used)
+        if (existing === value) {
+            this.moveToEnd(key, value);
+            return;
+        }
+
+        // Updating same key with a new tile - remove old tile and add new one
+        this.remove(key);
```
HarelM, line 60: Should this call onRemove? I'm honestly asking, I don't know the right answer.

wayofthefuture: I believe so:

https://github.com/maplibre/maplibre-gl-js/blob/main/src/tile/tile_cache.ts#L82-L85

Eventually on remove would be called for the same id.

wayofthefuture: It might be better to not call onRemove at all here and let the GC do its thing

wayofthefuture: onRemove propagates back to source.unloadTile which then goes back to the worker_source.removeTile. In here, `loaded` is used to store tiles by uid in the worker. uid is created in tile.js using uniqueId() which is an integer count++ on each function run, so every tile object has a new and unique uid. Even though the new tile may have the same tileID, it will absolutely have a unique uid, which means it still needs to be removed from the this.loaded cache in the worker. In other words, onRemove needs to be called otherwise the tiles would collect in the worker. Do some tracing and let me know if you agree...

HarelM: Ok, I think I know what should be the right approach here in order to get confident in the current solution.
1. We need to cover the current cache with tests so that we know exactly how it operates, which callbacks are called when and how it behaves in general (assuming this is not the tests that currently exists, if the tests cover the current functionality then we can skip this step).
2. After we have the tests in place, we can simply replace the implementation under the hood - no changes to logic (besides maybe the timers), without even changing the public API (take vs getAndRemove, clear vs reset etc)
3. After this is passing and we know that we haven't changed any logic, we can rename stuff.

I think this is the safest way to handle this.
If this is how you did it, then there's very low risk in introducing bugs, as the process above is fairly safe.

HarelM: I've read the cache code start to end and a few questions rose:
1. Why does it have an array? Most of the usage is referring to the first element, but not all. 
2. If the timeout were to be removed, how will this affect the code that listens to onRemove? What will happen if the tile's onRemove is not called when the tile expiries? 
3. The functionally as covered in the current tests (not those you added, but those that exist) is far from ideal, there's even a comment about how the cache should behave but it's not covered in a test, only in a comment, so not a great start...

wayofthefuture: Keep in mind that even though it says this._cache, I would think of it as this._viewportCache. It isn’t really a large tile cache it’s just a small cache for viewport rendering. It can't be that large because of the `filter` function which gets called when changing glyphs/symbols, so the cache size carries maintenance overhead.

While I am not completely certain:

1. The array appears to either be an optimization using old objects to keep a separate order array, and/or it was also needed to maintain older tile references to access in onRemove via the timers.
2. Tile expiration is just a timestamp, so when this time lapses the tile just sits in the cache. When it is removed via `take`, the time is checked and if expired it is discarded and unloaded in `_takeTileFromCache`.
3. I think it's simply a different functionality. The newer cache is simpler because of Map and operates on the assumption that expired tiles can remain in the cache and evicted/unloaded when they are retrieved or the cache needs the space.

If you read the 3rd line here https://www.npmjs.com/package/lru-cache it also supports the idea that a cache should be either TTL or LRU, not both.

HarelM: I need to find a day or two without interruptions to look at the tile manager code thoroughly and see how to break it down, I hope I'll be able to find the time next week, sorry for dragging this.
The above statement is not relevant to changes you are doing, as I'll keep on reviewing them and provide feedback (as best I can, the code is too complicated without a good reason...).

wayofthefuture: If you are good with keeping both 'raster' and 'vector' handled in tile_manager, then we could extract all retainment logic into tile_retainment.ts... versus having separate managers. This would cut out about 30% of the file. Just another idea...

HarelM: I think you'll need to pass "this" into this file, which is an anti-pattern.
Also regarding the above name I think there are simply two "storage" types: in-view-tiles and out-of-view-tiles, the current cache us the out of view as far as I understand. 
But I want to better understand other parts as well.

wayofthefuture: I think you can just pass this._tiles and retain, and change getLoadedTile to tile.hasData(), then you don't need to pass `this`...

Current cache is tiles that were loaded for a previous view... this._tiles is the current view. As tiles move out of view they are added to the cache...

Covering tiles -> add tile -> check the cache -> remove other this._tiles that aren't covering/substitute -> place them into cache -> repeat...

HarelM: Yeah, this is also my understanding. 
I wonder why there is a usage counter for a tile though...
In any case, a first step can be to move these parts to separate files as you suggested, onne for raster and one for vector.
Also renaming _tiles to inViewTiles and cache to outOfViewCache so that is clear why these two variables are needed and how they are different.

wayofthefuture: 'uses' to me seems like a hack that should be deprecated... if it doesn't work without it then we are missing something...

HarelM:
An initial stab at moving some code around. Not a big change overall...
- https://github.com/maplibre/maplibre-gl-js/pull/6756

wayofthefuture:
Well I resolved the merge conflicts and merged your split-tile-manager branch but now the 'files changed' page looks polluted. I guess after you merge your branch into main and I update this one it will be easier to read the changes...

HarelM:
I've merged it to main, so you'd have to resolve conflicts again, sorry...

wayofthefuture:
Got it merged... but still needs a little thought regarding remove/onRemove...

src/tile/tile_manager.ts
```diff
@@ -820,7 +853,8 @@ export class TileManager extends Evented {
             this._removeTile(id);
         }
 
-        this._outOfViewCache.reset();
+        // Clear the tile cache
```
HarelM, line 856: I don't think this comment is needed...

src/tile/tile_manager.ts
```diff
@@ -263,7 +266,8 @@ export class TileManager extends Evented {
             return;
         }
 
-        this._outOfViewCache.reset();
+        // Clear the tile cache
```
HarelM, line 269: Not needed...

HarelM:
I still think adding tests to current implementation is necessary to make sure we don't miss on a use case, for example maybe the cache array is used to store different wrap instances of a tile on the same key? IDK...

wayofthefuture:
Ok.

1. Different wrap instance of tile on same key.
2. Tile is actually expired on removal and not used (done).
3. Any others?

HarelM:
Current implementation, as in the old one, I know I don't fully understand the old behavior, so it's important to make sure we understand it before replacing it.

PR #6738, opened 2025-11-18
Make GeoJSONSource#setData faster

lucaswoj:
This PR brings some of the performance improvements we've added to `GeoJSONSource#updateData` in #4364 to the `GeoJSONSource#setData` method.

Specifically, it this PR removes the logic that passes the full serialized GeoJSON data from the worker thread to the main thread. In [my benchmark, adding 100k points to a map](https://jsbin.com/xahilipuye/edit?html,output), this makes `setData` about 22% faster vs main.

This changes some undocumented behavior of `Map#getStyle`. Prior to this PR, if you add a GeoJSON source with a data URL, wait for it to load, and then call `Map#getStyle`, you'd get the actual loaded GeoJSON features back, not the original source spec with the URL. Similarly, if you called `setData` with some improperly wound GeoJSON geometries, and then called `Map#getStyle`, you'd get the rewound geometries back. Even though this is undocumented behavior, I suspect at least a couple folks are relying on it. They will need to use the `GeoJSONSource#getData` method instead.

## Launch Checklist

 - [x] Link to related issues.
 - [x] ~Write tests for all new functionality.~
 - [x] Post benchmark scores.

### DIFF
--- a/src/source/geojson_source.test.ts
+++ b/src/source/geojson_source.test.ts
@@ -786,6 +786,24 @@ describe('GeoJSONSource.updateData', () => {
         const result = source.updateData({add: []} as GeoJSONSourceDiff, true);
         expect(result).toBeInstanceOf(Promise);
     });
+
+    test('throws error when updating data that is not compatible with updateData', async () => {
+        const initialData: GeoJSON.FeatureCollection = {
+            type: 'FeatureCollection',
+            features: [
+                {type: 'Feature', properties: {}, geometry: {type: 'Point', coordinates: [0, 0]}},
+            ]
+        };
+
+        const source = new GeoJSONSource('id', {data: initialData} as GeoJSONSourceOptions, mockDispatcher, undefined);
+        source.load();
+        await waitForEvent(source, 'data', (e: MapSourceDataEvent) => e.sourceDataType === 'metadata');
+
+        const errorPromise = waitForEvent(source, 'error', () => true);
+        source.updateData({add: [{type: 'Feature', id: 1, properties: {}, geometry: {type: 'Point', coordinates: [1, 1]}}]});
+        const error = await errorPromise;
+        expect(error.error.message).toBe('GeoJSONSource "id": GeoJSON data is not compatible with updateData');
+    });
 });
 
 describe('GeoJSONSource.getBounds', () => {


src/source/geojson_source.ts
```diff
@@ -397,7 +397,7 @@ export class GeoJSONSource extends Evented implements Source {
         const {data, diff} = this._pendingWorkerUpdate;
 
         const options: LoadGeoJSONParameters = extend({type: this.type}, this.workerOptions);
-        if (data) {
+        if (data !== undefined) {
```
lucaswoj, line 400: We need to explicitly check against `undefined` because `null` is a valid GeoJSON value.

src/source/geojson_source.ts
```diff
@@ -488,7 +480,7 @@ export class GeoJSONSource extends Evented implements Source {
 
         for (const id of prevIds.values()) {
             if (typeof id !== 'number' && this.promoteId == null) {
-                warnOnce(`GeoJSONSource "${this.id}": updateData is slower when using string GeoJSON feature IDs (e.g. "${id}"). Consider using promoteId or numeric IDs for better performance.`);
+                warnOnce(`GeoJSONSource "${this.id}": updateData is slower when using string GeoJSON feature IDs. Consider using promoteId or numeric IDs for better performance.`);
```
lucaswoj, line 483: Providing an example ID in this warning message can cause log spew. `warnOnce` needs the warning message to be exactly the same across calls but the example ID could change in every call.

src/source/geojson_source_diff.ts
```diff
@@ -73,6 +73,11 @@ export function isUpdateableGeoJSON(data: GeoJSON.GeoJSON | undefined, promoteId
         return true;
     }
 
+    // {} can be updated
+    if (data.type == null) {
+        return true;
+    }
```
lucaswoj, line 79: We have a couple dozen unit tests that pass `{}` as a GeoJSON data value. This isn't technically valid per the spec but I chose to allow the value rather than change all the tests.

src/source/geojson_source.ts
wayofthefuture: Agreed.

wayofthefuture:
Looks good to me!

HarelM (approved):

PR #6745, opened 2025-11-20
Line offset duplicates

mmc1718:
These changes update the offsetLine function in query_utils used to offset line geometries as part of querying whether a given linestring intersects a point or bounding box/polygon.

In cases of linestrings containing duplicate consecutive points:

offsetLine removes duplicate consecutive values from the line before performing the offset
In the case of no change across multiple points, points are skipped
This ensures offsetLines do not contain null values.

Fixes https://github.com/maplibre/maplibre-gl-js/issues/5431

## Launch Checklist

 - [x] Link to related issues.

### DIFF
--- a/src/style/style_layer/line_style_layer.test.ts
+++ b/src/style/style_layer/line_style_layer.test.ts
@@ -48,5 +48,4 @@ describe('LineStyleLayer', () => {
         lineLayer.setPaintProperty('line-gradient', null);
         expect(lineLayer.gradientVersion).toBeGreaterThan(gradientVersion);
     });
-
 });


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
+- Handle consecutive duplicate points in offsetLine to avoid null values in the output. Fixes [#5431](https://github.com/maplibre/maplibre-gl-js/issues/5431)
```
```suggestion
- Handle consecutive duplicate points in offsetLine to avoid null values in the output. Fixes [#5431](https://github.com/maplibre/maplibre-gl-js/issues/5431) (by [mmc1718](https://github.com/mmc1718))
```

src/style/style_layer/line_style_layer.test.ts
```diff
@@ -48,4 +52,75 @@ describe('LineStyleLayer', () => {
         lineLayer.setPaintProperty('line-gradient', null);
         expect(lineLayer.gradientVersion).toBeGreaterThan(gradientVersion);
     });
+
+    describe('queryIntersectsFeature', () => {
```
HarelM: Aren't these tests covered by the query test that use the json files?
Not that I mind having more tests, but I'm wondering if there are really needed...

HarelM:
Looks good overall, thanks!!
Just a small last comment to address before this can be merged.

src/style/style_layer/line_style_layer.test.ts
mmc1718: I guess there's overlap, this is testing the queryIntersectsFeature method specifically for LineStyleLayers whereas the integration test is calling map.queryRenderedFeatures which should then call the queryIntersectsFeature method.

If we want to avoid extra tests I would probably lean towards taking out the integration test?

HarelM: The integration tests can be used for native as well, if needed, I believe, maybe. 
Since there are unit tests I'm leaning towards removing those instead if the json tests...
I don't have strong feeling either way, your call.

mmc1718: I've updated this, removed the unit tests here in favor of keeping the integration test

src/style/style_layer/line_style_layer.test.ts
```diff
@@ -1,7 +1,11 @@
-import {describe, test, expect} from 'vitest';
+import {describe, test, expect, vi, beforeEach} from 'vitest';
 import {createStyleLayer} from '../create_style_layer';
 import {extend} from '../../util/util';
 import {type LineStyleLayer} from './line_style_layer';
+import {type QueryIntersectsFeatureParams} from '../style_layer';
```
HarelM: Does these addition require reverting as well? I'm not sure they are used in this file, aren't they?

mmc1718: Thanks for spotting this! Updated now

src/style/style_layer/line_style_layer.test.ts
```diff
@@ -48,4 +48,5 @@ describe('LineStyleLayer', () => {
         lineLayer.setPaintProperty('line-gradient', null);
         expect(lineLayer.gradientVersion).toBeGreaterThan(gradientVersion);
     });
+
```
HarelM:
```suggestion
```

Issue #6746, opened 2025-11-21
Uncaught AbortErrors in vector and raster sources


andrewda:
**maplibre-gl-js version**: 5.13.0

**browser**: Chrome v142

### Steps to Trigger Behavior

 1. Add a source with a `url` property.
 2. Immediately call `setSourceProperty`.
 3. The source will load properly, but an unhandled AbortError will be thrown.

### Link to Demonstration

https://jsbin.com/majuzunafa/edit?html,console,output

### Expected Behavior

The AbortError should be handled internally (likely just ignored, since the request is automatically re-initiated).

### Actual Behavior

The AbortError is propagated, resulting in an unhandled promise rejection.

PR #6747, opened 2025-11-21
Gracefully handle AbortErrors

andrewda:
Fixes https://github.com/maplibre/maplibre-gl-js/issues/6746

When TileJSON requests are aborted via the specified AbortController (e.g., via `source.setSourceProperty`), Maplibre currently throws an `AbortError` that cannot be easily caught in userland. This change catches and explicitly handles the `AbortError` in `loadTileJson` by returning null.

## Launch Checklist

 - [x] Link to related issues.
 - [x] Post benchmark scores.


HarelM:
I'm a bit concerned about changing the underlaying ajax infrastructure to throw an AbortError instead of AjaxError, although I do think it makes a lot of sense.
In general about errors were are not well handled in the code, which I changed when migrating from callbacks to promises, and the about part was definitely one of the concerns when I changed that code.
So I don't know what would be a good enough testing effort to make sure everything is working as expected.
Any ideas would be greatly appreciated here.

src/util/abort_error.ts
```diff
@@ -9,13 +9,15 @@ export const ABORT_ERROR = 'AbortError';
...
 }
 
 /**
  * Use this when you need to create an abort error.
  * @returns An error object with the message "AbortError"
  */
 export function createAbortError(): Error {
-    return new Error(ABORT_ERROR);
+    const abortError = new Error(ABORT_ERROR);
+    abortError.name = ABORT_ERROR;
```
HarelM: Is this a common practice?
I think abort errors are also sent from the worker to the main thread and I don't want them to break due to this change.

src/source/vector_tile_source.ts
```diff
@@ -111,8 +111,9 @@ export class VectorTileSource extends Evented implements Source {
             const tileJSON = await loadTileJson(this._options, this.map._requestManager, this._tileJSONRequest);
             this._tileJSONRequest = null;
             this._loaded = true;
-            this.map.style.tileManagers[this.id].clearTiles();
             if (tileJSON) {
+                this.map.style.tileManagers[this.id].clearTiles();
```
HarelM: I'm not sure this is the right place to add this call.

src/source/load_tilejson.test.ts
HarelM, line 1: This is very good! Thanks for adding all those tests!

src/util/abort_error.ts
andrewda: An error with _name_ "AbortError" is the [correct response](https://developer.mozilla.org/en-US/docs/Web/API/DOMException#aborterror) to an AbortController signal. Keeping the "AbortError" message primarily for backwards-compatibility and to not break the existing tests.

src/source/vector_tile_source.ts
andrewda: Without this change, `src/source/vector_tile_source.test.ts > VectorTileSource > setTiles only clears the cache once the TileJSON has reloaded` fails because `clearTiles` is called twice -- once when the initial request is aborted (and a `null` response returned from `loadTileJson`), and then subsequently when the second request is completed successfully.

My understanding is that the only change to functionality here is when a vector source URL returns a 200 empty response. Previously, empty TileJSON responses would clear the tiles -- now it does not. Either way, we currently don't have any tests that exercise this code path, so not sure if empty TileJSON responses are even something worth supporting...

andrewda:
@HarelM 
> I'm a bit concerned about changing the underlaying ajax infrastructure

Fair. Probably makes sense to document where we use these ajax calls and their interaction with AbortControllers. I'll put a table together and try to put some specific tests in place wherever possible.

src/source/vector_tile_source.ts
andrewda: Actually, an easier solution may just be to move the AbortError catch to the tile sources, rather than handling it in `loadTileJson`. That approach also lets us stay consistent with where the abort logic lives (AbortController and AbortError handling in the same file).

HarelM: Sounds right.

src/source/load_tilejson.ts
```diff
@@ -2,6 +2,7 @@ import {pick, extend, type TileJSON} from '../util/util';
 import {getJSON} from '../util/ajax';
 import {ResourceType} from '../util/request_manager';
 import {browser} from '../util/browser';
+import {isAbortError} from '../util/abort_error';
```
HarelM: Probably a leftover?

src/source/raster_tile_source.test.ts
```diff
@@ -178,15 +178,33 @@ describe('RasterTileSource', () => {
         expect((server.lastRequest as any).aborted).toBe(true);
     });
 
-    test('supports url property updates', () => {
+    test('supports url property updates', async () => {
```
HarelM, line 181: Does this cover the extra logic added for aborting? I'm not sure I see it in this test, am I missing anything?

src/source/vector_tile_source.ts
```diff
@@ -125,7 +126,11 @@ export class VectorTileSource extends Evented implements Source {
         } catch (err) {
             this._tileJSONRequest = null;
             this._loaded = true; // let's pretend it's loaded so the source will be ignored
-            this.fire(new ErrorEvent(err));
+
+            // only fire error event if it is not due to aborting the request
```
HarelM, line 130: Somewhat related to this change, can we rename `_tileJSONRequest` to something related to abort controller?

src/util/abort_error.ts
```diff
@@ -3,19 +3,22 @@
  */
 export const ABORT_ERROR = 'AbortError';
 
+export class AbortError extends Error {
```
HarelM, line 6: Does this survives worker-main thread serializaiton?

src/util/ajax.ts
```diff
@@ -178,9 +182,7 @@ async function makeFetchRequest(requestParameters: RequestParameters, abortContr
         parsePromise = response.text();
     }
     const result = await parsePromise;
-    if (abortController.signal.aborted) {
-        throw createAbortError();
-    }
+    abortController.signal.throwIfAborted();
```
HarelM, line 186: What type of error will this throw?

andrewda:
@HarelM Here's my deep dive on existing AJAX uses across the codebase:

*makeRequest*
- src/util/dispatcher.ts
	- `getGlobalDispatcher`: Existing Actor behavior is to [leave promises pending](https://github.com/maplibre/maplibre-gl-js/blob/a3058260156ccd0e985007187d6795c105bf23a4/src/util/actor.ts#L109-L121) when aborted. This seems potentially problematic, but resolving will require a larger refactor. For now, confirmed that there are no regressions with existing Actor tests.
- src/util/image_request.ts
	- `doImageRequest`: Already handled in `RasterDEMTileSource.loadTile`, `RasterTileSource.loadTile`, and `Style._loadSprite`. Additional AbortError handling added in `ImageSource.load`. Tests added for all 4 uses.

*getJSON*
- src/source/geojson_worker_source.ts
	- `loadGeoJSONFromUrl`:  Only used in `loadAndProcessGeoJSON`, which is subsequently only used in `loadData`. Aborts already tested in `geojson_worker_source.test.ts`.
- src/source/load_tilejson.ts
	- `loadTileJson`: Used in load functions of Vector and RasterTileSource. New abort tests in `raster_tile_source.test.ts` and `vector_tile_source.test.ts`.
- src/style/load_spirit.ts
	- `loadSprite`: Used in `Style._loadSprite`. New tests added to cover abort conditions.
- src/style/style.ts
	- `loadURL`: New tests added to cover abort condition.
- src/ui/map.ts
	- `_diffStyle`: Never aborted. 

*getArrayBuffer*
- src/source/vector_tile_worker_source.ts
	- `loadVectorTile`: Used in `loadTile`. Abort already tested.
- src/style/load_glyph_range.ts
	- `loadGlyphRange`: Never aborted.

src/source/load_tilejson.ts
andrewda: removed!

src/source/vector_tile_source.ts
andrewda: The "request" naming is actually pretty consistent across the codebase right now. Agreed that it's not ideal, but maybe worth a later refactor considering how many places need to be updated?

src/util/ajax.ts
andrewda: It throws the signal's `reason`, which defaults to an Error with `name = "AbortError"`.

https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/throwIfAborted

src/util/abort_error.ts
andrewda: I don't think it needs to -- AbortErrors are caught and handled in the worker. However, I did a test locally and `isAbortError` does continue to work after serialization.

src/source/raster_tile_source.test.ts
andrewda: Yeah, this updated test covers aborted requests. Specifically, `source.setUrl` will abort the pending request, and we check the server requests to verify that 2 requests were made, with the first being aborted. We are also checking that the `error` event is not called -- prior to this PR, it would have been called with an exception.

src/source/vector_tile_source.ts
HarelM: Fair enough.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
+- Fix AbortError handling in loadTileJson. ([#6747](https://github.com/maplibre/maplibre-gl-js/pull/6747)) (by [@andrewda](https://github.com/andrewda))
```
HarelM: This needs to be a bit more clear as it somewhat changes error handling for abort errors.
I would even add a yellow triangle here just in case.
I'm not sure if this is abr aking change that requires a breaking change version or not to be honest...

HarelM:
The following is what I was concerned about regarding serialization, I think that it's still ok since AbortError inherits from Error.
https://github.com/maplibre/maplibre-gl-js/blob/a3058260156ccd0e985007187d6795c105bf23a4/src/util/web_worker_transfer.ts#L112

Regarding your comment about actor abort behavior, if you think you can fix that as well as part of this change it can be great. 
I originally refactored actor from callbacks to promise and added support for abort controller and decided not to charge the behavior that existed prior to my changes (I also decided not to change the variables names, thus the request name still exists). The change was very large and I decided not to dive into abort flows.
It might be that this PR is the right place and time to address it as you seem to have dived deep into abort flows.
It can be a separate PR if you think this one is too big already.

CHANGELOG.md
andrewda: Yeah happy to go into more detail here and make it a little more clear. I don’t think this requires a breaking change version bump, though. Prior to this change, these AbortErrors would be thrown internally and could not be caught in userland without some hacking (e.g., adding a global unhandledRejection event listener). Off the top of my head, I can’t think of any reason why a user would need to be able to handle AbortErrors this way, other than to suppress them. For tile loading purposes, tile.aborted already exists as supported external abort-checking.

andrewda:
@HarelM Thanks for the context on the current state of the Actor behavior. I agree that a refactor makes sense, but I think it’s going to increase the scope substantially. From a quick test, changing how the Actor handles aborts breaks unit tests in a number of places, and I think it may even warrant some additional integration tests.

Happy to take on this work, but might make sense to do so separately from this PR.

HarelM:
A different PR is totally fine.

CHANGELOG.md
andrewda: Updated with more information -- let me know if you think there's still anything missing.

HarelM (approved):

PR #6756, opened 2025-11-23
Refactor some code out of tile manager

HarelM:
This introduces the `InViewTiles` class that stores all the tiles that are currently loaded in view.
It moves some of the logic related to these tiles to this class.
It moves the raster fade logic into another file (not a class, but just a method at this point).
It moves some methods and for loops around to better places than what they used to be.
This PR doesn't change any internal logic as far as I could see.

This is related to other PRs in this area, in order to improve the code in Tile Manager.
I couldn't find a lot of code related to vector handling that is worth a `tile_manager_vector.ts` file, but I might have overlooked.

## Launch Checklist

 - [x] Link to related issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 ## main
 
 ### ✨ Features and improvements
-- Improve performance of `GeoJSONSource#updateData` ([#6738](https://github.com/maplibre/maplibre-gl-js/pull/6738)) (by [@lucaswoj](https://github.com/lucaswoj))
 - Prevent redundant DEM border backfilling by tracking state, moved tile manager logic to helper files ([#6756](https://github.com/maplibre/maplibre-gl-js/pull/6756)) (by [@HarelM](https://github.com/HarelM))
 - Improve performance of GeoJSON `updateData`, `setData`, and overzoomed  tiles ([#6738](https://github.com/maplibre/maplibre-gl-js/pull/6738), [#6772](https://github.com/maplibre/maplibre-gl-js/pull/6772)) (by [@lucaswoj](https://github.com/lucaswoj))
 - _...Add new stuff here..._


src/source/raster_dem_tile_source.ts
```diff
@@ -126,7 +126,7 @@ export class RasterDEMTileSource extends RasterTileSource implements Source {
         const nx = (canonical.x + 1 + dim) % dim;
         const nxw = canonical.x + 1 === dim ? tileID.wrap + 1 : tileID.wrap;
 
-        const neighboringTiles = {};
+        const neighboringTiles: Record<string, {backfilled: boolean}> = {};
```
HarelM, line 129: `backfilled` is never checked, just assigned, I'm wondering if this is a potential optimization that can improve loading of hillshade etc...

src/style/style.ts
```diff
@@ -2042,15 +2042,6 @@ export class Style extends Evented {
         for (const id in this.tileManagers) {
             const tileManager = this.tileManagers[id];
             tileManager.setEventedParent(null);
-
-            if (tileManager._tiles) {
```
HarelM, line 2049: This part was moved inside the `onRemove` method of `tileManager`.

src/tile/tile_id.ts
```diff
@@ -242,5 +242,14 @@ function getQuadkey(z, x, y) {
     return quadkey;
 }
 
+export function compareTileId(a: OverscaledTileID, b: OverscaledTileID): number {
```
HarelM, line 245: No changes were made to this method, it was simply moved here.
There's another method that does sorting, but I'm not sure why it is inside another method...

src/tile/tile_manager_in_view_tiles.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    public getRenderableIds(bearingInRadians: number = 0, symbolLayer?: boolean): Array<string> {
+        const renderables: Array<Tile> = [];
+        for (const id of this.getAllIds()) {
+            if (this.isIdRenderable(id, symbolLayer)) {
+                renderables.push(this.getTileById(id));
+            }
+        }
+        if (symbolLayer) {
+            return renderables.sort((a_: Tile, b_: Tile) => {
```
HarelM, line 74: This is the sort method I was referring to...
I'm not sure why two sort operations are needed here, but I'm not sure I care enough to investigate...

src/tile/tile_manager_raster_dem.ts
```diff
@@ -0,0 +1,43 @@
+import {type Tile} from './tile';
+import {type InViewTiles} from './tile_manager_in_view_tiles';
+
+/**
+ * For raster terrain source, backfill DEM to eliminate visible tile boundaries
+ */
+export function backfillDEM(tile: Tile, inViewTiles: InViewTiles) {
```
HarelM, line 7: I'm not sure if this should be part of the raster dem source class or not.
It does use the `InViewTiles`, but otherwise this looks as if it should be handled by the source and not the tile manager. Maybe, IDK.

src/ui/map_tests/map_basic.test.ts
```diff
@@ -137,9 +137,9 @@ describe('Map', () => {
             await map.once('load');
             const fakeTileId = new OverscaledTileID(0, 0, 0, 0, 0);
             map.addSource('geojson', createStyleSource());
-            map.style.tileManagers.geojson._tiles[fakeTileId.key] = new Tile(fakeTileId, undefined);
+            map.style.tileManagers.geojson._inViewTiles.setTile(fakeTileId.key, new Tile(fakeTileId, undefined));
```
HarelM, line 140: And this is the reason I keep saying private field should not be used in unit tests, or at least used as less as possible...

src/tile/tile_manager.ts
```diff
@@ -276,21 +266,22 @@ export class TileManager extends Evented {
...
         }
 
-        this._cache.reset();
+        this._outOfViewCache.reset();
 
-        for (const i in this._tiles) {
-            if (shouldReloadTileOptions && this._source.shouldReloadTile && !this._source.shouldReloadTile(this._tiles[i], shouldReloadTileOptions)) {
+        for (const id of this._inViewTiles.getAllIds()) {
+            const tile = this._inViewTiles.getTileById(id);
+            if (shouldReloadTileOptions && this._source.shouldReloadTile && !this._source.shouldReloadTile(tile, shouldReloadTileOptions)) {
```
wayofthefuture: not related but `&& this._source.shouldReloadTile &&` is likely a no op

src/tile/tile_manager_in_view_tiles.ts
```diff
@@ -0,0 +1,80 @@
+import Point from '@mapbox/point-geometry';
+import {type LayerFeatureStates} from '../source/source_state';
+import {type Tile} from './tile';
+import {compareTileId, type OverscaledTileID} from './tile_id';
+
+export class InViewTiles {
+    private _inViewTiles: Record<string, Tile> = {};
```
wayofthefuture: consider `private _tiles: Record<string, Tile>`

src/tile/tile_manager_in_view_tiles.ts
```diff
@@ -0,0 +1,80 @@
...
+            const tile = this._inViewTiles[id];
+            tile.setFeatureState(featuresChanged, painter);
+        }
+    }
+
+    public getAllIds(sorted = false): string[] {
+        if (!sorted) {
+            return Object.keys(this._inViewTiles);
+        }
+        return Object.values(this._inViewTiles).map(tile => tile.tileID).sort(compareTileId).map(id => id.key);
```
wayofthefuture: consider invert and move up

src/tile/tile_manager_in_view_tiles.ts
```diff
@@ -0,0 +1,80 @@
...
+        if (symbolLayer) {
+            return renderables.sort((a_: Tile, b_: Tile) => {
+                const a = a_.tileID;
+                const b = b_.tileID;
+                const rotatedA = (new Point(a.canonical.x, a.canonical.y))._rotate(-bearingInRadians);
+                const rotatedB = (new Point(b.canonical.x, b.canonical.y))._rotate(-bearingInRadians);
+                return a.overscaledZ - b.overscaledZ || rotatedB.y - rotatedA.y || rotatedB.x - rotatedA.x;
+            }).map(tile => tile.tileID.key);
+        }
+        return renderables.map(tile => tile.tileID).sort(compareTileId).map(id => id.key);
```
wayofthefuture: I agree I think only 1 sort should be needed

src/tile/tile_manager.ts
```diff
@@ -132,10 +134,15 @@ export class TileManager extends Evented {
     }
 
     onRemove(map: Map) {
+        for (const id of this._inViewTiles.getAllIds()) {
+            const tile = this._inViewTiles.getTileById(id);
```
wayofthefuture: I see this pattern in many places... maybe this._inViewTiles.getAllTiles() could help

src/tile/tile_manager.ts
```diff
@@ -991,11 +756,12 @@ export class TileManager extends Evented {
...
+            if (!this._inViewTiles.isIdRenderable(id) && tile.state != 'errored') {
                 continue;
             }
-            if (tileIds.some(tid => tid.equals(this._tiles[id].tileID.canonical))) {
+            if (tileIds.some(tid => tid.equals(tile.tileID.canonical))) {
                 this._reloadTile(id, 'expired');
             }
         }
```
wayofthefuture, line 761: although unrelated not a fan the O(n) nested loop, `Set` may be better... odd we can't use _inViewTiles.getRenderables here

wayofthefuture (approved):
I gave it a good read, but can't really get into testing it as I'm out of town. Will have more time starting Wednesday. Looks good to me though... I added some slight suggestions...

src/tile/tile_manager.ts
HarelM: That's a good idea, I wanted to keep the code similar in some places, but sure, this can be simplified.

src/ui/map_tests/map_basic.test.ts
HarelM: True, but not by using them explicitly, instead, there should either be a mock or something like that, not these kind of calls...

src/tile/tile_manager.ts
HarelM: Do you want me to remove any part of this if statement?

src/tile/tile_manager_in_view_tiles.ts
HarelM: Done in bdbc879

src/tile/tile_manager_in_view_tiles.ts
HarelM: Done in bdbc879

src/tile/tile_manager.ts
HarelM: Done in bdbc879, other places are using the `id` field later in the loop, so I kept it.

src/tile/tile_manager.ts
wayofthefuture: I think the middle condition can be removed

src/tile/tile_manager.ts
HarelM: renderable is also sorting, and it doesn't include non-errored tiles.
I'm not sure it is dramatically different, but I didn't want to change the logic at this point in time.
We can build a set and then use it to reload the relevant tiles, but I don't think it will bring a dramatic performance improvement...

src/tile/tile_manager.ts
HarelM: Done in 0758d6b

src/tile/tile_manager.ts
```diff
@@ -445,10 +394,7 @@ export class TileManager extends Evented {
         const loadedDescendents: Record<string, Tile[]> = {};
 
         // enumerate current tiles and find the loaded descendents of each target tile
-        for (const id in this._tiles) {
-            const tile = this._tiles[id];
-            if (!tile.hasData()) continue;
-
+        for (const tile of this._inViewTiles.getAllTiles().filter(tile => tile.hasData())) {
```
wayofthefuture, line 397: 😕 on the extra loop liked the continue better

HarelM: I can revert this if you insist, but I kinda like it.
Are there other comments I should be addressing in your opinion?

wayofthefuture: I don't insist but the extra loops add up!

I feel like isrenderable code could be consolidated / cleaned up but likely not this PR...

HarelM: I think this is the same performane-wise, there's no really extra loop. When you check the if you do n checks and then continue to do the rest for n-x (x those with data).
It's the same number of operations in total...
2n-x in both cases.

wayofthefuture: Okay.forEach(commit => commit.merge({squash:true})) !!

HarelM: I'll add some tests to the sorting code and merge this. Thanks!

src/tile/tile_manager_in_view_tiles.ts
wayofthefuture: Once sentence from gpt: Symbol layers need special rotation and zoom sensitive sorting for stable text/icon symbol placement; all other layers rely on a simpler canonical tile ordering.

wayofthefuture:
I accidentally deleted your comment about tests fyi!

src/tile/tile_manager_in_view_tiles.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /**
+     * Get a currently loaded tile.
+     * - a cached tile is not a loaded tile
+     * @returns the tile if it's in view and had data, null otherwise.
+     */
+    public getLoadedTile(tileID: OverscaledTileID): Tile | null {
+        const tile = this.getTileById(tileID.key);
+        if (tile?.hasData()) {
```
wayofthefuture, line 56: if (!tile?.hasData() return null;
return tile;
???

src/tile/tile_manager_in_view_tiles.ts
HarelM: I kept it in the same place as it requires the value of `bearingInRadians` and it felt wrong to move it from there currently.
Maybe in a later stage.
I've added the relevant tests to cover its functionality.

src/tile/tile_manager_in_view_tiles.ts
HarelM: Too cryptic for me. I think it's good the way it is now. A few more character for readability don't hurt...

HarelM:
Ok, I've added tests to sorting, added some checks to backfill flag to avoid doing it twice when there's no need, and added tests for that.

src/tile/tile_manager_raster_dem.ts
HarelM: Let's keep it here for now...

PR #6781, opened 2025-11-27
Use _getTransformForUpdate for changes to min zoom and max zoom as it may apply changes to zoom

### LINKED ISSUE #6766: Changes to transform via setMinZoom or setMaxZoom can be superseded by stale transform copy when using transformCameraUpdate

**browser**: N/A

### Steps to Trigger Behavior

 1. Change min or max zoom via setMinZoom / setMaxZoom, ensuring zoom is updated due to it being outside of new bounds
 2. Zoom out via scroll
 3. Note that sometimes the previous state of the transform is re-instated, resetting your zoom, etc.

https://github.com/user-attachments/assets/e49d4386-9190-4a6c-be73-48db8fee87cc

### Link to Demonstration

https://stackblitz.com/edit/vitejs-vite-drzrraph

```html
<!DOCTYPE html>
<html lang="en">

<head>
    <script src=""></script>
    <link href="https://esm.sh/maplibre-gl@5.13.0/dist/maplibre-gl.css" rel="stylesheet">
    </link>
    <style>
        body {
            margin: 0;
        }

        #app {
            position: relative;
            height: 100vh;
            width: 100vw;
        }

        #map {
            width: 100%;
            height: 100%;
        }

        #change-zoom {
            position: absolute;
            top: 1rem;
            left: 1rem;
            z-index: 100;
        }

        #current-zoom {
            position: absolute;
            z-index: 100;
            font-size: 0.8rem;
            background-color: white;
            padding: 0.25rem;
            bottom: 1rem;
            left: 1rem;
        }
    </style>
</head>

<body>
    <div id="app">
        <button id="change-zoom">Change Zoom</button>
        <div id="map"></div>
        <div id="current-zoom"></div>
    </div>
    <script type="module">
        import maplibre from 'https://esm.sh/maplibre-gl@5.13.0'

        const map = new maplibre.Map({
            container: 'map',
            style: {
                name: 'MapLibre',
                layers: [
                    {
                        id: 'satellite',
                        type: 'raster',
                        source: 'maplibre',
                    },
                ],
                sources: {
                    maplibre: {
                        tiles: [
                            'https://server.arcgisonline.com/arcgis/rest/services/World_Imagery/MapServer/tile/{z}/{y}/{x}',
                        ],
                        type: 'raster',
                    },
                },
                version: 8,
            },
            center: [151.18068748043615, -33.96434511002358],
            zoom: 5,
            minZoom: 5,
            maxZoom: 18,
            transformCameraUpdate: (t) => t,
        });

        map.on('render', (e) => {
            const minZoom = e.target.getMinZoom();
            const maxZoom = e.target.getMaxZoom();
            const zoom = e.target.getZoom();
            document.querySelector(
                '#current-zoom'
            ).innerHTML = `${minZoom} - ${maxZoom} (${zoom})`;
        });

        document.querySelector('#change-zoom').addEventListener('click', () => {
            const currentMinZoom = map.getMinZoom();
            map.setMinZoom(currentMinZoom === 5 ? 8 : 5);
            map.setMaxZoom(currentMinZoom === 5 ? 16 : 18);
        });
    </script>
</body>

</html>
```

### Expected Behavior

transform continues to move forward and does not revert to old copies

### Actual Behavior

transform reverts to old copies which resets changes made

See downstream bug report: https://github.com/visgl/react-map-gl/issues/2565 which I will close now that I can reproduce with vanilla Maplibre.

Auspicus:
Adds `_getTransformForUpdate` calls to `setMinZoom` and `setMaxZoom` to allow `transformCameraUpdate` to apply changes to `zoom` and prevent stale transform copies from being reapplied.

From a correctness perspective, this probably makes more sense as a change to the transform (specifically zoom here) should be run through the `transformCameraUpdate` function to allow the user to apply their own changes to the transform. It serendipitously also fixes the bug with the transform copies being re-applied, although I'm not sure exactly why they were being re-applied in the first place.

`_getTransformForUpdate` is not called within `map.ts` so might be better suited to `camera.ts`.

Fixes #6766

todo:
- [x] add test which checks that `transformCameraUpdate` is called after `setXZoom` calls
- [x] add entry to `CHANGELOG.md`
- [x] add test for transform copy being re-applied?

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 
 ### 🐞 Bug fixes
 - Fix getting the right zoom for getElevationForLngLat ([#6825](https://github.com/maplibre/maplibre-gl-js/pull/6825)) (by [@HarelM](https://github.com/HarelM))
-- Fix issue [#6766](https://github.com/maplibre/maplibre-gl-js/issues/6766) where stale transform state was being applied after changing `minZoom` or `maxZoom` due to old copies from `transformCameraUpdate` taking precedence. `transformCameraUpdate` is now called from `setMinZoom` and `setMaxZoom` to allow user to control subsequent changes to `zoom` (by [@Auspicus](https://github.com/Auspicus))
+- Fix stale transform state being applied after changing `minZoom` or `maxZoom` due to old copies from `transformCameraUpdate` taking precedence. `transformCameraUpdate` is now called from `setMinZoom` and `setMaxZoom` to allow user to control subsequent changes to `zoom` [#6766](https://github.com/maplibre/maplibre-gl-js/issues/6766) (by [@Auspicus](https://github.com/Auspicus))
 - _...Add new stuff here..._
 
 ## 5.14.0


HarelM:
To be honest I'm not sure what's the point of map inheriting from camera as it is the only class the inherits from camera, they probably need to be united into a single class, but that's a bit besides the point here.
Can you add a test to cover the bug that this is fixing?
Also, any chance you could look for other public API related to map and transform that might need this fix as well?

HarelM:
Thanks for taking the time to open this PR and fix this issue!

HarelM:
Hi, any updates regarding the tests and changelog entry?

Auspicus:
Hey! I'll aim to get back to this sometime this coming week.

src/ui/map.ts
```diff
@@ -1144,11 +1144,11 @@ export class Map extends Camera {
         minZoom = minZoom === null || minZoom === undefined ? defaultMinZoom : minZoom;
 
         if (minZoom >= defaultMinZoom && minZoom <= this.transform.maxZoom) {
-            this.transform.setMinZoom(minZoom);
+            const tr = this._getTransformForUpdate();
+            tr.setMinZoom(minZoom);
+            this._applyUpdatedTransform(tr);
             this._update();
 
-            if (this.getZoom() < minZoom) this.setZoom(minZoom);
```
Auspicus, line 1150: I think this might be why the transform was occasionally reapplied. Race between `this._update` change and the `this.setZoom`

See the description on the `ITransformMutators.setMinZoom` (or `ITransformMutators.setMaxZoom`) where it's said that it will automatically constrain the zoom to the new range: https://github.com/maplibre/maplibre-gl-js/blob/main/src/geo/transform_interface.ts#L111-L120

Auspicus:
@HarelM I've rebased this, updated the `CHANGELOG.md` and added a test. I also adjusted it a little bit after realizing that `setMinZoom` and `setMaxZoom` on the transform was updating `zoom` itself and so the map doesn't need to set it as well.

src/ui/map.ts
HarelM: I see you have removed this condition, and I understand why, did you test that this is what's actually happening? i.e. that this if was indeed redundant?
Also it might be worth adding a test to make sure this is the behavior that we are expecting and that it is kept this way.

HarelM: Ahh, I see from your last comment that you have tested it.

PR #6790, opened 2025-11-30
chore: update the gl-js version in the docs

CommanderStorm:
The version of gl-js in the docs does not seem to be updated in a while.

This artefact was changed by the generate-docs step during testing a few style spec changes


HarelM:
This is updated as part of the docs build so when a version is released this is updated automatically as far as I know, but sure, we can update it here as well.

HarelM (approved):

PR #6791, opened 2025-11-30
Fix queryTerrainElevation

### LINKED ISSUE #6701: queryTerrainElevation() returns inconsistent values depending on zoom/pitch
**maplibre-gl-js version**: v5.11.0 (the same with v5.8.0)

**browser**: flutter_inappwebview v6.1.5 (the same with webview_flutter + flutter_linux_webview)

**Description of the issue**

For the same geographic coordinates (lng, lat), the map.queryTerrainElevation() function in maplibregl.Map returns significantly different elevation values depending on the current zoom level and camera pitch/angle.

**Impact**

1) **Markers** (maplibregl.Marker) **visibly “jump” on the map**, sometimes sinking below the terrain surface and sometimes popping back up.
2) **Elevation values can be off by hundreds of meters** (we observed deviations up to **300 m**), making the function unreliable for practical use.
3) While terrain height could theoretically be fetched from an external elevation service (which is strange to do when the map already has terrain), **this would not solve the visual jumping of objects**, which appears to be caused by the inconsistent results of queryTerrainElevation().

### Steps to Trigger Behavior

 1. Initialize a map: const map = new maplibregl.Map({ ... });
 2. Add markers: const marker = new maplibregl.Marker().setLngLat([lng, lat]).addTo(map);
 3. Log terrain elevation for a fixed coordinate: 
    map.on('zoomend', () => {
      const lngLat = { lng: 14.2663, lat: 47.4661 };
      const amsl = map.queryTerrainElevation(lngLat, { exaggerated: true });
      console.log("[queryTerrainElevation] =", amsl);
    });
4. Change zoom/pitch. The elevation value changes drastically even though the coordinates remain identical.
Attached is a video demonstrating the issue.

### Expected Behavior
queryTerrainElevation() should return consistent elevation values for the same coordinates, regardless of zoom level or camera pitch.

### Actual Behavior
The function returns different values depending on the current view, which leads to unstable marker placement and makes use of terrain elevation impossible in real applications.

### Request
Could you please clarify:
- Is this a known issue with terrain sampling?
- Is there a recommended workaround?
- How can we achieve stable and reliable elevation queries?
At the moment, this issue makes it impossible to use the terrain functionality safely.
Thank you!

https://github.com/user-attachments/assets/35d7ebe2-a298-4fb5-b453-beaf9329f513

HarelM:
## Launch Checklist

- Fixes #6701

Instead of just checking the in view tiles, this adds a check to see if the relevant tile is out of view but still in the cache in order to get the elevation from there when possible.
It also sets the zoom to be maximum in order to go up the zoom levels to find the right tile. Before this fix the starting zoom was the zoom of the center of the map, which might not be the correct zoom for all tiles.

 - [x] Link to related issues.


HarelM:
I'm still able to find where this breaks, I'll continue to try and find a better solution.

PR #6800, opened 2025-12-02
Use only bounds for geojson diff tile reload 

### LINKED ISSUE #6795: geoJsonSource.updateData not working properly after updating to 5.13.0

**browser**: Chrome 142

### Steps to Trigger Behavior

 1. press on the Hide button

also the text gets hidden if zooming around

Downgrading to 5.10.0 solves this issue
also when setting the `label-layer`s source to `shape-source` the text also gets hidden

### Link to Demonstration

https://jsbin.com/luyixayona/edit?html,console,output

### Expected Behavior

Hide features when they are removed

### Actual Behavior

Features dont get hidden

HarelM:
## Launch Checklist

- Fixes #6795

This changes the logic of tile refreshing for geojson
The idea behind this fix:
1. When updating the data, we know exactly which geometries were changed, so we use this information to return the relevant bounds of these geometries
2. It simplifies to only using bounds and not using ids of previous features

 - [x] Link to related issues.

### DIFF
--- a/src/source/geojson_source_diff.ts
+++ b/src/source/geojson_source_diff.ts
@@ -177,9 +177,9 @@ export function applySourceDiff(updateable: Map<GeoJSONFeatureId, GeoJSON.Featur
             if (!changeGeometry && !changeProps) continue;
 
             // clone once since we'll mutate
+            affectedGeometries.push(existing.geometry);
             const feature = {...existing};
             updateable.set(update.id, feature);
-            affectedGeometries.push(feature.geometry);
 
             if (changeGeometry) {
                 affectedGeometries.push(update.newGeometry);


HarelM:
@lucaswoj @wayofthefuture let me know what you think about the direction of this PR before I invest more time in fixing the tests and adding new tests.

wayofthefuture:
I like the way this is headed.

However, let's say a feature was moved from 1 tile to another tile. How do you know to invalidate the previous tile without the feature id? The new bounds would not affect the previous tile.

HarelM:
If a feature was moved from one tile to the other - I have the previous geometry just before I apply the update, and the new geometry.
So I "invalidate" tiles that have one of those geometries.

HarelM:
I've gone and updated the relevant tests. I've removed some tests that are "duplicated" in the sense that they are being tested in lat lng bounds and there's no need to retest them in geojson source.
I've added tests that are using only public API to mimic how the class is behaving.
There's still a weird issue with the "covering world tile", which causes a test to fail.
I added a similar test to lnglatbounds tests, which also fails to highlight the issue in a simpler test.
@lucaswoj, do let me know what you think about this test and if there's something wrong about it.

lucaswoj:
Please hold off on merging these changes until I have time to review later today. I think this introduces some new bugs.

HarelM:
Sure, take your time. I wouldn't merge this before you review it.

src/source/geojson_source.ts
```diff
@@ -466,70 +462,46 @@ export class GeoJSONSource extends Evented implements Source {
             this._data = {updateable: toUpdateable(this._data.geojson, promoteId)};
         }
 
-        if (this._data.updateable) {
-            applySourceDiff(this._data.updateable, diff, promoteId);
+        if (!this._data.updateable) {
+            return undefined;
         }
```
lucaswoj, line 475: ~This will break `updateData` for sources that have been loaded from a URL. Such sources will never have a truthy value for `this._data.updatable` on the main thread. Per #6668 this data is only loaded and available on the worker thread.~

lucaswoj: Ah, I see, this now returns the `GeoJSONSourceShouldReloadTileOptions` object directly. So, sources that have been loaded from a URL will still work with `updateData` but their performance will regress significantly.

src/source/geojson_source.ts
```diff
@@ -466,70 +462,46 @@ export class GeoJSONSource extends Evented implements Source {
...
-            const featureIndex = tile.latestFeatureIndex.featureIndexArray.get(i);
-            const feature = layers[GEOJSON_TILE_LAYER_NAME].feature(featureIndex.featureIndex);
-            const id = tile.latestFeatureIndex.getId(feature, GEOJSON_TILE_LAYER_NAME);
-            if (prevIds.has(id)) {
-                return true;
-            }
+        if (tile.state === 'unloaded') {
+            return false;
         }
-
```
lucaswoj, line 521: Removing this codepath will cause performance regressions in some cases. Although having two codepaths is ugly, invalidating tiles by id is much faster and more accurate than doing so by bounds.

src/source/geojson_source.ts
```diff
@@ -466,70 +462,46 @@ export class GeoJSONSource extends Evented implements Source {
...
      */
-    shouldReloadTile(tile: Tile, {nextBounds, prevIds}: GeoJSONSourceShouldReloadTileOptions) : boolean {
-        if (!tile.latestFeatureIndex) {
-            return tile.state !== 'unloaded';
+    shouldReloadTile(tile: Tile, {affectedBounds}: GeoJSONSourceShouldReloadTileOptions) : boolean {
+        if (tile.state === 'loading') {
+            return true;
         }
-
-        // Update the tile if it PREVIOUSLY contained an updated feature.
```
lucaswoj, line 511: > When moving a feature from one place to the other, we'll need to reload both the previous tile that it was in and the new tile is was moved to, which I'm not sure is handled in the previous code

Per this comment, this is the code that handles "the previous tile"

src/source/geojson_source.ts
```diff
@@ -466,70 +462,46 @@ export class GeoJSONSource extends Evented implements Source {
...
-            const feature = layers[GEOJSON_TILE_LAYER_NAME].feature(featureIndex.featureIndex);
-            const id = tile.latestFeatureIndex.getId(feature, GEOJSON_TILE_LAYER_NAME);
-            if (prevIds.has(id)) {
-                return true;
-            }
+        if (tile.state === 'unloaded') {
+            return false;
         }
-
         // Update the tile if it WILL NOW contain an updated feature.
```
lucaswoj: > When moving a feature from one place to the other, we'll need to reload both the previous tile that it was in and the new tile is was moved to, which I'm not sure is handled in the previous code

Per this comment, this is the code that handles "the new tile it was moved to"

lucaswoj:
Thanks for diving into this code @HarelM! This does fix the bug but it may regress performance in some cases. I've proposed another fix at #6801, ready for your review, that is performance neutral.

HarelM:
Can we measure how much performance degradation this is causing?
This code is a lot simpler to maintain and doesn't relay on feature index (including the timing related to it, which I find hard to make sure won't break in the future).

lucaswoj:
> Can we measure how much performance degradation this is causing?

There's not a single number that captures the whole picture. 

We can say that https://github.com/maplibre/maplibre-gl-js/pull/6562 would no longer apply to sources loaded from a URL, so updating 1 point out of 100k points would become ~100% slower after this PR merges.

https://github.com/maplibre/maplibre-gl-js/pull/6562 would also no longer apply in some cases involving more complex geometries, like an L-shaped line that crosses three out of 4 tiles in the viewport. We could create another benchmark that would become ~100% slower after this PR merges, for a non-URL-backed source.

HarelM:
Fair enough, can you take a look at the lat lang bounds failing test? I'm not sure I understand why it's failing.

HarelM:
It's worth mentioning that for tiles with a lot of features and small updates to the geojson, this route has better performance.
I'm not sure what I think about loading the geojson into memory when using url to improve performance, but that's also a possible route here.

lucaswoj:
> Fair enough, can you take a look at the lat lang bounds failing test? I'm not sure I understand why it's failing.

Just created https://github.com/maplibre/maplibre-gl-js/pull/6802 to investigate

HarelM:
@mwilsnd does the fix related to geojson undefined properties has to do with the feature index etc?
I'm wondering if the fix here is helps with your fix or if they are not related.

mwilsnd:
Take a look: #6803 - this just changes how the layers are constructed in the feature index for geojson sources, I'm not sure if this is fully related.

HarelM:
The features from the geojson are sent back for every tile, so it might be somewhat related, I'm not sure, I need to wrap my head around it...

src/source/geojson_source.ts
```diff
@@ -45,12 +44,7 @@ export type GeoJSONSourceShouldReloadTileOptions = {
     /**
      * Refresh all tiles that WILL contain these bounds.
      */
-    nextBounds: LngLatBounds[];
-
-    /**
-     * Refresh all tiles that PREVIOUSLY DID contain these feature ids.
-     */
-    prevIds: Set<GeoJSON.Feature['id']>;
+    affectedBounds: LngLatBounds[];
```
wayofthefuture, line 47: Since affectedBounds is the only property, would it be better to remove shouldReloadTileOptions and simply pass affectedBounds in the data event?

src/source/geojson_source.ts
```diff
@@ -455,7 +451,7 @@ export class GeoJSONSource extends Evented implements Source {
         }
     }
 
-    private _applyDiff(diff: GeoJSONSourceDiff): void {
+    private _applyDiff(diff: GeoJSONSourceDiff): GeoJSONSourceShouldReloadTileOptions {
```
wayofthefuture: Here you would be returning affectedBounds which makes more sense with _applyDiff imo

src/source/geojson_source_diff.ts
```diff
@@ -126,13 +125,16 @@ export function toUpdateable(data: UpdateableGeoJSON, promoteId?: string) {
...
  */
-export function applySourceDiff(updateable: Map<GeoJSONFeatureId, GeoJSON.Feature>, diff: GeoJSONSourceDiff, promoteId?: string): void {
+export function applySourceDiff(updateable: Map<GeoJSONFeatureId, GeoJSON.Feature>, diff: GeoJSONSourceDiff, promoteId?: string): GeoJSON.Geometry[] {
+    const affectedGeometries: GeoJSON.Geometry[] = [];
     if (diff.removeAll) {
         updateable.clear();
     }
     else if (diff.remove) {
         for (const id of diff.remove) {
+            affectedGeometries.push(updateable.get(id)?.geometry);
```
wayofthefuture: while I know you have the check in shouldReloadTile this would nevertheless push undefined values into the array...

src/source/geojson_source_diff.ts
```diff
@@ -141,6 +143,7 @@ export function applySourceDiff(updateable: Map<GeoJSONFeatureId, GeoJSON.Featur
         for (const feature of diff.add) {
             const id = getFeatureId(feature, promoteId);
             if (id != null) {
+                affectedGeometries.push(feature.geometry);
```
wayofthefuture: Currently updateable allows the addition of duplicate (replacing) feature ids, so you might want to consider checking/adding existing affected feature geometry here...

src/source/geojson_source_diff.ts
```diff
@@ -126,13 +125,16 @@ export function toUpdateable(data: UpdateableGeoJSON, promoteId?: string) {
  * 1. Remove operations (removeAll, remove)
  * 2. Add operations (add)
  * 3. Update operations (update)
+ * @returns an array of geometries that were affected by the diff
  */
-export function applySourceDiff(updateable: Map<GeoJSONFeatureId, GeoJSON.Feature>, diff: GeoJSONSourceDiff, promoteId?: string): void {
+export function applySourceDiff(updateable: Map<GeoJSONFeatureId, GeoJSON.Feature>, diff: GeoJSONSourceDiff, promoteId?: string): GeoJSON.Geometry[] {
+    const affectedGeometries: GeoJSON.Geometry[] = [];
     if (diff.removeAll) {
         updateable.clear();
```
wayofthefuture, line 139: while I realize you return undefined in applyDiff the return value is nevertheless incorrect... not sure if this matters

HarelM:
Thanks for the review! 
Note that I updated my initial message as feature updates were supported in the previous version, I just read the code wrong.

There are some performance changes (in some flows they are improved while in others they degrade).
I can't say which is better. 
I can say that I think this approach is a bit less fragile as it doesn't relay on feature index.

I'll work on the review comments nevertheless.

src/source/geojson_source.ts
HarelM: I think keeping it this way allows adding properties in the future if needed, so I'm fine with keeping it that way. 
I had the same thought myself, but decided to keep it as is.

src/source/geojson_source.ts
HarelM: Actually, changing this makes more sense in terms of API. I'll change that but keep the options object.

src/source/geojson_source_diff.ts
HarelM: Good point, I'll check before pushing...

src/source/geojson_source.ts
HarelM: Oh, wait, the problem is that there's a need to return undefined in some cases here, which is not equal to empty array of affected bound...

src/source/geojson_source_diff.ts
HarelM: Fixed in 755ec5f41ff0f9bd86e475e8cd1cb13e1585d4bd

src/source/geojson_source_diff.ts
HarelM: Fixed in 755ec5f41ff0f9bd86e475e8cd1cb13e1585d4bd

src/source/geojson_source.ts
HarelM: Fixed in 755ec5f41ff0f9bd86e475e8cd1cb13e1585d4bd

src/source/geojson_source.ts
HarelM: Let me know if the change I made here makes sense now: 755ec5f41ff0f9bd86e475e8cd1cb13e1585d4bd

src/source/geojson_source_diff.ts
HarelM: Can you clarify a bit what you mean here?
I can move the push before the "if", if that's what you mean.

wayofthefuture:
https://github.com/maplibre/maplibre-gl-js/pull/6809

src/source/geojson_source_diff.ts
```diff
@@ -162,10 +177,12 @@ export function applySourceDiff(updateable: Map<GeoJSONFeatureId, GeoJSON.Featur
             if (!changeGeometry && !changeProps) continue;
 
             // clone once since we'll mutate
-            feature = {...feature};
+            const feature = {...existing};
             updateable.set(update.id, feature);
+            affectedGeometries.push(feature.geometry);
```
wayofthefuture: readability only - consider moving this 1 line up 4 lines and change feature.geometry to existing.geometry

src/source/geojson_source_diff.test.ts
```diff
@@ -277,90 +278,126 @@ describe('applySourceDiff', () => {
         expect(updateable.has('point2')).toBeTruthy();
     });
 
+    test('adds a feature that exist and change its geometry', () => {
```
wayofthefuture, line 281: really nice!

src/source/geojson_source_diff.ts
HarelM: Ha, your code 😅

HarelM: Done in b37f167

HarelM:
Let me know if anything else is needed before merging this.

wayofthefuture:
Awesome!

PR #6801, opened 2025-12-02
Fix GeoJSONSource#updateData for symbol layers

### LINKED ISSUE #6795: geoJsonSource.updateData not working properly after updating to 5.13.0

lucaswoj:
Fixes #6795

## Problem

When calling `GeoJSONSource#updateData` to remove features, symbol layers would incorrectly keep displaying the removed features or hide features at the wrong times during zoom operations.

## Root Cause

The code was iterating through `tile.latestFeatureIndex.featureIndexArray` to determine which tiles needed reloading. However, `featureIndexArray` excludes symbol features. This meant that tiles containing symbol features would not be marked for reload when those features were updated or removed. 

## Solution

Changed the reload detection logic to iterate through `layers[GEOJSON_TILE_LAYER_NAME]`, which includes all features including symbol features. This ensures tiles are properly reloaded when any feature is updated, regardless of layer type.

## Testing

Added integration regression test at [test/integration/render/tests/regressions/mapbox-gl-js#6795/](vscode-webview://11tgqscajuv2gpg2sqhlq4st8cukgn7lv1bp2o529e00l2gua49h/test/integration/render/tests/regressions/mapbox-gl-js#6795/)

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Fix unwanted movement when moving a pitched terrain map at high latitudes; fix freezes when moving a pitched and rotated terrain map at low zoom ([#6775](https://github.com/maplibre/maplibre-gl-js/pull/6775)) (by [@larsmaxfield](https://github.com/larsmaxfield))
 - Fix issue with `static` modifier as part of mlt package ([#6796](https://github.com/maplibre/maplibre-gl-js/pull/6796)) (by [@HarelM](https://github.com/HarelM))
 - _...Add new stuff here..._
+- Fix `GeoJSONSource#updateData` for symbol layers ([#6801](https://github.com/maplibre/maplibre-gl-js/pull/6801) (by [@lucaswoj](https://github.com/lucaswoj)))
 
 ## 5.13.0
 


HarelM:
Since I biased towards my solution I'll let @wayofthefuture decide which solution he prefers.
Regardless, I would like to avoid updating render tests for this issue.
Also please take a look at how I changed the unit tests in geojson source to only relay on public APIs and not use "internal" methods.
You can wait with these request for changes above to a decision about which solution to use first, to avoid redundant work.

lucaswoj:
> I would like to avoid updating render tests for this issue.

I added a regression test for this bug. No existing tests were updated. We can remove the regression test if you prefer

PR #6802, opened 2025-12-02
Make `LngLatBounds#intersects` return `true` when bounds touch along an edge or at a corner 

lucaswoj:
Related to https://github.com/maplibre/maplibre-gl-js/pull/6800#issuecomment-3603436716 

`LngLatBounds#intersects` wasn't treating coincident edges as an intersection. I've updated the behavior so that bounds which touch along an edge or corner are now considered intersecting. This is more intuitive and consistent with how most spatial libraries handle intersection checks.

## Launch Checklist

 - [x] Link to related issues.
 - [x] ~Post benchmark scores.~

### DIFF
--- a/src/geo/lng_lat_bounds.ts
+++ b/src/geo/lng_lat_bounds.ts
@@ -305,13 +305,6 @@ export class LngLatBounds {
 
         if (!latIntersects) return false;
 
-        // Check if either bounds spans the full world (360 degrees or more)
-        const thisSpansWorld = this.getEast() - this.getWest() >= 360;
-        const otherSpansWorld = other.getEast() - other.getWest() >= 360;
-
-        // If either spans the full world, they intersect
-        if (thisSpansWorld || otherSpansWorld) return true;
-
         // Normalize longitudes to [-180, 180] range
         const thisWest = wrap(this.getWest(), -180, 180);
         const thisEast = wrap(this.getEast(), -180, 180);
@@ -319,8 +312,8 @@ export class LngLatBounds {
         const otherEast = wrap(other.getEast(), -180, 180);
 
         // Check if either bounds wraps around the antimeridian
-        const thisWraps = thisWest > thisEast;
-        const otherWraps = otherWest > otherEast;
+        const thisWraps = thisWest >= thisEast;
+        const otherWraps = otherWest >= otherEast;
 
         // Both wrap: they always intersect
         if (thisWraps && otherWraps) {


src/geo/lng_lat_bounds.test.ts
```diff
@@ -423,6 +423,12 @@ describe('LngLatBounds', () => {
                 const bounds = new LngLatBounds([-180, 5], [-175, 10]);
                 expect(tileBounds.intersects(bounds)).toBe(true);
             });
+
+            test('entire worlds tile should return true', () => {
+                const tileBounds = new LngLatBounds([270, -88.97061836630758], [-270, 88.97061836630758]);
+                const bounds = new LngLatBounds([0, 0], [10, 10]);
+                expect(tileBounds.intersects(bounds)).toBe(true);
+            });
```
lucaswoj, line 433: @HarelM Is this the test you're referring to? It appears to be passing on CI. Were you expecting  `expect(tileBounds.intersects(bounds)).toBe(false)`?

HarelM: Sorry, I must have fixed it when playing around with it, the following is failing:
```ts
            test('entire worlds tile should return true', () => {
                const tileBounds = new LngLatBounds([-270, -88.97061836630758], [270, 88.97061836630758]);
                const bounds = new LngLatBounds([0, 0], [0, 0]);
                expect(tileBounds.intersects(bounds)).toBe(true);
            });
```
These coordinates are the coordinates I get when I call `tileIdToLngLatBounds` for the 0,0,0 tile.

HarelM: Here a more "good looking test" that fails:
```ts
           test('entire worlds tile should return true', () => {
                const tileBounds = tileIdToLngLatBounds(new CanonicalTileID(0, 0, 0));
                const bounds = new LngLatBounds([0, 0], [0, 0]);
                expect(tileBounds.intersects(bounds)).toBe(true);
            });
```

lucaswoj: `LngLatBounds#intersects` wasn't treating coincident edges as an intersection. I've updated the behavior so that bounds which touch along an edge or corner are now considered intersecting. This is more intuitive and consistent with how most spatial libraries handle intersection checks.

src/geo/lng_lat_bounds.ts
```diff
@@ -290,19 +290,28 @@ export class LngLatBounds {
...
-            other.getSouth() > this.getNorth()
-        );
+        const latIntersects =
+            other.getNorth() >= this.getSouth() &&
+            other.getSouth() <= this.getNorth();
 
         if (!latIntersects) return false;
 
+        // Check if either bounds spans the full world (360 degrees or more)
+        const thisSpansWorld = this.getEast() - this.getWest() >= 360;
```
HarelM: Does this need "abs" to check if it spans the world in the opposite direction?

HarelM:
Thanks for looking into this. I've added a comment about one of the calculations.
Looks good otherwise.

src/geo/lng_lat_bounds.ts
```diff
@@ -290,19 +290,28 @@ export class LngLatBounds {
...
+        const otherSpansWorld = other.getEast() - other.getWest() >= 360;
+
+        // If either spans the full world, they intersect
+        if (thisSpansWorld || otherSpansWorld) return true;
```
HarelM: Actually, when I think about this a bit more, I don't think this is correct.
This can be 360 degrees east to west, but can still be a very narrow band north to south so that an area over the equator can be considered intersecting while it's not.
Or am I reading this wrong?

lucaswoj: I'm don't follow. Can you provide a specific test case that shows the case you're describing?

src/geo/lng_lat_bounds.ts
lucaswoj: As written, this method requires that "east" is greater than "west", even if the bounds span the antemeridian. AFIK this is the preferred way to handle antemeridian wrapping in MapLibre.

src/geo/lng_lat_bounds.ts
HarelM: My bad, latitude intersection is checked before this part, so if there's a narrow band around the world it would be caught earlier.
My test was the following:
```ts
            test('narrow band around the world ', () => {
                const tileBounds = new LngLatBounds([-270, -1], [270, 1]);
                const bounds = new LngLatBounds([3, 3], [4, 4]);
                expect(tileBounds.intersects(bounds)).toBe(false);
            });
```

src/geo/lng_lat_bounds.ts
HarelM: There's this method at the end of the file: `adjustAntiMeridian` for reference, I think the assumption is that you go from southwest to northeast, but there's no assumptions on which number is bigger, which is positive etc.

lucaswoj: Got it. I didn't know that was supported by our public API. This method was written primarily to support `GeoJSON#updateData` which holds that invariant. Let's address this separate issue in a separate PR.

lucaswoj: -> https://github.com/maplibre/maplibre-gl-js/issues/6804

HarelM: I don't follow why this needs a different PR...

lucaswoj: My mistake. I misread this and thought the 360 logic was present before this PR. Will fix.

lucaswoj: Fixed in [40cfc6b](https://github.com/maplibre/maplibre-gl-js/pull/6802/commits/40cfc6bb99152bada136346e865a9e7aea1c3ff5)

PR #6803, opened 2025-12-02
fix: Don't serialize GeoJSON vector tiles from the worker to support undefined properties

### LINKED ISSUE #6730: geoJSON source's updateData throws with undefined properties

**browser**: Chrome

### Steps to Trigger Behavior

 1. In the JSBin linked below, click anywhere to trigger updateData with an undefined property.

More generally this issue occurs when using updateData on a GeoJSON source and any of the feature's properties are undefined. 

### Link to Demonstration

https://jsbin.com/hiwawahida/edit?html,console,output

### Expected Behavior
Using updateData updates the features, working with undefined features just like setData does.  

This should work because the GeoJSON properties [types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ff41535d2d84f62f459fa56a9c462182ed9d1ab3/types/geojson/index.d.ts#L172) allow it. The fact that maplibre is using MVT which don't allow undefined properties under the hood should be considered an implementation detail, and maplibre should be responsible for stripping out invalid properties so that this case doesn't throw. This would match the behavior in the setData case. More detail on the reason this broke below. 

### Actual Behavior

Using updateData with an undefined property results in an uncaught error (`unknown feature value`). 

Console error:
```
at index.js:361:15
    at Mu (index.js:339:44)
    at wc.readFields (index.js:42:13)
    at new Eu (index.js:307:13)
    at Fu (index.js:385:23)
    at wc.readFields (index.js:42:13)
    at new ku (index.js:374:27)
    at wf.loadVTLayers (feature_index.ts:122:19)
    at ee.shouldReloadTile (geojson_source.ts:519:48)
    at Ie.reload (tile_manager.ts:282:91)
```

This is because we're hitting the case in mapbox/vector-tile's `readValueMessage` where the `value` read is null. In the 2.0.4 dependency, this case throws. Check out the code [here](https://www.npmjs.com/package/@mapbox/vector-tile/v/2.0.4?activeTab=code) at index.js line 360. Maplibre upgraded to this dependency in 5.6.2 via this PR https://github.com/maplibre/maplibre-gl-js/pull/6131. The previous version, 1.3.1, did not throw when encountering a null value (see https://www.npmjs.com/package/@mapbox/vector-tile/v/1.3.1?activeTab=code lib/vectortilelayer.js line 50). 

At a higher level, this likely happens because of the difference in how the "data" case is handled [here](https://github.com/maplibre/maplibre-gl-js/blob/53cb7999de2c3608f3e35cf3d6d35de5f106dcdf/src/source/geojson_source.ts#L400-L412) vs the "diff" case (corresponding to setData vs updateData).

mwilsnd:
## Launch Checklist

This fixes #6730 

By returning raw feature data from the web worker instead of serializing, we can regain support for undefined GeoJSON properties and get a nice performance boost.

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/source/worker_tile.ts
+++ b/src/source/worker_tile.ts
@@ -24,7 +24,6 @@ import type {PromoteIdSpecification} from '@maplibre/maplibre-gl-style-spec';
 import type {VectorTileLike} from '@maplibre/vt-pbf';
 import {type GetDashesResponse, MessageType, type GetGlyphsResponse, type GetImagesResponse} from '../util/actor_messages';
 import type {SubdivisionGranularitySetting} from '../render/subdivision_granularity_settings';
-
 export class WorkerTile {
     tileID: OverscaledTileID;
     uid: string | number;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Fix `queryTerrainElevation` to use higher zoom level tiles when possible ([#6791](https://github.com/maplibre/maplibre-gl-js/issues/6791)) (by [@HarelM](https://github.com/HarelM))
 - Fix unwanted movement when moving a pitched terrain map at high latitudes; fix freezes when moving a pitched and rotated terrain map at low zoom ([#6775](https://github.com/maplibre/maplibre-gl-js/pull/6775)) (by [@larsmaxfield](https://github.com/larsmaxfield))
 - Fix issue with `static` modifier as part of mlt package ([#6796](https://github.com/maplibre/maplibre-gl-js/pull/6796)) (by [@HarelM](https://github.com/HarelM))
+- `LngLatBounds#intersects` now returns `true` when bounds touch along an edge or at a corner ([#6802](https://github.com/maplibre/maplibre-gl-js/pull/6802)) (by [@lucaswoj](https://github.com/lucaswoj))
 - _...Add new stuff here..._
 
 ## 5.13.0
--- a/src/geo/lng_lat_bounds.test.ts
+++ b/src/geo/lng_lat_bounds.test.ts
@@ -1,6 +1,8 @@
 import {describe, test, expect} from 'vitest';
 import {LngLat} from './lng_lat';
 import {LngLatBounds} from './lng_lat_bounds';
+import {tileIdToLngLatBounds} from '../tile/tile_id_to_lng_lat_bounds';
+import {CanonicalTileID} from '../tile/tile_id';
 
 describe('LngLatBounds', () => {
     test('constructor', () => {
@@ -421,6 +423,12 @@ describe('LngLatBounds', () => {
             test('wrapping tile bounds at dateline intersects with negative longitude bounds', () => {
                 const tileBounds = new LngLatBounds([170, 0], [-170, 10]);
                 const bounds = new LngLatBounds([-180, 5], [-175, 10]);
+                expect(tileBounds.intersects(bounds)).toBe(true);
+            });
+
+            test('entire worlds tile should return true', () => {
+                const tileBounds = tileIdToLngLatBounds(new CanonicalTileID(0, 0, 0));
+                const bounds = new LngLatBounds([0, 0], [0, 0]);
                 expect(tileBounds.intersects(bounds)).toBe(true);
             });
         });
--- a/src/geo/lng_lat_bounds.ts
+++ b/src/geo/lng_lat_bounds.ts
@@ -290,16 +290,18 @@ export class LngLatBounds {
     /**
      * Checks if this bounding box intersects with another bounding box.
      *
+     * Returns true if the bounding boxes share any area, including cases where
+     * they only touch along an edge or at a corner.
+     *
      * This method properly handles cases where either or both bounding boxes cross
      * the antimeridian (date line).
      */
     intersects(other: LngLatBoundsLike): boolean {
         other = LngLatBounds.convert(other);
 
-        const latIntersects = !(
-            other.getNorth() < this.getSouth() ||
-            other.getSouth() > this.getNorth()
-        );
+        const latIntersects =
+            other.getNorth() >= this.getSouth() &&
+            other.getSouth() <= this.getNorth();
 
         if (!latIntersects) return false;
 
@@ -310,8 +312,8 @@ export class LngLatBounds {
         const otherEast = wrap(other.getEast(), -180, 180);
 
         // Check if either bounds wraps around the antimeridian
-        const thisWraps = thisWest > thisEast;
-        const otherWraps = otherWest > otherEast;
+        const thisWraps = thisWest >= thisEast;
+        const otherWraps = otherWest >= otherEast;
 
         // Both wrap: they always intersect
         if (thisWraps && otherWraps) {
@@ -329,10 +331,7 @@ export class LngLatBounds {
         }
 
         // Neither wraps: standard intersection check
-        return !(
-            otherWest > thisEast ||
-            otherEast < thisWest
-        );
+        return otherWest <= thisEast && otherEast >= thisWest;
     }
 
     /**


src/source/worker_tile.ts
```diff
@@ -43,6 +45,7 @@ export class WorkerTile {
 
     abort: AbortController;
     vectorTile: VectorTileLike;
+    geoJsonFeatures?: Feature[];
```
HarelM: Can't this be inferred from the `vectorTile`? It is mandatory to add this field?

mwilsnd: I think so - with the update to vt-pbf I can likely simplify things further.

HarelM: It doesn't look it is being used in the parse method, so I wonder if this can be removed from `WorkerTile` class...

src/source/vector_tile_worker_source.ts
```diff
@@ -140,6 +141,7 @@ export class VectorTileWorkerSource implements WorkerSource {
             }
 
             workerTile.vectorTile = response.vectorTile;
+            workerTile.geoJsonFeatures = response.isGeojson ? (response.vectorTile as GeoJSONWrapper).features : null;
```
HarelM: I wonder if this can be done better considering we have inheritance here with geojson and vector tile...

mwilsnd: Is reflecting the type of an object here more or less the same, performance-wise? No issues with changing this.

HarelM: Less about reflection, more about OOP... if you are doing `as GeoJSONWrapper` in a class that doesn't suppose to know about `GeoJSONWrapper` then it seems weird...

src/source/geojson_worker_source.test.ts
```diff
@@ -32,6 +32,11 @@ describe('reloadTile', () => {
             'geometry': {
                 'type': 'Point',
                 'coordinates': [0, 0]
+            },
+            'properties': {
```
HarelM, line 36: Probably better to split this out to a different test...

HarelM:
I can't say I'm loving this, seems to heckish, but I can't place my finger on a specific issue, what other alternatives do we have to solve the bug?

src/source/geojson_worker_source.ts
```diff
@@ -94,7 +93,10 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
 
         const geojsonWrapper = new GeoJSONWrapper(geoJSONTile.features, {version: 2, extent: EXTENT});
 
-        return toVirtualVectorTile(geojsonWrapper);
+        return {
+            vectorTile: geojsonWrapper,
+            isGeojson: true
+        };
```
lucaswoj, line 96: When we convert the GeoJSON data to a vector tile here, the data is encoded as a `UInt8Array`. This can be transferred to the main thread with zero overhead via the WebWorker `transfer` API. 

The plain GeoJSON data cannot take advantage of the `transfer` API and will be much slower. 

See https://developer.chrome.com/blog/transferable-objects-lightning-fast/ for more info.

lucaswoj: Tested against my performance benchmark (add 100k points, move one, https://jsbin.com/bubejanoha/edit?html,output) this branch is ~100% slower than main.

mwilsnd:
If it's the case that copying the raw feature data is slower than serializing the tile to protobuf, then the remaining options to fix this issue are a bit more involved:
* Custom binary serialization for GeoJSON worker tiles
* Perform feature tile culling on the main thread and pass those results to the feature index (we already get feature culling from geojsonvt on the worker, which has to run anyways for creating the buckets)
* Revert to mapbox/vector-tile-js v2.0.3 and pin it there, accept operating outside the vector tile specification
Or highly unlikely: petition to review the vector tile specification with support for undefined properties

I'm not certain what the best approach here is, we can't get away with just serializing to MVT anymore unless we pin an old version and stay there or get the spec changed.

lucaswoj:
Probably not a small lift, but can we serialize to MLT now?

HarelM:
No, we don't have a js derializer for mlt at the moment... :-/

lucaswoj:
I'm just getting up to speed on this issue, so I apologize if this is already answered somewhere.

Why can't we simply filter out any GeoJSON property keys that have `undefined` or `null` values?

mwilsnd:
While filtering would avoid serializing the now unsupported properties, that would result in those values not making it into the feature index. Since the GeoJSON  spec explicitly supports null/undefined properties, we want to see those keys persist in `queryRenderedFeatures`.

HarelM:
Because these are valid geojson definitions...
Not a great argument, but still...

HarelM:
I'm guessing mapbox should have the same issue when I think about it.
Have we tried opening an issue in the relevant repo? 
Maybe this can be made as an optional flag to the encoder so that by default it throws but it can be configured not to?

lucaswoj:
Did we return `null` / `undefined` from `queryRenderedFeatures` before https://github.com/maplibre/maplibre-gl-js/pull/6131? 

Is the goal of this PR restore pre-https://github.com/maplibre/maplibre-gl-js/pull/6131 behavior? or to expand our support for the GeoJSON spec in a new way?

HarelM:
As a side note, performance is less important than "error handling", i.e. if we need to solve this and the only way is by degrading performance, then we'll need to bite the bullet.
In the past this used to work without throwing exceptions, I'm not 100% sure that was the behavior, but it should be easy to check.

mwilsnd:
Undefined properties were previously supported. vt-pbf explicitly handles `undefined` so this has been the status quo for quite a while. I'm not sure what prompted Mapbox to make the change in vector-tile-js, or why it was done in a minor version increment, but I'd like to restore the old behavior.

src/source/geojson_worker_source.ts
mwilsnd: Running your tool, I'm seeing about a 15% regression on my system over an average of 40 samples in Firefox. Could you add this test to the benchmarks? I think it'd be really valuable.

lucaswoj: I've gotta step out for the afternoon but I'll try to get back to these questions in the next ~12 hrs.

lucaswoj: > Running your tool, I'm seeing about a 15% regression on my system over an average of 40 samples in Firefox. 

Are you sure you're running `npm run build-dist` on this branch and on `main`? The dist build is not generated automatically as part of the dev server. I'm certain that you'll see a measurable and noticeable difference. 

> Could you add this test to the benchmarks? I think it'd be really valuable.

I added a `GeoJSONDiff` benchmark that measures perf on a relatively small (10k features) source. You can manually edit the bench to test a large (100k features) source but the runtime is prohibitively long to include in the standard suite.

lucaswoj:
> Revert to mapbox/vector-tile-js v2.0.3 and pin it there, accept operating outside the vector tile specification

Are there any immediate consequences to doing this? I can help implement a longer term solution if this is an acceptable temporary bandaid. 

I've spent about 100 hours optimizing `GeoJSON#updateData` performance over the past few months. I am highly motivated to help find a solution that both preserves that perf improvement for me and fixes this behavior for you.

HarelM:
A user who needs this fixed can simply rollback to previous msplibre version, so we can take the time and fix this properly. 

Is it possible to use JSON.strinfigy with TextEncoder to pass the geojson data to the main thread instead of coverting it to MVT? There are other limitations of MVT that do no apply to geojson (like objects in properties) that I'll be happy to solve too.

HarelM:
If we are willing to entertain other formats, we can have a look at geobuf and geoarrow.
I don't know the implication of package size though...

ingalls:
@HarelM We've been benefiting enormously from the performance question on GeoJSON sources that @lucaswoj has been working on over the last couple of months. We have 10-100k points updated on a regular basis as we have live first responder locations for all of Colorado, as such we are likely more sensitive to performance loss as we've been working hard to try to get things as smooth as possible.

As such if the `undefined`/`null` values are something that we want to support moving forward, would it make more sense to simply fork the vector-tile-js library or pin it to the previous versions as a stop gap?

lucaswoj:
Another thought: the `GeoJSONSource` already has all the raw GeoJSON features on the main thread (`GeoJSONSource._data`), so we could read the properties directly from there instead of sending it b back and forth between threads.

src/data/feature_index.ts
```diff
@@ -114,9 +115,13 @@ export class FeatureIndex {
 
     loadVTLayers(): {[_: string]: VectorTileLayerLike} {
         if (!this.vtLayers) {
-            this.vtLayers = this.encoding !== 'mlt' 
-                ? new VectorTile(new Protobuf(this.rawTileData)).layers
-                : new MLTVectorTile(this.rawTileData).layers;
+            if (this.geoJsonFeatureData) {
+                this.vtLayers = new GeoJSONWrapper(this.geoJsonFeatureData, {version: 2, extent: EXTENT}).layers;
```
lucaswoj, line 119: How long does constructing `GeoJSONWrapper` take for a source with 100k points? We used to do this step on the worker thread.

HarelM: It's only a wrapper, it doesn't take right...

HarelM:
I think we came to the conclusion this is not the case for data set as irl, didn't we?

lucaswoj:
> I think we came to the conclusion this is not the case for data set as irl, didn't we?

It wouldn't be very hard to send the full GeoJSON data to the main thread in just that case

src/data/feature_index.ts
lucaswoj: Ah, right. I thought this was actually constructing some kind of vector tiles.

lucaswoj:
EDIT: moved to thread https://github.com/maplibre/maplibre-gl-js/pull/6803#discussion_r2589834500

HarelM:
Fetching the features on the main thread would still need to slice them up to tiles for querying which is also heavy. 
It maybe can be done lazily...

src/source/geojson_worker_source.ts
lucaswoj: I propped up a quick demo so that we can see the perf difference between this branch (https://lucaswoj.github.io/demo-6803/branch/index.html) and `main` (https://lucaswoj.github.io/demo-6803/main/index.html). 

I'm seeing this branch take about 1200ms for an update whereas main takes about 490ms, so it's about 2.5x slower on this branch.

If we can re-use the GeoJSON data already on the main thread rather than sending it from the worker with every tile, this change could be performance neutral. We will need to send the data from the worker to the main thread from `loadData` if the user is loading from a URL but that's not going to affect anyone in practice.

HarelM: Fetching the features on the main thread would still need to slice them up to tiles for querying which is also heavy.
It maybe can be done lazily... IDK...

lucaswoj: In #6810 I improved the automated benchmarks so that they can catch regressions like this:

Benchmark | #6810 | This branch
-- | -- | --
`GeoJSONSourceUpdateData` | 91.69 ms | 269.65 ms
`GeoJSONSourceSetData` | 269.65 ms | 645.79 ms

lucaswoj:
> Fetching the features on the main thread would still need to slice them up to tiles for querying which is also heavy. It maybe can be done lazily...

Could we keep using `FeatureIndex` for the spatial query, then look up the original feature properties by ID from `GeoJSONSource#_data.updatable` at the end? That way we get the undefined/null properties without re-slicing on the main thread.

lucaswoj:
> I think we came to the conclusion this is not the case for data set as irl, didn't we?

Just created https://github.com/maplibre/maplibre-gl-js/pull/6811 to provide the GeoJSON data back to the main thread when it's loaded from a URL

src/source/geojson_worker_source.ts
mwilsnd: Thanks! Prior to merging main my local branch was behind enough to not include your GeoJSONDiff benchmark. Not many great options here seeing this regression. Indirectly looking up properties after querying the feature index might be the best approach if it works.

lucaswoj: > Thanks! Prior to merging main my local branch was behind enough to not include your GeoJSONDiff benchmark.

If you pull in the latest changes from `main` you can compare this branch to main with one command. It takes a minute to run. Good time to get a glass of water, bother your pet, etc.

```
npm run benchmark -- GeoJSONSourceUpdateData && npm run benchmark -- GeoJSONSourceSetData 
```

> Indirectly looking up properties after querying the feature index might be the best approach if it works.

I think we can make it work! The one big hurdle to navigate will be supporting GeoJSON sources with string feature IDs or no feature IDs. I'm happy to chat more, help brainstorm or prototype but I respect that you're the primary issue owner here.

lucaswoj: I think there are a couple other performance solutions worth considering...

 1. pin to the previous `vector-tile-js` version 
 2. fork `vector-tile-js` to maintain support for `null`
 3. accept the breaking change (how many users are _really_ differentiating between `null` and `undefined` in `queryRenderedFeatures` results?)
 4. use a "sentinel" value to represent `null` in vector tiles (like the string `"__MAPLIBRE_NULL_VALUE__"`)

mwilsnd: 3 & 4 are certainly worth considering but I'm hesitant about using an older or modified version of vector-tile-js. For better or worse the current in-memory tile representation used is coupled to MVT, trying to ignore the specification (even if a minor deviation) doesn't sound too appealing to me.

While I question the usefulness of undefined/null GeoJSON properties, that specification allows for them so if possible I'd prefer a solution that complies with both formats. Other options like custom serialization, a forked vector tile deserializer or tile feature culling on the main thread would all add more complexity compared to a simple sentinel, so I'm currently leaning more towards that approach. An indirect lookup from the feature index would be the runner up, with the perk of not needing a special reserved property value.

@HarelM What are your thoughts on the possible approaches here? If we're fine introducing another reserved string then this whole PR becomes a lot more simple.

lucaswoj: >  that specification allows for them so if possible I'd prefer a solution that complies with both formats

`queryRenderedFeatures` has never supported the GeoJSON specification. This project's stance has consistently been that GeoJSON gets converted to a vector tile internally and that's what comes out of qRF. For example, it's never supported arrays or nested objects. Support for the `null` value was a quirk and does not indicate a broader commitment to full support of the GeoJSON specification.

HarelM: That's true, but there are a lot of bugs/feature requests on that topic, people keep finding it cumbersome.
If we could use feature index and be able to return the original feature with the original geometry from updatable that could be a significant win in terms of user expectations as users don't really care that we use MVT under the hood, they want their data back the same way they entered it. 
But this might be a breaking change, for good, but still a breaking change. 
If we manage to do it properly it's a good reason to start v6 breaking changes version. 
So if this is possible I'm in favor of solving geojson quirkiness once and for all.

HarelM:
@lucaswoj that's a very interesting solution, see my comment here: 
https://github.com/maplibre/maplibre-gl-js/pull/6803#discussion_r2591591956

src/source/geojson_worker_source.ts
```diff
@@ -92,9 +91,9 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
             return null;
         }
 
-        const geojsonWrapper = new GeoJSONWrapper(geoJSONTile.features, {version: 2, extent: EXTENT});
-
-        return toVirtualVectorTile(geojsonWrapper);
+        return {
+            vectorTile: new GeoJSONWrapper(geoJSONTile.features, {version: 2, extent: EXTENT})
+        };
```
wayofthefuture, line 96: Here we are returning the GeoJSONWrapper, however the querySourceFeatures test in tile.test.ts still uses a converted tile, possibly creating a false negative test result.

```ts
    describe('geojson tile', () => {
        const tile = new Tile(new OverscaledTileID(3, 0, 2, 1, 2), undefined);
        const geojsonWrapper = new GeoJSONWrapper(features);
        geojsonWrapper.name = GEOJSON_TILE_LAYER_NAME;
        tile.loadVectorData(
            createVectorData({rawTileData: fromVectorTileJs({layers: {[GEOJSON_TILE_LAYER_NAME]: geojsonWrapper}})}),
            createPainter()
        );
```

I have been messing around with this for a while now, and I think the current implementation in this PR would completely break querySourceFeatures for GeoJSON sources. I believe the test would need to be modified as well:

```ts
        tile.loadVectorData(
            createVectorData({vectorTile: geojsonWrapper}),
            createPainter()
        );
```

Or something like that, not sure though... I think tile.test.ts needs many more tests surrounding querySourceFeatures imo.

src/util/vectortile_to_geojson.ts
```diff
@@ -75,7 +75,9 @@ export class GeoJSONFeature {
             case 1: {
                 const points = [];
                 for (const line of vtCoords) {
-                    points.push(line[0]);
+                    for (const point of line) {
+                        points.push(point);
+                    }
```
wayofthefuture, line 80:
```ts
                // Vector tile point/multipoint both reside as nested array, however geojson point is not nested.
                const points = vtCoords[0];
                const coordinates = this.projectLine(points, x0, y0, size);
                this._geometry = points.length === 1 ?
                    {type: 'Point', coordinates: coordinates[0]} :  // Point is not nested like a vector tile point
                    {type: 'MultiPoint', coordinates};
                break;
```

the feature coordinate at the top of tile.test.ts should instead be???

```ts
    geometry: [[0, 0]]
```

wayofthefuture:
It's hard to imagine where anything other than the arraybuffer is not passed here...

```ts
        this.actor.registerMessageHandler(MessageType.loadTile, (mapId: string, params: WorkerTileParameters) => {
            return this._getWorkerSource(mapId, params.type, params.source).loadTile(params);
        });
```

wayofthefuture:
https://github.com/mapbox/vector-tile-js/issues/89

wayofthefuture:
@HarelM @mwilsnd In MapLibre vt-pbf, we have the following [line](https://github.com/maplibre/vt-pbf/blob/d8da4b7e13869b6ca15f2eee9fff629c55de2219/index.ts#L100):

```ts
    if (value === null) continue // don't encode null value properties
````

If we change this to:

```ts
    if (value == null) continue // don't encode null/undefined value properties
````

Could this be an easy solution to the problem in this PR? We are currently skipping null values so it makes sense to also skip undefined. If a property is undefined then querying it would return nullish anyway so undefined would then be supported.

**Comment by mwilsnd** [comment]
That would solve the immediate issue of vector-tile-js complaining, however it would also drop the keys associated with undefined values like in the simple filtering approach. That means keys with undefined values would be missing in a call to queryRenderedFeatures. Maybe not a big deal but it's a deviation from what the library has been doing in the past.

Discussing this topic during the monthly meeting, it seems like the more ideal long term solution would be to move towards MLT for our in-memory handling of GeoJSON. We don't yet have a JS deserializer for MLT so I'm open to a more simple workaround in the short to medium term while the MLT ecosystem improves. Something like the sentinel approach would restore the previous behavior users expect without the performance regression, buying time for us to push for the required MLT features to fully migrate off of the current MVT pathway. Future revisions of MLT would then potentially enable support for more of the GeoJSON spec like nested properties.

File: src/util/vectortile_to_geojson.ts (hunk 7 above)
Comment (mwilsnd) [reply src/util/vectortile_to_geojson.ts:80]: Yeah, I found that a bit surprising as well, but the geometry returned was in a different layout.

**Comment by wayofthefuture** [comment]
https://github.com/maplibre/vt-pbf/pull/16

**Comment by wayofthefuture** [comment]
> That would solve the immediate issue of vector-tile-js complaining... Maybe not a big deal but it's a deviation from what the library has been doing in the past.

IMO it's not complaining it's a throw that breaks GeoJSON sources via https://github.com/maplibre/maplibre-gl-js/pull/6131 for many implementations: which is the relevant deviation. I think priority 1 is that it works.

```ts
const o1 = {};
const o2 = {key: undefined};
o1['key'] === o2['key']; //true
```

Also I'm not sure about the state of MLT but it likely need's some time to mature before pursuing it for the worker... otherwise it slows the development of both.

HarelM:
Is this PR still relevant given the fix in vt-pbf?

mwilsnd:
Yeah, this shouldn't be needed right now. Closing, at least until we get better JS MLT serialization and we can revisit this.

PR #6811, opened 2025-12-04
Send GeoJSON data from worker to main if loaded from a URL

lucaswoj:
## Summary

Part of https://github.com/maplibre/maplibre-gl-js/issues/4364

This PR carves out an exception from the optimization added in #6668. When GeoJSON data is loaded from a URL, we now send it back to the main thread so that it can be:
- Returned from `map.getStyle()`
- Used in `GeoJSONSource#_getShouldReloadTileOptions` for faster `GeoJSONSource#updateData` operations (ref #6800)
- (Maybe someday) used in `queryRenderedFeatures` to return original feature properties (ref #6803)

When GeoJSON is provided inline or via diff updates, the main thread already has the data, so we skip sending it back (preserving the #6668 optimization).

This will make `GeoJSONSource#updateData` much faster when the GeoJSON data was loaded from a URL.

## Benchmarks

```
$ npm run benchmark -- GeoJSONSourceUpdateData && npm run benchmark -- GeoJSONSourceSetData 

> maplibre-gl@5.14.0 benchmark
> node --no-warnings --loader ts-node/esm test/bench/run-benchmarks.ts GeoJSONSourceUpdateData

Starting headless chrome at: http://localhost:9966/test/bench/versions/index.html
                                      v5.14.0                  main  set-data-url 85c1f56 
 GeoJSONSourceUpdateData                                92.3814 ms            92.6344 ms 

> maplibre-gl@5.14.0 benchmark
> node --no-warnings --loader ts-node/esm test/bench/run-benchmarks.ts GeoJSONSourceSetData

Starting headless chrome at: http://localhost:9966/test/bench/versions/index.html
                                   v5.14.0                  main  set-data-url 85c1f56 
 GeoJSONSourceSetData                               279.7234 ms           284.3850 ms 
```

## Launch Checklist

- [x] Link to related issues.
- [x] Post benchmark scores.

### DIFF
--- a/src/source/geojson_source.test.ts
+++ b/src/source/geojson_source.test.ts
@@ -506,7 +506,7 @@ describe('GeoJSONSource.update', () => {
         source.load();
         await promise;
 
-        expect((source as any)._data.geojson).toStrictEqual(hawkHill);
+        expect(source.serialize()).toStrictEqual({type: 'geojson', data: hawkHill});
     });
 
     test('fires event when metadata loads', async () =>  {
--- a/src/source/geojson_source.test.ts
+++ b/src/source/geojson_source.test.ts
@@ -497,34 +497,18 @@ describe('GeoJSONSource.update', () => {
     });
 
     test('updates _data.geojson when worker returns data from URL load', async () => {
-        const returnedData = {
-            type: 'FeatureCollection',
-            features: [{type: 'Feature', properties: {}, geometry: {type: 'Point', coordinates: [0, 0]}}]
-        } as GeoJSON.GeoJSON;
-
-        const mockDispatcher = wrapDispatcher({
-            sendAsync(_message: ActorMessage<MessageType>) {
-                return new Promise((resolve) => {
-                    setTimeout(() => resolve({data: returnedData}), 0);
-                });
-            }
-        });
-
-        const mapStub = {
-            _requestManager: {
-                transformRequest: (url) => ({url})
-            }
-        } as any;
-
-        const source = new GeoJSONSource('id', {data: 'https://example.com/data.geojson'} as GeoJSONSourceOptions, mockDispatcher, undefined);
-        source.map = mapStub;
+        const source = new GeoJSONSource('id', {data: 'https://example.com/data.geojson'} as GeoJSONSourceOptions, wrapDispatcher({
+            sendAsync() { return Promise.resolve({data: hawkHill}); }
+        }), undefined);
+        source.map = {_requestManager: {transformRequest: (url) => ({url})}} as any;
 
         const promise = waitForEvent(source, 'data', (e: MapSourceDataEvent) => e.sourceDataType === 'metadata');
         source.load();
         await promise;
 
-        expect((source as any)._data.geojson).toStrictEqual(returnedData);
+        expect((source as any)._data.geojson).toStrictEqual(hawkHill);
     });
+
     test('fires event when metadata loads', async () =>  {
         const mockDispatcher = wrapDispatcher({
             sendAsync(_message: ActorMessage<MessageType>) {
--- a/src/source/geojson_source.test.ts
+++ b/src/source/geojson_source.test.ts
@@ -495,6 +495,36 @@ describe('GeoJSONSource.update', () => {
         expect(transformSpy).toHaveBeenCalledTimes(1);
         expect(transformSpy.mock.calls[0][0]).toBe('https://example.com/data.geojson');
     });
+
+    test('updates _data.geojson when worker returns data from URL load', async () => {
+        const returnedData = {
+            type: 'FeatureCollection',
+            features: [{type: 'Feature', properties: {}, geometry: {type: 'Point', coordinates: [0, 0]}}]
+        } as GeoJSON.GeoJSON;
+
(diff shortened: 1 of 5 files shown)


HarelM:
How about fetching it in the main thread instead of serializing it from the worker? 
I think the browser will cache the response so the worker won't really fetch it again.
This is similar to what we did in the rtl plugin fetching.

lucaswoj:
There's no way to guarantee the browser will cache the response. If the response has a `Cache-Control: no-store` header, for example, it will certainly be fetched twice.

src/source/geojson_source.test.ts
```diff
@@ -495,6 +495,20 @@ describe('GeoJSONSource.update', () => {
...
+        const source = new GeoJSONSource('id', {data: 'https://example.com/data.geojson'} as GeoJSONSourceOptions, wrapDispatcher({
+            sendAsync() { return Promise.resolve({data: hawkHill}); }
+        }), undefined);
+        source.map = {_requestManager: {transformRequest: (url) => ({url})}} as any;
+
+        const promise = waitForEvent(source, 'data', (e: MapSourceDataEvent) => e.sourceDataType === 'metadata');
+        source.load();
+        await promise;
+
+        expect((source as any)._data.geojson).toStrictEqual(hawkHill);
```
HarelM: Is there a different way to check this using more public APIs and hopefully avoid `as any` as well?

src/source/geojson_source.ts
```diff
@@ -424,7 +424,8 @@ export class GeoJSONSource extends Evented implements Source {
                 return;
             }
 
-            if (diff) this._applyDiff(diff);
+            if (result.data) this._data.geojson = result.data;
```
HarelM: Should this be `this._data = { geojson: result.data }` in order to "clear" the others properties?

HarelM:
Does it make sense now to return the data in `getData` method not from the worker but from the main thread class?

lucaswoj:
> Does it make sense now to return the data in `getData` method not from the worker but from the main thread class?

As we discussed in #6738,  there are still cases where `GeoJSONSource#getData` returns a different value than `GeoJSONSource#serialize`

> ... if you called `setData` with some improperly wound GeoJSON geometries, and then called `Map#getStyle`, you'd get the rewound geometries back. Even though this is undocumented behavior, I suspect at least a couple folks are relying on it. They will need to use the `GeoJSONSource#getData` method instead.

I think it's worth having this method around for backwards compatibility. 

(Sidenote: If and when we're ready to make some breaking changes, I'd love to see `GeoJSONSource#serialize` and `Map#getStyle` become async and pull this giant dataset from the worker only when it's requested)

src/source/geojson_source.ts
lucaswoj: Done in 6af1a9b22

src/source/geojson_source.test.ts
lucaswoj: Done in 6af1a9b22

HarelM:
If you have ideas for breaking changes please add them to the following issue: 
- #6427

HarelM:
I'm not sure I'm following the last comment. If we now have the data on the main thread, as the user set it, what's the point of going to the worker to get it?
It can be out of scope for this PR but I don't understand why not do it.

lucaswoj:
I'll try to explain a different way: `GeoJSONSource#getData` doesn't return the data exactly "as the user set it." It returns data that has undergone some additional cleanup (namely, rewinding geometries via the `geojson-rewind`  package and possibly filtering via the `GeoJSONSourceOptions#filter` option). Our documentation doesn't explicitly say that `GeoJSONSource#getData` returns rewound geometries but it's possible that some users really depend on that. I see no reason to break `GeoJSONSource#getData` for them.

lucaswoj:
Oh! I might've explained the wrong part:

>  If we now have the data on the main thread, as the user set it, what's the point of going to the worker to get it?

The only reason we _must_ keep the full GeoJSON data on the main thread is to support the syncronous `GeoJSONSource#serialize` method. If that method were async, it's possible we could entirely get rid of the need to keep it in memory on the main thread. In some cases, this could save hundreds of megabytes of memory usage.

HarelM:
> The only reason we must keep the full GeoJSON data on the main thread is to support the syncronous GeoJSONSource#serialize method

I thinks this is no longer true given the recent changes with reload tile and the possible solution for undefined properties, or am I understanding this wrong?

HarelM:
In any case, we can always change that in a future PR, this code can be merged regardless of this discussion.

HarelM (approved):

lucaswoj:
> I thinks this is no longer true given the recent changes with reload tile and the possible solution for undefined properties, or am I understanding this wrong?

It could easily be true if we merged #6801 instead of #6800; and chose one of the _other_ possible solutions for undefined properties in #6803

HarelM:
> (Sidenote: If and when we're ready to make some breaking changes, I'd love to see GeoJSONSource#serialize and Map#getStyle become async and pull this giant dataset from the worker only when it's requested)

@lucaswoj we are now in the process of a version 6 breaking changes pre-release, so now would be a good time to do that if this is something you would like to push through. Keep me posted.

PR #6819, opened 2025-12-05
Deprecate isUpdateableGeoJSON in favor of cleaner input validation

wayofthefuture:
This PR removes the isUpdateableGeoJSON() function and consolidates validation and normalization logic into toUpdateable(). Previously, correctness depended on keeping two separate functions in sync, which risked logic drift and duplicated checks. By handling both validation and conversion in a single place, the update pipeline becomes simpler, more predictable, and easier to maintain. The caller now has a single, clear entry point: attempt to build an updateable map, and rely on undefined to signal invalid input.


HarelM (approved):
Yeah, I can see how this is cleaner...

PR #6835, opened 2025-12-08
Add funding-manifest-urls file

louwers:
Creating this file is required for applying to [FLOSS/fund](https://floss.fund/).

For example Python, https://www.python.org/funding.json

https://github.com/python/cpython/blob/main/.well-known/funding-manifest-urls

More info: https://fundingjson.org/

@HarelM If you don't want this file in `main`, I can add it to a branch, but then there would always be that branch that needs to be open. So I figured just committing it is a better solution.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -44,7 +44,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@fdbfb4d2750291e159f0156def62b853c2798ca2 # v4.31.5
+      uses: github/codeql-action/init@cf1bb45a277cb3c205638b2cd5c984db1c46a412 # v4.31.7
       with:
         languages: ${{ matrix.language }}
         # If you wish to specify custom queries, you can do so here or in a config file.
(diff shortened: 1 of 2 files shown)


louwers:
PR that adds `funding.json`: https://github.com/maplibre/maplibre.github.io/pull/502

HarelM:
Can you explain why `.well-known` folder is needed? Can't we use a regular `.md` file like we have for other stuff?
Isn't `.well-known` a part of a website?
Should this be added to `gh-pages` branch maybe?

louwers:
> Should this be added to gh-pages branch maybe?

Nope, it needs to exist at URL that starts with `https://github.com/maplibre/maplibre-gl-js`

image

HarelM:
This still applies to github pages branch:
https://github.com/maplibre/maplibre-gl-js/tree/gh-pages
It will also place it under 
maplibre.org/maplibre-gl-js/.well-known
But this is a good side effect, isn't it?

louwers:
> But this is a good side effect, isn't it?

It's not really relevant that it also exists on `maplibre.org`.

> This still applies to github pages branch:

Yes, so that would work as well. Commits are not usually made to the `gh-pages` branch directly though.

Let me know what you prefer, other projects seem to add it directly to `main` and that's what I did for MapLibre Native as well now.

HarelM:
Can we place it under the root folder instead of .well-known folder?

louwers:
@HarelM How about this?

HarelM:
Sure, why not...

PR #6870, opened 2025-12-16
Move to use reusable workflows

HarelM:
This PR moves the CI to use the reusable workflows.
It moves the benchmark builds to the relevant action instead of the release assuming the release will create a tag (which is what it does).
It deletes all the scripts that are now part of the reusable workflows.


HarelM:
I'll make sure these are working after merged obviously, but this is almost the last repo I'm planning to integrate this into, so these are battle tested by now.

PR #6873, opened 2025-12-17
Add test for undefined GeoJSON properties

### LINKED ISSUE #6730: geoJSON source's updateData throws with undefined properties

mwilsnd:
## Launch Checklist

- Fixes #6730

With vt-pbf updated to 4.2.0, this PR adds a simple test to ensure null & undefined properties don't break worker tile serialization.

 - [x] Link to related issues.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Fix `LngLatBounds#intersects` returning `false` for bounds spanning 360° or more ([#6863](https://github.com/maplibre/maplibre-gl-js/pull/6863)) (by [@lucaswoj](https://github.com/lucaswoj))
 - Fix getting the right zoom for getElevationForLngLat ([#6825](https://github.com/maplibre/maplibre-gl-js/pull/6825)) (by [@HarelM](https://github.com/HarelM))
 - Fix stale transform state being applied after changing `minZoom` or `maxZoom` due to old copies from `transformCameraUpdate` taking precedence. `transformCameraUpdate` is now called from `setMinZoom` and `setMaxZoom` to allow user to control subsequent changes to `zoom` [#6766](https://github.com/maplibre/maplibre-gl-js/issues/6766) (by [@Auspicus](https://github.com/Auspicus))
+- Fix GeoJSON source throwing with undefined properties [#6730](https://github.com/maplibre/maplibre-gl-js/issues/6730) (by [@wayofthefuture](https://github.com/wayofthefuture))
 - _...Add new stuff here..._
 
 ## 5.14.0


src/source/geojson_worker_source.test.ts
```diff
@@ -66,6 +66,51 @@ describe('reloadTile', () => {
...
+        const data = await source.reloadTile(tileParams as any as WorkerTileParameters);
+        expect(spy).toHaveBeenCalledTimes(1);
+        expect(data.featureIndex).toBeDefined();
+
+        // deserialize tile layers in the feature index
+        data.featureIndex.rawTileData = data.rawTileData;
+        const featureLayers = data.featureIndex.loadVTLayers();
+        expect(Object.keys(featureLayers)).toHaveLength(1);
+
+        // validate features from the index don't contain undefined or null properties
```
HarelM: I'm not sure this is the behavior we would like to keep in the future.
I believe this test mainly needs to verify that there's no crash as reported in the bug...

src/source/geojson_worker_source.test.ts
```diff
@@ -66,6 +66,51 @@ describe('reloadTile', () => {
...
+        expect(featureLayers['_geojsonTileLayer'].feature(0).properties).toStrictEqual({'stringProperty': 'string'});
```
HarelM: This layer has a constant definition we should use.

src/source/geojson_worker_source.test.ts
```diff
@@ -66,6 +66,51 @@ describe('reloadTile', () => {
...
+        const layers = [
+            {
+                id: 'mylayer',
+                source: 'sourceId',
+                type: 'symbol',
+            }
+        ] as LayerSpecification[];
+        const layerIndex = new StyleLayerIndex(layers);
+        const source = new GeoJSONWorkerSource(actor, layerIndex, []);
+        const spy = vi.spyOn(source, 'loadVectorTile');
```
HarelM: Is this mandatory or can this be removed?

mwilsnd: While not mandatory, it does sanity check that tile serialization is taking place. The presence of `rawTileData` should be sufficient however, so I can simplify.

src/source/geojson_worker_source.test.ts
mwilsnd: Sure, I can relax this check. I tend to prefer more explicit tests but there's likely little harm in being so pedantic here.

src/source/geojson_worker_source.test.ts
mwilsnd: Fixed

HarelM:
Thanks! Can you add a changelog entry for this so that it is clear the relevant bug was solved in this version?

mwilsnd:
Done. I've attributed this entry to @wayofthefuture for their PR on vt-pbf.

HarelM (approved):

HarelM:
Thanks! I'll merge this when the CI finishes.

wayofthefuture:
Thanks!

PR #6874, opened 2025-12-17
Add back release action name, fix benchmark upload

HarelM:
See title.


HarelM:
It seems that benchmark upload was not done after version 5.9 (for releases, it is still uploaded for main branch), might be related to some changes made there with the CI.
I hope this will solve this as well.

PR #6876, opened 2025-12-17
Allow manual run of benchmark upload in case of missing uploads.

HarelM:
This allow the manual upload of tagged versions to github pages and fixes the assumption that github.ref has the right information in the release workflow.

### DIFF
--- a/.github/workflows/upload-benchmarks.yml
+++ b/.github/workflows/upload-benchmarks.yml
@@ -25,8 +25,8 @@ jobs:
         if: inputs.tag != ''
         id: ref
         run: |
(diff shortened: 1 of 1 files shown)


HarelM:
Sorry for the extra work @CommanderStorm, this is hard to test so I'm relying on reading the code again and again...

HarelM:
I was wrong, the folders are string sorted to version 5.11.0 to 5.14.0 exists. Just the last version is missing, which is a change I made with a wrong assumption on how github action trigger one another.
But having a way to manually upload a version in these cases is also good.

Issue #2587, opened 2023-05-24
map.SetStyle(...) conflicts with custom render layers


measuredweighed:
**maplibre-gl-js version**: 3.0

**browser**: Safari, Chrome (likely all)

### Steps to Trigger Behavior

 1. Set a custom style using map.setStyle(...)
 2. Add a custom render layer like so: ```maplibre.on("style.load", () => {
maplibre.addLayer(customLayer);
});```
 3. Change the map style using map.setStyle(...) again

### Link to Demonstration

This behaviour can be seen live on [https://planefinder.net](https://planefinder.net). Tap the settings icon in the top right and toggle Map Appearance in the sidebar that appears (switching between "Road" and "Satellite"). Notice that the yellow aircraft marker layer is hidden by the appearance of the new style.

### Expected Behavior

We'd expect the style to switch without impacting our custom render layer.

### Actual Behavior

The custom render layer appears to be overlapped by the layers loaded from the new style. This approach has previously worked for MapLibre versions < 3.0 (and before that, Mapbox). I've attached a video showcasing the issue.

https://github.com/maplibre/maplibre-gl-js/assets/1234600/7ea68d7d-2a22-4f44-b7ae-9eb16396dd0c

### Addendum

Thank you for such a wonderful 3.0 release. It's wonderful to see the MapLibre project thriving. You've all done incredible work.

razorness:
Facing same issue. Looks like `style.load` is not getting fired on map instance.

https://github.com/maplibre/maplibre-gl-js/blob/bc64a2833e69f0cb1c10a717df3f82bf18e730f4/src/style/style.ts#L329

I have a small library for Vue bindings with a style switch. All custom sources and layers add them self on map on `style.load` event. 

Here is a branch for reproduction:
https://github.com/razorness/vue-maplibre-gl/tree/v3-style.load-not-fired

The magic happens in `src/lib/composable/useSource.ts`:

https://github.com/razorness/vue-maplibre-gl/blob/7731bd070468a505e55d9d85fff1cf9375b03b12/src/lib/composable/useSource.ts#L32

and

https://github.com/razorness/vue-maplibre-gl/blob/7731bd070468a505e55d9d85fff1cf9375b03b12/src/lib/composable/useSource.ts#L19-L25

I added a simple `console.log` to get some logging without disturbing tmings etc.

[...]

PR #6880, opened 2025-12-17
Emit style.load event on style diff

### LINKED ISSUE #4757: Need an event status identifier push, when I use setStyle(), change the layer
Now there is a scene in the project, when the user switches the map style, we need to redraw some layers on the map. The problem is that none of the event events in maplibregl can tell me that the new layer has been loaded and I can redraw now. I can only deal with this problem by setting a timer. It makes me feel very insecure.

lesbaa:
## Launch Checklist

Fixes an issue where reapplying a JSON style object via setStyle would not emit style.load, even when the internal diff detected changes. This caused style updates to be applied silently.

This affects working with custom layers, which need to be re-added after a style change. The fix ensures style.load fires when a JSON style is reapplied and so custom layers can be re-applied.

Addresses #2587 and likely resolves #4757.

 - [x] Link to related issues.

### DIFF
--- a/src/style/style.test.ts
+++ b/src/style/style.test.ts
@@ -886,7 +886,6 @@ describe('Style.setState', () => {
         style.loadJSON(styleJson);
         
         await style.once('style.load');
-        // console.log('styleJson', style);
 
         const newStyleJSON: StyleSpecification = {
             ...styleJson,


HarelM:
Why was the layer visibility stuff removed?

lesbaa:
@HarelM Huh, weird, I'm not sure 🤔, all I did was add the `fire('style.load')` call and added a test for it. I don't know why the layer visibility stuff was removed and part of the PR, I'll revert it...

HarelM:
Probably some bad merge. 
Thanks for the PR!

lesbaa:
@HarelM Very welcome. Anyway, reverted and updated.

HarelM:
Is there any way this is even is fired twice by mistake? Just asking to make sure not because I see a bug...

lesbaa:
I'll double check, but from my local testing, the `setState` call only seems to happen when the style is diffed from JSON, not from loaded styles or JSON loaded from a remote style. But I'll maybe add some more test cases to cover this just in case.

lesbaa:
Can confirm this only fires when a style is diffed.

package-lock.json
HarelM, line 1: This should be reverted and a changelog entry should be added. Thanks!!

lesbaa: Done 👍 .

Also, happy holidays!

src/style/style.test.ts
```diff
@@ -880,6 +880,28 @@ describe('Style.setState', () => {
...
+                {
+                    id: 'layerId2',
+                    type: 'background',
+                },
+                ...styleJson.layers,
+            ]
+        };
+
+        await style.once('style.load');
+        const spy = vi.spyOn(style, 'fire');
```
HarelM: It's better to use the "public API". i.e. `style.on('style.load', spy)`;

HarelM:
Sorry, missed a comment in the test, can you fix it and I'll merge this?

lesbaa:
> Sorry, missed a comment in the test, can you fix it and I'll merge this?

Updated 👍

src/style/style.test.ts
```diff
@@ -884,6 +884,9 @@ describe('Style.setState', () => {
         const style = createStyle();
         const styleJson = createStyleJSON();
         style.loadJSON(styleJson);
+        
+        await style.once('style.load');
+        // console.log('styleJson', style);
```
HarelM: Can this comment be removed?

lesbaa: done 👍

Issue #6872, opened 2025-12-17
Zoom in/out buttons should round zoom level


1ec5:
`NavigationControl`’s Zoom In and Zoom Out buttons always increment and decrement the zoom level by exactly one, even if the initial zoom level isn’t an integer. For example, if the map is at zoom level 14.5 and you click Zoom In, the resulting zoom level is 15.5. This is inconsistent with the <kbd>+</kbd> and <kbd>−</kbd> keyboard shortcuts, which round the zoom level to the nearest integer. The Zoom In and Zoom Out buttons should likewise round the zoom level.

These buttons should behave consistently with the keyboard shortcuts, because some people are only able to use one or the other at any given time. They have the same plus and minus signs, so the user would have no reason to believe they’d behave differently. Of the two behaviors, style authors would likely appreciate rounding to the nearest integer. Zoom stops are only evaluated at integer zoom levels: mapbox/mapbox-gl-js#13291. Moreover, raster basemap tiles (as opposed to satellite imagery) easily appear blurry at fractional zoom levels, so users would want an easy way to get to something more legible.

HarelM:
I think this is a good idea to unify between keyboard and zoom controls. 
Since this is a behavior change I would recommend doing it in a breaking change version, which is tracked here: 
- #6427

mizmay:
Anyone mind if I take a crack at this?

neodescis:
Agreed it would be good to unify the behaviors. Just know that the + and - keyboard actions don't necessarily just go to the next integer value; it appears that the current value is rounded **before** going to the next value. In other words, if you're currently at zoom level 14.6, hitting the + button results in a zoom level of 16.

mizmay:
Investigated. The +/- button behavior is controlled in [src/ui/camera.ts](src/ui/camera.ts):

`this.getZoom() + 1` => zoom 14.6 becomes 15.6 when + button is pressed

The keyboard controller is at [src/ui/handler/keyboard.ts](src/ui/handler/keyboard.ts):

`zoom: zoomDir ? Math.round(tr.zoom) + zoomDir * (e.shiftKey ? 2 : 1) : tr.zoom,` => zoom 14.6 becomes 16 when + key is pressed, or 17 with `Shift-+`

Noting this is already flagged as a breaking change.

Options:

1. Change the button behavior to match the keyboard behavior? Or 
2. Align both to the idea that +/- should take you to the nearest zoom?


[...]

PR #6888, opened 2025-12-19
Align zoom button behavior with keyboard shortcut

mizmay:
## Summary

Aligns the behavior of the zoom button controller with the keyboard shortcuts when it comes to fractional zooms, per #6872 

## Explanation

Currently, when you press the + or - button on a map controller, it zooms in or out *exactly* one zoom, i.e. 2.6 -> 3.6. However, if you press + or - on the keyboard, it rounds to the nearest zoom, and then zooms in or out, i.e. 2.6 -> 4.

Here's a video explanation of the change:

|  before | after |
|---|---|
| before | after|

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/ui/camera.ts
+++ b/src/ui/camera.ts
@@ -515,7 +515,7 @@ export abstract class Camera extends Evented {
     }
 
     /**
-     * Increases the map's zoom level by 1.
+     * Rounds to the nearest integer and then increases the map's zoom level by 1.
      *
      * Triggers the following events: `movestart`, `move`, `moveend`, `zoomstart`, `zoom`, and `zoomend`.
      *
@@ -533,7 +533,7 @@ export abstract class Camera extends Evented {
     }
 
     /**
-     * Decreases the map's zoom level by 1.
+     * Rounds to the nearest integer and then decreases the map's zoom level by 1.
      *
      * Triggers the following events: `movestart`, `move`, `moveend`, `zoomstart`, `zoom`, and `zoomend`.
      *
--- a/src/ui/handler/keyboard.ts
+++ b/src/ui/handler/keyboard.ts
@@ -12,10 +12,10 @@ const defaultOptions = {
  * The `KeyboardHandler` allows the user to zoom, rotate, and pan the map using
  * the following keyboard shortcuts:
  *
- * - `=` / `+`: Increase the zoom level by 1.
- * - `Shift-=` / `Shift-+`: Increase the zoom level by 2.
- * - `-`: Decrease the zoom level by 1.
- * - `Shift--`: Decrease the zoom level by 2.
+ * - `=` / `+`: Round to the nearest integer and then increase the zoom level by 1.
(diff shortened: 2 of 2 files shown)


mizmay:
Bummer Github keeps nuking the previews on my animated gifs, I'm not sure how small I have to make them or if it's worth it?

HarelM:
Would you like to introduce summer configuration to avoid a breaking change and in a breaking change version remove this configuration maybe? Not sure I'd know where to put that, but still worth asking...

mizmay:
@HarelM adding a config seems reasonable. BUT: if you have no idea, I definitely don't know what to suggest :). Time to think about it; alternative PR/implementation incoming....

src/ui/camera.ts
```diff
@@ -528,7 +528,7 @@ export abstract class Camera extends Evented {
      * ```
      */
     zoomIn(options?: AnimationOptions, eventData?: any): this {
-        this.zoomTo(this.getZoom() + 1, options, eventData);
+        this.zoomTo(Math.round(this.getZoom()) + 1, options, eventData);
```
1ec5, line 531: Developers might expect to call `zoomIn()` and `zoomOut()` programmatically as a shortcut for `setZoom()`. They might not expect `zoomIn()` and `zoomOut()` to have a subtle rounding behavior that `setZoom()` doesn’t have. I understand that that’s why we consider this a backwards-incompatible change. Modifying `NavigationControl` more directly would limit the fallout to where I think we could consider it a change in behavior but not an API change. If we still want a configuration option, then we could tuck it away in [`NavigationControlOptions`](https://maplibre.org/maplibre-gl-js/docs/API/type-aliases/NavigationControlOptions/).

mizmay: It is reasonable to expect `setZoom()` versus `zoomIn()` and `zoomOut()` to behave differently. The latter are relative, whereas the former is absolute. I disagree that this is the primary reason why this is a breaking change.

Let's please never forget that for all the developers there are in the world, there are _many more_ existing implementations of Maplibre. Those implementations have data layers, styles, and users. This change, however subtle, changes one of the most fundamental user interactions there is--that to my mind is why this is a breaking change.

Yes `NavigationControlOptions` is a choice, the other choice is to add e.g. `MapOptions.zoomRounding` as a global option.

1ec5: That’s fair, though do we actually need `zoomIn()` and `zoomOut()` to do exactly what the keyboard shortcuts do? To avoid being blocked on a major version, we could implement something for now that’s scoped to just `NavigationControl`, gated by an option on that class, and then later change the default, including by changing the implementation of these methods if desired.

HarelM: Another option is to add an optional parameter to this method to keep backwards compatibility (e.g. `roundZoom?: boolean` and remove it in the next major version.
We have an issue to track the breaking changes we would like to do in version 6, we can add it there.

PR #6894, opened 2025-12-19
fix: mapbox draw example needs maplibre specific CSS classes to work

tarwin:
Mapbox Draw plugin which works with Maplibre references Mapbox specific CSS classes for CSS cursor changes.

In the example of how to use Mapbox Draw plugin there was already one fix that showed how `styles` (internal) needed to be changed to work with Maplibre, so I added the required CSS with changed class names as well so others know how to use "correctly".


HarelM (approved):

HarelM:
Thanks!

PR #6907, opened 2025-12-23
Hide control characters in formatted string expressions

### LINKED ISSUE #6903: Visible control characters in `format` expressions
Some control characters are intended to affect text layout without any spacing. For example, [Diplomat](https://github.com/osm-americana/diplomat/) sometimes surrounds text with U+2068 FIRST STRONG ISOLATE (FSI) and U+2069 POP DIRECTIONAL ISOLATE (PDI) characters to prevent embedded right-to-left text from affecting the layout of adjacent left-to-right text or vice versa. When these characters occur in a string expression, they are invisible and nonspacing, as expected. But when these same characters occur in a `format` expression, they appear as either control pictures (when `glyphs` is specified) or blank spaces (when `glyphs` is unspecified). By definition, control characters should never be visible or take up space. `format` expressions should not render these characters any differently than plain string expressions.

[This example](https://jsbin.com/tapubub/4/edit?html,output) appears correct initially: the names in local languages are surrounded by parentheses without any spaces:

Nothing surrounding parentheses

If you comment out the line that replaces the `concat` expression with a `format` expression, spaces appear in place of the FSI and PDI characters:

Spaces surrounding parentheses

If you comment out the line that specifies `glyphs`, [FSI] and [PDI] markers appear in those spaces:

[FSI] and [PDI] surrounding parentheses

Reproduces in GL&nbsp;JS v5.15.0 in Firefox 147.0b5.

1ec5:
A control character that occurs at the beginning or end of an in put to a formatted string expression (`format`) is by definition invisible and nonspacing, so zero out its horizontal metrics. Along the way, I realized we weren’t actually testing glyph manager when leaving `glyphs` unset, so we have a basic test of that now.

Before | After
----|----
Spaced parentheses | Unspaced parentheses

Fixes #6903.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/render/glyph_manager.ts
+++ b/src/render/glyph_manager.ts
@@ -191,15 +191,18 @@ export class GlyphManager {
 
         const leftAdjustment = 0.5;
 
+        // By definition, control characters are invisible and nonspacing.
+        const isControl = /^\p{gc=Cf}+$/u.test(String.fromCodePoint(id));
+
         return {
             id,
             bitmap: new AlphaImage({width: char.width || 30 * textureScale, height: char.height || 30 * textureScale}, char.data),
             metrics: {
-                width: char.glyphWidth / textureScale || 24,
+                width: isControl ? 0 : (char.glyphWidth / textureScale || 24),
                 height: char.glyphHeight / textureScale || 24,
                 left: (char.glyphLeft / textureScale + leftAdjustment) || 0,
                 top: char.glyphTop / textureScale - topAdjustment || -8,
-                advance: char.glyphAdvance / textureScale || 24,
+                advance: isControl ? 0 : (char.glyphAdvance / textureScale || 24),
                 isDoubleResolution: true
             }
         };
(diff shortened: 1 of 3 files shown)


HarelM (approved):
Thanks!

src/render/glyph_manager.ts
```diff
@@ -191,15 +191,18 @@ export class GlyphManager {
...
 
+        // By definition, control characters are invisible and nonspacing.
+        const isControl = /^\p{gc=Cf}+$/u.test(String.fromCodePoint(id));
+
         return {
             id,
             bitmap: new AlphaImage({width: char.width || 30 * textureScale, height: char.height || 30 * textureScale}, char.data),
             metrics: {
-                width: char.glyphWidth / textureScale || 24,
+                width: isControl ? 0 : (char.glyphWidth / textureScale || 24),
```
HarelM, line 201: Is this code flow executed only for `format`? I would expect this to be a bug for regular text as well, shouldn't it?

1ec5: The code path runs regardless. I eventually discovered that unformatted text can exhibit this bug too, but only if the control character occurs [at the very beginning or end of the entire resolved string](https://jsbin.com/pagopey/1/edit?html,output), only after concatenating all the inputs. It’s easier to hit this bug with formatted strings because each section is laid out independently. And that’s a more realistic case for putting bidi control characters in a label.

PR #6908, opened 2025-12-24
Round zoom button keyboard

mizmay:
# Align Zoom Button and Keyboard Behavior with `legacyZoom` Flag

## Summary
This PR addresses the inconsistent zoom behavior between UI buttons and keyboard shortcuts as discussed in [Issue #6872](https://github.com/maplibre/maplibre-gl-js/issues/6872). It introduces a new threshold-based rounding logic that aligns both controllers and provides an intentional zoom-out bias, gated behind a backward-compatible `legacyZoom` flag (default: `true`).

See  [this demo app](https://mizmay.github.io/maplibre-gl-js/) for a visual exploration of the change.

### The Issue
The behavior between the keyboard `+`,`-` and the zoom control buttons was inconsistent:
- **UI Buttons** (`zoomIn`/`zoomOut`): Performed a simple `zoom ± 1` with no rounding, preserving fractional zoom levels.
- **Keyboard Shortcuts**: Used `Math.round(zoom) + step`, which rounded to the nearest integer.

Is this inconsistency intentional? The spec is unclear:
- Code comments describe identical functionality, "Increase or decrease zoom by 1" with no mention of rounding
- How the code is written suggests these different sections of the code base were not written in parallel
- The UI Buttons are customizable in ways the keyboard handlers are not, e.g. event triggers, animations etc.
- The keyboard handler includes an option to zoom in or out by 2 increments by holding down the shift key

Should the behaviors be inconsistent? **No**
- The current keyboard behavior is not transparent
- The discrepency may be confounding in some use cases, e.g. automation

Should both controllers zoom in/out to rounded integer zooms? **Yes, with reservations**.

**Pros:**
- In #6872 @1ec5 makes the case for rounded zooming. I am convinced, but not for identical reasons:
  - They have the same + and - signs so the user would have no reason to believe they would behave differently: *I suspect most users do not know the keyboard controls are there, and special keyboard shortcuts for developers are common, this may not be a strong argument*
  - Style authors would appreciate rounding to the nearest integer because the renderer only rerenders styles at integer zooms *renderer properties matter for zoom dependent layer styles, but style doesn't really factor in to UI controls and probably shouldn't*
- Possibly there is some qualitative improvement in sharpness of rendering

**Cons:**
  - If you zoom in and then out from a fractional zoom, rounding means you don't end up where you started.
  - The rounding creates asymmetry between the zoom in and zoom out increments, e.g. for zoom 14.5 you zoom in 1.5 zooms to 16, out 0.5 zoom to 14

### The Solution: Threshold Rounding

After experimenting with a [demo app](https://mizmay.github.io/maplibre-gl-js/), I propose applying a modified, threshold-based rounding approach to both the keyboard handler and UI button controller that uses the formula: `round(zoom - 0.1) ± step`.

This creates a **symmetric-like behavior with a zoom-out bias**:
- At exactly `.5` (e.g., `14.5`), zooming in goes to `15` (+0.5) and zooming out goes to `13` (-1.5).
- This "asymmetry" is intentional at the mid-point zoom levels: these are often chosen by map developers when initializing the map (e.g. 14.6 is much less common) 
- It makes for a better UI experience because it favors larger leaps when zooming out to quickly gain context, while maintaining gradual refinement when zooming in.

### The `legacyZoom` Flag
To ensure 100% backward compatibility for existing users, this new behavior is gated behind a `legacyZoom` option in `MapOptions`.
- `legacyZoom: true` (default): Preserves the original behavior (buttons don't round, keyboard uses standard rounding).
- `legacyZoom: false`: Enables the new aligned threshold-rounding behavior.

## Behavior Comparison (`legacyZoom: false`)

Note that rounding still produces asymmetry at other fractional zooms:

| Current Zoom | Action | Result | Distance | Note |
|:-------------|:-------|:-------|:---------|:-----|
| 14.5 | zoomIn | 15 | +0.5 | Fractional zoom correction |
| 14.5 | zoomOut | 13 | -1.5 | Additional Context |
| 14.6 | zoomIn | 16 | +1.4 | Above 0.6 threshold |
| 14.6 | zoomOut | 14 | -0.6 | |
| 14.4 | zoomIn | 15 | +0.6 | Below 0.6 threshold |
| 14.4 | zoomOut | 13 | -1.4 | |

## Alternatives Considered
- **Alignment approach** (Apply the keyboard handler logic to the UI buttons): This was implemented in PR#6888 but ultimately rejected because the asymmetric responses within the UI felt off.
- **Symmetric approach** (`Math.round(zoom + 0.1)` before incrementing the zoom in or out): This would result in equal ±0.5 jumps from the `.5` midpoint. This was rejected because the chosen approach felt more intuitive in testing.

## Suggested Additional Considerations

We should consider any implications for planimetric applications, e.g. turn-by-turn navigation.

## Implementation Details
- `src/ui/map.ts`: Added `legacyZoom` to `MapOptions`.
- `src/ui/camera.ts`: Updated `zoomIn`/`zoomOut` to check `legacyZoom`.
- `src/ui/handler/keyboard.ts`: Updated keydown logic to check `legacyZoom`.
- Comprehensive unit tests added to both `keyboard.test.ts` and `camera.test.ts` covering both flag states.

## Success Criteria
- [x] Alignment: Keyboard and buttons behave identically when `legacyZoom: false`.
- [x] Backward Compatibility: Default behavior is unchanged (`legacyZoom: true`).
- [x] Intentional asymmetry: Zoom-out is a bigger leap from the `.5` midpoint.
- [x] All tests passing.

## Benchmarking
Since this change only affects high-level camera methods triggered by user input, no performance impact on rendering or data processing is expected.

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## main
 
 ### ✨ Features and improvements
+- Add `legacyZoom` option (default: `true`) to optionally align zoom button and keyboard shortcut behavior using threshold-based rounding ([#6872](https://github.com/maplibre/maplibre-gl-js/issues/6872))
 - Add `padding` option to `Popup` class to prevent popups from being positioned too close to map container edges ([#5978](https://github.com/maplibre/maplibre-gl-js/issues/5978))
 - _...Add new stuff here..._
 


src/ui/handler/keyboard.ts
```diff
@@ -123,7 +128,11 @@ export class KeyboardHandler implements Handler {
                     easeId: 'keyboardHandler',
                     easing: easeOut,
 
-                    zoom: zoomDir ? Math.round(tr.zoom) + zoomDir * (e.shiftKey ? 2 : 1) : tr.zoom,
+                    zoom: zoomDir ? (
+                        map._legacyZoom
```
HarelM, line 132: I think using the flag for that is redundant as the behavior is not dramatically changed ut rather improved for the edge case of .5, so I would recommend removing this code from here.

src/ui/handler/keyboard.test.ts
```diff
@@ -233,4 +233,55 @@ describe('keyboard', () => {
         expect(spy.mock.calls[3][0].zoom).toBe(8);
 
     });
+
+    test('KeyboardHandler rounds fractional zoom before incrementing when legacyZoom is false', () => {
```
HarelM, line 237: If these tests are still needed I think it would be better to split them like you did with the camera tests to zoom in and zoom out separately.

src/ui/map.ts
```diff
@@ -152,6 +152,13 @@ export type MapOptions = {
      * @defaultValue 22
      */
     maxZoom?: number | null;
+    /**
+     * If `false`, zoom buttons and keyboard shortcuts align using threshold-based
+     * rounding. If `true` (default), preserves original behavior where buttons
```
HarelM, line 157: It is not clear which button this is reffering to since this is not an option for the navigation control.
I think this definition should be part of the navigation contol options and the zoomin and out methods should have an optional "round" flag.

HarelM:
I've added a few comments.
Also note that one of the unit tests you added are failing.

1ec5:
>> They have the same plus and minus signs, so the user would have no reason to believe they’d behave differently.
> 
> I suspect most users do not know the keyboard controls are there, and special keyboard shortcuts for developers are common, this may not be a strong argument

I agree that this is a superficial argument. The keyboard shortcuts are for laypeople too, depending on the map’s use case and users’ accessibility needs, but you’re right that laypeople typically won’t know or care whether they behave exactly like the on-screen buttons.

>> Moreover, raster basemap tiles (as opposed to satellite imagery) easily appear blurry at fractional zoom levels, so users would want an easy way to get to something more legible.
> 
> Possibly there is some qualitative improvement in sharpness of rendering

Yes, I was alluding to when a MapLibre map needs to display a [raster tile layer](https://maplibre.org/maplibre-gl-js/docs/examples/add-a-raster-tile-source/) that’s cartographic rather than photographic. As the OSM website is migrating to MapLibre, it’s receiving valid criticism for allowing fractional zooming in conjunction with those tiles: openstreetmap/openstreetmap-website#6618. The real solution is zoom snapping (#1591), but rounding zoom levels here would prevent a bounceback effect when zoom snapping is enabled.

src/ui/map.ts
```diff
@@ -152,6 +152,13 @@ export type MapOptions = {
...
+     * don't round and keyboard uses standard rounding.
+     * @defaultValue true
+     */
+    legacyZoom?: boolean;
```
1ec5, line 161: If we have any intention of keeping this option for the long term, I’d suggest using a more descriptive name like `zoomRounding`. Terms like “legacy”, “old”, or “new” can easily get awkward over time as the design changes further.

src/ui/camera.ts
```diff
@@ -517,6 +519,9 @@ export abstract class Camera extends Evented {
     /**
      * Increases the map's zoom level by 1.
      *
+     * If {@link MapOptions.legacyZoom} is `false`, the zoom level is first rounded to the nearest integer
+     * before incrementing, with a threshold of 0.6 (i.e. it rounds up only if the fractional part is \> 0.6).
```
1ec5, line 523: Consider adding a note here that, if the developer doesn’t want this heuristic when setting the zoom level programmatically, they can call `setZoom()` instead.

src/ui/map.ts
1ec5: It looks like the idea is to make this option more global so it can also affect `Map.zoomIn()` and `zoomOut()`. I had originally suggested scoping the change to just the navigation control, but the more sophisticated heuristic implemented in this PR will slightly change the behavior of the keyboard shortcuts too, so it would need a more general option if we give the developer any control over it.

src/ui/camera.ts
```diff
@@ -528,13 +533,19 @@ export abstract class Camera extends Evented {
      * ```
      */
     zoomIn(options?: AnimationOptions, eventData?: any): this {
```
1ec5, line 535: On main, double-clicking or double-tapping increments the zoom level. Additionally, on iOS or on macOS with a multitouch mouse, tapping with two fingers decrements the zoom level. Would it make sense to align these gestures with the keyboard shortcuts and on-screen buttons?

HarelM: I think it makes sense.

mizmay:
Thanks for the comments @HarelM and additional context, @1ec5. 

Here are my key takeaways:
- `legacyZoom` is too confusing, I will use a `roundZoom` flag more narrowly (e.g. only around the +/- buttons)
- Current relevance stems from OSM migration from Leaflet to Maplibre; since default Carto style renders as raster, avoiding fractional zooms is desirable.
- If we are aligning the keyboard controls and +/- buttons, it would make sense to align the double click / double tap zoom controls too.

On this last point, I'm investigating the current implementation and thinking through the more robust context from Minh's links to previous conversations around restricting to whole integer zooms.

I will close the PR and make a proposal in the original issue before opening a new PR.

PR #6915, opened 2025-12-26
Clarify nullability of Map.getGlyphs() and allow unsetting glyphs in setGlyphs()

shreeharshshinde:
## Summary

This PR clarifies and formalizes support for unsetting glyphs.

- `Map.getGlyphs()` / `getGlyphsUrl()` now document that they may return `null` when glyphs are unset.
- `Map.setGlyphs()` and `Style.setGlyphs()` accept `null` or `undefined` to unset glyphs.
- `Style.setGlyphs()` normalizes falsy values to `null` internally to keep state consistent.

## Tests

- Added a unit test covering setting and unsetting glyphs via `null` and `undefined`.
- Verified that glyph manager state is cleared when glyphs are unset.
- All unit tests pass (`npm run test-unit`).

### DIFF
--- a/src/style/style.test.ts
+++ b/src/style/style.test.ts
@@ -1285,26 +1285,23 @@ describe('Style.setGlyphs', () => {
         await style.once('style.load');
         style.update({zoom: 1} as EvaluationParameters);
 
-        // Set to a valid URL first
-        style.setGlyphs('https://foo.maplibre.org/font/{fontstack}/{range}.pbf');
-        expect(style.getGlyphsUrl()).toBe('https://foo.maplibre.org/font/{fontstack}/{range}.pbf');
-        expect(style.glyphManager.url).toBe('https://foo.maplibre.org/font/{fontstack}/{range}.pbf');
+        const glyphsUrl = 'https://foo.maplibre.org/font/{fontstack}/{range}.pbf';
 
-        // Mutate entries to ensure they get cleared
-        style.glyphManager.entries['TEST'] = {glyphs: {}, requests: {}, ranges: {}} as any;
-        expect(Object.keys(style.glyphManager.entries).length).toBeGreaterThan(0);
+        // Set glyphs
+        style.setGlyphs(glyphsUrl);
+        expect(style.getGlyphsUrl()).toBe(glyphsUrl);
 
-        // Unset with null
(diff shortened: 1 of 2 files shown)


src/style/style.test.ts
```diff
@@ -1278,6 +1278,34 @@ describe('Style.setGlyphs', () => {
...
+        await style.once('style.load');
+        style.update({zoom: 1} as EvaluationParameters);
+
+        // Set to a valid URL first
+        style.setGlyphs('https://foo.maplibre.org/font/{fontstack}/{range}.pbf');
+        expect(style.getGlyphsUrl()).toBe('https://foo.maplibre.org/font/{fontstack}/{range}.pbf');
+        expect(style.glyphManager.url).toBe('https://foo.maplibre.org/font/{fontstack}/{range}.pbf');
+
+        // Mutate entries to ensure they get cleared
+        style.glyphManager.entries['TEST'] = {glyphs: {}, requests: {}, ranges: {}} as any;
```
HarelM: This is a "private" API, it's better not to use that in a unit test.
Also the test is a bit too complicated for what you want to test with multiple assers. Consider simplifying it...

src/style/style.ts
```diff
@@ -1913,20 +1913,25 @@ export class Style extends Evented {
...
     }
 
-    setGlyphs(glyphsUrl: string | null, options: StyleSetterOptions = {}) {
+    setGlyphs(glyphsUrl: string | null | undefined, options: StyleSetterOptions = {}) {
         this._checkLoaded();
-        if (glyphsUrl && this._validate(validateStyle.glyphs, 'glyphs', glyphsUrl, null, options)) {
+
+        // Normalize falsy/undefined values to explicit `null` so the stylesheet
+        // and glyphManager consistently represent "unset" as `null`.
+        const url: string | null = glyphsUrl ?? null;
```
HarelM: I'm not sure this is needed, you can simply pass down null | undefined.

HarelM:
Added a few comments. The docs comment change is a good one, thanks!

shreeharshshinde:
Done @HarelM — simplified the test to rely only on the public API and removed assertions against private glyph manager internals; all checks passed successfully

PR #6926, opened 2025-12-31
Fix marker positions not updating when zoom or pitch are changed following a change in constraints

### LINKED ISSUE #6925: map.setMinZoom (or setMaxZoom) does not trigger markers to update until user interaction
**maplibre-gl-js version**: 5.15.0

**browser**: N/A

### Steps to Trigger Behavior

 1. Change minZoom or maxZoom such that the current zoom is out of bounds and will be updated to fall within new constraints
 2. Note that any markers added to the map will not move
 3. Interacting with the map via zoom, drag will cause the markers to move

### Link to Demonstration

```
<!doctype html>
<html lang="en">
    <head>
        <link href="../../dist/maplibre-gl.css" rel="stylesheet" />
        <style>
            html,
            body,
            #app,
            #map {
                margin: 0;
                width: 100vw;
                height: 100vh;
            }

            #change-zoom {
                position: absolute;
                top: 1rem;
                left: 1rem;
                z-index: 100;
            }

            #current-zoom {
                position: absolute;
                z-index: 100;
                font-size: 0.8rem;
                background-color: white;
                padding: 0.25rem;
                bottom: 1rem;
                left: 1rem;
            }
        </style>
    </head>

    <body>
        <div id="app">
            <button id="change-zoom">Change Zoom</button>
            <div id="map"></div>
            <div id="current-zoom"></div>
        </div>
        <script type="module">
            import "../../dist/maplibre-gl-dev.js";

                container: "map",
                style: "https://demotiles.maplibre.org/style.json",
                center: [151.18068748043615, -33.96434511002358],
                zoom: 5,
                minZoom: 4,
                maxZoom: 20,
                transformCameraUpdate: (t) => t,
            });

            const marker = new maplibregl.Marker({
                color: "#F00",
            })
                .setLngLat([151.2057, -33.8727])
                .addTo(map);

            map.on("render", (e) => {
                const minZoom = e.target.getMinZoom();
                const maxZoom = e.target.getMaxZoom();
                const zoom = e.target.getZoom();
                document.querySelector("#current-zoom").innerHTML =
                    `${minZoom} - ${maxZoom} (${zoom})`;
            });

            document
                .querySelector("#change-zoom")
                .addEventListener("click", () => {
                    const currentMinZoom = map.getMinZoom();
                    map.setMinZoom(currentMinZoom === 5 ? 4 : 5);
                    map.setMaxZoom(currentMinZoom === 5 ? 18 : 20);
                });
        </script>
    </body>
</html>
```

### Expected Behavior

Markers move immediately after changing zoom.

### Actual Behavior

Markers do not move until after user interaction.

Auspicus:
- Triggers `zoomstart`, `zoom`, `zoomend`, `movestart`, `move` and `moveend` when calling `setMinZoom` and `setMaxZoom`
- Triggers `pitchstart`, `pitch` and `pitchend`, `movestart`, `move` and `moveend` when calling `setMinPitch`  and `setMaxPitch`
- Markers will update if `setMinPitch`, `setMaxPitch`, `setMinZoom` or `setMaxZoom` adjust the transform
- Allows changes to `pitch` from `setMinPitch` or `setMaxPitch` to be controlled via `transformCameraUpdate` (following on from https://github.com/maplibre/maplibre-gl-js/pull/6781 where the same was done for zoom)

fixes #6925

todo:
- [x] tests
- [x] docs
- [x] ~benchmark? `zoomend` could be triggered frequently enough that markers are updated more frequently than necessary~ no longer using `zoomend` or `pitchend` events to update marker

## Launch Checklist

 - [x] Link to related issues. https://github.com/maplibre/maplibre-gl-js/issues/6925
 - [ ] Post benchmark scores.

### DIFF
--- a/src/ui/map_tests/map_pitch.test.ts
+++ b/src/ui/map_tests/map_pitch.test.ts
@@ -1,4 +1,4 @@
-import {beforeEach, test, expect,vi} from 'vitest';
+import {beforeEach, test, expect, vi} from 'vitest';
 import {createMap, beforeMapTest} from '../../util/test/util';
 
 beforeEach(() => {


src/ui/map.ts
```diff
@@ -1260,10 +1269,13 @@ export class Map extends Camera {
         }
 
         if (maxPitch >= this.transform.minPitch) {
-            this.transform.setMaxPitch(maxPitch);
+            const tr = this._getTransformForUpdate();
+            tr.setMaxPitch(maxPitch);
+            this._applyUpdatedTransform(tr);
             this._update();
-
-            if (this.getPitch() > maxPitch) this.setPitch(maxPitch);
```
Auspicus, line 1294: This isn't needed as the pitch is constrained by the transform helper.

src/ui/map.ts
```diff
@@ -1224,10 +1230,13 @@ export class Map extends Camera {
         }
 
         if (minPitch >= defaultMinPitch && minPitch <= this.transform.maxPitch) {
-            this.transform.setMinPitch(minPitch);
+            const tr = this._getTransformForUpdate();
+            tr.setMinPitch(minPitch);
+            this._applyUpdatedTransform(tr);
             this._update();
-
-            if (this.getPitch() < minPitch) this.setPitch(minPitch);
```
Auspicus, line 1258: This isn't needed as the pitch is constrained by the transform helper.

HarelM:
Thanks for taking the time to report the issue and open the PR.
I'm a bit concerned about all the events that were added, I would expect them to be fired only if the zoom/pitch are changed and not "always".
In theory moveend is fired alongside pitchend, right?
Button line, I'm not sure if this is the right solition, but I don't have a better idea how to solve it...

Auspicus:
@HarelM I've updated this PR to first check if the pitch or zoom must change before deciding to fire an event. I think the reason the `movestart`, `move` and `moveend` events don't cover this is because the change to `pitch` and `zoom` are a side-effect of constraining these values into the new range specified in `setMinZoom` `setMinPitch`, etc... AFAICT `move...` events are never triggered by these calls alone.

src/ui/map.ts
```diff
@@ -1184,10 +1190,16 @@ export class Map extends Camera {
         maxZoom = maxZoom === null || maxZoom === undefined ? defaultMaxZoom : maxZoom;
 
         if (maxZoom >= this.transform.minZoom) {
+            const zoomMustChange = this.transform.zoom > maxZoom;
             const tr = this._getTransformForUpdate();
             tr.setMaxZoom(maxZoom);
             this._applyUpdatedTransform(tr);
             this._update();
+            if (zoomMustChange) {
+                this.fire(new Event('zoomstart'))
```
Auspicus, line 1232: Similar pattern to: https://github.com/maplibre/maplibre-gl-js/blob/main/src/ui/camera.ts#L582-L584

src/ui/marker.ts
```diff
@@ -329,6 +329,8 @@ export class Marker extends Evented {
         map.getCanvasContainer().appendChild(this._element);
         map.on('move', this._update);
         map.on('moveend', this._update);
+        map.on('zoomend', this._update);
```
HarelM: What about unr gistering these events to avoid memory leak?

HarelM:
Can you add a changelog entry and I'll merge this?
Thanks!

Auspicus:
Hey @HarelM, I've updated the `CHANGELOG.md` and adjusted the PR a bit. I've removed the event listeners that were added for `zoomend` and `pitchend` in favor of just firing the `moveX` events in the same fashion that the `setVerticalFieldOfView` function does. I've also added a test and updated the doc comment on the `setMinX` and `setMaxX` functions.

src/ui/marker.test.ts
```diff
@@ -1206,4 +1206,80 @@ describe('marker', () => {
...
+        const map = createMap({width: 1024});
+        await map.once('load');
+        map.setZoom(0);
+        map.jumpTo({center: {lat: 1, lng: 1}});
+        const marker = new Marker()
+            .setLngLat([0, 0])
+            .setSubpixelPositioning(true)
+            .addTo(map);
+
+        const beforePitch = marker.getElement().style.transform;
```
HarelM: These are "internal" fields, I think it would be wiser not to assert on them, but instead concentrate on the "public" API - i.e. events in this case and marker location.

Auspicus: Some of the previous tests in this file were also asserting on the `marker.getElement().style.transform` and it doesn't seem like there's a good parallel for the screen position in the public API. Nevertheless, I changed the test coverage to instead only ensure that the map events are called since that was the real reason the transform was not updating.

PR #6935, opened 2026-01-06
fix(context restore): map is blank if resize is called before restore

ToHold:
## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

## Description
This PR fixes an issue where a blank map appears if a resize event is triggered before the WebGL context has been fully restored.

### The Problem
When a context loss occurs, the painter is destroyed and only recreated once the context is restored. Currently, the resize method attempts to use the painter immediately. If the resize happens during that "lost" if the map or the browser size is changed, it fails or enters an invalid state, resulting in a blank map.

### The Fix
I have added a check to ensure the painter is available and the context is ready before proceeding with the resize logic. This ensures the map state remains consistent during context restoration.

This issue has been observed and verified by users in my application. The fix is straightforward and prevents a critical visual failure.

## Related Issues
https://github.com/maplibre/maplibre-gl-js/issues/6242

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,12 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
-- Fix setting visibility on custom layer ([#6883](https://github.com/maplibre/maplibre-gl-js/issues/6883) (by [melitele](https://github.com/melitele))
+- Fix setting visibility on custom layer ([#6883](https://github.com/maplibre/maplibre-gl-js/issues/6883)) (by [melitele](https://github.com/melitele))
 - Hide leading and trailing control characters in `format` expressions ([#6907](https://github.com/maplibre/maplibre-gl-js/pull/6907)) (by [1ec5](https://github.com/1ec5))
 - Fix image sources being clipped at -180 and 180 longitude when terrain is enabled ([#4088](https://github.com/maplibre/maplibre-gl-js/issues/4088)) (by [pstaszek](https://github.com/pstaszek))
 - Fix map not immediately constraining to a valid zoom and center when changing projections ([#6892](https://github.com/maplibre/maplibre-gl-js/issues/6892)) (by [larsmaxfield](https://github.com/larsmaxfield))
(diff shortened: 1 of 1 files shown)


src/ui/map.ts
```diff
@@ -579,6 +579,12 @@ export class Map extends Camera {
         images: null
     };
 
+    /**
+     * @internal
+     * Flag to indicate whether the WebGL context is lost.
+     */
+    _isContextLost: boolean = false;
```
HarelM: Can we use the saved style as an indicator of context lost instead?

src/ui/map.ts
```diff
@@ -990,6 +996,26 @@ export class Map extends Camera {
...
+        }
+
+        this.fire(new Event('resize', eventData));
+
+        if (fireMoving) this.fire(new Event('moveend', eventData));
+
+        return this;
+    }
+
+    _resizeMap(constrainTransform = true) {
```
HarelM: Its not clear the difference between the two resize methods - resize and resizeMap, it's best to make the name more specific or maybe use "_resizeInternal" to indicate that this is part of the resize method...

ToHold: Should I rename _resizeInternal, or are the @internal tag and the comment sufficient?

HarelM: I think that calling it internal is better, there is the internal tag too. Both are probably even better...

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ### ✨ Features and improvements
 - Add `padding` option to `Popup` class to prevent popups from being positioned too close to map container edges ([#5978](https://github.com/maplibre/maplibre-gl-js/issues/5978))
+- Fix a blank map when a resize event is triggered before a WebGL context restoration ([#6935](https://github.com/maplibre/maplibre-gl-js/pull/6935))
```
HarelM: This should be under bug fixes and should have a "by" link.

src/ui/map.ts
```diff
@@ -990,6 +990,39 @@ export class Map extends Camera {
...
+        return this;
+    }
+
+    /**
+     * Resizes the map according to the dimensions of its
+     * `container` element.
+     *
+     * It does not trigger any events, and does not check for context loss.
+     *
+     * It is used internally and by {@link Map#resize}.
```
HarelM:
```suggestion
     * It is used internally and by {@link Map.resize}.
```

PR #6941, opened 2026-01-08
Implement `zoomSnap` API

### LINKED ISSUE #6872: Zoom in/out buttons should round zoom level

mizmay:
## Summary

Resolves #6872 by implementing a centralized `zoomSnap` API that rounds the zoom increment by the specified interval. Defaults to 0 for backward compatibility and common vector tile use cases where whole integer zooming is not required. Going forward, developers can specify a zoom increment (e.g., 1.0 or 0.5) for discrete zoom operations. All discrete input methods—including UI buttons, keyboard shortcuts, and double-click/tap interactions—have been refactored to respect this setting through a snapping helper in the Camera class. 

This functionality has been requested to align zoom behaviors across UI patterns (currently, keyboard zooming snaps to whole integers whereas HTML buttons, mouse click and double tap do not) as well as to prevent regressions on raster tile maps switching from Leaflet to Maplibre. It mimics Leaflet's [zoomSnap](https://leafletjs.com/reference.html#map-zoomsnap) API, though the implementation is different, and parallels the Maplibre `bearingSnap` API.

As time permits, I can add an implementation to address #1591 ; reviewers please advise if you'd prefer this in a separate PR. EDIT: added, scroll down...

cc/ @1ec5 

## Launch Checklist

 - [x] Link to related issues.
 - [x] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## main
 ### ✨ Features and improvements
-- ⚠️ Add `zoomSnap` option to `Map` to allow snapping zoom levels to a grid when zooming in and out; aligns the behavior across all UI patterns (keyboard, scroll wheel, on-screen buttons, double-click, double-tap. Previously, pressing +/- on the keyboard zoomed to rounded whole integers, now by default all zoom handlers zoom in or out by exactly one zoom, unless a value > 0 for `zoomSnap` is specified ([#6941](https://github.com/maplibre/maplibre-gl-js/pull/6941)) (by [@mizmay](https://github.com/mizmay))
+- ⚠️ Add `zoomSnap` option to `Map` to allow snapping zoom levels to a grid when zooming in and out; aligns the behavior across all UI patterns (keyboard, scroll wheel, on-screen zoom buttons, double-click, double-tap). Previously, pressing +/- on the keyboard zoomed to rounded whole integers, more or less than 1 zoom level when starting from a fractional zoom. Now any number can be specified for `zoomSnap`; a value of 1.0 produces the rounded whole integer behavior across all UI patterns. ([#6941](https://github.com/maplibre/maplibre-gl-js/pull/6941)) (by [@mizmay](https://github.com/mizmay))
 
 ### 🐞 Bug fixes
 - Fix incorrect shader selection for text labels with inline images ([#6956](https://github.com/maplibre/maplibre-gl-js/pull/6956)) (by [@ciscorn](https://github.com/ciscorn))


mizmay:
```
npm run benchmark -- Paint
 v5.15.0                          main  feat/zoom-snap-dirty 5db13b8 
 Paint                  10.3279 ms☠️                    9.2146 ms                     9.9970 ms⚠️
```

src/ui/control/navigation_control.test.ts
```diff
@@ -236,4 +236,17 @@ describe('NavigationControl', () => {
...
+    test('zoom buttons respect zoomSnap', () => {
+        map.setZoomSnap(1.0);
+        map.setZoom(9.7);
+        map.addControl(new NavigationControl());
+
+        const spy = vi.spyOn(map, 'zoomTo');
+        const zoomInButton = map.getContainer().querySelector('.maplibregl-ctrl-zoom-in');
+        simulate.click(zoomInButton);
+
+        // 9.7 + 1.0 = 10.7 -> snap to 11.0
```
HarelM: Optional: Maybe add this comment in the test "name" instead of here?

src/ui/handler/click_zoom.ts
```diff
@@ -29,7 +29,7 @@ export class ClickZoomHandler implements Handler {
             cameraAnimation: (map: Map) => {
                 map.easeTo({
                     duration: 300,
-                    zoom: this._tr.zoom + (e.shiftKey ? -1 : 1),
+                    zoom: map._snapZoom(this._tr.zoom + (e.shiftKey ? -1 : 1)),
```
HarelM: This looks as if it uses a private method. Is there a different way to solve this?

src/ui/handler/dblclick_zoom.test.ts
```diff
@@ -55,6 +58,47 @@ describe('dbclick_zoom', () => {
...
+            container: window.document.createElement('div'),
+            zoomSnap: 1.0,
+            zoom: 9.7
+        } as any);
+        const spy = vi.spyOn(map, 'easeTo');
+
+        const canvas = map.getCanvas();
+        simulate.touchstart(canvas, {touches: [{target: canvas, clientX: 0, clientY: 0}]});
+        simulate.touchend(canvas);
+        await new Promise(resolve => setTimeout(resolve, 100));
```
HarelM: I think there's a "sleep" utility method?

HarelM:
The only concern I have is with the internal method which I'd be happy if there was a better solution as it feels like we are piggy backing on the map/camera object, but I'm not sure I have a better idea how to solve this.

If you would like to add the other feature as part of this PR that's fine too, your call.

Looks good overall, I think some comments in the tests can be removed in favor of more detailed test name, but that's mostly nit picking...

mizmay:
> The only concern I have is with the internal method which I'd be happy if there was a better solution as it feels like we are piggy backing on the map/camera object, but I'm not sure I have a better idea how to solve this.

... if you do have a better idea I am all ears! EDIT: @HarelM I pushed a change to make that method public, but am hesitant because I suspect you would have suggested it?

I'll take a pass over the other comments, this is still my first PR so nitpick away please. It's useful calibration.

In the meantime I pushed another commit that snaps zooms when scrolling. I'm going to add @CommanderStorm as reviewer here, since he's been involved in the Carto migration and #1591 discussion. Frank, do you mind helping me test this?


https://github.com/user-attachments/assets/242a1fa9-608d-40e2-b9e6-1b3b060462a6

Notice from this video, when integer snapping is enabled it increases the sensitivity of the scroll wheel a lot. I don't know if that's a problem.

Parallel Leaflet implementation: sets a timer and delays for a fraction of a second before scrolling and then snapping. I don't know if that makes a difference in the UI.

Happy to push my demo app to the web.

src/ui/handler/click_zoom.ts
mizmay: Yes, I can promote it to public. That probably makes sense, especially if we're using this method across so many handlers...

mizmay: Done

src/ui/handler/dblclick_zoom.test.ts
```diff
@@ -55,6 +58,47 @@ describe('dbclick_zoom', () => {
...
+            container: window.document.createElement('div'),
+            zoomSnap: 1.0,
+            zoom: 9.7
+        } as any);
+        const spy = vi.spyOn(map, 'easeTo');
+
+        simulate.dblclick(map.getCanvas());
+        map._renderTaskQueue.run();
+
+        // 9.7 + 1.0 = 10.7 -> snap to 11.0
```
HarelM: I'm not super excited about these comments, see my previous code review comment...

src/ui/handler/dblclick_zoom.test.ts
```diff
@@ -55,6 +58,47 @@ describe('dbclick_zoom', () => {
...
+        simulate.touchstart(canvas, {touches: [{target: canvas, clientX: 0, clientY: 0}]});
+        simulate.touchend(canvas);
+        await sleep(100);
+        simulate.touchstart(canvas, {touches: [{target: canvas, clientX: 0, clientY: 0}]});
+        simulate.touchend(canvas);
+
+        // MapLibre's TapZoomHandler uses a 0ms timeout to reset state
+        await sleep(10);
+
+        // 9.7 + 1.0 = 10.7 -> snap to 11.0
```
HarelM: Same here.

src/ui/handler/dblclick_zoom.test.ts
```diff
@@ -149,4 +193,29 @@ describe('dbclick_zoom', () => {
...
     });
+
+    test('DoubleClickZoomHandler respects zoomSnap', () => {
+        const map = createMap({zoom: 9.7, zoomSnap: 1.0});
+        const spy = vi.spyOn(map, 'easeTo');
+
+        simulate.dblclick(map.getCanvas());
+        map._renderTaskQueue.run();
+
+        // 9.7 + 1.0 = 10.7 -> snap to 11.0
```
HarelM: Same here

src/ui/handler/dblclick_zoom.test.ts
```diff
@@ -149,4 +193,29 @@ describe('dbclick_zoom', () => {
...
+        map.remove();
+    });
+
+    test('DoubleClickZoomHandler double-tap respects zoomSnap', async () => {
+        const map = createMap({zoom: 9.7, zoomSnap: 1.0});
+        const spy = vi.spyOn(map, 'easeTo');
+
+        await simulateDoubleTap(map, 100);
+
+        // 9.7 + 1.0 = 10.7 -> snap to 11.0
```
HarelM: Same here... You get the point....

src/ui/handler/scroll_zoom.test.ts
```diff
@@ -5,14 +5,15 @@ import {DOM} from '../../util/dom';
 import simulate from '../../../test/unit/lib/simulate_interaction';
 import {setPerformance, beforeMapTest, createTerrain} from '../../util/test/util';
 
-function createMap() {
+function createMap(options: any = {}) {
```
HarelM: Can options be better typed?

src/ui/handler/scroll_zoom.test.ts
```diff
@@ -668,4 +669,120 @@ describe('ScrollZoomHandler', () => {
...
+            map.setZoom(10);
+
+            // Initial render
+            map._renderTaskQueue.run();
+
+            // 1. Simulate wheel event
+            simulate.wheel(map.getCanvas(), {type: 'wheel', deltaY: -simulate.magicWheelZoomDelta});
+            map._renderTaskQueue.run();
+
+            // 2. Advance time to finish the internal ScrollZoomHandler easing (200ms)
```
HarelM: Instead of these comments maybe use utility methods with the relevant name?

src/ui/handler/scroll_zoom.ts
```diff
@@ -297,7 +297,21 @@ export class ScrollZoomHandler implements Handler {
             }
 
             const fromScale = typeof this._targetZoom !== 'number' ? tr.scale : zoomScale(this._targetZoom);
-            this._targetZoom = tr.applyConstrain(tr.getCameraLngLat(), scaleZoom(fromScale * scale)).zoom;
+            const target = tr.applyConstrain(tr.getCameraLngLat(), scaleZoom(fromScale * scale)).zoom;
+
+            if (this._type === 'wheel' && this._map.getZoomSnap() > 0) {
```
HarelM: It might be better to handle this the same way/place the inertia is handled, i.e. at the end of the movement...?

src/ui/camera.ts
```diff
@@ -1161,6 +1183,21 @@ export abstract class Camera extends Evented {
...
     }
 
+    /**
+     * Snaps the provided zoom level to the nearest `zoomSnap` increment.
+     * If `zoomSnap` is 0 or less, the zoom level is returned unchanged.
+     *
+     * @param zoom - The zoom level to snap.
+     * @returns The snapped zoom level.
+     */
+    snapZoom(zoom: number): number {
```
HarelM: This doesn't really snap to the zoom, but rather calculate which zoom to snap to, so the name should be different a bit, getZoomToSnapTo or something else.
also we might consider moving this to be a utility method and not part of the map/camera API.
I'd like to avoid adding this to the public API as I don't think it's very useful "outside", but I might be wrong...

src/ui/handler_manager.ts
```diff
@@ -684,6 +688,34 @@ export class HandlerManager {
                 if (shouldSnapToNorth(this._map.getBearing())) {
                     this._map.resetNorth();
                 }
+
+                if (endEvents['zoomend'] && this._map.getZoomSnap() > 0) {
```
HarelM: This should probably be exported to a method.

src/ui/map.ts
```diff
@@ -3773,4 +3785,23 @@ export class Map extends Camera {
         this.style.setProjection(projection);
         return this._update(true);
     }
+
+    /**
+     * Sets the map's zoom snap level.
+     *
+     * @param snap - The zoom snap level to set.
+     */
+    setZoomSnap(snap: number): this {
```
HarelM: Doesn't look like this id needed...?

src/ui/map.ts
```diff
@@ -3773,4 +3785,23 @@ export class Map extends Camera {
...
+        super.setZoomSnap(snap);
+        return this;
+    }
+
+    /**
+     * Returns the map's current zoom snap level.
+     *
+     * @returns The map's current zoom snap level.
+     */
+    getZoomSnap(): number {
```
HarelM: Same here...

src/ui/handler/scroll_zoom.ts
```diff
@@ -297,7 +297,21 @@ export class ScrollZoomHandler implements Handler {
...
+                const snap = this._map.getZoomSnap();
+                const currentZoom = tr.zoom;
```
CommanderStorm: Did you mean this? (value is unused)


```suggestion
```

src/ui/camera.test.ts
```diff
@@ -3896,3 +3899,44 @@ describe('fitScreenCoordinates globe projection', () => {
...
+    test('snapZoom logic', () => {
+        const camera = createCamera({zoomSnap: 0.5});
+        expect(camera.snapZoom(9.1)).toBe(9.0);
+        expect(camera.snapZoom(9.4)).toBe(9.5);
+        expect(camera.snapZoom(9.7)).toBe(9.5);
+        expect(camera.snapZoom(9.8)).toBe(10.0);
+
+        camera.setZoomSnap(1.0);
+        expect(camera.snapZoom(9.4)).toBe(9.0);
+        expect(camera.snapZoom(9.6)).toBe(10.0);
```
CommanderStorm: Lets assert that the rounding direction is specified

```suggestion
        expect(camera.snapZoom(9.5)).toBe(10.0);
        expect(camera.snapZoom(9.6)).toBe(10.0);
```

src/ui/handler_manager.ts
```diff
@@ -684,6 +688,34 @@ export class HandlerManager {
...
+                    const zoom = this._map.getZoom();
+                    const snap = this._map.getZoomSnap();
+                    let snappedZoom;
+
+                    const deltaSinceStart = zoom - this._zoomStart;
+                    const threshold = 0.01; // Small threshold to detect intentional movement
```
CommanderStorm: What do you think about making this a fraction of `snap`?
For lower `snap`, a lower threshold likely feels better, for higher a slightly higher one.

The reasoning is that if we zoom in and then out, we should not be limited from returning to the original zoom, just because we did not hit the tiny 0.01 zoom threshold.
If snap is `1`, my change would make this threshold `0.05` (which is still small, but you can actually hit it).

Making this threshold be dependent on `snap` also has the advantage of allowing some crazy shenanigans with `snap >> 1`.

```suggestion
                    const threshold = snap / 20; // Small threshold to detect intentional movement
```

CommanderStorm (approved):
Really excited to see this happen.
Thank you, this helps me tremendously ❤️ 

I have three minor comments, but nothing in them would be blocking.
I agree that the one name is a bit awkward, but @HarelM  already mentioned that one.
I don't know what he thinks is not explanatory about the comments in the tests, those are very understandable to me.

src/ui/camera.ts
mizmay: But snapZoom is so snappy! Yes, I will look at this and see if I can improve the grammar.

> I'd like to avoid adding this to the public API as I don't think it's very useful "outside", but I might be wrong...

So revert back to internal method?

src/ui/handler_manager.ts
mizmay: I see the logic of what you are suggesting, I can try it but it is going to be tough to evaluate if it matters, where, and to whom.

Is this in response to building the branch, specifying `zoomSnap` in a Maplibre map, and scrolling the mouse wheel? It is REALLY hard not to overthink these patterns. There are a lot of (literal) moving parts, and what really matters is whether the movement feels natural when you are scrolling.

What Leaflet does instead: inserts a 40ms "debounce" delay at the beginning, and a 250ms ease-in at the end. Both are constants, the first is intended as a threshold to make sure the user is done scrolling (all I'm intending with that 0.01 value), the second to make things less jerky I think.

`snap >> 1` is just the crazy shenanigan zone! Pity the fool who would constrain that.

mizmay:
I will find some flow time to get back into this over the weekend or early next week. @HarelM sorry I forgot how many times I had arithmetic in comments, I will do a thorough pass and make it all a bit more polished. Those comments were useful for my dumb brain while developing the tests. They MIGHT be useful for understanding what the snap/rounding behavior looks like numerically, but people can figure it out.

Thanks to both of you for the other callouts as well.

src/ui/camera.ts
HarelM: Can this be an external utility method instead of a method in the map/camera class? i.e. in the utils.ts file or somewhere?

src/ui/handler_manager.ts
CommanderStorm: I am fine with leaving it as-is and reacting to user input.
Should be easy enough for them to do a PR, because they know the location.

wayofthefuture:
Yes it flies 🛸  on the free wheel 😜

src/ui/handler_manager.ts
wayofthefuture: Yes and also a comment explaining please.

wayofthefuture: Also this logic appears to be giving me odd behavior in ios mobile where the raster tile for the snapped zoom is not completely rendered. see level of detail control example on ios mobile to replicate.

mizmay:
9c581070-a71e-4caa-83b3-5379d05455e0_text

Thanks @wayofthefuture for joining the party... here's a Dune reference that clearly relates to #1591 but luckily I'll survive.

As suggested I:
- Renamed the `snapZoom()` method to something that makes it clearer that it's evaluating and optional parameter that by default does nothing: `evaluateZoomSnap()` (open to better names)
- Moved this method to `utils/utils.ts` and expanded it to handle directional snapping e.g. how you'd expect this to work if you were zooming with a scroll wheel
- Simplified the implementation in `handler_intertia.ts` so it's just referencing the utils method like everyone else
- Cleaned up the tests and comments and described them a bit more clearly (sorry there are a lot of them)

But the most significant thing I did was to add this method in `scroll_zoom.ts`, applied to mouse wheels but not to trackpads. @wayofthefuture @CommanderStorm , when you have a second, test it with your fancy peripherals again? I do not actually know from reading the logic in `scroll_zoom.ts` whether those will be detected as scroll wheels or track pads, but here's why it matters:

The way I am proposing to document this new option in the public API is: "... if `zoomSnap` is 1, the map will snap to whole integers **during discrete zoom operations**. If set to 0, zooming is continuous" (emphasis added). I'm thinking a scroll wheel can count as doing a "discrete zoom operation" and we can snap the zooms. A track pad doesn't, and we just can't (I tried... ).

TLDR, regardless of the value of `zoomSnap`, developers can still access fractional zooms using flyto() and other methods that allow you to specify an explicit numerical zoom, so messing with that would be rude, and users can access fractional zooms with track pads, which are continuous zoom devices, not discrete zoom operators.

Am I splitting the right hairs?

EDIT: it will still fly 🛸  if zoomSnap is set to 1.0, that is the price we pay. There could be some tweaking of the `easeTo` or animation durations that make it smoother, but I decided the existing logic in the handlers was doing a better job than the "threshold" logic @CommanderStorm and I were back and forth about (I decided the threshold just delayed the response and made the map feel laggy).

src/ui/handler/tap_zoom.ts
```diff
@@ -57,7 +58,7 @@ export class TapZoomHandler implements Handler {
             return {
                 cameraAnimation: (map: Map) => map.easeTo({
                     duration: 300,
-                    zoom: tr.zoom + 1,
+                    zoom: evaluateZoomSnap(tr.zoom + 1, map.getZoomSnap()),
```
wayofthefuture, line 61: This pattern is in many places. If you add a function to map.ts named maybe `resolveZoom`, then call the utility function from there if necessary, all of these implementations can be simplified? It would then look like your original implementation but the logic would be abstracted.

```ts
    //map.ts
    resolveZoom(zoom: number, delta?: number): number {
        if (this._zoomSnap > 0) {
            return calculateZoomSnap(zoom, this._zoomSnap, delta);  //from util.ts - renamed from evaluateZoomSnap
        }
        return zoom;
    }
```


But I could be completely wrong. ☹️ Sorry I'm very limited on time right now.

HarelM: There were previous comments about the fact that this adds to the public map API, which we would like to avoid in this case.

CHANGELOG.md
```diff
@@ -9,7 +9,7 @@
 ## 5.16.0
 
 ### ✨ Features and improvements
-
+- Add `zoomSnap` option to `Map` to allow snapping zoom levels to a grid when zooming in and out; aligns the behavior across all UI patterns (keyboard, scroll wheel, on-screen buttons, double-click, double-tap ([#6941](https://github.com/maplibre/maplibre-gl-js/pull/6941)) (by [@mizmay](https://github.com/mizmay))
```
HarelM: This should be moved up under main, as 5.16 was already released :-)

src/util/util.ts
```diff
@@ -497,6 +497,23 @@ export function zoomScale(zoom: number) { return Math.pow(2, zoom); }
...
 
+/**
+ * Evaluates the snapped zoom level based on zoomSnap. If zoomSnap is 0 or less, the zoom level is returned unchanged.
+ * If delta is provided, it performs directional snapping (ceil for zoom-in, floor for zoom-out).
+ * @param zoom - The input zoom level
+ * @param zoomSnap - The grid interval to snap to, e.g. 1.0 for 1.0 zoom levels, 0.5 for 0.5 zoom levels, etc.
+ * @param delta - Optional scroll delta or direction. If positive, snaps up; if negative, snaps down.
+ * @returns The snapped zoom level
+ */
+export function evaluateZoomSnap(zoom: number, zoomSnap: number, delta?: number): number {
```
HarelM, line 508: This looks good! Thanks!

src/util/util.test.ts
```diff
@@ -607,6 +607,35 @@ describe('util scaleZoom and zoomScale relation', () => {
...
+    });
+
+    test('evaluateZoomSnap directional logic', () => {
+        // Snap up
+        expect(evaluateZoomSnap(9.1, 1.0, 1.0)).toBe(10.0);
+        expect(evaluateZoomSnap(9.9, 1.0, 1.0)).toBe(10.0);
+        expect(evaluateZoomSnap(10.0, 1.0, 1.0)).toBe(10.0);
+        expect(evaluateZoomSnap(10.00000001, 1.0, 1.0)).toBe(11.0);
+
+        // Snap down
```
HarelM: this comment is probably an indication that there's a need to split this method into two methods.

[Optional] I would also consider removing the number of expects to cover the code, which probably requires like 5-6 expects?

HarelM:
This looks good over all, maybe we should focus on the simple stuff and leave the scroll wheel to a different PR?
In any case, let me know how you would like to move forward with this.

1ec5:
> users can access fractional zooms with track pads, which are continuous zoom devices, not discrete zoom operators

If the developer has set `zoomSnap` to 1, it’s probably because the content isn’t optimized for a fractional zoom level. Natively, both trackpads and scroll wheel mice can feature continuous scrolling, and both can have inertial scrolling, so a user doesn’t really choose a trackpad over a mouse in order to zoom continuously. Leaflet respects the snapping option regardless of the input device.

wayofthefuture:
I'm good with it as is and a different PR for the added functionality would be good because it's likely complex. Sorry about being late to the party and throwing my opinion.

1ec5:
Yeah, just to be clear, the main thing is the addition of this option. We can always expand it later to apply to more input devices, as long as that’s clearly the future direction. In the meantime, projects like OSM may be able to make do with workarounds if necessary.

src/util/util.test.ts
mizmay: Fixed

CHANGELOG.md
mizmay: Fixed. I was cursing about resolving a merge conflict on a changelog instead of paying attention... oops.

mizmay:
Ok thanks @HarelM I cleaned up the tests.

Writing the unit tests was the most challenging part of this change. The syntax and structures were new to me, I had to rely on AI. Once I went there, I really leaned on the tests for reassurance that I was hitting the right handlers and understanding what the code is doing. We've been through a lot together at this point, me and those tests. It's possible they could be dialed back even more. But the upside is that there is way more coverage now, so @1ec5 if someone is able to take on trackpad zoom snapping, there is coverage there now.

For my part, I've addressed #6872, which was originally only scoped to align the +/- UI buttons with the +/- keyboard buttons. They are aligned now, but since it was the odd one out, this PR changes the _keyboard_ interactions to align with the other handlers, rather than the UI buttons. I stand by this, the UI buttons need to work for vector tile maps too, and should not deviate from the expected behavior across mapping libraries without a more compelling reason. But it makes this a breaking change if anyone was truly relying on keyboard interactions for rounded zooms, and I don't like that.  

Thus we begin to address #1591 by allowing developers to use `zoomSnap: 1.0` to give their users the old rounded zoom keyboard behavior PLUS consistency across UI buttons, double clicks, and double taps. What I like about extending it to scroll wheels, instead of putting that logic in a separate PR is that it gives those interested in using `zoomSnap` something  to test and refine, hopefully quickly, so we can get it right or pull it out. Small side note: I recommend `zoomSnap: 0.5` instead of `zoomSnap 1.0` in most cases where zoom snapping is desired, including cartographic raster tiles. 

I am not comfortable extending this change to trackpads. As mentioned above with my deep reference to Dune, I tried. The snapping causes the map to fly forward, and when applied to a continuous scrolling motion, it loses focus, and the controllers just feel unstable. My second shortest summary of why I can't do it if Leaflet can: what Leaflet measures distance in pixels, Maplibre measures in frame rate, and Maplibre does A LOT with frame rate across GL JS and Native. Maplibre has a heuristic for detecting scroll wheels through a UI pattern, so within the library that makes `wheel` a discrete motion, not a category of peripherals. A click of the wheel and it flies when we snap the zoom, it has to if it is only allowed to land on whole integers, but it sticks to the patterns. Trackpad motion is continuous. To address that would require a larger analysis that gets uncomfortably close to the logic and assumptions around map animations and target zooms, which potentially touches not just zoom and pan operations, but also 3D, navigation, and globe. Hence my shortest summary: I am not that good of a developer.

@1ec5 I believe @CommanderStorm is good with this change and ready to implement it for the Carto OSM map use case. If he were not, I would be less keen on this whole change. Just confirming, with this explanation, does that sound good?

If not, as @HarelM suggests we can pull the scroll zoom logic out and put it in a separate PR. That is the more conservative approach, but also it might get stuck there.

1ec5:
> For my part, I've addressed #6872, which was originally only scoped to align the +/- UI buttons with the +/- keyboard buttons. They are aligned now, but since it was the odd one out, this PR changes the _keyboard_ interactions to align with the other handlers, rather than the UI buttons. I stand by this, the UI buttons need to work for vector tile maps too, and should not deviate from the expected behavior across mapping libraries without a more compelling reason. But it makes this a breaking change if anyone was truly relying on keyboard interactions for rounded zooms, and I don't like that.

Yes, I think that’s fair. It’s usually the developers/designers who want rounded zooms; end users don’t necessarily care. Once we have zoom snapping working for trackpads, that’ll solve the developer/designer need. In the meantime, if anyone needs rounded zooms but has a touch device, they can enable zoom snapping and use the keyboard or on-screen buttons, right?

> I am not comfortable extending this change to trackpads. As mentioned above with my deep reference to Dune, I tried. The snapping causes the map to fly forward, and when applied to a continuous scrolling motion, it loses focus, and the controllers just feel unstable. My second shortest summary of why I can't do it if Leaflet can: what Leaflet measures distance in pixels, Maplibre measures in frame rate, and Maplibre does A LOT with frame rate across GL JS and Native. Maplibre has a heuristic for detecting scroll wheels through a UI pattern, so within the library that makes `wheel` a discrete motion, not a category of peripherals. A click of the wheel and it flies when we snap the zoom, it has to if it is only allowed to land on whole integers, but it sticks to the patterns. Trackpad motion is continuous. To address that would require a larger analysis that gets uncomfortably close to the logic and assumptions around map animations and target zooms, which potentially touches not just zoom and pan operations, but also 3D, navigation, and globe. Hence my shortest summary: I am not that good of a developer.

Don’t be too hard on yourself. Accommodating touch devices across browsers is hard. It even requires an awful hack in Firefox: openstreetmap/iD#8649 webcompat/web-bugs#73148. I think Leaflet gets away with something passable by relying solely on scroll events, but I haven’t looked into how well that translates to the more advanced functionality in GL&nbsp;JS.

CommanderStorm:
> I am not that good of a developer.

What a load of bullcrap, you know it.

mizmay:
@CommanderStorm only Kwizatz Hardeach could write that trackpad logic

mizmay:
@HarelM I feel good about including the scroll wheel logic here instead of pulling it out and putting it in a separate PR, but what do you think? And please remind me, what further steps do I need to take to make sure people are aware of the breaking change(s)?

wayofthefuture:
> @CommanderStorm only Kwizatz Hardeach could write that trackpad logic

@mizmay So if I write the trackpad logic you'll call me Kwiz from now on? 🤣 🤣 😎

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ## main
 ### ✨ Features and improvements
-- _...Add new stuff here..._
+- Add `zoomSnap` option to `Map` to allow snapping zoom levels to a grid when zooming in and out; aligns the behavior across all UI patterns (keyboard, scroll wheel, on-screen buttons, double-click, double-tap ([#6941](https://github.com/maplibre/maplibre-gl-js/pull/6941)) (by [@mizmay](https://github.com/mizmay))
```
HarelM: In order to let people know there's a somewhat breaking change feature use the "warning" emoji as part of this line, check out other lines with this emoji.

HarelM:
So now if someone will set snap zoom and use a touch pad or free mouse wheel they'll get weird behavior, right? 
And we hope that someone that would like to use this feature with a touch pad or fly wheel will take the time to fix this in that case? 
I just want to make sure I understand it correctly. 
Just sharing a personal experience: when I changed the way rotate works on desktop and made some users angree because I broke the way they used to rotate the map I got a lot of heat but not a single PR to address the issue, and it was like 5 lines of very simple code. 
This is to highlight my experience around problematic code.
Don't get me wrong, I don't mind going that route, I think it's a healthy approach.

HarelM:
Looks good overall, just a small change to the changelog entry and this can be merged.

CHANGELOG.md
mizmay: I added the emoji and expanded the explanation in CHANGELOG

mizmay:
> So now if someone will set snap zoom and use a touch pad or free mouse wheel they'll get weird behavior, right? 

Not exactly...
- scroll wheel is going to feel more responsive with `zoomSnap` set to 1.0, because it's rounding up instead of just letting you roll along at fractional zooms
- trackpad behavior is unmodified
- touch pad behavior: pinch to zoom is unmodified, double-tap respects `zoomSnap`

Up to those requesting `zoomSnap` to decide if it's weird, I guess? 

The line I am drawing is that I am respecting all the existing heuristics around what constitutes a mouse wheel versus trackpad movement, and using the animations around responding to UI inputs as defined.

> And we hope that someone that would like to use this feature with a touch pad or fly wheel will take the time to fix this in that case? 

Yes.

> I just want to make sure I understand it correctly. Just sharing a personal experience: when I changed the way rotate works on desktop and made some users angree because I broke the way they used to rotate the map I got a lot of heat but not a single PR to address the issue, and it was like 5 lines of very simple code. This is to highlight my experience around problematic code. Don't get me wrong, I don't mind going that route, I think it's a healthy approach.

Thanks for highlighting, there is always potential for this to happen, the only "risk" I am taking with this PR is by forcing the scroll wheel to respect `zoomSnap`, and I'm doing it because @CommanderStorm and @1ec5 have eyes on it right now and want to use the feature.

The controversy (if there truly is one) is that this change may not fully address the desire expressed in #1591 to never allow fractional zooms, because I excluded the track pad from the change.

If there IS heat about no longer being able to snap to rounded zooms from the keyboard, it will be good to balance that against the desire to restrict landing on fractional zooms more fully.

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ## main
 ### ✨ Features and improvements
-- _...Add new stuff here..._
+- ⚠️ Add `zoomSnap` option to `Map` to allow snapping zoom levels to a grid when zooming in and out; aligns the behavior across all UI patterns (keyboard, scroll wheel, on-screen buttons, double-click, double-tap. Previously, pressing +/- on the keyboard zoomed to rounded whole integers, now by default all zoom handlers zoom in or out by exactly one zoom, unless a value > 0 for `zoomSnap` is specified ([#6941](https://github.com/maplibre/maplibre-gl-js/pull/6941)) (by [@mizmay](https://github.com/mizmay))
```
CommanderStorm: I'd like to be a bit more specific here.
I had to re-read the previous thread to get what is breaking here.

If we include the button handling, this should be much clearer

```suggestion
- ⚠️ Add `zoomSnap` option to `Map` to allow snapping zoom levels to a grid when zooming in and out; aligns the behavior across all UI patterns (keyboard, scroll wheel, on-screen buttons, double-click, double-tap. Previously, pressing +/- on the keyboard zoomed to rounded whole integers, while zoom buttons zoomed exactly one zoom level. Now by default all zoom handlers zoom in or out by exactly one zoom level, unless a value > 0 for `zoomSnap` is specified ([#6941](https://github.com/maplibre/maplibre-gl-js/pull/6941)) (by [@mizmay](https://github.com/mizmay))
```

mizmay: Previously and still the zoom buttons, double-click, and double-tap, everything but the keyboard zoom in exactly one zoom level, I don't think calling out the zoom buttons is more clear.

Does this improve things?

```suggestion
- ⚠️ Add `zoomSnap` option to `Map` to allow snapping zoom levels to a grid when zooming in and out; aligns the behavior across all UI patterns (keyboard, scroll wheel, on-screen zoom buttons, double-click, double-tap). Previously, pressing +/- on the keyboard zoomed to rounded whole integers, more or less than 1 zoom level when starting from a fractional zoom. Now any number can be specified for `zoomSnap`; a value of 1.0 produces the rounded whole integer behavior across all UI patterns. ([#6941](https://github.com/maplibre/maplibre-gl-js/pull/6941)) (by [@mizmay](https://github.com/mizmay))
```
Optionally I could add: `1.0 or 0.5 are preferred values for maps containing raster tiles that may be blurry at fractional zooms.` or further explain the scroll wheel change?

HarelM: Don't over think it... Only a handful of people will actually read this...

HarelM: Do let me know what you decide, I'll merge this once this comment is resolved.

mizmay:
Good to go 👍

PR #6956, opened 2026-01-13
Fix incorrect shader selection for text labels with inline images

ciscorn:
I found that when a tile mixes 'text labels with inline images' and 'plain text labels', the `iconsInText` flag gets overwritten by the last feature processed in the loop. That resets shader selection, so inline images end up going through the SDF-only shader and show up as garbled characters. This PR keeps the flag `true` once any inline image is detected.

Before fix:
bad

After fix: 
ok

```
	'text-field': [
		'case',
		['==', ['get', 'name'], 'foo'],
		['format', ['get', 'name'], ['image', 'osgeo']],
		['format', ['get', 'name']]
	]
```

## Launch Checklist

 - [ ] Post benchmark scores.


HarelM:
Thanks for taking the time to report this issue and fix it!
Can you maybe add a render test to showcase the fix? This way you can see that it fails before the fix and passes after it.

ciscorn:
@HarelM Thanks.
I just added a render test to cover the case where a plain-text label follows a label with an image. Is this sufficient?

Before fix (left) and `expected.png` (right)
image

HarelM (approved):
Great! Thanks!

PR #6963, opened 2026-01-14
Correct bounds calculation for GeoJSON with elevation (RFC 7946)

simonmnt:
Fixes a bug where 3D coordinates ([lng, lat, elevation]) caused incorrect bounding box calculations due to the use of `.flat(Infinity)`. [Support GeoJSON format (RFC 7946)](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.1)

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


src/util/geojson_bounds.ts
```diff
@@ -1,10 +1,21 @@
 import {LngLatBounds} from '../geo/lng_lat_bounds';
 
-function getCoordinatesFromGeometry(geometry: GeoJSON.Geometry): number[] {
+function extractCoordinates(coords: any[]): number[][] {
```
HarelM: Can you improve the type here? it can only be position array (and array of array etc), right?

HarelM:
I don't think we really support elevation in geojson, but I don't mind making the method a bit more robust.

simonmnt:
Elevation is optional but is available on GeoJSON [standard](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.1)

> A position is an array of numbers.  There MUST be two or more elements.  The first two elements are longitude and latitude, or easting and northing, precisely in that order and using decimal numbers.  Altitude or elevation MAY be included as an optional third element.

HarelM:
Yes, I'm familiar with the spec, but unfortunately maplibre doesn't handle it well right now. 
Nested undefined and null properties are not well supported for example although they are part of the spec...

CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
-- Fix incorrect shader selection for text labels with inline images ([#6956](https://github.com/maplibre/maplibre-gl-js/pull/6956)) (by [@ciscorn](https://github.com/ciscorn))
+- Fix correct bounds calculation for GeoJSON with elevation
```
HarelM: Please add a changelog entry with link to this PR and link to your gituhb handle.
Please also restore the previous entry.

simonmnt:
Hi @HarelM,

When this PR is released ? It's very help me on my current project.

Have a nice day.

HarelM:
The CI failed so this wasn't merged and released with the version that was released yesterday. 
I'm guessing it will have to wait for the next version then in about 2 weeks or so...

simonmnt:
The CI failed again ?

HarelM:
You need to update the build size json file...

HarelM:
You also need to resolve the changelog conflict...

HarelM:
The diff here looks weird, can you force push latest main with your changes maybe?

HarelM:
I'm still seeing this in the changelog diff unfortunately - version 5.18 changelog should not be seen as "new"... I'm not sure what's the best way to solve this though...
image

simonmnt:
I don't see CHANGELOG conflict…

HarelM:
Less about conflict, more about the fact that this add stuff "again" that are already there.
The diff here should only show the line you added, which is currently not the case as can be seen from the screenshot I've added...

simonmnt:
Do you want me to close this PR and open a cleaner one?

HarelM:
If it's not too much trouble, please do, sorry...

PR #6966, opened 2026-01-14
typings fix for  Discussions #5463

ricardovg4:
add aroundCenter typing for MapOptions.

No backports from mapbox.
https://github.com/maplibre/maplibre-gl-js/discussions/5463

### DIFF
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -388,10 +388,12 @@ export type MapOptions = {
      */
     experimentalZoomLevelsToOverscale?: number;
     /**
-     * Determines the map rotation horizontal direction. If set to true,
-     * the horizontal rotation direction has two directions, depending if
-     * rotation started above or below the center. If set to false the horizontal
-     * rotation direction will be the same for above and below the center.
+     * Determines the rotation interaction model:
+     * - When true: Uses "Orbital" logic where rotation is relative to the pivot center. 
+     *   Dragging right at the top rotates clockwise, while dragging right at the bottom
+     *   rotates counter-clockwise (like spinning a physical globe).
+     * - When false: Uses "Linear" logic where horizontal mouse movement translates directly
+     *   to bearing change regardless of cursor position.
      */
     aroundCenter?: boolean;
 };


src/ui/map.ts
```diff
@@ -387,6 +387,13 @@ export type MapOptions = {
      * @experimental
      */
     experimentalZoomLevelsToOverscale?: number;
+    /**
+     * Determines the map rotation horizontal direction. If set to true,
```
HarelM: I'm not sure this is the right definition, I think that the difference is between rotating around the center or rotating around the starting point of the drag or something like that.
I would recommend playing with it and seeing what it does, unless you already did that...

HarelM:

src/ui/map.ts
ricardovg4: Yes the variable name aroundCenter suggests the type of rotation you mention, however I think the only difference is the rotation horizontal direction if above or below the center point. The rotation is always kept around the map center. You can check these videos below:

aroundCenter = false;
https://github.com/user-attachments/assets/d0dc9d45-02b2-4bc4-9578-3316c3309492

aroundCenter = true;
https://github.com/user-attachments/assets/70045267-b4fc-49ff-bb3b-b6c64e98bdb0

Let me know if I'm mistaking something here. Thank you!

HarelM: This also changes the behavior of how sensitive the rotation is compared to how far you are from the center.
I think this was the reason for this variable name.
I changed that code a long time ago, so I need to dig into it again, but the idea is that is not pixel only movement, but also the far you are from the center the smaller the movement is - the idea was that you grap and move the map and it "stays" under your mouse, which is not the case when around center is false.
I'm not sure I was able to make my self clear, but try to drag the map in both options in different distance from the map center and see how it behaves...

HarelM: Sorry for nit-picking on the comment, but since we are adding it, we might as well make sure it is descriptive enough of the functionality it changes...

ricardovg4: I can't see the rotation being affected by distance between camera and the center of rotation being affected by aroundCenter. Maybe an improvement to the comment would be this:

Determines the rotation interaction model:
- When true: Uses "Orbital" logic where rotation is relative to the pivot center. Dragging right at the top rotates clockwise, while dragging right at the bottom rotates counter-clockwise (like spinning a physical globe).
- When false: Uses "Linear" logic where horizontal mouse movement translates directly to bearing change regardless of cursor position.

HarelM: I think that's accurate. Thanks.

PR #6969, opened 2026-01-15
feat: add cross-window support for popup windows and iframes

### LINKED ISSUE #6964: Support cross-window rendering (container in different window context)
## User Story

As a web developer, I can render a MapLibre map into a popup window while keeping the JavaScript context in the main window, so that I can build "pop-out" features with shared application state.

Modern web applications implement "pop-out" features that allow users to open content in separate windows while maintaining shared state. This is achieved using `window.open('about:blank')` combined with `ReactDOM.createPortal()` or direct DOM manipulation, where JS runs in the main window but UI renders into the popup's DOM.

Currently, MapLibre fails in this scenario because it uses window-scoped globals (`window`, `document`, `instanceof HTMLElement`, `instanceof MouseEvent`) that don't work across window contexts.

### Example use cases

-   **Multi-window web applications**: Trading platforms, analytics dashboards, or collaboration tools where users can pop out widgets to secondary monitors
-   **Electron/Tauri/WebView-based desktop apps**: Desktop applications built with web technologies often need to support multiple windows sharing the same JS context
-   **Picture-in-picture style features**: Detachable map panels that float independently while staying connected to the main application
-   **Kiosk and digital signage**: Multi-display setups where a single application drives content across multiple screens/windows

### Can it be done today? Is there a workaround?

**No viable workaround exists.**

The only alternative is to load MapLibre separately in the popup window, but this:

-   Breaks shared state with the main application
-   Requires duplicating the library and styles
-   Prevents React portal patterns that many apps rely on
-   Adds complexity for cross-window communication (postMessage, etc.)

### What if we do nothing?

Developers building multi-window web applications cannot use MapLibre for pop-out features. They must either:

1. Abandon the pop-out pattern for map-containing components
2. Use a different map library that supports cross-window rendering
3. Maintain a custom fork of MapLibre with cross-window fixes

This limits MapLibre's adoption in enterprise web applications and Electron/WebView-based desktop apps where multi-window support is a common requirement.

---

## Minimal Reproduction

Save as HTML and open in browser. Click the button to open a map in a popup window.

```html
<!DOCTYPE html>
<html>
<head>
    <script src="https://unpkg.com/maplibre-gl@5.0.0/dist/maplibre-gl.js"></script>
</head>
<body>
    <button onclick="openMapInPopup()">Open Map in Popup Window</button>
    <script>
        function openMapInPopup() {
            const popup = window.open('about:blank', '_blank', 'width=800,height=600');
            popup.document.write(`
                <link rel="stylesheet" href="https://unpkg.com/maplibre-gl@5.0.0/dist/maplibre-gl.css">
                <div id="map" style="width:100%;height:100%"></div>
            `);
            popup.document.close();

            // Container is in popup, but MapLibre JS runs in main window
            new maplibregl.Map({
                container: popup.document.getElementById('map'),
            });
        }
    </script>
</body>
</html>
```

**Symptoms:**
1. Map fails to render. Console shows: `Uncaught Error: Invalid type: 'container' must be a String or HTMLElement.`
2. Even after bypassing the container validation, the map renders but does not respond to any mouse or keyboard events.

## Proposed Implementation

Derive `ownerDocument` and `ownerWindow` from the container element instead of using globals:

1. Use `container.ownerDocument` and `ownerDocument.defaultView` instead of global `document`/`window`
2. Bind document/window-level events to the container's ownerDocument/ownerWindow
3. Use duck typing for event detection instead of `instanceof`
4. Pass `ownerWindow` to `requestAnimationFrame()` and `ResizeObserver`

I have a working implementation and would be happy to submit a PR.

Syncret:
This PR adds support for rendering MapLibre GL JS maps in popup windows (window.open) and iframes, fixing several issues that occur when the map container is in a different window context than the main window.
Fixes #6964

Changes:
- map.ts: Add _ownerWindow property; use nodeType check instead of instanceof HTMLElement; get ResizeObserver from ownerWindow
- handler_manager.ts: Add _ownerDocument/_ownerWindow; bind events to correct document/window instead of global
- util.ts: isPointableEvent uses duck typing via event target's window context to get correct constructor for instanceof checks
- browser.ts: frame() and frameAsync() accept optional targetWindow parameter for requestAnimationFrame calls
- style.ts, marker.ts, load_tilejson.ts, vector_tile_source.ts, raster_tile_source.ts: Pass targetWindow to browser.frameAsync()

Fixes issues where:
1. instanceof checks fail across window contexts
2. Events bound to wrong document/window don't fire
3. requestAnimationFrame pauses when main window is not focused

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


src/ui/handler_manager.ts
```diff
@@ -162,6 +162,18 @@ export class HandlerManager {
...
+     * The document that contains the map container element.
+     * Used for cross-window support when the map is in a popup window or iframe.
+     */
+    _ownerDocument: Document;
+    /**
+     * @internal
+     * The window that contains the map container element.
+     * Used for cross-window support when the map is in a popup window or iframe.
+     */
+    _ownerWindow: Window;
```
HarelM: Can this be calculate only when needed instead of storing it?

src/source/load_tilejson.ts
```diff
@@ -22,13 +22,14 @@ export async function loadTileJson(
     options: RasterSourceSpecification | RasterDEMSourceSpecification | VectorSourceSpecification,
     requestManager: RequestManager,
     abortController: AbortController,
+    targetWindow?: Window,
```
HarelM, line 25: This feels weird, can we store the window object in the browser "class" somehow instaed of passing it around?

src/ui/handler_manager.ts
```diff
@@ -217,17 +236,18 @@ export class HandlerManager {
             [el, 'wheel', {passive: false}],
             [el, 'contextmenu', undefined],
 
-            [window, 'blur', undefined]
+            // Cross-window support: use ownerWindow for popup window scenarios.
```
HarelM: No need to repeat the comment, just add it to the class' field as a tsdoc comment.

src/ui/map.ts
```diff
@@ -754,15 +777,18 @@ export class Map extends Camera {
...
             const throttledResizeCallback = throttle((entries: ResizeObserverEntry[]) => {
                 if (this._trackResize && !this._removed) {
                     this.resize(entries);
                     this.redraw();
                 }
             }, 50);
-            this._resizeObserver = new ResizeObserver((entries) => {
+            // Cross-window support: use the owning window's ResizeObserver.
+            // The global ResizeObserver may not properly observe elements in other windows.
+            const ResizeObserverClass = (this._ownerWindow as typeof window).ResizeObserver || ResizeObserver;
```
HarelM: These "as something" looks like bad typings definitions, please try and reduce these as much as possible.

src/ui/map.ts
```diff
@@ -754,15 +777,18 @@ export class Map extends Camera {
...
+            this._ownerWindow.addEventListener('online', this._onWindowOnline, false);
             let initialResizeEventCaptured = false;
             const throttledResizeCallback = throttle((entries: ResizeObserverEntry[]) => {
                 if (this._trackResize && !this._removed) {
                     this.resize(entries);
                     this.redraw();
                 }
             }, 50);
-            this._resizeObserver = new ResizeObserver((entries) => {
+            // Cross-window support: use the owning window's ResizeObserver.
```
HarelM: Remove comments and only keep them on the ownerWindow field.

HarelM:
Thanks for taking the time to open this PR! Besides the over verbosity in comments this looks good.

src/source/load_tilejson.ts
Syncret: I did consider it, but the browser object is a global singleton, so storing a single window reference there wouldn't work correctly when there are multiple Map instances in different windows on the same page. Passing targetWindow as a parameter allows each Map to use its own owner window context independently.

src/ui/handler_manager.ts
Syncret: Sure! I changed to getters that calculate on-demand with fallback to global document/window. Together updated the _ownerWindow in Map class.

src/ui/map.ts
Syncret: as typeof window is the most type-safe option I can come up with. ResizeObserver is a global in TypeScript, not a Window property, so type assertion seems to be required. We need the target window's ResizeObserver because the global one doesn't observe elements in other window contexts.

src/ui/map.ts
Syncret: Sure, I removed most redundant comments.

Syncret:
Hi @HarelM , I've addressed most feedbacks, simplified comments, and also added a unit test (map_cross_window.test.ts) to cover the new code and cross-window scenario. 
The only failing check is bundle size so I increased the bundle_size.json from 1014715 to 1016647. Let me know if it's the correct approach. 
The PR should be ready for review now. Thanks!

Syncret:
Hi, may I ask that once this PR is merged, roughly how long does it typically take for changes to land in a release? We have a feature that would benefit from this fix, and knowing the typical timeline would help us plan accordingly.

HarelM:
Usually we release twice a month, but if this is urgent I can release right after this gets merged.

Syncret:
> Usually we release twice a month, but if this is urgent I can release right after this gets merged.

Thanks @HarelM ! Twice a month is good enough for us.

src/ui/map.ts
```diff
@@ -754,15 +770,18 @@ export class Map extends Camera {
...
+            // The global ResizeObserver may not properly observe elements in other windows.
+            const ResizeObserverClass = (this._ownerWindow as typeof window).ResizeObserver ?? ResizeObserver;
```
HarelM: Isn't ownerWindow already of type window? I'm confused by this line...

src/ui/map.ts
```diff
@@ -754,15 +770,18 @@ export class Map extends Camera {
...
+            this._ownerWindow.addEventListener('online', this._onWindowOnline, false);
             let initialResizeEventCaptured = false;
             const throttledResizeCallback = throttle((entries: ResizeObserverEntry[]) => {
                 if (this._trackResize && !this._removed) {
                     this.resize(entries);
                     this.redraw();
                 }
             }, 50);
-            this._resizeObserver = new ResizeObserver((entries) => {
+            // Cross-window support: use the owning window's ResizeObserver.
```
HarelM: Maybe this part should be moved to a method to initialize resize observer with all its quirks?

src/ui/map.ts
```diff
@@ -733,6 +742,13 @@ export class Map extends Camera {
             }
         } else if (resolvedOptions.container instanceof HTMLElement) {
             this._container = resolvedOptions.container;
+        } else if (
+            // Cross-window support: use nodeType check as instanceof fails across windows
```
HarelM: Same for all this container logic? Move to a separate method?

src/ui/map_tests/map_cross_window.test.ts
```diff
@@ -0,0 +1,88 @@
...
+                interactive: false,
+                attributionControl: false,
+                style: {
+                    version: 8,
+                    sources: {},
+                    layers: []
+                }
+            });
+
+            // Map should be created successfully
```
HarelM: Please reduce test comments...

src/ui/map.ts
Syncret: @HarelM  Actually it also confused me.
The _ownerWindow getter returns type Window, but ResizeObserver is declared as a global variable in TypeScript's DOM types, not as a property of the Window interface.

If you go to the type definiation of window object, in lib.dom.d.ts, you can see:
`declare var window: Window & typeof globalThis;`
`declare var ResizeObserver: { ... };`

So typeof window = Window & typeof globalThis, which includes global constructors like ResizeObserver. The type assertion as typeof window is needed to access these globals from a Window-typed variable.

src/ui/map.ts
Syncret: Sure, I've moved this part to a private function _resolveContainer

src/ui/map_tests/map_cross_window.test.ts
Syncret: Sure, removed most redundant comments.

src/ui/map.ts
HarelM: Can you change the return type of `get _ownerWindow` to match this type instead of using casting or it will only move the problem there?

src/ui/map_tests/map_cross_window.test.ts
```diff
@@ -0,0 +1,79 @@
...
+            if (iframe && iframe.parentNode) {
+                iframe.parentNode.removeChild(iframe);
+            }
+        });
+
+        test('container from iframe works with Map', () => {
+            expect(iframeDocument).not.toBe(window.document);
+            expect(iframeWindow).not.toBe(window);
+
+            // Cross-window instanceof issue: container is NOT instanceof main window's HTMLElement
```
HarelM: Please remove this, this is what the expect is testing, no need for a comment.

src/ui/map_tests/map_cross_window.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+            expect(map).toBeTruthy();
+            expect(map.getContainer()).toBe(container);
+            expect(map._ownerWindow).toBe(iframeWindow);
+
+            const handlers = map.handlers;
+            expect(handlers._ownerWindow).toBe(iframeWindow);
+            expect(handlers._ownerDocument).toBe(iframeDocument);
+
+            // Basic map operations should work
```
HarelM: This comment is an indication that another test is needed here, this test is too long and have too many asserts...

src/ui/map_tests/map_cross_window.test.ts
```diff
@@ -0,0 +1,79 @@
...
+                    layers: []
+                }
+            });
+
+            expect(map).toBeTruthy();
+            expect(map.getContainer()).toBe(container);
+            expect(map._ownerWindow).toBe(iframeWindow);
+
+            const handlers = map.handlers;
+            expect(handlers._ownerWindow).toBe(iframeWindow);
```
HarelM: Can we check something else here? This is more of an internal API that I would like to avoid asserting on...

src/ui/map.ts
HarelM: Can you also move all the resize logic to a method like you did for the container?

Syncret: Done. 
Changed return type of _ownerWindow to typeof window which includes global constructors. 
Extracted this part to _setupResizeObserver() method.

src/ui/map_tests/map_cross_window.test.ts
Syncret: Done. Removed this comment.

src/ui/map_tests/map_cross_window.test.ts
Syncret: Done. Split into two tests: initializes with cross-window container and map operations work with cross-window container.

src/ui/map_tests/map_cross_window.test.ts
Syncret: Makes sense. I removed assertions on internal APIs (_ownerWindow, _ownerDocument). The test now only verifies that Map initializes and works correctly with cross-window containers. The other test "map operations work with cross-window container" will make sure normal operations will work for map.

HarelM:
Looks great now, thanks!!

Issue #6248, opened 2025-08-12
GeoJSON layers partially rendered twice with globe projection


pabueco:
Layers referencing GeoJSON sources seem to get rendered twice (partially) when globe projection is used.

Some possibly relevant details:
- The duplicate section seems to originate from the antimeridian and spread outwards east and west.
- The `buffer` property 'controls' how much (or up until which longitude) the layers is rendered a second time. The default appears to be `128`. A buffer of `0` essentially gets rid of the issue, but produces other artifacts. A buffer of `512` essentially causes shapes all around the globe to be rendered twice.

https://github.com/user-attachments/assets/71b3eb68-a956-4e36-89db-8245177810d5

**maplibre-gl-js version**: 5.6.2 (also tested with 5.0.0) 

**browser**: all, tested with Chrome 139 and Firefox 142

### Steps to Trigger Behavior

 1. Create map with globe projection (`style: { projection: { type: 'globe' } }`)
 2. Add GeoJSON source (somewhat close to the antimeridian) and layer(s) referencing that source
 3. Zoom out quite far or move camera close to poles
 4. Notice the layers seem to get rendered twice (at least partially)

### Link to Demonstration

https://codepen.io/distinctshapes/pen/wBKPwYa/b3a3fa2a4348302d82301cc68215b08d

### Expected Behavior

Layers only get rendered once, regardless of the view and used `buffer` value. 

### Actual Behavior

Layers get rendered twice (partially)

HarelM:
Thanks for taking the time to report this issue! 
Unfortunately, this is not the first issue with globe drawing twice and I'll guessing it is related to antimeridian you mentioned and stencil mode.
Feel free to look into it. 
Cc: @kubapelc @mwilsnd

pabueco:
I finally had time to look into this and it indeed seems to be related to the stencil mode. Or at least it was fixable (in some capacity) by using an adjusted stencil mode for drawing layers near the antimeridian. 

The PR with the fix is [here](https://github.com/maplibre/maplibre-gl-js/pull/6970). I'm not sure how viable this approach really is or what possible side effects there are, but it did fix the issue for me (for the most part), so I wanted to share it. Even if there are issues with it I was hoping it could serve as a starting point or inspiration for someone with more experience in this area.

maxargy:
Thank you @HarelM for shitching me to this issue.
As I sayd in my closed issue, increasing the buffer the glitch increase, deceasing to 0 no more glitches.
[...]

PR #6970, opened 2026-01-15
fix(globe): GeoJSON layers rendered twice near antimeridian

pabueco:
This is an attempt to fix issue https://github.com/maplibre/maplibre-gl-js/issues/6248

Note: While it seems to fix the issue for me I'm not sure how viable that approach/fix actually is.

The root cause for the bug outlined in the issue seems to be related to `geojson-vt` wrapping at the antimeridian and the fact that globe projection has no world copies. I first attempted to adjust the wrap function in `geojson-vt` to clip the features that cross the antimeridian, but that resulted in a overlap/line at the antimeridian, which I couldn't get rid of.

The approach that worked was leaving the `geojson-vt` wrapping as is and instead adjusting the stencil mode for tiles at/near the antimeridian to only draw on a pixel once per layer. 

This removed the overdraw entirely (before and after):

image image

The code used in this example is a slightly [adjusted codepen](https://codepen.io/distinctshapes/pen/wBKPwYa/b3a3fa2a4348302d82301cc68215b08d) linked in the issue.

As you can see the last thing that remains are the vertical lines at the edge of the previous overdraw. I feel like this should be fixable in a similar way but using the same fix in `draw_line.ts` did not help. 

However what did help was, changing the `buffer` value passed to `geojson-vt` to `256` (default is `128`). Either in the `workerOptions` [here](https://github.com/maplibre/maplibre-gl-js/blob/8d04baafb3300d83829fae6870ba16df7214ffac/src/source/geojson_source.ts#L192) or when adding the geojson source to the map:
```ts
map.addSource('rectangles', {
  'type': 'geojson',
  'data': geojson,
  'buffer': 256,
});
```

With this we get the exact result we're looking for:

image

Here is the [benchmark result](https://github.com/user-attachments/assets/6a64e0a5-3e13-4207-89c6-d082cd044781) for `LayerFill` (I assumed this is the only relevant test case).

### DIFF
--- a/src/render/draw_fill.ts
+++ b/src/render/draw_fill.ts
@@ -25,7 +25,7 @@ export function drawFill(painter: Painter, tileManager: TileManager, layer: Fill
         return;
     }
 
-    const {isRenderingToTexture} = renderOptions;
+    const {isRenderingToTexture, isRenderingGlobe} = renderOptions;
     const colorMode = painter.colorModeForRenderPass();
     const pattern = layer.paint.get('fill-pattern');
     const pass = painter.opaquePassEnabledForLayer() &&
@@ -37,7 +37,13 @@ export function drawFill(painter: Painter, tileManager: TileManager, layer: Fill
     if (painter.renderPass === pass) {
         const depthMode = painter.getDepthModeForSublayer(
             1, painter.renderPass === 'opaque' ? DepthMode.ReadWrite : DepthMode.ReadOnly);
-        drawFillTiles(painter, tileManager, layer, coords, depthMode, colorMode, false, isRenderingToTexture);
+        drawFillTiles(painter, tileManager, layer, coords, depthMode, colorMode, false, isRenderingToTexture, isRenderingGlobe);
+
+        // For globe projection, the fill pass uses overdraw prevention on antimeridian tiles,
+        // which corrupts their stencil values. Restore them before the outline pass.
+        if (isRenderingGlobe) {
+            painter.restoreAntimeridianClippingMasks(coords, isRenderingToTexture);
+        }
     }
 
     // Draw stroke
@@ -53,7 +59,13 @@ export function drawFill(painter: Painter, tileManager: TileManager, layer: Fill
         // the (non-antialiased) fill.
         const depthMode = painter.getDepthModeForSublayer(
             layer.getPaintProperty('fill-outline-color') ? 2 : 0, DepthMode.ReadOnly);
-        drawFillTiles(painter, tileManager, layer, coords, depthMode, colorMode, true, isRenderingToTexture);
+
+        drawFillTiles(painter, tileManager, layer, coords, depthMode, colorMode, true, isRenderingToTexture, isRenderingGlobe);
+
+        // Restore masks after outline pass for subsequent layers
+        if (isRenderingGlobe) {
+            painter.restoreAntimeridianClippingMasks(coords, isRenderingToTexture);
+        }
     }
 }
 
@@ -65,7 +77,8 @@ function drawFillTiles(
     depthMode: Readonly<DepthMode>,
     colorMode: Readonly<ColorMode>,
     isOutline: boolean,
-    isRenderingToTexture: boolean) {
+    isRenderingToTexture: boolean,
+    isRenderingGlobe: boolean) {
     const gl = painter.context.gl;
     const fillPropertyName = 'fill-pattern';
     const patternProperty = layer.paint.get(fillPropertyName);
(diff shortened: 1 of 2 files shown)


HarelM:
We are working geojson-vt fixes right now, please let me know if there are things that might be better of moved there.
I saw the code related to wrapping, maybe it would make sense to add `worldCopies` boolean to geojson options to let it know that it doesn't need to replicate the features.
It will also make things a bit faster I believe.
CC: @wayofthefuture

pabueco:
@HarelM That's what I thought about as well at first, and it did generally work, but I couldn't get around the following (rendering) issue:

So in `geojson-vt` I modified the wrap function to split the features into a left and right part and used that when rendering a globe. In maplibre itself this approach would only require slight changes to switch the wrap mode when rendering the globe. I can open a PR for this approach as well, if you're interested. But anyway, this essentially was the change in `geojson-vt`:

```js
export function wrapGlobe(features, options) {
    const result = [];
    
    const left = clip(features, 1, 0, 1, 0, -1, 2, options);
    if (left) result.push(...left);
    
    const right = clip(features, 1, 1, 2, 0, -1, 2, options);
    if (right) result.push(...shiftFeatureCoords(right, -1));
    
    return result;
}
```

This eliminates the double draw, but it introduces this line at the antimeridian for line layers. I guess this happens because we now have two features and they both receive their own outline.

image

But it's not just outlines, even if we just have a fill layer, we still get this very slight overlap at the antimeridian:

image

I could image this being solved when rendering, but I don't know how. Maybe you have an idea?

HarelM:
Yeah, I see what you mean.
If the fix in geojson is simpler and "more correct" I think it would make sense to have that there, although I'm not sure how easy it would be to make sure it works when you switch between globe and mercator.
regarding the lines you said that in both solutions you see the overdraw, right?

CC: @mwilsnd, maybe he can take a look from a graphics perspective on this.

pabueco:
I guess the geojson-vt solution is simpler, because it doesn't involve stencils and it seems more correct, because it actually adjusts the underlying data instead of just eliminating the symptoms when rendering. I mainly went with the current fix, because it worked in maplibre directly and didn't involve another dependency.

Handling the switch between globe and mercator could be done by adding a `onProjectionChange` method to (GeoJSON)sources, which is then called by `Style._setProjectionInternal` and can be used to update the worker (options). I didn't test this thoroughly at the time, but it seemed to work.

Yes, with both solutions we get the vertical lines (either via overdraw of fill layers or the stroke on line layers). In the first stencil based solution they 'move' based on the set `buffer` size. In the second solution they are located exactly at the antimeridian, because we disabled the wrapping. This is similar to setting `buffer: 0` with the first solution, but without the additional artefacts that disabling the buffer for the entire globe introduces.

wayofthefuture:
@pabueco Not sure if you are aware but we onboarded geojson-vt to Maplibre and are actively developing on it. So we welcome any improvements!

HarelM:
geojson-vt was on-borded and introduced into maplibre.
So fixing this issue there is possible if it's the right place.
@wayofthefuture mentioned that both geojson-vt and maplibre are wrapping features and we might want to remove the wrapping in the geojson-vt package and keep it only in maplibre.
I'm not sure I know what the right approach, but it seems that it might be a better solution, IDK.
@pabueco let me know if you would like to push this fix. If you need help setting anything up and testing please let me know.

wayofthefuture:
With 5.21.0 I see no overlap with the reproduction above... not related but the globe fading is quite nice lol

HarelM:
I took the codepen link from the linked issue and was still able to reproduce this with the latest version (5.21.0):

image

It requires some moving the globe around to make it visible, but not that hard to reproduce.

pabueco:
@HarelM Sorry for the delay here, finally getting back into it.

So preventing the wrapping in `geojson-vt` definitely seems to be the right approach. What remains is that without wrapping (so `worldCopies: false`), we now get two separate features (left and right of the antimeridian). This causes two issues (as mentioned in [this comment](https://github.com/maplibre/maplibre-gl-js/pull/6970#issuecomment-3767654564) earlier):
1. A small ~1px overlap of fill styles at the antimeridian
2. An extra line/border down the antimeridian if the feature has a stroke

As far as I can tell, these issues stem from the fact that we now actually cut off the features instead of wrapping them, which produces new (artificial) lines at the antimeridian. I tried a bunch of different approaches, but always landed on the idea that we need to filter out these edges in maplibre at some point. 

So the new proposed fix for the double rendering consists of:
- the new `worldCopies` flag in `geojson-vt` ([PR](https://github.com/maplibre/geojson-vt/pull/112))
- edge filtering in `maplibre` ([PR](https://github.com/maplibre/maplibre-gl-js/pull/7528))

This seems quite a bit cleaner (although a bit bigger) than the approach in this PR, so closing this one in favour of the PRs mentioned above.

HarelM:
Is it possible the the clipping is done in geojson-vt on the tile border instead of taking into account the buffer around the tile? If so, this might solve this...?
geojson-vt should mimic the regular vector tiles behavior, does this bug happens on regular vector tiles?

HarelM:
- Stopping the discussion here and continuing it in: #7528

Issue #6850, opened 2025-12-12
FR: Use etags to skip work for unchanged vector tiles


rivkamatan:
## User Story

As a developer working with maps that contain a lot of vector tile data and are used on underresourced machines, I would like to introduce the ability for the vector tile loading process to respect etag matches such that it will skip parsing and rendering when a tile that was loaded hasn’t changed.  

## Rationale

* Currently, Maplibre will respect the max-age header and refetch tiles at the given interval via the [TileManager](https://github.com/maplibre/maplibre-gl-js/blob/f6125d99c6885f2acae85f73c0d63f00f889b16c/src/tile/tile_manager.ts#L717) (expiry data is set via max-age headers [here](https://github.com/maplibre/maplibre-gl-js/blob/f6125d99c6885f2acae85f73c0d63f00f889b16c/src/tile/tile.ts#L456) and [here](https://github.com/maplibre/maplibre-gl-js/blob/f6125d99c6885f2acae85f73c0d63f00f889b16c/src/tile/tile.ts#L407)). However, vector tile loading, parsing, and rendering works the same way regardless of whether the tile has changed since we last requested it or not. 
* Even when there is a local or remote cache hit (304), such that the tile is exactly the same, all of the work that happens after getting the response from loadVectorTile is repeated. This includes [WorkerTile#parse()](https://github.com/maplibre/maplibre-gl-js/blob/f6125d99c6885f2acae85f73c0d63f00f889b16c/src/source/worker_tile.ts#L63) and firing a data event in [TileManager](https://github.com/maplibre/maplibre-gl-js/blob/bc547e98db900b4b29de85b42805ab35168acf45/src/tile/tile_manager.ts#L298) to trigger repaint and rendering. 
* If we maintained a mapping of tile ID to etag and compared the existing etag for a given tile with the etag from the tile fetch response, we could skip all of the `WorkerTile#parse()` work for that tile, and instead immediately set the loaded tile to the cached tile for that ID. This would reduce recomputation when the tile hasn’t changed. 
[...]

PR #6972, opened 2026-01-15
Skip work for vector tiles loaded from browser cache with etag match

rivkamatan:
Fixes https://github.com/maplibre/maplibre-gl-js/issues/6850. 

## Description
This PR introduces the ability for the vector tile loading process to use etag headers in order to skip duplicative parsing and rendering work when the tile that was requested hasn't changed. 

### Approach
- Update the `Source` API so that `loadTile` can optionally return a boolean, indicating if the tile was loaded from cache or not. 
- Store the latest etag on the `Tile` itself. Use the etag on the Tile in `VectorTileSource` to pass into the worker via the `loadTile` params . 
- Expose the etag header to the vector source worker.  
- When loading vector tiles, the worker compares the existing etag on the Tile provided in the params to the etag on the response header. If they're equivalent, the worker returns a new `WorkerTileResult` type indicating that the tile is unchanged, and skips tile parsing. When `VectorTileSource`  receives `data` with an `unchanged` type, it also skips parsing work and returns `true` from `loadTile`. 
- When `TileManager` receives `true` from `loadTile`, it skips modifications to the tile and doesn't fire a data event, which means repaint is avoided.

## Background
* Currently, Maplibre will respect the max-age header and refetch tiles at the given interval via the [TileManager](https://github.com/maplibre/maplibre-gl-js/blob/f6125d99c6885f2acae85f73c0d63f00f889b16c/src/tile/tile_manager.ts#L717) (expiry data is set via max-age headers [here](https://github.com/maplibre/maplibre-gl-js/blob/f6125d99c6885f2acae85f73c0d63f00f889b16c/src/tile/tile.ts#L456) and [here](https://github.com/maplibre/maplibre-gl-js/blob/f6125d99c6885f2acae85f73c0d63f00f889b16c/src/tile/tile.ts#L407)). However, vector tile loading, parsing, and rendering works the same way regardless of whether the tile has changed since we last requested it or not. 
* Even when there is a local or remote cache hit (304), such that the tile is exactly the same, all of the work that happens after getting the response from loadVectorTile is repeated. This includes [WorkerTile#parse()](https://github.com/maplibre/maplibre-gl-js/blob/f6125d99c6885f2acae85f73c0d63f00f889b16c/src/source/worker_tile.ts#L63) and firing a data event in [TileManager](https://github.com/maplibre/maplibre-gl-js/blob/bc547e98db900b4b29de85b42805ab35168acf45/src/tile/tile_manager.ts#L298) to trigger repaint and rendering. In the case of a cache hit, this work is not necessary.

## Launch Checklist
 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/source/vector_tile_source.ts
+++ b/src/source/vector_tile_source.ts
@@ -245,6 +245,7 @@ export class VectorTileSource extends Evented implements Source {
             }
             this._afterTileLoadWorkerResponse(tile, null);
         }
+        return false;
     }
 
     /**
--- a/src/tile/tile_manager.test.ts
+++ b/src/tile/tile_manager.test.ts
@@ -2337,9 +2337,6 @@ describe('TileManager.reload', () => {
 
     test('skips appropriate work if tile is loaded from cache hit', async () => {
         const tileManager = createTileManager();
-
-        const spyFireEvent = vi.spyOn(tileManager._source, 'fire');
-
         const tileID = new OverscaledTileID(1, 0, 1, 0, 1);
         const tileEtag = 'etag';
 
@@ -2350,9 +2347,8 @@ describe('TileManager.reload', () => {
         };
         const tile = await tileManager._addTile(tileID);
 
-        expect(spyFireEvent).toHaveBeenCalled();
-
-        const spyReload = vi.spyOn(tileManager, 'reload');
+        const dataEventSpy = vi.fn();
+        tileManager.on('data', dataEventSpy);
         tileManager._source.loadTile = async () => {
             tile.timeAdded = undefined;
             return true;
@@ -2363,7 +2359,7 @@ describe('TileManager.reload', () => {
         expect(tile.timeAdded).toBeDefined();
 
         // Cache hit should not trigger an event
-        expect(spyFireEvent).not.toHaveBeenCalledAfter(spyReload);
+        expect(dataEventSpy).not.toHaveBeenCalled();
 
         // Tile data should remain the same after a cache hit
         expect(tile.etag).toBe(tileEtag);


src/css/maplibre-gl.css
HarelM, line 1: Can you check the history of this file? I think this was reverted to this notation explicitly.
In any case, I don't think it is related to this PR so I would suggest reverting it or introducing in a different PR.

src/source/geojson_worker_source.test.ts
```diff
@@ -46,12 +46,14 @@ describe('reloadTile', () => {
...
         const firstData = await source.reloadTile(tileParams as any as WorkerTileParameters);
+        expect(firstData.type).toBe('processed');
         expect(spy).toHaveBeenCalledTimes(1);
 
         // second call won't give us new rawTileData
         let data = await source.reloadTile(tileParams as any as WorkerTileParameters);
         expect('rawTileData' in data).toBeFalsy();
-        data.rawTileData = firstData.rawTileData;
+        expect(data.type).toBe('processed');
+        (data as WorkerTileProcessedResult).rawTileData = (firstData as WorkerTileProcessedResult).rawTileData;
```
HarelM: Not sure I understand why this casting and assignment is done here...?

src/source/geojson_worker_source.test.ts
```diff
@@ -100,11 +102,12 @@ describe('reloadTile', () => {
...
         // load vector data from geojson, passing through the tile serialization step
         const data = await source.reloadTile(tileParams as any as WorkerTileParameters);
-        expect(data.featureIndex).toBeDefined();
+        expect(data.type).toBe('processed');
+        expect((data as WorkerTileProcessedResult).featureIndex).toBeDefined();
 
         // deserialize tile layers in the feature index
-        data.featureIndex.rawTileData = data.rawTileData;
-        const featureLayers = data.featureIndex.loadVTLayers();
+        (data as WorkerTileProcessedResult).featureIndex.rawTileData = (data as WorkerTileProcessedResult).rawTileData;
```
HarelM: Same here...

src/source/source.ts
```diff
@@ -89,9 +89,10 @@ export interface Source {
     /**
      * This method does the heavy lifting of loading a tile.
      * In most cases it will defer the work to the relevant worker source.
+     * Returns true if tile was loaded from browser cache and parsing was skipped. Further processing and repainting of the tile is not necessaryin this case.
```
HarelM:
```suggestion
     * @returns `true` if tile was loaded from browser cache and parsing was skipped. Further processing and repainting of the tile is not necessary in this case. `false` otherwise.
```

src/source/vector_tile_source.ts
```diff
@@ -231,6 +232,8 @@ export class VectorTileSource extends Evented implements Source {
                 return;
             }
             this._afterTileLoadWorkerResponse(tile, data);
+
+            return data && data.type === 'unchanged';
```
HarelM: This might not return a boolean in some cases - use `data?.type === 'unchanged' ? true : false`.

src/source/vector_tile_source.ts
```diff
@@ -269,6 +272,10 @@ export class VectorTileSource extends Evented implements Source {
             tile.resourceTiming = data.resourceTiming;
         }
 
+        if (data && data.etag) {
```
HarelM: use `.?` operator.

src/source/vector_tile_worker_source.ts
```diff
@@ -70,12 +84,23 @@ export class VectorTileWorkerSource implements WorkerSource {
     async loadVectorTile(params: WorkerTileParameters, abortController: AbortController): Promise<LoadVectorTileResult> {
         const response = await getArrayBuffer(params.request, abortController);
         try {
-            const vectorTile = params.encoding !== 'mlt' 
-                ? new VectorTile(new Protobuf(response.data)) 
+            if (params.etag && response.etag && params.etag === response.etag) {
```
HarelM, line 87: This check, as far as I understand, still means that the full tile is fetched from the network. I would expect to see the etag passed down to the ajax `getArrayBuffer` call in order to use the etag before fetching the entire tile.
I might be reading this wrong so please keep me honest here.

src/source/vector_tile_worker_source.ts
```diff
@@ -119,16 +144,16 @@ export class VectorTileWorkerSource implements WorkerSource {
                 return null;
             }
 
-            if (overzoomParameters) {
+            if (overzoomParameters && response.type === 'full') {
```
HarelM: This seems like a weird addition to this if, does this mean that having the etag breaks the slice-instead-of-overzoom feature?

src/tile/tile_manager.ts
```diff
@@ -186,8 +186,8 @@ export class TileManager extends Evented {
 
     async _loadTile(tile: Tile, id: string, state: TileState): Promise<void> {
         try {
-            await this._source.loadTile(tile);
-            this._tileLoaded(tile, id, state);
+            const isTileLoadedFromCacheHit = await this._source.loadTile(tile);
+            this._tileLoaded(tile, id, state, isTileLoadedFromCacheHit === true);
```
HarelM, line 190: This should be a boolean... why is the `isTileLoadedFromCacheHit === true` used here?

HarelM:
I did a quick review, and it looks good overall I think.
I'm not sure I'm fully on board with splitting the types into two types instead of making optional fields, but I understand why you chose to do it, and I guess it's fine...

I would be interested to know if this can be used for raster source as well, not that you'd need to implement it as part of this PR, but I would like to know that the design decisions made in this PR make sense in order to implement this if needed in the future for raster or other sources.

src/css/maplibre-gl.css
rivkamatan: This is unintentional, I will revert

src/source/geojson_worker_source.test.ts
rivkamatan: Since the `expect` statement doesn't narrow the type of data, there would be an error that `rawTileData` may not exist on the new `WorkerTileResult` type without the cast. I can implement this type fix a different way if you'd prefer.

src/tile/tile_manager.ts
rivkamatan: The type inferred was `boolean | void` here and `void` isn't falsey. I'm not sure why it was inferred that way since when I try in the TS playground it isn't an issue. Have you encountered that before?

src/source/vector_tile_worker_source.ts
rivkamatan: The browser itself is storing the etag as part of the cache management. It will send the etag representing the cached version of the resource along with the request, and if the server's version of the resource has not changed, the server will respond with a `304 Not Modified` status with no body. The browser will then use the existing version of the resource as the response body for the request. 

However, all this happens at a layer of abstraction deeper than the ajax utils have access to. The 304 is reported at this level as a 200 status since the browser was able to successfully load the requested item. Therefore, comparing the old etag on the Tile with the response's etag header is the best way to detect if a 304 occurred.  

A network request is still happening here (there's no way to avoid this if you want to check whether the tile has changed since it was last requested) but the bytes don't have to go over the network. (And if they did, at least we'd be saving work at this step by not repainting the tile).

src/source/vector_tile_worker_source.ts
rivkamatan: Good flag. I don't think it has to break the feature, but I'm not totally sure what the right interaction between the features is. Do you think it would make sense to say that if we get an overzoomed tile it should always have the 'full' type so that we always process + repaint in response to it?

rivkamatan:
> I would be interested to know if this can be used for raster source as well, not that you'd need to implement it as part of this PR, but I would like to know that the design decisions made in this PR make sense in order to implement this if needed in the future for raster or other sources.

While I haven't looked as deeply at other source's loading processes, I'd expect the approach in this PR to work for other sources. The work required would primarily be in the other source's `loadTile` implementation: (1) exposing the etag header from the network response, (2) comparing it to the one saved to the tile to skip work within that source's loadTile implementation, and (3) either return an unchanged result type and skip work, or save the etag to the tile and process as usual. 
Work upstream of that, like skipping repaint when you get an unchanged result type, should be reusable across source types.

src/source/vector_tile_worker_source.ts
HarelM: What about the request to fetch a tile after it was expired but it's still the same? 
I'm not expecting to avoid network calls, but I am expecting to the the etag in the ajax class in some way.

src/source/vector_tile_worker_source.ts
HarelM: Overzoom tiles are requesting lower level zoom from the network and then splicing the returned tile.
Why did you decided to add that it? Did it break a test?

src/tile/tile_manager.ts
HarelM: I don't think so, probably somewhere where nothing is returned so the type is either boolean or void, see if you can spot a code path that doesn't return, or add return value explicitly to loadTile method.

src/source/vector_tile_worker_source.ts
rivkamatan: There's no `response.rawData` or `response.vectorTile` (needed in the if statement) if the response type is unchanged.

src/tile/tile_manager.ts
rivkamatan: If I add a return value explicitly, it will break the existing Source API, so I didn't think that was worth it

src/source/vector_tile_worker_source.ts
rivkamatan: The browser will handle sending the etag automatically via an [if-none-match](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/If-None-Match) header. That happens a level of abstraction deeper than our AJAX utils, via the browser's cache, so we don't need to deal with etag logic in the maplibre ajax code.

HarelM: Have you tested it to make sure this is what's happening? Just to be on the safe side...

src/tile/tile_manager.ts
HarelM: I would prefer to have the API a bit more strict, but it might be a breaking change as you said, so it might be worth doing for version 6 and not right now.
I'll add it to the list of version 6 breaking changes.
- #6427

src/source/vector_tile_source.ts
```diff
@@ -231,6 +232,8 @@ export class VectorTileSource extends Evented implements Source {
                 return;
             }
             this._afterTileLoadWorkerResponse(tile, data);
+
+            return data?.type === 'unchanged';
```
HarelM, line 236: What about returning something in other places in this method to make it a bit more explicit?

src/tile/tile_manager.test.ts
```diff
@@ -2335,6 +2335,41 @@ describe('TileManager.reload', () => {
 
     });
 
+    test('skips work if tile is loaded from cache hit', async () => {
+        const tileManager = createTileManager();
+
+        const spyTileLoaded = vi.spyOn(tileManager, '_tileLoaded');
```
HarelM: Is there a way to test this without using internal method stubbing for the class under test?

src/tile/tile.test.ts
```diff
@@ -124,6 +125,28 @@ describe('querySourceFeatures', () => {
         expect(spy).toHaveBeenCalledWith();
     });
 
+    test('loadVectorData does not unload existing data if the data type is unchanged', () => {
+        const tile = new Tile(new OverscaledTileID(1, 0, 1, 1, 1), undefined);
+
+        const spy = vi.spyOn(tile, 'unloadVectorData');
```
HarelM, line 131: Can we do this test by using the "public" API?

src/source/vector_tile_worker_source.ts
HarelM: Ahh, I see, can you move this if condition to be below the added `response.type === 'unchanged'` check?
This way there will not be a need to check if it's full since if it is unchanged it will return early before this if, I think, do check me on this.

src/source/vector_tile_worker_source.test.ts
```diff
@@ -246,6 +249,150 @@ describe('vector tile worker source', () => {
...
+        server.respondWith(request => {
+            request.respond(200, {
+                'Content-Type': 'application/pbf',
+                'ETag': newEtag,
+                'Cache-Control': newCacheControl
+            }, '');
+        });
+
+        // Mock PBF and VT so the mock server response body doesn't cause errors
+        vi.doMock('pbf', () => ({default: class Protobuf { constructor() { return {} as Protobuf; }}}));
```
HarelM: Is there a different way to achieve this test without mocking `pbf` and `@mapbox/vector-tile`? Seems a bit of an overkill, I think...

src/source/geojson_worker_source.test.ts
HarelM: Maybe just create a local variable before the expects? I don't know...

HarelM:
I just remembered we added recently `shouldReloadTile` as part of the source interface, can we use that instead of returning something from loadTile or is it too late down the code flow?

HarelM:
This looks good overall, sorry for dragging the PR...

src/source/vector_tile_source.ts
rivkamatan: Sure! I wasn't sure if it was better to keep surface area of this change down or to make it more explicit so kept it lower. Happy to add

src/tile/tile_manager.test.ts
rivkamatan: Not understanding the concern here - `spyOn` just wraps the method without stubbing it unless a mock implementation is provided. Is the issue that the test is looking at internal methods at all?

src/tile/tile.test.ts
rivkamatan: Do you mean by checking the tile's state directly? The issue I had with that is that `loadVectorData` rewrites the `state` field to `loaded` immediately after calling `unloadVectorData` so I didn't see a good way to check.

src/tile/tile_manager.test.ts
HarelM: Yes, if we change the internal method we need to change the test, which is less than ideal...

src/source/vector_tile_worker_source.test.ts
rivkamatan: I agree that the mocking is a lot. The issue was that I couldn't find a way to return a valid string response from the mock server as the request body (providing invalid strings there means that PBF and MVT will throw an error down the line). Do you know a good way to return a validly-formatted test string for the response body?

src/tile/tile.test.ts
HarelM: Maybe it's too specific to test, or you might want to make sure something else is happening...

src/source/vector_tile_worker_source.test.ts
HarelM: I think there are other tests that do it, IIRC, if you can't find one let me know and I'll look too. It should be possible.

rivkamatan:
> I just remembered we added recently `shouldReloadTile` as part of the source interface, can we use that instead of returning something from loadTile or is it too late down the code flow?

It's not in a good part of the code flow for this application unfortunately. In this case, we don't want to skip the loading part of the codepath because we do need to send a network request so the server can tell us whether the etag has changed or not. We just want to skip what happens *after* the load. `shouldReloadTile` is useful if you know you can skip the entire load process.

src/tile/tile_manager.test.ts
rivkamatan: Got it, I can check that the timing fields updated appropriately instead of spying on the method directly

src/tile/tile.test.ts
rivkamatan: I could do this by checking that the vector data fields aren't changed, e.g. `buckets` is a different object before and after running `loadVectorData`. I do think spying on the method is the most direct way to test this though. It's based on the existing test directly above it.

src/source/vector_tile_worker_source.ts
rivkamatan: Yes! I've tested it with a development server so I can experiment with max-age and etag headers, and the tile requests consistently have a 304 status when they should.

src/tile/tile_manager.ts
```diff
@@ -295,15 +295,21 @@ export class TileManager extends Evented {
...
+        if (!isTileLoadedFromCacheHit && tile.selfFading) {
             tile.fadeEndTime = tile.timeAdded + this._rasterFadeDuration;
         }
 
         if (previousState === 'expired') tile.refreshedUponExpiration = true;
         this._setTileReloadTimer(id, tile);
+
+        if (isTileLoadedFromCacheHit) {
+            // No data event or further processing is needed because the tile is unchanged.
+            return;
```
rivkamatan, line 310: @HarelM By the way, do we need to add a new event type that is fired here to signal the source data is done loading without triggering the effect of repainting?

HarelM: I honestly don't know, let start without firing a new event and see if we encounter any issues.

HarelM:
> It's not in a good part of the code flow for this application unfortunately. In this case, we don't want to skip the loading part of the codepath because we do need to send a network request so the server can tell us whether the etag has changed or not. We just want to skip what happens after the load. shouldReloadTile is useful if you know you can skip the entire load process.

I'm not sure this is what happens for the goejson and I recall this check is done after calling the geojson worker source (by it might be just the set/update data).

The more interesting question is if there's a good way to combine those, as they seems to try and support very similar behavior and I think it's worth the extra thought to see if there's a way to unite them to some extent.

I'm not sure it makes a lot of sense, but if you could try and deep a bit deeper to better explain why both sources can't use this optional method it would be great.
Sorry again for the extra work, I just want to make sure we are not create two similar mechanisms for the same problem...

src/source/vector_tile_source.ts
HarelM: I think there's still one missing at the end of the method...?

src/tile/tile_manager.test.ts
```diff
@@ -2335,6 +2335,40 @@ describe('TileManager.reload', () => {
...
+        tileManager._source.loadTile = async (tile) => {
+            tile.state = 'loaded';
+            tile.etag = tileEtag;
+            return false;
+        };
+        const tile = await tileManager._addTile(tileID);
+
+        expect(spyFireEvent).toHaveBeenCalled();
+
+        const spyReload = vi.spyOn(tileManager, 'reload');
```
HarelM: Consider removing this spy and just making sure spyFireEvent is not fired.

src/tile/tile_manager.test.ts
```diff
@@ -2335,6 +2335,40 @@ describe('TileManager.reload', () => {
 
     });
 
+    test('skips appropriate work if tile is loaded from cache hit', async () => {
+        const tileManager = createTileManager();
+
+        const spyFireEvent = vi.spyOn(tileManager._source, 'fire');
```
HarelM: Can't we simply use `tileManager._source.on('something', spy)`?

rivkamatan:
> The more interesting question is if there's a good way to combine those, as they seems to try and support very similar behavior and I think it's worth the extra thought to see if there's a way to unite them to some extent.

It's a good flag, thank you for pointing it out! Looking into it more, I'm still thinking the `shouldReloadTile` solves a pretty different problem than this one. In more detail: 

The `shouldReloadTile` callback is used here https://github.com/maplibre/maplibre-gl-js/blob/e00b513418b3221af7a6dd964fcf19d68ae86349/src/tile/tile_manager.ts#L257-L279. The code flow on this path is:

(1) `TileManager#reload`
(2) `TileManager#_reloadTile`
(3)`TileManager#_loadTile` 
----(4) `Source#loadTile` 
-------(5) Source-specific worker `loadTile` implementation 
----(6) Source-specific post-worker response 
(7) `TileManager#_tileLoaded`
(8) Map responds to ‘data’ event from source
… etc. 

The `shouldReloadTile` callback causes the code to short-circuit **before step 2**, which means we don’t even try to load the tile at all, and we don’t update any timing or expiration fields on the tile.  It relies on information that we have available before trying to load the tile at all to skip reloads.
The vector etag optimization needs to **complete step 5** to discover that the tile has not changed, and needs to do some work in steps 6 and 7 to update the timing and expiration fields like timeAdded and refreshedUponExpiration on the tile, since we did make an attempt to load the tile. It can only fully skip step 8 and beyond. It relies on information that we get from the server after trying to load the tile and then learning from the server that it hasn't changed.

As a result, we can’t omit the code in this PR that allows the worker to report to the source in step (5) that the data is unchanged, and allows the source to report to the tile manager in step (6) that the tile is unchanged so the tile manager can skip firing the event and skip steps (8) and beyond. 

If we wanted to remove shouldReloadTile from the Source API, and we were okay with GeoJson tiles having their timing/expiration data (e.g. timeAdded, refreshedUponExpiration) updated even in the shouldReloadTile false case, I see a way to do that. I suspect that’s not desirable though – let me know if I’m wrong here and I can get into more detail on that option.

HarelM:
Fair enough, I guess making the tests pass would be good enough to allow merging this.
Thanks for all the hard work on this!

wayofthefuture:
I appreciate your work on this, but I believe this PR is not ready. In my opinion this is a code smell:

```
const isTileLoadedFromCacheHit = await this._source.loadTile(tile);
this._tileLoaded(tile, id, state, isTileLoadedFromCacheHit === true);
```

We have multiple caches so the fact this is etag related creates more confusion as well as the inlined conditional. The addition of the 4 multiple types to support etag is not necessary in my opinion and it can be supported by using the cacheControl object which is already passed through the pipeline. I believe this can be done without the addition of extra types - and the naming of these types is not related to etag even though the purposes of them is for etag. If the tile is unmodified then the state in tile manager should remain unchanged - except for the reload timer / expiry data. Also, tile.refreshedUponExpiration can be deleted as it is unused.

The code referenced above is basically the end of the pipeline, and my vision is more like this:

```
//tile_manager.ts - _loadTile()
const result = await this._source.loadTile(tile);
if (result.etagUnmodified) {
    this._setTileReloadTimer(id, tile);
    return;
}
this._tileLoaded(tile, id, state);
```
Then an estimate of the changes elsewheres:

```

//vector_tile_worker_source.ts - loadVectorTile
//this needs to be extracted out - then we don't need the two return types.
const response = await getArrayBuffer(params.request, abortController);
//delete:
if (params.etag && response.etag && params.etag === response.etag) {
    return {
        type: 'unchanged',
        etag: response.etag,
        cacheControl: response.cacheControl,
        expires: response.expires
    };
}

//vector_tile_worker_source.ts - loadTile
//the caller can perform the etag check and call loadVectorTile if needed
const tileResponse = await this.downloadTileData(params, abortController);
const etagUnmodified = (params.etag && params.etag === tileResponse.etag);
if (etagUnmodified) {
    const cacheControl = {etagUnmodified};
    const result: WorkerTileResult = extend({}, cacheControl);
    return result;
}
const response = await this.loadVectorTile(params, abortController);
//now we only have 1 return type

//vector_tile_source.ts - I suggest simply pass WorkerTileResult here or we can create a loadTileResult if needed.
async loadTile(tile: Tile): Promise<WorkerTileResult> {
...
const result = await tile.actor.sendAsync({type: messageType, data: params}, tile.abortController);
this._afterTileLoadWorkerResponse(tile, result);
return result;
//now we've removed the confusing boolean
...
//_afterTileLoadWorkerResponse
if (!data?.etagUnmodifed) tile.loadVectorData(data, this.map.painter);
or inside of tile.loadVectorData() instead
if (data?.etagUnmodified) return;
if (this.hasData()) ...
```

Or something like that... I think these changes eliminate the need for all the extra types, eliminate the need for the boolean or to pass conditionals through parameters, and leverage the already existing object pipeline for cache control. If I am missing anything please let me know...

wayofthefuture:
@rivkamatan @HarelM I should be careful throwing up recommendations like above... because I just tried to implement my suggestion to only find that this becomes a types can-of-worms due to geojson worker. I was able to get etag really simplified and almost finished, but the breaking factor was that geojson worker loadVectorTile is used by the vector worker, the parameters are locked, so we can't modify it without affecting geojson. Once the other PR I have up is merged I think this PR becomes much easier because we can modify the vector worker without breaking the geojson worker. I'm happy to throw up my branch if you guys want to see, or I could throw up the etag changes on top of the other branch, but nevertheless my suggestions above were great in theory but did not actually work, so I apologize for that!

HarelM:
@rivkamatan are you ok with the idea of merging the geojson refactoring first, before this PR, in order to simplify the code here? 
I wanted to reduce the work for you in terms of conflict resolution since you opened this PR before the other one, but if the other PR can simplify the code here I prefer to "move it up the queue".
Let me know what you think.

wayofthefuture:
As I was trying to resolve the merge conflicts on the other branch I decided to just add the etag support as a branch off that branch - here is my take on it. It is not perfect - the types are a pain - but this is the best I could do:

https://github.com/maplibre/maplibre-gl-js/compare/geojson-worker-abstract...geojson-worker-abstract-etag?expand=1

HarelM:
Here's the relevant PR by @wayofthefuture:
- https://github.com/maplibre/maplibre-gl-js/pull/7074

I don't see significant changes between the solution, besides the simplification of the `isTileLoadedFromCacheHit` which I think is a good simplification, although @rivkamatan might have solved some other issues there, not sure...

rivkamatan:
@HarelM @wayofthefuture Thank you for the thoughts here! Merging the geojson abstraction first sounds like the right path to keep these changes independent. 

I took a look at #7074 and the idea looks neat to me. Would you like me to update this PR to match that idea and own merging it, or should I review that one and close this one?

HarelM:
@rivkamatan your call, whichever is easier for you. We can update the changelog to include you from credit pount of view.

rivkamatan:
Closing this in favor of #7074, thank you @wayofthefuture for the work on this improvement!

PR #6988, opened 2026-01-19
Refactor geojson_worker_source and test

wayofthefuture:
The GeoJSON worker had an odd construction due to a test. The test was modified to spy on the class method instead of passing a method in the constructor. Methods were consolidated into the class and the constructor was removed. I didn't change any of the supercluster code even though the compare doesn't look right on github, looks good on webstorm though.

### DIFF
--- a/.github/workflows/test-all.yml
+++ b/.github/workflows/test-all.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           persist-credentials: false
-      - uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v4
+      - uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v4
         with:
           node-version-file: '.nvmrc'
       - run: npm ci
@@ -49,7 +49,7 @@ jobs:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           persist-credentials: false
-      - uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v4
+      - uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v4
         with:
           node-version-file: '.nvmrc'
       - name: Start display server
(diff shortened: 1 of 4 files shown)


src/source/geojson_worker_source.test.ts
```diff
@@ -392,13 +392,14 @@ describe('loadData', () => {
     });
 
     test('loadData should process cluster change with no data', async () => {
-        const mockCreateGeoJSONIndex = vi.fn(createGeoJSONIndex);
-        const worker = new GeoJSONWorkerSource(actor, layerIndex, [], mockCreateGeoJSONIndex);
+        const worker = new GeoJSONWorkerSource(actor, layerIndex, []);
+        const spy = vi.spyOn(worker, '_createGeoJSONIndex');
```
HarelM, line 396: I originally added this method to the constructor. The idea behind this is coming from SOLID principles related to inversion of control.
This change basically "knows" the internal methods of geojson source, which I would like to avoid in order to make sure the tests are maintainable.
I've recently suffered a refactoring pain when a lot of the tests were replacing and using internal methods instead of using the public API of the class.
So I would like to avoid this pain in the future if possible...

wayofthefuture: You're the pro but I believe tests should be built around the code and not the other way around.

The createGeoJSONIndexFunc in the constructor is very confusing.

Testing of private methods seems important -- rather than attempting to test a private method through another public method. The former allows for both options and more granular testing.

wayofthefuture: And hey...

Code readability is based on SOLID principle too right?

And I can relate... I got burned once by marriage but that doesn't mean I shouldn't ever get married again right?

Touché? Squash and merge? 🤣 😂 🤣 ❗ ❗

HarelM: > You're the pro but I believe tests should be built around the code and not the other way around.

The other part of this equation, and the interesting one, is that if you are having a hard time testing your code, it means that you need to change it, so writing tests should make you change your code, so that it will be easier to test it.

> The createGeoJSONIndexFunc in the constructor is very confusing.

I agree, but this is a classic factory pattern, I wish it was used more in this repo instead of all the "new" sprinkled around, which makes tests harder to write and break the "I" of the SOLID principles.

> Testing of private methods seems important -- rather than attempting to test a private method through another public method. The former allows for both options and more granular testing.

If there's an internal method that you can't get to from the tests you are trying to write using the public APIs, then you might need to split the code into smaller pieces.

The bottom line is that the public API of a file or class should be tested in unit tests.
The pain I felt was not about the changes I needed to make to the code, the pain was because I knew it's a bad habit to allow using internal/private fields in unit tests but I still allowed it when reviewing code...

wayofthefuture: Is there any solution that moves these functions inside of the class and is more readable?

HarelM: You might be able to replace the classes with vitest so that calling `new Somthing` will trigger some vitest magic and return a mocked class instead of the real one.
I'm trying to use this option as little as possible because it's hard to see it when you are looking at a test, but it's better than mocking methods of the class you are testing...

PR #7002, opened 2026-01-23
fix: correct context for setTerrain in setStyle (Fixes #6824)

### LINKED ISSUE #6824: Unexpected error when terrain is loading when `setStyle` is called
**maplibre-gl-js version**: 5.14.0 (latest)

**browser**: All browsers i believe

### Steps to Trigger Behavior

 1. Set a map with terrain.
 2. Let the map flying or zoom in.
 3. When flying or zoom in is still happening, use `map.setStyle()` to set a new style with terrain. 

### Link to Demonstration

[Example](https://jsfiddle.net/0uq6p5gc/3/)

To use this example:

1. Click the compass button to let the map move.
2. Before map move is finished, set the map to another style.
3. Error occurs.

### Expected Behavior

No error and the map is able to set a new style with terrain.

### Actual Behavior

There is an error:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'shaderPreludeCode')
useProgram	@	painter.ts:729
(anonymous)	@	draw_terrain.ts:23
maybeDrawDepthAndCoords	@	painter.ts:653
render	@	painter.ts:519
_render	@	map.ts:3483
(anonymous)	@	map.ts:3617
(anonymous)
```

### Use-case
Users should be able to change basemap layers freely no matter the terrain is ready or not.

### Explanation
I belive that when `map.setStyle()` is called, the `map.painter.style.projection` is immediately set to undefined at initialization untill the correct projection is set. As a result, if the terrain is still drawing, in the `Painter.useprogram()` implementation, this following line will lead to an error
```
const projectionPrelude = forceSimpleProjection ? shaders.projectionMercator : projection.shaderPreludeCode;
```
I'm new to maplibre internal implementation so lmk if I'm wrong.

Might be related to #6184 and #6817

### Solution

Change the following line in `Painter.maybeDrawDepthAndCoords()`:
```
-        if (!this.style || !this.style.map || !this.style.map.terrain) {
+        if (!this.style || !this.style.map || !this.style.map.terrain || !this.style.projection) {
```
I could create a PR but I need to learn how to properly write tests as per the contribution guideline and see if it breaks anything else..

biplavbarua:
This PR fixes a race condition where calling `setStyle` while terrain is active/loading caused a crash due to incorrect `this` binding in `Style.setState`.

### Fix
Correctly binds `this.map` when calling `setTerrain` during style update operations.

### Verification
- Reproduction: Confirmed the crash/warning is gone with the fix.
- Tests: All unit tests passed.

Fixes #6824

### DIFF
--- a/src/style/style.ts
+++ b/src/style/style.ts
@@ -936,7 +936,7 @@ export class Style extends Evented {
                     operations.push(() => this.setSprite.apply(this, op.args));
                     break;
                 case 'setTerrain':
-                    operations.push(() => this.map.setTerrain.apply(this, op.args));
+                    operations.push(() => this.map.setTerrain.apply(this.map, op.args));
                     break;
                 case 'setSky':
                     operations.push(() => this.setSky.apply(this, op.args));


HarelM:
Looks like you took a wrong branch and tied to add fixes to it.
Please take the time to review the PR before you open it.
Or at least set it to draft so that I'd know it's not ready for review.
This is the last warning, next time I'll simply close the PR.

HarelM:
Please submit a different PR, and make sure to review it yourself first.

PR #7007, opened 2026-01-25
Fix no update on source feature state changes when terrain is enabled

### LINKED ISSUE #6231: map.setFeatureState freezes updates under high CPU load when zooming.
**maplibre-gl-js version**: 5.6.2

**browser**: Chrome 139 / Firefox 141 

### Steps to Trigger Behavior

1. Use a large GeoJSON source with terrain enabled.
(I’m not entirely sure terrain is the cause, the issue might be related to high memory usage instead. 🤔)
2. Call setFeatureState in response to a mousemove event.
3. Create a layer with paint or layout properties depending on feature-state.
4. Move the mouse over the map while zooming out.

### Link to Demonstration

https://jsbin.com/geyakevage/edit?html,console,output

https://github.com/user-attachments/assets/146ec1b0-1964-4041-8620-2d4714004d50

### Expected Behavior

5. Layers depending on feature-state should continue updating.
6. If CPU usage is high, updates might temporarily freeze, but they should eventually resume after a short wait ⏳.

### Actual Behavior

5. Even though setFeatureState is still being called, the map stops updating accordingly.
6. Waiting does not resolve the issue, updates never resume.
7. Performing a zoom in/out action is required to “unlock” the updates.

The behavior is inconsistent: sometimes it happens immediately, sometimes not at all.
If anyone has ideas or workarounds, I’d be happy to test them out and share the results 🙌😊

pstaszek:
- Fixes #6231 

There was no mechanism to update terrain tile textures on source feature state change. Proposed solution is to introduce `revision` for source state. When revision of a source state changes the `RenderToTexture` class compares it with local reference, and update terrain tile textures if needed.

Before:

https://github.com/user-attachments/assets/c931fc05-f815-4de4-8ada-169f6acd5ca7

After:

https://github.com/user-attachments/assets/9f22dd6f-d564-475f-a269-5781ca346767

Sample example to verify the code before/after (from #6231).
```HTML
<!DOCTYPE html>
<html>

<head>
    <meta charset="utf-8" />
    <title>MapLibre Demotile - OMT</title>
    <meta name="viewport" content="initial-scale=1,maximum-scale=1,user-scalable=no" />
    <style>
        body {
            margin: 0;
            padding: 0;
        }

        #map {
            position: absolute;
            top: 0;
            bottom: 0;
            width: 100%;
        }
    </style>
</head>

<body>
    <div id="map"></div>
    <script>
        function createPolygonFeature(coords) {
            const featureId = `${coords[0][0]}-${coords[0][1]}`;
            return {
                id: featureId,
                type: "Feature",
                geometry: {
                    type: "Polygon",
                    coordinates: [coords],
                },
                properties: {
                    shape: "rectangle",
                },
            };
        }

        function createCollection() {
            const step = 0.2;
            const size = 0.75 * step;
            const geoJson = {
                type: "FeatureCollection",
                features: [],
            };

            for (let lng = -1; lng < 1; lng += step) {
                for (let lat = 50; lat > 48; lat -= step) {
                    const feature = createPolygonFeature([
                        [lng, lat],
                        [lng + size, lat],
                        [lng + size, lat - size],
                        [lng, lat - size],
                        [lng, lat],
                    ]);

                    geoJson.features.push(feature);
                }
            }

            return geoJson;
        }

        let hoveredFeature = null;

        const map = (window.map = new maplibregl.Map({
            container: "map",
            zoom: 6,
            center: [0, 49],
            style: {
                version: 8,
                sources: {},
                layers: [],
            },
        }));
        const canvas = map.getCanvasContainer();

        const terrainSource = {
            type: "raster-dem",
            tiles: [
                "https://s3.amazonaws.com/elevation-tiles-prod/terrarium/{z}/{x}/{y}.png",
            ],
            encoding: "terrarium",
            tileSize: 256,
            maxzoom: 15,
        };

        map.on("load", () => {
            map.addSource("terrarium-terrain", terrainSource);

            map.setTerrain({
                source: "terrarium-terrain",
                exaggeration: 1,
            });

            map.addSource("base", {
                type: "geojson",
                data: createCollection(),
                generateId: true,
            });
            map.addLayer({
                source: "base",
                id: "base-hover",
                type: "line",
                paint: {
                    "line-color": [
                        "case",
                        ["boolean", ["feature-state", "hover"], false],
                        "#ff0000",
                        "transparent",
                    ],
                    "line-opacity": 1,
                    "line-width": 7,
                },
            });
            map.addLayer({
                source: "base",
                id: "base-fill",
                type: "fill",
                paint: {
                    "fill-color": "#ff5600",
                    "fill-opacity": [
                        "case",
                        ["boolean", ["feature-state", "hover"], false],
                        1,
                        0.1,
                    ],
                },
            });
            map.addLayer({
                source: "base",
                id: "base-stroke",
                type: "line",
                paint: {
                    "line-color": "#278cda",
                    "line-opacity": 1,
                    "line-width": 3,
                },
            });

            map.on("mousemove", `base-fill`, (e) => {
                if (!e.features || e.features.length < 1) return;
                const nextFeature = e.features[0];

                if (hoveredFeature && nextFeature.id === hoveredFeature.id) {
                    console.log("bypass");
                    return;
                } else {
                    console.log("set feature state");
                }
                if (hoveredFeature) {
                    map.setFeatureState(
                        { source: "base", id: hoveredFeature.id },
                        { hover: false }
                    );
                }
                map.setFeatureState(
                    { source: "base", id: nextFeature.id },
                    { hover: true }
                );
                canvas.style.cursor = "pointer";
                hoveredFeature = nextFeature;
            });
            map.on("mouseleave", `base-fill`, () => {
                if (hoveredFeature) {
                    map.setFeatureState(
                        { source: "base", id: hoveredFeature.id },
                        { hover: false }
                    );
                }
                hoveredFeature = null;
                canvas.style.cursor = "";
            });
        });
    </script>
</body>

</html>
```

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
That's a very clever solution! Kudos!

src/render/render_to_texture.ts
```diff
@@ -63,11 +63,16 @@ export class RenderToTexture {
      * a list of all layer-ids which should be rendered
      */
     _renderableLayerIds: Array<string>;
+    /**
+     * map of source-id to source-revision
+     */
+    _sourceRevisions: Record<string, number>;
```
HarelM: Does it make sense to utilize _coordsAscendingStr for this somehow? It looks like both do a  similar thing...

pstaszek: Yes! Done, you had a very good point, now it requires much less code.
I also changed naming a bit and added/changed comments to better reflect the content of those objects.

src/render/render_to_texture.ts
```diff
@@ -98,26 +98,29 @@ export class RenderToTexture {
...
-        this._coordsAscendingStr = {};
+        this._rttFingerprint = {};
         for (const id of style._order) {
             const layer = style._layers[id], source = layer.source;
             if (LAYERS[layer.type]) {
-                if (!this._coordsAscendingStr[source]) {
-                    this._coordsAscendingStr[source] = {};
+                if (!this._rttFingerprint[source]) {
+                    this._rttFingerprint[source] = {};
+                    const revision = style.tileManagers[source] ? style.tileManagers[source].getState().revision : 0;
```
HarelM: Will it make it a bit shorter to use `style.tileManagers[source]?.getState().revision ?? 0`

HarelM: It might be worth to also reduce indentation by calling `continue` for some ifs, this code is very nested...

src/render/render_to_texture.ts
```diff
@@ -36,10 +36,11 @@ export class RenderToTexture {
...
     _coordsAscending: {[_: string]: {[_:string]: Array<OverscaledTileID>}};
     /**
-     * create a string representation of all to tiles rendered to render-to-texture tiles
-     * this string representation is used to check if tile should be re-rendered.
+     * fingerprint string representing the unique state of source tiles and revision
+     * for a given render-to-texture tile. Used to detect changes and trigger re-rendering.
+     * Format: "sorted_tile_keys#revision"
      */
-    _coordsAscendingStr: {[_: string]: {[_:string]: string}};
+    _rttFingerprint: {[sourceId: string]: {[rttTileKey: string]: string}};
```
HarelM: [nit picking] `_rttFingerprints` (extra s) to distinguish it from the per tile fingerprint?

src/render/render_to_texture.test.ts
```diff
@@ -130,7 +136,7 @@ describe('render to texture', () => {
                 ]
             }
         });
-        expect(rtt._coordsAscendingStr).toStrictEqual({maine: {'923': '923'}});
+        expect(rtt._rttFingerprint).toStrictEqual({maine: {'923': '923#0'}});
```
HarelM: It is possible to use a more "public" API here?
I know you didn't wrote this test, but I'd be happy to fix this if possible.

Consider this optional.

src/render/render_to_texture.ts
pstaszek: Sure, done.

src/render/render_to_texture.ts
pstaszek: Changed.

src/render/render_to_texture.test.ts
pstaszek: I'll try tomorrow to refactor this test.

I have side question - is it possible to check which and why render tests failed? I can't see any information when I check the preview of failed job.

HarelM: Yes, you should be able to download from the summary of the run the report of the relevant render test and open the html to see the differences.
You can run that test locally to try and solve it.

pstaszek: I changed the test + added a few more to avoid using internals for testing.

Regarding render tests - I think there was some error during job execution, and because it didn't finish, there was no report. Unfortunately there was also no any other information about why it didn't succeed. Locally it runs with no issues, and the newer runs also went without problems.

src/render/render_to_texture.ts
pstaszek: Sorry, I didn't notice your second comment before, simplified a bit this logic.

CHANGELOG.md
```diff
@@ -18,7 +19,7 @@
 
 - Fix incorrect shader selection for text labels with inline images ([#6956](https://github.com/maplibre/maplibre-gl-js/pull/6956)) (by [@ciscorn](https://github.com/ciscorn))
 - Fix marker positions not updating when zoom or pitch are changed following a change in constraints ([#6925](https://github.com/maplibre/maplibre-gl-js/issues/6925)) (by [auspicus](https://github.com/auspicus))
-
+`
```
```suggestion

```

PR #7028, opened 2026-01-28
Add click event handler for Marker

### LINKED ISSUE #7018: Add click event handler for Marker
## User Story

As a developer I can add click handler to Marker easily, so that I don't need to define the event handler in the HTML element

## Rationale

Currently if I want to add a click event handler to Marker, I need to create the marker element and then add it to it, such as in this example: https://maplibre.org/maplibre-gl-js/docs/examples/add-custom-icons-with-markers/
I then need to take care of the cleanup as well.
However for drag events I can add the handler more easily easily using the `.on` function, such as in https://maplibre.org/maplibre-gl-js/docs/examples/create-a-draggable-marker/

## Impact

Easier usage of Markers

ganesh8068:
- Resolves #7018

### **Title: feat(marker): add support for click events on Marker**

**Description:** This PR adds support for listening to `click` events on the Marker class, similar to how drag events are handled. This allows developers to easily attach click listeners directly to the marker instance using` .on('click', fn)`.

**Changes:**

- Added `_onClick` listener to the Marker class.
- Added logic to fire a 'click' event when the marker element is clicked.
- Added a unit test to verify that the click event is fired correctly.

How to Test:

1. Run unit tests: `npm run test-unit `or` npx vitest run src/ui/marker.test.ts`

2. Verify that the "Marker fires click event when clicked" test passes.


src/ui/marker.ts
```diff
@@ -291,6 +293,7 @@ export class Marker extends Evented {
             // prevent focusing on click
             e.preventDefault();
         });
+        this._element.addEventListener('click', this._onClick);
```
HarelM: There's a need to remove event listening when removing the marker I believe.

ganesh8068: image
now it's ok??

ganesh8068:
Thanks for the feedback! I've updated the code to attach the listener in `addTo` and remove it in `remove` now check the code.

HarelM:
Please make sure to follow the template next time.
In this specific case a changelog entry is missing.

src/ui/marker.ts
```diff
@@ -339,6 +341,7 @@ export class Marker extends Evented {
         // would close once the event propagated to `map` due to the
         // `Popup._onClickClose` listener.
         this._map.on('click', this._onMapClick);
+        this._element.addEventListener('click', this._onClick);
```
HarelM: I would expect this event to be added "similar" to `dragstart` and `mousedown` in the constructor. Is there a reason why you added it here instead?

src/ui/marker.ts
```diff
@@ -370,6 +373,7 @@ export class Marker extends Evented {
             this._map.off('touchmove', this._onMove);
             delete this._map;
         }
+        this._element.removeEventListener('click', this._onClick);
```
HarelM: I'm guessing the other listeners should be removed as well, or not removed at all, this should be consistent with `dragstart` and `mousedown` I believe.

HarelM:
Thanks for looking into this.

src/ui/marker.ts
ganesh8068: Fixed — the click listener is now attached in the constructor, consistent with `dragstart` and `mousedown` .

src/ui/marker.ts
ganesh8068: Agreed. The click listener now follows the same lifecycle as other marker element listeners and no longer has special cleanup logic.

HarelM (approved):

HarelM:
I still tend to think there's a memory leak here, but if no one complains then I guess we are OK...
Thanks for contributing!

PR #7031, opened 2026-01-30
fix: release running task after error so more tasks can run

### LINKED ISSUE #6093: task_queue.run(...) doesn't handle errors so it gets stuck on currentlyRunning

**browser**: Chrome 137

### Steps to Trigger Behavior

 1. have a map with the GlobeControl
 2. Zoom in and out of the map while alternating projection between mercator and globe
 3. And error related to `getRayDirectionFromPixel` will be thrown because of the changing of projections
 4. The error will bubble up causing the task_queue to get stuck on `currentlyRunning`
 5. You can no longer interact with the map or other controls and each interaction a new error about "attempting to run(), but is already running" will be thrown

### Link to Demonstration

You can reproduce this on the following example https://maplibre.org/maplibre-gl-js/docs/examples/create-a-heatmap-layer-on-a-globe-with-terrain-elevation/

### Expected Behavior

No error and freeze of the map should happen because of changing projections while zoom in/out or any other issue

### Actual Behavior

On this case the first error happens because of changing projections. 

As it can be seen on the screenshot for the error "not implemented yet" this is cause by the map render, it will call the `globe_camera_helper` and `handleMapControlsRollpitchBearingZoom` from `vertical_perspective_camera_helper.ts` but actually the `getRayDirectionFromPixel` would be from the `mercator_transform.ts` throwing the "not implemented yet" error.

This cause the second issue that is what I am mainly reporting here. Because this error is not handle at `task_queue.run(...)` and it just bubbles up until the try catch at `map.triggerRepaint()` the internal `this._currentRunnin` doesn't change to `false`, which causes that any future `task_queue.run(...)` call throws "Attempting to run(),  but it is already running"

#### Video

https://github.com/user-attachments/assets/ad8d87af-3d8d-4476-ab52-1d8bda81879a

#### Screenshots of the errors
Image
Image

### Additional info

I suppose this issue is actually including 2 issues, but I am focusing on the second one regarding task_queue, as I believe this is similar to other issues like https://github.com/maplibre/maplibre-gl-js/issues/6011 

This errors like the one from changing projection should break the task_queue and cause the map to stop working.

UberMouse:
- Fixes #6093

When a task throws, the queue now catches per task, runs the remaining tasks, releases the running flag and rethrows the first error after the loop.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


src/util/task_queue.ts
```diff
@@ -45,14 +45,16 @@ export class TaskQueue {
         // on the next run, not the current run.
         this._queue = [];
 
-        for (const task of queue) {
-            if (task.cancelled) continue;
-            task.callback(timeStamp);
-            if (this._cleared) break;
+        try {
```
HarelM: Shouldn't the task execution be wrapped in try catch and not the entire for loop?

UberMouse: this ensures that exiting the function always releases the `_currentlyRunning` lock

HarelM: What will happen if I add two tasks - the first that throws and the second that should pass.
And then call the `run` method? Will the second task run or be "ignored"?

UberMouse: Don't know. Don't have any idea how this Task system works, my boss just complained about this error so I found the linked issue and did a PR for the solution they suggested and never bothered to PR. Hopefully someone who knows anything about this Task system can check it out

HarelM: I would recommend trying to take a deeper look then, otherwise I'll have a hard time merging this as is...

HarelM:
Any updates on this?

src/util/task_queue.ts
birkskyum: >  Will the second task run or be "ignored"?

It's ignored, which isn't great.

I modified it so it now catches per task instead, runs the rest, and rethrows the first error after the loop, so the map doesn't freeze and the error still shows up for debugging. I've added a unit test for the two-task case.

birkskyum:
> Any updates on this?

@HarelM I've picked up this ticket, because I kept hitting this map freeze while working on a bug. It's in a better state now.

birkskyum (approved):

HarelM:
Might be worth trying to understand what causes task to throw and prevent that or try-catch inside these task. 
Nevertheless, this is a good improvement.

PR #7040, opened 2026-02-01
fix: Incorrect text placement along line in 3d mode (#7039)

### LINKED ISSUE #7039: Incorrect text placement along line in 3d mode
Text rendering along lines breaks sometimes in 3d.

I root caused this to getDEMElevation being called with coordinates beyond the bounds the terrain tile being used, [which returns `0`](https://github.com/maplibre/maplibre-gl-js/blob/00f3dc2327390fae65ace6f00090925dd1a33ee3/src/render/terrain.ts#L162), this leads to text pointing downwards sometimes. 

**maplibre-gl-js version**: 5.17.0

**browser**: Chrome stable

### Steps to Trigger Behavior

 1. view https://openskimap.org/#15.61/47.152287/11.861808/-152.9/51
 2. see the text for Ahornbahn is not aligned to the line properly

### Link to Demonstration

https://openskimap.org/#15.61/47.152287/11.861808/-152.9/51

I used this for testing the fix:
https://gist.github.com/russellporter/a1c420a3eeaaa0f30ec07ff0a96d471d

### Expected Behavior

Image

### Actual Behavior

Image

russellporter:
Fixes #7039, text rendering along lines breaks sometimes in 3d.

I root caused this to getDEMElevation being called with coordinates beyond the bounds the terrain tile being used, which returns 0, this leads to text pointing downwards sometimes.

I modified that function to compute the correct tile and fetch the elevation from it.

I suspect there is a better solution than what I came up with. It seems awkward that we need to reconstitute the correct tile to use inside getDEMElevation.

Before:
Screenshot 2026-02-01 at 11 15 15

After:
Screenshot 2026-02-01 at 11 17 15

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -60,6 +60,7 @@ import type {Tile} from './tile/tile';
 import type {GeoJSONFeatureDiff, GeoJSONFeatureId, GeoJSONSourceDiff} from './source/geojson_source_diff';
 import type {QueryRenderedFeaturesOptions, QuerySourceFeatureOptions} from './source/query_features';
 import type {RequestTransformFunction, ResourceType} from './util/request_manager';
+import {EXTENT} from './data/extent';
 import type {OverscaledTileID} from './tile/tile_id';
 import type {PositionAnchor} from './ui/anchor';
 import type {ProjectionData} from './geo/projection/projection_data';
@@ -375,5 +376,6 @@ export {
     now,
     setNow,
     restoreNow,
-    isTimeFrozen
+    isTimeFrozen,
+    EXTENT
 };


HarelM:
So why not pass the "correct" tileID to this method then?

russellporter:
Thanks for taking a look so quick!

The problem is the code that uses this function operates in the coordinate system of a single tile, it doesn't have access to the tile itself afaict.

The tile gets captured here: https://github.com/maplibre/maplibre-gl-js/blob/00f3dc2327390fae65ace6f00090925dd1a33ee3/src/symbol/placement.ts#L239

and that gets passed around and ends up being called here:

https://github.com/maplibre/maplibre-gl-js/blob/00f3dc2327390fae65ace6f00090925dd1a33ee3/src/symbol/projection.ts#L178

It makes sense to me that the projection code doesn't need to juggle two sets of coordinates (tile and x/y) and that this is abstracted somehow.

One idea that comes to mind is moving this coordinates "normalization" into `_getTerrainElevationFunc` - at that point we could transform the x/y/tile params so they are properly passed to getElevation. I suppose the recursion hack I have now wouldn't be needed then.

HarelM:
Sounds better I think.

russellporter:
@HareIM on further inspection, i found the coordinates normalization is required by a few callsites so I extended terrain.ts with an additional function that can be used in these cases.

I also added a render test that proves the bug is fixed

Screenshot 2026-02-07 at 11 16 36

src/tile/tile_id.ts
```diff
@@ -242,6 +242,41 @@ function getQuadkey(z, x, y) {
     return quadkey;
 }
 
+export function normalizeTileCoordinates(
```
HarelM: Should this be part of the `OverscaledTileID`? like the other methods - wrap, scaleTo, etc.

HarelM: Also please add tsDocs comments on this method to better explain what it does - what does normalize means in this case...

src/tile/tile_id.ts
```diff
@@ -242,6 +242,41 @@ function getQuadkey(z, x, y) {
...
+        newWrap -= Math.ceil(-newCanonicalX / dim);
+        newCanonicalX = ((newCanonicalX % dim) + dim) % dim;
+    } else if (newCanonicalX >= dim) {
+        newWrap += Math.floor(newCanonicalX / dim);
+        newCanonicalX = newCanonicalX % dim;
+    }
+
+    return {
+        tileID: new OverscaledTileID(tileID.overscaledZ, newWrap, z, newCanonicalX, newCanonicalY),
+        x: newX,
```
HarelM: It is possible to only return a new `OverscaledTileID` without the extra x and y and use what's in the `OverscaledTileID` maybe?

HarelM:
The new solution is much better, thanks!

src/render/terrain.ts
```diff
@@ -180,6 +180,21 @@ export class Terrain {
...
+    /**
+     * Get elevation for coordinates that may be outside the tile bounds [0, EXTENT).
+     * Normalizes coordinates to the appropriate neighbor tile before lookup.
+     * @param tileID - the tile the coordinates are relative to
+     * @param x - x coordinate, may be outside [0, EXTENT)
+     * @param y - y coordinate, may be outside [0, EXTENT)
+     * @param extent - optional, default 8192
+     * @returns the elevation
+     */
+    getElevationCrossTile(tileID: OverscaledTileID, x: number, y: number, extent: number = EXTENT): number {
```
HarelM: When wouldn't we want to use this? I would expect to avoid adding a new method to the terrain calls and update `getElevation`, is it possible? Does it makes sense?

src/tile/tile_id.ts
HarelM: Ahh, the overscaltileid is only the tile coordinates, but there's a need to get the in-tile position as well, after the normalization...

src/render/terrain.ts
russellporter: That was my original solution. However, it seemed wrong that `getElevation` would need to accept invalid coordinates and then clean them up under the hood - coordinates should be already valid when they are being passed in - thats a responsibility of the caller. Ideally the code that produced the invalid coordinates should be fixed, although that seems tricky in this case, so this is a compromise.

What do you think?

src/tile/tile_id.ts
russellporter: > Ahh, the overscaltileid is only the tile coordinates, but there's a need to get the in-tile position as well, after the normalization...

👍

src/render/terrain.ts
HarelM: I don't think there should be two methods that does almost the same thing as part of the "piblic" API of terrain. I would recommend simply normalizing the coordinates when calling getElevation and remove this method. 
Your original solution involved recursion, this is not what I suggest here, simply move this normalize call to inside get elevation. 
The other solution is to do it outside this class, but then you need to hunt down all these places, and I don't think it's worth the effort...

russellporter: Understood, thank you. I've updated the impl accordingly

src/tile/tile_id.ts
russellporter: Addressed 👍

HarelM:
Just make sure to export EXTENT since this is not part of OverscaleTileID, which is somewhat "public".

PR #7056, opened 2026-02-04
fix: store encoding when caching raw tile data for MLT queryRenderedFeatures

dannote:
## Launch Checklist

`latestEncoding` is declared on `Tile` (line 75) and read in the `else` branch of `loadVectorData` (line 231), but never written when `rawTileData` first arrives. This causes `queryRenderedFeatures` to use the PBF parser for MLT-encoded tiles, throwing `Unimplemented type: 7`.

The fix adds `this.latestEncoding = data.encoding` alongside the existing `this.latestRawTileData = data.rawTileData` assignment.

**Repro:** set `encoding: "mlt"` on a vector source, call `map.queryRenderedFeatures()` → throws `Unimplemented type: 7`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@
 ### 🐞 Bug fixes
 
 - Fix incorrect popup location in case of terrain and `jumpTo` ([#7267](https://github.com/maplibre/maplibre-gl-js/issues/7267)) (by [@HarelM](https://github.com/HarelM))
-- _...Add new stuff here..._
 - Fix memory leak in VideoSource: remove `playing` event listener and pause video on source removal ([#7279](https://github.com/maplibre/maplibre-gl-js/pull/7279)) (by [@johanrd](https://github.com/johanrd))
 - Fix memory leak where typed array views retained StructArray buffers after GPU upload, preventing garbage collection ([#7280](https://github.com/maplibre/maplibre-gl-js/pull/7280)) (by [@johanrd](https://github.com/johanrd))
 - Fix raster DEM tiles getting stuck in `"reloading"` state ([#7277](https://github.com/maplibre/maplibre-gl-js/pull/7277)) (by [@katemihalikova](https://github.com/katemihalikova))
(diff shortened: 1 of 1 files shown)


HarelM:
While I don't think your fix is wrong, this is being set in featureIndex when the tile arrives from the worker, doesn't it? 
I tested QRF as part of mlt implementation and didn't see this issue. 
Can you add a test maybe to illustrate the issue, or share a simple reproduction?

Don't get me wrong, I think the fix here surely can't hurt and is probably needed, I'm just wondering how I missed it...

dannote:
@HarelM Here's my case: I have a click handler that shows feature info in a popup:

```js
map.on("click", (e) => {
  const features = map.queryRenderedFeatures(e.point);
  showPopup(features);
});
```

The source uses `encoding: "mlt"`. When I click on the map, I get `Unimplemented type: 7`.

Looking at the code, `this.latestEncoding` is never set when `rawTileData` first arrives — only `this.latestFeatureIndex.encoding` gets assigned. Later, the else-branch tries to read `this.latestEncoding` and it's undefined.

HarelM:
I'm using the following example to try and reproduce, but I can't:
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <title>Display a map with MLT</title>
    <meta property="og:description" content="Initialize a map in an HTML element with MapLibre GL JS. This example is using a style that uses MLT (MapLibre Tiles). The example is very similar to the regular one except the tiles are in a different encoding." />
    <meta charset='utf-8'>
    <style>
    </style>
</head>
<body>
<div id="map"></div>
<script>
        style: 'https://demotiles.maplibre.org/tiles-mlt/plain.json', // style URL
        zoom: 1, // starting zoom
        maplibreLogo: true
    });

    map.on('load', () => {
        map.on("click", (e) => {
            const features = map.queryRenderedFeatures(e.point);
            console.log(features);
        });
    });
</script>
</body>
</html>
```

Here's the console log for a random click I did:
image


What's your setup? Can you share a reproduction?

PR #7058, opened 2026-02-05
Simplify and abstract the GeoJSON Worker

wayofthefuture:
I’m sure this PR will likely be vetoed, but giving it a go anyway.

This PR disables inheritance between vector tile worker and geojson worker, allowing both to independently implement the workertile interface.

Following the geojson worker back to the main thread through the vector tile worker is a spaghetti nightmare.  If there was ever a case for composition against inheritance, this likely fits the bill, even though this refactor does not provide composition. 

Vector tile worker handles multiple functions unrelated to geojson, such as cache control, experimental overzooming, overzoomed tile cache, and request timing for individual tiles - all of which have no place in the geojson worker. Furthermore, the re-load tile logic is segmented, as the behavior differs and refers back to the inherited class if the tile is already loaded.

Both files contain about a combined 750 lines of code but are only sharing about 50 lines. 

For all of the above reasons I think this makes geojson source worker overly complex - and should be made independent/abstracted into its own class that implements the workertile interface separately - in my opinion.


src/source/geojson_worker_source.ts
```diff
@@ -60,7 +66,14 @@ type GeoJSONIndex = GeoJSONVT | Supercluster;
  * `new GeoJSONWorkerSource(actor, layerIndex, customLoadGeoJSONFunction)`.
  * For a full example, see [mapbox-gl-topojson](https://github.com/developmentseed/mapbox-gl-topojson).
  */
-export class GeoJSONWorkerSource extends VectorTileWorkerSource {
+export class GeoJSONWorkerSource implements WorkerSource {
+    actor: IActor;
+    layerIndex: StyleLayerIndex;
+    availableImages: Array<string>;
+    fetching: {[_: string]: FetchingState };
```
HarelM: Maybe  you can move the tile loading logic to a class both worker sources can use?

src/source/geojson_worker_source.ts
```diff
@@ -97,6 +117,126 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
...
+    /**
+     * Implements {@link WorkerSource.reloadTile}.
+     *
+     * If the tile is loaded, reload by re-parsing the already available tile data.
+     * Otherwise, such as after a setData() call, we load the tile fresh.
+     *
+     * @param params - the parameters
+     * @returns A promise that resolves when the tile is reloaded
+     */
+    reloadTile(params: WorkerTileParameters): Promise<WorkerTileResult> {
```
HarelM: Can you move this down to reduce diff changes? I see it in the diff as removed and added.

src/source/geojson_worker_source.ts
```diff
@@ -152,13 +292,13 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
...
         const resourceTimingData = perf.finish();
+        if (!resourceTimingData) return;
+
         // it's necessary to eval the result of getEntriesByName() here via parse/stringify
         // late evaluation in the main thread causes TypeError: illegal invocation
-        if (resourceTimingData) {
-            result.resourceTiming = {};
-            result.resourceTiming[params.source] = JSON.parse(JSON.stringify(resourceTimingData));
-        }
+        result.resourceTiming = {[params.source]: JSON.parse(JSON.stringify(resourceTimingData))};
```
HarelM: Why was this changed?

src/source/geojson_worker_source.ts
```diff
@@ -97,6 +117,126 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
...
+        if (workerTile.status === 'parsing') {
+            const result = await workerTile.parse(workerTile.vectorTile, this.layerIndex, this.availableImages, this.actor, params.subdivisionGranularity);
+
+            // If we have canceled the original parse, make sure to pass the rawData from the original fetch.
+            if (this.fetching[uid]) {
+                const {rawData} = this.fetching[uid];
+                delete this.fetching[uid];
+
+                // Transferring a copy of rawData because the worker needs to retain its copy.
+                return extend({rawTileData: rawData.slice(0), encoding: params.encoding}, result);
```
HarelM: Does the worker here needs to retain a copy still (maybe for reload...)?
Also I would expect encoding here to be "hardcoded" as opposed to the vector source.

HarelM:
Bastard the coverage part, I find this PR a good way to avoid problem with geojson and vector source as they are different enough...
This is a good call!

src/source/geojson_worker_source.ts
wayofthefuture: always de-nesting

src/source/geojson_worker_source.ts
wayofthefuture: I don't know... it's odd. But I wanted to retain the same behavior. I also don't know why we would re-parse here versus storing the promise and awaiting the in progress parse. Would be nice if we could figure out how to get this raw data completely out however

HarelM:
Keep in mind that I would like the following PR to be merged before this one, so you might need to resolve some conflicts, sorry...
- https://github.com/maplibre/maplibre-gl-js/pull/6972

wayofthefuture:
> Keep in mind that I would like the following PR to be merged before this one, so you might need to resolve some conflicts, sorry...

Yup that's a big ouch but a perfect example of why geojson worker needs to be separate.

wayofthefuture:
> Bastard the coverage part, I find this PR a good way to avoid problem with geojson and vector source...

Coverage looking better now

wayofthefuture:
@HarelM Well I hope I didn't get too carried away... between the composition class and parse helper I think this is really turning out quite nice. Since request timing isn't needed for loadTile in the geojson worker it really cleaned that up too. Not looking forward to eTag!

src/source/geojson_worker_source.ts
```diff
@@ -73,28 +81,102 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
...
+
+            const parseState = {rawData};
+            this.tileState.setParsing(uid, parseState);  // Keep data so reloadTile can access if parse is canceled.
+            try {
+                return await this._parseWorkerTile(workerTile, params, parseState);
+            } finally {
+                this.tileState.clearParsing(uid);
+            }
+        } catch (err) {
+            workerTile.status = 'done';
```
HarelM, line 130: Really? In case of an error the status should be `done` and not `error`?

src/source/geojson_worker_source.ts
```diff
@@ -73,28 +81,102 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
...
+    }
 
-        const geoJSONTile = this._geoJSONIndex.getTile(canonical.z, canonical.x, canonical.y);
-        if (!geoJSONTile) {
-            return null;
+    private async _reloadLoadedTile(params: WorkerTileParameters): Promise<WorkerTileResult> {
+        const uid = params.uid;
+
+        const workerTile = this.tileState.getLoaded(uid);
+        if (!workerTile) throw new Error('Should not be trying to reload a tile that was never loaded or has been removed');
```
HarelM, line 140: Maybe add some information on which tile this is for better debugability?

src/source/geojson_worker_source.ts
```diff
@@ -73,28 +81,102 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
...
         this._createGeoJSONIndex = createGeoJSONIndexFunc;
     }
 
     /**
      * Retrieves and sends loaded vector tiles to the main thread.
      */
-    override async loadVectorTile(params: WorkerTileParameters, _abortController: AbortController): Promise<LoadVectorTileResult | null> {
-        const canonical = params.tileID.canonical;
+    loadVectorTile(params: WorkerTileParameters): LoadVectorTileResult | null {
+        if (!this._geoJSONIndex) throw new Error('Unable to parse the data into a cluster or geojson');
```
HarelM, line 95: Maybe add some data to the error to indicate which tile this is for better debugability?

src/source/geojson_worker_source.ts
```diff
@@ -73,28 +81,102 @@ export class GeoJSONWorkerSource extends VectorTileWorkerSource {
...
+            const parseState = this.tileState.consumeParsing(uid);
+            return await this._parseWorkerTile(workerTile, params, parseState);
         }
 
-        const geojsonWrapper = new GeoJSONWrapper(geoJSONTile.features, {version: 2, extent: EXTENT});
+        // If there was no vector tile data on the initial load, don't try and reparse the tile.
+        if (workerTile.status === 'done' && workerTile.vectorTile) {
+            return await this._parseWorkerTile(workerTile, params);
+        }
+    }
```
HarelM, line 154: Am I miss-reading this or there's a code flow that doesn't return anything?

src/source/geojson_worker_source.ts
HarelM: Well, I now saw that this is how it is in vector source, so I guess this code hasn't changed...

src/source/geojson_worker_source.ts
HarelM: Well, I now saw that this is how it is in vector source, so I guess this code hasn't changed...

HarelM:
The code is now structured a lot better than before. Thanks for this.
I'll be merging this as I don't see any changed to the original behavior, so there's no reason not to merge this.
If there are still things you would like to change feel free to add them in a later PR.

HarelM (approved):

HarelM:
The only thing I found "repeated" is the reload tile method, but I think that's a small price to pay.

PR #7070, opened 2026-02-06
Feat/geolocate element

ganesh8068:
**Title**: feat: Support `<geolocation>` HTML element in **GeolocateControl**

### Description:

This PR introduces support for the experimental `<geolocation>` HTML element within the **GeolocateControl**. This enhancement aligns with the goal of verifying new features and improving user experience on supported browsers.

### Changes

- New Feature: Added `useGeolocationElement` option to **GeolocateControlOptions**.

  - When enabled, it renders the native `<geolocation>` element if supported by the browser.
  - Falls back to the standard button implementation if unsupported.
 
**Example**: Added `docs/assets/examples/geolocate-html-element.html` to demonstrate the new functionality, as recommended for new features. 

**Tests**: Added comprehensive unit tests in `src/ui/control/geolocate_control.test.ts` to ensure no regressions and verify the new logic (element creation, event handling, fallback).

**Changelog**: Updated `CHANGELOG.md` with a new entry under **Features and improvements**.


HarelM:
Can you attach a video showcasing how this would look? 
If the UX of this control is not in-par with the current control I would consider moving this to a plugin.

ganesh8068:
Here is a video demonstrating the `<geolocation>` element in action.


image
image



**Regarding the UX:** The control uses the native browser UI (the `<geolocation>` element). This is by design to improve user trust and ensure higher permission success rates, as detailed in the [Chrome Developers Blog](https://developer.chrome.com/blog/geolocation-element).

- **UX Consistency:** It purposely looks different (native browser style) to signal that it is a trusted browser action, distinct from the custom MapLibre controls.

- **Fallback:** It gracefully falls back to the standard MapLibre button design on browsers that don't support it, so the "standard" UX is preserved where the new API isn't available.

- **Plugin vs Core:** Since this is a standard HTML element that might become common, having it as a built-in option (`useGeolocationElement: true`) seems appropriate for the core library, similar to how we support standard HTML buttons. Moving it to a plugin might ensure strict visual consistency but would add friction for users wanting to adopt this new web standard.

HarelM:
Thanks for the details, but this control is just bad UX compared to the correct control. 
Looks like the file input browsers have which also look hideous. 
I prefer to have that as a plugin and not in the core if someone would like to use it. Sorry...

ganesh8068:
Understood. I appreciate the feedback regarding the UX consistency. I'll close this PR and consider publishing this as a separate plugin for users who prefer the native browser behavior. Thanks for reviewing!

HarelM:
Thanks for taking the time to raise this! I hope you'll find other places you can contribute, sorry for the cold shoulder...

PR #7073, opened 2026-02-07
refactor!: `UrlSearchParams` based `Hash`

### LINKED ISSUE #7006: Refactor `Hash` to use `URLSearchParams`
`Hash` currently does a lot of manual [string building](https://github.com/maplibre/maplibre-gl-js/blob/b8f34428adc052583c7197daaf11b5e6deb54343/src/ui/hash.ts#L70-L80) and [parsing](https://github.com/maplibre/maplibre-gl-js/blob/b8f34428adc052583c7197daaf11b5e6deb54343/src/ui/hash.ts#L134-L151). This makes it tricky for clients to extend the class to support, say, other hash parameters that might need to coexist with the ones managed by MapLibre. `URLSearchParams` would simplify the class a bit. It should be supported by all the browsers we care about.

CommanderStorm:
## Launch Checklist

 - [x] Link to related issues.

This PR refactors `Hash` to be based on `UrlSearchParams`.
Tests were pulled to another PR.

CC @1ec5

Resolves https://github.com/maplibre/maplibre-gl-js/issues/7006


src/ui/hash.test.ts
```diff
@@ -358,7 +395,19 @@ describe('hash', () => {
 
             expect(hash._isValidHash(hash._getCurrentHash())).toBeFalsy();
         });
+        
+        test('invalidate hash, only one values', () => {
+            window.location.hash = '#24';
 
+            expect(hash._isValidHash(hash._getCurrentHash())).toBeFalsy();
+        });
```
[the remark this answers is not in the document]
CommanderStorm:
```suggestion
        test('invalidate hash, only one value', () => {
            window.location.hash = '#24';

            expect(hash._isValidHash(hash._getCurrentHash())).toBeFalsy();
        });
```

src/ui/hash.ts
```diff
@@ -65,41 +65,32 @@ export class Hash {
...
-                return part;
-            }).filter(a => a);
-            if (!found) {
-                parts.push(`${hashName}=${hash}`);
-            }
-            return `#${parts.join('&')}`;
+            const params = this._getHashParams();
+            params.set(this._hashName, hash);
+            // Manually build the string to avoid URL encoding the hash value
+            return `#${this._buildHashString(params)}`;
```
1ec5: If the issue is only that `/` gets escaped despite being legal in the URL fragment, would replacing `%2F` have the desired effect?

```js
return params.toString().replaceAll('%2F', '/');
```

Otherwise, unescaping the string should be equivalent, avoiding the need to build a string manually:

```js
return decodeURIComponent(params.toString());
```

CommanderStorm: fair

src/ui/hash.ts
```diff
@@ -121,42 +105,35 @@ export class Hash {
     };
 
     _updateHashUnthrottled = () => {
-        // Replace if already present, else append the updated hash string
-        const location = window.location.href.replace(/(#.*)?$/, this.getHashString());
-        window.history.replaceState(window.history.state, null, location);
+        window.location.hash = this.getHashString();
```
HarelM: I'm not sure this is equivalent to the previous code - this updates the hash and then replaces the state, which might add a "new" entry to the hash's history...?

src/ui/hash.test.ts
```diff
@@ -276,7 +313,7 @@ describe('hash', () => {
 
         map.setCenter([2.0, 1.0]);
 
-        expect(window.location.hash).toBe('#baz&map=7/1/2/135/60&foo=bar');
+        expect(window.location.hash).toBe('#baz=&map=7/1/2/135/60&foo=bar');
```
HarelM: Why was this changed? Is it possible that this might be a "breaking change"?

src/ui/hash.test.ts
```diff
@@ -462,4 +513,204 @@ describe('hash', () => {
 
         expect(map).toBeTruthy();
     });
+
+    test('hash with special characters in other parameters', () => {
+        const hash = createHash('map')
+            .addTo(map);
+
+        // Set up hash with URL in another parameter
+        window.location.hash = '#map=10/3/-1&returnUrl=https://example.com&filter=a&b=';
```
HarelM, line 532: What happens if you have `https://example.com/&filter=a&b=` instead (note the addition of "/" after .com)?

src/ui/hash.test.ts
```diff
@@ -462,4 +513,204 @@ describe('hash', () => {
...
+
+        // Should decode and parse correctly
+        expect(map.getZoom()).toBe(10);
+        expect(map.getCenter().lat).toBe(3);
+        expect(map.getCenter().lng).toBe(-1);
+
+        // Update and verify encoding is preserved where needed
+        map.setZoom(8);
+
+        expect(window.location.hash).toBe('#map=8/3/-1&foo=bar');
```
HarelM: This doesn't look like it preserves the "%2F", am I missing anything?

src/ui/hash.test.ts
```diff
@@ -462,4 +513,204 @@ describe('hash', () => {
...
+        expect(window.location.hash).toContain('foo=bar');
+
+        // Double ampersand
+        window.location.hash = '#map=10/3/-1&&foo=bar';
+        hash._onHashChange();
+        map.setZoom(12);
+        expect(window.location.hash).toContain('map=12/3/-1');
+        expect(window.location.hash).toContain('foo=bar');
+
+        // Leading ampersand
```
HarelM: Split test and have this comment as the test name.

src/ui/hash.test.ts
```diff
@@ -462,4 +513,204 @@ describe('hash', () => {
...
+        map.setZoom(13);
+        expect(window.location.hash).toContain('map=13/3/-1');
+        expect(window.location.hash).toContain('foo=bar');
+    });
+
+    test('hash with empty parameter values', () => {
+        const hash = createHash('map')
+            .addTo(map);
+
+        // Empty map value should be invalid
```
HarelM: I would consider removing all these comments inside the test, tests should be readable enough without these comments.

src/ui/hash.test.ts
```diff
@@ -462,4 +513,204 @@ describe('hash', () => {
...
+        expect(Math.abs(map.getCenter().lng)).toBeCloseTo(180, 2);
+
+        // Near north pole, positive dateline
+        window.location.hash = '#10/85.05/180';
+        hash._onHashChange();
+        expect(map.getZoom()).toBe(10);
+        expect(map.getCenter().lat).toBeCloseTo(85.05, 1);
+        expect(map.getCenter().lng).toBeCloseTo(180, 2);
+
+        // Bearing at exact ±180° boundary
```
HarelM: Split test...

src/ui/hash.ts
```diff
@@ -121,42 +105,35 @@ export class Hash {
...
+        window.history.replaceState(window.history.state, null, window.location.href);
```
CommanderStorm:
```suggestion
        const location = window.location.href.replace(/(#.*)?$/, this.getHashString());
        window.history.replaceState(window.history.state, null, location);
```

src/ui/hash.test.ts
CommanderStorm: No, URL encoding is getting removed

src/ui/hash.test.ts
CommanderStorm: I have no clue.
This change is the result of simplifying `_buildHashString` to `decodeURIComponent(params.toString())`
https://github.com/maplibre/maplibre-gl-js/pull/7073/changes/578887c634715eb49f2bcc8bad205041381c0017..615b59c48a350538bdfeb4f76ccbe3a1900bd617

CommanderStorm:
Lets split adding tests and changing behaviour, this way the actual changes are more clear

HarelM:
Sure, no problem.

src/ui/hash.test.ts
1ec5: I think it makes sense to continue to unescape `%2F`, since it gets annoying in the usual `map=` parameter. Slashes in the URL fragment don’t need to be escaped anyhow. However, other encoding might be useful. The class is written generically enough that a parameter could end up with freeform text containing a `&` that needs to be escaped. My other suggestion in https://github.com/maplibre/maplibre-gl-js/pull/7073#discussion_r2778218629 would’ve solved this problem, prettifying `%2F` as `/` but leaving all the other escaping in place.

1ec5: The other character to watch out for is the space, which `URLSearchParams` encodes as `+` rather than `%20`.

src/ui/hash.test.ts
1ec5: Apparently `foo=` is the canonical form of `foo`. `URLSearchParams` [doesn’t distinguish](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams#empty_value_vs._no_value). I’m pretty sure this is consistent with how most software interprets query strings, but here we’re reusing query strings as fragments. Technically we get to decide the correct behavior of flag parameters. As long as we use `URLSearchParams` for both parsing and reconstructing the fragment, it shouldn’t cause any problems but could trip up tests like this one.

src/ui/hash.ts
```diff
@@ -121,42 +107,35 @@ export class Hash {
...
+                params.delete(keys[0]);
+            }
         }
-        let location = window.location.href.replace(/(#.+)?$/, replaceString);
-        location = location.replace('&&', '&');
-        window.history.replaceState(window.history.state, null, location);
+
+        const newHash = decodeURIComponent(params.toString());
+        window.location.hash = newHash ? `#${newHash}` : '';
+        window.history.replaceState(window.history.state, null, window.location.href);
```
CommanderStorm:
```suggestion
        const newHash = decodeURIComponent(params.toString());
        const replaceString = newHash ? `#${newHash}` : ''
        let location = window.location.href.replace(/(#.+)?$/, replaceString);
        location = location.replace('&&', '&');
        window.history.replaceState(window.history.state, null, location);
```

src/ui/hash.ts
```diff
@@ -65,40 +65,26 @@ export class Hash {
...
-                }
-                return part;
-            }).filter(a => a);
-            if (!found) {
-                parts.push(`${hashName}=${hash}`);
-            }
-            return `#${parts.join('&')}`;
+            const params = this._getHashParams();
+            params.set(this._hashName, hash);
+            return `#${decodeURIComponent(params.toString())}`;
```
1ec5: If we don’t like the trailing `=`, we could do another find and replace, your call:

```suggestion
            return `#${decodeURIComponent(params.toString()).replace(/=&|=$/g, '')}`;
```

src/ui/hash.ts
```diff
@@ -65,40 +65,26 @@ export class Hash {
...
-                if (part[0] === this._hashName) {
-                    keyval = part;
-                }
-            });
-            return (keyval ? keyval[1] || '' : '').split('/');
+            return (params.get(this._hashName) || '').split('/');
         }
+        // For unnamed hashes, get the first key
+        const keys = Array.from(params.keys());
+        const hash = keys.length > 0 ? keys[0] : '';
```
1ec5: Nit: In case you feel like playing a round of code golf with me:

```suggestion
        const hash = [...params.keys()][0] ?? '';
```

1ec5 (approved):

src/ui/hash.test.ts
```diff
@@ -335,7 +335,7 @@ describe('hash', () => {
 
             hash._removeHash();
 
-            expect(window.location.hash).toBe('#baz&foo=bar');
+            expect(window.location.hash).toBe('#baz=&foo=bar');
```
CommanderStorm:
```suggestion
            expect(window.location.hash).toBe('#baz&foo=bar');
```

src/ui/hash.test.ts
```diff
@@ -273,7 +273,7 @@ describe('hash', () => {
 
         map.setCenter([2.0, 1.0]);
 
-        expect(window.location.hash).toBe('#baz&map=7/1/2/135/60&foo=bar');
+        expect(window.location.hash).toBe('#baz=&map=7/1/2/135/60&foo=bar');
```
CommanderStorm:
```suggestion
        expect(window.location.hash).toBe('#baz&map=7/1/2/135/60&foo=bar');
```

src/ui/hash.ts
```diff
@@ -121,30 +106,25 @@ export class Hash {
...
+            params.delete(this._hashName);
+        } else {
+            // For unnamed hash (#zoom/lat/lng&other=params), remove first entry
+            const keys = Array.from(params.keys());
+            if (keys.length > 0) {
+                params.delete(keys[0]);
+            }
         }
+
+        const newHash = decodeURIComponent(params.toString());
```
CommanderStorm:
```suggestion
        const newHash = decodeURIComponent(params.toString()).replace(/=&|=$/g, '');
```

HarelM:
I'm a bit lost with all these commits and review comments, please ping me when a new review is required.

CommanderStorm:
@HarelM the PR is done, sorry for the messy history, not my best work.

So basically, this PR boils down to "how should this be normalized and should this be normalized?"
- if yes, then the previous version
- if no, then the current version
- if shmaybe (neighter one), then the previous version. See tests what I mean

We can aso put this into v6

HarelM:
I think using urlsearchpatams is important, but this is a subtle breaking change. 
Let's add that to the list of v6 breaking changes. 
I want to push v6 forward, we have enough reasons to release a breaking change version and we haven't released one in over a year, so it's time.

HarelM:
I've added it to the list.

1ec5:
We can easily avoid the breaking change if desired with a one-liner. That was the question in https://github.com/maplibre/maplibre-gl-js/pull/7073#issuecomment-3881537974.

HarelM:
I might have miss read it then, avoiding a breaking change would be better, we can introduce the breaking change later if it's only a single line change.

CommanderStorm:
I don't think this can be done without a subtle behaviour change if we want to use `UrlSearchParams`, that is what I meant by previous version.

src/ui/hash.test.ts
```diff
@@ -383,10 +383,10 @@ describe('hash', () => {
             expect(hash._isValidHash(hash._getCurrentHash())).toBeTruthy();
         });
 
-        test('invalidate hash with slashes encoded as %2F', () => {
+        test('validate hash with slashes encoded as %2F', () => {
             window.location.hash = '#10%2F3.00%2F-1.00';
 
-            expect(hash._isValidHash(hash._getCurrentHash())).toBeFalsy();
+            expect(hash._isValidHash(hash._getCurrentHash())).toBeTruthy();
```
1ec5, line 389: I’d characterize this as a routine bug fix rather than a backwards-incompatible change. I don’t think anyone would be expecting the `%2F` to persist, and we continue to recognize it anyways.

src/ui/hash.ts
```diff
@@ -65,40 +65,25 @@ export class Hash {
...
+            const params = this._getHashParams();
+            params.set(this._hashName, hash);
+            return `#${decodeURIComponent(params.toString()).replace(/=&/g, '&').replace(/=$/g, '')}`;
         }
 
         return `#${hash}`;
     }
 
+    _getHashParams = () => {
+        return new URLSearchParams(window.location.hash.replace('#', ''));
```
1ec5, line 77: If `hash` is set, its first character is guaranteed to be `#`. The previous code only removed the leading `#`, whereas this removes any occurrence.

src/ui/hash.test.ts
```diff
@@ -610,7 +610,7 @@ describe('hash', () => {
         expect(map.getZoom()).toBe(10);
 
         map.setZoom(5);
-        expect(window.location.hash).toBe('#map=5/3/-1&empty=');
+        expect(window.location.hash).toBe('#map=5/3/-1&empty');
```
1ec5, line 613: whatwg/url#427 led me down quite a rabbit hole of discussions…

As I mentioned earlier, I think we could have leeway to make this change. The fact that our hashes generally conform to the query string syntax is just a convention on our part. I also don’t think we necessarily need to hold the hash behavior to the same standard as a formal public API, since `Hash` isn’t publicly exposed by any APIs yet.

That said, if this is a source of concern nonetheless, anything calling `_getHashParams()` would need to separately look for flags:

```js
const hash = window.location.hash.slice(1);
const params = new URLSearchParams(hash);
const flags = [...hash.matchAll(/(?<=^|&)(\w+)(?=&|$)/g).map(p => p[1])];
// …
const result = [...params.entries().map(([k, v]) => flags.includes(k) ? k : `${k}=${v}`)].join('&');
return `#${result}`;
```

Kind of cryptic…

src/ui/hash.test.ts
HarelM: A case where this is needed for example is if you have a link to a file or location as part of the query parameters, then you can't use "/" but you have to use the encoded way. I'm not sure this answers any question, but I wanted to give a use case.
If this is kept the same and doesn't break, that's great.

src/ui/hash.test.ts
HarelM: The question is who is relaying on this functionality to work this way, is somewhat a concern I have. The methods in the hash my not be public, setting it as part on map initialization is public and reading the address bar value is public as well.
I'm just saying that waiting a bit longer for version 6 might be safer for this.
It might be regarded as a bug fix though if we want to cut ourselves some slack 😀

HarelM:
I'm a bit lost, where are we with this PR? Waiting for version 6?

CommanderStorm:
Yes, I think that is the best bet.

HarelM:
@CommanderStorm can you add a changelog with a "breaking change" triangle icon?
I'll merge this once you add the changelog entry.

CommanderStorm:
done

HarelM (approved):

PR #7074, opened 2026-02-07
Add Etag support to the coming vector/geojson worker refactor.

wayofthefuture:
This is a PR against the current PR #7058 that creates more separation between vector and geojson workers. Putting up this draft PR so we can visualize the differences (and show patch coverage) between the new refactored workers and the coming etag support, which is modifying many of the same files. Feel free to comment. Am trying to figure out the best way to abstract the vector/geojson workers from each other while maintaining the coming etag support in a clean way.

The etag support in this PR tries to leverage the existing workertile result pipeline, and adds a LoadTileResult type to loadTile - which carries an etagUnmodified property. I figured this result type would be beneficial as it allows further properties to be added in the future to loadTile. In order to separate downloading from tile creation (for etag), the network request and creation of the vector tile data are separated so that when the etag hasn't changed the processing does not occur. Coincidentally now both loadVectorTile's in both workers are synchronous and separate tile processing from any network requests.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -102,7 +102,7 @@
         "postcss-cli": "^11.0.1",
         "postcss-inline-svg": "^6.0.0",
         "pretty-bytes": "^7.1.0",
-        "puppeteer": "^24.37.3",
+        "puppeteer": "^24.37.5",
         "react": "^19.2.4",
         "react-dom": "^19.2.4",
         "regenerate": "^1.4.2",
@@ -2970,9 +2970,9 @@
       "license": "MIT"
     },
     "node_modules/@puppeteer/browsers": {
-      "version": "2.12.1",
-      "resolved": "https://registry.npmjs.org/@puppeteer/browsers/-/browsers-2.12.1.tgz",
-      "integrity": "sha512-fXa6uXLxfslBlus3MEpW8S6S9fe5RwmAE5Gd8u3krqOwnkZJV3/lQJiY3LaFdTctLLqJtyMgEUGkbDnRNf6vbQ==",
+      "version": "2.13.0",
+      "resolved": "https://registry.npmjs.org/@puppeteer/browsers/-/browsers-2.13.0.tgz",
+      "integrity": "sha512-46BZJYJjc/WwmKjsvDFykHtXrtomsCIrwYQPOP7VfMJoZY2bsDF9oROBABR3paDjDcmkUye1Pb1BqdcdiipaWA==",
       "dev": true,
       "license": "Apache-2.0",
       "dependencies": {
@@ -5820,9 +5820,9 @@
       }
     },
     "node_modules/b4a": {
-      "version": "1.7.4",
-      "resolved": "https://registry.npmjs.org/b4a/-/b4a-1.7.4.tgz",
-      "integrity": "sha512-u20zJLDaSWpxaZ+zaAkEIB2dZZ1o+DF4T/MRbmsvGp9nletHOyiai19OzX1fF8xUBYsO1bPXxODvcd0978pnug==",
+      "version": "1.8.0",
+      "resolved": "https://registry.npmjs.org/b4a/-/b4a-1.8.0.tgz",
+      "integrity": "sha512-qRuSmNSkGQaHwNbM7J78Wwy+ghLEYF1zNrSeMxj4Kgw6y33O3mXcQ6Ie9fRvfU/YnxWkOchPXbaLb73TkIsfdg==",
       "dev": true,
       "license": "Apache-2.0",
       "peerDependencies": {
@@ -5905,14 +5905,15 @@
       }
     },
     "node_modules/bare-stream": {
-      "version": "2.7.0",
-      "resolved": "https://registry.npmjs.org/bare-stream/-/bare-stream-2.7.0.tgz",
-      "integrity": "sha512-oyXQNicV1y8nc2aKffH+BUHFRXmx6VrPzlnaEvMhram0nPBrKcEdcyBg5r08D0i8VxngHFAiVyn1QKXpSG0B8A==",
+      "version": "2.8.0",
+      "resolved": "https://registry.npmjs.org/bare-stream/-/bare-stream-2.8.0.tgz",
+      "integrity": "sha512-reUN0M2sHRqCdG4lUK3Fw8w98eeUIZHL5c3H7Mbhk2yVBL+oofgaIp0ieLfD5QXwPCypBpmEEKU2WZKzbAk8GA==",
       "dev": true,
       "license": "Apache-2.0",
       "optional": true,
       "dependencies": {
-        "streamx": "^2.21.0"
+        "streamx": "^2.21.0",
+        "teex": "^1.0.1"
       },
       "peerDependencies": {
         "bare-buffer": "*",
@@ -13356,18 +13357,18 @@
       }
     },
     "node_modules/puppeteer": {
-      "version": "24.37.3",
-      "resolved": "https://registry.npmjs.org/puppeteer/-/puppeteer-24.37.3.tgz",
-      "integrity": "sha512-AUGGWq0BhPM+IOS2U9A+ZREH3HDFkV1Y5HERYGDg5cbGXjoGsTCT7/A6VZRfNU0UJJdCclyEimZICkZW6pqJyw==",
+      "version": "24.37.5",
+      "resolved": "https://registry.npmjs.org/puppeteer/-/puppeteer-24.37.5.tgz",
+      "integrity": "sha512-3PAOIQLceyEmn1Fi76GkGO2EVxztv5OtdlB1m8hMUZL3f8KDHnlvXbvCXv+Ls7KzF1R0KdKBqLuT/Hhrok12hQ==",
       "dev": true,
       "hasInstallScript": true,
       "license": "Apache-2.0",
       "dependencies": {
-        "@puppeteer/browsers": "2.12.1",
+        "@puppeteer/browsers": "2.13.0",
         "chromium-bidi": "14.0.0",
         "cosmiconfig": "^9.0.0",
         "devtools-protocol": "0.0.1566079",
-        "puppeteer-core": "24.37.3",
+        "puppeteer-core": "24.37.5",
         "typed-query-selector": "^2.12.0"
       },
       "bin": {
@@ -13378,13 +13379,13 @@
       }
     },
     "node_modules/puppeteer-core": {
-      "version": "24.37.3",
-      "resolved": "https://registry.npmjs.org/puppeteer-core/-/puppeteer-core-24.37.3.tgz",
-      "integrity": "sha512-fokQ8gv+hNgsRWqVuP5rUjGp+wzV5aMTP3fcm8ekNabmLGlJdFHas1OdMscAH9Gzq4Qcf7cfI/Pe6wEcAqQhqg==",
+      "version": "24.37.5",
+      "resolved": "https://registry.npmjs.org/puppeteer-core/-/puppeteer-core-24.37.5.tgz",
+      "integrity": "sha512-ybL7iE78YPN4T6J+sPLO7r0lSByp/0NN6PvfBEql219cOnttoTFzCWKiBOjstXSqi/OKpwae623DWAsL7cn2MQ==",
       "dev": true,
       "license": "Apache-2.0",
       "dependencies": {
-        "@puppeteer/browsers": "2.12.1",
+        "@puppeteer/browsers": "2.13.0",
         "chromium-bidi": "14.0.0",
         "debug": "^4.4.3",
         "devtools-protocol": "0.0.1566079",
@@ -15183,6 +15184,17 @@
       "dependencies": {
         "base64-js": "^1.3.1",
         "ieee754": "^1.1.13"
+      }
+    },
+    "node_modules/teex": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/teex/-/teex-1.0.1.tgz",
+      "integrity": "sha512-eYE6iEI62Ni1H8oIa7KlDU6uQBtqr4Eajni3wX7rpfXD8ysFx8z0+dri+KWEPWpBsxXfxu58x/0jvTVT1ekOSg==",
+      "dev": true,
+      "license": "MIT",
+      "optional": true,
+      "dependencies": {
+        "streamx": "^2.12.5"
       }
     },
     "node_modules/terser": {
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
     "postcss-cli": "^11.0.1",
     "postcss-inline-svg": "^6.0.0",
     "pretty-bytes": "^7.1.0",
-    "puppeteer": "^24.37.3",
+    "puppeteer": "^24.37.5",
     "react": "^19.2.4",
     "react-dom": "^19.2.4",
     "regenerate": "^1.4.2",


src/util/ajax.ts
```diff
@@ -11,7 +11,7 @@ export const GLOBAL_DISPATCHER_ID = 'global-dispatcher';
 /**
  * A type used to store the tile's expiration date and cache control definition
  */
-export type ExpiryData = {cacheControl?: string | null; expires?: Date | string | null};
+export type ExpiryData = {cacheControl?: string | null; expires?: Date | string | null; etag?: string; etagUnmodified?: boolean};
```
HarelM: The rest of the PR is straightforward, but this seems a bit weird to me - the addition of `etagUnmodified` to `ExpiryData`, can it be avoided somehow?

wayofthefuture: yes, done

HarelM:
@wayofthefuture I'll let you handle this merge, I tried it myself but was not 100% sure...

wayofthefuture:
All done!

HarelM:
@rivkamatan can you take a look, better yet if you could test it against your use case? 
If so I can merge this and release a new version.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 - Add support for click events on Marker ([#7028](https://github.com/maplibre/maplibre-gl-js/pull/7028)) (by [@ganesh8068](https://github.com/ganesh8068))
 - Simplify and abstract the GeoJSON Worker ([#7058](https://github.com/maplibre/maplibre-gl-js/pull/7058)) (by [@wayofthefuture](https://github.com/wayofthefuture))
 - Add `pseudo` option to `FullscreenControl` to force CSS-based fullscreen mode instead of native fullscreen API. Users might want this because this is faster on some devices ([#7076](https://github.com/maplibre/maplibre-gl-js/pull/7076)) (by [@CommanderStorm](https://github.com/CommanderStorm))
+- Add Etag support to the vector/geojson worker refactor ([#7074](https://github.com/maplibre/maplibre-gl-js/pull/7074)) (by [@wayofthefuture](https://github.com/wayofthefuture))
```
rivkamatan: "Etag support" may be too vague, maybe something like "Optimize vector tile loading by using vector tile etags to skip processing work when a reloaded tile has not changed"

src/source/vector_tile_source.ts
```diff
@@ -20,6 +20,13 @@ export type VectorTileSourceOptions = VectorSourceSpecification & {
     tileSize?: number;
 };
 
+export type LoadTileResult = {
+    /**
+     * Indicates that the tile request was validated via ETag and it was unmodified.
+     */
+    etagUnmodified?: boolean;
```
rivkamatan: Is it desirable to expose the etag unmodified concept as high up as the Source API? They are not necessarily relevant for all source types and it seems fairly low-level as a concept. 

I like that this lets us return a named parameter instead of a plain boolean.

HarelM: I second this comment, unmodified is enough I think, no need to ge into details as to why in terms of API.

src/source/vector_tile_source.ts
```diff
@@ -272,6 +284,11 @@ export class VectorTileSource extends Evented implements Source {
         if (data && this.map._refreshExpiredTiles) {
             tile.setExpiryData(data);
         }
+
+        if (data && data.etag) {
```
HarelM: Use "?."

src/source/vector_tile_worker_source.test.ts
```diff
@@ -412,15 +439,53 @@ describe('vector tile worker source', () => {
...
+            request: {url: 'http://localhost:2900/faketile.pbf'},
+            etag: '"v1"',
+            subdivisionGranularity: SubdivisionGranularitySetting.noSubdivision,
+        } as any as WorkerTileParameters);
+
+        server.respond();
+        const res = await promise;
+
+        expect(res).toBeDefined();
+        expect((res as any).etagUnmodified).toBe(true);
```
HarelM: Better casting would be better here... (Avoid as any)

src/source/vector_tile_worker_source.test.ts
```diff
@@ -412,15 +439,53 @@ describe('vector tile worker source', () => {
...
+
         expect(res.resourceTiming[0]).toEqual(
             {'duration': 250, 'entryType': 'measure', 'name': 'http://localhost:2900/faketile.pbf', 'startTime': 100}
         );
     });
+
+    test('VectorTileWorkerSource.loadTile returns etagUnmodified when etag matches response and skips parsing', async () => {
+        const source = new VectorTileWorkerSource(actor, new StyleLayerIndex(), []);
+
+        const loadVectorTileSpy = vi.spyOn(source, 'loadVectorTile');
```
HarelM: Is there a more "publish API"-ish way to test this maybe?

src/source/vector_tile_source.ts
wayofthefuture:
```
private _afterTileLoadWorkerResponse(tile: Tile, data: WorkerTileResult) {
        if (data && data.resourceTiming) {
            tile.resourceTiming = data.resourceTiming;
        }

        if (data && this.map._refreshExpiredTiles) {
            tile.setExpiryData(data);
        }

        if (data && data.etag) {
            tile.etag = data.etag;
        }

File: src/source/vector_tile_worker_source.ts [hunk 7]
```diff
@@ -26,7 +26,7 @@ export type LoadVectorTileResult = {
 } & ExpiryData;
 
 export type AbortVectorData = () => void;
-export type LoadVectorData = (params: WorkerTileParameters, abortController: AbortController) => Promise<LoadVectorTileResult | null>;
+export type LoadVectorData = (params: WorkerTileParameters, rawData: ArrayBuffer) => LoadVectorTileResult;
```
HarelM: Is this type still needed? I think it was mainly used for the inheritance part, wasn't it?

src/tile/tile_manager.test.ts
```diff
@@ -2562,3 +2562,22 @@ describe('TileManager::refreshTiles', () => {
...
+
+describe('TileManager / etag', () => {
+    test('skips _tileLoaded when Source.loadTile returns etagUnmodified', async () => {
+        const tileManager = createTileManager();
+        tileManager.onAdd(undefined);
+
+        const tileID = new OverscaledTileID(0, 0, 0, 0, 0);
+        const tile = new Tile(tileID, 512);
+
+        const tileLoadedSpy = vi.spyOn(tileManager as any, '_tileLoaded');
```
HarelM: Can we avoid the "as any" here?

src/tile/tile_manager.ts
```diff
@@ -186,10 +187,18 @@ export class TileManager extends Evented {
 
     async _loadTile(tile: Tile, id: string, state: TileState): Promise<void> {
         try {
-            await this._source.loadTile(tile);
+            const result = await this._source.loadTile(tile) as LoadTileResult;
```
HarelM, line 190: Why is castinig needed here?

src/tile/tile_manager.ts
```diff
@@ -186,10 +187,18 @@ export class TileManager extends Evented {
...
+
+            // Tile data has not changed - reset the reload timer and skip loading the tile.
```
HarelM: Redundant commet, the code is self explanatory

CHANGELOG.md
HarelM: Add credit to @rivkamatan

src/source/vector_tile_source.ts
HarelM: Worth fixing the first one too, the second one is not using data at the second condition...

src/tile/tile_manager.ts
wayofthefuture: Because of the source interface return type  Promise<LoadTileResult | void>

HarelM: Can we change the source interface then? Does it make sense to charge it?

src/source/vector_tile_worker_source.ts
```diff
@@ -25,28 +25,13 @@ export type LoadVectorTileResult = {
     resourceTiming?: Array<PerformanceResourceTiming>;
 } & ExpiryData;
 
-export type AbortVectorData = () => void;
-export type LoadVectorData = (params: WorkerTileParameters, abortController: AbortController) => Promise<LoadVectorTileResult | null>;
-
-/**
- * The {@link WorkerSource} implementation that supports {@link VectorTileSource}.
```
HarelM, line 32: It's still worth keeping some details about this class, even if it now not intended to be easily reused 😀

src/source/vector_tile_worker_source.ts
```diff
@@ -144,13 +137,10 @@ export class VectorTileWorkerSource implements WorkerSource {
         return result;
     }
 
-    _getExpiryData(response: LoadVectorTileResult): ExpiryData {
-        const {expires, cacheControl} = response;
-
+    _getExpiryData({expires, cacheControl}: ExpiryData): ExpiryData {
```
HarelM: Isn't etag part of expiry data now? I would expect to see an assignment here...?

src/source/vector_tile_worker_source.ts
```diff
@@ -132,6 +119,12 @@ export class VectorTileWorkerSource implements WorkerSource {
         }
     }
 
+    _getEtagUnmodifiedResult(response: ExpiryData, timing: RequestPerformance): WorkerTileResult {
+        const cacheControl = this._getExpiryData(response);
+        const resourceTiming = this._finishRequestTiming(timing);
+        return extend({etagUnmodified: true as const}, cacheControl, resourceTiming);
```
HarelM, line 128: Wasn't the name modifid?

HarelM: No, that's a different name, my bad...

src/tile/tile_manager.ts
wayofthefuture: If we remove the `void` from source.ts then all sources loadTile() function complains

rivkamatan: I agree that other Sources need to be able to not return anything, otherwise it would be a break. Does changing it to LoadTileResult | undefined make it easier? Then we could check if the result is null and have it do type narrowing instead of casting.

src/tile/tile_manager.ts
```diff
@@ -186,10 +187,17 @@ export class TileManager extends Evented {
 
     async _loadTile(tile: Tile, id: string, state: TileState): Promise<void> {
         try {
-            await this._source.loadTile(tile);
+            const result = await this._source.loadTile(tile) as LoadTileResult;
+
+            if (result?.unmodified) {
+                this._setTileReloadTimer(id, tile);
```
rivkamatan: Are none of the other fields that are modified in _tileLoaded, like timeAdded, relevant at all here? 

Regardless, I would be somewhat concerned about not sharing the method since it seems easy to get wrong in the unmodified case if other work is ever added to `_tileLoaded`. It's not that the tile was not loaded, it's just that it was not changed when it was loaded, so it may be fragile to separate them.

src/tile/tile_manager.test.ts
```diff
@@ -2562,3 +2562,20 @@ describe('TileManager::refreshTiles', () => {
...
+
+describe('TileManager / etag', () => {
+    test('skips _tileLoaded when Source.loadTile returns unmodified', async () => {
+        const tileManager = createTileManager();
+        tileManager.onAdd(undefined);
+
+        const tileID = new OverscaledTileID(0, 0, 0, 0, 0);
+        const tile = new Tile(tileID, 512);
+
+        const tileLoadedSpy = vi.spyOn(tileManager, '_tileLoaded');
```
rivkamatan: Having the test check if _tileLoaded was called doesn't seem like precisely the right test and it relies on spying on internal methods. The most important part of this case is that the data event does *not* get fired so repainting doesn't happen. Can we test that? I have an example of doing that in #6972 tile_manager.test.ts

src/tile/tile.ts
```diff
@@ -204,6 +205,8 @@ export class Tile {
      * @param justReloaded - `true` to just reload
      */
     loadVectorData(data: WorkerTileResult, painter: any, justReloaded?: boolean | null) {
+        if (data?.etagUnmodified === true) return;
```
rivkamatan: Will it cause issues down the line if we don't set the tile state to 'loaded' at this point? eg will it get stuck in 'reloading' state even though it did actually successfully load?

src/source/worker_source.ts
```diff
@@ -91,8 +92,16 @@ export type WorkerTileResult = ExpiryData & {
         [_: string]: StyleImage;
     } | null;
     glyphPositions?: GlyphPositions | null;
+    etagUnmodified?: false;
 };
 
+export type WorkerTileWithoutData = ExpiryData & {
+    etagUnmodified: true;  // Strict for type narrowing
+    resourceTiming?: Array<PerformanceResourceTiming>;
```
rivkamatan, line 100: The resourceTiming property is the same on both of them, maybe we should share it like how ExpiryData is shared for clarity.

src/source/vector_tile_source.ts
```diff
@@ -265,13 +277,18 @@ export class VectorTileSource extends Evented implements Source {
...
+        if (data?.resourceTiming) {
             tile.resourceTiming = data.resourceTiming;
         }
 
         if (data && this.map._refreshExpiredTiles) {
             tile.setExpiryData(data);
         }
+
+        if (data?.etag) {
+            tile.etag = data.etag;
```
rivkamatan: This was in my initial PR, but rereading this, I think that if the tile has an etag and the data does not have an etag when reloading, the tile's etag should be overwritten with undefined - since the data's missing etag means we don't know whether the old etag is the right match for the current content. So unconditionally:
```suggestion
            tile.etag = data?.etag;
```
It is an edge case since servers should reliably send etags with specific kinds of content but it is probably best not to persist stale etags regardless.

src/source/vector_tile_worker_source.ts
```diff
@@ -102,16 +84,25 @@ export class VectorTileWorkerSource implements WorkerSource {
...
+            // Download the tile data from the network.
+            const tileResponse = await getArrayBuffer(params.request, abortController);
+
+            // Tile data hasn't changed (etag support) - return an unmodified result
+            if (params.etag && params.etag === tileResponse.etag) {
+                this.tileState.finishLoading(uid);
+                return this._getEtagUnmodifiedResult(tileResponse, timing);
+            }
+
+            const tileResult = this.loadVectorTile(params, tileResponse.data);
```
rivkamatan, line 95: Is it still accurate to call this method `loadVectorTile` when it isn't doing the network interaction? It is more involved in converting the ArrayBuffer into a valid VectorTile format than loading the data.

wayofthefuture: Both vector and geojson have similar naming. loadVectorTile -> LoadVectorTileResult. Not sure if there would be better naming.

src/tile/tile_manager.ts
wayofthefuture: If the tile has not changed then why would anything change in the manager? It basically a no-op

src/tile/tile.ts
wayofthefuture: The tile is set to 'loaded' on the first vector load - trace it forward from the reloadTimer function in tile manager and the state does not get changed again - unless I'm missing something please let me know

src/source/vector_tile_source.ts
wayofthefuture: Not sure my thinking is that the etag is tied to the data so it should remain... I believe the guard is higher up

```
            // Tile data hasn't changed (etag support) - return an unmodified result
            if (params.etag && params.etag === tileResponse.etag) {
                this.tileState.finishLoading(uid);
                return this._getEtagUnmodifiedResult(tileResponse, timing);
            }
```

tag === undefined -> false -> it skips this block always reloads...

src/tile/tile_manager.test.ts
wayofthefuture: Yes that's a good idea I'll copy it over

src/tile/tile_manager.ts
wayofthefuture: when i change it to undefined it does the same thing it breaks all the other sources types

src/tile/tile.ts
wayofthefuture: Interesting find -> I can find no occurrence of .state = 'reloading' anywhere in the codebase

src/tile/tile_manager.ts
rivkamatan: The contents of the tile haven't changed but the metadata around timing could have changed, because the tile was refetched, which is why you call `this._setTileReloadTimer` here. I'm suggesting that it might be easier to follow and keep correct in the future if the post-load logic is kept in the same function instead of split out.

Not a strong opinion, more of a style thing, feel free to disagree.

src/tile/tile.ts
rivkamatan: It's not assigned directly that way, `reload` calls _reloadTile with `state`: 
https://github.com/maplibre/maplibre-gl-js/blob/50a6313d249361c85df87f3e60550c297798da02/src/tile/tile_manager.ts#L275
which then calls _loadTile

rivkamatan: For the timer path, it calls _reloadTile with 'expired' state: 
https://github.com/maplibre/maplibre-gl-js/blob/50a6313d249361c85df87f3e60550c297798da02/src/tile/tile_manager.ts#L723

That sets `tile.state = state`:
https://github.com/maplibre/maplibre-gl-js/blob/50a6313d249361c85df87f3e60550c297798da02/src/tile/tile_manager.ts#L293

Then it calls _loadTile, which calls the source's loadTile method. That uses _afterTileLoadWorkerResponse and then tile.loadVectorData to reset the state to 'loaded'. 

So when called from the reload timer would get stuck in 'expired' state if we don't set it back to 'loaded' here.

src/source/vector_tile_source.ts
rivkamatan: That's if the `params.etag` is undefined which would mean the existing tile doesn't have an etag. I'm saying that if `data.etag` (the *new* data's etag response) does not have an etag, we should overwrite the tile's existing `etag` property because it's now stale.

src/source/vector_tile_worker_source.ts
rivkamatan: The `loadTile` naming still makes sense but this specific method is doing something more like `encodeVectorTile` now, since more of the work has moved to the main method.

src/tile/tile.ts
wayofthefuture: I agree with this good find.

src/source/vector_tile_source.ts
wayofthefuture: Ok I understand will do

src/source/vector_tile_worker_source.ts
wayofthefuture: i think this is more of a decode versus an encode... so I think the name works. Also it's a big rename change to do this so might be better for a diff pr/discussion

src/tile/tile_manager.ts
wayofthefuture: Reading this I think we have a bigger issue that reload is only called for inview tiles but load is called for both so I'm going to make some changes.

wayofthefuture:
@rivkamatan Good points I have made some changes to hopefully address them. @HarelM Might want to take a look at recents commits.

HarelM:
Let me know when this is ready for final review after all of @rivkamatan 's comments have been addressed.

wayofthefuture:
@HarelM I believe this is ready...

HarelM:
Looks good, can you update the build size and make sure the newly added code in tile.ts is covered by tests?

HarelM:
@rivkamatan you can open a PR against the geojson-worker-abstract-etag to add a test to cover the added code in tile.ts if you want to push this faster, this is now the only "blocker".

PR #7076, opened 2026-02-07
feat: Add `pseudo` option to `FullscreenControl` to force CSS-based fullscreen mode instead of native fullscreen API

### LINKED ISSUE #6778: FullScreenControl option to force pseudo fullscreen
## User Story

As a developer of a Map application I would like a option to force the `FullScreenControl` to always use pseudo fullscreen (viewport of browser, instead of native fullscreen). 

## Rationale

1. The native full screen can take multiple seconds to "open" (window UI change). The full map is first resized/shown after the transition to full screen is done (at least on macOS). Resizing the map to just cover the browser viewport is instantaneous.
2. Fullscreen can only have one map open at a time. On applications containing a small map widget for example, where you would like to compare the maps in two different browser windows, with pseudo fullscreen you could enlarge each map to the viewport of its respective browser window and view those windows side-by-side.

Pseudo fullscreen is already available and used in the FullScreenControl as a fullback, just not controllable via an option: https://github.com/maplibre/maplibre-gl-js/blob/d16956f51aa9f0ae7ac792b49c115043d89708fc/src/ui/control/fullscreen_control.ts#L170-L182

I would suggest to make pseudo fullscreen available to developers via an option so they can enforce it over the native fullscreen, i.e. like this:

```js
map.addControl(new FullScreenControl({
    pseudo: true,
}), "top-left");
```

While it is of course possible to create a custom control with this behavior, since the functionality is already there, just not accessible via an option, it think adding such a option would be beneficial, and should also have no impact on existing integrations.

## Impact

n/a

CommanderStorm:
## Launch Checklist

 - [x] Link to related issues.

Resolves #6778 by adding the `pseudo` option to the `FulllScreenControll`
CC @stroebjo


src/ui/control/fullscreen_control.ts
```diff
@@ -154,7 +162,9 @@ export class FullscreenControl extends Evented implements IControl {
     };
 
     _exitFullscreen() {
-        if (window.document.exitFullscreen) {
+        if (this._pseudo) {
+            this._togglePseudoFullScreen();
+        } else if (window.document.exitFullscreen) {
```
HarelM, line 166: Not that you changed it, but you can use "x in y" to avoid the "as any" casting.

CommanderStorm: actually, we need to keep the webkit around for iPhone support.. what a mess

CommanderStorm: https://caniuse.com/?search=exitFullscreen

src/ui/control/fullscreen_control.test.ts
```diff
@@ -112,4 +112,70 @@ describe('FullscreenControl', () => {
...
+            value: true,
+            writable: true,
+        });
+
+        const map = createMap();
+        const mapContainer = map.getContainer();
+
+        const fullscreen = new FullscreenControl({pseudo: true});
+        map.addControl(fullscreen);
+        const control = map._controls.find((ctrl) => {
```
HarelM: Can't we use the control we created? I don't like poking inside internal map variables...

src/ui/control/fullscreen_control.test.ts
```diff
@@ -112,4 +112,70 @@ describe('FullscreenControl', () => {
...
+        const requestFullscreenSpy = vi.fn();
+        mapContainer.requestFullscreen = requestFullscreenSpy;
+
+        const fullscreen = new FullscreenControl({pseudo: true});
+        map.addControl(fullscreen);
+        const control = map._controls.find((ctrl) => {
+            return Object.prototype.hasOwnProperty.call(ctrl, '_fullscreen');
+        }) as FullscreenControl;
+
+        control._onClickFullscreen();
```
HarelM: Can we use a button with automation ID instead of using a "private" method?

src/ui/control/fullscreen_control.test.ts
```diff
@@ -112,4 +112,70 @@ describe('FullscreenControl', () => {
...
+
+        // Verify native fullscreen was NOT called
```
HarelM: These commets are redundant

HarelM:
Braided summer tests nit picking this looks good.

src/ui/control/fullscreen_control.ts
CommanderStorm: it is nessesary (`window.document.webkit..` has type `never`), but we can some of the functions.

We don't work in the browsers that these manual polyfills were designed for anyhow (ff < 64, Safari < 12, IE).

We need to keep the `webkit..` around for iPhone support.. what a mess

src/ui/control/fullscreen_control.test.ts
CommanderStorm: not sure what you mean by automation ID, but I think you mean `_fullscreenButton`?

src/ui/control/fullscreen_control.test.ts
```diff
@@ -112,4 +111,65 @@ describe('FullscreenControl', () => {
...
+        const map = createMap();
+        const mapContainer = map.getContainer();
+
+        // Mock requestFullscreen to verify it's NOT called when pseudo is true
+        const requestFullscreenSpy = vi.fn();
+        mapContainer.requestFullscreen = requestFullscreenSpy;
+
+        const fullscreen = new FullscreenControl({pseudo: true});
+        map.addControl(fullscreen);
+        const control = map._controls.find((ctrl) => {
```
HarelM: Same here - remove find control code.

src/ui/control/fullscreen_control.test.ts
HarelM: Not really :-) when testing a UI element, the common practice is to add an automation ID, something like `data-id="fullscreen-button"` to the button itself, and then use `document.querySelector(...)` to find this element.
This means that you are creating a contract between the automation and the UI, but you are not limiting how the internal properties of this class are called.
So in the future you could easily change `_fullscreenButton` to `_fullScreenButtonWithIcon` (stupid example, but you get the point).
The idea, from my point of view, when I review tests is to avoid as much as possible the use of `_` fields in the tests. I would even say that I would prefer to have them private, but this might break a lot of plugins that are relaying on those fields (which is not a great architecture anyway, but it is what it is...).
This is not to say that there are no tests that use these "private" fields, but I've seen the impact of this when refactoring, and I would like to avoid future tests using "private" fields as much as possible.

src/ui/control/fullscreen_control.ts
```diff
@@ -154,27 +161,25 @@ export class FullscreenControl extends Evented implements IControl {
...
+        if (this._pseudo) {
+            this._togglePseudoFullScreen();
+        } else if (this._container.requestFullscreen) {
             this._container.requestFullscreen();
-        } else if ((this._container as any).mozRequestFullScreen) {
-            (this._container as any).mozRequestFullScreen();
-        } else if ((this._container as any).msRequestFullscreen) {
-            (this._container as any).msRequestFullscreen();
         } else if ((this._container as any).webkitRequestFullscreen) {
+            // due to https://caniuse.com/mdn-api_element_requestfullscreen
```
HarelM, line 182: I think you can use `if ('webkitRequestFullscreen' in this._container) { ...` and avoid some `as any`, can't you?

src/ui/control/fullscreen_control.ts
HarelM: See my comment here:
https://github.com/maplibre/maplibre-gl-js/pull/7076/changes#r2778661054

I'm not saying anything about the code, just the improved typings, hopefully, I didn't test it.
If this doesn't help then you can ignore these comments.

CommanderStorm: No that does not work.

Basically this is the case and why it does not work

```ts
let document = {}
if ("foo" in document){
   // document.foo is now never
   document.foo(); // 😑 tsc says no
}
```

src/ui/control/fullscreen_control.test.ts
CommanderStorm: Are we already doing this somewhere?

I am having trouble finding a place where these automation IDs are defined.

src/ui/control/fullscreen_control.ts
HarelM: Ok, thanks!

Issue #4858, opened 2024-10-18
RFC - High-level performance metrics


vcschapp:
This is a request for comments to start a discussion about some possible features. Let me know if it's not the right place, as I'm happy to move this issue or create a new one elsewhere.

### Summary

This is a proposal to instrument MapLibre GL JS with a small set (no more than 2–3) of high-level performance metrics that can be used to drive sustained improvements in initial map load performance and interactivity. A decision framework for selecting metrics to track is given and three metrics are suggested based on the framework: underutilized network time, unrendered resource time, and cancelled network time.

The decision framework and the suggested metrics are independent proposals. For example, the decision framework may be more or less right but there may be better alternatives to the proposed metrics that more closely meet the framework goals; or the decision framework may require refinement.

### Motivation

The things that get measured are the things that improve.

We want to improve the end-user experienced performance of interactive maps on MapLibre, including initial map load time and the time to completely respond to end user interactions like pans and zooms. To improve it, we need to measure it. To be more specific, we need to measure performance in a way that can feed mechanisms that drive performance improvements and prevent regressions.

### Decision Framework

The following tenets are proposed for selecting metrics to instrument into MapLibre:

1. **Easy to understand**. A person with a basic high-level understanding of how map rendering client software works should be able to understand what each specific metric is measuring, why the quantity measured is important, and how each specific metric is meaningfully different from the others.
2. **Limited number**. Measuring too many quantities diffuses the signal, reduces the probability that the metrics are easy to understand, and makes it more likely that nobody is paying attention. The right number of metrics just a handful, around 2 or 3.
3. **Directional, not diagnostic**. The purpose of the metrics is to drive behavior that keeps end-user experienced performance moving in the right direction. It is not to detect specific bottlenecks or identify specific changes that need to be made. The envisioned model, once the metrics are in place, is to select a metric to optimize, select a goal value, and create an issue, contract, or work order to drive the chosen metric down to the goal value without regressing the other metrics. The means to achieve the goal will be left to the person or persons doing the work.
4. **Isolate changeable client behavior**. The end-user experience of maps is influenced by both client-side and server-side factors. Since MapLibre does not control the server side, and the server side can be independently optimized by those who do control it, any useful metric must try to exclude server-side factors such as download TTLB to the extent possible.
[...]

PR #7077, opened 2026-02-07
feat: Add a demo how to implement performance metrics using our callbacks

CommanderStorm:

It is optional and should treeshake out nicely.
For the CDN build, it cannot treeshake out obviously, but that is not a thing we can prevent.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -36,7 +36,6 @@
         "@rollup/plugin-json": "^6.1.0",
         "@rollup/plugin-node-resolve": "^16.0.3",
         "@rollup/plugin-replace": "^6.0.3",
-        "@rollup/plugin-strip": "^3.0.4",
         "@rollup/plugin-terser": "^1.0.0",
         "@rollup/plugin-typescript": "^12.1.4",
         "@stylistic/eslint-plugin": "^5.10.0",
@@ -3423,29 +3422,6 @@
       "license": "MIT",
       "dependencies": {
         "@rollup/pluginutils": "^5.0.1",
-        "magic-string": "^0.30.3"
-      },
-      "engines": {
-        "node": ">=14.0.0"
-      },
-      "peerDependencies": {
-        "rollup": "^1.20.0||^2.0.0||^3.0.0||^4.0.0"
-      },
-      "peerDependenciesMeta": {
-        "rollup": {
-          "optional": true
-        }
-      }
-    },
-    "node_modules/@rollup/plugin-strip": {
-      "version": "3.0.4",
-      "resolved": "https://registry.npmjs.org/@rollup/plugin-strip/-/plugin-strip-3.0.4.tgz",
-      "integrity": "sha512-LDRV49ZaavxUo2YoKKMQjCxzCxugu1rCPQa0lDYBOWLj6vtzBMr8DcoJjsmg+s450RbKbe3qI9ZLaSO+O1oNbg==",
-      "dev": true,
-      "license": "MIT",
-      "dependencies": {
-        "@rollup/pluginutils": "^5.0.1",
-        "estree-walker": "^2.0.2",
         "magic-string": "^0.30.3"
       },
       "engines": {
--- a/package.json
+++ b/package.json
@@ -44,7 +44,6 @@
     "@rollup/plugin-json": "^6.1.0",
     "@rollup/plugin-node-resolve": "^16.0.3",
     "@rollup/plugin-replace": "^6.0.3",
-    "@rollup/plugin-strip": "^3.0.4",
     "@rollup/plugin-terser": "^1.0.0",
     "@rollup/plugin-typescript": "^12.1.4",
     "@stylistic/eslint-plugin": "^5.10.0",


HarelM:
That might be an interesting addition to Maputnik for example, when editing a style.
Thanks for looking into this.

CommanderStorm:
> That might be an interesting addition to Maputnik for example

Could you give more details?
How would you use this? (aka what metrics do you want)

HarelM:
I don't have idea for new matrices, just the ones you made public, to show them in Maputnik on one side of the map, or when debug is on...

src/util/performance.test.ts
```diff
@@ -1,4 +1,339 @@
...
+            expect(metrics2.totalFramesCount).toBe(1);
+        });
+    });
+});
+
+describe('RequestPerformance', () => {
+    let requestParams: RequestParameters;
 import {RequestPerformance} from './performance';
 
 describe('RequestPerformance', () => {
```
CommanderStorm:
```suggestion
});

describe('RequestPerformance', () => {
```

test/build/bundle_size.json
```diff
@@ -1 +1 @@
-1022428
+1023622
```
```suggestion
1022428
```

src/ui/map_tests/map_webgl.test.ts
```diff
@@ -24,6 +24,22 @@ test('does not fire "webglcontextlost" after remove has been called', () => {
...
+    map.on('webglcontextlost', spy);
+    // Dispatch the event manually because at the time of this writing, gl does not support
+    // the WEBGL_lose_context extension.
+    expect(() => {
+        canvas.dispatchEvent(new window.Event('webglcontextlost'));
+    }).not.toThrow();
+    expect(spy).toHaveBeenCalled();
+    map.remove();
+});
+
```
CommanderStorm, line 42:
```suggestion
```

src/ui/map_tests/map_webgl.test.ts
```diff
@@ -24,22 +24,6 @@ test('does not fire "webglcontextlost" after remove has been called', () => {
     expect(spy).not.toHaveBeenCalled();
 });
```
CommanderStorm, line 26:
```suggestion

test('handles "webglcontextlost" when map is created without style', () => {
    // This test verifies fix for #7022 - map should not throw when WebGL context
    // is lost before the style is loaded (i.e., when style is null/undefined)
    const map = createMap({deleteStyle: true});
    const canvas = map.getCanvas();
    const spy = vi.fn();
    map.on('webglcontextlost', spy);
    // Dispatch the event manually because at the time of this writing, gl does not support
    // the WEBGL_lose_context extension.
    expect(() => {
        canvas.dispatchEvent(new window.Event('webglcontextlost'));
    }).not.toThrow();
    expect(spy).toHaveBeenCalled();
    map.remove();
});
```

src/ui/map_tests/map_webgl.test.ts
```diff
@@ -39,7 +39,6 @@ test('handles "webglcontextlost" when map is created without style', () => {
     expect(spy).toHaveBeenCalled();
     map.remove();
 });
```
CommanderStorm, line 41:
```suggestion
});

```

src/util/performance_observer/observer.ts
CommanderStorm, line 1: @vcschapp
might it be possible to get an early review from you if the basic architecture since you wrote the RFC at https://github.com/maplibre/maplibre-gl-js/issues/4858 ?

docs/README.md
```diff
@@ -25,6 +25,8 @@ npm run build-css
 Then generate the docs files:
 
 ```bash
+npm run build-prod
+npm run build-css
```
CommanderStorm:
```suggestion
```

birkskyum:
Just for reference, here's a link to external MapTiler bench for public GL JS api - it doesn't replace this PR, but maybe compliments it.
- https://labs.maptiler.com/sdk-benchmarks/

CommanderStorm:
(as discussed in the meeting now)

@birkskyum all of these are browser-native metrics, so things you can get without any instrumentation.
I am proposing a few more invasive metrics such as time to first load vs time to full load, frame timing, ability to enable performance.mark.

There are likely a few more metrics that are missing.
The following PR proposed 3 different metrics
- https://github.com/maplibre/maplibre-gl-js/issues/4858 

They were:
- Underutilized network time
- Unrendered resource time
- Cancelled network time

In the end, this system needs to be extendable to enable all of these peformance mesurement usecases without being a special purpose API such as [`MapOptions.collectResourceTiming`](https://maplibre.org/maplibre-gl-js/docs/API/type-aliases/MapOptions/#collectresourcetiming)

test/examples/display-performance-metrics.html
```diff
@@ -0,0 +1,153 @@
...
+    const intervalId = setInterval(updateMetricsDisplay, 100); // Update every 1 second
+
+    map.on("remove", () => {
+        clearInterval(intervalId); // Clear the interval when the map is removed
+    });
+
+    map.on("error", (e) => {
+        console.error("Map error:", e.error);
+        metricsElement.innerHTML = `Error loading map: ${e.error.message}`;
+        clearInterval(intervalId); // Also clear on error
```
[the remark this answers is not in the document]
CommanderStorm:
```suggestion
    map.on("error", (e) => {
        console.error("Map error:", e.error);
        clearInterval(intervalId);
```

test/examples/display-performance-metrics.html
HarelM, line 1: This example turned out to be a bit long - ~140 lines, is there a way to simplify it? May just show part of the matrices and link to the docs for more?

src/util/performance_observer/rendering.ts
```diff
@@ -0,0 +1,79 @@
...
+export class RenderingPerformanceObserver implements IPerformanceObserver {
+    private _lastFrameTime?: number;
+    private _lastFrameDuration: number;
+    private _totalFrameTime = 0;
+    private _totalFrameCount = 0;
+    private _totalDroppedFrameCount = 0;
+
+    /** {@inheritDoc} */
+    observe(type: PerformanceEventType, timestamp: number): void {
+        if (type ===  'startOfFrame') {
```
HarelM: [nit picking] can these ifs be reverted to reduce indentation?

src/util/performance_observer/observer.ts
```diff
@@ -0,0 +1,49 @@
...
+ * Observer interface for performance monitoring
+ * @group Performance
+ */
+export interface IPerformanceObserver {
+    /**
+     * Called when a performance event occurs
+     * @param type - the type of event
+     * @param timestamp - when the event occurred
+     */
+    observe(type: PerformanceEventType, timestamp: number): void;
```
HarelM: I think I'm having trouble with this API. Feel like the observers should know to which events to listen without passing the even type to the observe method.
I would expect to see something link "getObserverByEvent(eventType)" and then observer.observe, or observer.log.
IDK... something feels weird with this design but I can't place it...
I think the way the tests look is what got me confused as I see a test like the following:
```
observer.observe('startOfFrame', 0);
observer.observe('startOfFrame', 10);  // >60 fps
observer.observe('startOfFrame', 20);  // >60 fps
observer.observe('startOfFrame', 30);  // >60 fps
```
This is a weird looking test which I think caused by a weird API.
But I'm not sure I have a better idea how to solve this...

src/util/performance_observer/marker.ts
```diff
@@ -0,0 +1,31 @@
+import {type PerformanceEventType, type IPerformanceObserver} from './observer';
+
+/**
+ * Marks points in time when certain events occur
+ * 
+ * This is useful because they are visible in the performance timeline
+ * @group Performance
+ */
+export class MarkingPerformanceObserver implements IPerformanceObserver {
+    private _markedEvents: Set<string> = new Set();
```
HarelM: What is this used for? I see it as "write only" but I don't see where it is being read...?

src/ui/map.ts
```diff
@@ -686,7 +693,8 @@ export class Map extends Camera {
     transformConstrain: TransformConstrainFunction | null;
 
     constructor(options: MapOptions) {
-        PerformanceUtils.mark(PerformanceMarkers.create);
+        const performanceSubject = new PerformanceSubject(options.performanceObservers ?? []);
```
HarelM: Why not use the `_performanceSubject` member instead of a "temporary" one?

CommanderStorm: We are right at the start of the constructor, so `_performanceSubject` does not exist yet.

src/util/performance_observer/marker.ts
CommanderStorm: It is used to know which events to clear in `disconnect`

src/util/performance_observer/observer.ts
CommanderStorm: The alternative that I thought of was `.observeStartOfFrame()` and such, but that did not seem like the most ergonomic way to do this.

Not sure.

Query based APIs are another abstraction layer that I would like to avoid.

CommanderStorm: Basically, `.observeStartOfFrame` means that `PerformanceSubject` is likely inlined.

Both are fine, we just have to make a choice

HarelM:
Sorry for the delayed review, I need to dedicate time for this to fully understand what you did here, so it might take a bit longer than usual...

src/util/performance_observer/observer.ts
HarelM: Is it possible to remove the number from the API then? Keep browser.now inside the observer maybe?

HarelM:
Finally had some time to look at this more closely.
Thanks for opening the PR.
I still find this API weird.
From a benchmark perspective and given the current benchmarks, I don't feel like this is solving the problem well.
What I would like to achieve is the ability to take any (future) version of maplibre-gl-js and be able to compare performance without needing to build it like we do for the benchmarks today.
I believe that in order to do that only raising event is needed.
If all the events are fired, one could listen to these events and calculate the relevant performance matrices they are interested in.
All the logic around different observers can then move to a plugin or external code and avoid all this complexity inside the code.
So I think the bottom line is, that an event based architecture is what I would like to see push forward as part of this PR and not internal specific classes for performance that might not be what other people are looking to measure.
The example that you wrote should still be able to provide an initial code to showcase how one could measure things related to maplibre.

Sorry it took me so long to review. If the above is not clear ping me on slack and we can have a deeper chat on this.
THANKS!

src/source/geojson_worker_source.ts
```diff
@@ -1,5 +1,6 @@
 import {getJSON} from '../util/ajax';
-import {RequestPerformance} from '../util/performance';
+import {RequestPerformance} from '../util/request_performance';
+import rewind from '@mapbox/geojson-rewind';
```
CommanderStorm:
```suggestion
```

src/ui/map.ts
```diff
@@ -22,7 +22,7 @@ import {Event, ErrorEvent, type Listener} from '../util/evented';
 import {type MapEventType, type MapLayerEventType, MapMouseEvent, type MapSourceDataEvent, type MapStyleDataEvent} from './events';
 import {TaskQueue} from '../util/task_queue';
 import {throttle} from '../util/throttle';
-import {PerformanceMarkers, PerformanceUtils} from '../util/performance';
+import {webpSupported} from '../util/webp_supported';
```
```suggestion
```

typedoc.json
```diff
@@ -7,7 +7,8 @@
         "Geography and Geometry",
         "Handlers",
         "Sources",
-        "Event Related"
+        "Event Related",
+        "Performance"
```
CommanderStorm:
```suggestion
        "Event Related"
```

CommanderStorm:
>  order to do that only raising event is needed

I guess it can be implemented this way.
Much higher overhead likely though..

typedoc.json
```diff
@@ -31,8 +31,3 @@
     "externalSymbolLinkMappings": {
         "@maplibre/maplibre-gl-style-spec": {
             "ProjectionSpecification": "https://maplibre.org/maplibre-style-spec/projection/",
```
CommanderStorm, line 33:
```suggestion
            "ProjectionSpecification": "https://maplibre.org/maplibre-style-spec/projection/",
            "LayerSpecification": "https://maplibre.org/maplibre-style-spec/layers/",
            "SourceSpecification": "https://maplibre.org/maplibre-style-spec/sources/"
        }
    }
  }

```

typedoc.json
```diff
@@ -35,4 +35,5 @@
             "SourceSpecification": "https://maplibre.org/maplibre-style-spec/sources/"
         }
     }
-  }
\ No newline at end of file
+  }
+
```
CommanderStorm:
```suggestion
  }
```

test/build/sourcemaps.test.ts
```diff
@@ -84,6 +84,6 @@ describe('main sourcemap', () => {
         const s1 = setMinus(actualEntriesInSourcemapJSON, expectedEntriesInSourcemapJSON);
         expect(s1.length).toBeLessThan(5);
         const s2 = setMinus(expectedEntriesInSourcemapJSON, actualEntriesInSourcemapJSON);
-        expect(s2.length).toBeLessThan(16);
+        expect(s2.length).toBeLessThan(17);
```
CommanderStorm:
```suggestion
        expect(s2.length).toBeLessThan(16);
```

src/source/geojson_worker_source.ts
```diff
@@ -374,3 +374,4 @@ function getSuperclusterOptions({superclusterOptions, clusterProperties}: LoadGe
     };
     return superclusterOptions;
 }
+
```
CommanderStorm:
```suggestion
```

src/source/vector_tile_worker_source.ts
```diff
@@ -237,3 +237,4 @@ export class VectorTileWorkerSource implements WorkerSource {
         this.tileState.removeLoaded(params.uid);
     }
 }
+
```
CommanderStorm:
```suggestion
```

src/ui/map.ts
```diff
@@ -3934,3 +3926,4 @@ export class Map extends Camera {
         return this._update(true);
     }
 }
+
```
```suggestion
```

src/util/request_performance.ts
```diff
@@ -0,0 +1,35 @@
...
+
+            // cleanup
+            performance.clearMarks(this.start);
+            performance.clearMarks(this.end);
+            performance.clearMeasures(this.measure);
+        }
+
+        return resourceTimingData;
+    }
+}
```
CommanderStorm:
```suggestion
}

```

HarelM:
I've posted a message in the original issue, let's give it a few more days before I merge this.
We can always add more events as we see fit in the future to allow more fine grained measurements.

HarelM:
No input... merging.

PR #7078, opened 2026-02-08
test: More thoroughly test `Hash`

CommanderStorm:
## Launch Checklist

I think the actual changes in https://github.com/maplibre/maplibre-gl-js/pull/7073 are getting hard to review.
Lets pull the testcase changes out so that the actual changes are more obvious.

### DIFF
--- a/src/ui/hash.test.ts
+++ b/src/ui/hash.test.ts
@@ -494,7 +494,6 @@ describe('hash', () => {
 
         map.remove();
 
-        expect(map).toBeTruthy();
     });
 
     test('hash with URL in other parameter does not change', () => {


src/ui/hash.test.ts
```diff
@@ -359,6 +373,18 @@ describe('hash', () => {
             expect(hash._isValidHash(hash._getCurrentHash())).toBeFalsy();
         });
 
+        test('invalidate hash, only one value', () => {
+            window.location.hash = '#24';
+
+            expect(hash._isValidHash(hash._getCurrentHash())).toBeFalsy();
```
HarelM, line 379: I know this was not changed as part of this PR, but if there was a better way to test these internal methods without calling them explicitly, it would've been better...

src/ui/hash.test.ts
```diff
@@ -386,80 +412,270 @@ describe('hash', () => {
...
+            createHash().addTo(map);
+            map.setZoom(3);
+            expect(window.location.hash).toBe('#3/0/0');
+            expect(window.location.href).toBe('http://localhost/#3/0/0');
+            map.setCenter([2.0, 1.0]);
+            expect(window.location.hash).toBe('#3/1/2');
+            expect(window.location.href).toBe('http://localhost/#3/1/2');
+        });
+
+        test('http://localhost#', () => {
```
HarelM, line 446: How is this different from the test just before the previous one? `test('http://localhost/#', () => {`

src/ui/hash.test.ts
```diff
@@ -386,80 +412,270 @@ describe('hash', () => {
...
+            createHash().addTo(map);
+            map.setZoom(5);
+            expect(window.location.hash).toBe('#5/0/0');
+            expect(window.location.href).toBe('http://localhost/#5/0/0');
+            map.setCenter([2.0, 1.0]);
+            expect(window.location.hash).toBe('#5/1/2');
+            expect(window.location.href).toBe('http://localhost/#5/1/2');
+        });
+
+        test('http://localhost', () => {
```
HarelM, line 478: Same here - same test just before the previous one...?

src/ui/hash.test.ts
```diff
@@ -386,80 +412,270 @@ describe('hash', () => {
...
-        expect(window.location.hash).toBe('#3/1/2');
-        expect(window.location.href).toBe('http://localhost/#3/1/2');
+    test('map.remove', () => {
+        const container = window.document.createElement('div');
+        Object.defineProperty(container, 'clientWidth', {value: 512});
+        Object.defineProperty(container, 'clientHeight', {value: 512});
+
+        map.remove();
+
+        expect(map).toBeTruthy();
```
HarelM: `expect(() =>map.remove()).not.toThrow()`? I'm not sure what this is testing...

src/ui/hash.test.ts
```diff
@@ -386,80 +412,270 @@ describe('hash', () => {
...
+            .addTo(map);
+
+        window.location.hash = '#map=10/3/-1&foo=bar&';
+        hash._onHashChange();
+        map.setZoom(11);
+        expect(window.location.hash).toBe('#map=11/3/-1&foo=bar');
+
+    });
+
+    test('hash with double ampersand', () => {
```
HarelM, line 546: `hash with double ampersand changes to a valid one`?

src/ui/hash.test.ts
```diff
@@ -386,80 +412,270 @@ describe('hash', () => {
...
+            .addTo(map);
+
+        window.location.hash = '#map=10/3/-1&&foo=bar';
+        hash._onHashChange();
+        map.setZoom(12);
+        expect(window.location.hash).toBe('#map=12/3/-1&foo=bar');
+
+    });
+
+    test('hash with leading ampersand', () => {
```
HarelM: `hash with leading ampersand gets removed`?

src/ui/hash.test.ts
```diff
@@ -386,80 +412,270 @@ describe('hash', () => {
...
+        map.setZoom(13);
+        expect(window.location.hash).toBe('#map=13/3/-1&foo=bar');
+    });
+
+    test('hash with empty parameter values should be invalid', () => {
+        const hash = createHash('map')
+            .addTo(map);
+
+        window.location.hash = '#map=&foo=bar';
+        expect(hash._onHashChange()).toBeFalsy();
```
HarelM, line 572: Not sure I follow what this is testing... From the names of the methods it seems like it's calling `_onHashChange` which I'm not is descriptive enough.
Consider removing this test or improving its readability.

src/ui/hash.test.ts
```diff
@@ -386,80 +412,270 @@ describe('hash', () => {
...
-        expect(window.location.href).toBe('http://localhost/#4/1/2');
+    describe('geographic boundary values', () => {
+        let hash: Hash;
+
+        beforeEach(() => {
+            hash = createHash()
+                .addTo(map);
+        });
+
+        test('Near south pole, dateline (map may clamp to Mercator limits)', () => {
```
HarelM: `map will`? Not sure why use "may" in a test that should be deterministic...

src/ui/hash.test.ts
```diff
@@ -386,80 +412,270 @@ describe('hash', () => {
...
+        expect(map1.getZoom()).toBe(7);
+        expect(map1.getCenter().lat).toBe(5);
+        expect(map1.getCenter().lng).toBe(6);
+
+        hash2._onHashChange();
+        expect(map2.getZoom()).toBe(12);
+        expect(map2.getCenter().lat).toBe(7);
+        expect(map2.getCenter().lng).toBe(8);
+
+        // Clean up
```
HarelM, line 674: use `beforeEach` and `afterEach`?

HarelM:
Sorry for spamming... I hope this is not too much...

src/ui/hash.test.ts
CommanderStorm:
```suggestion
    test('hash with leading ampersand removes leading ampersand', () => {
```

src/ui/hash.test.ts
CommanderStorm:
```suggestion
        test('Near south pole, dateline', () => {
```

src/ui/hash.test.ts
CommanderStorm: these are existing tests (=did not change) that are nessesary due to the manual parsing of the hash

src/ui/hash.test.ts
CommanderStorm: Not sure. It is a legacy test.
Not throwing should also be tested by just being a test.

```suggestion
```

src/ui/hash.test.ts
HarelM: I think there are duplicate tests here, can you verify me and remove duplication?
For example:
L424-L432:
```ts
            window.location.href = 'http://localhost/#';
            createHash().addTo(map);
            map.setZoom(3);
            expect(window.location.hash).toBe('#3/0/0');
            expect(window.location.href).toBe('http://localhost/#3/0/0');
            map.setCenter([2.0, 1.0]);
            expect(window.location.hash).toBe('#3/1/2');
            expect(window.location.href).toBe('http://localhost/#3/1/2');
        });
```
L447-L454
```ts
            window.location.href = 'http://localhost#';
            createHash().addTo(map);
            map.setZoom(4);
            expect(window.location.hash).toBe('#4/0/0');
            expect(window.location.href).toBe('http://localhost/#4/0/0');
            map.setCenter([2.0, 1.0]);
            expect(window.location.hash).toBe('#4/1/2');
            expect(window.location.href).toBe('http://localhost/#4/1/2');
```

CommanderStorm: if we were using the way proposed by @1ec5 yes, but `http://localhost#` and `http://localhost/#` are different urls on paper at least, even though they normalise to the same thing.

This is purely a test that the normalisation works as expected, so can be removed in the next PR.

src/ui/hash.test.ts
CommanderStorm: This is cleanup only for this test, I don't think `beforeEach` makes it more redable.

src/ui/hash.test.ts
CommanderStorm: I don't know how to do this.

Given that we really want these to be correct, I think not using the public API is better here.

Also, the methods are not marked as private, so technically they are public api ^^

src/ui/hash.test.ts
HarelM: Oh, I missed the extra "/". Ignore me...

PR #7079, opened 2026-02-08
fix: guard against unhydrated layout in PauseablePlacement

### LINKED ISSUE #1835: SymbolLayer crashes on style update with `Cannot read properties of undefined (reading 'get')`
## Description

This is a copy of this issue https://github.com/visgl/react-map-gl/issues/1844 that also occurs on the latest version of maplibre (ie. 2.4.0).

This is what I have in my console : 

```
maplibre-gl.css?7443:82 Uncaught TypeError: Cannot read properties of undefined (reading 'get')
    at new LayerPlacement (maplibre-gl.css?7443:82:1)
    at PauseablePlacement.continuePlacement (maplibre-gl.css?7443:82:1)
    at Style._updatePlacement (maplibre-gl.css?7443:82:1)
    at Map._render (maplibre-gl.css?7443:82:1)
    at map._render (mapbox.ts:557:1)
    at maplibre-gl.css?7443:82:1
```

I updated the codesandbox from the source issue to do it with the latest version of maplibre : https://codesandbox.io/s/maplibre-issue-symbol-layer-wvqbiy 
Just click on the map to see the error.

The error comes from this line: https://github.com/maplibre/maplibre-gl-js/blob/main/src/style/pauseable_placement.ts#L22
The `layout` attribut on `styleLayer` can be `undefined` sometimes but  Idon't know why...

garethbowker:
Fixes #1835

When a symbol layer is added via `addLayer()`, there's a brief window where the layer
exists in `_layers` with `type === 'symbol'` but `layout` hasn't been initialized yet.

If `continuePlacement()` hits that layer during this window, `LayerPlacement`'s
constructor calls `layout.get('symbol-sort-key')` on `undefined`, which throws
`Cannot read properties of undefined (reading 'get')`.

The fix adds a guard that skips layers whose `layout` property hasn't been hydrated yet.
These layers will be picked up on the next placement cycle once they're fully initialized.

Tested with ~77 dynamic layers (~23 symbol layers) being added in quick succession
during map init. Haven't seen the crash since applying this fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,6 @@
 ### 🐞 Bug fixes
 
 - Fix incorrect popup location in case of terrain and `jumpTo` ([#7267](https://github.com/maplibre/maplibre-gl-js/issues/7267)) (by [@HarelM](https://github.com/HarelM))
-- _...Add new stuff here..._
 - Fix memory leak in VideoSource: remove `playing` event listener and pause video on source removal ([#7279](https://github.com/maplibre/maplibre-gl-js/pull/7279)) (by [@johanrd](https://github.com/johanrd))
 - Fix memory leak where typed array views retained StructArray buffers after GPU upload, preventing garbage collection ([#7280](https://github.com/maplibre/maplibre-gl-js/pull/7280)) (by [@johanrd](https://github.com/johanrd))
 - Fix raster DEM tiles getting stuck in `"reloading"` state ([#7277](https://github.com/maplibre/maplibre-gl-js/pull/7277)) (by [@katemihalikova](https://github.com/katemihalikova))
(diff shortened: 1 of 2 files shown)


HarelM:
Do you think you'll be able to add a test (probably a unit test) to cover this bug fix?

src/style/pauseable_placement.ts
```diff
@@ -103,6 +103,7 @@ export class PauseablePlacement {
             const layer = layers[layerId];
             const placementZoom = this.placement.collisionIndex.transform.zoom;
             if (layer.type === 'symbol' &&
+                (layer as any as SymbolStyleLayer).layout &&
```
HarelM: Is there a way to avoid the casting here? Doesn't `layer.type === 'symbol` narrows the type?

garethbowker:
Thanks for the review @HarelM!

On the cast: Yeah, `layers` is typed as `{[_: string]: StyleLayer}` rather than a discriminated union, so the `=== 'symbol'` check doesn't narrow. I've switched to the existing `isSymbolStyleLayer()` type guard (from `symbol_style_layer.ts`), which does the same runtime check but with a `layer is SymbolStyleLayer` return type - so TypeScript narrows properly and both `as any as` casts are eliminated.

On the test: Added a unit test that constructs a `PauseablePlacement` with a mock symbol layer where `layout` is `undefined` and verifies `continuePlacement` skips it without throwing. The underlying race condition (style recalculation not finishing before the placement loop iterates) is hard to reproduce deterministically, but this test validates the guard itself.

src/style/pauseable_placement.test.ts
```diff
@@ -0,0 +1,41 @@
+import {describe, expect, test} from 'vitest';
+import {PauseablePlacement} from './pauseable_placement';
+import {MercatorTransform} from '../geo/projection/mercator_transform';
+import type {StyleLayer} from './style_layer';
+
+describe('PauseablePlacement', () => {
+    // When a style change adds or reorders symbol layers, the placement loop
```
HarelM: Consider linking to the relevant issue instead of this long intro comment, or the relevant PR...

PR #7080, opened 2026-02-08
Mention Rotation in the  `NavigationControl.ResetBearing` message

CommanderStorm:
## Launch Checklist

 - [x] Link to related issues.

This continues https://github.com/maplibre/maplibre-gl-js/pull/6717
Resolves https://github.com/maplibre/maplibre-gl-js/issues/6716

### Problem statement

Users do not discover that the `NavigationControl` bearing button supports map rotation by dragging the button.

From a user’s perspective, the control presents itself as a “reset to north” button, both in iconography and tooltip text. After the map has been reset to north, the control continues to communicate the same single-purpose behavior.

As a result, users reasonably conclude that:

- the control only resets bearing, and
- rotation is not supported via this control.

There is currently no affordance or feedback (except that we show the grab handle instead of the pointer) that suggests an additional interaction (e.g. dragging) is possible. Discoverability relies on accidental interaction or prior knowledge, which leads to confusion and missed functionality.

The issue is therefore discoverability of rotation capability, not correctness of the current behavior.

## Solution

We can either:

- remove this behaviour (it is arguably rather non-standard)
- document it in the helptext

For backwards compat, I chose the latter


HarelM (approved):
I think it has a poor UX, dragging on top of such a small button is not great. 
But changing the text can't hurt...

PR #7094, opened 2026-02-09
fix: handle WebGL context loss when style is not loaded

### LINKED ISSUE #7022: Javascript error when WebGL context is lost before map style is loaded  (v5.12+)
**maplibre-gl-js version**: 5.12 and later

**browser**: every

If the WebGL context is lost and the map style has not yet been loaded, a javascript error occures in the following line in `map.ts`:

https://github.com/maplibre/maplibre-gl-js/blob/ecfba847c0abb4bfa7aab619eec376dc1c10b93d/src/ui/map.ts#L3432

Just form looking at the code it's pretty obvious. `this.style._layers` is accessed without any null/undefined checks. If the map style is not yet loaded and parsed (which might happen), `this.style` is null/undefined. This piece of code was introduced in version 5.12 with this PR:  https://github.com/maplibre/maplibre-gl-js/pull/6398

kaigritun:
## Summary

Fix TypeError when WebGL context is lost before the map style is loaded.

Fixes #7022

## Problem

When a map is created without providing a `style` option (or before the style finishes loading), `this.style` is `null`. The `_contextLost` handler was then trying to:
1. Iterate over `this.style._layers` - causes `TypeError: Cannot read properties of null (reading '_layers')`
2. Call `this.style.destroy()` - would also fail

## Solution

Added a null check around the style-dependent operations in the `_contextLost` handler. The `_getStyleAndImages()` method already handles the null case properly, so it's called before the null check.

## Testing

Added a new unit test that creates a map without a style and dispatches a `webglcontextlost` event, verifying that:
1. No error is thrown
2. The event handler is still called properly

All existing tests continue to pass.

## Minimal reproduction (from issue)

```html
<!DOCTYPE html>
<html>
  <head>
    <script src="https://unpkg.com/maplibre-gl@5.16.0/dist/maplibre-gl.js"></script>
  </head>
  <body>
    <div id="map"></div>
    <script>
          container: "map"
        });

      const canvas = document.getElementsByTagName("canvas")[0];
      const context = canvas.getContext("webgl2") || canvas.getContext("webgl");
      context.getExtension("WEBGL_lose_context").loseContext();
    </script>
  </body>
</html>
```

### DIFF
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -358,7 +358,9 @@ export type MapOptions = {
      */
     pixelRatio?: number;
     /**
-     * If false, style validation will be skipped. Useful in production environment.
+     * If false, style validation will be skipped.
+     * Useful in production environments due to enabling tree-shaking of the validation code in some environments and minor performance improvements.
+     * Disabling this option comes at the cost of less clear error messages
      * @defaultValue true
      */
     validateStyle?: boolean;
@@ -396,7 +398,7 @@ export type MapOptions = {
     experimentalZoomLevelsToOverscale?: number;
     /**
      * Determines the rotation interaction model:
-     * - When true: Uses "Orbital" logic where rotation is relative to the pivot center. 
+     * - When true: Uses "Orbital" logic where rotation is relative to the pivot center.
      *   Dragging right at the top rotates clockwise, while dragging right at the bottom
      *   rotates counter-clockwise (like spinning a physical globe).
      * - When false: Uses "Linear" logic where horizontal mouse movement translates directly
@@ -3716,7 +3718,7 @@ export class Map extends Camera {
         this._container.removeEventListener('scroll', this._onMapScroll, false);
         this._container.classList.remove('maplibregl-map');
 
-        PerformanceUtils.clearMetrics();
+        PerformanceUtils.remove();
 
         this._removed = true;
         this.fire(new Event('remove'));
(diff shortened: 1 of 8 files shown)


src/ui/map.ts
```diff
@@ -3428,22 +3428,26 @@ export class Map extends Camera {
...
         this.painter.destroy();
 
-        // check if style contains custom layers to warn user that they can't be restored automatically
-        for (const layer of Object.values(this.style._layers)) {
-            if (layer.type === 'custom') {
-                console.warn(`Custom layer with id '${layer.id}' cannot be restored after WebGL context loss. You will need to re-add it manually after context restoration.`);
-            }
+        this._lostContextStyle = this._getStyleAndImages();
+
+        if (this.style) {
```
HarelM: Nit picking: Can you invert the if to reduce indentation and improve readabiality?

HarelM:
Thanks for taking the time to report this issue!
Can you add a changelog entry please?

kaigritun:
Added the changelog entry and inverted the if statement for better readability. Thanks for the review!

MichaelDeBoey:
I'm not making any claims about the quality of their work, but I wanted to let you know that @kaigritun is a fully-autonomous non-human actor
https://socket.dev/blog/ai-agent-lands-prs-in-major-oss-projects-targets-maintainers-via-cold-outreach

HarelM:
Thanks for the heads up @MichaelDeBoey, this PR is a pretty simple bug fix so there's no harm in it, but I'll continue monitor suspicious PRs.

PR #7099, opened 2026-02-10
feat： Add render hooks to Painter for external render passes

cavencj:
Some advanced integrations require executing external rendering logic at specific points in the render pipeline. While custom layers provide a flexible extension mechanism, they are rendered as part of the layer-based pipeline and cannot be inserted at arbitrary render pass boundaries.

This PR introduces a lightweight render hook mechanism to Painter that allows external rendering logic to be executed at well-defined points in the render pipeline, without modifying the existing layer system or render order.
The goal is to enable advanced integrations (e.g. external renderers, custom GPU passes, Three.js) while keeping the core render flow intact and easy to maintain.

What’s added

A small, synchronous render hook API on Painter

Four hook phases aligned with the existing render passes:

- beforeOpaque

- afterOpaque

- beforeTranslucent

- afterTranslucent

Hooks receive the active Painter instance and current RenderOptions

No existing rendering logic is changed; hooks are executed strictly in addition to the current pipeline.


HarelM:
`painter` in not really part of the public API, the `Map` class is usually where things are configured/registered.
I would also recommend adding an example of how to use these hooks in the examples folder. It would also help to get some more context on a problem that this solves that can't be solved otherwise, or the alternative is a lot more complicated.
Also some tests would be great to cover the new code.
I think the changes are minimal and powerful at the same time, which is good.
What will happen on `contextlost` event when I think the painter is destroyed? Should there be a warning printed to console like we do for custom layers?
Thanks!!

HarelM:
No response, if this is still relevant please open a new PR.

PR #7110, opened 2026-02-12
feat: expose original style URL via map.getStyleUrl()

### LINKED ISSUE #7109: Feature: Expose the original style URL via map.getStyleUrl()
## User Story

As a plugin developer, I can retrieve the original style URL used to initialize the map, so that I can fetch the original basemap style and identify which layers came from the basemap vs. layers added dynamically.

When a map is initialized with a style URL, that URL is not accessible later. The `map.getStyle()` method returns the current style object which includes all layers - both from the original basemap and any layers added dynamically via `map.addLayer()`. There's no way to distinguish between them.

### Example use cases

- **Layer control plugins** that want to group basemap layers as a single "Basemap" entry while listing user-added layers individually
- **Style management tools** that need to know what basemap is being used
- **Map state serialization** that needs to store the original style URL separately from dynamic modifications

### Current workaround

Currently, developers must either:
1. Ask users to manually provide the style URL as a separate option
2. Track the style URL themselves before passing it to the Map constructor
3. Have no way to identify original basemap layers

## Impact

Without this feature, plugin developers cannot reliably identify which layers came from the original basemap style. This forces either:
- Manual configuration by end users (poor developer experience)
- Complex workarounds that require intercepting map initialization
- Inability to provide features that distinguish basemap from user layers

## Proposed API

```typescript
// Get the style URL (returns undefined if style was provided as object)
const styleUrl = map.getStyleUrl();

// After setStyle() with a URL, it would update accordingly
map.setStyle('https://example.com/style.json');
map.getStyleUrl(); // 'https://example.com/style.json'

// If style was set as an object, returns undefined
map.setStyle({ version: 8, sources: {}, layers: [] });
map.getStyleUrl(); // undefined
```

giswqs:
## Summary

- Add `map.getStyleUrl()` method that returns the URL used to load the current style
- Returns `null` if the style was loaded from a JSON object
- Preserves URL across style diffing and fallback scenarios

## Motivation

Closes #7109

When a map is initialized with a style URL, that URL is not accessible later. The `map.getStyle()` method returns the current style object which includes all layers - both from the original basemap and any layers added dynamically. There's no way to distinguish between them.

This feature enables plugin developers to:
- Build layer control plugins that group basemap layers as a single "Basemap" entry
- Implement style management tools that need to know what basemap is being used
- Serialize map state with the original style URL separately from dynamic modifications

## API

```typescript
// Get the style URL (returns null if style was provided as object)
const styleUrl = map.getStyleUrl();

// After setStyle() with a URL, it returns the URL
map.setStyle('https://example.com/style.json');
map.getStyleUrl(); // 'https://example.com/style.json'

// If style was set as an object, returns null
map.setStyle({ version: 8, sources: {}, layers: [] });
map.getStyleUrl(); // null
```

## Test plan

- [x] Added unit tests for `Style.getStyleUrl()` in `style.test.ts`
- [x] Added unit tests for `Map.getStyleUrl()` in `map_style.test.ts`
- [x] All 2710 existing unit tests pass


src/ui/map.ts
```diff
@@ -2124,7 +2126,11 @@ export class Map extends Camera {
             warnOnce(
                 `Unable to perform style diff: ${e.message || e.error || e}.  Rebuilding the style from scratch.`
             );
+            // When falling back, use the JSON and preserve the URL if one was provided
             this._updateStyle(style, options);
+            if (sourceUrl && this.style) {
```
HarelM: This doesn't look like it is the responsibility of this map class to change this field...

HarelM:
I think that both map and style classes are updating an internal property of the style class, which is not a good pattern. 
Please see if you can change that...

src/ui/map.ts
giswqs: Thanks for the feedback! I've refactored to address this:

1. Added a `setStyleUrl()` public method to the Style class
2. Added an optional `sourceUrl` parameter to `loadJSON()` for the diff fallback case
3. Updated Map to use `setStyleUrl()` instead of directly accessing `_styleUrl`
4. Created `_updateStyleWithSourceUrl()` helper to pass the URL through `loadJSON()`

Now all `_styleUrl` updates happen within the Style class methods.

src/ui/map.ts
```diff
@@ -2124,10 +2126,29 @@ export class Map extends Camera {
             warnOnce(
                 `Unable to perform style diff: ${e.message || e.error || e}.  Rebuilding the style from scratch.`
             );
-            this._updateStyle(style, options);
+            // When falling back, pass the sourceUrl to loadJSON via _updateStyle
+            this._updateStyleWithSourceUrl(style, options, sourceUrl);
         }
     }
 
+    private _updateStyleWithSourceUrl(style: StyleSpecification, options?: StyleSwapOptions & StyleOptions, sourceUrl?: string) {
```
HarelM: This looks like a bug fix...? Is it related to this PR? I'm not sure I like this waiting approach... Is there an alternative?

src/style/style.ts
```diff
@@ -1921,6 +1926,25 @@ export class Style extends Evented {
...
+        return this._styleUrl;
+    }
+
+    /**
+     * Sets the URL of the style. This is used internally when the style is loaded from a URL
+     * or when the style URL needs to be updated during diff operations.
+     *
+     * @param url - The style URL, or null if the style is not loaded from a URL.
+     */
+    setStyleUrl(url: string | null) {
```
HarelM: Is it possible not to expose this method and handle updating the style in the relevant place without exposing this?
This API is misleading as it looks like you can set the style URL and have the style updated using this method, but that's not the case...

src/ui/map_tests/map_style.test.ts
```diff
@@ -514,4 +514,96 @@ describe('getStyle', () => {
         });
     });
 
+    describe('getStyleUrl', () => {
+        test('returns null when style is loaded from JSON object', async () => {
```
HarelM: These looks like duplicated tests, I would only keep tests here that are relevant to the changes of the map.ts class as part of this PR.
Seems like setting a styleURL and getting it back is enough as other scenarios are covered by the style.test.ts file (except the added _updateStyleWithSourceUrl logic, which I already reviewed and left my comments on).

src/ui/map.ts
HarelM: See my other comments. Thanks for looking into this.

giswqs:
Thanks for the detailed feedback! I've addressed all the concerns in the latest commit:

1. **Removed `setStyleUrl()` method** - You're right that it was misleading. Now all `_styleUrl` updates happen internally within Style class methods only.

2. **Removed `_updateStyleWithSourceUrl()` helper** - Simplified by adding an optional `sourceUrl` parameter to `_updateStyle()` instead of having a separate method with the "waiting approach".

3. **Added `sourceUrl` parameter to `setState()`** - When diffing succeeds, `_styleUrl` is now updated internally within `setState()` rather than Map reaching into Style's internal state. This keeps all `_styleUrl` management encapsulated within the Style class.

4. **Reduced duplicate tests** - Removed the tests from `map_style.test.ts` that were duplicating what's already covered in `style.test.ts`. Kept only the map-specific tests (setting/getting URL through Map interface).

5. **Added tests for `setState` with `sourceUrl`** - Added proper test coverage in `style.test.ts` to verify the new parameter works correctly.

All tests pass locally. Let me know if there's anything else that needs adjustment!

src/ui/map.ts
```diff
@@ -2166,16 +2166,17 @@ export class Map extends Camera {
         }
     }
 
-    _updateDiff(style: StyleSpecification, options?: StyleSwapOptions & StyleOptions) {
+    _updateDiff(style: StyleSpecification, options?: StyleSwapOptions & StyleOptions, sourceUrl?: string) {
```
HarelM, line 2169: Would it make sense to add this to the options object?

src/ui/map.ts
```diff
@@ -2196,6 +2197,32 @@ export class Map extends Camera {
...
+     * map.setStyle('https://example.com/style.json');
+     * map.getStyleUrl(); // 'https://example.com/style.json'
+     *
+     * // If style was set as an object, returns null
+     * map.setStyle({ version: 8, sources: {}, layers: [] });
+     * map.getStyleUrl(); // null
+     * ```
+     */
+    getStyleUrl(): string | null {
+        if (this.style) {
```
HarelM, line 2220: Use "?." operator.

HarelM:
It is better to answer in the relevant thread and link to the relevant commit, this way I can track which review comment we are talking about...

HarelM:
- This PR is stale and replaced by #8327

PR #7113, opened 2026-02-12
refactor: simplify Hash class with structured parsing helpers

### LINKED ISSUE #7006: Refactor `Hash` to use `URLSearchParams`

pierreeurope:
## Summary

Refactor the `Hash` class to replace manual string building/parsing with structured `_parseHash()` and `_serializeParams()` helper methods.

Closes #7006

## Motivation

As noted in #7006, the Hash class does a lot of manual string building and parsing, making it tricky to extend or maintain. This PR extracts the parsing and serialization into dedicated methods.

## Approach

Rather than using `URLSearchParams` directly (which encodes slashes as `%2F` and colons as `%3A`, breaking the hash value format), this PR introduces two lightweight helpers:

- **`_parseHash()`**: Parses the hash fragment into an `Array<[string, string | null]>` — preserving parameter order, the distinction between `key` (no equals) and `key=` (empty value), and unencoded characters in values
- **`_serializeParams()`**: Serializes the array back to a `&`-separated string

These naturally handle edge cases (trailing/double/leading ampersands) through the split-and-filter approach.

## Changes

- **`getHashString()`**: Uses `_parseHash()` + find/update/push + `_serializeParams()` instead of manual `.split('&').map().filter()`
- **`_getCurrentHash()`**: Uses `_parseHash()` + `.find()` instead of manual `.split('&').map().forEach()`
- **`_removeHash()`**: Uses `_parseHash()` + `.filter()` + `_serializeParams()` instead of manual string replacement

## Tests

All 52 existing tests pass unchanged — no behavior changes.

### DIFF
--- a/src/ui/hash.ts
+++ b/src/ui/hash.ts
@@ -65,41 +65,74 @@ export class Hash {
         if (pitch) hash += (`/${Math.round(pitch)}`);
 
         if (this._hashName) {
-            const hashName = this._hashName;
+            const params = this._parseHash();
             let found = false;
-            const parts = window.location.hash.slice(1).split('&').map(part => {
-                const key = part.split('=')[0];
-                if (key === hashName) {
+            for (let i = 0; i < params.length; i++) {
+                if (params[i][0] === this._hashName) {
+                    params[i] = [this._hashName, hash];
                     found = true;
-                    return `${key}=${hash}`;
+                    break;
                 }
-                return part;
-            }).filter(a => a);
-            if (!found) {
-                parts.push(`${hashName}=${hash}`);
             }
-            return `#${parts.join('&')}`;
+            if (!found) {
+                params.push([this._hashName, hash]);
+            }
+            return `#${this._serializeParams(params)}`;
         }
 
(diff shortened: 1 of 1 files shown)


HarelM:
Why is this PR better than the other that's already open?

pierreeurope:
You're right, I missed #7073 — apologies for the duplicate. That PR takes the correct approach of using actual `URLSearchParams`. Closing this in favor of that one.

PR #7114, opened 2026-02-12
refactor: replace deprecated substr() with startsWith()

pierreeurope:
Replace `String.substr()` with `String.startsWith()` in `program_configuration.ts`.

`substr()` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) and `startsWith()` is more readable and semantic for prefix checking.

**Before:**
```ts
name.substr(0, 9) === 'u_pattern' || name.substr(0, 12) === 'u_dasharray_'
```

**After:**
```ts
name.startsWith('u_pattern') || name.startsWith('u_dasharray_')
```

### DIFF
--- a/src/data/program_configuration.ts
+++ b/src/data/program_configuration.ts
@@ -187,7 +187,7 @@ class CrossFadedConstantBinder implements UniformBinder {
     }
 
     getBinding(context: Context, location: WebGLUniformLocation, name: string): Partial<Uniform<any>> {
-        return (name.substr(0, 9) === 'u_pattern' || name.substr(0, 12) === 'u_dasharray_') ?
+        return (name.startsWith('u_pattern') || name.startsWith('u_dasharray_')) ?
             new Uniform4f(context, location) :
             new Uniform1f(context, location);
     }


src/data/program_configuration.ts
```diff
@@ -187,7 +187,7 @@ class CrossFadedConstantBinder implements UniformBinder {
     }
 
     getBinding(context: Context, location: WebGLUniformLocation, name: string): Partial<Uniform<any>> {
-        return (name.substr(0, 9) === 'u_pattern' || name.substr(0, 12) === 'u_dasharray_') ?
+        return (name.startsWith('u_pattern') || name.startsWith('u_dasharray_')) ?
```
HarelM, line 190: That's not exactly the same logic...

pierreeurope:
Re: @HarelM's inline comment — the logic is equivalent: `name.substr(0, 9) === 'u_pattern'` extracts the first 9 characters and compares to the 9-char string `'u_pattern'`, which is exactly what `startsWith('u_pattern')` does. Same for `substr(0, 12) === 'u_dasharray_'` → `startsWith('u_dasharray_')`.

`startsWith` is cleaner and avoids the deprecated `substr()` method.

HarelM:
Right, sorry, I missed the 0 at the begging, my mistake...

PR #7115, opened 2026-02-12
Fix #7082: Prevent popup close button from receiving automatic keyboard focus

### LINKED ISSUE #7082: Popup close button receives keyboard focus (Tab) when popup auto-opens
Hello MapLibre Team,

I’m facing an accessibility / UX issue with MapLibre GL JS popups and wanted to report it.
Issue summary
When a popup is opened automatically (for example, on map load), the popup’s close (×) button becomes focusable via the Tab key. As a result:
* Pressing `Tab` moves focus directly to the popup close button
* This happens even if the user has not interacted with the map
* The browser may also scroll the page to the map area unexpectedly

This behavior is problematic when the map is embedded below other page content, because keyboard focus jumps into the map UI without user intent.

[html.html](https://github.com/user-attachments/files/25175383/4.html.html)

pierreeurope:
## Description
Fixes #7082

When a popup auto-opens (e.g., via marker click), the close button was receiving automatic keyboard focus, creating unexpected keyboard navigation behavior.

## Changes
- Added `tabindex="-1"` to the popup close button to exclude it from automatic focus
- Updated `focusQuerySelector` to exclude buttons with `tabindex="-1"`
- Updated tests to reflect the new behavior

## Behavior
**Before:**
- Close button received focus when popup opened with no other focusable content
- Pressing Tab would immediately focus the close button

**After:**
- Close button does not receive automatic focus
- Close button remains clickable and accessible via mouse
- Keyboard focus stays on previously focused element or first focusable content element
- Close button is still accessible but not in the automatic tab order

## Testing
All existing popup tests pass with updated expectations.

### DIFF
--- a/src/ui/popup.test.ts
+++ b/src/ui/popup.test.ts
@@ -729,7 +729,7 @@ describe('popup', () => {
         expect(window.document.activeElement).toBe(dummyFocusedEl);
     });
 
-    test('Close button is focused if it is the only focusable element', () => {
+    test('Close button is not automatically focused when popup opens', () => {
         const dummyFocusedEl = window.document.createElement('button');
         window.document.body.appendChild(dummyFocusedEl);
         dummyFocusedEl.focus();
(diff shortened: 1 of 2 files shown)


HarelM:
I don't think the issue was concluded that this bug needs a solution in maplibre and not in the outside code. 
From an accessibility point of view it makes sense to focus on the close button when a popup opens.

pierreeurope:
You're right — the issue doesn't have consensus that this needs a fix in MapLibre, and from an accessibility standpoint focusing the close button makes sense. Closing this.

PR #7116, opened 2026-02-12
Fix #6565: CTRL+Click drag rotation not working on Firefox

### LINKED ISSUE #6565: CTRL+Click Drag rotation not working on firefox
**maplibre-gl-js version**: 5.9.0

**browser**: Firefox

### Steps to Trigger Behavior

 1. Open the Firefox browser (I'm on OSX but I'm not sure that's relevant)
 2. Visit an example in the maplibre docs, e.g. https://maplibre.org/maplibre-gl-js/docs/examples/3d-terrain/ or https://maplibre.org/maplibre-gl-js/docs/examples/add-a-hillshade-layer/
 3. Press the ctrl key and click and drag on the map

### Expected Behavior

Ctrl+click and drag should allow the user to rotate the map. It should also allow them to pitch the map when that functionality is enabled. The DragRotateHandler states that this is supported behavior and this behavior exists on Chrome.

### Actual Behavior

Nothing happens. The map does not pan or rotate or pitch at all.

pierreeurope:
## Description
Fixes #6565

CTRL+Click drag rotation didn't work on Firefox because Firefox reports CTRL+Click as a RIGHT_BUTTON press with ctrlKey set to true, rather than a LEFT_BUTTON press with ctrlKey set.

## Changes
- Removed the `!e.ctrlKey` check from the RIGHT_BUTTON condition in `generateMouseRotationHandler`
- Now accepts both RIGHT_BUTTON clicks (with or without CTRL)
- This allows Firefox's CTRL+Click (reported as RIGHT_BUTTON + ctrlKey) to trigger rotation

## Behavior
**Before:**
- CTRL+Click drag on Firefox did nothing
- Worked fine on Chrome/Safari

**After:**
- CTRL+Click drag works on Firefox for map rotation
- Still works on Chrome/Safari
- Right-click without CTRL also works for rotation

## Testing
Tested the rotation behavior works correctly with the simplified condition.

### DIFF
--- a/src/ui/handler/mouse.ts
+++ b/src/ui/handler/mouse.ts
@@ -62,7 +62,7 @@ export function generateMouseRotationHandler({enable, clickTolerance, aroundCent
     const mouseMoveStateManager = new MouseMoveStateManager({
         checkCorrectEvent: (e: MouseEvent): boolean =>
             (DOM.mouseButton(e) === LEFT_BUTTON && e.ctrlKey) ||
-            (DOM.mouseButton(e) === RIGHT_BUTTON && !e.ctrlKey),
+            (DOM.mouseButton(e) === RIGHT_BUTTON),
     });
     return new DragHandler<DragRotateResult, MouseEvent>({
         clickTolerance,


HarelM:
I think right clicked is used for context menu event, have you tested this fix with other handlers?

pierreeurope:
Fair point about the context menu handler — I haven't tested the interaction with other handlers thoroughly enough to be confident this doesn't break right-click context menu behavior. Closing rather than risk introducing a regression.

PR #7117, opened 2026-02-12
fix: recover from task callback errors to prevent map freeze

### LINKED ISSUE #6093: task_queue.run(...) doesn't handle errors so it gets stuck on currentlyRunning

pierreeurope:
## Summary

Wrap the task execution loop in `TaskQueue.run()` with `try/finally` to ensure `_currentlyRunning` is always reset, even when a task callback throws.

Fixes #6093

## Problem

When a task callback throws an error (e.g., during globe↔mercator projection transitions where `getRayDirectionFromPixel` throws "not implemented yet"), the `_currentlyRunning` flag is never reset to `false`. This causes every subsequent `run()` call to throw:

> Attempting to run(), but is already running.

This permanently freezes the map — no further interactions or renders are possible.

## Fix

Move the cleanup code (`_cleared = false`, `_currentlyRunning = false`) into a `finally` block so it executes regardless of whether a task throws. The error still propagates to the caller, but the queue recovers and can be used again.

## Test

Added a test verifying that after a task throws, the queue accepts and runs new tasks normally.

### DIFF
--- a/src/util/task_queue.test.ts
+++ b/src/util/task_queue.test.ts
@@ -100,6 +100,17 @@ describe('TaskQueue', () => {
         expect(after).toHaveBeenCalledTimes(1);
     });
 
+    test('Recovers from task callback errors and allows subsequent run()', () => {
+        const q = new TaskQueue();
+        q.add(() => { throw new Error('task error'); });
+        expect(() => q.run()).toThrow('task error');
+        // Queue should be recoverable — not stuck in "already running" state
+        const cb = vi.fn();
+        q.add(cb);
+        q.run();
+        expect(cb).toHaveBeenCalledTimes(1);
+    });
+
     test('TaskQueue.clear() interrupts currently-running queue', () => {
         const q = new TaskQueue();
         const before = vi.fn();
--- a/src/util/task_queue.ts
+++ b/src/util/task_queue.ts
@@ -45,14 +45,16 @@ export class TaskQueue {
         // on the next run, not the current run.
         this._queue = [];
 
-        for (const task of queue) {
-            if (task.cancelled) continue;
-            task.callback(timeStamp);
-            if (this._cleared) break;
+        try {
+            for (const task of queue) {
+                if (task.cancelled) continue;
+                task.callback(timeStamp);
+                if (this._cleared) break;
+            }
+        } finally {
+            this._cleared = false;
+            this._currentlyRunning = false;
         }
-
-        this._cleared = false;
-        this._currentlyRunning = false;
     }
 
     clear() {


HarelM:
Thanks for taking the time and interest in maplibre.
This is now the second PR which attempts to solve something an on going PR is .
If this is AI generated, please open one PR at a time so I can review it and provide feedback. Spamming with multiple PRs is not the right approach...

pierreeurope:
You're right, and I apologize for the spam — I should have gone one at a time. I've just closed 4 of the weaker PRs where I wasn't confident enough in the fix.

Regarding this PR specifically: the task queue fix is a straightforward try/finally that ensures `_currentlyRunning` resets after an error. It doesn't change the behavior of any other code path — the error still propagates, but the queue becomes usable again instead of permanently locking up. The test covers this exact scenario.

I'll slow down and focus on quality going forward.

pierreeurope:
Closing - didn't realize there was already an ongoing PR for this. Apologies for the noise.

PR #7118, opened 2026-02-12
Fix #6765: Map.setPadding no longer interrupts ongoing flyTo animations

### LINKED ISSUE #6765: Map.flyTo animation stops when calling Map.setPadding
maplibre-gl-js version 5.13.0
Firefox

### Steps to Trigger Behavior

 1. call ```map.flyTo({center:[7, 47]});```
 2. during animation call ```map.setPadding({top: 10, bottom: 10, left: 10, right: 10})```

### Link to Demonstration
https://jsbin.com/wimuwafafa/edit?html,output

### Expected Behavior
I would expect the animation to finish and the map to be centered at [7,47] in the end.
Padding could be applied immediately or after the animation.

### Actual Behavior
The animation abruptly stops. Center of the map depends on timing of the setPadding call.

pierreeurope:
## Description
Fixes #6765 

This PR modifies the behavior of `Map.setPadding()` to no longer interrupt ongoing camera animations like `flyTo()`.

## Changes
- Modified `Camera.setPadding()` to update padding directly on the transform without calling `jumpTo()`
- The method now checks if an animation is in progress using `isEasing()` before firing movestart/moveend events
- Padding updates are applied immediately but no longer stop ongoing animations
- Added test case to verify that flyTo animations continue when setPadding is called

## Before
When `setPadding()` was called during a `flyTo()` animation, it would call `jumpTo()`, which in turn calls `stop()`, abruptly ending the animation.

## After
`setPadding()` now applies padding changes directly without interrupting animations. The padding update is immediate, and the flyTo animation completes as expected.

## Testing
- All existing camera tests pass
- New test added to verify flyTo continues after setPadding

## Breaking Changes
None. This is a bug fix that makes behavior more intuitive. Users who want the old behavior (stopping animations) can call `map.stop()` before `map.setPadding()`.

### DIFF
--- a/src/ui/camera.test.ts
+++ b/src/ui/camera.test.ts
@@ -572,6 +572,22 @@ describe('setPadding', () => {
         expect(currentPadding.right).toBe(padding1.right);
         expect(currentPadding.bottom).toBe(padding.bottom);
     });
+
+    test('does not interrupt ongoing flyTo animation', () => {
+        const camera = createCamera();
+        
+        camera.flyTo({center: [10, 10], duration: 100});
+        expect(camera.isEasing()).toBe(true);
+        
+        // Call setPadding during the animation - should not stop it
+        camera.setPadding({left: 100, top: 50, right: 100, bottom: 50});
+        
+        // Animation should still be running (not interrupted)
+        expect(camera.isEasing()).toBe(true);
+        
+        // Padding should be applied immediately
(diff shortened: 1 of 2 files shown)


HarelM:
The padding logic is very complicated, how confident are you that this doesn't break anything else?

pierreeurope:
You're right that the padding logic is complicated. I'm not confident enough that this doesn't break other padding scenarios. Closing this — it needs more thorough investigation and end-to-end testing than I've done.

PR #7119, opened 2026-02-12
Fix #6744: Prevent padding accumulation between flyTo and fitBounds

### LINKED ISSUE #6744: Map.flyTo padding not being cleared -> fitBounds is summing it up with it's own padding argument.
maplibre-gl-js version: 5.13.0

browser: chrome 141.0.7390.77

I think there were few issues about this already, but I don't see any open one, yet still this bug exists so creating new one.

Related issues:
https://github.com/mapbox/mapbox-gl-js/issues/11831
https://github.com/maplibre/maplibre-gl-js/issues/4095 - this was closed, but not resolved
https://github.com/maplibre/maplibre-gl-js/discussions/2984 -> there is no solution to my knowledge, mapbox says that they fixed it somehow

### Steps to Trigger Behavior

Steps to Trigger Behavior
1. Call function map.FlyTo with padding ->
```
 map.flyTo({
   center: coordinates[0],
   padding: {
      top: 50,
      bottom: 0,
      left: 0,
      right: 0
   }
})
```
2. Call function map.fitBounds ->
```
map.fitBounds(bounds, {
 padding: 0
});
```

### Link to Demonstration

https://jsbin.com/pifejixubo/edit?html,output

https://github.com/user-attachments/assets/5c240d19-9205-4c16-822c-e9f3c400eb92

### Expected Behavior

Padding is not summed up

### Actual Behavior

Padding is summed up which breaks UI of the app

pierreeurope:
## Description
Fixes #6744

This PR fixes a bug where padding from `flyTo()` would persist and be added to the padding specified in subsequent `fitBounds()` calls, causing incorrect zoom calculations.

## Problem
When calling `flyTo()` with padding, that padding is set on the transform and persists. When `fitBounds()` is then called with its own padding, the `cameraForBoxAndBearing` function was summing both paddings together:

```typescript
const edgePadding = tr.padding;  // Old padding from flyTo
const availableWidth = (tr.width - (edgePadding.left + edgePadding.right + padding.left + padding.right));
const availableHeight = (tr.height - (edgePadding.top + edgePadding.bottom + padding.top + padding.bottom));
```

This resulted in incorrect camera calculations and broken UI behavior.

## Solution
Removed the addition of `edgePadding` (the transform's current padding) in the `cameraForBoxAndBearing` calculation. Now only the explicitly requested padding parameter is used:

```typescript
const availableWidth = (tr.width - (padding.left + padding.right));
const availableHeight = (tr.height - (padding.top + padding.bottom));
```

## Changes
- Modified `cameraForBoxAndBearing()` in `src/geo/projection/camera_helper.ts`
- Removed the `edgePadding` variable and its usage in width/height calculations
- Added comprehensive test to verify padding is not accumulated between animations

## Testing
- All existing camera and fitBounds tests pass
- New test verifies that fitBounds produces the same result regardless of previous flyTo padding

## Breaking Changes
None. This is a bug fix that restores the expected behavior.

### DIFF
--- a/src/geo/projection/camera_helper.ts
+++ b/src/geo/projection/camera_helper.ts
@@ -154,8 +154,6 @@ export function updateRotation(args: UpdateRotationArgs) {
 }
 
 export function cameraForBoxAndBearing(options: CameraForBoundsOptions, padding: PaddingOptions, bounds: LngLatBounds, bearing: number, tr: IReadonlyTransform): CameraForBoxAndBearingHandlerResult {
-    const edgePadding = tr.padding;
-
     // Consider all corners of the rotated bounding box derived from the given points
     // when find the camera position that fits the given points.
 
@@ -184,8 +182,8 @@ export function cameraForBoxAndBearing(options: CameraForBoundsOptions, padding:
     // Calculate zoom: consider the original bbox and padding.
     const size = upperRight.sub(lowerLeft);
 
-    const availableWidth = (tr.width - (edgePadding.left + edgePadding.right + padding.left + padding.right));
-    const availableHeight = (tr.height - (edgePadding.top + edgePadding.bottom + padding.top + padding.bottom));
+    const availableWidth = (tr.width - (padding.left + padding.right));
+    const availableHeight = (tr.height - (padding.top + padding.bottom));
     const scaleX = availableWidth / size.x;
     const scaleY = availableHeight / size.y;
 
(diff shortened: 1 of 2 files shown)


HarelM:
This logic is complicated and a lot of people have tried to solve it in the past, how confident are you that this will solve this issue? 
What test did you do to make sure it doesn't break other padding scenarios?

pierreeurope:
Same as #7118 — the padding system has too many interacting parts for me to be confident this naive fix doesn't break legitimate use cases (e.g., persistent viewport padding set via setPadding() should still affect fitBounds). Closing.

PR #7126, opened 2026-02-13
Fix texture corruption after WebGL context loss (#2811)

### LINKED ISSUE #2811: Map sometimes renders custom images or font characters instead of basemap tiles
We have recently seen very intermittent, odd behavior of the base map layer showing up as a set of our internal icons tiled across the map or a set of letters tiled across the map. 

In one case, it appears a base map layer has partially loaded with these icons overlayed on top of it.

**Here is my best attempt at a text description of the symptoms. See the video below (https://github.com/maplibre/maplibre-gl-js/issues/2811#issuecomment-2249318129) for a great example.**

* Light gray (“white”) basemap tiles, often with other non-basemap layers still visible  
  * In particular, GeoJSON source layers don’t *seem* to be affected  
* Sometimes basemap streets, some place names, or highway badges are still visible  
* Sometimes basemap tiles are replaced by font glyph texture  
* Sometimes basemap tiles are replaced by potpacked user-added icons texture (from calls to `Map.addImage()`)  
* Symptoms above can change when panning around the map or zooming in/out  
  * Results in flickering/flashing of map tiles  
  * A given tile can display any of the above incorrect textures  
* A given symptom *seems* to be different for every map tile, although the flickering behavior will show within a single integer zoom level while zooming  
  * Crossing a zoom level *seems to* “change the symptom”: may go from glyph texture at one zoom level to gray, or user-added icons texture, at the next zoom level  
* Trigger *seems* to happen right after bringing the app into focus again after it’s been running in the background. However, it has happened quickly after app launch  
* Have confirmed instances with the exact same symptoms (including video evidence/photos on some) for:
  - iPhone app (Capacitor which has a WebView in it aka. Safari) (countless users)
  - iPhone Safari Mobile (a few users)
  - Windows running Chrome (2 users)
  - Windows running Firefox (1 user)
* It seems as though the 2d canvas context that contains the style sprites is cleared when this bug happens, and otherwise appears to be populated. Not knowing the architecture, I’m not sure if this is unexpected behavior or normal
* A little different every time

** Helpful images / videos **

Screenshot 2023-06-27 at 4 50 04 PM

And in the second, it appears to just be the set of english characters tiled.
Screenshot 2023-06-27 at 4 50 31 PM

image
image
image
image

I have caught the bug twice on a dev device, but did not catch the transition from working -> broken while attached and using developer tools. Once the map experience is broken, there are no console logs.

I am using:

maplibre-gl: 3.1.0
and 
react-map-gl: ^7.0.25

_Originally posted by @davidemerritt in https://github.com/maplibre/maplibre-gl-js/discussions/2738_

birkskyum:
When the browser reclaims GPU resources (e.g. tab backgrounded on mobile), WebGL textures can get mixed up.  Raster map tiles end up displaying glyph atlas characters or icon sprites instead of the actual imagery.

This adds a guard in `Texture.bind()` that checks whether the underlying WebGL handle still matches the one originally created for this texture, and restores it if not.

Includes a unit test that verifies texture handles are correctly restored.

Fixes #2811
- #2811

#### Render Test 
Before
actual

After

expected

## Launch Checklist

 - [X] Link to related issues.
 - [ ] Post benchmark scores.


test/integration/render/run_render_tests.ts
```diff
@@ -677,6 +677,24 @@ async function getImageFromStyle(styleForTest: StyleWithTestData, page: Page): P
                     case 'pauseTiles':
                         map.style.tileManagers[operation[1]].pause();
                         break;
+                    case 'corruptTextures': {
```
HarelM: Other connect lost test are done in the browsers tests, I prefer to keep render test as simple as possible.
Or maybe a unit test would be enough here...

HarelM:
Nice work!! Added a minor comment about the render test.

test/integration/render/run_render_tests.ts
birkskyum: I've added a unit test instead - it does the job fine. The render test helped me verify it was in fact a proper reproduction because i wanted to visually see that this is the same, but now that we know a unit test can sufficiently check.

src/render/texture.ts
```diff
@@ -87,6 +91,12 @@ export class Texture {
     bind(filter: TextureFilter, wrap: TextureWrap, minFilter?: TextureFilter | null) {
         const {context} = this;
         const {gl} = context;
+
+        // #2811: restore handle if it was corrupted after context loss
```
HarelM: I think the text can be removed, the field has this information in a tsdoc comment, which is better.

HarelM (approved):
Added some nitpicking comment, but otherwise approved.

test/integration/render/run_render_tests.ts
HarelM: Makes sense. Thanks!

PR #7128, opened 2026-02-13
Fix Firefox "Alpha-premult deprecated for non-DOM uploads" warning (#2030)

### LINKED ISSUE #2030: Firefox WebGL warning (Alpha-premult and y-flip are deprecated for non-DOM-Element uploads).

birkskyum:
Firefox logs a warning every time we ask WebGL to handle alpha premultiplication for raw pixel data (sprite atlases, color ramps, etc). It only wants us to use that WebGL flag for actual DOM elements like images and canvases For raw byte arrays, it expects us to do it ourselves.

This moves premultiplication into JS for raw data uploads, keeping the WebGL path for DOM elements.

Fixes #2030
- #2030

Before

Screenshot 2026-02-13 at 21 46 45

After
Screenshot 2026-02-13 at 21 44 20

## Launch Checklist

 - [X] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/render/texture.ts
+++ b/src/render/texture.ts
@@ -1,7 +1,6 @@
 import type {Context} from '../gl/context';
 import type {RGBAImage, AlphaImage} from '../util/image';
 import {premultiplyAlpha} from '../util/image';
-import {isImageBitmap} from '../util/util';
 
 export type TextureFormat = WebGLRenderingContextBase['RGBA'] | WebGLRenderingContextBase['ALPHA'];
 export type TextureFilter = WebGLRenderingContextBase['LINEAR'] | WebGLRenderingContextBase['LINEAR_MIPMAP_NEAREST'] | WebGLRenderingContextBase['NEAREST'];
@@ -15,6 +14,10 @@ type EmptyImage = {
 
 type DataTextureImage = RGBAImage | AlphaImage | EmptyImage;
 export type TextureImage = TexImageSource | DataTextureImage;
+
+function hasDataProperty(image: TextureImage): image is DataTextureImage {
+    return 'data' in image;
+}
 
 /**
  * @internal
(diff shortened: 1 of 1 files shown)


src/render/texture.ts
```diff
@@ -59,23 +60,32 @@ export class Texture {
...
 
         if (resize) {
             this.size = [width, height];
 
-            if (image instanceof HTMLImageElement || image instanceof HTMLCanvasElement || image instanceof HTMLVideoElement || image instanceof ImageData || isImageBitmap(image)) {
+            if (isDomSource) {
                 gl.texImage2D(gl.TEXTURE_2D, 0, this.format, this.format, gl.UNSIGNED_BYTE, image);
             } else {
-                gl.texImage2D(gl.TEXTURE_2D, 0, this.format, width, height, 0, this.format, gl.UNSIGNED_BYTE, (image as DataTextureImage).data);
+                let data = (image as DataTextureImage).data;
```
HarelM: Is it possible to avoid this casting somehow? Maybe invert the if?

birkskyum: We can use a "hasDataProperty" type guard instead, then there's no untyped cast.

src/render/texture.test.ts
```diff
@@ -51,4 +51,14 @@ describe('Texture', () => {
         texture.bind(gl.LINEAR, gl.CLAMP_TO_EDGE);
         expect(texture.texture).toBe(originalHandle);
     });
+
+    test('premultiplyAlpha produces correct output', () => {
```
HarelM, line 55: This test needs to be next to its method, in image.test.ts or something...

HarelM:
Thanks for this PR!
How confident are you about the performance of this? While I understand it might be good to resolve a Firefox warning, if this is worse for chrome, which is a lot more popular, I'm not sure it's worth it. What do you think?

birkskyum:
As far as I can tell it should be fine. Firefox only optimizes the CPU path we're now using (JS premult), so they benefit. For Chrome, I ran the Paint benchmark comparing main vs this branch:

Screenshot 2026-02-14 at 15 03 05]


This branch is actually 3.4% faster (9.50ms vs 9.83ms) - ran it again and saw 2.9% improvement


Why minimal impact:

- Texture uploads happen rarely (just at map load and style changes, not every frame)
- We're processing tiny amounts of data: line gradients are 1KB, sprite atlases load once
- Raster tiles (biggest perf factor) use images/canvas and aren't affected

PR #7142, opened 2026-02-17
fix: setPadding no longer interrupts ongoing flyTo animations (#6765)

### LINKED ISSUE #6765: Map.flyTo animation stops when calling Map.setPadding

pierreeurope:
## Problem

When `map.setPadding()` is called while a `flyTo` (or other camera animation) is in progress, the animation abruptly stops mid-flight. This happens because `setPadding` delegates to `jumpTo`, which unconditionally calls `this.stop()`.

Fixes #6765.

## Root Cause

```ts
// camera.ts — before this fix
setPadding(padding: PaddingOptions, eventData?: any): this {
    this.jumpTo({padding}, eventData);  // jumpTo always calls this.stop() first
    return this;
}
```

## Solution

Rewrite `setPadding` to apply padding directly to the transform without touching the animation state:

- If an animation **is in progress**: padding is applied immediately to the current transform, `this._padding` is not reset, and only a `move` event fires (no `movestart`/`moveend` — those belong to the animation lifecycle).
- If **no animation** is in progress: `movestart`, `move`, and `moveend` fire as before.

The ongoing `flyTo` center/zoom animation continues uninterrupted.

## Changes

- `src/ui/camera.ts`: Rewrite `setPadding` to avoid calling `jumpTo` and thus avoid `stop()`.
- `src/ui/camera.test.ts`: Add test asserting `isEasing()` remains `true` after `setPadding` is called during a `flyTo`.

## Testing

```bash
npm run lint                                         # clean
npx vitest run --config vitest.config.unit.ts src/ui/camera.test.ts  # 293 passed
```

The new test (`does not interrupt ongoing flyTo animation`) passes alongside all existing `setPadding` and `flyTo` tests.


HarelM:
That's not an appropriate way to conduct in this open source space.
The following PR was closed and there's no mention of it here.
- https://github.com/maplibre/maplibre-gl-js/pull/7118
Also the PR template is completely ignored and there are missing things that are required there.
I've also asked to open a single PR at a time.
I'm now assuming this is AI generated and closing all PRs related to this account.
If this is not the case you are welcome to join the next monthly meeting and appeal this decision.

PR #7143, opened 2026-02-18
fix: allow map communication in srcdoc iframes with null origin

### LINKED ISSUE #7047: Map does not render in an iframe
**maplibre-gl-js version**: 3.6.2+

**browser**: Any

### Steps to Trigger Behavior

 1. Create an HTML page with an iframe
 2. Add the map element inside the iframe
 3. Open the HTML page

### Link to Demonstration

Copy the below HTML code into https://jsbin.com to reproduce the issue.

[jsbin.xacadatayu.2.html](https://github.com/user-attachments/files/25024683/jsbin.xacadatayu.2.html)

### Expected Behavior

The map embedded in the iframe renders.

### Actual Behavior

The map embedded in the iframe does not render.

pierreeurope:
## Problem

When MapLibre GL JS is embedded inside an `<iframe srcdoc="...">" element, browsers assign the [opaque origin](https://html.spec.whatwg.org/multipage/origin.html#concept-origin-opaque) `'null'` (the string) to that document. The `receive()` method in `actor.ts` checks that the message sender's origin matches the receiver's origin before processing messages. This check fails when one side is a null-origin srcdoc iframe and the other is a worker created from a CDN-hosted script (which retains the CDN's real origin), causing the map to silently fail to render.

This was a regression introduced in #3233 (the actor callback→promise migration), noted in #7047.

## Root Cause

`actor.ts` line ~150:

```ts
if (data.origin !== 'file://' && location.origin !== 'file://' &&
    data.origin !== 'resource://android' && location.origin !== 'resource://android' &&
    data.origin !== location.origin) {
    return;
}
```

`file://` and `resource://android` are already exempted because those environments cannot provide a meaningful same-origin guarantee. The `'null'` opaque origin in srcdoc iframes is in exactly the same category.

## Fix

Extend the allowed-origin list with `'null'` so that maps hosted inside srcdoc iframes can communicate with their workers. The implementation extracts the check into a small helper for readability:

```ts
const allowedOrigin = (origin: string) =>
    origin === 'file://' || origin === 'resource://android' || origin === 'null';
if (!allowedOrigin(data.origin) && !allowedOrigin(location.origin) && data.origin !== location.origin) {
    return;
}
```

## Tests

Added a new unit test to `src/util/actor.test.ts`:

> **should process a message when sender origin is "null" (srcdoc iframe)**

All **16** actor unit tests pass.

## Checklist

- [x] I've read the [contribution guidelines](https://github.com/maplibre/maplibre-gl-js/blob/main/CONTRIBUTING.md)
- [x] Unit test added for the new behaviour
- [x] No backport from Mapbox proprietary code

Fixes #7047

### DIFF
--- a/src/util/actor.test.ts
+++ b/src/util/actor.test.ts
@@ -277,4 +277,21 @@ describe('Actor', () => {
 
         expect(spy).toHaveBeenCalled();
     });
+
+    test('should process a message when sender origin is "null" (srcdoc iframe)', async () => {
+        // Browsers assign a null origin to srcdoc iframes. When MapLibre is hosted inside
+        // such an iframe the worker's origin may differ from the iframe's 'null' origin.
+        // We must allow communication in this case, similar to file:// environments.
+        const worker = workerFactory() as any as WorkerGlobalScopeInterface & ActorTarget;
+        const actor = new Actor(worker, '1');
+
+        const spy = vi.fn().mockReturnValue(Promise.resolve({}));
+        worker.worker.actor.registerMessageHandler(MessageType.getClusterExpansionZoom, spy);
+
+        actor.target.postMessage({type: MessageType.getClusterExpansionZoom, data: {} as any, origin: 'null'});
+
+        await sleep(0);
+
+        expect(spy).toHaveBeenCalled();
+    });
 });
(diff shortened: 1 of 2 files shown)


HarelM:
This is believed to be AI generated and is missing the PR template.
The discussion suggest to close the issue and not fix it.
Closing this as spam.

PR #7150, opened 2026-02-20
Add optional boxZoomEnd callback for custom Shift-drag behavior

### LINKED ISSUE #6397: Add an option for customizing box zoom action
## User Story

Map apps sometimes need a way to select multiple features on a map, and shift-drag is a convenient handler for this. By default, it is used for zooming to a selected portion of the map, but it would be nice to have a way to customize the control so that it would do something else instead of zooming.

## Rationale

Reimplementing this kind of selection logic requires a lot of code boilerplate, and it's a relatively common scenario. Adding an option to `boxZoom` handler is on the other hand quite easy (a 1-2 lines of code change).

## Alternatives

A current workaround is to hack into the `boxZoom` instance and force it to return a different handler function on mouseup, but this is obviously fragile:

```js
map.boxZoom.mouseupWindow = function (e, p1) {
    const p0 = this._startPos;
    if (map.boxZoom.constructor.prototype.mouseupWindow.call(this, e, p1)) return {
        cameraAnimation() {
            // do something else instead of zooming
            console.log(p0, p1);
        }
    }
};
```

itisyb:
Fixes #6397.

This adds a small extension point to box zoom so Shift-drag can be reused for custom actions (for example, feature selection) without overriding internal methods.

### What changed

- Added `BoxZoomHandlerOptions` in `src/ui/handler/box_zoom.ts`.
- Added optional `boxZoomEnd` callback:
  - `(map, startPos, endPos, originalEvent) => BoxZoomEndResult | void`
  - returning `cameraAnimation` overrides the default `fitScreenCoordinates` animation.
- Updated `MapOptions.boxZoom` to `boolean | BoxZoomHandlerOptions` in `src/ui/map.ts`.
- Passed `options.boxZoom` into `boxZoom.enable(...)` in `src/ui/handler_manager.ts`.
- Kept default behavior unchanged when no callback is provided.

### Tests

Added unit tests in `src/ui/handler/box_zoom.test.ts` for:

- callback invocation on box-zoom end
- skipping default zoom when callback is set
- running custom `cameraAnimation` returned by callback
- preserving default zoom when `boxZoom: {}` has no callback

### Verification

- `npm run lint -- src/ui/handler/box_zoom.ts src/ui/map.ts src/ui/handler_manager.ts src/ui/handler/box_zoom.test.ts`
- `npm run typecheck`
- `npm run test-unit`
- `npm run build-dev && npm run build-prod && npm run test-build`
- `npm run build-css && npm run test-integration`


HarelM:
I find the API a bit cryptic to understand. 
Can you also add to this PR a usage example (in the form of the existing examples in this repo) to better showcase how this can and should be used?

itisyb:
using a supported hook (boxZoomEnd) instead of overriding internals.
Default behavior still works exactly the same with boxZoom: true

So now we can do something like custome hooks to select those dots like mentioned instead of zooming 

Default will normally Zoom and if we add 

Before/after screenshots
image
image

src/ui/handler/box_zoom.ts
```diff
@@ -57,12 +84,16 @@ export class BoxZoomHandler implements Handler {
...
      * Enables the "box zoom" interaction.
      *
+     * @param options - Options object.
      * @example
      * ```ts
      * map.boxZoom.enable();
      * ```
      */
-    enable() {
+    enable(options?: BoxZoomHandlerOptions | boolean) {
```
HarelM: Why use the enable method and not the constructor? The constructor already has some kind of options object...

src/ui/handler/box_zoom.ts
```diff
@@ -7,6 +7,32 @@ import type {Map} from '../map';
 import type Point from '@mapbox/point-geometry';
 import {type Handler} from '../handler_manager';
 
+/**
+ * Result type for a custom box zoom end callback.
+ */
+export type BoxZoomEndResult = {
+    /**
+     * A custom camera animation to run instead of the default fit-to-box behavior.
```
HarelM: The description here is not informative enough, and it should not return "any".

src/ui/handler/box_zoom.test.ts
```diff
@@ -3,9 +3,15 @@ import {Map} from '../map';
...
-function createMap(clickTolerance) {
-    return new Map({style: '', container: DOM.create('div', '', window.document.body), clickTolerance});
+import type {BoxZoomHandlerOptions} from './box_zoom';
+
+function createMap(clickTolerance, boxZoom?: boolean | BoxZoomHandlerOptions) {
+    return new Map({
+        style: '',
+        container: DOM.create('div', '', window.document.body),
+        clickTolerance,
+        ...(boxZoom === undefined ? {} : {boxZoom})
```
HarelM: I'm guessing this line can look simpler...?

src/ui/handler/box_zoom.test.ts
```diff
@@ -40,6 +46,71 @@ describe('BoxZoomHandler', () => {
         map.remove();
     });
 
+    test('runs custom box zoom end callback instead of default zoom animation', () => {
```
HarelM: I think this test is redundant given the next test...

src/ui/handler_manager.ts
```diff
@@ -78,6 +78,11 @@ export interface Handler {
     readonly renderFrame?: () => HandlerResult | void;
 }
 
+/**
+ * Callback that performs a one-off camera animation on the map.
```
HarelM: An example of such a method would be helpful, as it's not clear what a void returning method should do.

itisyb: Should I add example in comment or a separate file,
also that dots example i created would that work

src/ui/handler/box_zoom.ts
```diff
@@ -5,7 +5,34 @@ import {TransformProvider} from './transform-provider';
...
+
+/**
+ * Result returned by {@link BoxZoomEndHandler}.
+ * Return `void` to skip the default zoom behavior, or return `cameraAnimation` to run a custom camera action.
+ */
+export type BoxZoomEndResult = {
+    /**
+     * A custom camera action to run after `boxzoomend`.
+     */
+    cameraAnimation?: CameraAnimationCallback;
```
HarelM: What's the added value of returning this and not just simply do the animation as part of the zoom end method?
I'm a bit confused...
I understand there's a need to return a camera animation method, but why not when overidering boxzoomend return en "empty" method (`() => {}`) and assume all the relevant code is running inside this method?

itisyb: I followed the same pattern we already use in click_zoom: handlers return a HandlerResult, and HandlerManager runs camera side effects in one place. That keeps behavior consistent with the rest of the input pipeline.
If you don’t need camera movement, boxZoomEnd can run your custom logic directly and just return void

HarelM: Is the click_zoom part of the public API? I can't find it.
I'm asking because returning a cameraAnimation callback is very implementation specific and I'm not sure the user should bother with internal implementation details.
Instead getting a callback and doing whatever you want and returning void would make more sense to me, in terms of public API. But if retuning void will not allow doing some camera animation/movement then that's not a good enough API.
This is why I'm asking if retuning `cameraAnimation` is a must or not.
According to what I understand from you answer, the user can take care of it inside the handler method, so returning void would be better I think...

itisyb: make sense,
updated boxZoomEnd to a plain void callback so users can run custom logic directly inside the callback
if no boxZoomEnd is provided.. Shift drag still does the normal fit-to-box zoom.

src/ui/handler_manager.ts
HarelM: Since this is not exported as part of this PR making the docs here better is of less priority, but a small example of a camera animation callback might be interesting to add here - just 1-5 lines of code, nothing big.

Regarding the "interactive" example of selecting features, I think it would be great to add it as part of this PR and this feature.

Check out the examples folder under the tests to see all the examples that later on appear in maplibre.org docs pages.

HarelM:
Let me know what you decide about the example page, I think it can be a good addition, but this can also be merged as is, your call.

HarelM:
Looks great, thanks! 
Can you simplify the example somehow? Maybe reduce some css styling? Use an existing geojson or move the features to a new geojson file? 
It's currently ~150 lines, but I'd like it to be shorter if possible, hopefully half of the current size.

itisyb:
> Looks great, thanks! Can you simplify the example somehow? Maybe reduce some css styling? Use an existing geojson or move the features to a new geojson file? It's currently ~150 lines, but I'd like it to be shorter if possible, hopefully half of the current size.

I have made it super simplified... also removed select highlighting let me know if that's required

HarelM (approved):

HarelM (approved):

itisyb:
I think bundle min size is failing so I have increased it let me know if this is a wrong approach @HarelM

PR #7165, opened 2026-02-24
Fix: Allow opaque ("null") origins in Actor message filtering

### LINKED ISSUE #7047: Map does not render in an iframe

pcardinal:
## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

**Description**

- Fix src\util\actor.ts to accept opaque origins ("null") alongside file:// and resource://android.
- Add an explanatory comment near the origin check.

https://github.com/pcardinal/maplibre-gl-js/blob/7bafcde7a3c4f02c5480bedd1c2049e9581e831a/src/util/actor.ts#L150-L152

- Add a unit test covering the "null" origin case.

https://github.com/pcardinal/maplibre-gl-js/blob/7bafcde7a3c4f02c5480bedd1c2049e9581e831a/src/util/actor.test.ts#L253-L265

**Issue**
Fixes #7047

The introduced change (migration “Async Actor”, PR https://github.com/maplibre/maplibre-gl-js/pull/3233, related to security https://github.com/maplibre/maplibre-gl-js/issues/3239) adds a strict origin check in receive():
if (... && data.origin !== location.origin) return;

In a srcdoc iframe, the page origin is often null (opaque), while the MapLibre worker can have a different origin (e.g., a CDN).
The result in v5.0.0: worker messages are rejected, so the map doesn't initialize.

**Tests**

 npm run test-unit -- --run src/util/actor.test.ts


src/util/actor.ts
```diff
@@ -147,7 +147,9 @@ export class Actor implements IActor {
     receive(message: {data: MessageData}) {
         const data = message.data;
         const id = data.id;
-        if (data.origin !== 'file://' && location.origin !== 'file://' && data.origin !== 'resource://android' && location.origin !== 'resource://android' && data.origin !== location.origin) {
+        const isOpaqueOriginCase = data.origin === 'null' || location.origin === 'null';
```
HarelM: Are you sure about the strict comparison to string (`=== 'null'`)?

src/util/actor.ts
```diff
@@ -147,7 +147,9 @@ export class Actor implements IActor {
...
+        // Ignore cross-origin messages except for opaque origins and known file/resource cases.
+        if (data.origin !== 'file://' && location.origin !== 'file://' && data.origin !== 'resource://android' && location.origin !== 'resource://android' && data.origin !== location.origin && !isOpaqueOriginCase) {
```
HarelM: Looks like the checks in this if are somewhat duplicated.
I would recommend trying to generalize it and there's a lot of duplicated code here which looks very similar. Maybe define a "allow" or "un-allow" list and check in a for loop, or `filter`/`some`?

HarelM:

src/util/actor.ts
pcardinal: Yes, I'm quite certain of that.
I ran tests using the maplibre-gl-dev.js file, in which I added lines of code using console.log()


  ```
        console.log('***location.origin');
        console.log('value =', location.origin);
        console.log('typeof =', typeof location.origin);
        console.log('=== null ?', location.origin === null);
        console.log("=== 'null' ?", location.origin === 'null');
        console.log('***data.origin');
        console.log('value =', data.origin);
        console.log('typeof =', typeof data.origin);
        console.log('=== null ?', data.origin === null);
        console.log("=== 'null' ?", data.origin === 'null');
```



I have the following results when the code is running in a worker 

***location.origin 
value = http://localhost:5500 
typeof = string 
=== null ? false 
=== 'null' ? false 

***data.origin 
value = null 
**typeof = string** 	&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;<= data.origin is a string
=== null ? false 
**=== 'null' ? true**        &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;<= data.origin is 'null'




I have the following results when the code is running in actor.ts (out of worker)

***location.origin
value = null
**typeof = string**          &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; <= location.origin is a string
=== null ? false
**=== 'null' ? true**        &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;<= location.origin is 'null'

***data.origin
value = http://localhost:5500
typeof = string
=== null ? false
=== 'null' ? false

Also according to the Web specification, for an opaque origin (sandbox, data:, certain contexts), `origin` is serialized as a "null" string, not as a null value.
Therefore, `=== 'null'` is the correct test.

File: src/util/actor.ts (hunk 2 above)
Comment (pcardinal) [reply src/util/actor.ts]: Is is better and acceptable for you ?

```
    const SPECIAL_ORIGINS = new Set(['file://', 'resource://android']);
    const isSpecialOrigin = (origin: string) => SPECIAL_ORIGINS.has(origin);
    const isOpaqueOriginCase = data.origin === 'null' || location.origin === 'null';
    const isSameOrigin = data.origin === location.origin;
    const isKnownSpecialCase = isSpecialOrigin(data.origin) || isSpecialOrigin(location.origin);

    // Ignore cross-origin messages except for opaque origins and known file/resource cases.
    if (!isSameOrigin && !isKnownSpecialCase && !isOpaqueOriginCase) {
        return;
    }
```

I refactored the origin guard to remove duplicated checks by extracting isSameOrigin, isKnownSpecialCase, and isOpaqueOriginCase predicates. Behavior is unchanged, but readability/maintainability is better.

Comment (HarelM) [reply src/util/actor.ts]: I think array can be used instead of set to simplify things, opaque can be treated as special origin as well.
And I think you should be able to loop on the array and check data and location.
I hope it can reduce the amount of code but I'm not sure...

Comment (pcardinal) [reply src/util/actor.ts]: It should be better like that...

```
    const SPECIAL_ORIGINS = ['file://', 'resource://android', 'null'];
    const origins = [data.origin, location.origin];

    const isSameOrigin = data.origin === location.origin;
    const hasSpecialOrigin = origins.some((origin) => SPECIAL_ORIGINS.includes(origin));

    // Ignore cross-origin messages except for special origins.
    if (!isSameOrigin && !hasSpecialOrigin) {
        return;
    }
```

HarelM: Yup, something like that.

HarelM (approved):

PR #7172, opened 2026-02-25
Add updateable support for GeoJSON-VT

wayofthefuture:
In progress...

<br>

 - [ ] Post benchmark scores.

### DIFF
--- a/src/source/geojson_source.test.ts
+++ b/src/source/geojson_source.test.ts
@@ -8,7 +8,7 @@ import {extend} from '../util/util';
 import {SubdivisionGranularitySetting} from '../render/subdivision_granularity_settings';
 import {MercatorTransform} from '../geo/projection/mercator_transform';
 import {sleep, waitForEvent} from '../util/test/util';
-import {type ActorMessage, ClusterIDAndSource, GeoJSONWorkerSourceLoadDataResult, MessageType} from '../util/actor_messages';
+import {type ActorMessage, type ClusterIDAndSource, type GeoJSONWorkerSourceLoadDataResult, MessageType} from '../util/actor_messages';
 import type {IReadonlyTransform} from '../geo/transform_interface';
 import type {Dispatcher} from '../util/dispatcher';
 import type {RequestManager} from '../util/request_manager';


wayofthefuture:
@HarelM It looks sooooo much better...

HarelM:
Ok, only one test is failing now.
I'm able to see the difference in terms of slightly different rendering.
When I tired to use the old suplercluster library the test passes.
When I debugged it a bit I see that getTile for 0,0,0 is returning a different number of features (61 vs 58) which explains the rendering diff.
So there's a need to debug this further and see what changed.
I'm guessing it is related to the changes I made, might be the dead code part that I removed.
I think it should be fairly easy to reproduce this in a geojson-vt test.
I'll see if I can work on that next week.

wayofthefuture:
@HarelM I copied places.json into geojson-vt and old supercluster lib and compared the 0,0,0 tile generated:
```ts
test('generates clusters properly', () => {
    const index = new ClusterTileIndex({radius: 25});
    index.load(places2.features);
    const tile = index.getTile(0, 0, 0);
    writeFileSync('out.json', JSON.stringify(tile.features));
});
```
Both tiles contained 58 points, and JSON.stringify(tile.features)===JSON.stringify(tile.features) (new geojson-vt -> old supercluster lib) was an exact match in both libraries. The only difference was the tile metadata {transformed: true, source: null, x: 0, y: 0, z: 0} in the newer.

In other words, no progress, but wanted to let you know.

HarelM:
Did you pass the same cluster parameters?

wayofthefuture:
Only the radius

HarelM:
I was able to reproduce this fairly easily on maplibre by replacing the initialization in the factory method. What you did sounds like what I would've done to try and reproduce this, but I'll give it anothe go early next week nevertheless.

wayofthefuture:
Found it:

```
    // wraps features (ie extreme west and extreme east)
    features = wrap(features, options);
```
https://github.com/maplibre/geojson-vt/blob/ada402f655274d09c77b903f759535edad6a23e4/src/geojsonvt.ts#L79-L80

```
    if (!options.cluster) features = wrap(features, options);
```

fixes it. however, when applying a source diff the features would be wrapped there as well. so i think the solution may be to update expected.png in the test.

HarelM:
Can you better explain why updating the image is the right solition? 
I don't mind updating the image, just curious as to why.

wayofthefuture:
The source data contains wrapped features. It would be complex to have wrapped for geojsonvt and unwrapped for supercluster. It's a confusing but very slight difference between the images:

expected / actual:
image

HarelM:
Feel free to update the image and I'll merge this. I think both options are valid I guess, aligning then to the same processing makes sense.

src/source/geojson_source.test.ts
```diff
@@ -268,7 +268,8 @@ describe('GeoJSONSource.update', () => {
                     tolerance: 4,
                     buffer: 256,
                     lineMetrics: false,
-                    generateId: true
+                    generateId: true,
+                    updateable: true
```
HarelM: This is now always on, can we remove this parameter?

src/source/geojson_source.ts
```diff
@@ -195,7 +199,8 @@ export class GeoJSONSource extends Evented implements Source {
                 extent: EXTENT,
                 maxZoom: this.maxzoom,
                 lineMetrics: options.lineMetrics || false,
-                generateId: options.generateId || false
+                generateId: options.generateId || false,
+                updateable: true
```
HarelM: I meant this one...

src/source/geojson_source.ts
```diff
@@ -286,8 +291,16 @@ export class GeoJSONSource extends Evented implements Source {
      * @returns a promise which resolves to the source's actual GeoJSON data
      */
     async getData(): Promise<GeoJSON.GeoJSON> {
-        const options: LoadGeoJSONParameters = extend({type: this.type}, this.workerOptions);
-        return this.actor.sendAsync({type: MessageType.getData, data: options});
+        if (this._data.url) {
+            await this.once('data'); // wait for loading to complete
```
HarelM, line 294: There's a need to cover this with a test I guess... Let me know if you want to do it...

test/examples/update-geojson-polygons.html
```diff
@@ -0,0 +1,194 @@
...
+
+        return {
+            type: 'Polygon',
+            coordinates: [coordinates]
+        };
+    }
+
+    map.on('load', () => {
+        const features = [];
+        rectangles.forEach(rect => {
```
HarelM: Can we use regular for llop?

test/examples/update-geojson-polygons.html
HarelM, line 1: This file is a bit long, the example is great but I'm wondering if it can be simplified a bit.

HarelM:
Might be interesting to increase coverage while we are at it.

src/source/geojson_source.ts
wayofthefuture: No because geojson-vt default options is false... want to change it there?

src/source/geojson_source.test.ts
wayofthefuture: See below

src/source/geojson_source.ts
HarelM: We can pass the relevant parameter here hardcoded, can't we? We don't have to change the default in geojson-vt package, don't we? 
I mean not to pass it to the worker and instead pass it to the geojson initialization options.

wayofthefuture: Yes

wayofthefuture: I'm good with either way

HarelM: I think making the change only here makes more sense.

wayofthefuture:
In order to hardcode updateable I had to remove the non-updateable tests.. let me know what you think. The tests weren't very good anyway because they simulated geojsonvt worker options that don't exist. But we still have to consider geojson sources without ids...

wayofthefuture:
Fyi also see the `misc changes` commit above

PR #7184, opened 2026-02-26
feat: async RequestTransformFunction

### LINKED ISSUE #7180: Async RequestTransformFunction
## User Story

I am running a private map tile server that requires any request to bear an access token. As the lifetime of an access token is short, I would like to refresh it when the map makes a tile request; i.e., in a [`RequestTransformFunction`](https://maplibre.org/maplibre-gl-js/docs/API/type-aliases/RequestTransformFunction/).

The current `RequestTransformFunction` cannot be `async`, however, token refreshing usually involves an asynchronous HTTP request. Therefore, we cannot wait for the token refreshed in `RequestTransformFunction`.

### What are some example use cases?

Here is what I would like to do:

```ts
let map = new Map({
  container: 'map',
  center: [-122.420679, 37.772537],
  zoom: 13,
  style: style_object,
  hash: true,
  // async RequestTransformFunction
  transformRequest: async (url, resourceType) => {
    if(resourceType === 'Tile' && url.startsWith('https://myHost')) {
      const token = await refreshToken(); // asynchronously refresh the access token
      return {
        url,
        headers: { Authorization: `Bearer ${token}` }
      }
    }
  }
});
```

### Can it be done today? Is there a workaround?

I can trigger a refresh of an access token from `RequestTransformFunction`, but the subsequent tile request fails as the retrieval of a refreshed token has to wait for another cycle of the event loop. 

Or, I can also trigger a refresh of an access token prior to map tile requests, but it would be difficult to prevent some map tile requests from failing before the token refresh completes.

## Impact

As I described above, the token refresh and the token retrieval happen in separate cycles of the event loop, and some map tile requests will end up with authorization errors. Smooth chaining of the token refresh and the map tile request is not possible.

## Remarks

I can offer a PR. You can find my changes in [my fork](https://github.com/codemonger-io/maplibre-gl-js/tree/async-transform-request).

kikuomax:
## Proposed changes

- Add `Promise<RequestParameters>` to the return type of `RequestTransformFunction`; i.e., it may be `async`.
- Reorganize code where `transformRequest` is invoked so that no type errors are reported and existing unit tests won't break.
- Introduce new unit tests that test `async transformRequest`. They are basically variants of existing unit tests involving `transformRequest`.

## Related issue

- closes #7180

## Launch Checklist

 - [x] Link to related issues.
 - N/A Include before/after visuals or gifs if this PR includes visual changes.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/source/raster_dem_tile_source.test.ts
+++ b/src/source/raster_dem_tile_source.test.ts
@@ -356,6 +356,7 @@ describe('RasterDEMTileSource', () => {
         const source = createSource({url: '/source.json'});
         const promise = waitForMetadataEvent(source);
 
+        await sleep(0);
         server.respond();
         await promise;
 
@@ -367,6 +368,7 @@ describe('RasterDEMTileSource', () => {
             setExpiryData() {}
         } as any as Tile;
         const tilePromise = source.loadTile(tile);
+        await sleep(0);
         server.respond();
         await tilePromise;
         expect(tile.state).toBe('loaded');
--- a/src/source/vector_tile_source.test.ts
+++ b/src/source/vector_tile_source.test.ts
@@ -7,7 +7,7 @@ import {OverscaledTileID} from '../tile/tile_id';
 import {Evented} from '../util/evented';
 import {RequestManager} from '../util/request_manager';
 import fixturesSource from '../../test/unit/assets/source.json' with {type: 'json'};
-import {getMockDispatcher, getWrapDispatcher, waitForEvent, waitForMetadataEvent} from '../util/test/util';
+import {getMockDispatcher, getWrapDispatcher, sleep, waitForEvent, waitForMetadataEvent} from '../util/test/util';
 import {type Map} from '../ui/map';
 import {type WorkerTileParameters} from './worker_source';
 import {SubdivisionGranularitySetting} from '../render/subdivision_granularity_settings';


HarelM:
This looks a bit too complicated with all the `instanceof`...
Can these be reduced somehow? In async methods you can await a non promise returning method, so I think some of the code can be simplified, not sure about the rest of the code.
Maybe wrap the transfrom request in a promise to make it easier to work with? IDK...

kikuomax:
@HarelM 

> This looks a bit too complicated with all the `instanceof`... Can these be reduced somehow? In async methods you can await a non promise returning method, so I think some of the code can be simplified, not sure about the rest of the code. Maybe wrap the transfrom request in a promise to make it easier to work with? IDK...

I did it intentionally not to break the existing unit tests. I first tried to simply prepend `await` to `transformRequest` calls, but doing so caused a lot of timeout errors in the unit tests. It seems that `await` introduces an additional event loop cycle whether the awaited value is a promise or not. I did not want to change the existing unit tests to make sure I did not break anything currently working and I ended up with the current solution.

kikuomax:
@HarelM
I am working on reducing the `instanceof Promise` checks to simple `await`. A lot of exiting unit tests have to be modified, but I found it manageable. I am going to make this PR draft until I finish the modification. I will let you know when updates are completed.

HarelM:
Thanks!

kikuomax:
@HarelM 
I was able to replace all the `instanceof Promise` checks with simple `await`.

The only concern is that the extra microtask delay introduced by unconditionally awaiting `transformRequest` might cause subtle errors in some applications that, maybe unintentionally, depend on strict timing of the `transformRequest` resolution, as the unit tests broke.

src/ui/control/attribution_control.test.ts
```diff
@@ -326,7 +329,10 @@ describe('AttributionControl', () => {
         map.on('data', spy);
         await map.once('load');
         map.addSource('1', {type: 'raster-dem', url: '/source.json'});
-        server.respond();
+        setTimeout(() => {
```
HarelM: Use "sleep" instead?

src/source/geojson_source.test.ts
```diff
@@ -158,6 +158,28 @@ describe('GeoJSONSource.setData', () => {
             });
         };
         source.setData('http://localhost/nonexistent');
+        await sleep(0); // to resolve pending `await`s
```
HarelM: No need to keep these comments, sleep(0) is self explanatory.

src/source/image_source.test.ts
```diff
@@ -45,6 +45,11 @@ class StubMap extends Evented {
 describe('ImageSource', () => {
     stubAjaxGetImage(undefined);
     let server: FakeServer;
+    // delay server.respond so that it happens after the pending request is made
+    // this works around a microtask delay introduced by unconditionally awaiting the transformRequest result
+    const delayServerRespond = () => {
```
HarelM: Maybe move this to a test util file? I would prefer to avoid using this and just use sleep if possible as this can't be awaited...

src/source/load_tilejson.test.ts
```diff
@@ -3,11 +3,17 @@ import {fakeServer, type FakeServer} from 'nise';
...
 import {ABORT_ERROR} from '../util/abort_error';
+import {sleep} from '../util/test/util';
 
 import {type RasterSourceSpecification} from '@maplibre/maplibre-gl-style-spec';
 
 describe('loadTileJson', () => {
     let server: FakeServer;
+    // delay server.resond so that it happens after the pending request is made
+    // this works around a microtask delay introduced by unconditionally awaiting the transformRequest result
+    const delayServerRespond = () => {
```
HarelM: See previous message....

src/ui/map.ts
```diff
@@ -2152,10 +2152,10 @@ export class Map extends Camera {
         }
     }
 
-    _diffStyle(style: StyleSpecification | string, options?: StyleSwapOptions & StyleOptions) {
+    async _diffStyle(style: StyleSpecification | string, options?: StyleSwapOptions & StyleOptions) {
```
HarelM, line 2154: Can this method avoid using "then" now with this change?

HarelM:
Looks a lot better now, thanks! 
I've added some minor comments about delayedServerResponde and tests stuff.
There's also some map async code that I think is worth solving, but otherwise great work!

src/ui/control/attribution_control.test.ts
kikuomax: Good catch! The following code should do the same:

```ts
await sleep(0);
server.respond();
```

src/source/image_source.test.ts
kikuomax: All of the occurrences of `delayServerRespond()` can be replaced with the code I wrote in [this comment](https://github.com/maplibre/maplibre-gl-js/pull/7184#discussion_r2871251122). I do not think `delayServerRespond` is necessary in the end.

src/source/load_tilejson.test.ts
kikuomax: Please see https://github.com/maplibre/maplibre-gl-js/pull/7184#discussion_r2871265241.

src/ui/map.ts
kikuomax: I think so. I will change it.

src/ui/control/attribution_control.test.ts
HarelM: let's use that then.

kikuomax:
@HarelM Would mind me rebasing my branch to solve the conflict? Or would you prefer a merge commit?

HarelM:
Whichever works for you, I'll squash merge it at the end...

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 - Add Etag unmodified support to optimize vector tile reloading ([#7074](https://github.com/maplibre/maplibre-gl-js/pull/7074)) (by [@rivkamatan](https://github.com/rivkamatan and [@wayofthefuture](https://github.com/wayofthefuture))
 - Add `boxZoom.boxZoomEnd` option to customize the action after Shift-drag box selection ([#6397](https://github.com/maplibre/maplibre-gl-js/issues/6397))
 - Add updateable support for GeoJSON-VT ([#7172](https://github.com/maplibre/maplibre-gl-js/issues/7172)) (by [@wayofthefuture](https://github.com/wayofthefuture) and [HarelM](https://github.com/HarelM))
+- Make `setTransformRequest` accept an async function in addition to a sync function.
```
```suggestion
- Make `setTransformRequest` accept an async function in addition to a sync function. ([#7184](https://github.com/maplibre/maplibre-gl-js/issues/7184)) (by [@kikuomax
](https://github.com/kikuomax))
```

src/source/geojson_source.test.ts
```diff
@@ -158,6 +158,28 @@ describe('GeoJSONSource.setData', () => {
...
+        source.map = {
+            _requestManager: {
+                transformRequest: async (url) => { return {url}; }
+            } as any as RequestManager
+        } as any;
+        const spy = vi.fn();
+        source.actor.sendAsync = (message: ActorMessage<MessageType>) => {
+            return new Promise((resolve) => {
+                if (message.type === MessageType.loadData) {
+                    expect((message.data as any).request.collectResourceTiming).toBeTruthy();
```
HarelM: In other tests I moved this expect to the lower part of the test by passing the message to the spy, can you change this was well here (this was a recent change I did to these tests)?

src/source/geojson_source.test.ts
```diff
@@ -251,10 +273,11 @@ describe('GeoJSONSource.update', () => {
         });
 
         new GeoJSONSource('id', {data: {}} as GeoJSONSourceOptions, mockDispatcher, undefined).load();
+        await sleep(); // to resolve pending `await`s
```
HarelM: Remove comment please.

src/source/geojson_source.test.ts
```diff
@@ -366,7 +392,7 @@ describe('GeoJSONSource.update', () => {
 
         // Wait for initial data to be loaded
         source.load();
-        await waitForEvent(source, 'data', (e: MapSourceDataEvent) => e.sourceDataType === 'metadata');
+        await sleep(0);
```
HarelM: This looks like it changes the timing of the test, are you sure about this change?

HarelM:
I've added some comments after latest merge from main, overall looks good.
See that you take care of the remaining comments and I'll merge this.
THANKS!

src/source/geojson_source.test.ts
kikuomax: Sorry, it is a trace of my trial and error. It is an unnecessary change.

src/source/geojson_source.test.ts
kikuomax: Sorry, It escaped from my grep.

src/source/geojson_source.test.ts
kikuomax: The second `sleep(0)` in this test case is intended to wait until all the updates are applied. I modified this test case so that it matches the actual behavior. `spy` used to be called twice, however, in the latest `main` branch, it is called three times whether `async transformRequest` is introduced or not.

HarelM: I'm not sure I'm fully following as in main it passes the unit tests and is only called 2 times.
Nevertheless, if this is intentional and correct let's keep it.

src/ui/map.ts
HarelM: Any updates on this?

src/source/geojson_source.test.ts
kikuomax: Would like me to replace other occurrences of `spy()` with `spy(message)` in this file?

HarelM: Yes please, I might have missed some by mistake, sorry for the extra work.

kikuomax:
@HarelM Sorry, I have pushed minor changes after your approval.

HarelM:
No worries. Thanks for all the work on this! A great addition to this library.

HarelM:
Can you check if the conflicts can be resolved? I was sure this was already merged...

HarelM:
Seems like I don't have permissions to update your fork, so you'll have to do it on your own...

kikuomax:
@HarelM I have resolved the conflicts.

PR #7198, opened 2026-02-28
Example 3D Tiles

### LINKED ISSUE #2758: support loading B3DM format data
hello, when I want load B3DM with map, but tips: `deck: initialization of ScenegraphLayer({id: 'ceshi-3d-scenegraph-http://localhost:5173/static/20230628/Tile_-3340_9261_0000/Tile_-3340_9261_0000_-0001_0007.b3dm'}): Cannot convert undefined or null to object`;

but  I use `mapbox-gl` , it is right.

so , plan to support data like B3DM?

thank you.

hh-hang:
https://github.com/maplibre/maplibre-gl-js/discussions/3378#discussioncomment-15953887

Add a 3D Tiles example.


jo-chemla:
I'm not part of maplibre team, just a user, but you might want to avoid adding b3dm binary data files (the `obj2tiles` directory) to this PR and instead rely on open-data OGC 3d-tiles datasets, remotely accessible on public endpoints, listed eg here [CesiumGS/3d-tiles/RESOURCES open-data](https://github.com/CesiumGS/3d-tiles/blob/main/RESOURCES.md#open-data). 

Also, looking at the screenshot, it looks like the tileset is not correctly georeferenced on top of the maplibre globe. The open-data section shares georeferenced datasets, so you probably can tweak your transform code to handle matching origins/coordinate systems - also could rely on NASA-AMMOS/3DTilesRendererJS [ReorientationPlugin ](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/three/plugins/README.md#reorientationplugin), see [this example](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/ff99ee37c7cc3cd30f51251d89e50f2e511cccfc/example/three/googleMapsAerial.js#L40C1-L41C1) 

```js
	tiles.registerPlugin( new ReorientationPlugin( { lat: 35.6586 * MathUtils.DEG2RAD, lon: 139.7454 * MathUtils.DEG2RAD } ) );
```

HarelM:
I agree that using an existing 3D tiles source would be better than adding these files here.

hh-hang:
> I'm not part of maplibre team, just a user, but you might want to avoid adding b3dm binary data files (the `obj2tiles` directory) to this PR and instead rely on open-data OGC 3d-tiles datasets, remotely accessible on public endpoints, listed eg here [CesiumGS/3d-tiles/RESOURCES open-data](https://github.com/CesiumGS/3d-tiles/blob/main/RESOURCES.md#open-data).
> 
Our company actually has many online 3D Tiles datasets, but due to data confidentiality we can’t upload them to the example. So I downloaded a representative public dataset and will switch the example’s 3D Tiles to Cesium’s official online data.I will directly reference Cesium’s official public datasets for the example. Thank you for your suggestion.

> Also, looking at the screenshot, it looks like the tileset is not correctly georeferenced on top of the maplibre globe.
> 
Regarding the issue: my understanding is that 3D Tiles generally include a transform matrix. For photogrammetry models that matrix encodes rotation and translation from the model origin into specific Cartesian coordinates. In my example I convert those Cartesian coordinates to latitude/longitude, use that lat/lon as the origin in three.js, then translate the 3D Tiles coordinates to the origin and apply the inverse rotation matrix so the 3D Tiles are placed correctly. Because the example doesn’t load terrain and 3D Tiles themselves include elevation, the model ends up floating above the map and does not snap to the ground.

>he open-data section shares georeferenced datasets, so you probably can tweak your transform code to handle matching origins/coordinate systems - also could rely on NASA-AMMOS/3DTilesRendererJS ReorientationPlugin
>
This approach ensures accurate placement and is simpler, but I don’t think it’s very representative, since most users will use their own data.

hh-hang:
> Thanks for taking the time to open this PR.
> I agree that using an existing 3D tiles source would be better than adding these files here.

Got it — I'll switch to using the existing 3D tiles source. Thanks for the suggestion.

hh-hang:
I noticed that Cesium's official sample datasets often lack the transform property in the tileset.json. Does this imply that these 3D Tiles are already stored in world-fixed Cartesian coordinates (ECEF)? If so, my current method—which likely relies on explicit matrix information—might not be compatible with datasets that don't provide it.Because I am unable to apply the proper inverse matrix
b5f4c83850051ba6e7c4ddf447030d77

jo-chemla:
Those are correctly georeferenced 3D-tiles tileset, indeed usually expressed in ECEF EPSG:4978 - although the spec now accepts broader coordinate system, which is present in the tileset root metadata. 

See eg this [sandcastle](https://sandcastle.cesium.com/#c=fZJva9swEMa/ymEGdUYseW73pnPDRlrYYKGlTffKsCry1dYmS0Ynx6Sl333+S1M29krc3e95dHeSqmrrPLwHQbBGUk0Fj85WkAVyiLLgU2YyI60hD3uFLTq4AIPtRLMfQy6c+bU1XiiDLgsWg9K7AzxnBmC08Eojoe88RCuUn13G4/RyO5ZZ38O90+FJ6X1N55zTKcMmapF8lLBWkNgpOhCTtuJYUUSF9FFdWm8robQWBfLryy/rb1t+tbn7ebNJ4iTh093sF1lz0ncH00SMJBpktVOV8mqPxESehxP+BnyyttrasM/APMtyjI528hVFrkxxo7wsb4UpcBLADGyEL5m3tx0mDIUxixfL/yPR2cdjKInjTjSFY7rv8wWk6O6EEJ2zbvG6eKuRaVuED1d9AbQdGpwnOId3z4Pi5WFwyUywDFLyB42r3uGzGn9J070IY9xjVWvhkfiukb+7hUqiXpfyWZLmag8qv/jHrwCpBVFXeWy0vlNPmAWrlHf8G9nU4PUenRaHHik/rL6PScZYyrvwb5W3Vu+EO3L8Aw) that loads one of the open-data samples (it can be useful to use an allow-cors plugin on chrome/firefox

Also see [this example](https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/three/index.html#https://soest.virtualcitymap.de/datasource-data/HOSTING-Soest/3D-Mesh_2023/tileset.json) based on 3DTilesRendererJS which does accept passing a tileset url as hash. If you use any of these tileset urls into eg qgis, you see they are correctly georeferenced and can be displayed on top of eg Google 3D Cities 3Dtiles tileset.

test/examples/add-a-3d-tiles-using-threejs.html
```diff
@@ -0,0 +1,216 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>Add a 3D model using three.js</title>
```
HarelM: This should match the file name kebab-case.

.cspell.json
HarelM, line 1: Can the change here be reduced to only the added word?

pnpm-lock.yaml
HarelM, line 1: Please remove this file

test/examples/add-a-3d-tiles-using-threejs.html
```diff
@@ -0,0 +1,216 @@
...
+                    return;
+                }
+
+                const scale = 1;
+                const sphere = new THREE.Sphere();
+                tiles.getBoundingSphere(sphere);
+                const center = sphere.center.clone();
+                const root = (tiles ).root;
+
+                console.log('tiles',tiles)
```
HarelM: I don't think we should add console.log in the example...

test/examples/add-a-3d-tiles-using-threejs.html
```diff
@@ -0,0 +1,216 @@
...
+                camera.projectionMatrix.multiply(localTransform);
+                renderer.resetState();
+                renderer.render(scene, camera);
+                if (tiles) tiles.update();
+                mapInstance?.triggerRepaint();
+            },
+        };
+
+        map.on('style.load', () => {
+            console.log("style.load");
```
HarelM: Same here

.cspell.json
hh-hang: ok

test/examples/add-a-3d-tiles-using-threejs.html
HarelM: Also, if possible use `await map.once('style.load')`.

test/examples/add-a-3d-tiles-using-threejs.html
```diff
@@ -0,0 +1,216 @@
...
+                    console.warn("tileset does not have a transform matrix, using default matrix");
+                }
+                loadedTileSetHandled = true;
+
+                const { lng, lat, alt } = ecefToLngLatAlt(center.x, center.y, center.z);
+                map.jumpTo({ center: [lng, lat], zoom: 18, pitch: 60 });
+                updateLocalTransform([lng, lat, alt + altOffset]);
+
+                const rotationMat3 = new THREE.Matrix3().set(m[0], m[1], m[2], m[8], m[9], m[10], -m[4], -m[5], -m[6]); // The original model is based on a Z-up coordinate system (oblique photography), apply this matrix transformation
+                // const rotationMat3 = new THREE.Matrix3().set(-m[0], -m[1], -m[2], -m[4], -m[5], -m[6], -m[8], -m[9], -m[10]); // The original model is based on a Y-up coordinate system (consistent with three.js), apply this matrix transformation
```
HarelM: Is this comment relevant?

test/examples/add-a-3d-tiles-using-threejs.html
```diff
@@ -0,0 +1,216 @@
...
+                const center = sphere.center.clone();
+                const root = (tiles ).root;
+
+                console.log('tiles',tiles)
+                let m = [1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1];
+                // Get matrix
+                if (root.transform) {
+                    m = root.transform;
+                } else {
+                    console.warn("tileset does not have a transform matrix, using default matrix");
```
HarelM: Is this interesting for this example?

test/examples/add-a-3d-tiles-using-threejs.html
```diff
@@ -0,0 +1,216 @@
...
+                if (loadedTileSetHandled) {
+                    tiles?.removeEventListener("load-tileset", loadTileSet);
+                    return;
+                }
+
+                const scale = 1;
+                const sphere = new THREE.Sphere();
+                tiles.getBoundingSphere(sphere);
+                const center = sphere.center.clone();
+                const root = (tiles ).root;
```
HarelM: remove space and parenthesis.

HarelM:
I've added some minor comments, looks great otherwise, thanks!!

hh-hang:
> I've added some minor comments, looks great otherwise, thanks!!
>
Thanks, I’ve made the changes and am preparing to update.

test/examples/add-a-3d-tiles-using-threejs.html
```diff
@@ -0,0 +1,209 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <title>Add a 3D tiles using three.js</title>
```
HarelM: "a 3D tiles" is probably not the right gramer...
Should be: Add 3D tiles using Three.js, son't forget to change the file name.

hh-hang: Okay, it has been updated.

.cspell.json
HarelM: This still needs addressing...

test/examples/add-3d-tiles-using-threejs.html
```diff
@@ -0,0 +1,209 @@
...
+                camera.projectionMatrix.fromArray(args.defaultProjectionData.mainMatrix);
+                camera.projectionMatrix.multiply(localTransform);
+                renderer.resetState();
+                renderer.render(scene, camera);
+                if (tiles) tiles.update();
+                mapInstance?.triggerRepaint();
+            },
+        };
+
+        map.on('style.load', () => {
```
HarelM: Is it possible to use `once` here?

hh-hang: Yes. Should I change it?

HarelM: Yes please.

test/examples/add-3d-tiles-using-threejs.html
```diff
@@ -198,7 +198,7 @@
             },
         };
 
-        map.on('style.load', () => {
+        map.once('style.load', () => {
```
HarelM: please use `await map.once('style.load')` instead.

test/examples/add-3d-tiles-using-threejs.html
hh-hang: Okey.

test/examples/add-3d-tiles-using-threejs.html
hh-hang: It has been updated.

HarelM:
Please fix the cspell file and I'll merge this, thanks.

hh-hang:
> Please fix the cspell file and I'll merge this, thanks.
>

I’m not quite sure exactly which parts of the cspell file need to be modified. Could you please point them out? Thank you.

HarelM:
I would recommend downloading the previous version and modify it without changing formatting. You can see how it looks in GitHub diff page.

.cspell.json
```diff
@@ -251,7 +251,8 @@
     "zensical",
     "Zfrom",
     "zoomto",
-    "ZULN"
+    "ZULN",
+	"dtiles"
```
HarelM: Can you sort this alphabetically? Sorry for the extra work...

hh-hang: Not at all, it's done!

hh-hang:
> See eg this [sandcastle](https://sandcastle.cesium.com/#c=fZJva9swEMa/ymEGdUYseW73pnPDRlrYYKGlTffKsCry1dYmS0Ynx6Sl333+S1M29krc3e95dHeSqmrrPLwHQbBGUk0Fj85WkAVyiLLgU2YyI60hD3uFLTq4AIPtRLMfQy6c+bU1XiiDLgsWg9K7AzxnBmC08Eojoe88RCuUn13G4/RyO5ZZ38O90+FJ6X1N55zTKcMmapF8lLBWkNgpOhCTtuJYUUSF9FFdWm8robQWBfLryy/rb1t+tbn7ebNJ4iTh093sF1lz0ncH00SMJBpktVOV8mqPxESehxP+BnyyttrasM/APMtyjI528hVFrkxxo7wsb4UpcBLADGyEL5m3tx0mDIUxixfL/yPR2cdjKInjTjSFY7rv8wWk6O6EEJ2zbvG6eKuRaVuED1d9AbQdGpwnOId3z4Pi5WFwyUywDFLyB42r3uGzGn9J070IY9xjVWvhkfiukb+7hUqiXpfyWZLmag8qv/jHrwCpBVFXeWy0vlNPmAWrlHf8G9nU4PUenRaHHik/rL6PScZYyrvwb5W3Vu+EO3L8Aw) that loads one of the open-data samples (it can be useful to use an allow-cors plugin on chrome/firefox
> 
> Also see [this example](https://nasa-ammos.github.io/3DTilesRendererJS/example/bundle/three/index.html#https://soest.virtualcitymap.de/datasource-data/HOSTING-Soest/3D-Mesh_2023/tileset.json) based on 3DTilesRendererJS which does accept passing a tileset url as hash. If you use any of these tileset urls into eg qgis, you see they are correctly georeferenced and can be displayed on top of eg Google 3D Cities 3Dtiles tileset.
>

Thanks for the explanation, I'll look into it.

HarelM (approved):

safwat-halaby:
I have some improvements to this example that I would like to submit. Before tidying a PR I would like some feedback. My code is working for my datasets and I want to make sure it is generic. Here is the plan.

## root.transform

The current example assumes that `root.transform` exists and that it has the information needed to orient the model so that its up side faces up in the web Mercator plane. The 3dtiles model will be misplaced if these assumptions aren't true.

The code calculates a "center" in `const { lng, lat, alt } = ecefToLngLatAlt(center.x, center.y, center.z);`. It translates the model to those lngLat, then it rotates it based on information from `root.transform`.


I think it is always possible to calculate the necessary rotation matrix solely based on the `lngLat` of the `center` without any reliance on `root.transform`. It worked for my testcase (which has no root.transform) and for the current example at the docs (`https://pelican-public.s3.amazonaws.com/3dtiles/agi-hq/tileset.json`).


## Anchor point improvements

The "origin/anchor" point to mapLibre's plane is arbitrary chosen to be the `lngLat` of the center of the model bounding sphere (`const { lng, lat, alt } = ecefToLngLatAlt(center.x, center.y, center.z);`), the rest of the model vertices are relative to that anchor point, in cartesian meters. This rapidly looses precision when moving away from the anchor point. On a city-scale dataset this is noticable. I propose we put that anchor point at Maplibre's `map.getCenter()`, this can be calculated at every render, or only when the last calculated anchor point is more than, say, 1km away from `map.getCenter()`. This ensures perfect precision even on huge datasets. (edit: we can fallback to the center if the camera is significantly out of the models' bounds or to the point nearest to `map.getCenter` that is still within the bounds)

safwat-halaby:
Further info regarding `root.transform`: 

I think that `3d-tile-renderer` already applies the `.transform` matrices internally. If that is true, then from our perspective it is always an ECEF world. In such a world, all that we need to rotate the model properly is to know the `lngLat` of the anchor point we're rotating around.

Since my code is working whether there is or isn't a `.transform` in the dataset I think my hypothesis is correct but please let me know if I am missing something.

HarelM:
Thanks for taking the time to try and improve this example!
I think it would be best to open a PR with the relevant changes you would like to do and tag the people in this PR for feedback.
It would be easier to understand what you are proposing to change and how.

jo-chemla:
Nice to see you're willing to improve on this `maplibre + 3Dtiles` example!

From what I recall indeed 3DTilesRendererJS default mode is `EPSG:4978 ECEF CRS` coordinate reference system, and its [camera controls](https://github.com/NASA-AMMOS/3DTilesRendererJS/tree/8a5fbf07b8b0f2e1bc46bf52a598cd41e0f077d1/src/three/renderer/controls)  (`Globe/Environment`) use that to define the camera matrix updates (target lookat, cam center position, up/side vectors) based on user mouse input. Floating Point precision errors roughly cancels out when applying camera world matrix inverse to the composed node/leaf world matrix when camera is close to tile, which makes the rendering numerically stable (no jitter when camera moves, or mesh edges clamped like a scale effect).

Since Garrett made the whole library very well composable/extensible, there is however a great plugin that could be useful to having a ~perfect match between the maplibre coordinate system representation and the 3D-tiles tileset, which is the [ReorientationPlugin](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/src/three/plugins/API.md#reorientationplugin). I imagine this could be useful to making the new iteration of this example more future-proof, and working with all tileset and transforms configurations. 

Regarding the fact that the difference between the tileset data and underlying globe becomes more and more noticeable after a few kilometers, I would imagine this more being a underlying data/tileset issue rather than the current maplibre internals, but I have not looked into it at all. We do produce 3D-scans of cities, and the projection system used for the processing has to be aware of this from the beginning. It would be nice to see if there is the same height discrepancy if you overlay your tileset on top of say Google Photorealistic 3D tiles, both via 3DTilesRendererJS (see eg [googleMapsAerial.js example](https://github.com/NASA-AMMOS/3DTilesRendererJS/blob/master/example/three/googleMapsAerial.js)) or on top of a cesium terrain/globe via cesiumjs in their [Google 3D Tiles sandcastle](https://sandcastle.cesium.com/?id=google-photorealistic-3d-tiles), which would help better identify the root cause for the error - be it the data or the rendering stack. 

Your idea of defining the tileset center at init, and eventually updating it when distance to center reaches threshold looks also good way to better overcome precision loss, but I would assume you could re-use the map center maplibre exposes that is probably used at every render to project terrain to screen at no additional cost. 

> It is worth noting that https://github.com/maplibre/maplibre-gl-js/pull/7523 makes room for much flexibility in 3d tiles. A transparent RGB terrain with disableTerrainSkirts: true allows us to create style spec polygons that trace the ground of a 3dtile model. Would such an example be useful in the docs?

Is your idea to have the polygon be used as a cut-out within the 3D-tiles tileset? Would be nice indeed! [Here](https://sandcastle.cesium.com/?id=aec-clipping) is a nice recent cesiumjs way of doing it using a user-defined clipping polygon, and another nice [PR/discussion](https://github.com/CesiumGS/cesium/pull/13244#issuecomment-3991180930) about of getting geometry on the CPU to compute the clipping polygon automatically that evolved into something more generic over there https://github.com/CesiumGS/cesium/pull/13330. 

Hope this helps kickstart ideas into this new implementation!

EDIT: just had a quick scroll and noticed I already suggested a few messages above in the thread, hope this is not a duplicate!
https://github.com/maplibre/maplibre-gl-js/pull/7198#issuecomment-3983443280

safwat-halaby:
Thank you for the feedback!

I'll be submitting a pr soon.

At this point I am certain the issue is in the specific Maplibre 3dtiles demo that is currently in the docs. The demo transforms the tiles so that they appear on some specific anchor lat/lon, and from there on it uses the ECEF units as cartesian meters on the web mercator plane. This works well until you're far enough from the anchor. I'll provide some screenshots of before/after the patch in the PR. 

I built my own function to reorient the 3dtiles properly without reliance on `root.transform`. I wasn't aware of the `ReorientationPlugin` and if it provides the same functionality I'll use it instead. Thank you!

> Is your idea to have the polygon be used as a cut-out within the 3D-tiles tileset?

My idea is to be able to draw whatever the maplibre style spec can currently draw (icons, lines, polygons, etc) on top of 3dtiles. such that they appear to be sitting on the surface on the 3dtiles and not awkwardly below or above. Supplying a height map to maplibre along with a transparent background allows that easily. From maplibre's perspective the style spec features are sitting on an "invisible surface". I think I'll submit a separate PR for such a demo.

@HarelM I've been thinking that the demo is too low-level. From a UX perspective, I think I would have wanted 3dtiles loading to be as simple as:

```
const 3dTilesLayer = create3dTilesLayer('https://example.com/tileset.json');
map.addLayer(3dTileLayer);
```

In the demo I intend to encapsulate all the low level functionality in some `create3dTilesLayer` function. But I've been thinking about the form of bundling.
- Maybe this should be a maplibre built-in function?
- Maybe this should be some plugin e.g.

```
import {create3dTilesLayer} from `maplibre-threejs-3dtiles`
```

HarelM:
Usually a plugin is the solution to what you said, the problem is maintenance. So if you think you can commit to maintain it, making it a plugin would be the right approach.
Including it in maplibre will require a lot of code for 3D tiles, and we don't want to bloat the package on something that most users do not need/want.

safwat-halaby:
I have a glue library here: https://github.com/safwat-halaby/maplibre-gl-three

It is still a work in progress. It is based off the original example shown here but with some polishing and upgrades, including solving the issue raised by @hh-hang 

I'll open a proper PR with more explanation when things are more polished but feel free to have a look for now.

safwat-halaby:
@hh-hang I am wondering if I could reuse `https://pelican-public.s3.amazonaws.com/3dtiles/agi-hq/tileset.json` in the examples in [maplibre-gl-three](https://github.com/safwat-halaby/maplibre-gl-three). Is this publicly licensed? Should I credit anyone for the data?

By the way, in the upcoming version I will be taking vertical datums into account: If a terrain is loaded to maplibre in addition to the 3d-tiles, a 3d-tile will not need any vertical offseting and will sit just right! Unless of course there are actual vertical errors in the data.

Also, I suspect this particular dataset has a ~234 meter vertical error AFTER datum corrections. Which means it would still require a ~234 offset in the upcoming version:

Dataset height above ellipsoid: ~310 meters
Dataset height above geoid (above sea level): ~344 meters
Expected height above see level at -75.596, 40.038: ~110 meters
Expected height above ellipsoid at -75.596, 40.038: ~76 meters

jo-chemla:
This dataset is from cesium itself, their headquarters, shared on the [open-data](https://github.com/CesiumGS/3d-tiles/blob/main/RESOURCES.md#open-data) section of their 3D-tiles ecosystem Resource Page. There are other datasets, note not all are easy to use, some are published as older versions of the 3d-tiles spec, some with not well supported extensions etc. But a pretty interesting resource nonetheless if you want to benchmark your implementation against various sources!

hh-hang:
>Also, I suspect this particular dataset has a ~234 meter vertical error AFTER datum corrections. Which means it would still require a ~234 offset in the upcoming version:

I agree. I noticed the same issue after adding terrain. I suspect this particular 3D Tiles dataset has an inherent vertical offset or elevation error, so applying an additional offset after terrain alignment seems to be the most reliable approach for now. I use the same approach in Cesium as well.

PR #913, opened 2022-01-28
Fix stale tiles being shown when calling VectorTileSource#setTiles while the map is moving

vanilla-lake:
<changelog>Fix stale tiles being shown when calling VectorTileSource#setTiles while the map is moving</changelog>

Before this PR, when calling `VectorTileSource#setTiles`, the source cache was cleared before the TileJSON was reloaded asynchronously. If the map was moving while `VectorTileSource#setTiles` was called, it was possible for the cache to be repopulated with the old tiles if `VectorTileSource#loadTile` happened to be called before the TileJSON had been reloaded. This was due to the `VectorTileSource#tiles` property, which is used by `VectorTileSource#loadTile`, only being updated asynchronously after the TileJSON had reloaded.

## Launch Checklist

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.
 - [x] Write tests for all new functionality.
 - [x] Manually test the debug page.
 - [x] Add an entry inside this element for inclusion in the `maplibre-gl-js` changelog: `<changelog></changelog>`.

PR #7213, opened 2026-03-03
Fix VectorTileSource flicker when using `setUrl` or `setTiles`

### LINKED ISSUE #7185: `VectorTileSource` flickers when using `setUrl` or `setTiles`
When updating a vector tile source using `setTiles` or `setUrl`, layers tied to this source disappear until they are loaded again. This behavior doesn't happen with raster tile sources.

**maplibre-gl-js version**: 5.19.0

### Steps to Trigger Behavior

 1. Add a vector tile source, and a layer to this source.
 2. Change the source's data with `setUrl` or `setTiles`.
 3. The layer disappears briefly (depending on how long the new tiles take to load).

### Link to Demonstration

https://jsbin.com/milecejuxo/

In this demo, the map on top displays countries in red using a vector tile source, while the bottom map displays numbers using a raster tile source.
Click on the button below a map to update this map's source using `setUrl` (the URL is set to the same one, but the behavior is the same as if it was a different URL).

### Expected Behavior

The source should reload smoothly by keeping previous tiles on screen until the new ones are loaded.

In my opinion, keeping previous tiles rendered should be the default behavior when using `setUrl` and `setTiles`. If one wants not to continue displaying tiles while the new ones load, it can easily be done by removing and recreating the source. It will also be consistent with how raster tile sources behave.

madoci:
Fixes #7185 

When using `setUrl` or `setTiles` on a vector tile source, layers don't flicker anymore. Instead, each tile remains on screen until the new one is loaded.

The following images show an extreme case where tiles are completely different from one another, to better visualize the change.

Before:
ezgif-78e67489b25b5f1a

After:
ezgif-74d2f43ad5a4beba

## Implementation

The implementation is a copy of how [`RasterTileSource`](https://github.com/maplibre/maplibre-gl-js/blob/c58a6368ae871987a38de43b30bbf2a72c3d9e55/src/source/raster_tile_source.ts#L91) handles it.

## A few notes

This PR is an implementation proposal, I'm unsure of how this will impact users. Further inquiries show that this issue has been talked about and tackled in many ways. Here's a few information I found about this:
- In https://github.com/mapbox/mapbox-gl-js/pull/8048 (the PR adding `setTiles` and `setUrl`), there are several comments asking for it not to clear the cache. This supports the idea that the behavior implemented in this PR is what users expect.
- #415 talks about this issue, and a solution has been provided in #5806, but doesn't work with `setUrl` and `setTiles`.
- This PR changes what has been done in #913, but since the tiles are not cleared at all and refreshed, the underlying issue they fixed should still be fixed (I tested it and it seems to work fine).

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/source/vector_tile_source.test.ts
+++ b/src/source/vector_tile_source.test.ts
@@ -7,7 +7,7 @@ import {OverscaledTileID} from '../tile/tile_id';
 import {Evented} from '../util/evented';
 import {RequestManager} from '../util/request_manager';
 import fixturesSource from '../../test/unit/assets/source.json' with {type: 'json'};
-import {getMockDispatcher, getWrapDispatcher, sleep, waitForEvent, waitForMetadataEvent} from '../util/test/util';
+import {getMockDispatcher, getWrapDispatcher, waitForEvent, waitForMetadataEvent} from '../util/test/util';
 import {type Map} from '../ui/map';
 import {type WorkerTileParameters} from './worker_source';
 import {SubdivisionGranularitySetting} from '../render/subdivision_granularity_settings';


HarelM:
Removing a reference to the tile manager and raising an event send like the right approach.

PR #7214, opened 2026-03-03
Fix(raster): cancel in-flight raster tile requests when source URL changes - issue #7149

### LINKED ISSUE #7149: source.setUrl does not abort in-flight raster tile requests (tilejson completed)
**maplibre-gl-js version**: 5.18.0

**browser**: Firefox, Chromium

### Summary

Calling `source.setUrl(...)` on a raster source does not abort in-flight raster tile requests if the tilejson request has already completed. The abort only occurs if `setUrl` is called before the tilejson response arrives. As a result, the map continues loading tiles from the old URL(s) after `setUrl` is invoked.

### Steps to Trigger Behavior

1. Add a raster source and raster layer.
2. Wait until the tilejson request completes (so tile URLs are known) but before all raster tiles have finished loading.
3. Call `source.setUrl(newUrl)`.
4. Observe that in-flight raster tiles continue to load for the previous URL(s).

### Link to Demonstration

Example demonstrating this behavior using a custom protocol: https://output.jsbin.com/toxawohare

If the source is removed and re-added the inflight tile-requests are correctly aborted: https://jsbin.com/pidolexoji. This mitigates my problem, but avoiding visual flickering using this pattern is tedious.

### Expected Behavior

Abort inflight tile requests for previous URL when `setUrl` is called.

### Actual Behavior

Inflight requests are only aborted if `setUrl` is called before the tilejson is returned.

pcardinal:
## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

**Summary**

- This PR fixes a raster source lifecycle issue where calling RasterTileSource.setUrl could leave old tile requests running.
- When the URL changes, in-flight tile requests are now aborted deterministically through TileManager.abortAllRequests, with cache-side support in TileCache.
- The change ensures in-flight raster requests are cancelled before reloading source metadata.

**Root cause**
When a raster source URL was updated, the source reload path did not consistently cancel all active requests tied to the previous URL/state.

**Changes included**
- Updated RasterTileSource.setUrl to trigger request cancellation through the tile manager before reload.
- Updated request-cancellation behavior in TileManager.abortAllRequests.
- Updated cache-level request cancellation in TileCache.abortAllRequests to ensure out-of-view pending raster requests are aborted.

**Tests added/updated**
- raster_tile_source.test.ts - setUrl aborts in-flight raster tile requests through TileManager
- tile_manager.test.ts - coverage for TileManager.abortAllRequests
- tile_cache.test.ts - coverage for TileCache.abortAllRequests
 

**How to validate locally**
npm run test-unit -- src/source/raster_tile_source.test.ts
npm run test-unit -- src/tile/tile_manager.test.ts
npm run test-unit -- src/tile/tile_cache.test.ts

**Impact**

- No public API changes.
- Internal behavior change only for raster request cancellation on URL update.
- Prevents stale requests/flicker and reduces unnecessary network activity.

**Issue**
Fixes #7149


src/source/raster_tile_source.ts
```diff
@@ -163,6 +163,12 @@ export class RasterTileSource extends Evented implements Source {
      * @param url - A URL to a TileJSON resource. Supported protocols are `http:` and `https:`.
      */
     setUrl(url: string): this {
+        // Cancels all in-flight raster tile requests for this source
+        const tileManager = this.map?.style?.tileManagers?.[this.id];
```
HarelM: There has to be a better way, this assums too much about the internals of tile manager, map, etc.

madoci: Would adding an optional attribute to `MapSourceDataEvent` and firing a data event be better ? It will then be handled in `TileManager._dataHandler()`.

Also, it might be outside of this PR's scope, but wouldn't this benefit vector tile source as well ?

HarelM: Yeah, I think raising an event is probably a better approach.
If it can be done for vector tiles as well even better.

pcardinal:
Vector and raster tiles are adressed
If this works, the title of this issue may need to be changed.

src/tile/tile_manager.ts
```diff
@@ -798,19 +813,24 @@ export class TileManager extends Evented {
     private _dataHandler(e: MapSourceDataEvent) {
         if (e.dataType !== 'source') return;
 
-        if (e.sourceDataType === 'metadata') {
-            this._sourceLoaded = true;
+        const internal = e as MapSourceDataEvent & {abortPendingTileRequests?: boolean};
+
+        // Event interne d'abort (émis avant metadata/content)
```
HarelM: Please avoid non english comments...

src/tile/tile_manager.ts
```diff
@@ -798,19 +813,24 @@ export class TileManager extends Evented {
     private _dataHandler(e: MapSourceDataEvent) {
         if (e.dataType !== 'source') return;
 
-        if (e.sourceDataType === 'metadata') {
-            this._sourceLoaded = true;
+        const internal = e as MapSourceDataEvent & {abortPendingTileRequests?: boolean};
```
HarelM: This parameter should be part of the event type.

src/tile/tile_manager.ts
```diff
@@ -125,6 +125,21 @@ export class TileManager extends Evented {
...
 
+    abortAllRequests() {
+        if (this._inViewTiles) {
+            for (const id of this._inViewTiles.getAllIds()) {
+                const tile = this._inViewTiles.getTileById(id);
+                if (!tile) continue;
+
+                tile.aborted = true;
+                tile.abortController?.abort();
+                void (this._source as any)?.abortTile?.(tile);
```
HarelM: We should not cast to any. If this is required it should be part of the source API.

src/source/vector_tile_source.ts
```diff
@@ -245,12 +264,13 @@ export class VectorTileSource extends Evented implements Source {
         } catch (err) {
             delete tile.abortController;
 
-            if (tile.aborted) {
+            if (tile.aborted || isAbortError(err)) {
+                tile.state = 'expired';
```
HarelM: Setting tile state as expired when the tile request is aborted is confusing. Is there a better way to address this?

src/source/vector_tile_source.ts
```diff
@@ -229,14 +235,27 @@ export class VectorTileSource extends Evented implements Source {
                 tile.reloadPromise = {resolve, reject};
             });
         }
+
+        const restartQueuedReload = () => {
```
HarelM: Can this be a private method maybe?

src/source/vector_tile_source.ts
```diff
@@ -156,11 +153,19 @@ export class VectorTileSource extends Evented implements Source {
...
+
+        // Abort immédiat des requêtes de tuiles en vol
+        this.fire(new Event('data', {
+            dataType: 'source',
+            abortPendingTileRequests: true
+        } as any));
 
-        this.load(true);
+        callback();
+        this.load(true); // garder 1 argument
```
HarelM: English please.

src/source/vector_tile_source.ts
```diff
@@ -229,14 +235,27 @@ export class VectorTileSource extends Evented implements Source {
                 tile.reloadPromise = {resolve, reject};
             });
         }
+
+        const restartQueuedReload = () => {
+            if (!tile.reloadPromise) return;
+            const reloadPromise = tile.reloadPromise;
+            tile.reloadPromise = null;
+            this.loadTile(tile).then(reloadPromise.resolve).catch(reloadPromise.reject);
```
HarelM: Use async await?

src/source/raster_tile_source.ts
```diff
@@ -139,8 +139,13 @@ export class RasterTileSource extends Evented implements Source {
             this._tileJSONRequest = null;
         }
 
-        callback();
+        // Demande d'annulation des tuiles en vol, sans dépendre de map.style.tileManagers
+        this.fire(new Event('data', {
+            dataType: 'source',
+            abortPendingTileRequests: true
+        } as any));
```
HarelM: Avoid using "as any" if possible

src/source/raster_tile_source.ts
```diff
@@ -160,7 +165,7 @@ export class RasterTileSource extends Evented implements Source {
     /**
      * Sets the source `url` property and re-renders the map.
      *
-     * @param url - A URL to a TileJSON resource. Supported protocols are `http:` and `https:`.
+     * @param url -  A URL to a TileJSON resource. Supported protocols are `http:` and `https:`.
```
HarelM: Remove extra space.

HarelM:
Thanks for the time spent on this! 
I haven't reviewed the test code yet, let's make sure we address the production code comments first.

src/ui/events.ts
```diff
@@ -472,12 +472,13 @@ export type MapSourceDataEvent = MapLibreEvent & {
...
     tile: any;
+    shouldReloadTileOptions: GeoJSONSourceShouldReloadTileOptions;
 
     /**
-     * Options to determine whether a tile should be reloaded.
+     * Internal flag used by sources to request aborting in-flight tile requests.
      * @internal
      */
-    shouldReloadTileOptions: GeoJSONSourceShouldReloadTileOptions;
+    abortPendingTileRequests?: boolean;
```
HarelM, line 486: I'm not sure this is the same as `shouldReloadTileOptions`, is it? In previous code geojson source was the one using this in order to reload tiles since they were out dated, is this functionality kept in a different way?

src/tile/tile_manager.ts
```diff
@@ -798,15 +817,19 @@ export class TileManager extends Evented {
...
             return;
         }
 
-        if (e.sourceDataType !== 'content' || !this._sourceLoaded || this._paused) {
+        if (e.sourceDataType === 'metadata') {
+            this._sourceLoaded = true;
             return;
         }
 
+        if (e.sourceDataType !== 'content' || !this._sourceLoaded || this._paused) return;
```
HarelM: I suggest to keep the previous "{" to reduce code changes for easier review.

src/tile/tile_manager.ts
```diff
@@ -207,8 +225,9 @@ export class TileManager extends Evented {
     }
 
     _abortTile(tile: Tile) {
-        if (this._source.abortTile)
-            this._source.abortTile(tile);
+        if (this._source.abortTile) {
+            void this._source.abortTile(tile);
```
HarelM: What does `void` mean here?

src/tile/tile_cache.ts
```diff
@@ -209,6 +209,15 @@ export class TileCache {
             this.remove(r.value.tileID, r);
         }
     }
+    abortAllRequests() {
+        for (const key in this.data) {
```
HarelM: You can use `Object.entries` to go over all the entries.

src/tile/tile_cache.ts
```diff
@@ -209,6 +209,15 @@ export class TileCache {
...
+            for (const entry of this.data[key]) {
+                if (entry.value && entry.value.abortController) {
```
HarelM: You can use "?" operator.

src/source/vector_tile_source.ts
```diff
@@ -245,9 +260,12 @@ export class VectorTileSource extends Evented implements Source {
         } catch (err) {
             delete tile.abortController;
 
-            if (tile.aborted) {
+            if (tile.aborted || isAbortError(err)) {
+                tile.state = 'unloaded';
+                void this._restartQueuedReload(tile);
```
HarelM: Won't this cause a loop to try loading a tile again?
What will happen in case of zoom in/out where tiles are aborted since this are not needed anymore and not because the source data changed?

src/tile/tile_manager.ts
```diff
@@ -125,6 +125,24 @@ export class TileManager extends Evented {
         this._updated = false;
     }
 
+    abortAllRequests() {
+        if (this._inViewTiles) {
+            for (const id of this._inViewTiles.getAllIds()) {
```
HarelM: nit picking: The if can probably be removed by using `this._inViewTiles?.getAllIds() ?? []` I think, thus reducing indentation and complexity.

HarelM:
I haven't reviewed the tests yet as I want to make sure the business logic is correct first.
Once the code looks good I'll do a second review for the tests.

src/tile/tile_manager.ts
```diff
@@ -125,6 +125,27 @@ export class TileManager extends Evented {
...
+            const tile = this._inViewTiles.getTileById(id);
+            if (!tile) continue;
+
+            const hasInFlightRequest = tile.state === 'loading' || !!tile.abortController;
+            if (!hasInFlightRequest) continue;
+
+            tile.aborted = true;
+
+            if (this._source?.abortTile) {
+                void this._source.abortTile(tile);
```
HarelM: What does `void` mean here?

src/tile/tile_manager.ts
```diff
@@ -125,6 +125,27 @@ export class TileManager extends Evented {
...
+            } else {
+                tile.abortController?.abort();
+            }
+        }
+
+        this._outOfViewCache?.abortAllRequests?.();
```
HarelM: I'm not sure the second ?. is needed. Is there more than one implementation of outOfViewCache that won't have the newly added `abortAllRequests` method?

src/source/vector_tile_source.ts
```diff
@@ -287,14 +325,12 @@ export class VectorTileSource extends Evented implements Source {
 
         tile.loadVectorData(data, this.map.painter);
 
-        if (tile.reloadPromise) {
-            const reloadPromise = tile.reloadPromise;
-            tile.reloadPromise = null;
-            this.loadTile(tile).then(reloadPromise.resolve).catch(reloadPromise.reject);
-        }
+        void this._restartQueuedReload(tile);
```
HarelM: Remove `void`.

src/source/vector_tile_source.ts
```diff
@@ -275,6 +293,26 @@ export class VectorTileSource extends Evented implements Source {
         };
     }
 
+    private async _restartQueuedReload(tile: Tile): Promise<void> {
```
HarelM: Is there a better name to this method? `restartQueuedReload` does not explain what is restarted...

Also the comment below about restart is a bit vague...

HarelM:
This changes the internals of tile loading, which I think is risky I believe, and a part of the core of maplibre.
How confident are you that this fix does not introduce other issues related to tile (re)loading?

src/source/vector_tile_source.ts
pcardinal: change for _processQueuedReloadIfRetained

src/source/vector_tile_source.ts
pcardinal: removed

src/tile/tile_manager.ts
pcardinal: second ? removed

src/tile/tile_manager.ts
pcardinal: void removed

pcardinal:
_This changes the internals of tile loading, which I think is risky I believe, and a part of the core of maplibre.
How confident are you that this fix does not introduce other issues related to tile (re)loading?_

That's the billion-dollar question, I agree that tile loading is a sensitive core area, so I kept the fix intentionally narrow.

On the raster side, I am more confident then on the vectore side

The change is limited to two behaviors:

Cancel in-flight tile requests when source properties change, so stale requests cannot complete after a source switch.
Reset the tile aborted state before a new load, so valid retries are not blocked by a previous abort.
It does not change tile selection, cache sizing, or the normal scheduling flow.

To reduce risk, I added focused regression coverage around cancel/reload behavior and queued reload handling, and then ran the full test matrix:

pcardinal:
_This changes the internals of tile loading, which I think is risky I believe, and a part of the core of maplibre.
How confident are you that this fix does not introduce other issues related to tile (re)loading?_

That's the billion-dollar question, I agree that tile loading is a sensitive core area, so I kept the fix intentionally narrow.

On the raster side, I am more confident then on the vectore side

The change is limited to two behaviors:

Cancel in-flight tile requests when source properties change, so stale requests cannot complete after a source switch.
Reset the tile aborted state before a new load, so valid retries are not blocked by a previous abort.
It does not change tile selection, cache sizing, or the normal scheduling flow.

To reduce risk, I added focused regression coverage around cancel/reload behavior and queued reload handling, and then ran the full test matrix.

https://pcardinal.github.io/7149_raster_prod.html
https://pcardinal.github.io/7149_raster_dev.html

https://pcardinal.github.io/7149_vect_prod.html
https://pcardinal.github.io/7149_vect_dev.html

src/source/raster_tile_source.test.ts
```diff
@@ -30,7 +31,7 @@ describe('RasterTileSource', () => {
     });
 
     afterEach(() => {
-        server.restore();
+        vi.restoreAllMocks();
```
HarelM, line 39: Why was the `server.restore()` removed?

src/source/raster_tile_source.test.ts
```diff
@@ -339,4 +340,175 @@ describe('RasterTileSource', () => {
...
+        expect(tile.aborted).toBe(false);
+    });
+
+    test('setSourceProperty aborts in-flight TileJSON request, emits abortPendingTileRequests, then calls load(true)', () => {
+        const source = createSource({
+            tiles: ['http://example.com/{z}/{x}/{y}.png']
+        });
+
+        const requestAbort = vi.fn();
+        (source as any)._tileJSONRequest = {abort: requestAbort};
```
HarelM: Is there a better way to do it instead of manipulating the internal of the unit under test?

src/source/raster_tile_source.test.ts
```diff
@@ -339,4 +340,175 @@ describe('RasterTileSource', () => {
...
+        });
+
+        source.setSourceProperty(callback);
+
+        expect(requestAbort).toHaveBeenCalledTimes(1);
+        expect((source as any)._tileJSONRequest).toBeNull();
+        expect(callback).toHaveBeenCalledTimes(1);
+        expect(loadSpy).toHaveBeenCalledTimes(1);
+        expect(loadSpy).toHaveBeenCalledWith(true);
+        expect(sequence).toEqual(['abort-event', 'callback', 'load']);
```
HarelM: spys have a better way to handle this, it's better to use them instead of creating a parallel infrastructure.

src/source/raster_tile_source.test.ts
```diff
@@ -339,4 +340,175 @@ describe('RasterTileSource', () => {
...
+        });
+
+        const isAbortEvent = (e: any) => e?.abortPendingTileRequests === true || e?.data?.abortPendingTileRequests === true;
+        let sawAbortEvent = false;
+
+        source.on('data', (e: any) => {
+            if (isAbortEvent(e)) sawAbortEvent = true;
+        });
+
+        const loadSpy = vi.spyOn(source, 'load').mockResolvedValue(undefined as any);
```
HarelM: This manipulates the unit under test and will restrict future refactoring, is there a way to make sure this behavior is in place by using the "public" APIs?

src/source/vector_tile_source.test.ts
```diff
@@ -302,6 +302,81 @@ describe('VectorTileSource', () => {
...
+        const sendAsync = vi.fn((_message) => {
+            if (!resolveFirstRequest) {
+                return new Promise((resolve) => {
+                    resolveFirstRequest = resolve;
+                });
+            }
+            return Promise.resolve({});
+        });
+
+        source.dispatcher = getWrapDispatcher()({sendAsync});
```
HarelM: Not entirely related to this change, but I think it would be better to move the initialization of the dispatcher to the `createSource` method in this file.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### 🐞 Bug fixes
 - _...Add new stuff here..._
+- Fix: Add proper cancellation of in-flight raster/vector tile requests on setUrl ([#7149](https://github.com/maplibre/maplibre-gl-js/pull/7149))(by [@pcardinal](https://github.com/pcardinal))
```
```suggestion
- Fix: Add proper cancellation of in-flight raster/vector tile requests on setUrl ([#7149](https://github.com/maplibre/maplibre-gl-js/pull/7149)) (by [@pcardinal](https://github.com/pcardinal))
```

src/tile/tile_manager.ts
```diff
@@ -125,6 +125,27 @@ export class TileManager extends Evented {
         this._updated = false;
     }
 
+    abortAllRequests() {
+
+        for (const id of this._inViewTiles.getAllIds()) {
+            const tile = this._inViewTiles.getTileById(id);
+            if (!tile) continue;
+
+            const hasInFlightRequest = tile.state === 'loading' || !!tile.abortController;
```
HarelM: nit-picking: Can you use `tile.abortController != null` instead of `!!`?

src/tile/tile_manager.test.ts
```diff
@@ -2586,3 +2683,31 @@ describe('TileManager / etag', () => {
         expect(tile.etag).toBe(tileEtag);
     });
 });
+
+describe('TileManager.abortAllRequests', () => {
+    test('aborts requests from in-view and out-of-view tiles', () => {
```
HarelM: This test is not testing well the new functionality, it mocks too much...

src/tile/tile_manager.test.ts
HarelM, line 1: The tests here are using too many internal fields and non public APIs, I suggest to rewrite them to use public APIs and mock a lot less than what they are mocking right now.

src/tile/tile_cache.ts
```diff
@@ -209,6 +209,13 @@ export class TileCache {
             this.remove(r.value.tileID, r);
         }
     }
+    abortAllRequests() {
```
HarelM, line 212: Does this need to change the state of the tile?

src/tile/tile_cache.test.ts
```diff
@@ -188,4 +188,33 @@ describe('BoundedLRUCache', () => {
...
-});
\ No newline at end of file
+
+    describe('TileCache#abortAllRequests', () => {
+        test('aborts all entries that have an abortController', () => {
+            const cache = new TileCache(10, () => {});
+            const abortA = vi.fn();
+            const abortB = vi.fn();
+
+            (cache as any).data = {
```
HarelM: Don't use `as any`, instead use the public APIs of this class.

src/source/vector_tile_source.ts
```diff
@@ -220,23 +233,46 @@ export class VectorTileSource extends Evented implements Source {
...
         let messageType: MessageType.loadTile | MessageType.reloadTile = MessageType.reloadTile;
-        if (!tile.actor || tile.state === 'expired') {
+        if (!tile.actor || tile.state === 'expired' || tile.state === 'unloaded') {
             tile.actor = this.dispatcher.getActor();
             messageType = MessageType.loadTile;
         } else if (tile.state === 'loading') {
+            const pendingReload = tile.reloadPromise;
             return new Promise<void>((resolve, reject) => {
-                tile.reloadPromise = {resolve, reject};
+                if (pendingReload) {
```
HarelM: I'm not sure I understand this code, what is it meant for?

src/source/vector_tile_source.ts
```diff
@@ -27,6 +27,10 @@ export type LoadTileResult = {
     unmodified?: boolean;
 };
 
+function hasRenderableState(tile: Tile): boolean {
```
HarelM: This looks like a method that should be part of the Tile class instead of here...?

src/source/vector_tile_source.test.ts
```diff
@@ -488,4 +592,155 @@ describe('VectorTileSource', () => {
...
+        expect(tile.state).toBe('expired');
+        expect(abort).toHaveBeenCalledTimes(1);
+        expect(tile.abortController).toBeUndefined();
+        expect(sendAsync).toHaveBeenCalledWith({
+            type: MessageType.abortTile,
+            data: {uid: 45, type: source.type, source: source.id}
+        });
+    });
+
+    test('abortTile sets tile state to unloaded and aborts controller even when actor is missing', async () => {
```
HarelM: Is missing actor interesting in any way? When can it really happen that actor is missing?

src/source/vector_tile_source.ts
HarelM: It's still not clear what this means... sorry...

HarelM:
I've went over the tests and they have a lot of fixes needed, also the coverage can probably be improved.
The logic itself is still a bit confusing to read.

src/source/raster_tile_source.test.ts
pcardinal: Yes, that was a test-cleanup regression. I restored fake server cleanup while keeping Vitest mock cleanup.

src/source/raster_tile_source.test.ts
pcardinal: I removed the tests that were forcing _tileJSONRequest and asserting on _options, and replaced them with checks based on public effects (data event + callback/load order via invocationCallOrder) in raster_tile_source.test.ts:383 and vector_tile_source.test.ts:600.
I also centralized the abortPendingTileRequests event predicate with a helper in raster_tile_source.test.ts:12 and vector_tile_source.test.ts:17.

src/source/raster_tile_source.test.ts
pcardinal: Done. Ordering assertions now use invocationCallOrder instead of a manual sequence array in raster_tile_source.test.ts:395 and vector_tile_source.test.ts:612.

src/source/vector_tile_source.test.ts
pcardinal: Done. The helper now accepts a custom actor to inject sendAsync at source creation time, instead of reassigning source.dispatcher in the targeted test, in vector_tile_source.test.ts:21, with usage at vector_tile_source.test.ts:310.

src/tile/tile_manager.ts
pcardinal: Done, replaced with an explicit null/undefined check in tile_manager.ts:133.

src/tile/tile_manager.test.ts
pcardinal: Rewritten: no more ad-hoc object passed to the prototype; we now test a real TileManager instance with real test tiles in tile_manager.test.ts:2687.

src/tile/tile_cache.ts
pcardinal: I aligned abort semantics by setting tile.aborted=true before abortController.abort in tile_cache.ts:212.
Tests were rewritten using public cache APIs (add), without overwriting cache.data, in tile_cache.test.ts:193.

src/source/vector_tile_source.test.ts
pcardinal: I removed it along with the other no-op test, since they are not very relevant to the business behavior targeted by this change. The useful retained test is the one preserving renderable state for expired tiles in vector_tile_source.test.ts:668.

src/tile/tile_cache.test.ts
pcardinal: Done

src/source/vector_tile_source.ts
pcardinal: I clarified intent by splitting it explicitly into reload queuing in vector_tile_source.ts:303 and conditional draining of queued reload in vector_tile_source.ts:326, and updated call sites in vector_tile_source.ts:242.

src/source/raster_tile_source.test.ts
pcardinal: const loadSpy = vi.spyOn(source, 'load');

▸ Simple spy without mock → lets load() execute for real
▸ Verifies that load() was called with the correct arguments
▸ Allows the actual logic of load() to run and emit its own events
▸ No restrictions on future refactorings

src/source/vector_tile_source.ts
pcardinal: I moved the renderable-state rule into the Tile model so this behavior is owned in one place instead of inside the source. The vector source now uses that Tile-level helper instead of a local function, and Tile.hasData now delegates to the same logic to keep semantics consistent. I implemented it as a static Tile helper so existing tests that use tile-like objects (not full Tile instances) remain compatible. I also reran the targeted vector-source unit tests, and they pass.

New test 'Tile renderable states' in src/tile/tile.test.ts

src/tile/tile_manager.test.ts
pcardinal: see your following remark

src/source/vector_tile_source.ts
pcardinal: removed

src/source/raster_tile_source.test.ts
HarelM: Did you forget to commit it?

pcardinal:
I didn’t push everything yet. 

 

De : Harel M ***@***.***> 
Envoyé : 14 mars 2026 11:27
À : maplibre/maplibre-gl-js ***@***.***>
Cc : pcardinal ***@***.***>; Author ***@***.***>
Objet : Re: [maplibre/maplibre-gl-js] Fix(raster): cancel in-flight raster tile requests when source URL changes - issue #7149 (PR #7214)

 

@HarelM commented on this pull request.

  _____  

In src/source/raster_tile_source.test.ts <https://github.com/maplibre/maplibre-gl-js/pull/7214#discussion_r2935405777> :

> @@ -30,7 +31,7 @@ describe('RasterTileSource', () => {
     });
 
     afterEach(() => {
-        server.restore();
+        vi.restoreAllMocks();

Did you forget to commit it?

pcardinal:
"Apologies for the force-push. I had some sync issues with the changelog, but the branch is now clean and includes the core fix along with the new unit tests for raster and vector sources."

src/tile/tile_manager.test.ts
```diff
@@ -188,6 +188,37 @@ describe('TileManager.addTile', () => {
 
     });
 
+    test('falls back to cache miss after out-of-view cache reset', () => {
+        const tileID = new OverscaledTileID(0, 0, 0, 0, 0);
+        let load = 0;
+        let unload = 0;
```
HarelM: Is it possible to use `vi.fn` and check for `toHaveBeenCalledTimes(...)`?
Would be great to do it for the loadTile as well.

src/tile/tile.ts
```diff
@@ -146,6 +146,10 @@ export class Tile {
         this.state = 'loading';
     }
 
+    static hasRenderableState(tile: {state: TileState}): boolean {
```
HarelM: This looks like a duplicate of the `hadData` method, so I don't think it's needed, simply use `hadData()`.

src/tile/tile.test.ts
```diff
@@ -200,6 +200,26 @@ describe('Tile.isLessThan', () => {
     });
 });
 
+describe('Tile renderable states', () => {
+    test('hasRenderableState and hasData follow the same state table', () => {
```
HarelM: I'm guessing this is also redundant if the other method is removed...

src/source/vector_tile_source.ts
```diff
@@ -287,14 +352,14 @@ export class VectorTileSource extends Evented implements Source {
...
-            const reloadPromise = tile.reloadPromise;
-            tile.reloadPromise = null;
-            this.loadTile(tile).then(reloadPromise.resolve).catch(reloadPromise.reject);
-        }
+        this._drainQueuedReloadIfTileRetained(tile);
     }
 
     async abortTile(tile: Tile): Promise<void> {
+        tile.aborted = true;
+        if (!Tile.hasRenderableState(tile)) {
```
HarelM: use `!tile.hasData()` instead.

src/source/raster_tile_source.test.ts
```diff
@@ -339,4 +345,148 @@ describe('RasterTileSource', () => {
         await expect(loadPromise).resolves.toBeUndefined();
         expect(tile.state).toBe('unloaded');
     });
+
+    test('loads tile after previous abort flag was set', async () => {
```
HarelM, line 412: I don't see an abort flag in the test, am I missing anything?

src/source/raster_tile_source.test.ts
```diff
@@ -339,4 +345,148 @@ describe('RasterTileSource', () => {
...
+
+        const callback = vi.fn();
+        const loadSpy = vi.spyOn(source, 'load');
+
+        source.setSourceProperty(callback);
+
+        const abortEventIndex = onData.mock.calls.findIndex(([e]) => isAbortPendingTileRequestsEvent(e));
+        expect(abortEventIndex).toBeGreaterThan(-1);
+
+        const abortEventOrder = onData.mock.invocationCallOrder[abortEventIndex];
```
HarelM: Looks like a lot of logic in this test is about call order, is it that important to make sure the call order is kept in the future so that this test will be meaningful?

src/source/raster_tile_source.test.ts
```diff
@@ -339,4 +345,148 @@ describe('RasterTileSource', () => {
...
+
+        source.setUrl('http://localhost:2900/source2.json');
+
+        expect(
+            events.some((e) => isAbortPendingTileRequestsEvent(e))
+        ).toBe(true);
+        expect(loadSpy).toHaveBeenCalledWith(true);
+    });
+
+    test('setUrl emits abortPendingTileRequests and updates url (no TileManager internals)', () => {
```
HarelM: Can an expect added to the previous test to cover this test as well?

src/source/raster_tile_source.test.ts
```diff
@@ -339,4 +345,148 @@ describe('RasterTileSource', () => {
...
+        source.setUrl('http://localhost:2900/source2.json');
+
+        expect(
+            events.some((e) => isAbortPendingTileRequestsEvent(e))
+        ).toBe(true);
+        expect(source.url).toBe('http://localhost:2900/source2.json');
+        expect(loadSpy).toHaveBeenCalledWith(true);
+    });
+
+    test('load emits error event on TileJSON network error (non-abort)', async () => {
```
HarelM: Is this test related to the change in this PR? I don't mind the additional test, but I'm not sure I understand why it was added...

src/tile/tile.ts
pcardinal: This was redundant with hasData, so I removed hasRenderableState and switched the call sites to use hasData directly. I also removed the redundant test.

src/tile/tile.test.ts
pcardinal: You are right, this test is redundant once hasRenderableState is removed. I removed the Tile renderable states test and kept hasData as the single source of truth for renderable states.

src/source/vector_tile_source.ts
pcardinal: Good point, I agree. I replaced the check with !tile.hasData() in abortTile, so we rely on the existing Tile instance method instead of the redundant static helper.

src/source/raster_tile_source.test.ts
pcardinal: Good catch. The abort flag is set in the test fixture on the tile object (aborted: true) before calling loadTile, but that was not very explicit. I updated the test to assert the initial flag (expect(tile.aborted).toBe(true)) and then verify it is reset to false after the load succeeds.

src/source/raster_tile_source.test.ts
pcardinal: Yes, absolutely. I merged the coverage by adding expect(source.url).toBe('http://localhost:2900/source2.json') to the previous setUrl test, then removed the redundant second test.

src/source/raster_tile_source.test.ts
pcardinal: You are right, that test is not directly related to the change in this PR. I had added it while touching the TileJSON load path, but it expands the scope without helping validate the abort-related behavior here, so I removed it to keep the PR focused.

src/source/raster_tile_source.test.ts
pcardinal: Yes, for this path the order matters. We need abortPendingTileRequests before callback and load(true) to avoid stale-request races, so keeping the order check makes this test meaningful.

src/tile/tile_manager.test.ts
pcardinal: Thanks, great suggestion. I updated the test to use vi.fn spies for both loadTile and unloadTile, and replaced the manual counters with toHaveBeenCalledTimes assertions. This now checks loadTile call counts as well (1 before cache reset, then 2 after the reset path).

src/ui/events.ts
pcardinal: No, they are not the same.

shouldReloadTileOptions is still the GeoJSON-specific selective reload path:

computed in maplibre-gl-js/src/source/geojson_source.ts
sent on the content event in maplibre-gl-js/src/source/geojson_source.ts
forwarded to TileManager.reload in maplibre-gl-js/src/tile/tile_manager.ts
used to filter which tiles reload in maplibre-gl-js/src/tile/tile_manager.ts, via GeoJSONSource.shouldReloadTile in maplibre-gl-js/src/source/geojson_source.ts
abortPendingTileRequests is a separate signal for immediate cancellation of in-flight requests when source properties change:

emitted in maplibre-gl-js/src/source/vector_tile_source.ts and maplibre-gl-js/src/source/raster_tile_source.ts
handled in maplibre-gl-js/src/tile/tile_manager.ts
So the previous GeoJSON stale-tile reload behavior is still kept. This change adds an earlier abort phase for in-flight requests.

src/source/raster_tile_source.test.ts
HarelM: Why not simply set it to `true` instead of asserting? Or have the server response with an abort response?

src/source/raster_tile_source.test.ts
HarelM: I would recommend using the same spy then and making sure the relevant class are in place. This is a unit test, there should not be a lot of events and you should be able to simply asset on which invocation happened when explicitly. This code is still too complicated for this unit test...

src/source/raster_tile_source.test.ts
```diff
@@ -339,4 +345,118 @@ describe('RasterTileSource', () => {
...
+        expect(loadSpy).toHaveBeenCalledWith(true);
+    });
+
+    test('load ignores AbortError from TileJSON request', async () => {
+        const source = createSource({
+            tiles: ['http://example.com/{z}/{x}/{y}.png']
+        });
+
+        const abortError = new Error('aborted');
+        (abortError as any).name = 'AbortError';
```
HarelM: Is this `as any` really needed? I tried to assign a name for an error and I didn't get any typescript errors...?

src/source/raster_tile_source.test.ts
```diff
@@ -339,4 +345,118 @@ describe('RasterTileSource', () => {
...
+
+        expect(onError).not.toHaveBeenCalled();
+    });
+
+    test('load emits error event when TileJSON is malformed (parser rejection)', async () => {
+        const source = createSource({
+            tiles: ['http://example.com/{z}/{x}/{y}.png']
+        });
+
+        vi.spyOn(loadTileJSONModule, 'loadTileJson').mockRejectedValueOnce(new Error('Invalid TileJSON payload'));
```
HarelM: Is it possible to return an invalid tileJson as a server response instead of mocking this method?

src/source/vector_tile_source.test.ts
```diff
@@ -37,6 +42,12 @@ function createSource(options, transformCallback?, clearTiles = () => {}) {
     return source;
 }
 
+function withHasData<T extends {state?: string}>(tile: T): T & {hasData: () => boolean} {
```
HarelM: This isn't a great method name, how about `createTile` or `createMockTile`?

src/source/vector_tile_source.test.ts
```diff
@@ -37,6 +42,12 @@ function createSource(options, transformCallback?, clearTiles = () => {}) {
...
+    const tileWithHasData = tile as T & {hasData: () => boolean};
+    tileWithHasData.hasData = () => tile.state === 'loaded' || tile.state === 'reloading' || tile.state === 'expired';
```
HarelM: I'm not entirely sure this is needed, can't you simply override the relevant behavior of this method's return value when you need to in the relevant test?

src/source/vector_tile_source.test.ts
HarelM: Look like the usage of this method is to create an object for the `loadTile` method, and there's a lot of `as any as Tile` in the tests that can be removed if this method had a better return value...

src/source/vector_tile_source.test.ts
```diff
@@ -302,6 +314,80 @@ describe('VectorTileSource', () => {
...
+        const initialLoad = source.loadTile(tile);
+        const queuedReloadA = source.loadTile(tile);
+        const queuedReloadB = source.loadTile(tile);
+
+        resolveFirstRequest({});
+
+        await initialLoad;
+        await Promise.race([
+            Promise.all([queuedReloadA, queuedReloadB]),
+            new Promise((_, reject) => setTimeout(() => reject(new Error('queued reload promises did not resolve')), 200))
```
HarelM: It's not clear what this test is checking. Is it checking that only one of the two is resolved? Is there a different way to check this and make the test more readable?

src/source/vector_tile_source.test.ts
```diff
@@ -476,16 +591,104 @@ describe('VectorTileSource', () => {
...
+        });
+
+        const loadSpy = vi.spyOn(source, 'load').mockResolvedValue(undefined as any);
+        const callback = vi.fn();
+        const onData = vi.fn();
+
+        source.on('data', onData);
+        source.setSourceProperty(callback);
+
+        const abortEventIndex = onData.mock.calls.findIndex(([e]) => isAbortPendingTileRequestsEvent(e));
```
HarelM: See my comment on the other test. Consider simplifying it.

src/source/vector_tile_source.test.ts
```diff
@@ -476,16 +591,104 @@ describe('VectorTileSource', () => {
...
+
+        expect(abort).toHaveBeenCalledTimes(1);
+        expect(tile.abortController).toBeUndefined();
+        expect(sendAsync).toHaveBeenCalledWith({
+            type: MessageType.abortTile,
+            data: {uid: 42, type: source.type, source: source.id}
+        });
+    });
+
+    test('abortTile sets tile state to unloaded, aborts controller, and sends MessageType.abortTile', async () => {
```
HarelM, line 697: Can this test and `abortTile aborts controller and sends MessageType.abortTile` be merged to a single test? Seems like they are similar enough, or aren't they? If they are not, then the test name should be more specific.

src/source/vector_tile_source.ts
```diff
@@ -245,9 +261,14 @@ export class VectorTileSource extends Evented implements Source {
         } catch (err) {
             delete tile.abortController;
 
-            if (tile.aborted) {
+            if (tile.aborted || isAbortError(err)) {
+                if (!wasRenderableBeforeLoad) {
```
HarelM, line 265: Looks like this condition is checked in both cases before calling `_drainQueuedReloadIfTileRetained`, does it makes sense to include it in the method?

HarelM: Never mind, I see another invocation of this method.

src/source/vector_tile_source.ts
```diff
@@ -275,6 +296,50 @@ export class VectorTileSource extends Evented implements Source {
...
+    private _enqueueReloadForLoadingTile(tile: Tile): Promise<void> {
+        const previousQueuedReload = tile.reloadPromise;
+        return new Promise<void>((resolve, reject) => {
+            if (!previousQueuedReload) {
+                tile.reloadPromise = {resolve, reject};
+                return;
+            }
+
+            // Keep all callers waiting for a reload while this tile is already loading.
+            // Resolve/reject cascades so every queued caller settles when the queued reload completes.
```
HarelM: This is like the 3-5 time I'm reading this and it still isn't clear what's happening here.
I think the comment is explaining it , but I still find this code combersome.
Is there a better way to manage a queue instead of this obscure promise inside promise thing?
It's also might be better to change the name of  `tile.reloadPromise` to something like `reloadResolveReject` as it's not really promise but something that can resolve or reject a promise, and I think it adds to the confusion (to be fair I think I gave it that "bad" name).

src/source/vector_tile_source.ts
```diff
@@ -287,14 +352,14 @@ export class VectorTileSource extends Evented implements Source {
...
-        if (tile.reloadPromise) {
-            const reloadPromise = tile.reloadPromise;
-            tile.reloadPromise = null;
-            this.loadTile(tile).then(reloadPromise.resolve).catch(reloadPromise.reject);
-        }
+        this._drainQueuedReloadIfTileRetained(tile);
     }
 
     async abortTile(tile: Tile): Promise<void> {
+        tile.aborted = true;
```
HarelM, line 347: I wonder why aborted is not a state of the tile and instead it's a boolean field, but that is probably outside the scope of this PR...

src/source/vector_tile_source.ts
```diff
@@ -287,14 +352,14 @@ export class VectorTileSource extends Evented implements Source {
...
+        if (!tile.hasData()) {
+            tile.state = 'unloaded';
```
HarelM, line 349: Wouldn't this be a reaction to the below abort call from the abort controller? Is this mandatory to set this?

src/source/raster_tile_source.test.ts
HarelM: If there are other tests that are outside the scope of this PR let's move them to a different PR to add coverage, I would still like to have them if they are testing some scenarios that we don't want to break by mistake.

src/tile/tile_manager.test.ts
```diff
@@ -188,6 +188,35 @@ describe('TileManager.addTile', () => {
 
     });
 
+    test('falls back to cache miss after out-of-view cache reset', () => {
```
HarelM: I think this is another test not related to this PR, right?

src/tile/tile_manager.test.ts
```diff
@@ -347,6 +376,25 @@ describe('TileManager.removeTile', () => {
 
     });
 
+    test('calls abortTile before unloadTile for unfinished tile', () => {
+        const tileID = new OverscaledTileID(0, 0, 0, 0, 0);
+        const calls: string[] = [];
```
HarelM, line 352: Same as previous comment with the spies.

src/tile/tile_manager.test.ts
```diff
@@ -515,6 +563,53 @@ describe('TileManager / Source lifecycle', () => {
...
+        const tileManager = createTileManager({});
+        const abortAllSpy = vi.spyOn(tileManager, 'abortAllRequests');
+        const reloadSpy = vi.spyOn(tileManager, 'reload').mockImplementation(() => {});
+        const updateSpy = vi.spyOn(tileManager, 'update').mockImplementation(() => undefined as any);
+
+        tileManager._sourceLoaded = true;
+        tileManager._paused = false;
+        tileManager.transform = new MercatorTransform() as any;
+
+        (tileManager as any)._dataHandler({dataType: 'source', abortPendingTileRequests: true});
```
HarelM: Please use public APIs only, this `as any` is a sign of a bad test.

src/tile/tile_manager.test.ts
```diff
@@ -515,6 +563,53 @@ describe('TileManager / Source lifecycle', () => {
         expect(tileManager.loaded()).toBeFalsy();
     });
 
+    test('ignores content events after abortPendingTileRequests until metadata arrives', () => {
+        const tileManager = createTileManager({});
+        const abortAllSpy = vi.spyOn(tileManager, 'abortAllRequests');
```
HarelM: I'm not sure that's the right way to test how this class behaves. It's best to either mock other classes this class uses or check events and other public APIs.

src/tile/tile_manager.test.ts
```diff
@@ -515,6 +563,53 @@ describe('TileManager / Source lifecycle', () => {
...
+        expect(updateSpy).toHaveBeenCalledTimes(0);
+
+        (tileManager as any)._dataHandler({dataType: 'source', sourceDataType: 'metadata'});
+        (tileManager as any)._dataHandler({dataType: 'source', sourceDataType: 'content', sourceDataChanged: true});
+
+        expect(reloadSpy).toHaveBeenCalledTimes(1);
+        expect(updateSpy).toHaveBeenCalledTimes(1);
+    });
+
+    test('forwards sourceDataChanged and shouldReloadTileOptions to reload', () => {
```
HarelM, line 571: See my previous comment, this test mocks too much of the class under test.

src/tile/tile_manager.test.ts
```diff
@@ -2586,3 +2698,42 @@ describe('TileManager / etag', () => {
...
+describe('TileManager.abortAllRequests', () => {
+    test('aborts in-flight in-view requests and delegates to out-of-view cache', () => {
+        const tileManager = createTileManager();
+        tileManager.onAdd(undefined);
+
+        const loadingTile = tileManager.addTile(new OverscaledTileID(1, 0, 1, 0, 1));
+        const nonLoadingTile = tileManager.addTile(new OverscaledTileID(1, 0, 1, 1, 1));
+        nonLoadingTile.state = 'loaded';
+
+        const abortTileSpy = vi.spyOn(tileManager._source, 'abortTile');
```
HarelM: I would feel better to simply check the state of the tiles and avoid making sure some internal methods are called, as those can change in the future.

src/tile/tile_manager.test.ts
```diff
@@ -2586,3 +2698,42 @@ describe('TileManager / etag', () => {
...
+
+    test('aborts tiles that expose an abortController even when not loading', () => {
+        const tileManager = createTileManager();
+        tileManager.onAdd(undefined);
+
+        const tile = tileManager.addTile(new OverscaledTileID(2, 0, 2, 0, 0));
+        tile.state = 'loaded';
+        tile.abortController = {abort: vi.fn()} as unknown as AbortController;
+
+        const abortTileSpy = vi.spyOn(tileManager._source, 'abortTile');
```
HarelM: If the outcome is that the tile is aborted by calling `abortAllRequests` I think that's a good enough test, no need to drill down to the exact method that was called.

src/source/raster_tile_source.test.ts
pcardinal: Good point, and I updated the test in that direction.

Instead of setting aborted to true and then asserting the same value, the test now follows the real runtime sequence:

Start a first tile load.
Abort it in flight through the tile AbortController (and mark aborted as true, as happens in runtime).
Verify the tile ends in unloaded state.
Start a second load and let the server return 200.
Verify the tile loads successfully and aborted is reset to false.
I did not use a server-side “abort response” because this abort path is client-side cancellation, not an HTTP error contract. Simulating the in-flight client abort is closer to production behavior and gives better regression coverage.
Also, I used tile.aborted = true manually to stay consistent with the existing testing patterns in this file.

src/source/raster_tile_source.test.ts
pcardinal: Good point. I've simplified both tests by spying on [source.fire](vscode-file://vscode-app/c:/Users/pierr/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html) directly instead of using a separate [source.on('data', ...)](vscode-file://vscode-app/c:/Users/pierr/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html) listener. This gives a single, uniform spy for all calls.

Since [setSourceProperty](vscode-file://vscode-app/c:/Users/pierr/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html) always fires the abort event as the very first call (synchronously), we can access it directly at calls[0] and compare invocationCallOrder by position — no findIndex, no some(), no helper needed. The helper isAbortPendingTileRequestsEvent was only used in these two tests and has been removed.

One thing worth noting: [load()](vscode-file://vscode-app/c:/Users/pierr/AppData/Local/Programs/Microsoft%20VS%20Code/ce099c1ed2/resources/app/out/vs/code/electron-browser/workbench/workbench.html) fires dataloading synchronously as its first line, so fireSpy ends up with two calls total — I've added a comment to make that expectation explicit rather than silently relying on it.

src/source/raster_tile_source.test.ts
pcardinal: You’re right, the cast was unnecessary. I updated the test to remove as any and assign the name directly on the Error instance (abortError.name = 'AbortError'). This keeps the test strictly typed and avoids an unnecessary type escape.

src/source/raster_tile_source.test.ts
pcardinal: Yes, it is possible, and I implemented it that way.

I replaced the mocked parser rejection with a real malformed TileJSON HTTP response in:
maplibre-gl-js/src/source/raster_tile_source.test.ts

What changed:

Added a server handler that returns:
a valid TileJSON on the first request (initial onAdd load),
an invalid JSON body on the second request to trigger parser failure:
maplibre-gl-js/src/source/raster_tile_source.test.ts
Switched source creation to URL-based loading:
maplibre-gl-js/src/source/raster_tile_source.test.ts
Waited for initial metadata load, then called load(true), responded from server, and asserted the error event:
maplibre-gl-js/src/source/raster_tile_source.test.ts
maplibre-gl-js/src/source/raster_tile_source.test.ts
Validation:

Ran: npm run test-unit -- raster_tile_source.test.ts
Result: 1 test file passed, 20/20 tests passed.

src/source/vector_tile_source.test.ts
pcardinal: Implemented both suggestions.

Renamed the helper from withHasData to createMockTile to better reflect its purpose as a test tile factory.
Updated the helper to return a Tile-typed mock instead of a narrower ad hoc shape, which allowed removal of the redundant as any as Tile casts in the loadTile tests.
Also adjusted the abort-related mocks to satisfy the AbortController type correctly, so the test file remains type-safe after the cleanup.
The targeted unit test file still passes after these changes.

src/source/vector_tile_source.test.ts
pcardinal: hasData() is called at the very start of loadTile() on every tile passed to it: 
`const wasRenderableBeforeLoad = tile.hasData();`

and again inside abortTile(). Because plain object mocks don't inherit from Tile, hasData would be undefined on every mock tile without this helper — calling loadTile would throw at runtime rather than testing anything meaningful.

The helper avoids repeating hasData: () => state === 'loaded' || ... on all 16 call sites. If you'd prefer the logic inlined on each mock, that's valid — but since the behaviour is tied to state (which each mock already sets), the shared function keeps that coupling in one place and scales better if hasData's semantics ever change.

src/source/vector_tile_source.test.ts
pcardinal: You are right that the original version was hard to read.
The intent is not to check that only one of the two queued promises resolves. It is checking two things:

1. both queued callers resolve
2. only one worker reload is actually executed for those queued calls

In other words, while the first load is in flight, multiple reload requests are coalesced into a single follow-up reload, and all waiting callers are resolved when that reload completes.

To make this clearer, I rewrote the test to be deterministic:

- control the first request and the queued reload with explicit resolvers
- assert that queued promises are still pending after the first request resolves
- then resolve the queued reload and assert both queued promises resolve
- keep the sendAsync call count assertion at 2 (initial load + one queued reload)

This removes the timeout-based Promise.race pattern and makes the behavior under test explicit.

src/source/vector_tile_source.test.ts
pcardinal: I simplified it to match the other test. Instead of listening for data and scanning the recorded events, it now spies on fire directly, asserts that the first emitted event is the abortPendingTileRequests signal, and checks the call order against the callback and load(true). This keeps the test narrower and easier to read.

src/source/vector_tile_source.test.ts
pcardinal: Done — the two tests were identical except that the second one also asserted tile.state === 'unloaded'. The first was therefore a strict subset and has been removed. The single remaining test now covers all three behaviors at once: controller abort, abortController cleared on tile, and MessageType.abortTile sent to the worke

src/source/vector_tile_source.ts
pcardinal: Thanks for calling this out. I agree the previous implementation was hard to read.

I refactored the reload-while-loading flow to use an explicit queue of waiters instead of chaining resolve/reject callbacks through nested promises. This makes the behavior easier to follow:

- when a tile is already loading, each caller is added to a queued waiters list
- when loading finishes, we run at most one queued reload
- then we resolve or reject all queued callers together

I also renamed the tile field from reloadPromise to queuedReloadWaiters, since it stores resolver pairs rather than an actual Promise.

Behavior is unchanged, but the control flow and naming are now much clearer. I validated this with the targeted unit tests for both:

- single reload while loading
- multiple queued reload callers resolved by one queued worker reload

src/source/vector_tile_source.ts
pcardinal: aborted is a boolean (not a TileState) because it represents a different axis of information than state:

1. state (state: TileState;) models data/render lifecycle, not request cancellation
It answers: “what data status does this tile currently have?” (loading, loaded, reloading, unloaded, errored, expired) in tile.ts:47.
  
2. aborted (aborted: boolean;) is a transient control flag for in-flight async work
It answers: “should the current async load result be ignored/cancelled?” It is checked right after worker response/error handling in vector_tile_source.ts:248 and vector_tile_source.ts:264, and set in abort path at vector_tile_source.ts:347.

3. A tile can keep a meaningful render state while a request is aborted
For example, when reloading an already renderable tile, aborting should not necessarily force a new lifecycle state; it may keep prior usable data and just stop/ignore the current fetch. The code explicitly only forces unloaded in some cases (e.g. when it had no renderable data before), see vector_tile_source.ts:249 and vector_tile_source.ts:265.

4. This pattern is used across the manager as request-level metadata
`(method) TileManager.abortAllRequests(): void` marks `tile.aborted = true` independently of lifecycle state in tile_manager.ts:136, which reinforces that cancellation is orthogonal to the main tile state machine.

So in short: state is “what the tile is,” aborted is “what to do with the current request.” Keeping them separate avoids state explosion and preserves correct rendering behavior during cancellations.

src/source/vector_tile_source.ts
pcardinal: Setting `tile.state = 'unloaded'` here is intentional and complementary to `AbortController.abort()`. The controller only cancels the in-flight async work; it does not guarantee the tile lifecycle state is updated consistently. Marking non-renderable tiles as unloaded ensures we do not keep a loading/aborted tile in an inconsistent state, while preserving renderable tiles (hasData()) to avoid flicker.

src/tile/tile_manager.test.ts
pcardinal: Removed. This test seems unrelated to the PR scope since it exercises TileManager cache reset behavior on reload, not the abortPendingTileRequests path.

src/tile/tile_manager.test.ts
pcardinal: which comment ?

src/tile/tile_manager.test.ts
pcardinal: I cleaned up the two neighboring unit tests to use public APIs only and removed the private/internal access patterns introduced in this thread.

**What was changed:**

1. Refactored the test at tile_manager.test.ts:537 to stop calling private handlers and stop using internal state mutation.
2. Refactored the adjacent test at tile_manager.test.ts:571 with the same approach.
3. Replaced direct private calls and state poking with public behavior via source events:

      - onAdd(...)
      - update(transform)
      - getSource().fire(new Event('data', ...))

**Validation:**

1. Targeted test run for “ignores content events after abortPendingTileRequests until metadata arrives” passed.
2. Targeted test run for “forwards sourceDataChanged and shouldReloadTileOptions to reload” passed.


**Scope note:**

I only changed code authored in this conversation, as requested, and left other existing as any usages untouched.

src/tile/tile_manager.test.ts
pcardinal: Thanks for the feedback. I updated the two latest tests to avoid asserting TileManager internals and to focus on observable behavior.

- test('ignores content events after abortPendingTileRequests until metadata arrives'
- test('forwards sourceDataChanged and shouldReloadTileOptions to reload




What changed:

1. Removed spies on TileManager methods (abortAllRequests, reload, update).
2. Kept event-driven setup through public source data events.
3. Asserted behavior through public/collaborator effects instead:

    - loaded() transitions after abortPendingTileRequests
    - reload behavior inferred from Source.loadTile call count changes
    - forwarding of shouldReloadTileOptions verified via Source.shouldReloadTile arguments

I reran the two updated tests, and both pass.

src/tile/tile_manager.test.ts
pcardinal: see previous comments R568 and R576

src/tile/tile_manager.test.ts
pcardinal: Done! I've refactored both tests to remove the internal spy assertions and focus only on observable behavior—the tile states.

The tests now:

1. First test: Directly assert that loadingTile.aborted is true and nonLoadingTile.aborted is not—checking only the public state changes
2. Second test: Assert that tile.aborted is true after calling abortAllRequests()—verifying the end result without coupling to internal method calls

This approach follows the project guidelines : "Prefer assertions on public behavior (emitted events, public state, return values) over private/internal method calls." The tests are now more resilient to future implementation refactors.

src/tile/tile_manager.test.ts
pcardinal: see previous comment "Comment on line R2711"

src/source/raster_tile_source.test.ts
pcardinal: Thanks for the feedback — done.

Three changes were made to keep the tests in scope:

1. setUrl test — public assertion instead of internal spy: replaced the fire spy (which inspects an internal method) with a source.on('data', ...) listener and a typed helper isAbortPendingTileRequestsEvent. This asserts the same behaviour through the public event API, consistent with the project's test guidelines.

2. Last test renamed and refocused: 'load emits error event when TileJSON is malformed (parser rejection)' has been replaced with 'load emits error event on TileJSON network error (non-abort)'. The server now responds with a 500 on reload instead of malformed JSON, which directly covers the real-world failure path guarded by this PR (non-abort network errors must reach the error event). An explicit expect(onError.mock.calls[0][0].error.status).toBe(500) assertion was added.
 
3. Malformed-JSON parser path removed from this PR: that scenario (valid 200 with unparseable body) is useful coverage but is unrelated to the abort/reload fix. I'm happy to add it in a follow-up PR if that's still wanted.

All 22 tests pass.

HarelM:
This PR had become stale.
I suggest closing it and splitting the PR to raster and vector tile changes (and other test improvements).
I think it will make it easier to review and also allow merging separately and reverting if there's a need.
Closing... Sorry for the long delay in response, version 6 needed my attention...

PR #7243, opened 2026-03-11
fix(double click zoom): makes it possible to disable double click zoom

qmatkovski:
I noticed one can't disable double-click zoom really - and that's for two reasons.

First, the click and tap handlers ignore their enabled state.

Second, the double-click shim handler reports its own enabled state as `undefined` by default (no defaults set on its dependencies), which is treated as `true` down the line.

This fixes both issues.

### DIFF
--- a/src/ui/handler/click_zoom.ts
+++ b/src/ui/handler/click_zoom.ts
@@ -25,6 +25,9 @@ export class ClickZoomHandler implements Handler {
     }
 
     dblclick(e: MouseEvent, point: Point) {
+        if (!this._enabled) {
+            return;
+        }
         e.preventDefault();
         return {
             cameraAnimation: (map: Map) => {
--- a/src/ui/handler/shim/dblclick_zoom.ts
+++ b/src/ui/handler/shim/dblclick_zoom.ts
@@ -50,7 +50,7 @@ export class DoubleClickZoomHandler {
      * @returns `true` if the "double click to zoom" interaction is enabled.
      */
     isEnabled() {
-        return this._clickZoom.isEnabled() && this._tapZoom.isEnabled();
+        return !!(this._clickZoom.isEnabled() && this._tapZoom.isEnabled());
     }
 
     /**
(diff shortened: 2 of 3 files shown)


src/ui/handler/shim/dblclick_zoom.ts
```diff
@@ -50,7 +50,7 @@ export class DoubleClickZoomHandler {
      * @returns `true` if the "double click to zoom" interaction is enabled.
      */
     isEnabled() {
-        return this._clickZoom.isEnabled() && this._tapZoom.isEnabled();
+        return !!(this._clickZoom.isEnabled() && this._tapZoom.isEnabled());
```
HarelM, line 53: Can you elaborate why this is needed? I would recommend not returning undefined from click and tap instead of monkey patch this call.

src/ui/handler/tap_zoom.ts
```diff
@@ -47,6 +47,10 @@ export class TapZoomHandler implements Handler {
     }
 
     touchend(e: TouchEvent, points: Array<Point>, mapTouches: Array<Touch>) {
+        if (!this._enabled) {
```
HarelM, line 50: Shouldn't this be checked in the touchstart instead of touchend?

src/ui/handler/click_zoom.ts
```diff
@@ -25,6 +25,9 @@ export class ClickZoomHandler implements Handler {
     }
 
     dblclick(e: MouseEvent, point: Point) {
+        if (!this._enabled) {
```
HarelM, line 28: I think the enable state is checked in the handler manager to avoid calling these methods, isn't it?

HarelM:
The PR has a template to make sure the relevant things are covered.
For example unit tests and changelog entry.
Please make sure to add those.

qmatkovski:
I did some more debugging and figured out I'm on the wrong track. The main issue is rather in the other library I'm using (@birkskyum/maplibre-gl-draw) which makes wrong assumptions on its end, and I think it's better to fix the issue there.

PR #7258, opened 2026-03-13
feat: restrict `sideEffects` to CSS files in package.json

CommanderStorm:
## Launch Checklist

We are missing the `sideEffects` flag in our package-json, which means that bundlers currently have to be extra cautios when optimising us.
The `*.css` is something that we can likely also remove, but this is a bit unclear to me if they are actually side effect free
This is uneesseary and we can instead just let them optimise us.

I am putting this into v6 since it is fairly hard to predict if this is actually safe.
According to our tests it is, and we are not using side-effect ridden things in the main apis that I looked at (plus in the many more minor APIs that I let claude audit).


HarelM:
I need to read about what this means, never saw it before. Thanks for taking the time to open this PR.

HarelM (approved):

HarelM:
It is worth a changelog entry? I'm not sure, your call.
Feel free to merge this either way.

PR #7267, opened 2026-03-16
Adds elevation to jumpTo method in case of terrain

### LINKED ISSUE #7178: Wrong popup location after `jumpTo()` when terrain is enabled
**maplibre-gl-js version**: 5.0.0, 5.19.0

**browser**: Chrome

### Steps to Trigger Behavior

 1. Open https://jsbin.com/wedisodada/1/edit?html,output
 2. Drag the map so the popup is near the bottom of the screen.
 3. Click the "Jump" button in the upper right corner. The popup is now in the wrong location.
 4. Drag the map slightly. The popup jumps to the correct location.

### Link to Demonstration

https://jsbin.com/wedisodada/1/edit?html,output

### Expected Behavior

Popup always stays in the right location.

### Actual Behavior

Popup location is wrong after `jumpTo()` when terrain is enabled.

HarelM:
## Launch Checklist

Fixes #7178

I've also refactored the popup class a bit to remove almost empty events callbacks.

The cause of this issue is that the pixel matrix is calculated with the current elevation and not with the target elevation.
So when placing the popup while listening to the move event, the matrix is somewhat incorrect.
When the moves end and the data from the elevation is updated the popup is not updated.
This does not solve completely the case of 3D terrain but reduces the probability of this bug.
It also follows the existing code to update elevation as part of the movement like fly-to and ease-to.
@chrneumann it is possible that this might solve some terrain collision issues in case the terrain is already loaded? In any case, your feedback here would be greatly valued.

 - [x] Link to related issues.
 - [x] Write tests for all new functionality - tests were updated as they already cover this code flow.


HarelM:
No, the event changes are just refactoring of the popup class as part of my attempts to fix this issue, but they do not change the logic at all, just simplified the code a bit and make it more type friendly.

HarelM:
The fix itself is only the if addition in the jumpTo method.

HarelM:
I've updated the initial thread message with this information.

PR #7279, opened 2026-03-18
Fix memory leak in VideoSource: remove playing listener and pause on removal

johanrd:
I saw some strange behavior in the memory profile when working with videos.

VideoSource registers a `playing` event listener on the video element that captures `this.map` in its closure, but never removes it. Since the browser media subsystem can keep a playing video element alive, the closure retains a reference to the entire Map object graph after `map.remove()` or `removeSource()`.

Fix: store the handler reference, remove it in a new `onRemove()` override, and pause the video.

## Launch Checklist
- [ ] Link to related issues.
- [x] ~~Post benchmark scores.~~ — N/A, no perf-sensitive path changed

Cowritten by claude

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
-- _...Add new stuff here..._
+- Fix memory leak in VideoSource: remove `playing` event listener and pause video on source removal ([#7279](https://github.com/maplibre/maplibre-gl-js/pull/7279)) (by [@johanrd](https://github.com/johanrd))
 
 ## 5.20.2
 


src/source/video_source.ts
```diff
@@ -56,6 +56,7 @@ export class VideoSource extends ImageSource {
     urls: Array<string>;
     video: HTMLVideoElement;
     roundZoom: boolean;
+    _onPlaying: (() => void) | null = null;
```
HarelM: Is it possible to solve this memory leak without using this method, or at least not override it in the tests?

johanrd: yes, good call. Switched to a private arrow class field now. Tests no longer touch the internal handler, but still verify the observable behavior — that removeEventListener('playing', ...) is called and the video is paused on removal.

PR #7280, opened 2026-03-18
Free StructArray typed views after GPU upload to reduce memory

johanrd:
After uploading static vertex/index buffers to the GPU, the existing code does `delete array.arrayBuffer` to free the JS heap copy. However, the typed array views (`uint8`, `int16`, `float32`, etc.) hold strong internal references to the underlying `ArrayBuffer`, so it is never garbage collected.

Add `StructArray.freeBufferAfterUpload()` which replaces the buffer with an empty one, recreates all views via `_refreshViews()`, then deletes the `arrayBuffer` property (preserving the existing falsy-check contract in `ProgramConfiguration` binder `upload()` methods).

## Benchmark scores
Measured locally using real PBF tile data (`mbsv5-6-18-23.vector.pbf`) with FillBucket and LineBucket using [this gist](https://gist.github.com/johanrd/9f4644618a8dc7d6f3793812130b845d)

| | Old (`delete arrayBuffer`) | New (`freeBufferAfterUpload`) |
|---|---|---|
| Per tile (fill + line only) | ~762 KB retained | 0 KB retained |
| 20 tiles, 100 StructArrays | 14.9 MB retained | 0 KB retained |
| Views still alive? | 100/100 | 0/100 |

This is only fill + line layers (2 of ~6 bucket types). Real maps with symbol, circle, and fill-extrusion layers would retain more.

## Note on `arrayBuffer` type declaration

`StructArray` declares `arrayBuffer: ArrayBuffer`, but after upload the property is deleted (making it `undefined`). This type lie is pre-existing — the old code already did `delete array.arrayBuffer` — and fixing it would require adding guards at 10+ call sites across `vertex_buffer.ts`, `index_buffer.ts`, `struct_array.ts`, and `program_configuration.ts`. Left as-is to keep this PR focused on the memory fix; could be addressed in a follow-up.

## Checklist
- [ ] Link to related issues.
- [x] Post benchmark scores. — included in PR body

Cowritten by claude

### DIFF
--- a/src/data/program_configuration.ts
+++ b/src/data/program_configuration.ts
@@ -244,7 +244,7 @@ class SourceExpressionBinder implements AttributeBinder {
     }
 
     upload(context: Context) {
-        if (this.paintVertexArray && this.paintVertexArray.arrayBuffer.byteLength) {
+        if (this.paintVertexArray?.arrayBuffer.byteLength) {
             if (this.paintVertexBuffer && this.paintVertexBuffer.buffer) {
                 this.paintVertexBuffer.updateData(this.paintVertexArray);
             } else {
@@ -320,7 +320,7 @@ class CompositeExpressionBinder implements AttributeBinder, UniformBinder {
     }
 
     upload(context: Context) {
-        if (this.paintVertexArray && this.paintVertexArray.arrayBuffer.byteLength) {
+        if (this.paintVertexArray?.arrayBuffer.byteLength) {
             if (this.paintVertexBuffer && this.paintVertexBuffer.buffer) {
                 this.paintVertexBuffer.updateData(this.paintVertexArray);
             } else {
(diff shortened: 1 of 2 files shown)


HarelM:
Thanks for taking the time to open this PR. Any chance to tone down the comments a bit?

src/util/struct_array.ts
```diff
@@ -191,6 +191,20 @@ abstract class StructArray {
...
         throw new Error('_refreshViews() must be implemented by each concrete StructArray layout');
     }
+
+    /**
+     * Release the JS heap copy of the buffer data after it has been uploaded
+     * to the GPU. Replaces the buffer with an empty one and recreates all
+     * typed array views so they no longer retain the original ArrayBuffer.
+     */
+    freeBufferAfterUpload() {
+        // Recreate views over an empty buffer so they release their reference
```
HarelM: I would recommend removing this comment or adding it to the method's tsdocs.

src/util/struct_array.ts
```diff
@@ -191,6 +191,20 @@ abstract class StructArray {
...
+        // to the original ArrayBuffer, then delete the property to preserve
+        // the falsy-check contract used by ProgramConfiguration binder upload().
+        this.arrayBuffer = new ArrayBuffer(0);
+        this._refreshViews();
+        delete this.arrayBuffer;
```
HarelM: Why not keep it as 0 sized array?

johanrd: Hi, thanks for the review, good call.

you mean Keep the 0-sized ArrayBuffer and change the guards to `.arrayBuffer?.byteLength > 0`? 

it would mean cleaner typing, but touches more files. i can do that.

HarelM: I don't know if it makes sense, but from a code perspective it looks weird to assign, refresh and then delete. 
If there's a need to check if it's empty or not calling another method to encapsulate it makes sense here. i.e. `isEmpty()`

johanrd: I think it makes sense, when i tried it: https://github.com/maplibre/maplibre-gl-js/pull/7280/changes/c1ef39ea75dc7b663ba9177832a7558843c8a909

src/data/program_configuration.ts
```diff
@@ -244,7 +244,7 @@ class SourceExpressionBinder implements AttributeBinder {
     }
 
     upload(context: Context) {
-        if (this.paintVertexArray && this.paintVertexArray.arrayBuffer) {
+        if (this.paintVertexArray && this.paintVertexArray.arrayBuffer.byteLength) {
```
HarelM: Can the `?.` operator be used here?

PR #7281, opened 2026-03-18
feat: add referrerPolicy option to RequestParameters

### LINKED ISSUE #7278: Cannot set referrer policy in transform request resulting in tiles not loading when using OpenStreetMap
OpenStreetMap is now enforcing their [usage policy](https://operations.osmfoundation.org/policies/nominatim/) of requiring `referer` to be sent in the header for tile requests. This means if your server or website has a restrictive `referrerPolicy` like `"same-origin"` then no `referer` is sent.

This should be fixable by setting a more relaxed `referrerPolicy` in the request options, however when providing one to `transformRequest`, it does not copy the property to the options parameter.

[OpenLayers](https://github.com/openlayers/openlayers/pull/17089) and [Leaflet](https://github.com/Leaflet/Leaflet/pull/9883) are both also dealing with this issue.

**maplibre-gl-js version**:

5.20.2

### Steps to Trigger Behavior

 1. Have `referrerPolicy` set up such that `referrer` is not sent by default, such as `"same-origin"` either in the server host or index file meta
 2. Use OpenStreetMap as the source
 3. Some tiles have errors as `referer` is not being sent in the header
 4. Specify `referrerPolicy` in `transformRequest` to allow `referer`, such as `"origin-when-cross-origin"`
 5. `referer` is still not sent in header

### Link to Demonstration

https://jsbin.com/juqadesize/edit?html,output

### Expected Behavior

Setting the `RequestParameters` to have a less restrictive `referrerPolicy` to allow `referer` to be sent

### Actual Behavior

`referrerPolicy` is not copied to the `fetch.Request` `RequestInit` parameters

### Proposed Fix

1. Add `referrerPolicy?: ReferrerPolicy` to `ajax.RequestParameters`
2. Add `referrerPolicy: requestParameters.referrerPolicy,` to the `RequestInit` object in `ajax.MakeFetchRequest`

Bingtagui404:
Closes #7278

## Summary

OpenStreetMap now enforces sending a `Referer` header for tile requests. Sites with restrictive `referrerPolicy` (e.g. `same-origin`) cannot override this via `transformRequest` because `referrerPolicy` was not forwarded to the fetch `Request`.

## Changes

- Add `referrerPolicy?: ReferrerPolicy` to the `RequestParameters` type
- Pass `referrerPolicy` through to the `Request` constructor in `makeFetchRequest`
- Add CHANGELOG entry

## Usage

```ts
const map = new Map({
  transformRequest: (url, resourceType) => {
    return {
      url,
      referrerPolicy: 'origin-when-cross-origin'
    };
  }
});
```

### DIFF
--- a/src/util/ajax.test.ts
+++ b/src/util/ajax.test.ts
@@ -278,4 +278,35 @@ describe('ajax', () => {
         });
 
     });
+
+    describe('referrerPolicy', () => {
+
+        test('should pass referrerPolicy to fetch Request', async () => {
+            global.fetch = originalFetch;
+
+            const fetchSpy = vi.spyOn(global, 'fetch').mockResolvedValue(new Response(new ArrayBuffer(1)));
+
+            await getArrayBuffer({url: 'http://example.com/test.json', referrerPolicy: 'origin-when-cross-origin'}, new AbortController());
+
+            expect(fetchSpy).toHaveBeenCalledTimes(1);
+            const request = fetchSpy.mock.calls[0][0] as Request;
+            expect(request.referrerPolicy).toBe('origin-when-cross-origin');
+
+            fetchSpy.mockRestore();
+        });
+
+        test('should default referrerPolicy to empty string when not provided', async () => {
+            global.fetch = originalFetch;
+
+            const fetchSpy = vi.spyOn(global, 'fetch').mockResolvedValue(new Response(new ArrayBuffer(1)));
+
+            await getArrayBuffer({url: 'http://example.com/test.json'}, new AbortController());
+
+            expect(fetchSpy).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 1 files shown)


HarelM:
Does it make sense to add a unit test or add this filed to an existing test? 
Looks good otherwise, thanks for taking the time to implement this.

Bingtagui404:
Added unit tests in b1d1e2f:

1. Verifies `referrerPolicy: 'origin-when-cross-origin'` is forwarded to the fetch `Request`
2. Verifies the default is `''` (empty string) when `referrerPolicy` is not provided — confirming no behavior change for existing callers

Both tests spy on `fetch` to inspect the actual `Request` object constructed by `makeFetchRequest`.

Issue #7069, opened 2026-02-06
`_defineProperty is not defined` after upgrading to maplibre-gl 5.17.0


gzurbach:
After upgrading maplibre-gl from 5.16.0 to 5.17.0, the application crashes at runtime with the following error when interacting with the map (e.g. zooming):

```
_defineProperty is not defined
```

This issue does not occur after downgrading to version 5.16.0.

The error appears to be caused by untranspiled code relying on Babel helpers that are not provided at runtime.

This looks like a packaging / transpilation regression in `5.17.0`, where Babel helpers such as `_defineProperty` are referenced but not included or inlined.

BRaimbault:
Hi!

I think I have a similar error here. I suspect that the changes in https://github.com/maplibre/maplibre-gl-js/pull/6983 introduced usage of modern class fields and private methods, which require ES2022+ and could be the cause.

HarelM:
Can you find the PR that caused this? I don't have a good way to reproduce this...
Is it the usage of "..." that's causing this?

HarelM:
Also probably similar to this one: 
- https://github.com/maplibre/maplibre-gl-js/issues/6680
- https://github.com/maplibre/maplibre-gl-js/issues/6429

wayofthefuture:
@BRaimbault Can you be more specific about #6983? I don't see anything in this PR which has not already existed in Maplibre prior.

neodescis:
@HarelM: Should we define what bundlers we want to support, and create some build+run tests using them? It might take some doing (and some maintenance), but I think it would prevent these problems in the future.

neodescis:
> [@BRaimbault](https://github.com/BRaimbault) Can you be more specific about [#6983](https://github.com/maplibre/maplibre-gl-js/pull/6983)? I don't see anything in this PR which has not already existed in Maplibre prior.

Nothing jumps out at me in this PR either, but keep in mind that it's only the worker thread that doesn't get these bundler-introduced functions, since the bundler doesn't know how to add them to the blob-ified script.

HarelM:
It might be enough to define some linting rules maybe... IDK... we need to add some safe guards...

HarelM:
It might be the "...", but it was not introduced in this PR, I think.

BRaimbault:
[...]

PR #7283, opened 2026-03-18
test: add build test for ES2020 compatibility

claudiobgit:
## Summary

- Adds a build test that catches ES2022+ class field declarations in the UMD bundle
- The test transforms dist/maplibre-gl.js with esbuild (already available as a transitive dependency of Vite) targeting es2020 and verifies no downleveling helpers are injected

## Context

When bundlers like esbuild (used by Vite) target < ES2022, they transform class field declarations into helper calls. These helpers are defined at ESM module scope but the worker code is serialized via .toString() into a Web Worker blob — the worker does not have access to them, causing a runtime crash.

This test is intentionally **failing** with the current packages that still emit ES2022+ class fields. Once the updated packages (with ES2020 targets) are bumped in this repo, the test will pass.

- Fixes: #7069

## Test plan

- [ ] Test fails with current packages (confirms it catches the regression)
- [ ] After updating @maplibre/mlt and @maplibre/geojson-vt to new versions, rebuild and test passes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -75,6 +75,7 @@
         "devtools-protocol": "^0.0.1596832",
         "diff": "^8.0.3",
         "dts-bundle-generator": "^9.5.1",
+        "esbuild": "^0.27.4",
         "eslint": "^9.39.2",
         "eslint-plugin-html": "^8.1.4",
         "eslint-plugin-import": "^2.32.0",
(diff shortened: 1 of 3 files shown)


HarelM:
I think esbuild needs to be added to dev dependencies for this to run the tests...?

claudiobgit:
You are right we need it. Sorry. Is this something we can do or we need a workaround?

HarelM:
Just add it as devDependencies.

HarelM:
Lock file update is missing I believe.

claudiobgit:
True. I pushed it now. Thanks

HarelM:
Looks like this is passing even though geojson-vt and mlt packages were not updated here, so I'm guessing the test it not checking what it should...?

claudiobgit:
I will investigate later today what is happening as I was catching class
fields locally.

Claudio Bernardini

Il gio 19 mar 2026, 13:31 Harel M ***@***.***> ha scritto:

claudiobgit:
@HarelM I investigated why the test passes. It turns out `main` already has `@maplibre/mlt@1.1.8` and `@maplibre/geojson-vt@6.0.4` in the lockfile (the fixed versions). So the build produces a dist without ES2022+ class fields, and the test passes correctly.

I verified locally that the test logic works: with the old packages (1.1.7/6.0.3), esbuild transform produces `__publicField` helpers → test fails. With the new packages (1.1.8/6.0.4) → test passes.

So the test is actually working as intended — it will catch any future regression (e.g., if a dependency reintroduces ES2022+ class fields).

I will clean up the PR to a single commit (test file + esbuild devDependency) rebased on current main.

claudiobgit:
I tested the fix locally by running npm pack on my maplibre-gl-js clone (which has the updated @maplibre/mlt@1.1.8 and @maplibre/geojson-vt@6.0.4 in its lockfile), then installed the resulting tarball in my Strapi plugin project that uses Vite with target: 'es2020'. The __publicField is not defined **error is gone** — the map loads and works correctly. This confirms that once maplibre-gl publishes a new npm release with the updated dependencies, the issue will be fully resolved for end users and the test will block any > 'es2020' in CI. 
Are we all good?

claudiobgit:
Just pulled last 5.21.0 npm package now in my application and everything is fine.

I am a happy mapper again! Thanks for the support. 

immagine

PR #7284, opened 2026-03-18
Fix RasterDEMTileSource tiles stuck in reloading state

### LINKED ISSUE #7277: Tiles from RasterDEMTileSource are stuck in state "reloading" and block map.loaded() or idle event
**maplibre-gl-js version**: 5.16, 5.20.2 (latest), `main`

**browser**: Firefox 147.0.2

### Steps to Trigger Behavior

 1. Add layer with raster-dem source and wait for map being idle
 2. Switch projection
 3. `map.loaded()` is stuck returning `false`; no `idle` event is dispatched

This hinders our ability to do something when map is done loading after projection change.

### Link to Demonstration

1) https://jsbin.com/difobekiji/1/edit?js,console,output

Adds a raster-dem source and hillshade layer, provides button to switch projection, shows current state of `map.loaded()`, logs `idle` event to console. Bug is triggered just by clicking on _"Switch to Globe"_ - map loaded never changes to `true` and no `idle` event is logged to console. This continues indefinitely until zoom changes (to load new tiles), however then another projection switch triggers the bug again.

2) https://maplibre.org/maplibre-gl-js/docs/examples/add-a-hillshade-layer.html

Open browser console. Run `map.setProjection({type: 'globe'});` to switch projection. Run `map.loaded()` - it returns `false`. This continues indefinitely until zoom changes (to load new tiles), however then another projection switch triggers the bug again.

### Expected Behavior

`map.loaded()` eventually returns `true` and `idle` event is dispatched

### Actual Behavior

`map.loaded()` never returns `true` and `idle` event is never dispatched

---

It seems to me that [`RasterDEMTileSource`](https://github.com/maplibre/maplibre-gl-js/blob/main/src/source/raster_dem_tile_source.ts) class never calls `tile.state = 'loaded';` (nor does anything with `rawImageData`) for tiles that do not satisfy the condition on line 87 (the stuck tiles do have `tile.actor` and are not expired). I'm not exactly sure what should happen with these tiles, hence opening an issue instead of a PR.

katemihalikova:
Fixes #7277. As per your comment in the issue, I dived into the code more. Comparing lifecycle of tiles from other sources, I believe this PR fixes behavior of `RasterDEMTileSource` and puts it somehow in line with logic in `RasterTileSource` and `VectorTileSource`. DEM tiles are no longer stuck in `"reloading"` state, `map.loaded()` eventually returns `true` and `idle` event gets dispatched correctly. A unit test that is failing with `main` but passing with this PR is included.

[Benchmark scores](https://github.com/user-attachments/assets/bd53b36a-7a63-487f-b341-b10de041d32c)

## Launch Checklist

 - [x] Link to related issues.
 - ~Include before/after visuals or gifs if this PR includes visual changes.~
 - ~Document any changes to public APIs.~
 - [x] Post benchmark scores.

### DIFF
--- a/src/source/raster_dem_tile_source.ts
+++ b/src/source/raster_dem_tile_source.ts
@@ -84,10 +84,11 @@ export class RasterDEMTileSource extends RasterTileSource implements Source {
                     baseShift: this.baseShift
                 };
 
+                if (tile.actor && tile.state !== 'expired' && tile.state !== 'reloading') {
+                    return;
+                }
                 if (!tile.actor || tile.state === 'expired') {
                     tile.actor = this.dispatcher.getActor();
-                } else if (tile.state !== 'reloading') {
-                    return;
                 }
                 const data = await tile.actor.sendAsync({type: MessageType.loadDEMTile, data: params});
                 tile.dem = data;


src/source/raster_dem_tile_source.ts
```diff
@@ -86,12 +86,12 @@ export class RasterDEMTileSource extends RasterTileSource implements Source {
 
                 if (!tile.actor || tile.state === 'expired') {
                     tile.actor = this.dispatcher.getActor();
-                    const data = await tile.actor.sendAsync({type: MessageType.loadDEMTile, data: params});
-                    tile.dem = data;
-                    tile.needsHillshadePrepare = true;
-                    tile.needsTerrainPrepare = true;
-                    tile.state = 'loaded';
                 }
+                const data = await tile.actor.sendAsync({type: MessageType.loadDEMTile, data: params});
```
HarelM, line 93: Moving this outside the if will result in loading it multiple time in case this is in loading state, wouldn't it?
Should the if be updated to be "more correct" instead?
I would even consider negating the "if" and retuning instead of keeping this code inside the if for better readability.

HarelM:
Added a minor comment, thanks for taking the time to open this PR!

src/source/raster_dem_tile_source.ts
katemihalikova: Good point, I haven't considered a tile that reloads while still loading. This means there should be a branch that does nothing, as you pointed out. As I don't want to replace actor on reloading (it's the same in other sources), I need three branches (one that adds actor and loads tile, one that reuses actor and loads time, one that does nothing). I've pushed a variant where the conditions are shortest, but I'm not 100% sure about readability, would something like this be better?
```ts
if (tile.actor && tile.state !== 'expired' && tile.state !== 'reloading') {
    return;
} else if (!tile.actor || tile.state === 'expired') {
}
const data = ...
```
or
```ts
if (tile.actor && tile.state !== 'expired' && tile.state !== 'reloading') {
    return;
} else if (tile.state !== 'reloading') {
}
const data = ...
```

HarelM: I think that if (...) return is preferred.
Then don't use else if, but instead just if (!actor) - assign actor.
I think it help read this code better.

katemihalikova: Thanks! I've changed it according to your comment. I've left the second `if` condition in full (it's the same in other sources).

PR #7285, opened 2026-03-18
Ensure GPU finishes draw calls before benchmark timing

xavierjs:
Hi,

 ## Issue

 Benchmarks involving rendering return wrong results. The end time is measured once `map._render()` method is called but we do not ensure that the GPU call stack is empty. We need to make sure that the GPU has processed all the draw calls
 before measuring this timestamp.
 If we start chrome without hardware acceleration (with CLI parameters `--use-gl=swiftshader --disable-web-security`), we can notice that the browser remains unresponsive and laggy for a very long time, whereas the benchmark result is almost
 as good as if we get hardware acceleration.

 ## Proposed solution

 To make sure that the GPU has processed all its draw calls, we read a pixel at the end of the benchmark.
 Of course, the rendering benchmark results became significantly worse.

 ## Benchmarks

 If we consider this benchmark for example: http://localhost:9966/test/bench/versions/index.html?compare=#LayerSymbol (we render a map with symbol layers).

 **Before:**
 Google Chrome, Macbook Pro M4 Max: `0.413ms`
 Google Chrome, Macbook Pro M4 Max with swiftshaders (hardware acceleration disabled): `0.440ms`

 **After:**
 Google Chrome, Macbook Pro M4 Max: `2.57ms`
 Google Chrome, Macbook Pro M4 Max with swiftshaders (hardware acceleration disabled): `171ms`

 ## Misc

 No related issues.
 This stackoverflow post also advises using `gl.readPixels` for WebGL benchmarking: https://stackoverflow.com/questions/38710140/how-do-i-benchmark-a-webgl-shader

 This PR does not include any visual changes and does not require additional tests. The public API remains the same (we are not supposed to use the added methods in the public API so I prefixed them with `_`).

## Launch Checklist
 - [X] Link to related issues.
 - [X] Post benchmark scores.


HarelM:
Is it possible to change the benchmarks without changing the "production" code? I would like to move away as much as possible from this "benckmarks knows the internal of the code" approach.

xavierjs:
> Is it possible to change the benchmarks without changing the "production" code? I would like to move away as much as possible from this "neckmarks knows the internal of the code" approach.

Sure, it is done in the last commit.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## main
 ### ✨ Features and improvements
 - Add `referrerPolicy` option to `RequestParameters` to allow controlling the referrer policy for tile requests ([#7278](https://github.com/maplibre/maplibre-gl-js/issues/7278))
+- Wait for the GPU to finish its callstack for rendering benchmarks
```
HarelM: Please add pr number and author.

HarelM:
Thanks! Added a minor comment, looks good otherwise.

xavierjs:
> Thanks! Added a minor comment, looks good otherwise.

Thank you for your fast feedback, I have just added the link to the PR in the Changelog file.

PR #7286, opened 2026-03-18
Fix GeolocateControl movestart listener leak on removal

johanrd:
When `trackUserLocation` is enabled, `GeolocateControl._finishSetupUI` registers a `movestart` handler on the map as an anonymous function. Because `onRemove()` has no reference to that function, it can never call `this._map.off('movestart', ...)` to remove it.

This causes two problems:
1. **Memory leak**: the closure retains the `GeolocateControl` instance (and everything it references) for the lifetime of the map, even after `removeControl()`.
2. **Crash**: if the control was in `ACTIVE_LOCK` state when removed, the leaked handler calls `this._map.isZooming()` on the next `movestart` — but `this._map` was set to `undefined` in `onRemove()`, causing a `TypeError`.

Fix: extract the anonymous handler into a named class field (`_onMoveStart`) so it can be removed in `onRemove()`. Also add a defensive `if (!this._map) return` guard.

## Test plan

|---|---|---|
| Listener leak (count stays on map) | `expected 1 to be 0` | passes |
| Crash (`isZooming` on undefined) | `TypeError: Cannot read properties of undefined (reading 'isZooming')` | passes |

## Checklist
- [ ] Link to related issues.
- [x] ~~Post benchmark scores.~~ — N/A, no perf-sensitive path changed

Cowritten by claude


src/ui/control/geolocate_control.test.ts
```diff
@@ -880,6 +880,32 @@ describe('GeolocateControl with no options', () => {
         expect(geolocate._circleElement.style.width).toBeTruthy();
     });
 
+    test('GeolocateControl removes movestart listener from map on removal', async () => {
+        const geolocate = new GeolocateControl({trackUserLocation: true});
+        map.addControl(geolocate);
+        await sleep(0);
+
+        const movestartBefore = (map._listeners?.movestart ?? []).length;
```
HarelM: I'm a bit tired of repeating myself when reviwing the tests, please make sure to review this on your own before opening a PR or mark the PR as draft while you review your changes.

johanrd: @HarelM Hi, sorry about that, and thanks for the detailed, quick and honest reviews. 

I've removed the listener-count test (a regression there would also cause the crash test to fail I think). Also rewrote the type error test to use public API.

HarelM (approved):

PR #7288, opened 2026-03-19
Fix Terrain GPU resource leak when disabling terrain

johanrd:
*Be ware: This is not from a use case I have had myself, but asked claude for memory vulnerabilities that could be inspired from the other recent memory leak prs*:

`setTerrain(null)` drops the `Terrain` object without destroying its GPU resources: FBO, depth/coords textures, placeholder textures, and cached meshes (~25 MB at 1080p). These leak every toggle because WebGL resources require explicit `gl.deleteTexture()` / `gl.deleteFramebuffer()`.

Fix: add `Terrain.destroy()` and call it from `setTerrain(null)`.

Note: `map.remove()` is not affected — `loseContext()` frees everything at the driver level. This only leaks when toggling terrain while the map stays alive.

### Implementation notes

- `tileManager.destruct()` is called inside `Terrain.destroy()` so all cleanup is encapsulated in one place.
- **FBO double-delete is harmless:** `Framebuffer.destroy()` deletes its current color attachment, then `_fboCoordsTexture.destroy()` / `_fboDepthTexture.destroy()` delete the same GL handle again for whichever was attached. This is a no-op per the WebGL spec and matches the existing resize teardown pattern in `getFramebuffer()`.

## Test plan

|---|---|---|
| `destroy()` releases GPU resources | `TypeError: terrain.destroy is not a function` | Resources nulled, mesh cache empty |

## Checklist
- [ ] Link to related issues.

Cowritten by claude

### DIFF
--- a/src/render/terrain.test.ts
+++ b/src/render/terrain.test.ts
@@ -383,7 +383,7 @@ describe('Terrain', () => {
         });
     });
 
-    test('destroy releases GPU resources', () => {
+    test('destroy does not throw', () => {
         const context = new Context(gl);
         const painter = {
             context,
@@ -399,17 +399,7 @@ describe('Terrain', () => {
         terrain.getFramebuffer('coords');
         terrain.getTerrainMesh(new OverscaledTileID(0, 0, 0, 0, 0));
 
-        expect(terrain._coordsTexture).toBeTruthy();
-        expect(terrain._fbo).toBeTruthy();
-        expect(Object.keys(terrain._meshCache).length).toBeGreaterThan(0);
-
-        terrain.destroy();
-
-        expect(terrain._coordsTexture).toBeNull();
-        expect(terrain._fbo).toBeNull();
-        expect(terrain._fboCoordsTexture).toBeNull();
-        expect(terrain._fboDepthTexture).toBeNull();
-        expect(Object.keys(terrain._meshCache)).toHaveLength(0);
+        expect(() => terrain.destroy()).not.toThrow();
     });
 
 });


src/ui/map.ts
```diff
@@ -2312,7 +2312,10 @@ export class Map extends Camera {
 
         if (!options) {
             // remove terrain
-            if (this.terrain) this.terrain.tileManager.destruct();
+            if (this.terrain) {
+                this.terrain.destroy();
+                this.terrain.tileManager.destruct();
```
HarelM: Move that insidethe terrain destroy method.

src/render/terrain.test.ts
```diff
@@ -383,4 +384,64 @@ describe('Terrain', () => {
         });
     });
 
+    test('destroy should free GPU resources', () => {
```
HarelM: Too specific, let's just make sur destroy doesn't throw an exception when called.

src/render/terrain.test.ts
```diff
@@ -383,4 +383,33 @@ describe('Terrain', () => {
...
+            style: {projection: null},
+        } as any as Painter;
+        const tileManager = {_source: {tileSize: 512}, destruct: vi.fn()} as any as TileManager;
+        const terrain = new Terrain(painter, tileManager, {} as any as TerrainSpecification);
+
+        terrain.getCoordsTexture();
+        terrain.getFramebuffer('coords');
+        terrain.getTerrainMesh(new OverscaledTileID(0, 0, 0, 0, 0));
+
+        expect(terrain._coordsTexture).toBeTruthy();
```
HarelM: This is still using "private" fields...

PR #7289, opened 2026-03-19
Memory: Cap tile texture reuse pool to prevent unbounded VRAM growth

johanrd:
`Painter.saveTileTexture()` pushes every unloaded tile's GPU texture into a reuse pool (`_tileTextures`) with no size limit. During quick zoom out or flyto textures accumulate faster than they are reused, causing unbounded VRAM growth. Can cause WebGL process (context lost).

Fix: add a static cap (`MAX_TEXTURE_POOL_SIZE_PER_BUCKET = 50`) and call `texture.destroy()` when the pool is full. 50 textures per size bucket (~50 MB for 512x512 RGBA tiles) should be above typical reuse needs?

## Test plan

|---|---|---|
| Pool exceeds cap after 150 saves | 150 textures pooled | 50 pooled, 100 destroyed |

## Checklist
- [ ] Link to related issues.

Cowrittten by claude

### DIFF
--- a/src/render/painter.test.ts
+++ b/src/render/painter.test.ts
@@ -40,16 +40,23 @@ describe('tile texture pool', () => {
         return new Texture(painter.context, image, gl.RGBA);
     }
 
-    test('saveTileTexture caps pool size', () => {
+    test('saveTileTexture caps pool size and destroys excess', () => {
         const painter = createPainterWithPool();
+        const cap = Painter.MAX_TEXTURE_POOL_SIZE_PER_BUCKET;
 
-        for (let i = 0; i < 150; i++) {
-            painter.saveTileTexture(createTexture(painter, 256));
+        const textures: Texture[] = [];
+        for (let i = 0; i < cap + 100; i++) {
+            const tex = createTexture(painter, 256);
+            textures.push(tex);
+            painter.saveTileTexture(tex);
         }
 
         let reused = 0;
         while (painter.getTileTexture(256)) reused++;
-        expect(reused).toBe(Painter.MAX_TEXTURE_POOL_SIZE_PER_BUCKET);
+        expect(reused).toBe(cap);
+
+        const destroyed = textures.filter(t => t.texture === null).length;
+        expect(destroyed).toBe(100);
 
         painter.destroy();
     });


src/render/painter.test.ts
```diff
@@ -25,3 +26,39 @@ test('Render must not fail with incompletely loaded style', () => {
...
+    }
+
+    test('saveTileTexture caps pool size and destroys excess textures', () => {
+        const painter = createPainterWithPool();
+        const cap = Painter.MAX_TEXTURE_POOL_SIZE_PER_BUCKET;
+
+        for (let i = 0; i < cap; i++) {
+            painter.saveTileTexture(createTexture(painter, 256));
+        }
+        expect(painter._tileTextures[256].length).toBe(cap);
```
HarelM: That's a private field, I would avoid using it. Cache limit are hard to test so it might not be worth the trouble...

src/render/painter.ts
```diff
@@ -685,12 +685,16 @@ export class Painter {
...
+
     saveTileTexture(texture: Texture) {
         const textures = this._tileTextures[texture.size[0]];
         if (!textures) {
             this._tileTextures[texture.size[0]] = [texture];
-        } else {
+        } else if (textures.length < Painter.MAX_TEXTURE_POOL_SIZE_PER_BUCKET) {
             textures.push(texture);
+        } else {
+            texture.destroy();
```
HarelM, line 697: Is this using lru mechanism?

johanrd: Currently push/pop (LIFO), not LRU — when the pool is full the incoming texture is destroyed. Since all textures of the same size are interchangeable (they're overwritten via texture.update() on reuse), it should no matter which one is kept. Could also switch to LRU/evict-oldest, but as I understand it it is not necessary in this case.

src/render/painter.test.ts
johanrd: thanks, good point. found a public way to test it now, i think, but `texture === null` is maybe still a bit specific.

src/render/painter.ts
HarelM: Fair enough.

src/render/painter.test.ts
HarelM: I think it's good enough.

HarelM:
Please resolve the conflict and I'll merge this.

PR #7294, opened 2026-03-19
Cleanup: Remove legacy browser compatibility code (Baseline 2016)

johanrd:
Remove compatibility shims and fallbacks for browsers that MapLibre no longer needs to support. This PR targets baseline 2016. MapLibre's ES2019 tsconfig target implies a browser floor of Chrome 73 / Firefox 64 / Safari 12.1 (all 2018-2019, ([context](https://github.com/maplibre/maplibre-gl-js/discussions/3450))) — all cleanups targets APIs standardized well before that .

**Bundle size: -674 bytes** (1,049,231 → 1,048,557, measured locally; `bundle_size.json` not updated yet) 

### Changes included (considered safe to cleanup)

| File | What | Baseline since | Ref |
|---|---|---|---|
| `context.ts` | Remove `MOZ_`/`WEBKIT_` prefixed `EXT_texture_filter_anisotropic` lookups | Chrome 34, FF 47, Safari 9.1 (2016) | [caniuse](https://caniuse.com/mdn-api_ext_texture_filter_anisotropic) |
| `dom.ts` | Remove `testProp()`, `MozUserSelect`, `msUserSelect`; hardcode `transform`. | transform: Safari 9 (2015) | [caniuse transform](https://caniuse.com/transforms2d), [user-select](https://caniuse.com/mdn-css_properties_user-select) |
| `dom.ts` | Remove passive event listener `'passive' in options` compat check | Chrome 51, FF 49, Safari 10 (2016) | [caniuse](https://caniuse.com/passive-event-listener) |
| `marker.ts` | Remove `charCode`/`keyCode` fallback, use `e.code` only | Chrome 48, FF 38, Safari 10.1 (March 2017) | [caniuse](https://caniuse.com/keyboardevent-code) |
| `throttled_invoker.ts` | Remove `MessageChannel` existence check and `setTimeout` fallback | Baseline Sept 2015 | [MDN](https://developer.mozilla.org/en-US/docs/Web/API/MessageChannel#browser_compatibility) |
| `time_control.ts` | Remove `performance.now()` existence check and `Date.now` fallback | Baseline 2012 | [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Performance/now#browser_compatibility) |
| `transferable_grid_index.ts` | Replace `Array.prototype.slice.call()` IE/Safari workaround with `Array.from()` | Chrome 45, FF 38, Safari 10 (2016) | [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/slice#browser_compatibility) |
| `dom.ts`, `handler_manager.ts`, `navigation_control.ts` | Remove now-trivial `DOM.addEventListener`/`removeEventListener` wrappers; inline direct calls | Follow-up to passive event listener removal above | |
| `dom.ts`, `marker.ts`, `popup.ts`, `box_zoom.ts` | Remove `DOM.setTransform()` wrapper and `transformProp` indirection; inline `el.style.transform = value` | transform: Safari 9 (2015) | [caniuse](https://caniuse.com/transforms2d) |  

### Investigated but not removed (still needed)

| What | Why | Ref |
|---|---|---|
| `-webkit-user-select` (`dom.ts`) | Safari still requires the webkit prefix | [caniuse](https://caniuse.com/mdn-css_properties_user-select) |
| Fullscreen API prefixes (`fullscreen_control.ts`) | Safari unprefixed only in 16.4 (March 2023) | [caniuse](https://caniuse.com/mdn-api_document_fullscreenchange_event) |
| `isSafari()` worker count (`worker_pool.ts`) | Perf optimization from A/B testing, not a workaround | |
| OffscreenCanvas / ImageBitmap / VideoFrame detection | Legitimate runtime feature detection | |

## Checklist
- [ ] Link to related issues.
- [x] ~~Include before/after visuals or gifs if this PR includes visual changes.~~ — N/A
- [x] ~~Write tests for all new functionality.~~ — N/A, pure removal
- [x] ~~Document any changes to public APIs.~~ — N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Add `referrerPolicy` option to `RequestParameters` to allow controlling the referrer policy for tile requests ([#7278](https://github.com/maplibre/maplibre-gl-js/issues/7278))
 - Remove Edge 18 WebP detection workaround; always send `Accept: image/webp` header for image requests ([#7293](https://github.com/maplibre/maplibre-gl-js/pull/7293)) (by [@johanrd](https://github.com/johanrd))
 - Remove legacy browser compatibility code targeting IE11 and pre-2016 browsers ([#7294](https://github.com/maplibre/maplibre-gl-js/pull/7294)) (by [@johanrd](https://github.com/johanrd))
+- Remove legacy `DOM.remove()` and `DOM.mouseButton()` wrappers; use native APIs directly (baseline 2015) ([#7295](https://github.com/maplibre/maplibre-gl-js/pull/7295)) (by [@johanrd](https://github.com/johanrd))
 
 ### 🐞 Bug fixes
 
--- a/src/ui/marker.ts
+++ b/src/ui/marker.ts
@@ -374,7 +374,7 @@ export class Marker extends Evented {
             delete this._map;
         }
         this._element.removeEventListener('click', this._onClick);
-        DOM.remove(this._element);
+        this._element.remove();
         if (this._popup) this._popup.remove();
         return this;
     }
--- a/src/ui/control/attribution_control.ts
+++ b/src/ui/control/attribution_control.ts
@@ -83,7 +83,7 @@ export class AttributionControl implements IControl {
 
     /** {@inheritDoc IControl.onRemove} */
     onRemove() {
-        DOM.remove(this._container);
+        this._container.remove();
 
         this._map.off('styledata', this._updateData);
         this._map.off('sourcedata', this._updateData);
--- a/src/ui/control/control.ts
+++ b/src/ui/control/control.ts
@@ -28,7 +28,7 @@ export type ControlPosition = 'top-left' | 'top-right' | 'bottom-left' | 'bottom
  *     }
  *
  *     onRemove() {
- *         this._container.parentNode.removeChild(this._container);
+ *         this._container.remove();
  *         this._map = undefined;
  *     }
  * }
--- a/src/ui/control/fullscreen_control.ts
+++ b/src/ui/control/fullscreen_control.ts
@@ -91,7 +91,7 @@ export class FullscreenControl extends Evented implements IControl {
 
     /** {@inheritDoc IControl.onRemove} */
     onRemove() {
-        DOM.remove(this._controlContainer);
+        this._controlContainer.remove();
         this._map = null;
         window.document.removeEventListener(this._fullscreenchange, this._onFullscreenChange);
     }
--- a/src/ui/control/geolocate_control.ts
+++ b/src/ui/control/geolocate_control.ts
@@ -302,7 +302,7 @@ export class GeolocateControl extends Evented implements IControl {
             this._accuracyCircleMarker.remove();
         }
 
-        DOM.remove(this._container);
+        this._container.remove();
         this._map.off('movestart', this._onMoveStart);
         this._map.off('zoom', this._onUpdate);
         this._map.off('move', this._onUpdate);
--- a/src/ui/control/globe_control.ts
+++ b/src/ui/control/globe_control.ts
@@ -38,7 +38,7 @@ export class GlobeControl implements IControl {
 
     /** {@inheritDoc IControl.onRemove} */
     onRemove() {
-        DOM.remove(this._container);
+        this._container.remove();
         this._map.off('styledata', this._updateGlobeIcon);
         this._map.off('projectiontransition', this._updateGlobeIcon);
         this._globeButton.removeEventListener('click', this._toggleProjection);
--- a/src/ui/control/logo_control.ts
+++ b/src/ui/control/logo_control.ts
@@ -63,7 +63,7 @@ export class LogoControl implements IControl {
 
     /** {@inheritDoc IControl.onRemove} */
     onRemove() {
-        DOM.remove(this._container);
+        this._container.remove();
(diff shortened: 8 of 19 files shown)


src/ui/marker.ts
```diff
@@ -501,13 +501,7 @@ export class Marker extends Evented {
...
 
     _onKeyPress = (e: KeyboardEvent) => {
-        const code = e.code;
-        const legacyCode = e.charCode || e.keyCode;
-
-        if (
-            (code === 'Space') || (code === 'Enter') ||
-            (legacyCode === 32) || (legacyCode === 13) // space or enter
-        ) {
+        if (e.code === 'Space' || e.code === 'Enter') {
```
HarelM, line 505: Are there constant that can be used instead of these string by any chance?

src/util/throttled_invoker.ts
```diff
@@ -26,11 +24,6 @@ export class ThrottledInvoker {
         this._triggered = true;
         if (this._channel) {
             this._channel.port1.postMessage(true);
```
HarelM: Use `this._channel?.port1.postMessage?` Although I'm not sure I see how channel can be undefined...?

src/util/time_control.ts
```diff
@@ -2,9 +2,7 @@
  * Manages time flow with optional freezing capability for deterministic rendering.
  */
 class TimeManager {
-    private _realTime = typeof performance !== 'undefined' && performance && performance.now ?
-        performance.now.bind(performance) :
-        Date.now.bind(Date);
+    private _realTime = performance.now.bind(performance);
```
HarelM: Is it possible to remove the `bind` here as well? I want to avoid using those as much as possible.

src/util/transferable_grid_index.ts
```diff
@@ -101,10 +101,7 @@ export class TransferableGridIndex {
         const min = this.min;
         const max = this.max;
         if (x1 <= min && y1 <= min && max <= x2 && max <= y2 && !intersectionTest) {
-            // We use `Array.slice` because `this.keys` may be a `Int32Array` and
-            // some browsers (Safari and IE) do not support `TypedArray.slice`
-            // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/slice#Browser_compatibility
-            return Array.prototype.slice.call(this.keys);
+            return Array.from(this.keys);
```
HarelM: The previous comment was talking about `slice`, should the code here use slice instead of `Array.from`?

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## main
 ### ✨ Features and improvements
 - Add `referrerPolicy` option to `RequestParameters` to allow controlling the referrer policy for tile requests ([#7278](https://github.com/maplibre/maplibre-gl-js/issues/7278))
+- Remove legacy browser compatibility code targeting IE11 and pre-2016 browsers (by [@johanrd](https://github.com/johanrd))
```
HarelM: Please add a PR link as well, thanks!!

HarelM:
Added some minor comments, but overall this looks great, thanks!!

src/util/time_control.ts
johanrd: changed to arrow function

src/util/transferable_grid_index.ts
johanrd: Changed to [...this.keys] (spread). Plain .slice() doesn't work here because this.keys can be an Int32Array at runtime

src/util/throttled_invoker.ts
johanrd: Replaced the if (this._channel) guard with this._channel?.port1.postMessage(true) and typed _channel as MessageChannel | undefined (since remove() deletes it)

src/ui/marker.ts
johanrd: 'Space' and 'Enter' are the standard [KeyboardEvent.code](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code) values. Note that e.code is marked as [limited availabilty on mobile](https://caniuse.com/?search=KeyboardEvent%3A+code) due to mobile virtual keyboards returning `''`, but that should not be relevant here — this handler is for keyboard/screen-reader accessibility, not virtual keyboard input.

The alternative would be e.key === ' ' || e.key  === 'Enter' (or keep the legacy) but that's less readable.

johanrd:
@HarelM thanks for the feedback, the removal of `{passive: false}` enabled some further cleanup of the DOM helpers, so you would need to have a look again I'm afraid, 

the only small uncertainty is the KeyboardEvent.code https://github.com/maplibre/maplibre-gl-js/pull/7294#discussion_r2959965901 as i see it, but i think it should be ok.

src/ui/marker.ts
HarelM: Fair enough.

src/util/time_control.ts
```diff
@@ -2,9 +2,7 @@
  * Manages time flow with optional freezing capability for deterministic rendering.
  */
 class TimeManager {
-    private _realTime = typeof performance !== 'undefined' && performance && performance.now ?
-        performance.now.bind(performance) :
-        Date.now.bind(Date);
+    private _realTime = () => performance.now();
```
HarelM: Given that this is a call invocation of a single line, wouldn't it be better to completly remove it?

PR #7296, opened 2026-03-20
Fix missing encoding value for cached vector tiles

ted-piotrowski:
## Launch Checklist

I found this bug while experimenting with the new 'mlt' file format. In some instances where I tilted the map to 3D and back to 2D and then panned the map, the encoding property on some tiles would come back as undefined during this check:

https://github.com/maplibre/maplibre-gl-js/blob/779cd294739f44b2a200ac0b3a235af763931816/src/data/feature_index.ts#L117

This resulted in Maplibre attempting to decode the 'mlt' tile as an 'mvt' and returning "Unimplemented Type 6 error". I tested this fix with my specific setup and found it to resolve the issue.

I might need a bit of direction to get this one across the line in terms of tests (if required) and benchmarks but just wanted to open the PR as it is a pretty minor bug fix. Please let me know what next steps I should take to get it across the line.

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
The code is straight forward and it might be merged as is as I probably forgot to add this line as part of the mlt implementation I did, but it would be great if there's a unit test that can reproduce the issue somehow.
If you understand how this scenario is can happen maybe you could create a test to simulate it?
Do you think it is related to move a tile out af view and back into view so it changes from in view tiles to out of view cache (just a wild guess)?

HarelM:
I just remembered that there was already a PR with this fix:
- #7056

I'll merge this other one since this is an important fix, but it would be great if you could find the time to write a test that will reproduce the issue.

HarelM:
Closing this one as the other PR was merged, tests would be very welcomed though.

PR #7303, opened 2026-03-20
Fix unhandled framebuffer error when tab wakes from sleep

johanrd:
I saw this error in Sentry from a production app:
- `14:21:04` — user navigates to a page with map, map loads with satellite tiles + heatmap
- `14:46:38` — last user interaction
- `07:07:49` — next morning, tile fetches resume (tab wakes up after ~16 hours)
- `07:38:34` — `Error: Framebuffer is not complete` fires, unhandled

## Root cause?
The `Framebuffer` constructor throws whenever `gl.checkFramebufferStatus()` isn't `FRAMEBUFFER_COMPLETE`. On a *transient* GPU resource loss that doesn't fire a `webglcontextlost` event, this turns a self-healing blip into a hard, unhandled error.

The existing context-loss recovery can't catch this: for the error to reach the global handler, `webglcontextlost` can't have fired (otherwise `this.style` would be null and `_render()` couldn't run). So it's a *soft* loss the `webglcontextlost`/`webglcontextrestored` machinery never sees — distinct from the null-style crashes in #7710/#7726, which come from unguarded public accessors *after* a real context loss.

> NB! the above diagnosis is inferred from the symptom, not independently proven.

## Change

Remove the framebuffer-completeness check and let WebGL handle an incomplete framebuffer the way it already does — the draws no-op and the map recovers on the next repaint. Same defensive direction taken for the sibling null-style accessors in #7726. With the check gone the error can no longer originate in our code, so the now-dead `isFramebufferNotCompleteError` carve-out in `triggerRepaint()` and the `framebuffer_error` helper are removed too.

## Behavior

| Scenario | Before | After |
|---|---|---|
| Transient incomplete framebuffer (e.g. tab wakes from sleep) | Constructor throws → unhandled global error | No throw; draws no-op, map recovers next repaint |
| Persistent incomplete framebuffer | Throws (crashes the frame) | No throw; region renders blank/stale *(the tradeoff)* |

If a signal is preferred for the persistent case, the call site could `fire(new ErrorEvent(...))` before bailing instead of staying silent.

## Checklist

Cowritten by claude Opus 4.8


src/gl/framebuffer.ts
```diff
@@ -28,6 +28,11 @@ export class Framebuffer {
         } else if (hasStencil) {
             throw new Error('Stencil cannot be set without depth');
         }
+    }
+
+    checkFramebufferStatus() {
```
HarelM: This isn't really checking status, does it?

johanrd: @HarelM Good catch — I think this name is papering over a deeper question about whether the check should exist at all. Some history:

It started as a debug-only assert: https://github.com/maplibre/maplibre-gl-js/blob/00d702689a734152bfd0fb1e69f4c1ed46a6610a/src/gl/framebuffer.js#L26 as a sanity check for "did I forget an attachment?" in the debug build.

#1485 turned it into a production throw, and [you flagged exactly this risk on that PR](https://github.com/maplibre/maplibre-gl-js/pull/1485#issuecomment-1214208763):

>The only main "issue" I see work this approach is that there is a functionally change in this commit: before this code simply didn't exist in production so if there was an assert that might be triggered but didn't cause any actual error the user wouldn't know as opposed to now where there might be "throws" that cause actual problems.

That seems to reflect what's showing up in Sentry now.

Then https://github.com/maplibre/maplibre-gl-js/pull/5266 tightened the https://github.com/maplibre/maplibre-gl-js/blob/e6d6923dcf1b6a562af762f18b7e06fd35d74e75/src/ui/map.ts#L3384 but specifically carved out framebuffer-not-complete.

### Zooming out
Given all that, **I'd suggest deleting ensureFramebufferComplete() and all 5 call sites** instead of renaming it. After that, isFramebufferNotCompleteError and the redraw() try/catch guard an error that can no longer be thrown from our code, and can go too.

The alternative is developing an active recovery path (force context loss/restore on failure), but honestly WebGL may handle a transient framebuffer blip fine on its own if we just get out of the way. Happy to restructure the PR either direction.

HarelM: Given the history you dug, I trust you understand this better than me, so if you think a direction is worth pursuing, feel free to do so.
I would like to get @birkskyum and @ToHold 's input here since @birkskyum recently fixed a long lasting bug with framebuffer misalignment and @ToHold did a lot of work around recovery from context loss.

HarelM: CC: @mvanhorn who's also added some PRs around context restored.

Is there a sudden interest in making this work well which wasn't a big motivation before?

johanrd: > Is there a sudden interest in making this work well which wasn't a big motivation before?

I am no authority on the solution, but did my best to answer why there is a sudden interest.

**Pre-2025:** Essentially no active context-loss work for ~4 years.  **Then  [#6242](https://github.com/maplibre/maplibre-gl-js/issues/6242) was opened 2025-08-12** — first reproducible [jsbin](https://jsbin.com/gesoyacivu/1/edit?html,js,console,output) showing `_contextRestored` leaves the map blank.

**Since then:**

| Date | PR | Notes |
|---|---|---|
| 2025-11-04 | [#6398](https://github.com/maplibre/maplibre-gl-js/pull/6398) | ToHold's fix for #6242: [`this.painter.destroy()`](https://github.com/maplibre/maplibre-gl-js/blob/c87db5b135ddefe60ad711077f6ae2106f59a324/src/ui/map.ts#L3289) + [iterate `this.style._layers`](https://github.com/maplibre/maplibre-gl-js/blob/c87db5b135ddefe60ad711077f6ae2106f59a324/src/ui/map.ts#L3292) + [`this.style = null`](https://github.com/maplibre/maplibre-gl-js/blob/c87db5b135ddefe60ad711077f6ae2106f59a324/src/ui/map.ts#L3306) in `_contextLost`, [`setStyle(_lostContextStyle.style)`](https://github.com/maplibre/maplibre-gl-js/blob/c87db5b135ddefe60ad711077f6ae2106f59a324/src/ui/map.ts#L3312) on restore. |
| 2026-01-07 | [#6935](https://github.com/maplibre/maplibre-gl-js/pull/6935) | [Early-return guard in `resize()`](https://github.com/maplibre/maplibre-gl-js/blob/1b949645a00f255dd136aae646b185d02d0ab1a3/src/ui/map.ts#L995-L996) — failure mode from #6398's `painter.destroy()`. |
| 2026-02-10 | [#7094](https://github.com/maplibre/maplibre-gl-js/pull/7094) | [Null-style early-return](https://github.com/maplibre/maplibre-gl-js/blob/67c6b3382d06ba8dcfb7d72b973453740322b722/src/ui/map.ts#L3435-L3438) in `_contextLost` — failure mode from #6398 iterating `this.style._layers`. Fixes [#7022](https://github.com/maplibre/maplibre-gl-js/issues/7022) ("v5.12+" regression). |
| 2026-02-13 | [#7126](https://github.com/maplibre/maplibre-gl-js/pull/7126) | Unrelated: texture handle corruption. Fixes [#2811](https://github.com/maplibre/maplibre-gl-js/issues/2811) (opened 2023-07-06). |
| 2026-04-11 | [#7446](https://github.com/maplibre/maplibre-gl-js/pull/7446) | [`this.style` null-check](https://github.com/maplibre/maplibre-gl-js/blob/f2072c313752926cf3dd6d142a42e78ec8a1edb9/src/ui/map.ts#L3535-L3536) in `_contextRestored` — failure mode from #6398 setting `this.style = null`. Fixes [#7432](https://github.com/maplibre/maplibre-gl-js/issues/7432) ("v5.22+" regression). |
| now | #7303 | Framebuffer-not-complete on tab wake (Sentry production error; not in #6242 lineage). |

Tu my understanding, #6398 introduced an eager teardown approach that turned "GL handles invalid between lost/restored" into "this.style can be null at unexpected moments" — #6935, #7094, and #7446 each guard a path that became  null-reachable as a result.

HarelM:
Regardless of the need to get some feedback from other developers you'd need to resolve conflicts.

HarelM:
@johanrd any updates on this? We are in the middle of version 6 pre-release, so if there's a breaking change worth introducing, now is the time.

mvanhorn:
@HarelM hit context-restored bugs in #7446 while building a webgl-heavy app, just patching what surfaced. No coordinated push.

johanrd:
@HarelM hi, sorry for not coming back to you on this before now. To be honest, I already felt that I stretched my competence already when doing the "digging into the history" part of the diagnosis – I don't have enough experience with WebGL to confidently stand by one solution above another.

By Occam's razor it's compelling to let maplibre get out of the way and let WebGL recover on its own, so I updated the PR now so that it just removes the framebuffer-completeness check (constructor throw + the dead isFramebufferNotCompleteError carve-out + the helper). 

Now there is something more concrete to review / discuss, at least.

HarelM:
Overall, this looks good to me assuming you've tested it and see that these errors are no longer happening and that the context lost is handled correctly.

HarelM (approved):

Issue #7271, opened 2026-03-17
Enable touch zoom speed control


gowestmen:
I created my own custom basemap. I took special care to ensure that it is “progressive,” meaning that the graphical rendering of the data (roads, rivers, city names, buildings, etc.) changes smoothly as the zoom level changes, thereby reducing the risk of abrupt style changes.

To highlight this progressive aspect of my basemap, I reduced the zoom speed when using the scroll wheel, using the following methods: 

`map.scrollZoom.setWheelZoomRate()   // for the mouse scroll wheel`
 `map.scrollZoom.setZoomRate()   // for touchpad zoom`

Now, I would like to reduce the zoom speed for touch interactions (pinch-to-zoom and “tap and drag”), but it seems there is no equivalent to the two methods above.

I am not really sure how contributing to Maplibre works, but I would be happy to help !

HarelM:
There are alot if readmes secreted in statistic places in this repo. Reading those should help you get things going. 
The best way to start contributing is trying to change the code and open a PR.
The PR template has instructions on what you should be looking for after changing the code - tests, changelog etc.
If those doesn't work you can join the slack channel (maplibre-gl-js) to ask some questions. 
If all of this doesn't work let me know and I'll try to help. 
Generally speaking this shouldn't be complicated to achieve, I think.

laem:
> I created my own custom basemap. I took special care to ensure that it is “progressive,” meaning that the graphical rendering of the data (roads, rivers, city names, buildings, etc.) changes smoothly as the zoom level changes, thereby reducing the risk of abrupt style changes.

Do you have something that we can test online ? Sounds interesting !

gowestmen:
Thank you for your interest in this work. 

In fact, this is part of a larger French research project called [LostInZoom](https://lostinzoom.github.io/home/). The project aims to understand how users interact with interactive maps and how to improve them.

We have deployed a basemap on a small area that attempts to incorporate this “progressive” aspect. You can test it here: 
https://lostinzoom.huma-num.fr/thegoodmap/

For now, only the basemap is presented, but we plan to deploy a map prototype with new types of zoom interactions (notably including zoom speed control).

laem:
Cool thanks ! You might be interested by what we're doing on cartes.app. 

We have a big base.ts file as a dynamic style. https://codeberg.org/cartes/web/src/branch/master/app/styles/base.ts

[...]

PR #7305, opened 2026-03-20
Expose touch zoom speed and trigger threshold 

gowestmen:
## Launch Checklist

 - [x] Allow the touch zoom speed to be adjusted via  
        `map.touchZoomRotate.setTwoFingersZoomSpeed() // for two fingers pinch zoom`
        `map.touchZoomRotate.setTapDragZoomSpeed() // for tap drag zoom`
        These methods modify a property used to multiply the zoom delta during its calculation.
       Also added a method that allows to set the zoom threshold for two-finger interactions. Previously, this value was fixed (0.1); it can now be modified via `map.touchZoomRotate.setZoomThreshold()` to adjust the sensitivity to pinch gestures (a higher value means that zooming will be triggered less easily when the pinch gesture is very slight).
 - [x] #7271 

### DIFF
--- a/src/ui/handler/two_fingers_touch.ts
+++ b/src/ui/handler/two_fingers_touch.ts
@@ -198,7 +198,7 @@ export class TwoFingersTouchZoomHandler extends TwoFingersTouchHandler {
         if (!this._active && Math.abs(getZoomDelta(this._distance, this._startDistance!)) < this._zoomThreshold) return;
         this._active = true;
         return {
-            zoomDelta: (getZoomDelta(this._distance, lastDistance))*this._twoFingersZoomSpeed,
+            zoomDelta: (getZoomDelta(this._distance, lastDistance)) * this._twoFingersZoomSpeed,
             pinchAround
         };
     }


HarelM:
Note that I'm in the middle of releasing a new version, so if you want this released soon you'll need to hurry up :-)

src/ui/handler/tap_drag_zoom.ts
```diff
@@ -70,7 +72,7 @@ export class TapDragZoomHandler implements Handler {
             this._active = true;
 
             return {
-                zoomDelta: dist / 128
+                zoomDelta: (dist / 128)*this._tapDragZoomSpeed
```
HarelM:
```suggestion
                zoomDelta: (dist / 128) * this._tapDragZoomSpeed
```

src/ui/handler/two_fingers_touch.ts
```diff
@@ -25,10 +25,16 @@ abstract class TwoFingersTouchHandler implements Handler {
     _vector?: Point;
     _startVector?: Point;
     _aroundCenter?: boolean;
+    _twoFingersZoomSpeed?: number;
+    _zoomThreshold?: number;
 
     /** @internal */
     constructor() {
         this.reset();
+        if (this._twoFingersZoomSpeed === undefined) {
```
HarelM: How can this if be false in the constructor?

src/ui/handler/tap_drag_zoom.ts
```diff
@@ -14,14 +14,16 @@ export class TapDragZoomHandler implements Handler {
...
+    _tapDragZoomSpeed: number;
 
     constructor() {
 
         this._tap = new TapRecognizer({
             numTouches: 1,
             numTaps: 1
         });
-
+        if (this._tapDragZoomSpeed === undefined) {
```
HarelM: How can this if be false in the constructor?

src/ui/handler/shim/two_fingers_touch.ts
```diff
@@ -109,4 +109,43 @@ export class TwoFingersTouchZoomRotateHandler {
...
+    * Modify the speed of two fingers touch zoom
+    * @param zoomSpeedRatio - 1 The ratio used to multiply two fingers zoom delta value (resulting in speed changes).
+    * @example
+    * Speed up two fingers zoom
+    * ```ts
+    * map.touchZoomRotate.setTwoFingersZoomSpeed(1.25);
+    * ```
+    */
+    setTwoFingersZoomSpeed(zoomSpeedRatio: number) {
+        this._touchZoom._twoFingersZoomSpeed = zoomSpeedRatio;
```
HarelM: It might be better to add a public `setZoomSpeed` to the relevant class instead of using the "private" fields...

HarelM:
I think some unit tests are missing here, can you add those please?

src/ui/handler/two_fingers_touch.ts
gowestmen: I think I miss understood how constructor works. Since I just want to set the default value of _twoFingersZoomSpeed to 1, I guess it should works if I just keep `this._twoFingersZoomSpeed = 1`

src/ui/handler/tap_drag_zoom.ts
gowestmen: Same as 
> I think I miss understood how constructor works. Since I just want to set the default value of _twoFingersZoomSpeed to 1, I guess it should works if I just keep `this._twoFingersZoomSpeed = 1`

src/ui/handler/shim/two_fingers_touch.ts
gowestmen: My idea was that by declaring this method here, in the `TwoFingersTouchZoomRotateHandler` class, users could access it via `map.touchZoomRotate.setTwoFingersZoomSpeed()` or `map.touchZoomRotate.setTapDragZoomSpeed()`, which doesn't seem illogical to me. 

Do you mean creating a `setZoomSpeed()` method directly in the `TwoFingersTouchHandler` and `TapDragZoomHandler` classes, from which `TwoFingersTouchZoomRotateHandler` inherits?

gowestmen:
> Note that I'm in the middle of releasing a new version, so if you want this released soon you'll need to hurry up :-)

I think I was too late. Should I modify my CHANGELOG.md file to put my changes in the current "main" part, or add them in the "5.21.0" part ?

HarelM:
Yes, please update the changelog to move the change up under main.

src/ui/handler/shim/two_fingers_touch.ts
HarelM: There's no inheritance, but yes, add public methods.

gowestmen: I've modified my code so that there is now a setZoomSpeed() method in the TapDragZoomHandler and TwoFingersTouchHandler classes.

Users can access these methods using the following code: 
`map.touchZoomRotate._touchZoom.setZoomSpeed()` and  `map.touchZoomRotate._tapDragZoom.setZoomSpeed()`

gowestmen:
> I think some unit tests are missing here, can you add those please?

I'm not very familiar with unit tests. I generated a few using AI for the methods I'm trying to add. I tested them locally by running `npm run test-unit -- example.test.ts`, and it seemed to work.

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ## main
 ### ✨ Features and improvements
-- _...Add new stuff here..._
+- Add `_twoFingersZoomSpeed` & `_tapDragZoomSpeed` properties and expose methods to change it to control zoom speed. Add `_zoomThreshold` and relevant method to control sensitivity to pinch move that trigger zoom ([#7271](https://github.com/maplibre/maplibre-gl-js/issues/7271))
```
HarelM, line 3: Internal field are less interesting when looking at the changelog, I advise to comment on what public APIs were added/changed.

src/ui/handler/two_fingers_touch.ts
```diff
@@ -167,10 +195,10 @@ export class TwoFingersTouchZoomHandler extends TwoFingersTouchHandler {
     _move(points: [Point, Point], pinchAround: Point | null): HandlerResult | void {
         const lastDistance = this._distance!;
         this._distance = points[0].dist(points[1]);
-        if (!this._active && Math.abs(getZoomDelta(this._distance, this._startDistance!)) < ZOOM_THRESHOLD) return;
+        if (!this._active && Math.abs(getZoomDelta(this._distance, this._startDistance!)) < this._zoomThreshold) return;
         this._active = true;
         return {
-            zoomDelta: getZoomDelta(this._distance, lastDistance),
+            zoomDelta: (getZoomDelta(this._distance, lastDistance))*this._twoFingersZoomSpeed,
```
HarelM:
```suggestion
            zoomDelta: (getZoomDelta(this._distance, lastDistance)) * this._twoFingersZoomSpeed,
```

src/ui/handler/two_fingers_touch.ts
```diff
@@ -167,10 +195,10 @@ export class TwoFingersTouchZoomHandler extends TwoFingersTouchHandler {
     _move(points: [Point, Point], pinchAround: Point | null): HandlerResult | void {
         const lastDistance = this._distance!;
         this._distance = points[0].dist(points[1]);
-        if (!this._active && Math.abs(getZoomDelta(this._distance, this._startDistance!)) < ZOOM_THRESHOLD) return;
+        if (!this._active && Math.abs(getZoomDelta(this._distance, this._startDistance!)) < this._zoomThreshold) return;
         this._active = true;
         return {
-            zoomDelta: getZoomDelta(this._distance, lastDistance),
+            zoomDelta: (getZoomDelta(this._distance, lastDistance)) * this._twoFingersZoomSpeed,
```
HarelM, line 201: I think it will be better to include in inside the `getZoomDelta` method instead of here.

src/ui/handler/two_fingers_touch.ts
```diff
@@ -127,6 +131,32 @@ abstract class TwoFingersTouchHandler implements Handler {
...
         return !!this._active;
     }
+
+    /**
+    * Modify the speed of two fingers touch zoom
+    * @param zoomSpeedRatio - 1 The ratio used to multiply two fingers zoom delta value (resulting in speed changes).
+    * @example
+    * Speed up two fingers zoom
+    * ```ts
+    * map.touchZoomRotate._touchZoom.setZoomSpeed(1.25);
```
HarelM, line 141: That's not a great example, as it uses `_` field which is internal...

src/ui/handler/tap_drag_zoom.test.ts
```diff
@@ -112,4 +112,52 @@ describe('tap_drag_zoom', () => {
...
+        map._renderTaskQueue.run();
+
+        const zoomNormal = map.getZoom();
+
+        map.remove();
+
+        const map2 = createMap();
+        const target2 = map2.getCanvas();
+
+        map2.touchZoomRotate._tapDragZoom.setZoomSpeed(2);
```
HarelM, line 141: If this is the expected way to use this feature it's not a good solution, this needs to be part of the public API of `touchZoomRotate` instance/class.

Also looking at the test here, it looks like this should be split into two tests.

src/ui/handler/two_fingers_touch.ts
```diff
@@ -127,6 +131,32 @@ abstract class TwoFingersTouchHandler implements Handler {
...
+    /**
+    * Modify the threshold to trigger two fingers zoom
+    * @param zoomThreshold - 0.1 The threshold value used to trigger two fingers zoom interaction depending on pinch move size.
+    * @example
+    * Increaze the two fingers zoom trigger threshold (a wider pinch gesture will be required to trigger the interaction)
+    * ```ts
+    * map.touchZoomRotate._touchZoom.setZoomThreshold(0.3);
+    * ```
+    */
+    setZoomThreshold(zoomThreshold: number) {
```
HarelM, line 157: How does one revert this change to "default" value?
In other places we usually allow setting `undefined` as an input parameter to revert to default value. Please look at other APIs and see if this would be a good approach here too.

gowestmen: For now, the only way to do this is to re-execute the function by passing 0.1 (the default value) as a parameter.

For example, the existing function for configuring the wheel zoom rate does not work if called like this: `map.scrollZoom.setWheelZoomRate(undefined)` or `map.scrollZoom.setWheelZoomRate()` (it even disables wheel zooming). The behavior appears to be the same with this `setZoomThreshold()` function.

I assume it is possible to set a default value for the `zoomThreshold` argument of the `setZoomThreshold()` function, so that it can be used as you mention.

src/ui/handler/tap_drag_zoom.test.ts
gowestmen: Okay, would this kind of function work?
`setZoomSpeed(zoomSpeedRatio: number, zoomType: string) {
if (zoomType == 'pinch') {this._touchZoom._twoFingersZoomSpeed = zoomSpeedRatio;} 
else if (zoomType == 'tap_drag')  {this._tapDragZoom._tapDragZoomSpeed = zoomSpeedRatio;}}`

Users should then be able to use the function by calling `map.touchZoomRotate.setZoomSpeed(2, ‘tap_drag’)`. 

As for testing, I’m only testing the function with a ratio of 2, which increases the zoom speed. Do you mean I should also test with a reduction ratio, or do you have something else in mind?

HarelM:
Closing this one in favor of 
- #7384

If this is something else please open a different PR.

PR #7320, opened 2026-03-24
Add missing promoteId parameter to geojson worker

### LINKED ISSUE #7315: GeoJSONSource.updateData() ignores feature property updates when using (string) promoteId
**maplibre-gl-js version**: 5.21.0
Regressed in 5.20.0+
Worked in: 5.19.0  

**browser**: Chrome 146

### Steps to Trigger Behavior

 1. Create a GeoJSON source using a string-based `promoteId`
 2. Update feature properties using the `.updateData()` function of the geojson source. Reference the features via their promoted Id.

### Link to Demonstration

https://jsbin.com/boqimoy/edit?html,js,console,output

Expected behavior in the reproduction:

Clicking the button in the top left corner should change the color of the rendered circle.

This can be accomplished by:
- Downgrading to 5.19.0, or
- Referencing features via their native `id` (without using `promoteId`)

When using `promoteId`, updates are not applied regardless of whether the promoted value is a string or a number.

### Expected Behavior

Features can be referenced via their promoted id in the `GeoJSONSourceDiff` supplied to the `.updateData()` function.

This behavior is described in the [docs](https://maplibre.org/maplibre-gl-js/docs/API/classes/GeoJSONSource/#updatedata):

> This approach requires unique IDs for every feature in the source. The IDs can either be specified on the feature, or by using the promoteId option to specify which property should be used as the ID.

### Actual Behavior

- Feature properties are not updated when referenced via promoted ID
- No errors or warnings in console
- Behavior silently fails

HarelM:
## Launch Checklist

- Fixes #7315 

This adds the missing promoteId parameters to geojson-vt initialization options.
I was not able to write a reasonable test to check this since the code that copies the goejson-vt options is inside the factory method "outside" the class.
I thought about moving all the logic inside the class and just leave the `new GeoJSONVT(...)` in the factory method, but that seems a bit too heavy for this fix.
IDK...

 - [x] Link to related issues.
 - [x] Update test for the new functionality

### DIFF
--- a/src/source/geojson_worker_source.test.ts
+++ b/src/source/geojson_worker_source.test.ts
@@ -563,8 +563,8 @@ describe('loadData', () => {
                 clusterOptions: {},
             },
             clusterProperties: {
-                "max": ["max", ["get", "scalerank"]],
-                "sum": ["+", ["get", "scalerank"]],
+                'max': ['max', ['get', 'scalerank']],
+                'sum': ['+', ['get', 'scalerank']],
             }
         } as LoadGeoJSONParameters)).resolves.toBeDefined();
         expect(updateSpy).toHaveBeenCalled();


HarelM:
Should I be changing the `promoteId` to be passed as part of the geojsonvt options instead of "outside" this object? I think the only reason to pass it to the worker now, as opposed to before is to pass it down to the geojson-vt, so this options my be suited there, right?

wayofthefuture:
Yes

HarelM:
I changed the worker communication object to align with the geojson options object since now everything is one object and there's no need to split the options, as far as I see.
@wayofthefuture let me know if you think I missed anything.
This is somewhat of an expected step given the change we did in geojson-vt to "inline" supercluster.
Previous related PR:
- #7172

wayofthefuture (approved):
Looks good

HarelM:
I'll add missing coverage while I'm at it...

PR #7330, opened 2026-03-25
fix: handle undefined projection in setProjection to prevent setState crash

### LINKED ISSUE #7314: Style diff in globe projection falls back to full rebuild due to setProjection handling
**maplibre-gl-js version**: 5.21.0

**browser**: Chrome 146, Firefox 140

When globe projection is active, `map.setStyle()` always falls back to a full style rebuild instead of performing an incremental diff. This causes unnecessary tile re-fetching, a visible flash, and resets the map projection.

### Root Cause

`Style.setState()` serializes the current style and diffs it against the incoming style. In globe projection, the serialized style includes projection: {"type":"globe"}, but most style specs don't include a projection field at all. The diff produces a `setProjection` command with args: [undefined], and the handler crashes trying to access it:

"Unable to perform style diff: Cannot read properties of undefined (reading 'type'). Rebuilding the style from scratch."

In Mercator projection, projection is not present in the origin serialized style either, so the diff succeeds.

### Link to Demonstration

https://jsbin.com/gofikev/edit?html,js,output

1. Open the demo (starts in Mercator)
2. Click "Switch style" - observe smooth, instant switch (diff works)
3. Click "Toggle projection" to switch to Globe
4. Click "Switch style" - observe full repaint, tiles reload, and the console warning

A commented-out workaround is included that uses `transformStyle` to carry forward projection from the serialized style, preventing the diff from seeing the change.

### Expected Behavior:
projection is a viewport property like center, zoom, bearing, pitch, and roll - all of which are skipped as no-ops in `_getOperationsToPerform()`. setProjection should be handled the same way, so the style diff doesn't attempt to reset the projection when the target style simply omits it.

ashwinuae:
`style.setState()` crashes when switching from a style with an explicit `projection` field to one without. The `diff()` function emits `{command: 'setProjection', args: [undefined]}`, and `setProjection(undefined)` crashes on `undefined.type`. The fix skips `setProjection` diff commands in `_getOperationsToPerform` via `continue`, matching the existing pattern for other map-owned viewport properties (`setCenter`, `setZoom`, `setBearing`, `setPitch`, `setRoll`). Projection is owned by the map, not the style and `setState` shouldn't mutate it as a side effect of a style diff.                                           

Fixes #7314

## Launch Checklist
 - [x] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### 🐞 Bug fixes
 - Make `fitBounds` and `fitScreenCoordinates` respect the `zoomSnap` map option by snapping the zoom level down to keep bounds fully visible ([#7332](https://github.com/maplibre/maplibre-gl-js/issues/7332) (by [@CommanderStorm](https://github.com/CommanderStorm))
 - Make `jumpTo`, `easeTo`, and `flyTo` respect the `zoomSnap` map option by snapping the zoom level to the nearest valid increment ([#7333](https://github.com/maplibre/maplibre-gl-js/issues/7333) (by [@CommanderStorm](https://github.com/CommanderStorm))
-- Fix `setState` crash when switching styles while globe projection is active ([#7314](https://github.com/maplibre/maplibre-gl-js/issues/7314))
+- Fix `setState` crash when switching styles while globe projection is active ([#7314](https://github.com/maplibre/maplibre-gl-js/issues/7314)) (by [@ashwinuae](https://github.com/ashwinuae))
 
 ## 5.21.1
 


HarelM:
> Projection is owned by the map, not the style and setState shouldn't mutate it as a side effect of a style diff.

I don't think I share that view, as opposed to zoom, center etc which can be defined in the style but are ignored when using set style I would expect projection to change the projection when specified in the style.

Moreover, I'm not sure I'm aligned with the fact that setting zoom, center, etc only work on initial load and not afterwards, but that's a different story and probably related to some historic stuff. 
I would consider supporting undefined for setProjection to "return to default" kind of behavior, I think it might solve this issue as well, wouldn't it?

HarelM:
Thanks for taking the time to open this PR BTW!

ashwinuae:
Good point, on re-read I can see that I overstated the "projection is owned by the map" framing in the original description. You're right that projection is part of the style spec and should be applied when explicitly specified.

The issue again is with styles that don't specify a projection; since most styles out there don't include a projection field, the diff produces `setProjection(undefined)` whenever the user switches styles in globe mode which crashes because `setProjection` doesn't handle undefined. A "return to default" approach would fix the crash here, but to me the silent projection reset feels off from a UX standpoint.

I think a good in-between approach is to only skip when the arg is undefined, but still apply explicit projection changes:                                                                                                                                                                                                                                                             
```
case 'setProjection':
    if (op.args[0] !== undefined) {
        operations.push(() => this.setProjection.apply(this, op.args));
    }
    break;
```

So style A (globe) → style B (mercator) changes projection as expected, but style A (globe) → style B (no projection field) preserves the current projection without crashing. 

(And yes, totally understandable w.r.t the historic stuff.)

Lemme know if this looks good, and I'll proceed to push!

HarelM:
Can you elaborate a bit on the use case here? 
Why would you expect the projection to be kept and not "reset"?

I think the approach of checking for non undefined projection is a good middle ground and definitely better than crashing, I just want to better understand why you think it's a bad UX to reset the projection in case it's not defined in the style (since the default is Mercator it makes sense not to specify it in the style).

neodescis:
What do we do with other style properties? I don't think we skip over anything else (besides properties related to a user navigating around the map) that isn't in the style and leave it as-is, do we? For example, sky, light, sprite, transition? I really think we need to be consistent. Furthermore, changing this behavior would be a (potentially large for some) breaking change.

Also, it's not really so much work to query the current projection and fold it into a new style before calling setStyle, is it?

Obviously the crash should be fixed though!

ashwinuae:
Thanks for the comments; they helped give me more clarity! I guess I was fixated on the projection handling alone and the "skip vs reset" question when the answer should be consistency. @neodescis you're right that sky, light, sprite, etc. all get reset when missing from the target style, so projection should behave the same way. The fix should belong in `setProjection` itself then; handle undefined by falling back to 'mercator' (same as the default when no projection is specified).

I've updated the PR now, lemme know if it looks fine. Thanks! :)

neodescis:
LGTM, but I'll let @HarelM take a look and merge. Thanks @ashwinuae!

src/style/style.ts
```diff
@@ -1685,15 +1685,16 @@ export class Style extends Evented {
 
     setProjection(projection: ProjectionSpecification) {
         this._checkLoaded();
-        this.stylesheet.projection = projection;
+        const resolvedProjection = projection ?? {type: 'mercator'};
+        this.stylesheet.projection = resolvedProjection;
```
HarelM: I think it would be better to update the style with the given projection, otherwise the style you set and the style you serialize might be different.
I.e. set the value to the stylesheet object but resolve to default for the setProjection method. Makes sense?

HarelM: Probably also worth adding a test for the serialization part as well.

ashwinuae: Ah yep, makes sense. Done now!

CHANGELOG.md
```diff
@@ -3,9 +3,9 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
-- _...Add new stuff here..._
 - Make `fitBounds` and `fitScreenCoordinates` respect the `zoomSnap` map option by snapping the zoom level down to keep bounds fully visible ([#7332](https://github.com/maplibre/maplibre-gl-js/issues/7332) (by [@CommanderStorm](https://github.com/CommanderStorm))
 - Make `jumpTo`, `easeTo`, and `flyTo` respect the `zoomSnap` map option by snapping the zoom level to the nearest valid increment ([#7333](https://github.com/maplibre/maplibre-gl-js/issues/7333) (by [@CommanderStorm](https://github.com/CommanderStorm))
+- Fix `setState` crash when switching styles while globe projection is active ([#7314](https://github.com/maplibre/maplibre-gl-js/issues/7314))
```
```suggestion
- Fix `setState` crash when switching styles while globe projection is active ([#7314](https://github.com/maplibre/maplibre-gl-js/issues/7314)) (by [@ashwinuae](https://github.com/ashwinuae))
```

PR #7332, opened 2026-03-26
feat(camera): implement `zoomSnap` for `fitBounds` and `fitScreenCoordinates`

CommanderStorm:
We sort of missed this. It is triggered by the `fitBounds` and respective screen space methods.


src/ui/camera.ts
```diff
@@ -840,7 +840,11 @@ export abstract class Camera extends Evented {
         const tr = this.transform;
         const bounds = new LngLatBounds(p0, p1);
 
-        return this.cameraHelper.cameraForBoxAndBearing(options, padding, bounds, bearing, tr);
+        const result = this.cameraHelper.cameraForBoxAndBearing(options, padding, bounds, bearing, tr);
+        if (result && this._zoomSnap) {
+            result.zoom = evaluateZoomSnap(result.zoom, this._zoomSnap, -1);
```
HarelM, line 845: Should this be hardcoded -1 or should snapZoom value be passed here?

CommanderStorm: `-1` is good, this is "please zoom out", don't round the zoom to the nearest.
For this operation, rounding would be suboptimal and not what I would expect when a user asks to fit this bbox in the viewport.

test/build/bundle_size.json
```diff
@@ -1 +1 @@
-1048557
+1023660
```
```suggestion
1049586
```

PR #7338, opened 2026-03-26
feat(camera): Make `jumpTo`, `easeTo`, and `flyTo` respect the `zoomSnap` map option by snapping the zoom level to the nearest valid increment

CommanderStorm:
Not sure if we want to add this here as well. I think it makes sense for consistency.
There is no "this is better completely" because these are all functions that get zoom passed as the input, so essentially we are partially rejecting user input on this one.
I think this is still sommehwat what we should to though


CommanderStorm:
> What's the use case for this?

To not have to have the same code externally, so the "not having to think about it" ^^

PR #7347, opened 2026-03-28
feat: add auto hide compass on neutral values

Its-Just-Nans:
Add an auto hide option for the compass

Magic commands are

```sh
bun install
bun run watch-dev
```

## Launch Checklist

### DIFF
--- a/src/ui/control/navigation_control.ts
+++ b/src/ui/control/navigation_control.ts
@@ -114,7 +114,6 @@ export class NavigationControl implements IControl {
                 }
                 return Math.abs(value) < epsilon;
             });
-            console.log({transforms, areNeutral});
             if (areNeutral) {
                 this._compass.style.display = 'none';
             }else{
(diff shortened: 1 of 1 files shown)


HarelM:
It's not clear what this feature is aimed to solve and what this new parameter is doing. Can you please elaborate?

Its-Just-Nans:
> Thanks for taking the time to open this PR.
> It's not clear what this feature is aimed to solve and what this new parameter is doing. Can you please elaborate?

It's used to auto hide the compass when the compass is on "default" setting (not rotation)

HarelM:
I'm not sure this should be part of this library to be honest, why not simply solve this using CSS?

Its-Just-Nans:
> why not simply solve this using CSS?

I'm very curious on how you can do a CSS only solution for that ??

HarelM:
Listen to rotation event and apply a csd class for 0 degrees...?

Its-Just-Nans:
> Listen to rotation event and apply a csd class for 0 degrees...?

Like having

```js

map.on('moveend', (e)=> {
// Find the navigation control in e, then the compass, then parse the transform style props, then hide if necessary
})
```

Something like this ?

HarelM:
Something like that, yeah.

Its-Just-Nans:
> Something like that, yeah.

That's why I thought a MR would be useful. Since we don't have to do the reparsing. And this option is generic over all different maps

Feel free to close if you don't think that way

HarelM:
I tend to think this is very specific to your use case, sorry...

PR #7350, opened 2026-03-29
test: add more render tests for various visibility/line/.. things

CommanderStorm:
## Launch Checklist

During unrelated work, I noticed that a few of our properties are missing a bit of render-test coverage. This improves the coverage in this area.

I used claude for generating the test data, since this is essentially pattern matching with existing tests, I think this is fair.
I reviewed every test thoroughly.


test/integration/render/tests/line-miter-limit/default/expected.png
CommanderStorm, line 1: This looks funny and is likely more in the "bug" category.
Still good to have the test since this is the default..

test/integration/render/tests/text-ignore-placement/false/expected.png
HarelM, line 1: What's the difference between this one and:
`‎test/integration/render/tests/text-ignore-placement/true/`?
I would expect the render test to be different, shouldn't they?

HarelM:
Thanks for taking the time to add these tests!
It's worth considering which tests are actually valuable to add as it takes longer for the CI to run all these tests, and since these are end to end test and not unit test they run a bit slow.
I've seen a few cases where I did not understand what the test is testing, when setting something to true or false yield the same results (about 2-3 tests).
I would consider adding global-state related visibility tests if those are not already present (since you added some visibility tests, global-state is an important aspect I think).

Overall, these are good to have, THANKS!

test/integration/render/tests/text-ignore-placement/false/expected.png
CommanderStorm: `text-ignore-placement` is anoying. Works exactly counterintitive

test/integration/render/tests/symbol-avoid-edges/false/expected.png
CommanderStorm, line 1: https://maplibre.org/maplibre-style-spec/layers/#symbol-avoid-edges is lying.
This was never implemnted in GL-JS.

CommanderStorm: Actually, it was just removed in favor of global placement

test/integration/render/tests/icon-keep-upright/false/symbol-placement-line-center/expected.png
HarelM, line 1: I think the name of this folder is misleading, as it seems that it's related to line-placement while it's not.

test/integration/render/tests/icon-keep-upright/true/symbol-placement-line-center/expected.png
HarelM, line 1: See previous comment.

test/integration/render/tests/symbol-avoid-edges/false/expected.png
HarelM: Is this image different than `test/integration/render/tests/symbol-avoid-edges/true`?

HarelM:
I've added a few more comments.

test/integration/render/tests/symbol-avoid-edges/false/expected.png
CommanderStorm: Negative, since this property is siletly ignored.
But if we ignore a prop, we should also have tests that we ignore it

CommanderStorm: and the CI is not slow becasue of this part, but because we are not on rolldown. If we were on rolldown, the 4m building the dist would fall away instead of the 2.5m of actually running tests

HarelM:
Rolldown doesn't support the complicated build we have here with the worker code shared etc, but maybe for next version we can drop the commonjs support and have be able to use more modern build tools...

CommanderStorm:
> I would consider adding global-state related visibility tests if those are not already present (since you added some visibility tests, global-state is an important aspect I think).

There is a lot of work left in this department, but I think this is making a dent there

HarelM:
I'm having a hard time reviewing this as it's hard to keep track of what was fixed, added etc.
Can this PR be split according to stuff that were approved and keep new tests in another PR?
Sorry for the extra work, but I'm lost here... 😕

CommanderStorm:
needs to be deconstucted

PR #7351, opened 2026-03-29
feat(line): implement data driven styling for `line-{cap,round-limit,miter-limit}`

CommanderStorm:
 
I think we just forgot this part...
CI will fail as long as we don't adjust the style spec to allow this

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -117,7 +117,7 @@
         "tslib": "^2.8.1",
         "typedoc": "^0.28.18",
         "typedoc-plugin-markdown": "^4.11.0",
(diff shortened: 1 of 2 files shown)


HarelM:
Looks good, thanks for looking into this.
Do you happen to know why this was left unimplemented? Maybe it has a performance penalty that makes this feature problematic?

CommanderStorm:
> Do you happen to know why this was left unimplemented

It seems it was not done due to the complexities around line-dash-array it seems.

- https://github.com/mapbox/mapbox-gl-js/pull/5018/

PR #7352, opened 2026-03-30
Clamp mercator horizon for high pitch angles

Timorleiderman:
### Motivation
- Prevent negative horizon values and ground disappearance when the camera pitch exceeds about 89 degrees in Mercator mode.
- This surfaced in downstream testing in a web game that uses MapLibre for a near-ground missile-tracking camera, where high pitch angles caused the ground to disappear.

### Description
- Clamp the pitch to `maxMercatorHorizonAngle` inside `getMercatorHorizon()` in `src/geo/projection/mercator_utils.ts` before computing tangents.
- Update `src/geo/projection/mercator_utils.test.ts` so 90 degree and 95 degree pitch cases expect a horizon of `0`.
- No public API or style-spec changes are included.

### Testing
- `npm run test-unit -- src/geo/projection/mercator_utils.test.ts`
- Downstream manual validation in the web game: high-pitch ground view no longer loses the ground plane once the fix is applied.

### Notes
- This is a small Mercator projection bug fix, so I am opening it directly as a PR rather than starting with a separate issue.

### DIFF
--- a/src/geo/projection/mercator_utils.test.ts
+++ b/src/geo/projection/mercator_utils.test.ts
@@ -36,7 +36,7 @@ describe('mercator utils', () => {
         transform.setPitch(90);
         const horizon = getMercatorHorizon(transform);
 
-        expect(horizon).toBeCloseTo(-9.818037813626313, 10);
+        expect(horizon).toBe(0);
     });
 
     test('getMercatorHorizon95', () => {
@@ -45,7 +45,7 @@ describe('mercator utils', () => {
         transform.setPitch(95);
         const horizon = getMercatorHorizon(transform);
 
-        expect(horizon).toBeCloseTo(-75.52102888757743, 10);
+        expect(horizon).toBe(0);
     });
     describe('getProjectionData', () => {
         test('return identity matrix when not passing overscaledTileID', () => {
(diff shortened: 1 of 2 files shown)


HarelM:
Can you share a before and after videos maybe? I'm keen to see how the game looks 😀

Can you also check if this breaks the functionality of allowing the camera to look above the horizon which was added a while back (with clamp to ground false)? 

Cc: @NathanMOlson

NathanMOlson:
I do not think this change should be made in Maplibre GL JS. For me, allowing pitch angles above 90 is desired behavior. If the camera pitch is far enough above the horizon, the ground _should_ be offscreen. The unit and render tests show the desired behavior.

My suggestion is to clamp the pitch angle outside the library if you don't want to allow pitch angles above 90. If you want the library to do it for you, take a look at [maxPitch](https://maplibre.org/maplibre-gl-js/docs/API/classes/Map/#setmaxpitch) and see if that will serve your needs.

HarelM:
Yeah, I kind of suspected this is what you would say. Makes sense.

HarelM:
No response. closing.
If this is still relevant please open a different PR.

PR #7364, opened 2026-04-01
 GPU performance optimization: early culling of transparent symbols in vertex shaders   

xavierjs:
Hi,

## Issue

The total opacity of symbols (a symbol can be either a character or an icon) is calculated in the vertex. This total opacity is often very low—almost transparent—to achieve the fade-in effect when zooming in. However, even for opacities close to zero, the fragment shader is fully executed for every pixel of the symbol, resulting in significant overdraw.

In this video, we highlight in opaque red the text symbols that would be displayed with an opacity less than 0.1 (i.e., transparent or nearly transparent):

https://github.com/user-attachments/assets/3c0afde5-4a28-411a-a47a-6af8cfd410f8

## Proposed solution

We evaluate the total opacity as early as possible in the vertex shader and exit the shader if the value is too low. This approach offers two main benefits:

* The fragment shader is not executed at all (previously, it was executed at least once per pixel of the rendered symbol, including a texture fetch).
* We also skip the most compute-intensive part of the vertex shader.

## Benchmarks

If we consider this benchmark for example: http://localhost:9966/test/bench/versions/index.html?compare=#LayerSymbol (we render a map with symbol layers).
We need to change the zoom level of the benchmark (otherwise there are no symbols fading-in)
in test/bench/benchmarks/layer.ts, in the setup function we change the zoom level from 16 to 13.15.

**Before:**
Google Chrome, Macbook Pro M4 Max: `2.46ms`
Google Chrome, Macbook Pro M4 Max with swiftshaders (hardware acceleration disabled): `177ms`

**After:**
Google Chrome, Macbook Pro M4 Max: `1.51ms`
Google Chrome, Macbook Pro M4 Max with swiftshaders (hardware acceleration disabled): `68.5ms`

## Misc

No related issues.

## Launch Checklist
 - [X] Link to related issues.
 - [X] Post benchmark scores.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## main
 ### ✨ Features and improvements
 - _...Add new stuff here..._
+-  GPU performance optimization: early culling of transparent symbols in vertex shaders ([#7364](https://github.com/maplibre/maplibre-gl-js/pull/7364)) (by [@xavierjs](https://github.com/xavierjs))
 - Add example showing how to measure map performance using built-in events (`load`, `idle`, `render`) ([#7077](https://github.com/maplibre/maplibre-gl-js/pull/7077)) (by [@CommanderStorm](https://github.com/CommanderStorm))
 
 ### 🐞 Bug fixes


HarelM:
I'm not a shader expert, and this looks OK.
Can you check that this syntax is supported for webgl1 as we're trying to continue supporting it for the time being.

xavierjs:
> Thanks for taking the time to open this PR! I'm not a shader expert, and this looks OK. Can you check that this syntax is supported for webgl1 as we're trying to continue supporting it for the time being.

I double-checked and I am sure this change is compatible with WebGL1 :)

PR #7367, opened 2026-04-01
Optimize matrix inversions and reduce GPU stalls 

xavierjs:
Hi,

## Issue

To compute matrix inversion, we use the general method `mat4.invert` from the `gl-matrix` package: https://github.com/toji/gl-matrix/blob/master/src/mat4.js#L293
This function compute the invert matrix using the general auto-adjoint formulae.

## Proposed solution

In most cases, the matrix to invert has a specific format (transformation matrix, projection matrix, ...), so we can compute the inverse with far less operations.

In this PR, we also avoid some useless allocations, and we add 2 minor optimizations:

* to test if we are running a `WebGL2` context, we just run `gl instanceof WebGL2RenderingContext;`
* we avoid calling `gl.getParameter(gl.MAX_TEXTURE_SIZE);` at each raster tile loading (in general `gl.getParameter` can stall the GPU for some configuration, we should cache its value when possible)

## Benchmarks

The performance improvement is not drastic. There is a 7% gain on the `Placement` benchmark.

## Misc

No related issues.

## Launch Checklist

 - [X] Link to related issues.
 - [X] Post benchmark scores.


src/util/fast_maths.ts
HarelM, line 1: Maybe call this fast_inverse_matrix.ts?

src/util/fast_maths.ts
```diff
@@ -0,0 +1,111 @@
+/*
+(xbourry): this class implements fast mathematical operations
```
HarelM: What's xburry?

src/util/fast_maths.ts
HarelM: Also all the functions in this class should be covered by unit tests.

HarelM:
Are you sure the fast matrix inversion is safe and won't cause issue on some edge cases?
Should there be a check to make sure a matrix is "of the right type" before applying the fast inversion? I'm asking to make sure this was given the right thought not because I think it has to be there.

src/util/fast_maths.ts
HarelM: File name comment needs to be addressed or responded to :-)

src/util/fast_maths.test.ts
HarelM, line 1: Missing test for skew inversion.
I would be a lot more at ease with more than just one test for each inversion, these are at the core of this library and shouldn't have bugs.

xavierjs: I have added a bunch of cases for all the 3 inversion operation (transform, projection, skew), let me know if you would like more test

src/util/fast_maths.test.ts
```diff
@@ -0,0 +1,177 @@
+import {describe, test, expect} from 'vitest';
+import {mat4, vec3, quat} from 'gl-matrix';
+import {fastInvertTransformMat4, fastInvertProjMat4, fastInvertSkewMat4} from './fast_maths';
+
+function compare_matrix(mat: mat4, matRef: mat4){
```
HarelM, line 5: Use `expectToBeCloseToArray` instead.

src/util/fast_maths.test.ts
```diff
@@ -0,0 +1,177 @@
...
+            scaleXY: 50,
+            rotAxis: [0.3, 0.4, 0.5],
+            rotAngle: -0.01
+        },{
+            translate: [42, -7, 13],
+            scaleXY: 4,
+            rotAxis: [1, -2, 3],
+            rotAngle: 2.1
+        }].forEach(({translate, scaleXY, rotAxis, rotAngle}) => {
+            console.log(translate, scaleXY, rotAxis, rotAngle);
```
HarelM: There should not be console.log in the tests.

src/util/fast_maths.test.ts
```diff
@@ -0,0 +1,177 @@
...
+            translate: [0, 0, -9999],
+            scaleXY: 50,
+            rotAxis: [0.3, 0.4, 0.5],
+            rotAngle: -0.01
+        },{
+            translate: [42, -7, 13],
+            scaleXY: 4,
+            rotAxis: [1, -2, 3],
+            rotAngle: 2.1
+        }].forEach(({translate, scaleXY, rotAxis, rotAngle}) => {
```
HarelM: Use regular for loop.

src/symbol/projection.ts
```diff
@@ -930,9 +931,9 @@ export function xyTransformMat4(out: vec4, a: vec4, m: mat4) {
  * Returns a new array of the projected points.
  * Does not modify the input array.
  */
+const inverseLabelPlaneMatrix = mat4.create();
```
HarelM: This should be above the tsdoc comment, but I'm not sure I understand why it is "global", can you explain please?

xavierjs:
> Thanks for taking the time to open this PR! Are you sure the fast matrix inversion is safe and won't cause issue on some edge cases? Should there be a check to make sure a matrix is "of the right type" before applying the fast inversion? I'm asking to make sure this was given the right thought not because I think it has to be there.

I would be difficult to test the matrix type because a full test may be as costly as using the adjoint matrix formulae. Where we use specific matrix inversion there is already strong hypothesis on the matrix format (like we know it is and it should be a projection matrix, or a transform matrix, ...).

HarelM:
Ok, thanks. Please make sure to address the other code review comments.

xavierjs:
> Ok, thanks. Please make sure to address the other code review comments.

I think I implemented feedback or answered to all comments in this PR.

HarelM:
There are still two open comments that were not addressed and no replay, have you forgot to push some code?
In any case, there are conflicts to resolve before this can be merged.

src/symbol/projection.ts
HarelM: This one is still open...

src/symbol/projection.ts
```diff
@@ -17,6 +17,12 @@ import {WritingMode} from '../symbol/shaping';
 import {findLineIntersection} from '../util/util';
 import {type UnwrappedTileID} from '../tile/tile_id';
 import {type StructArray} from '../util/struct_array';
+import {fastInvertSkewMat4} from '../util/fast_maths';
+
+/**
+ * Pre-allocate objects to avoid online allocation
+ */
+const tmpMat4 = mat4.create();
```
HarelM, line 25: Does this really affect performance enough that's "worth" the risk?

HarelM:
Added a last comment, thanks for the improved description.

xavierjs:
I did some mess with a rebase, I am fixing it

HarelM:
I'm concerned about this change, but I don't have any way to reduce my concerns.
I'll publish this and hope that someone will report an issue in case they find one...

HarelM (approved):

Issue #7348, opened 2026-03-29
GeolocateControl ignores fitBoundsOptions.maxZoom when map projection is vertical-perspective


Lievesley:
**maplibre-gl-js version**: 5.21.1

**browser**: Chrome / Safari

### Steps to Trigger Behavior

1. Create a map and set the projection to `vertical-perspective`.
2. Add a `GeolocateControl` with `trackUserLocation: true` and `fitBoundsOptions: { maxZoom: 15 }`.
3. Click the geolocate button when the map is in `vertical-perspective`.

### Link to Demonstration

Paste this into https://jsbin.com.
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
    <title>MapLibre vertical-perspective geolocate repro</title>
    <meta name="viewport" content="width=device-width, initial-scale=1" />
    <link
      rel="stylesheet"
      href="https://unpkg.com/maplibre-gl@5.21.1/dist/maplibre-gl.css"
    />
    <style>
      html, body, #map {
        margin: 0;
        width: 100%;
        height: 100%;
      }

      .panel {
        position: absolute;
        top: 12px;
        left: 12px;
        z-index: 2;
        background: rgba(255, 255, 255, 0.96);
        padding: 10px 12px;
        border-radius: 8px;
        font: 12px/1.4 sans-serif;
        box-shadow: 0 2px 12px rgba(0, 0, 0, 0.18);
        max-width: 360px;
      }

      .mono {
        font-family: ui-monospace, SFMono-Regular, Menlo, monospace;
      }

      button {
        margin-right: 8px;
        margin-top: 6px;
      }
    </style>
  </head>
  <body>
    <div class="panel">
      <div><strong>Repro:</strong> GeolocateControl + vertical-perspective</div>
      <div>Projection: <span id="projection" class="mono">vertical-perspective</span></div>
      <div>Zoom: <span id="zoom" class="mono">7.00</span></div>
      <div>Geolocate max zoom: <span class="mono">14</span></div>
      <div style="margin-top:8px">
        1. Open this in full JSBin output mode over HTTPS.<br />
        2. Allow location access for the page.<br />
        3. Leave projection on <span class="mono">vertical-perspective</span>.<br />
        4. Click the native geolocate button.<br />
        5. Compare with <span class="mono">mercator</span>.
      </div>
      <div style="margin-top:8px">
        Expected: geolocate should not zoom above 14.<br />
[...]

PR #7372, opened 2026-04-02
feat: add maxZoom for vertical perspective

CommanderStorm:
Resolves https://github.com/maplibre/maplibre-gl-js/issues/7348


src/geo/projection/vertical_perspective_camera_helper.ts
```diff
@@ -202,7 +202,7 @@ export class VerticalPerspectiveCameraHelper implements ICameraHelper {
         }
 
         // Compute target zoom from the obtained scale.
-        result.zoom = clonedTr.zoom + scaleZoom(smallestNeededScale);
+        result.zoom = Math.min(clonedTr.zoom + scaleZoom(smallestNeededScale), options.maxZoom);
```
CommanderStorm, line 205: we do the same thing in the camera helper

HarelM:
Thanks for fixing this! I'll merge this, but just to be on the safe side it's worth making sure the tests are failing without the fix.

PR #7378, opened 2026-04-02
fix: prevent stale async style loads from completing after style clear

### LINKED ISSUE #7136: throwIfAborted not supported in older browsers
**maplibre-gl-js version**: 5.18.0

**browser**: Just tested on Google Chrome 87. Caniuse.com states it's not supported before Google Chrome 100

### Steps to Trigger Behavior

 1. Visit any of the examples on [maplibre.org](https://maplibre.org/maplibre-gl-js/docs/examples/add-a-canvas-source/) in a browser older than Google Chrome 100
 2. Open console, see an error like `.signal.throwIfAborted is not a function (ajax.ts:186)`

Seems to originate from here: https://github.com/maplibre/maplibre-gl-js/blob/5102ccbb85bf7c936b890c8a640a0836dfc1c880/src/util/ajax.ts#L186

### Expected Behavior
Support for older browsers. As of today, [caniuse.com](https://caniuse.com/?search=throwIfAborted) reports that around 5% of browsers do not support `throwIfAborted`

### Actual Behavior
Not compatible with older browsers

### LINKED ISSUE #7377: Stale setStyle(url) load can apply after setStyle(null) or style replacement
maplibre-gl-js version: 5.21.1, also reproducible on main at commit ae1cd7b76e085d8ab33c28a63dadc498e431cf1f (2026-04-02)

browser: Safari 18.5 on macOS, also reproducible on iPhone Safari

Steps to Trigger Behavior

Create a map instance.
Configure `transformRequest` so that a style URL request resolves asynchronously rather than immediately.
Call `map.setStyle('style.json', {diff: false})`.
Before the async `transformRequest` resolves, clear or replace the style, for example with `map.setStyle(null, {diff: false})`.
Then let the original `transformRequest` resolve and allow the original style request to complete.

Link to Demonstration

I do not have a public minimal repro URL yet, but the issue reproduces locally with an async `transformRequest` and a style change that clears the style before the original style URL load completes.

Relevant behavior:

A stale `setStyle(url)` load can continue into `Style#_load(...)` even after that style has already been cleared or replaced.

Problematic code path:

// style.ts
const request = await this.map._requestManager.transformRequest(url, ResourceType.Style);
const response = await getJSON<StyleSpecification>(request, this._loadStyleRequest);
this._loadStyleRequest = null;
this._load(response.data, options, previousStyle);

The issue is that `transformRequest(...)` is async. If the style is cleared or replaced while that await is still pending, the old load can resume afterward and still call `_load(...)` unless the in-flight request is verified to still be current.

Expected Behavior

If a style load has been cleared or replaced before its async request pipeline completes, the stale request should be ignored and must not continue loading style state.

Actual Behavior

A stale async style load can complete after the style was cleared or replaced, causing obsolete style initialization to run after that load should already have been discarded.

Root Cause

`loadURL()` stores an abort controller for the current style load, but an older async load can resume after `transformRequest(...)` resolves and continue unless it checks whether it is still the active request before proceeding.

Notes

This is easiest to reproduce when `transformRequest` is async, because that creates a window where `setStyle(url)` has started but has not yet reached the actual fetch/load step.
The fix is to keep the abort handle associated with the specific load attempt and ignore completion from stale requests once that attempt has been superseded.

Lievesley:
## Launch Checklist

This PR fixes a race in async style loading. If setStyle(url) was still waiting on an async transformRequest and the style was cleared or replaced in the meantime, the stale request could still complete and apply its old style.

The change keeps the abort controller tied to the specific load attempt, checks that the request is still current before continuing, and avoids clearing a newer in-flight request by mistake. It also adds a regression test covering setStyle(null) during async style resolution and updates the changelog text to describe the actual failure mode.

- Fixes #7136
- Fixes #7377
 
### DIFF
--- a/src/ui/map_tests/map_style.test.ts
+++ b/src/ui/map_tests/map_style.test.ts
@@ -357,10 +357,8 @@ describe('setStyle', () => {
         const map = createMap({deleteStyle: true});
         const initialTransform = map.transform;
         const initialPainterTransform = map.painter.transform;
-        let projectionTransitionFired = false;
-        map.on('projectiontransition', () => {
-            projectionTransitionFired = true;
-        });
+        const projectionTransitionSpy = vi.fn();
+        map.on('projectiontransition', projectionTransitionSpy);
         map.setTransformRequest(() => transformRequest);
 
         map.setStyle('style.json', {diff: false});
@@ -372,7 +370,7 @@ describe('setStyle', () => {
         await sleep(0);
 
         expect(map.style).toBeUndefined();
-        expect(projectionTransitionFired).toBe(false);
+        expect(projectionTransitionSpy).not.toHaveBeenCalled();
         expect(map.transform).toBe(initialTransform);
         expect(map.painter.transform).toBe(initialPainterTransform);
     });


src/style/style.ts
```diff
@@ -1686,7 +1686,7 @@ export class Style extends Evented {
     setProjection(projection: ProjectionSpecification) {
         this._checkLoaded();
         const resolvedProjection = projection ?? {type: 'mercator'};
```
CommanderStorm, line 1698: I don't think `projection` can be nullish according to the type definition, so this cannot actually happen, or?

So what are you passing as a parameter here?

HarelM:
This might be related to a recent fix that was merged: 
- https://github.com/maplibre/maplibre-gl-js/pull/7330#discussion_r3002969580

Note that in the linked comment I explained why I think that this is not the right fix...

src/style/style.ts
Lievesley: I updated the fix and description so it addresses the underlying cause better. I added one test that only succeeds with changes in this PR. Please take another look @CommanderStorm.

src/style/style.ts
```diff
@@ -426,15 +426,24 @@ export class Style extends Evented {
...
             options.validate : true;
 
-        const request = await this.map._requestManager.transformRequest(url, ResourceType.Style);
         this._loadStyleRequest = new AbortController();
         const abortController = this._loadStyleRequest;
         try {
-            const response = await getJSON<StyleSpecification>(request, this._loadStyleRequest);
-            this._loadStyleRequest = null;
+            const request = await this.map._requestManager.transformRequest(url, ResourceType.Style);
+            abortController.signal.throwIfAborted();
```
HarelM: I think someone mentioned that this is a new method that's not suppoted well in all browsers?

src/style/style.ts
```diff
@@ -426,15 +426,24 @@ export class Style extends Evented {
...
+
+            const response = await getJSON<StyleSpecification>(request, abortController);
+            // Clear this request only if it is still the active style load. A stale
```
HarelM: Nit pickng: This comments should be inside the if pleas

src/style/style.ts
HarelM: One can pass undefined to revert to default projection. It might be worth updating the method signature to indicate this better.

src/ui/map_tests/map_style.test.ts
```diff
@@ -346,6 +346,30 @@ describe('setStyle', () => {
...
+    test('stale style URL load does not complete after style is cleared during async transformRequest', async () => {
+        server.respondWith('style.json', JSON.stringify(createStyle()));
+
+        let resolveTransformRequest: (value: {url: string}) => void;
+        const transformRequest = new Promise<{url: string}>((resolve) => {
+            resolveTransformRequest = resolve;
+        });
+
+        const map = createMap({deleteStyle: true});
+        const migrateProjectionSpy = vi.spyOn(map, 'migrateProjection');
```
HarelM: Is there a better way to test this instead of spying on this method?

CHANGELOG.md
```diff
@@ -113,7 +117,7 @@
 
 ### ✨ Features and improvements
 
-- Refactor _updateWorkerData ([#6983](https://github.com/maplibre/maplibre-gl-js/pull/6983)) (by [@wayofthefuture](https://github.com/wayofthefuture))
+- Refactor \_updateWorkerData ([#6983](https://github.com/maplibre/maplibre-gl-js/pull/6983)) (by [@wayofthefuture](https://github.com/wayofthefuture))
```
HarelM: Maybe use "`" instead?

HarelM:
Thanks for all the changelog fixes, it was automatic, right?

src/style/style.ts
neodescis: There's an open issue on it, actually:  https://github.com/maplibre/maplibre-gl-js/issues/7136

Lievesley:
@HarelM  yes changelog was formatted automatically, I only added a description for the bug manually.

src/style/style.ts
Lievesley: I added a backward compatible helper function.
abortController.signal.throwIfAborted() was still used in ajax.ts so I updated that too.

src/style/style.ts
Lievesley: Updated

src/ui/map_tests/map_style.test.ts
Lievesley: I updated it to check the external interface only

CHANGELOG.md
Lievesley: Updated

src/style/style.ts
Lievesley: Updated projection to be optional

src/ui/map_tests/map_style.test.ts
```diff
@@ -346,6 +346,37 @@ describe('setStyle', () => {
...
+
+        let resolveTransformRequest: (value: {url: string}) => void;
+        const transformRequest = new Promise<{url: string}>((resolve) => {
+            resolveTransformRequest = resolve;
+        });
+
+        const map = createMap({deleteStyle: true});
+        const initialTransform = map.transform;
+        const initialPainterTransform = map.painter.transform;
+        let projectionTransitionFired = false;
```
HarelM: Use `vi.fn()` and `expect(...).not.toHaveBeenCalled()`.

src/style/style.ts
HarelM: THANKS!

Issue #5048, opened 2024-11-15
Symbol for polygon not placed at center.


exotfboy:
Hi:

When adding symbols to polygons with `symbol-placement` set to `point`, it is expected that the symbol will be placed at the center of the polygon.

However, in some situations, this is not the case. [Live Example](https://codesandbox.io/p/sandbox/maplibre-polygon-label-s4267p).

As shown in the example, the symbol for `A` is correctly centered, while the symbol for `B` is not.

What's going on?

HarelM:
I would consider this as an issue with the limitation of accuracy for geojson source.
Geojson source only build tiles to zoom 15 or something similar, so the accuracy at zoom 21 is very low, which I would assume what's causing this calculation issues.
But you are more than welcome to check my theory.
You can also try out making vector tiles with higher resolution to see if it helps.

exotfboy:
1 I will have a try if possible.

2 Any fallback instead of vector tiles at the moment?

HarelM:
Maybe `addProtocol` where you could return the relevant tiles you create "on the fly", just maybe...

exotfboy:
@HarelM 

Since I do not have the environment to create vector tiles from the server side. So I tried to generate vector tiles on the fly as you said, it does not change anything. Have a look at [this](https://codesandbox.io/p/sandbox/maplibre-polygon-label-vt-onfly-gxvm78).

HarelM:
Thanks! But this still users geijson-vt which I think is where the big is. 
Try to use vt-pbf or something that can serialize a geojson object to a vector tile.
No sure it will help but world be interesting to see if it changes anything...

exotfboy:
Indeed the `vt-pbf` is used:

```
  const tile = tileIndex.getTile(z, x, y); //use geojson-vt create vector tile with json format
  if (tile == null) return { data: null };
  const buff = vtpbf.fromGeojsonVt({ json: tile }); //use vt-pbf to generate pbf format.
```

HarelM:
I would advise to try and create a tile using a different method and then try to read it and see if you still have a problem.

exotfboy:
Finally I tried to upload my data to mapbox studio to generate vector tiles. And things does [not change](https://codepen.io/exotfboy/pen/ZEgZvxM).

HarelM:
Ok, thanks for taking the time to try this out, this rules that the problem is with the geojson conversation to MVT, and leaves text location precision I guess...

[...]

PR #7380, opened 2026-04-02
fix: high-zoom symbol placement inside polygon inaccuracy

CommanderStorm:

Two commits to prove that this actually fixes the issue

Resolves https://github.com/maplibre/maplibre-gl-js/issues/5048


src/util/find_pole_of_inaccessibility.ts
HarelM, line 1: Can we remove the debug flag from here? I don't see when and how one would use it, unit tests and render tests are the proper way to debug this I believe.

HarelM:
Can you also check the coverage report? Maybe a unit test is needed here...?

src/util/find_pole_of_inaccessibility.test.ts
```diff
@@ -1,22 +1,70 @@
-import {test, expect} from 'vitest';
+import {describe, test, expect} from 'vitest';
 import Point from '@mapbox/point-geometry';
-import {findPoleOfInaccessibility} from './find_pole_of_inaccessibility';
+import {findPoleOfInaccessibility, getCentroidCell} from './find_pole_of_inaccessibility';
```
HarelM, line 3: If not mandatory, I would avoid exporting a method just for tests.
It should be covered by the existing tests, shouldn't it?

CommanderStorm: It is much harder to read the test if the differentiation centroid vs pole of inaccessibility is not clear by this import.

It kind of is covered, but the assertion could be stronger by this export.
Lests keep it, it does not have impact on consumers, only makes the tests better

HarelM:
Can you move the changelog entry up? Sorry for the extra work...
Looks good otherwise.

CHANGELOG.md
CommanderStorm, line 7:
```suggestion
- Fix polygon text label placement drifting far from center for convex polygons at high zoom due to coordinate rounding in geojson-vt ([#7380](https://github.com/maplibre/maplibre-gl-js/pull/7380)) (by [@CommanderStorm](https://github.com/CommanderStorm))
- _...Add new stuff here..._
```

CHANGELOG.md
```diff
@@ -15,7 +15,7 @@
 - UX: Clarify error message language so if layout and paint properties are confused in `setPaintProperty` or `setLayoutProperty` ([#6954](https://github.com/maplibre/maplibre-gl-js/pull/6954)) (by [@Willjfield](https://github.com/Willjfield) and [@CommanderStorm](https://github.com/CommanderStorm))
 
 ### 🐞 Bug fixes
-
+- Fix polygon text label placement drifting far from center for convex polygons at high zoom due to coordinate rounding in geojson-vt ([#7380](https://github.com/maplibre/maplibre-gl-js/pull/7380)) (by [@CommanderStorm](https://github.com/CommanderStorm))
 - Fix startup crash caused by a stale async style load completing after the style was cleared or replaced ([#7377](https://github.com/maplibre/maplibre-gl-js/issues/7377))
```
```suggestion

- Fix startup crash caused by a stale async style load completing after the style was cleared or replaced ([#7377](https://github.com/maplibre/maplibre-gl-js/issues/7377))
```

CHANGELOG.md
```diff
@@ -25,6 +26,7 @@
 - Fix `fitBounds` ignoring `maxZoom` option in `vertical-perspective` projection ([#7372](https://github.com/maplibre/maplibre-gl-js/issues/7372)) (by [@CommanderStorm](https://github.com/CommanderStorm))
 - Prevent stale async style loads from completing after style clear ([#7378](https://github.com/maplibre/maplibre-gl-js/pull/7378)) (by [@Lievesley](https://github.com/Lievesley))
 - Fix broken example for `fill-pattern` ([#7326](https://github.com/maplibre/maplibre-gl-js/pull/7326)) (by [@k-yle](https://github.com/k-yle))
+- _...Add new stuff here..._
```
```suggestion
```

PR #7392, opened 2026-04-03
chore: add @typescript-eslint/prefer-optional-chain

CommanderStorm:
I think this lint makes the code better, but if we don't like the look of this we can also go the other way.
Currently we are a bit inconsistent in this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,11 @@
 ## main
+### ✨ Features and improvements
+- _...Add new stuff here..._
+
+### 🐞 Bug fixes
+- _...Add new stuff here..._
+
+## 5.22.0
 
 ### ✨ Features and improvements
 
@@ -6,7 +13,6 @@
 - GPU performance optimization: early culling of transparent symbols in vertex shaders ([#7364](https://github.com/maplibre/maplibre-gl-js/pull/7364)) (by [@xavierjs](https://github.com/xavierjs))
 - Add example showing how to measure map performance using built-in events (`load`, `idle`, `render`) ([#7077](https://github.com/maplibre/maplibre-gl-js/pull/7077)) (by [@CommanderStorm](https://github.com/CommanderStorm))
 - UX: Clarify error message language so if layout and paint properties are confused in `setPaintProperty` or `setLayoutProperty` ([#6954](https://github.com/maplibre/maplibre-gl-js/pull/6954)) (by [@Willjfield](https://github.com/Willjfield) and [@CommanderStorm](https://github.com/CommanderStorm))
-- _...Add new stuff here..._
 
 ### 🐞 Bug fixes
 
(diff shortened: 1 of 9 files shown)


HarelM (dismissed):
Definitely better, thanks!
I think there's a need to update to eslint 10, so you might want to do that first...

HarelM:
Why was bundle size affected by this change?

CommanderStorm:
> Why was bundle size affected by this change?

we are on ES2016, so the transpiled output is apparently sligtly larger on this op.

I would advise we bump this in v6

test/build/bundle_size.json
```diff
@@ -1 +1 @@
-1049586
+1053145
```
CommanderStorm, line 1:
```diff
diff --git i/test/build/bundle_size.json w/test/build/bundle_size.json
index 135d71ea3..bcc8eb2bd 100644
--- i/test/build/bundle_size.json
+++ w/test/build/bundle_size.json
@@ -1 +1 @@
-1053145
+1040152
diff --git i/tsconfig.json w/tsconfig.json
index ce088ad70..b812b8836 100644
--- i/tsconfig.json
+++ w/tsconfig.json
@@ -14,7 +14,7 @@
     "resolveJsonModule": true,
     "skipLibCheck": false,
     "strict": false,
-    "target": "ES2016",
+    "target": "ES2020",
     "lib": [
       "ESNext",
       "DOM",
```

HarelM:
Larger how? Doesn't it get transpiled to exactly what it used to be?

CommanderStorm:
> If the object accessed or function called using this operator is [undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined) or [null](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null), the expression short circuits and evaluates to [undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined) instead of throwing an error.

So emulating this using "is falsy" only is a bit more code.
Think of void functions, functions and emty arrays.

HarelM:
Sorry for taking so long to reply, but I still don't understand why this is 3-4k larger. What is the javascript equivalent that is more verbose for all those cases?

CommanderStorm:
> but I still don't understand why this is 3-4k larger

there are 178 `?.` here.
So 3559 / 178 = 19.9 chars/bytes on avg.

When I look at typescript, this matches what I expect.
Lets remember that we still ship typescript targeting `ES2016`

Example:

```diff
{
  "compilerOptions": {
-   "target": "ES2016",
+   "target": "ES2020",
    "module": "ESNext",
    "strict": true,
    "outDir": "dist"
  }
}
```
on

```ts
export function test(foo: { bar: number } | null | undefined) {
  return foo?.bar;
}
```

```bash
npx tsc
```
```diff
export function test(foo) {
-     return foo === null || foo === void 0 ? void 0 : foo.bar;
+     return foo?.bar;
}
```

after mnification, that LGTM

HarelM:
I would argue that style have been using the long notation, but I guess the void 0 verbosity explains it. Thanks!

HarelM:
You've opened a bunch of PRs, let me know if there's an order to review them.

PR #7397, opened 2026-04-03
chore: resolve a few lints around return types

CommanderStorm:
Needs https://github.com/maplibre/maplibre-gl-js/pull/7392 merged first

### DIFF
--- a/src/ui/popup.test.ts
+++ b/src/ui/popup.test.ts
@@ -395,7 +395,7 @@ describe('popup', () => {
         ['bottom-left',  new Point(10, containerHeight - 10),                   'translate(0,-100%) translate(7px,-7px)'],
         ['left',         new Point(10, containerHeight / 2),                    'translate(0,-50%) translate(10px,0px)'],
         ['bottom',       new Point(containerWidth / 2, containerHeight / 2),    'translate(-50%,-100%) translate(0px,-10px)']
-    ] as const satisfies readonly [PositionAnchor, Point, string][];
+    ] as const satisfies ReadonlyArray<[PositionAnchor, Point, string]>;
 
     const anchorCases = cases.map(([anchor, point]) => [anchor, point] as const);
     test.each(anchorCases)(


src/ui/map.ts
```diff
@@ -934,7 +934,11 @@ export class Map extends Camera {
      * map.removeControl(navigation);
      * ```
      */
+<<<<<<< better-return-types
+    removeControl(control: IControl): this {
+=======
     removeControl(control: IControl): Map {
+>>>>>>> main
```
```suggestion
    removeControl(control: IControl): this {
```

src/render/draw_debug.ts
```diff
@@ -141,9 +141,9 @@ export function selectDebugSource(style: Style, zoom: number): TileManager | nul
             selectedSource = source;
         }
     };
-    vectorSources.forEach((source) => considerSource(source));
+    vectorSources.forEach(considerSource);
```
HarelM: Can we add a lint rule to avoid forEach?

src/render/terrain.test.ts
```diff
@@ -399,7 +399,7 @@ describe('Terrain', () => {
         terrain.getFramebuffer('coords');
         terrain.getTerrainMesh(new OverscaledTileID(0, 0, 0, 0, 0));
 
-        expect(() => terrain.destroy()).not.toThrow();
+        expect(() => { terrain.destroy(); }).not.toThrow();
```
HarelM: Not sure this extra characters are actually needed... I'm not sure what I feel about this rule...

CommanderStorm: This fn (`destroy`) does returns void.

So this way it is clearer that this actually does not have a return type.
-> no returing `void` returns

HarelM: I understand, but I'm not sure it's actually clearer this way. 
If the rule was that you need return for every lambda it would've been more clear, but just addind {} to avoid not returning for a method that returns void doesn't clarifies it a lot...

src/render/uniform_binding.test.ts
```diff
@@ -20,7 +20,7 @@ describe('Uniform Binding', () => {
 
         const context = {
             gl: {
-                uniform1i: () => { expect(true).toBeTruthy(); }
+                uniform1i: () => expect(true).toBeTruthy()
```
CommanderStorm, line 23: yea, sure..

Unrelated to this PR though

src/render/draw_debug.ts
CommanderStorm: will be more code to review, but sure

src/render/terrain.test.ts
CommanderStorm: Ok, I changed it the other way around

src/source/geojson_worker_source.ts
```diff
@@ -298,9 +298,8 @@ export class GeoJSONWorkerSource implements WorkerSource {
         if (compiled.result === 'error') {
             throw new Error(compiled.value.map(err => `${err.key}: ${err.message}`).join(', '));
         }
-        
-        const predicate = (feature: GeoJSON.Feature) => compiled.value.evaluate({zoom: 0}, feature as any);
-        return predicate;
+
+        return (feature: GeoJSON.Feature) => compiled.value.evaluate({zoom: 0}, feature as any);
```
HarelM, line 302: This is less readable, is this a lint replacement or manual one?

CommanderStorm: manual, but I also think that `_getFilterPredicate` should not need this value there.

src/tile/tile_manager.test.ts
```diff
@@ -1,16 +1,16 @@
 import type {StyleSpecification} from '@maplibre/maplibre-gl-style-spec';
-import {describe, beforeEach, afterEach, test, expect, vi} from 'vitest';
+import {afterEach, beforeEach, describe, expect, test, vi} from 'vitest';
 import {TileManager} from './tile_manager';
-import {type Source, addSourceType} from '../source/source';
-import {Tile, FadingRoles, FadingDirections} from './tile';
+import {addSourceType, type Source} from '../source/source';
```
HarelM, line 4: Would be great to have a rule for import ordering too I guess...

src/ui/map_tests/map_disable_handlers.test.ts
```diff
@@ -27,13 +27,10 @@ const handlerNames = [
     'doubleClickZoom',
     'touchZoomRotate'
 ];
-handlerNames.forEach((handlerName) => {
-    test(`disable "${handlerName}" handler`, () => {
-        const options = {};
-        options[handlerName] = false;
-        const map = createMap(options);
+test.each(handlerNames)('disable "%s" handler', (handlerName) => {
```
HarelM, line 30: regular for would be better here as well...

CommanderStorm: regular?

src/tile/tile_manager.test.ts
CommanderStorm: formatting != linting, but yes we should move to something better for formatting. not this PR

src/util/web_worker_transfer.ts
```diff
@@ -103,10 +103,8 @@ function isRegistered(input: unknown): boolean {
         return false;
     }
     const classRegistryKey = getClassRegistryKey(input);
-    if (classRegistryKey && classRegistryKey !== 'Object') {
-        return true;
-    }
-    return false;
+    return classRegistryKey && classRegistryKey !== 'Object';
```
HarelM, line 106: That's not the same I believe. Returning boolean by checking on an object can retrun a non-null response which is not a boolean.

src/source/geojson_worker_source.ts
HarelM: I don't have strong feelings about this TBH...

src/ui/map_tests/map_disable_handlers.test.ts
HarelM: for..of instead of test.each.

HarelM:
All these rules will make it hard to transition to different tooling, but I don't know of a good alternative, so I'm fine with this direction...

src/util/web_worker_transfer.ts
CommanderStorm: We check here if the string `classRegistryKey` is neither `''` (aka falsy) or `Object`.

There is no js-object here.

src/util/web_worker_transfer.ts
```diff
@@ -103,10 +103,8 @@ function isRegistered(input: unknown): boolean {
...
+
```
CommanderStorm:
```suggestion
    return classRegistryKey && classRegistryKey !== 'Object';
```

src/ui/popup.test.ts
```diff
@@ -395,12 +395,12 @@ describe('popup', () => {
         ['bottom-left',  new Point(10, containerHeight - 10),                   'translate(0,-100%) translate(7px,-7px)'],
         ['left',         new Point(10, containerHeight / 2),                    'translate(0,-50%) translate(10px,0px)'],
         ['bottom',       new Point(containerWidth / 2, containerHeight / 2),    'translate(-50%,-100%) translate(0px,-10px)']
-    ] as Array<[PositionAnchor, Point, string]>).forEach((args) => {
-        const anchor = args[0];
-        const point = args[1];
-        const transform = args[2];
+    ] as const satisfies readonly [PositionAnchor, Point, string][];
```
CommanderStorm:
```suggestion
    ] as const satisfies ReadonlyArray<[PositionAnchor, Point, string]>;
```

src/util/web_worker_transfer.ts
HarelM: This returns classRegistryKey in case it's not an empty string I think. 
If you add a return value to the method it will highlight the error, I think. 
I've seen bugs on this type of return, so I'm sensitive to it...

PR #7403, opened 2026-04-04
chore: simplify types `Array<T>` -> `T[]` for simple types

CommanderStorm:
currrently, we have both `Array` and `[]` -> lets only have the latter.


CommanderStorm:
> I hope this is automated

mostly, except the changes to how we do codegen and the decision to keep `Array<T>` for "complex" types, because the extra verbosity is there better.
Think
- `[number, string[]][]` being less clear than `Array<[number, string[]]>`.
- `("foo"|"bar")[]` being less clear than `Array<"foo"|"bar">`.

PR #7404, opened 2026-04-04
chore!: bump tsconfig target to `ES2022`

CommanderStorm:
`ES2022` was the version of JS that browsers reliably supported in 2022.

We are currently on 2016, but are using 2020 features (-> if users bundle us, this is relevant), so realistically, this is not actually a bump, but maybe this might.

This change aligns us with the wider ecosystem, but if someone has not updated their browser in 5 years, but wants to update this would be problematic -> a breakign change.

This PR will fail the bundle size check, but since we are not going to merge it now, that is intentional. this part will need to change once we are ready to pivot towards v6


HarelM:
@CommanderStorm can you make this PR pass CI so we can push it forward?
Also would you prefer to set it to 2020 or straight to 2022?

CommanderStorm:
> Also would you prefer to set it to 2020 or straight to 2022

I would prefer 2022, because this will enable lazy loading modules.
This is interesting because in our init process some things are less important than others, so for example the svg work of oliver would be great to have initalised lazily.
Maybe also others, not sure.
We will have to benchmark that.

HarelM:
Sure, no problem. Can you prepare this PR to be merged (Including changelog update)?
Thanks!

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## main
 ### ✨ Features and improvements
+
+- ⚠️ The TypeScript target has been updated to ES2022.
+  This results in smaller bundles and improved  runtime performance by relying on modern JavaScript features and reducing transpilation. Consumers targeting browsers or using some tooling released before 2022 may need to transpile MapLibre or update. This change also aligns all internal build configurations to a single target instead of ES2016 + ES2019, avoiding inconsistencies in emitted code. ([#7404](https://github.com/maplibre/maplibre-gl-js/pull/7404)) (by [@CommanderStorm](https://github.com/CommanderStorm))
```
```suggestion
  This results in smaller bundles and improved runtime performance by relying on modern JavaScript features and reducing transpilation. Consumers targeting browsers or using some tooling released before 2022 may need to transpile MapLibre or update. This change also aligns all internal build configurations to a single target instead of ES2016 + ES2019, avoiding inconsistencies in emitted code. ([#7404](https://github.com/maplibre/maplibre-gl-js/pull/7404)) (by [@CommanderStorm](https://github.com/CommanderStorm))
```

HarelM (approved):
With the addition of ESM and removal of commonjs it should be possible to transpile this code to older version if needed I assume.

HarelM (approved):

PR #7409, opened 2026-04-05
chore: add logical-assignment-operators

CommanderStorm:
simplifies the code, but won't make it shorter until #7404 

logical assigment is es2021


src/data/bucket/symbol_bucket.ts
```diff
@@ -478,7 +478,7 @@ export class SymbolBucket implements Bucket {
                 const formattedText = Formatted.factory(resolvedTokens);
 
                 // on this instance: if hasRTLText is already true, all future calls to containsRTLText can be skipped.
-                const bucketHasRTLText = this.hasRTLText = (this.hasRTLText || containsRTLText(formattedText));
+                const bucketHasRTLText = (this.hasRTLText ||= containsRTLText(formattedText));
```
HarelM: I don't like these single line multiple variable assignment to be honest.
Better to split this into 2 lines...

src/data/bucket/symbol_bucket.ts
```diff
@@ -533,7 +533,7 @@ export class SymbolBucket implements Bucket {
                     if (!section.image) {
                         const doesAllowVerticalWritingMode = allowsVerticalWritingMode(text.toString());
                         const sectionFont = section.fontStack || fontStack;
-                        const sectionStack = stacks[sectionFont] = stacks[sectionFont] || {};
+                        const sectionStack = (stacks[sectionFont] ||= {});
```
HarelM: Same here...

src/render/program.ts
```diff
@@ -234,8 +234,8 @@ export class Program<Us extends UniformBindings> {
         }
 
         for (const segment of segments.get()) {
-            const vaos = segment.vaos || (segment.vaos = {});
-            const vao: VertexArrayObject = vaos[layerID] || (vaos[layerID] = new VertexArrayObject());
+            const vaos = (segment.vaos ||= {});
+            const vao: VertexArrayObject = (vaos[layerID] ||= new VertexArrayObject());
```
HarelM: Same here...

src/symbol/cross_tile_symbol_index.ts
```diff
@@ -332,9 +330,8 @@ export class CrossTileSymbolIndex {
             if (styleLayer.id !== symbolBucket?.layerIds[0])
                 continue;
 
-            if (!symbolBucket.bucketInstanceId) {
-                symbolBucket.bucketInstanceId = ++this.maxBucketInstanceId;
-            }
+            ++this.maxBucketInstanceId;
```
HarelM: This is not exactly the same logic, is it?

src/tile/tile.ts
```diff
@@ -297,17 +297,13 @@ export class Tile {
             this.imageAtlasTexture.destroy();
         }
 
-        if (this.imageAtlas) {
-            this.imageAtlas = null;
-        }
+        this.imageAtlas &&= null;
```
HarelM: Just = null without the if or &&?

src/tile/tile.ts
```diff
@@ -297,17 +297,13 @@ export class Tile {
...
 
         if (this.glyphAtlasTexture) {
             this.glyphAtlasTexture.destroy();
         }
 
-        if (this.dashPositions) {
-            this.dashPositions = null;
-        }
+        this.dashPositions &&= null;
```
HarelM: Same as above

src/ui/handler/scroll_zoom.ts
```diff
@@ -236,9 +236,7 @@ export class ScrollZoomHandler implements Handler {
     _start(e: MouseEvent) {
         if (!this._delta) return;
 
-        if (this._frameId) {
-            this._frameId = null;
-        }
+        this._frameId &&= null;
```
HarelM, line 239: Same as above

HarelM: Maybe better to set it to false as this is a boolean, or even change its name to indicate it's not holding an id.

HarelM:
Thanks! Done my review, looks good overall, added a few comments.

src/symbol/cross_tile_symbol_index.ts
```diff
@@ -332,9 +330,8 @@ export class CrossTileSymbolIndex {
...
+            symbolBucket.bucketInstanceId ||= this.maxBucketInstanceId;
```
CommanderStorm:
```suggestion
             if (!symbolBucket.bucketInstanceId) {
                 this.maxBucketInstanceId += 1;
                 symbolBucket.bucketInstanceId = this.maxBucketInstanceId;
             }
```

src/symbol/cross_tile_symbol_index.ts
CommanderStorm: ups

src/symbol/cross_tile_symbol_index.ts
```diff
@@ -332,9 +330,10 @@ export class CrossTileSymbolIndex {
             if (styleLayer.id !== symbolBucket?.layerIds[0])
                 continue;
 
-            if (!symbolBucket.bucketInstanceId) {
-                symbolBucket.bucketInstanceId = ++this.maxBucketInstanceId;
-            }
+             if (!symbolBucket.bucketInstanceId) {
+                 this.maxBucketInstanceId += 1;
+                 symbolBucket.bucketInstanceId = this.maxBucketInstanceId;
+             }
```
CommanderStorm:
```suggestion
            if (!symbolBucket.bucketInstanceId) {
                this.maxBucketInstanceId += 1;
                symbolBucket.bucketInstanceId = this.maxBucketInstanceId;
            }
```

CommanderStorm:
it is very hard to do lints against a moving target. can we +1 instead of bike shedding on the small stuff please?

src/tile/tile.ts
CommanderStorm: this is not a code change

src/tile/tile.ts
CommanderStorm: this is not a code change

src/ui/handler/scroll_zoom.ts
CommanderStorm: this is not a code change. Lets stick to the previous logic.

src/tile/tile.ts
HarelM: I don't understand this statement.
Why use `&&=` when it's not needed?
Let's do change the code if it's weird, like it is now...

HarelM:
Can you fix the `&&= null` with a simple `= null` please?
I don't want to introduce something that's cumbersome to read.
If you don't want to change the logic (which I understand, but don't agree), I would prefer to keep the "if", it's more readable.

src/tile/tile.ts
CommanderStorm: I did not change the code on this one and don't want to investigate this further.
This is not a code change, it is exactly the same logic as before.

CommanderStorm:
@HarelM
This is very nitpicky. Either close the PR or do changes yourself if you want to nitpic at this level.

This change does overall improve the codebase.
I won't investigate the three cases where we might be able to simplify the code further.
That is not a code change.

HarelM (approved):

PR #7412, opened 2026-04-05
Feat LayerRegistry

FalkF:
These are a couple classes I use to handle turning on and off multiple complex features on the map.
This is intended for big projects that have things dynamically turning on and off, user settings for features etc.
It correctly inserts all layers belonging to a `FeatureLayer` (I'm terrible with names), loads needed sources and manages the whole lifecycle of the layer. Including the possibility of unloading before and reloading (everything that should be visible) after map style changes. 

This is not 100% ready, to be merged, I encourage you to check out the branch and play around with the concept a little.
Just create a LayerRegistry, add a MapResource and Feature Layer, you now can use the Registry to toggle the layer on and off, it will be inserted correctly according to the order `addLayer` is called. 

Ideas highly welcome!

## Launch Checklist

Not sure if mapbox has something like this, I dont know mapbox.

 - [ ] Post benchmark scores.

### DIFF
--- a/src/layer-registry/LayerRegistry.ts
+++ b/src/layer-registry/LayerRegistry.ts
@@ -4,6 +4,11 @@ import type {Map} from '../ui/map';
 
 type FeatureLayerSettings = { [layerName: string]: boolean };
 
+/**
+ * LayerRegistry manages the lifecycle of feature layers and their associated map resources.
+ * It ensures that layers are loaded and unloaded in the correct order, and that sources are available before layers are loaded.
+ * This helps manage the lifecycle of multiple complex features and their associated map resources.
+ */
 export class LayerRegistry<T extends FeatureLayerSettings> {
     private sources: MapResource[] = [];
     private featureLayers: FeatureLayer[] = [];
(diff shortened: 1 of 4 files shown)


HarelM:
Can this be made to a plugin then? I don't see that it uses any internal APIs...

FalkF:
Absolutely could be a plugin, I hoped it might be small, worry free and usefull enough to maybe make it into main. With some refinement ofc.
Having a way to toggle things on and off reliably without having to write code to respect correct layer order is kind of essential to some usecases. I found a couple threads with people looking for similar things. (toggling stuff on and off easily and having a system to load sources when features that need it are visible)
If you still would rather have it as a plugin I will set one up. 👍

HarelM:
I definitely agree there's a strong use case, but nothing is preventing it from being kept as a plugin, so I think it's a better way to start. If the plugin had a lot of downloads and the maintenance is high to keep it up to date with latest version of maplibre it would be a good argument to introduce it, but I think now it's a bit premature.

PR #7413, opened 2026-04-05
chore: better internal types

CommanderStorm:
this PR is best reviewed commit by commit.

This starts working through the backlog of weird typing decisions that have been made.
The goal is to have stricter types, where these are just a few of the many fixes nesseary to achive this


src/style/parse_glyph_pbf.ts
```diff
@@ -5,6 +5,16 @@ const border = 3;
 
 import type {StyleGlyph} from './style_glyph';
 
+interface RawGlyph {
```
HarelM: I'm trying to keep interfaces for things that has methods and "worth" inheritance, this looks like a type...?

src/style/parse_glyph_pbf.ts
```diff
@@ -13,7 +23,7 @@ function readFontstacks(tag: number, glyphs: StyleGlyph[], pbf: Protobuf) {
 
 function readFontstack(tag: number, glyphs: StyleGlyph[], pbf: Protobuf) {
     if (tag === 3) {
-        const {id, bitmap, width, height, left, top, advance} = pbf.readMessage(readGlyph, {});
+        const {id, bitmap, width, height, left, top, advance} = pbf.readMessage(readGlyph, {} as RawGlyph);
```
HarelM, line 26: This means all `RawGlyph` fields are optional which will avoid the need to call `as RawGlyph` here.

src/tile/tile_manager.ts
```diff
@@ -192,8 +194,8 @@ export class TileManager extends Evented {
         } catch (err) {
             tile.state = 'errored';
 
-            if (err.status !== 404) {
-                this._source.fire(new ErrorEvent(err, {tile}));
+            if (!(err instanceof AJAXError) || err.status !== 404) {
```
HarelM: I'm not sure about this change with regards to `addProtocol` use case...

src/util/actor.ts
```diff
@@ -85,7 +85,7 @@ export class Actor implements IActor {
     }
 
     registerMessageHandler<T extends MessageType>(type: T, handler: MessageHandler<T>) {
-        this.messageHandlers[type] = handler;
+        (this.messageHandlers as Record<T, MessageHandler<T>>)[type] = handler;
```
HarelM, line 88: Why was this casting added?

test/integration/render/run_render_tests.ts
```diff
@@ -888,7 +889,7 @@ function applyDebugParameter(options: RenderOptions, page: Page) {
             console.log(`${message.type().substring(0, 3).toUpperCase()} ${messages.filter(Boolean)}`);
         });
 
-        page.on('pageerror', ({message}) => { console.error(message); });
+        page.on('pageerror', (error) => { console.error(error instanceof Error ? error.message : String(error)); });
```
HarelM: Is it possible to use `ensureError(error).message` here as well?

src/geo/projection/mercator_covering_tiles_details_provider.ts
```diff
@@ -8,7 +9,8 @@ import {type CoveringTilesDetailsProvider} from './covering_tiles_details_provid
 
 export class MercatorCoveringTilesDetailsProvider implements CoveringTilesDetailsProvider {
 
-    distanceToTile2d(pointX: number, pointY: number, _tileID: {x: number; y: number; z: number}, aabb: Aabb): number {
+    distanceToTile2d(pointX: number, pointY: number, _tileID: {x: number; y: number; z: number}, boundingVolume: IBoundingVolume): number {
+        const aabb = boundingVolume as Aabb;
```
HarelM, line 13: I find this casting weird, it is possible to avoid it?

HarelM:
I've added some comment, thanks!

src/style/parse_glyph_pbf.ts
CommanderStorm: Ok, there does not exist a rule for this, but writing rules is simple. I added one.

src/style/parse_glyph_pbf.ts
CommanderStorm: why?
The type that this produces is not optional..

src/util/actor.ts
CommanderStorm: with this "deduction guide" what T really is in this method one of the lints that I want to add does not trigger.
The type does not change, I just help TS figure out that T actually is the key.

src/tile/tile_manager.ts
CommanderStorm: tabled for the future

src/geo/projection/mercator_covering_tiles_details_provider.ts
CommanderStorm: the type of this method IS `IBoundingVolume`.
`Aabb` implementes `IBoundingVolume`.
Under strict ts, that is not the same, for our usage it is

src/tile/tile_manager.ts
```diff
@@ -2,6 +2,8 @@ import {create as createSource} from '../source/source';
 
 import {Tile} from './tile';
 import {ErrorEvent, Event, Evented} from '../util/evented';
+import {AJAXError} from '../util/ajax';
```
CommanderStorm:
```suggestion
```

HarelM (approved):

HarelM:
Not sure what I feel about adding a locally written lint rule as this might break when upgrading eslint, so it might be worth just keeping the changes and remove the extra rule... IDK... your call...

CommanderStorm:
It it breaks, we remove the lint. It is as easy as that.

HarelM:
@CommanderStorm can you fix the warning that this rule added please?
image

PR #7420, opened 2026-04-06
Fix fill-outline-color fallback for feature-state expressions

### LINKED ISSUE #4234: `feature-state` with `fill-outline-color` breaks rendering

**browser**: Tested on Firefox and Chrome

### Steps to Trigger Behavior
 1. Create a fill layer with the [fill-outline-color](https://maplibre.org/maplibre-style-spec/layers/#fill-outline-color) property
 2. Upon adding the layer to the map, an error will be raise from `packColor`: `TypeError: can't access property "r", color is null`
 

### Link to Demonstration
https://jsbin.com/rugodev/1/edit?html,output

### Expected Behavior
The outline color should update reflecting the feature-state

### Actual Behavior
The layer rendering breaks and errors are thrown

itisyb:
## Summary
- preserve `fill-outline-color` fallback to `fill-color` when `feature-state` is missing so fill rendering does not break during initial paint population
- apply the same fallback behavior to both source and composite expression binders, and refresh it after paint updates so later `fill-color` changes stay in sync
- add unit coverage for missing state, feature-state updates, fallback refreshes, and invalid color warnings

## Testing
- `npm run lint -- src/data/program_configuration.ts src/data/program_configuration.test.ts`
- `npm run test-unit -- src/data/program_configuration.test.ts`

## Notes
- I attempted to add a render regression too, but `npm run build-dev` is currently blocked by unrelated existing TypeScript errors in `src/data/bucket/line_bucket.ts` and `src/source/geojson_*`, so this PR keeps the regression coverage focused on unit tests.

Fixes #4234


itisyb:
Also created a small demo but don't think need to push/show for this case

HarelM:
I don't think there's an issue with building or render test as CI passes without any issues as far as I know. 
From the initial post it's hard to understand what the changes are and why these are the right places to do them, can you elaborate a bit more on the solution?

itisyb:
> Thanks for taking the time to open this PR. I don't think there's an issue with building or render test as CI passes without any issues as far as I know. From the initial post it's hard to understand what the charges are and why these are the right places to do them, can you elaborate a bit more on the solution?



> Thanks for taking the time to open this PR. I don't think there's an issue with building or render test as CI passes without any issues as far as I know. From the initial post it's hard to understand what the charges are and why these are the right places to do them, can you elaborate a bit more on the solution?

Before the fix, there was a problem with how fill-outline-color was handled when using feature-state.

If the feature-state value was missing, the expression could return null. The rendering system didn’t treat that as “no value” - it treated it like an actual color. Then when packColor() tried to use it, it attempted to access color.r on null, which caused a runtime error: “Cannot read properties of null (reading 'r')”.

it already had a fallback rule where fill-outline-color defaults to fill-color if not specified. But that logic only applied in the constant (non–data-driven) case. In the feature-state (data-driven) path, that fallback wasn’t applied, so instead of defaulting correctly, it crashed.

src/data/program_configuration.ts
```diff
@@ -194,38 +209,42 @@ class CrossFadedConstantBinder implements UniformBinder {
...
+    zoom: number;
     maxValue: number;
 
+    fillColorFallback: FillColorFallback;
     paintVertexArray: StructArray;
     paintVertexAttributes: StructArrayMember[];
     paintVertexBuffer: VertexBuffer;
 
-    constructor(expression: SourceExpression, names: string[], type: string, PaintVertexArray: {
+    constructor(expression: SourceExpressionWithRawEvaluation, names: string[], type: string, zoom: number, fillColorFallback: FillColorFallback, PaintVertexArray: {
```
HarelM, line 231: This looks too specific to fill color, are you sure this is the right place to add this code? I would expect to see this code in the fill bucket/fill program or something related to fill, and not in this "shared" program configuration code.

src/data/program_configuration.ts
```diff
@@ -56,6 +57,20 @@ type PaintOptions = {
...
 };
 
+type FillColorFallback = PossiblyEvaluatedPropertyValue<Color>['value'] | null;
+
+type SourceExpressionWithRawEvaluation = ZoomConstantExpression<'source'>;
+
+type CompositeExpressionWithRawEvaluation = ZoomDependentExpression<'composite'>;
+
+function getFillColorFallback(layer: TypedStyleLayer, property: string): FillColorFallback {
+    return property === 'fill-outline-color' ? (layer.paint as any).get('fill-color').value : null;
```
HarelM: Is there a way to avoid this `as any`?

src/data/program_configuration.ts
HarelM: Seems like this is still relevant after the recent force-push...

src/data/program_configuration.ts
itisyb: I have fixed this, let me know if something else is needed :)

PR #7411, opened 2026-04-05
[PoC] WebGPU Backend

birkskyum:
Early port of the drawables-based WebGPU backend from Native to GL JS, without really changing the webgl1/2 backend - this is largely additive to avoid breaking existing functionality. The goal here is to understand the scope required for landing webgpu, and the bundle size impact (we're looking at ~200kb before gzip). This impl. does not use Luma.GL

Screenshot 2026-04-05 at 15 23 23
Screenshot 2026-04-05 at 13 10 32
Screenshot 2026-04-05 at 15 23 09

What's working
- Most layers (fill, line, x-pattern, x-outline, symbol, circle)
- 3d Buildings (bit of z-fighting)
- 3D Terrain and hillshade
- Raster
- Deckgl layer integration

What's missing
- Sky/Fog
- Globe

To keep the diff "small" i didn't move all the webgl code to a separte webgl folder, but it would arguably be good similar to native to have webgl and webgpu well separted - it would be easier after webgl has migrated to drawables.

I've added a backend flag to the render test harness, webgl1/webgl2/webgpu to force using one of them.

## Launch Checklist

 - [X] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [ ] Briefly describe the changes in this PR.
 - [ ] Link to related issues.
 - [ ] Include before/after visuals or gifs if this PR includes visual changes.
 - [ ] Write tests for all new functionality.
 - [ ] Document any changes to public APIs.
 - [ ] Post benchmark scores.
 - [ ] Add an entry to `CHANGELOG.md` under the `## main` section.

Issue #2606, opened 2023-05-28
WebGPU support


birkskyum:
In order to be ready for when WebGPU browser support increases, it makes sense to start preparing for this feature now.

Enabled by default in stable release ([caniuse.com](https://caniuse.com/?search=webgpu)):
* Chrome/Blink - Supported, since v113, Released May 2nd '23
* Apple/WebKit - Implementing ([progress tracking](https://bugs.webkit.org/buglist.cgi?bug_status=UNCONFIRMED&bug_status=NEW&bug_status=ASSIGNED&bug_status=REOPENED&field0-0-0=product&field0-0-1=component&field0-0-2=alias&field0-0-3=short_desc&field0-0-4=status_whiteboard&field0-0-5=content&order=changeddate%20DESC%2Cbug_status%2Cpriority%2Cassigned_to%2Cbug_id&query_format=advanced&type0-0-0=substring&type0-0-1=substring&type0-0-2=substring&type0-0-3=substring&type0-0-4=substring&type0-0-5=matches&value0-0-0=webgpu&value0-0-1=webgpu&value0-0-2=webgpu&value0-0-3=webgpu&value0-0-4=webgpu&value0-0-5=%22webgpu%22), behind feature flag in stable Safari)
* Firefox/Gecko - Implementing ([progress tracking](https://bugzilla.mozilla.org/show_bug.cgi?id=1616739))

We can modernize our code with WebGL2 in ways that will make a transition to WebGPU easier. These issues can be tracked here as well:
* #2604 
* #2605

HarelM:
Do we want to track this in an issue or a discussion?

zhangyiatmicrosoft:
Based on my investigation, webgpu is not there (yet) if it is VM or nodejs, this means it is nearly impossible to run test cases on CI.
I hope headless Chrome can add it soon enough.

hpugis:
You can pay attention to [GEngine,](https://github.com/GEngine-js/GEngine) the engine provides a Model class that can be used for basic rendering packaging, very flexible and easy to use

hpugis:
`<script type="module">
	import { Model, Context, Texture, RenderTarget, Attachment } from "webgpu-gengine-js";
	import { mat4, vec3 } from "gl-matrix";
	const init = async () => {
		//init context
		const context = new Context({
			canvas: null,
			container: document.getElementById("app"),
			pixelRatio: 1
		});
		const { canvas, presentationFormat } = context;
		await context.init();
[...]

PR #7422, opened 2026-04-06
refactor: webgl isolation

birkskyum:
- rename `/gl` to `/webgl`
- move all webgl draw files to /webgl/draw
- move all webgl program files to /webgl/program/ 
- updates all the imports in the remaining files
- update painter to use a drawFunctions object, so for now `draw.circle` will reach the webgl drawCircle, but it gives the option for us to introduce a webgpu draw function.

Related to
- #2606 
- #7411 

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

### DIFF
--- a/src/render/painter.ts
+++ b/src/render/painter.ts
@@ -18,21 +18,8 @@ import {ColorMode} from '../webgl/color_mode';
 import {CullFaceMode} from '../webgl/cull_face_mode';
 import {Texture} from '../webgl/texture';
 import {Color} from '@maplibre/maplibre-gl-style-spec';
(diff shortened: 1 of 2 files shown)


src/render/painter.ts
```diff
@@ -84,10 +84,31 @@ export type RenderOptions = {
  * @internal
  * Initialize a new painter object.
  */
+export type DrawFunctions = {
```
HarelM: This looks more like an "index" object. I would consider moving it to `../webgl/draw/index` and export it from there, this way all the internal logic of where other files are resides inside it.

PR #7384, opened 2026-04-03
feat: add touch zoom controls

itisyb:
## Summary
- add `map.touchZoomRotate.setZoomRate()` to control pinch and double-tap-drag zoom speed
- add `map.touchZoomRotate.setZoomThreshold()` to control pinch activation sensitivity
- add unit tests and a changelog entry

Closes #7271

PR #7436, opened 2026-04-09
Symbol SDF Performance Optimization: Render Halo and Glyph in a Single Pass (-40% Time Reduction)

xavierjs:
Hello,

## Issue

Text is rendered as SDF symbols, with each glyph represented as a separate symbol. To enhance contrast, a white halo is often added around the text. When a halo is present, rendering occurs in two passes: the first pass draws the halo, and the second pass renders the glyphs themselves. Both passes use the same shader, `symbol_SDF.glsl`, with a uniform variable indicating whether the current pass is for the halo or the glyph.

This means the fragment shader runs for all pixels within the glyph, including those that are fully transparent. The vertex shader is also executed for each glyph (4 times if it is represented by a quad).

## Proposed solution

Both the glyph and halo values are computed within the fragment shader, and the final composited result of the text and halo is rendered in a single pass. This approach reduces the number of shader executions by half.

Here is a screenshot portion of this example: http://localhost:9966/test/examples/filter-symbols-by-text-input.html
glyphHaloView

Now we enable the overdraw inspector by running `map.showOverdrawInspector=true`in the JS console. This animated GIF shows with and without the optimization. There is less overdraw for glyphs with a halo:
glyphHaloOverdraw

## Benchmarks

There is no benchmark for rendering SDF symbols with a halo (`LayerSymbol` benchmark renders characters without a halo).
This PR introduces a new benchmark for that, `LayerSymbolWithHalo`. Here are the results:

**Before:**
Google Chrome, Macbook Pro M4 Max: `4.47ms`
Google Chrome, Macbook Pro M4 Max with swiftshaders (hardware acceleration disabled): `335ms`

**After:**
Google Chrome, Macbook Pro M4 Max: `3.13ms` (30% time reduction)
Google Chrome, Macbook Pro M4 Max with swiftshaders (hardware acceleration disabled): `186ms` (44% time reduction - it shoud theoretically tend to 50% with even crappier GPU)

## Tests

`npm run test-render`runs exactly as before. I just had to update the expected images for the halo tests.
Before, the expected images only had the halo rendered.
With this change, it is not possible to render only the halo. We need to render the glyph + the halo.

## Misc

No related issues.

## Launch Checklist
 - [X] Link to related issues.
 - [X] Post benchmark scores.


test/integration/render/tests/text-halo-width/default/expected.png
HarelM, line 1: This looks like there was a halo added when the original picture didn't have one...?

test/integration/render/tests/text-halo-width/function/expected.png
HarelM, line 1: This one looks worse than before...?

test/integration/render/tests/text-halo-width/property-function/expected.png
HarelM, line 1: Oh I see, the Halo overrides the text for every letter, this might be a degradation in appearance, isn't it?

xavierjs: Yes, if there is glyph overlap then the rendering is different.
Theoretically it should not happen with collision detection.
Unfortunately it is not possible to get exactly the same rendering in this case.

I think this example is nice to figure out what is the best:

Image

Image

If there is a small overlap, I would say that the current rendering is the best since it makes the halo effect less obvious.
But if there is a large overlap, I would go for the new rendering since it ensures that some labels at least will be visible.

xavierjs:
Now it renders the halo in 2 passes only if there is no overlap between glyph halos (it computes the overlap using text-letter-spacing and halo width).
The render tests are passing and the rendering is the same. If there is an overlap between 2 different text labels, and if there is no overlap between glyphs (so 1 pass halo rendering), the result will be more legible with this PR since the halo of the latest rendered text label will be rendered above the first rendered text label, so at least 1 text label will be legible (instead of 0).

src/webgl/program/symbol_program.ts
```diff
@@ -223,7 +225,8 @@ const symbolSDFUniformValues = (
         glCoordMatrix, translation, isText, texSize, pitchedScale), {
         'u_gamma_scale': (pitchWithMap ? Math.cos(transform.pitch * Math.PI / 180.0) * transform.cameraToCenterDistance : 1),
         'u_device_pixel_ratio': painter.pixelRatio,
-        'u_is_halo': +isHalo
+        'u_is_halo': +isHalo,
```
HarelM: Maybe use `isHalo ? 1 : 0` to make this clearer?

test/integration/render/tests/text-halo-width/property-function/expected.png
HarelM, line 1: This still looks worse than before, is it possible to fix this?

HarelM:
In what cases this improvement is really valuable?
I would assume that in most cases, the halo rendering will need the two passes, as in most cases the letters are close to each other, are they?
If the latest code only helps some edge cases then it might not be worth the extra complexity, wouldn't you agree?

xavierjs:
> In what cases this improvement is really valuable? I would assume that in most cases, the halo rendering will need the two passes, as in most cases the letters are close to each other, are they? If the latest code only helps some edge cases then it might not be worth the extra complexity, wouldn't you agree?

If I run this example (very standard map around WDC): http://localhost:9966/test/examples/filter-symbols-by-text-input.html
I set a logpoint here: https://github.com/xavierjs/maplibre-gl-js/blob/a5bd908aa46ca895f97e0b63772ecde006cd645f/src/webgl/draw/draw_symbol.ts#L481
to log `isGlyphOverlap.`

Then it renders:
* 179 times with `isGlyphOverlap=true`, so without the optimization
* 864 times with `isGlyphOverlap=false`, so with the optimization

Then it depends on the style, the map, the resolution,... But overall I think it is worth it.

CHANGELOG.md
```diff
@@ -1,9 +1,13 @@
 ## main
 ### ✨ Features and improvements
 - _...Add new stuff here..._
+<<<<<<< HEAD
```
HarelM: Please fix this...

src/webgl/program/symbol_program.ts
```diff
@@ -223,7 +225,8 @@ const symbolSDFUniformValues = (
         glCoordMatrix, translation, isText, texSize, pitchedScale), {
         'u_gamma_scale': (pitchWithMap ? Math.cos(transform.pitch * Math.PI / 180.0) * transform.cameraToCenterDistance : 1),
         'u_device_pixel_ratio': painter.pixelRatio,
-        'u_is_halo': +isHalo
+        'u_is_halo': + ((isHalo) ? 1 : 0 ),
```
HarelM: I think the `+` and `(` is no longer needed here...

HarelM:
So the render tests all pass with no changes, right? I just want to make sure I'm reading this right.
If this is the case and you see improved performance I think we can have this merged.
Thanks for all the hard work around this!

src/webgl/program/symbol_program.ts
HarelM:
```suggestion
        'u_is_halo': isHalo ? 1 : 0,
```

HarelM (approved):

CHANGELOG.md
```diff
@@ -12,7 +12,7 @@
 
 ### ✨ Features and improvements
 
-- Add `touchZoomRotate.setZoomRate()` and `touchZoomRotate.setZoomThreshold()` to customize touch zoom speed and pinch sensitivity ([#7271](https://github.com/maplibre/maplibre-gl-js/issues/7271))
+- Add `touchZoomRotate.setZoomRate()` and `touchZoomRotate.setZoomThreshold()` to customize touch zoom speed and pinch sensitivity ([#7271](https://github.com/maplibre/maplibre-gl-js/issues/7271)) (by [@gowestmen](https://github.com/gowestmen))
```
CommanderStorm, line 15: https://github.com/maplibre/maplibre-gl-js/pull/7384

```suggestion
- Add `touchZoomRotate.setZoomRate()` and `touchZoomRotate.setZoomThreshold()` to customize touch zoom speed and pinch sensitivity ([#7271](https://github.com/maplibre/maplibre-gl-js/issues/7271)) (by [@itisyb](https://github.com/itisyb))
```

HarelM: My bad, sorry...

xavierjs:
> So the render tests all pass with no changes, right? I just want to make sure I'm reading this right. If this is the case and you see improved performance I think we can have this merged. Thanks for all the hard work around this!

Yes there are no change in the render tests now. Thank you Harel for your great feedback and review!

PR #7440, opened 2026-04-09
docs: split html and js and indicate the new items

CommanderStorm:
basically, this 

image


build/generate-docs.ts
```diff
@@ -93,6 +155,10 @@ async function generateExamplesFolder() {
     const examplesFolder = path.join('test', 'examples');
     const files = fs.readdirSync(examplesFolder).filter(f => f.endsWith('html'));
     const maplibreUnpkg = `https://unpkg.com/maplibre-gl@${packageJson.version}/`;
+    const newExamples = getNewExamples();
+    if (newExamples.size > 0) {
+        console.log(`${newExamples.size} of ${files.length} examples marked as "new": ${newExamples.map(f => path.basename(f).join(', ')}`);
```
HarelM: Left over?

HarelM:
How does the markdown of this looks like?
As far as I understand tabs are not considered regular markdown
for example the following doesn't look like tabs here:
 
=== Tab1

content

=== Tab2

content 2


So if I'm not mistaken, this is zensical (mkdocs) specific which I would like to avoid to be able to port the docs if needed in the future.

Same for the header.

While I think this is a nice addition to the docs, I think it's not worth the complexity and "vendor lock". sorry... :-/

CommanderStorm:
> So if I'm not mistaken, this is zensical (mkdocs) specific which I would like to avoid to be able to port the docs if needed in the future.
> 
> Same for the header.
> 
> While I think this is a nice addition to the docs, I think it's not worth the complexity and "vendor lock". sorry... :-

if we have another solution we want to migrate to they likey support tabs.
That is super common, even though evryone has their own syntax for how to do this.

Realistically, there will always be coupling to the docs framework. Just have a look at our social cards or how we do admonations..
You change the syntax and move on. No biggie.
This is not like this is a one-way street, we can just remove the feature if it hinders us.

When looking at the examples, they look much more approchable this way if we "hide" the ugly html and show only the js.

build/generate-docs.ts
```diff
@@ -93,6 +155,10 @@ async function generateExamplesFolder() {
...
+    }
```
CommanderStorm:
```suggestion
```

HarelM:
> You change the syntax and move on. No biggie.

I completely disagree, as I was the person who moved the docs from mapbox's proprietary framework and deleted the docusaurus approach I can tell you that I've worked hard to keep it simple, maintainable, and readable.
I'd like to keep it this way.
The example are searchable from google, and from the website itself.
Adding specific markdown that's not a standard is something I'd like to avoid.
Having all the example in a single place is easy to read, copy and maintain.
I understand the value you are looking to add, don't get me wrong, but I have seen how this gets complicated fast...

HarelM:
If the above message sounds too aggressive I'm sorry, I'll be happy to discuss this online if you'd like.

build/generate-docs.ts
```diff
@@ -33,17 +35,77 @@ Import declarations are omitted from the examples for brevity.
...
+    let codeBlock: string;
+    if (jsContent) {
+        const jsBlock = indentBlock(`\`\`\`js\n${jsContent}\n\`\`\``);
+        const htmlBlock = indentBlock(`\`\`\`html\n${htmlContent.trimEnd()}\n\`\`\``);
+        codeBlock = `=== "Only JS"\n\n${jsBlock}\n\n=== "Full HTML"\n\n${htmlBlock}`;
+    } else {
+        codeBlock = `\`\`\`html\n${htmlContent}\n\`\`\``;
+    }
+
+    return `${frontmatter}
```
CommanderStorm, line 93: @HarelM
this here is how you migrate to a pure md version if the framework that you are planning to migrate to does not support tabs.
Any other syntax is a one line change.

Removing the "new" banner is similarly simple to do.

I really see no reason why we should make the docs worse in terms of how approchable they are and how we advertise our new features.

In the conversations that I have had, the examples are the main way how people are interacting with our docs, so we should do our best to make them approchable.

```suggestion
    const jsContent = extractJsFromHtml(htmlContent);

    let codeBlock: string;
    if (jsContent) {
        const jsBlock = indentBlock(`\`\`\`js\n${jsContent}\n\`\`\``);
        const htmlBlock = indentBlock(`\`\`\`html\n${htmlContent.trimEnd()}\n\`\`\``);
        codeBlock = `<details><summary>Only JS</summary>\n\n${jsBlock}\n\n</details><details><summary>Full HTML</summary>\n\n${htmlBlock}\n\n</details>`;
    } else {
        codeBlock = `\`\`\`html\n${htmlContent}\n\`\`\``;
    }

    return `
```

HarelM:
Let's try to find a time to talk about it online please.

test/examples/add-a-3d-model-to-globe-using-threejs.html
```diff
@@ -4,6 +4,7 @@
 <head>
     <title>Add a 3D model to globe using three.js</title>
     <meta property="og:description" content="Use a custom style layer with three.js to add a 3D model to a globe." />
+    <meta property="og:created" content="2025-06-25" />
```
HarelM, line 7: Looks like most of the dates are 2025-06-25, which I find weird, are you sure about this?

test/build/examples.test.ts
```diff
@@ -0,0 +1,91 @@
...
+ */
+function findLine(lines: string[], pattern: RegExp): number | null {
+    const idx = lines.findIndex(l => pattern.test(l));
+    return idx >= 0 ? idx + 1 : null;
+}
+
+describe('Example HTML files', () => {
+    const exampleFiles = globSync('test/examples/*.html').sort();
+
+    test.each(exampleFiles)('%s has required meta tags', (file) => {
```
HarelM: Can this use a regular for loop?

test/build/examples.test.ts
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Find the 1-based line number of the first line matching `pattern`,
+ * or null if not found.
+ */
+function findLine(lines: string[], pattern: RegExp): number | null {
+    const idx = lines.findIndex(l => pattern.test(l));
+    return idx >= 0 ? idx + 1 : null;
+}
+
+describe('Example HTML files', () => {
```
HarelM, line 5: These are nice checks, thanks!
Can you add a check to make sure the file name and title match (considering kebab-casing)?

build/generate-docs.ts
```diff
@@ -33,17 +35,69 @@ Import declarations are omitted from the examples for brevity.
     return intro;
 }
 
-function generateMarkdownForExample(title: string, description: string, file: string, htmlContent: string): string {
-    return `
+function isNewExample(htmlContentLines: string[]): boolean {
+    const createdLine = htmlContentLines.find(l => l.includes('og:created'));
+    if (!createdLine) return false;
+    const match = createdLine.match(/content=["'](\d{4}-\d{2}-\d{2})["']/);
```
HarelM, line 41: Is there a better way to parse these files? maybe xml2js or something similar?

build/generate-docs.ts
```diff
@@ -33,17 +35,69 @@ Import declarations are omitted from the examples for brevity.
...
+    if (!match) return false;
+    const createdDate = new Date(match[1]);
+    const sixMonthsAgo = new Date();
+    sixMonthsAgo.setMonth(sixMonthsAgo.getMonth() - 6);
+    return createdDate >= sixMonthsAgo;
+}
+
+function extractJsFromHtml(htmlContent: string): string | null {
+    const scriptRegex = /<script(?![^>]*\bsrc\b)(?![^>]*type="importmap")[^>]*>([\s\S]*?)<\/script>/g;
```
HarelM, line 50: Same as above, I think a proper html parser would be better here...

build/generate-docs.ts
```diff
@@ -33,17 +35,69 @@ Import declarations are omitted from the examples for brevity.
...
+    const indent = ' '.repeat(spaces);
+    return text.split('\n').map(line => line.length > 0 ? `${indent}${line}` : '').join('\n');
+}
+
+function generateMarkdownForExample(title: string, description: string, file: string, htmlContent: string, isNew: boolean): string {
+    const frontmatter = isNew ? '---\nstatus: new\n---\n' : '';
+    const jsContent = extractJsFromHtml(htmlContent);
+
+    let codeBlock: string;
+    if (jsContent) {
```
HarelM, line 85: Are there examples without jscontent? Isn't this condition always true for our use case?

build/generate-docs.ts
```diff
@@ -33,17 +35,69 @@ Import declarations are omitted from the examples for brevity.
...
+function extractJsFromHtml(htmlContent: string): string | null {
+    const scriptRegex = /<script(?![^>]*\bsrc\b)(?![^>]*type="importmap")[^>]*>([\s\S]*?)<\/script>/g;
+    const scripts: string[] = [];
+    let match;
+    while ((match = scriptRegex.exec(htmlContent)) !== null) {
+        if (match[1].trim()) {
+            scripts.push(match[1]);
+        }
+    }
+    if (scripts.length === 0) return null;
```
HarelM, line 58: I prefer the generation to fail or throw in this case.

build/generate-docs.ts
```diff
@@ -33,17 +35,69 @@ Import declarations are omitted from the examples for brevity.
...
+
+    let js = scripts.join('\n\n');
+
+    // Strip common leading whitespace
+    const lines = js.split('\n');
+    const nonEmptyLines = lines.filter(l => l.trim().length > 0);
+    if (nonEmptyLines.length > 0) {
```
HarelM, line 65: Is this if really needed?

build/generate-docs.ts
```diff
@@ -33,17 +35,69 @@ Import declarations are omitted from the examples for brevity.
...
+        const minIndent = Math.min(...nonEmptyLines.map(l => l.match(/^\s*/)[0].length));
+        if (minIndent > 0) {
```
HarelM, line 67: I'm not sure what this code is doing, but I would prefer to have some linting or formatting ran on the html examples instead of complicating the code here with indent calculations...

test/build/examples.test.ts
```diff
@@ -0,0 +1,91 @@
+import {describe, test, expect} from 'vitest';
+import {globSync} from 'glob';
+import fs from 'fs';
+
+/**
+ * Show a snippet of the file around `lineNum` (1-based) with a gutter,
+ * highlighting the target line. If lineNum is null, shows the first few
+ * lines of <head> as context.
+ */
+function snippet(lines: string[], lineNum: number | null, file: string): string {
```
HarelM: This is nice, but I think it's an overkill, let's keep it simple...

test/build/examples.test.ts
```diff
@@ -0,0 +1,91 @@
...
+
+describe('Example HTML files', () => {
+    const exampleFiles = globSync('test/examples/*.html').sort();
+
+    test.each(exampleFiles)('%s has required meta tags', (file) => {
+        const content = fs.readFileSync(file, 'utf-8');
+        const lines = content.split('\n');
+        const errors: string[] = [];
+
+        // Check og:description
```
HarelM: This should not be a comment, but instead part of the test name.

test/build/examples.test.ts
```diff
@@ -0,0 +1,91 @@
...
+            const loc = findLine(lines, /og:description/);
+            errors.push(
+                `error: \`og:description\` content is empty\n${
+                    snippet(lines, loc, file)}\n` +
+                '  = help: provide a meaningful description, e.g.:\n' +
+                '          <meta property="og:description" content="Demonstrates how to ..." />'
+            );
+        }
+
+        // Check og:created
```
HarelM: Same here, this should be a test, not a comment inside a test.

test/build/examples.test.ts
```diff
@@ -0,0 +1,91 @@
...
+            const loc = findLine(lines, /og:created/);
+            errors.push(
+                `error: \`og:created\` has invalid date format "${createdMatch[1]}"\n${
+                    snippet(lines, loc, file)}\n` +
+                '  = help: use YYYY-MM-DD format, e.g.:\n' +
+                '          <meta property="og:created" content="2025-10-31" />'
+            );
+        }
+
+        if (errors.length > 0) {
```
HarelM: No need to collect errors, just use regular assertion.

HarelM:
I've added some comments, mostly for readability and reduced complexity.
Once addressed this can be merged.
Thanks for working on this!

test/examples/add-a-3d-model-to-globe-using-threejs.html
CommanderStorm: Yes, that was the last major change for all.
https://github.com/CommanderStorm/maplibre-gl-js/commit/f9140dc2c2ff2b77bdac4d882ce1f2452e2293e6

I am not digging further, this is only used to control if something is new or not.
I don't plan on adding this into the UI

test/build/examples.test.ts
CommanderStorm: lets use vitests features instead

build/generate-docs.ts
CommanderStorm: regex works fine for this usecase

build/generate-docs.ts
CommanderStorm: This is defenstive to still show something if the parser breaks.
I think it is fine to have

test/build/examples.test.ts
CommanderStorm: Better error messages are better, lets keep it

test/build/examples.test.ts
CommanderStorm: We usually put the menthod here that is being tested.
There is no such method, so not applicable

build/generate-docs.ts
CommanderStorm: I think it is overkill to add a dependency for this

build/generate-docs.ts
CommanderStorm: You cannot have this be low maintenance and let the code throw errors on syntax errors.

Those are diverging goals that are fundamentally not the same.

build/generate-docs.ts
CommanderStorm: Yes, this removes the common leading whitespace.
We want leading whitespace removed

build/generate-docs.ts
CommanderStorm: this is not about the html, we don't want random indents on the JS code we extract.
We want that code de-indented to be 

```js
fn foo {
...
}
```

instead of 


```js
                fn foo {
                ...
                }
```

test/build/examples.test.ts
```diff
@@ -0,0 +1,87 @@
...
+        const lines = content.split('\n');
+
+        const descriptionMatch = content.match(/<meta\s+property=["']og:description["']\s+content=["']([^"']*)["']/);
+        if (!descriptionMatch) {
+            const loc = findLine(lines, /<head/i);
+            expect.fail(
+                ++missing \`og:description\` meta tag\n${
+                    snippet(lines, loc, file)}\n` +
+                '  = help: add inside <head>:\n' +
+                '          <meta property="og:description" content="A short description of what this example demonstrates." />'
```
CommanderStorm:
```suggestion
                `missing \`og:description\` meta tag\n${
                    snippet(lines, loc, file)}\n` +
                '  = help: add inside <head>:\n' +
                '          <meta property="og:description" content="A short description of what this example demonstrates." />'
```

HarelM (approved):

PR #7441, opened 2026-04-09
docs: make the guides on the examples-docs overview page more discoverable to help people discover which guides we have

CommanderStorm:
During the conversations that I have had it came up that people have trouble discovering which guides exist and what gl-js has to offer.
For example knowing that we have terrain3d is fairly unknown (presumably) because it happens many pages down.
We should make this better discoverable to market ourselves better.

Users should not need to have to jump via google and instead should be able to have a good overview page that shares more guides per page.
This is said overview page, which cards instead of the half-page-images we had before.

looks now like this:

| before | after |
|--------|--------|
| image | image |

In terms of scope this is fairly simple to add the cards to the codegen and adjust the css to be responsive/plesant to look at/match the rest of our docs.


build/generate-docs.ts
```diff
@@ -33,36 +35,106 @@ Import declarations are omitted from the examples for brevity.
...
         const outputPath = imagePath.replace('.png', '.webp');
         promises.push(sharp(imagePath).webp({quality: 90, lossless: false}).toFile(outputPath));
-        indexMarkdown += `
-## [${indexArrayItem.title}](./${indexArrayItem.mdFileName})
-
-![${indexArrayItem.description}](${outputPath.replace('docs/', '../')}){ loading=lazy }
-
-${indexArrayItem.description}
+        const desc = indexArrayItem.description || '';
+        indexMarkdown += `<a class="example-card" href="./${indexArrayItem.mdFileName}">
```
HarelM: Nope, sorry, markdown only, no HTML.
I've seen what happened to maplibre.io repo, it's full of HTML and CSS and it's unmaintainable. I don't want that here, sorry...

docs/assets/extra.css
```diff
@@ -8,8 +8,83 @@
...
 }
-  
+
 [data-md-color-scheme="slate"] {
     --md-primary-fg-color: #295DAA;
     --md-accent-fg-color: #568ad6;
 }
+
+/* Examples overview card grid */
+.examples-grid {
```
HarelM, line 15: The opposite of KISS... I'd like to avoid this please.

build/generate-docs.ts
CommanderStorm: but that is simply impossible in pure markdown.
I cannot do a table that flexes from 1 to 3 element.

I also don't think this is unmaintanable..
It is fairly simple html and just where I cannot get away with this otherwise

You cannot with a straight face tell me that this is better...

image

docs/assets/extra.css
CommanderStorm: I don't see a way to do this.

How can we get to a more overviewy Overview page without custom styling and without using any of zenzicals features?
It really does not make sense to me why you want this to be worse.

build/generate-docs.ts
```diff
@@ -33,36 +35,106 @@ Import declarations are omitted from the examples for brevity.
...
+        const filenames = new Set<string>();
+        for (const line of output.split('\n')) {
+            const trimmed = line.trim();
+            if (trimmed && trimmed.startsWith('test/examples/')) {
+                filenames.add(path.basename(trimmed));
+            }
+        }
+        return filenames;
+    } catch {
+        return new Set<string>();
```
louwers: This smells.

build/generate-docs.ts
```diff
@@ -1,5 +1,7 @@
 import fs from 'fs';
 import path from 'path';
+
+
 import typedocConfig from '../typedoc.json' with {type: 'json'};
```
CommanderStorm, line 4:
```suggestion
import path from 'path';
import typedocConfig from '../typedoc.json' with {type: 'json'};
```

HarelM:
@CommanderStorm are you planning on pushing this forward or should I close it?

CommanderStorm:
Yes, but since this based of the other PR this needed to be merged first.
Can put some engineering behind the changes that you asked regarding inlining the things that are possible to inline

build/generate-docs.ts
```diff
@@ -102,21 +102,27 @@ ${codeBlock}
...
         const outputPath = imagePath.replace('.png', '.webp');
         promises.push(sharp(imagePath).webp({quality: 90, lossless: false}).toFile(outputPath));
-        indexMarkdown += `
-## [${indexArrayItem.title}](./${indexArrayItem.mdFileName})
-
-![${indexArrayItem.description}](${outputPath.replace('docs/', '../')}){ loading=lazy }
-
-${indexArrayItem.description}
+        const desc = indexArrayItem.description || '';
+        indexMarkdown += `<a class="example-card" href="./${indexArrayItem.mdFileName.replace(/.md$/, '/')}" style="border:1px solid var(--md-default-fg-color--lightest);border-radius:0.5rem;text-decoration:none!important;color:inherit!important;transition:box-shadow 0.2s,transform 0.2s;display:flex;flex-direction:column">
```
CommanderStorm: I am not under the oppunion that these inline styles are more maintainable, but if that is what merging this PR requries, so be it.

I would strongly prefer to keep the css in extra.css and keep this code un-obvuscated

docs/assets/extra.css
```diff
@@ -8,8 +8,27 @@
...
-  
+
 [data-md-color-scheme="slate"] {
     --md-primary-fg-color: #295DAA;
     --md-accent-fg-color: #568ad6;
 }
+
+/* stylelint-disable selector-class-pattern */
+
+/* Example card hover */
```
HarelM: I would recommend to remove it, most users browse from a mobile phone and this extra complexity of hover effects is redundant.

build/generate-docs.ts
HarelM: Can you move the replace out of this part to increase readability?
I think you can also remove some over complicated css stuff and just keep the basics for it to look as a grid on the desktop assuming this is what this is for...

docs/assets/extra.css
```diff
@@ -8,8 +8,27 @@
...
+.example-card:hover {
+    box-shadow: 0 4px 16px rgb(0 0 0 / 0.12);
+    transform: translateY(-2px);
+}
+
+[data-md-color-scheme="slate"] .example-card:hover {
+    box-shadow: 0 4px 16px rgb(0 0 0 / 0.4);
+}
+
+/* Hide the right sidebar (TOC) when it has no entries */
```
HarelM, line 71: Is it possible to ask Zensical to implement this maybe?

docs/assets/extra.css
CommanderStorm: I would recomend to keep it, it makes it more obvious that this is a link on desktop

build/generate-docs.ts
CommanderStorm: Which parts are overcomplicated? They all serve a purpose

build/generate-docs.ts
```diff
@@ -102,21 +102,27 @@ ${codeBlock}
...
+<div style="position:relative;overflow:hidden;border-radius:0.5rem 0.5rem 0 0">
+<img src="${outputPath.replace('docs/', '../')}" loading="lazy" alt="${desc}" style="width:100%;aspect-ratio:12/5;object-fit:cover;display:block">
```
CommanderStorm:
```suggestion
        const cardFileName = indexArrayItem.mdFileName.replace(/.md$/, '/');
        const cardImg = outputPath.replace('docs/', '../');
        indexMarkdown += `<a class="example-card" href="./${cardFileName}" style="border:1px solid var(--md-default-fg-color--lightest);border-radius:0.5rem;text-decoration:none!important;color:inherit!important;transition:box-shadow 0.2s,transform 0.2s;display:flex;flex-direction:column">
<div style="position:relative;overflow:hidden;border-radius:0.5rem 0.5rem 0 0">
<img src="${cardImg}" loading="lazy" alt="${desc}" style="width:100%;aspect-ratio:12/5;object-fit:cover;display:block">
```

docs/assets/extra.css
CommanderStorm: It is, but lets get some progress instead, ok?

docs/assets/extra.css
```diff
@@ -2,8 +2,27 @@
...
+
+/* Example card hover */
+.example-card:hover {
+    box-shadow: 0 4px 16px rgb(0 0 0 / 0.12);
+    transform: translateY(-2px);
+}
+
+[data-md-color-scheme="slate"] .example-card:hover {
+    box-shadow: 0 4px 16px rgb(0 0 0 / 0.4);
+}
```
CommanderStorm, line 45: @HarelM would this (imo UX regression) enable us to merge this finally?

```suggestion

/* stylelint-disable selector-class-pattern */
```

build/generate-docs.ts
CommanderStorm: https://github.com/maplibre/maplibre-gl-js/pull/7441/commits/062ff6f43eb4dbdbfad2d31e5b779c29f76cd83d was added as the review-requested compromise, but I really don't like how this obvuscates the css.

I would much prefer the intial version that is clearer to read..

mizmay:
Hopefully we all agree that:
- The MapLibre GJ JS guides page is overwhelming and difficult to navigate, which makes guides difficult to discover.
- We want to avoid customizing Zensical, as that defeats the purpose of having an automatic doc generator, and also customizations like this tend to be brittle across mobile, different browsers, etc.

As far as customization goes, we can probably also all agree that adding an external CSS file is less brittle than mixing random bits of HTML in the markdown. But it is still a customization. The marginal benefit may not be worth the cost, and it is reasonable for maintainers to be skeptical.

A few I can suggest, if the custom CSS solution is unacceptable:
1. Instead of custom css or injected HTML, could the equivalent of this plug-in work: https://www.neoteroi.dev/mkdocs-plugins/cards/ , and if so can we register we want it with the Zensical devs?
2. Let's curate the guides! No layout will improve this page as much as better organization and focus, and a plan to remove/consolidate guides where we have too many, commission, create and post guides where there are gaps.
3. If pursuing it with Zensical doesn't make sense, let's get sharper on what the final layout should be, for example using grids https://zensical.org/docs/authoring/grids/ might require *less* custom CSS

CommanderStorm:
> customizations like this tend to be brittle across mobile, different browsers

I’m not fully convinced this applies in this case.
The approach I’m proposing is fairly minimal - standard markdown plus a small amount of CSS within a codegen-framework we already maintain.
It is reeeaally not a lot of code that we are talking about!

It’s not relying on anything exotic, just well-supported features like CSS Grids.

From what I’ve tested, it behaves consistently across viewports and modern browsers. If there are specific concerns about edge cases or environments where this might break, I can look into those and validate them.
I don't think there are any bugs or shortcomings.

> Instead of custom css or injected HTML, could the equivalent of this plug-in work: [https://www.neoteroi.dev/mkdocs-plugins/cards/](https://www.neoteroi.dev/mkdocs-plugins/cards/)

> using grids https://zensical.org/docs/authoring/grids/ might require *less* custom CSS

My understanding is that using more advanced features of our documentation framework has already been considered out of scope in earlier discussions (in the previous PR with @HarelM and above in this PR).

There is also the built-in Zensical Grids functionality, but as I understand it, the concern there is "portability".
@HarelM set the red line at "is standard markdown" and cards are unlikely to be portable between docs, unless we use html for that. Html+CSS is portable and solves this constraints well imo.

I did experiment with Zensical Grids, but wasn’t able to achieve the desired layout (link behavior, image placement, margins, hover states, visual consistency like spacing, radii, and colors). These are relatively straightforward to implement with a small amount of CSS.
As you can see form [the files tab](https://github.com/maplibre/maplibre-gl-js/pull/7441/commits), this is not chnanging a lot..
To name a concrete limitation: I could not link to the page underneath, could not make the new button appear how + where it should and could not make the 3-wide-layout work.

So that does not sattisfy my design coonstraints that it actually needs to look good and work well for the end user...

-> This is what this PR is adding all of these, the cards I am adding work!

This might be a skill issue and this might be possible in zenzical, not sure.
I would like to consider this an unrelated refactoring if anything.

> Let's curate the guides

Different PR. Lets focus on the change at hand, ok?
This PR is only about increasing the discoverability, not changing the content.

HarelM:
My main concern is less about this specific feature, but more about "let's customize the Sh%4t out of the docs".
This is not a theoretical concern - maplibre.prg is mega customized where adding a single newsletter page adds extra html, react and what not, and still to this day, when you browse from a mobile device you can scroll the page sideways due to all the customization that were made there. Every design change there is painful, and I've seen PRs that did it and broke it to pieces.

"It is reeeaally not a lot of code that we are talking about!" is what starts this, and I've seen it in my career countless times.

We've introduced tabs in a previous PR, so card won't be a stretch now, if you think it's a good middle ground.

To be honest, given that we have more than 130 examples, I don't think putting them 3 in a row, 2 in a row or anything like that will really make them more or less discoverable, it's simply too many to really scroll through.

We have a search capability, and SEO for google discoverability, which is what we should focus on making sure works well.

If you want to group the examples into "families" to make them more discoverable that's an option too. I've seen content website like news, hiking routes etc, go that route where you add more hierarchy to allow better discoverability. I'm not sure I have a good idea how to group the examples to be honest, but I think it's a better solution than what's suggested here.

The bottom line is that I think this solution doesn't really addresses the problem statement well, it's a nice addition with the cost of future pain. So from engineering perspective, not a great direction.

To be fair, I did try to tell myself "hey, why not, it's just a few lines of CSS, how harmful can it be". But the more I think about it the harder I'm having telling myself that... Sorry... :-/

CommanderStorm:
As I said, that would be a refactoring at best.
I CANNOT make this work unsing zenzcal cards. I have TRIED.

I guess we are stuck with the shit design

mizmay:
@CommanderStorm thanks for continuing to push for improvements on the docs and MapLibre frontmatter. I know it's been frustrating and confounding at times.

I want to push back gently on "stuck with the shit design." The work you've done here has real value: you've explored the problem in depth, tried approaches, and hit concrete limitations, that's progress that is worth preserving.

I'm going to open an issue to capture the full problem space as best I can so this work isn't lost and the next person isn't starting from zero. Please take a look and edit or comment to make sure it accurately reflects what you tried and where it fell short? You have more detail on the Zensical grids limitations than anyone for all we know, plus you came up with some novel and interesting POCs with this implementation. It may not be a one-shot deal, but I'd like to help move this discussion forward.

HarelM:
Can you move the css back to the extra.css file. I agree it's not better the current way. 
Sorry for the extra work...

CommanderStorm:
It is this way? 😉

build/generate-docs.ts
```diff
@@ -102,21 +102,29 @@ ${codeBlock}
...
-        indexMarkdown += `
-## [${indexArrayItem.title}](./${indexArrayItem.mdFileName})
-
-![${indexArrayItem.description}](${outputPath.replace('docs/', '../')}){ loading=lazy }
-
-${indexArrayItem.description}
+        const desc = indexArrayItem.description || '';
+        const cardFileName = indexArrayItem.mdFileName.replace(/.md$/, '/');
+        const cardImg = outputPath.replace('docs/', '../');
+        indexMarkdown += `<a class="example-card" href="./${cardFileName}">
```
HarelM, line 124: Can you split the image creation and markdown writing as this can cause random writing order. Saw this bug when I tried to see how cards would look.

docs/assets/extra.css
HarelM: You can keep the hover.

build/generate-docs.ts
CommanderStorm: Sounds implausibly.
I am not changing the order here and writitng order does not matter.

If `indexArray` is non-determinisically ordered, that might happen, but not otherwise.
That is a different bug though and should be handled in another PR if that.

https://github.com/maplibre/maplibre-gl-js/blob/4d3a32d2a19d0bbe081e0ebd97afd7e9ae4bdb95/build/generate-docs.ts#L148

HarelM: The order of the examples should be the same as the file names order.
But the promises to `sharp` can return in unexpected order.
So it would be better to first run the loop of promises with the `sharp` calls and only after that call add all the pages to the markdown.
It will also make the code clearer to read as the markdown writing won't be part of the promises code...

CommanderStorm: No, we are not awaiting this in the loop.
Sharp is not the culprit, at least that does not make any sense.
The promise does not include the part where the md is being concatinated together.

It is because of the missing grouping and sorting, different PR.

HarelM: Oh, right, still worth splitting this part out of the loop to keep the markdown handling separate from the sharp calls, they are not really related.

CommanderStorm: they both rely on `outputPath`, so they are related in that.
I think lets merge and change the grouping in another PR.

If you want to nitpick more, why don't you just commit a change to the PR please?

HarelM: > If you want to nitpick more, why don't you just commit a change to the PR please?

That's uncalled for.

HarelM:
Please fix code hygiene.

build/generate-docs.ts
CommanderStorm: The current logic works as intended, and I don’t see a clear benefit in splitting this into another loop—it would duplicate the dependency on `outputPath` and spread related logic across multiple places.

Given that, I’d prefer to keep this as-is for now. If you feel strongly about restructuring it, feel free to propose a concrete change or follow-up PR and I’m happy to take a look.

build/generate-docs.ts
```diff
@@ -110,18 +110,29 @@ async function packImages(indexArray: HtmlDoc[]) {
     for (const indexArrayItem of indexArray) {
         const imagePath = `docs/assets/examples/${indexArrayItem.mdFileName.replace('.md', '.png')}`;
         const outputPath = imagePath.replace('.png', '.webp');
-        promises.push(sharp(imagePath).webp({quality: 90, lossless: false}).toFile(outputPath));
+        promises.push(sharp(imagePath).webp({quality: 90, lossless: false}).toFile(
+        ));
```
CommanderStorm:
```suggestion
        promises.push(sharp(imagePath).webp({quality: 90, lossless: false}).toFile());
```

build/generate-docs.ts
```diff
@@ -110,18 +110,28 @@ async function packImages(indexArray: HtmlDoc[]) {
     for (const indexArrayItem of indexArray) {
         const imagePath = `docs/assets/examples/${indexArrayItem.mdFileName.replace('.md', '.png')}`;
         const outputPath = imagePath.replace('.png', '.webp');
-        promises.push(sharp(imagePath).webp({quality: 90, lossless: false}).toFile(outputPath));
+        promises.push(sharp(imagePath).webp({quality: 90, lossless: false}).toFile());
```
CommanderStorm:
```suggestion
        promises.push(sharp(imagePath).webp({quality: 90, lossless: false}).toFile(outputPath));
```

HarelM (approved):

PR #7442, opened 2026-04-09
feat(marker): support number for opacity options

### LINKED ISSUE #7433: Globe, markers and popups
**maplibre-gl-js version**: 5.22.0

**browser**: chrome/edge/firefox

On a 3d globe, I set some markers to display a popup when the marker is clicked.

I modified the marker class to show the mouse pointer (cursor: help).

When the markers are hidden behind the globe (with the default opacity of 0.2), the mouse cursor changes (it shouldn't ?), but the marker doesn't display a popup when clicked (this seems correct to me).

However, if I set any value for opacityWhenCovered (even 0, the marker isn't visible), the mouse cursor changes and the popup appears when the marker is clicked (this doesn't seem correct to me).

YuChunTsao:
The issue #7433 highlighted a bug where manually setting `opacityWhenCovered` (e.g., to 0) would still allow the marker's popup to appear when the marker was behind the globe.

Previously, `opacity` and `opacityWhenCovered` in `MarkerOptions` and `setOpacity()` only accepted `string`, which required users to pass values like `'0.5'` instead of the more natural `0.5`.
 
- This PR extends the type to `string | number` for a more intuitive API. The value is converted internally via `String()` before being applied to `element.style.opacity`, so existing behavior is fully preserved.
- Enable custom styling for occluded markers through the new `.maplibregl-marker-covered` CSS class.

https://github.com/user-attachments/assets/b4859a42-697a-4de2-a503-5676c62e6b7e

## Launch Checklist

 - [x] Link to related issues.


YuChunTsao:
I noticed that the results in the example environment are different from what I see in my local dev environment. I'm going to investigate this further and will update the PR once I find the cause.

src/ui/marker.ts
```diff
@@ -861,21 +868,21 @@ export class Marker extends Evented {
     /**
      * Sets the `opacity` and `opacityWhenCovered` properties of the marker.
      * When called without arguments, resets opacity and opacityWhenCovered to defaults
-     * @param opacity - Sets the `opacity` property of the marker.
-     * @param opacityWhenCovered - Sets the `opacityWhenCovered` property of the marker.
+     * @param opacity - Sets the `opacity` property of the marker. Accepts any valid CSS opacity value as a number or string. Defaults to `1`.
+     * @param opacityWhenCovered - Sets the `opacityWhenCovered` property of the marker. Accepts any valid CSS opacity value as a number or string. Defaults to `0.2`.
```
HarelM: That's not entirely correct given the code below, only if the two are not passed then it is reverted to default values, otherwise it uses only what was passed...

src/ui/marker.ts
```diff
@@ -601,6 +607,7 @@ export class Marker extends Evented {
 
         if (this._popup?.isOpen() && centerIsInvisible) this._popup.remove();
         this._element.style.opacity = centerIsInvisible ? this._opacityWhenCovered : this._opacity;
+        this._element.classList.toggle('maplibregl-marker-occluded', centerIsInvisible);
```
HarelM: The API is using the term "covered", should the class use the same terms as the API?
(`opacityWhenCovered`).

src/ui/marker.test.ts
HarelM, line 1: There should be a test or two to check the newly added class I think.

test/examples/style-markers-when-occluded.html
HarelM, line 1: I'm not sure this is worth an example, I don't think it add enough.
If you still think this is valuable, please make sure to link it from the setOpacity and popup tsdocs comments with `@see`.

src/ui/marker.test.ts
YuChunTsao: I've added four tests covering the `maplibregl-marker-covered`

test/examples/style-markers-when-occluded.html
YuChunTsao: I've removed it and instead documented `maplibregl-marker-covered` directly in the Marker class `JSDoc` as a CSS Classes section

src/ui/marker.ts
YuChunTsao: I've renamed the class to `maplibregl-marker-covered` to be consistent with the `opacityWhenCovered` option name in the API.

src/ui/marker.ts
YuChunTsao: You're right, the description I added was inaccurate. I've reverted the `@param` descriptions back to the original wording to avoid any confusion.

HarelM (approved):
THANKS!

PR #7446, opened 2026-04-10
fix: guard against null style in _contextRestored

mvanhorn:
## Launch Checklist

- [x] Link to related issues.

## Description

When a map is initialized without a style (which is valid per the TypeScript typings), `_contextLost` sets `this.style = null`. On context restore, `_contextRestored` skips the `setStyle` call (since `_lostContextStyle.style` is null) but still tries to access `this.style.imageManager.images`, which throws.

The fix adds a null check on `this.style` before restoring cached images. If there's no style, there are no images to restore.

### Test

Added a test that creates a map without a style, triggers context loss then context restore, and verifies no error is thrown.

Fixes #7432

### DIFF
--- a/src/ui/map_tests/map_webgl.test.ts
+++ b/src/ui/map_tests/map_webgl.test.ts
@@ -41,6 +41,22 @@ test('handles "webglcontextlost" when map is created without style', () => {
     map.remove();
 });
 
+test('handles "webglcontextrestored" when map is created without style', async () => {
+    // This test verifies fix for #7432 - map should not throw when WebGL context
+    // is restored but this.style is null (map initialized without style)
+    const map = createMap({deleteStyle: true});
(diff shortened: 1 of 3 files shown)


src/ui/map_tests/map_webgl.test.ts
```diff
@@ -41,6 +41,22 @@ test('handles "webglcontextlost" when map is created without style', () => {
     map.remove();
 });
 
+test('handles "webglcontextrestored" when map is created without style', async () => {
+    // This test verifies fix for #7432 - map should not throw when WebGL context
```
HarelM: I don't think this extra comment is needed, the rest looks good though, thanks!

mvanhorn:
Removed the comment in 47c3571. Thanks for the review!

HarelM (approved):

PR #7451, opened 2026-04-11
Improve ability to import scripts in workers

HarelM:
## Launch Checklist

I wanted to improve the situation with maplibre-contours since I needed it to be able to support other protocols.
In order to do that I decided to try and use the `importScriptInWorkers` API.
I was able to achieve what I wanted but found out there are two missing things that will improve ergonomics:
1. Add ability to send message to the workers. This is important since loading the script is only allowing "hardcoded" initialization, but if there are parameters that need to be passed to the worker there's no easy way to do it. global dispacher is the way to do it, this is how it's done with RTL plugin and in general the way to communicate with the workers.
2. Add `makeRequest` available in the worker. When you use `addProtocol` in one plugin but need to use this in anther plugin the `makeRequest` is the way to activate this. So this is also needed in the worker context.

Here's my process around this issue:
- https://github.com/IsraelHikingMap/Site/issues/2454

I did think about implementing the contour source in maplibre, but decided to avoid this at this point in time. 

 - [x] Link to related issues.
 - [ ] ~Write tests for all new functionality.~ only public APIs updated.

### DIFF
--- a/src/util/dispatcher.ts
+++ b/src/util/dispatcher.ts
@@ -74,10 +74,10 @@ export class Dispatcher {
 let globalDispatcher: Dispatcher;
 
 /**
- * This method is used to get the global dispatcher that is shared across all maps instances.
+ * This function is used to get the global dispatcher that is shared across all maps instances.
  * It is used by the main thread to send messages to the workers, and by the workers to send messages back to the main thread.
  * If you import a script into the worker and need to send a message to the workers to pass some parameters for example, 
- * you can use this method to get the global dispatcher and send a message to the workers.
+ * you can use this function to get the global dispatcher and send a message to the workers.
  * @returns The global dispatcher instance.
  */
 export function getGlobalDispatcher(): Dispatcher {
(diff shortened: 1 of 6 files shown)


birkskyum (approved):

src/util/dispatcher.ts
```diff
@@ -73,6 +73,13 @@ export class Dispatcher {
 
 let globalDispatcher: Dispatcher;
 
+/**
+ * This method is used to get the global dispatcher that is shared across all maps instances.
```
louwers: Nit: it's not a method.

louwers (approved):

src/util/dispatcher.ts
HarelM: Why not? `export function getGlobalDispatcher(): Dispatcher {...`

louwers: It's a free standing function right? Not a method on some class.

HarelM: I was under the impression that a method and a function are interchangeable in terms of naming. But I don't mind changing the comment to use the word function instead of method.

PR #7452, opened 2026-04-12
chore: fix warning in CI

CommanderStorm:
also enables warnings as errors for this so that this does not happen again

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -10,7 +10,7 @@ import preferTypeForDataShapes from './build/eslint-rules/prefer-type-for-data-s
 
 export default [
     {
-        ignores: ['build/*.js', 'build/rollup/**', 'staging/**', 'coverage/**', 'node_modules/**', 'docs/**', 'dist/**']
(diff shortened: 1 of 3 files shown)


eslint.config.js
```diff
@@ -10,7 +10,7 @@ import preferTypeForDataShapes from './build/eslint-rules/prefer-type-for-data-s
 
 export default [
     {
-        ignores: ['build/*.js', 'build/rollup/**', 'staging/**', 'coverage/**', 'node_modules/**', 'docs/**', 'dist/**']
+        ignores: ['build/*.js', 'build/rollup/**', 'staging/**', 'coverage/**', 'node_modules/**', 'docs/**', 'dist/**', '**/*_generated.js']
```
HarelM, line 13: Maybe generalize build folder for js files (first entry)?

HarelM (approved):

eslint.config.js
CommanderStorm: na, this is about the benchmark scripts that we codegen, not the build folder

HarelM: The first argument is: `build/*.js`... but I sure...

PR #7453, opened 2026-04-12
feat!: Drop webgl1 support

CommanderStorm:
This PR is the part of v6 that will drop the webgl1 support.
We have talked about this for a while and I don't think we have gotten a lot of pushback.

This blocks a few of the more agressive performance fixes for 3d terrain, such as not needing 2 frames every frame (-> async)

Resolves:
- #3947


test/build/sourcemaps.test.ts
```diff
@@ -80,6 +80,6 @@ describe('main sourcemap', () => {
         const s1 = setMinus(actualEntriesInSourcemapJSON, expectedEntriesInSourcemapJSON);
         expect(s1.length).toBeLessThan(5);
         const s2 = setMinus(expectedEntriesInSourcemapJSON, actualEntriesInSourcemapJSON);
-        expect(s2.length).toBeLessThan(17);
+        expect(s2.length).toBeLessThan(18);
```
CommanderStorm, line 83: @birkskyum since you touched this last, what does this mean? For me this test is a bit magical and I don't fully get what is beign tested, but likely this just me not having worked with sourcemaps 😓

HarelM:
I've added a link in the following issue, just in case:
- #6427

src/ui/default_locale.ts
```diff
@@ -24,4 +24,7 @@ export const defaultLocale = {
     'CooperativeGesturesHandler.WindowsHelpText': 'Use Ctrl + scroll to zoom the map',
     'CooperativeGesturesHandler.MacHelpText': 'Use ⌘ + scroll to zoom the map',
     'CooperativeGesturesHandler.MobileHelpText': 'Use two fingers to move the map',
+    'Map.WebGL2NotSupported.Full': 'We are sorry, but it seems that your browser does not support WebGL, a technology for rendering 3D graphics on the web.',
```
birkskyum: It should say WebGL2 instead of just webgl

src/ui/default_locale.ts
```diff
@@ -24,4 +24,7 @@ export const defaultLocale = {
...
+    'Map.WebGL2NotSupported.Short': 'WebGL is required to display this map.',
```
CommanderStorm:
```suggestion
    'Map.WebGL2NotSupported.Full': 'We are sorry, but it seems that your browser does not support WebGL2, a technology for rendering 3D graphics on the web.',
    'Map.WebGL2NotSupported.Short': 'WebGL2 is required to display this map.',
```

src/ui/default_locale.ts
```diff
@@ -27,4 +27,5 @@ export const defaultLocale = {
     'Map.WebGL2NotSupported.Full': 'We are sorry, but it seems that your browser does not support WebGL2, a technology for rendering 3D graphics on the web.',
     'Map.WebGL2NotSupported.Short': 'WebGL2 is required to display this map.',
     'Map.WebGL2NotSupported.LearnMore': 'Read more',
+    'Map.WebGL2NotSupported.LearnMoreUrl': 'https://wiki.openstreetmap.org/wiki/This_map_requires_WebGL',
```
birkskyum: This page only speak of "WebGL" being required - not a single reference to webgl2. I don't know if we best make a separate page for webgl2, or if a note on the existing page about  maplibre v6+.

CommanderStorm: Once we upgrade OSM I will have to upgrade that wiki anyhow. So should be the require a different docs.

Alternatively, we can just have a dup of that, but it would likely end up being a dup..

HarelM:
@CommanderStorm can you move this PR from draft to ready for review and resolve conflicts?
I think we can push it forward now.

src/util/test/null_gl.ts
HarelM, line 1: This used to be covered with a package, is this no longer the case?
If so, can we remove the package?
Can we move this code into the setup_mock_webgl2 instead of here?
Does vitest have a better looking way to create a mock object? maybe a mock proxy or something?
Bottom line, this file can be improved.

src/ui/map.ts
```diff
@@ -3465,26 +3468,52 @@ export class Map extends Camera {
...
-                throw new Error(msg);
-            }
+            this._showWebGL2Error(webglcontextcreationerrorDetailObject);
+            return;
         }
 
         this.painter = new Painter(gl, this.transform);
     }
 
+    _showWebGL2Error(webglcontextcreationerrorDetailObject: any) {
```
HarelM: I tend to think this is an overkill for this library.
Each library should decide how the UI for this should look.
I would advise to keep it as it was, if there's a hard requirement to add this, let's understand why raising the event is not enough.

src/geo/projection/globe_projection_error_measurement.ts
```diff
@@ -175,55 +172,44 @@ export class ProjectionErrorMeasurement {
...
-                sync,
-            };
-        } else {
-            // Read it back later.
-            this._readbackQueue = {
-                frameNumberIssued: this._updateCount,
-                sync: null,
-            };
-        }
+        // Read back into PBO
```
HarelM, line 175: The previous code checked that `_pbo` is defined, should this check be added back?

src/css/maplibre-gl.css
```diff
@@ -905,3 +905,39 @@ a.maplibregl-ctrl-logo.maplibregl-compact {
     left: 0 !important;
     z-index: 99999;
 }
+
+.maplibregl-webgl-error {
```
HarelM: See my other comment about adding UI for this case.

src/ui/map.ts
CommanderStorm: the code a user for this currently has to write is fairly non-obvious if they want to catch this.
I doubt that the average does this because of this.
I would think that if we have an error message this would be much better.
Plus, if a user does not like the way the error is worded, they can either
- localise it
- override this method via subclassing

Here is the code that is nessary currently, which is not great.
I would think that if we had this in the library, this would be a better average expericece.

https://github.com/openstreetmap/openstreetmap-website/blob/16761f19d731fc9b6f5c9bdf7167829dd3fbe6d0/app/assets/javascripts/maplibre/map.js#L33

CommanderStorm: Also, now this works via the regular event firing mechanism, so does not need the weird constructor hacking that I did.

src/util/test/null_gl.ts
CommanderStorm: > is this no longer the case

Yes, the package we have covers webgl, but not webgl2
Given that the lversion is adtive, we need both

> If so, can we remove the package?

I can inline said package into this file, but we need both currently.
We can move to luma.gls' null_gl adapter once we move there, this is modeled after this a bit

> Can we move this code into the setup_mock_webgl2 instead of here?

Could you give me a link to what that is?
It does not exist in google or this repo.

> Does vitest have a better looking way to create a mock object? maybe a mock proxy or something?

vitest does not have a gpu adapter with all of the wegl2 constants that we need or at leas I did not find any.

src/geo/projection/globe_projection_error_measurement.ts
CommanderStorm: `_pbo` should now be always defined given that we alwayxs have webgl2 enabled.
It is not avaliable after destroy, but that case I think also means that we are not rendering any more.

src/css/maplibre-gl.css
CommanderStorm: not actionable

src/ui/map.ts
CommanderStorm: | before | now |
|--------|--------|
| image | image |

HarelM: If you would like to fire a specific error using an error class I'll fully support this, but let's avoid adding design related charges as part of this PR. This applies to the CSS as well. 
We can decide if we want to introduce this not as part of this change. 
Let's keep this PR only about removal of webgl1.

CommanderStorm: I mean I can split it into two if you insist, but this is a fairly small change, 10 lines or so.
Should be fine.

It is also very much closely related.
I think it is a better default given that most users will not customise this, so we should have a better default.
And we likely should have this given that we will now fail on IE and such..

src/util/test/null_gl.ts
CommanderStorm: > Can we move this code into the setup_mock_webgl2 instead of here?

setup_mock_webgl2 is for our unit tests only, it is not for the render tests and such, so lets keep the two separate

src/ui/map.ts
HarelM: If you include the CSS it's a bit more than 10 lines, and I'm not fully on board this is really needed, especially if you take into consideration all kind of responsiveness issues, shadow DOM etc.
I really don't want to get into that rabbit hole, let's just raise an event. If needed, create an error class for it.

src/css/maplibre-gl.css
```diff
@@ -906,3 +906,39 @@ a.maplibregl-ctrl-logo.maplibregl-compact {
...
+
+@media (width <= 480px) {
+    .maplibregl-webgl-error .maplibregl-webgl-error-full {
+        display: none;
+    }
+
+    .maplibregl-webgl-error .maplibregl-webgl-error-short {
+        display: block;
+    }
+}
```
CommanderStorm:
```suggestion
```

src/ui/map.ts
```diff
@@ -3466,26 +3469,52 @@ export class Map extends Camera {
...
+        fullLink.textContent = learnMore;
+
+        const shortMsg = DOM.create('p', 'maplibregl-webgl-error-short', errorDiv);
+        shortMsg.textContent = `${shortText} `;
+        const shortLink = DOM.create('a', undefined, shortMsg);
+        shortLink.href = webglLink;
+        shortLink.target = '_blank';
+        shortLink.rel = 'noopener noreferrer';
+        shortLink.textContent = learnMore;
+
```
```suggestion
```

src/ui/default_locale.ts
```diff
@@ -24,4 +24,8 @@ export const defaultLocale = {
     'CooperativeGesturesHandler.WindowsHelpText': 'Use Ctrl + scroll to zoom the map',
     'CooperativeGesturesHandler.MacHelpText': 'Use ⌘ + scroll to zoom the map',
     'CooperativeGesturesHandler.MobileHelpText': 'Use two fingers to move the map',
+    'Map.WebGL2NotSupported.Full': 'We are sorry, but it seems that your browser does not support WebGL2, a technology for rendering 3D graphics on the web.',
+    'Map.WebGL2NotSupported.Short': 'WebGL2 is required to display this map.',
+    'Map.WebGL2NotSupported.LearnMore': 'Read more',
+    'Map.WebGL2NotSupported.LearnMoreUrl': 'https://wiki.openstreetmap.org/wiki/This_map_requires_WebGL',
```
CommanderStorm:
```suggestion
```

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
...
+- ⚠️ WebGL (v1) support has been removed; WebGL2 is now required.
+  In practical terms, this will not change how you interact with the map.
+  This enables performance improvements (e.g. line opacity), Terrain3D enhancements, and several bug fixes.
+  WebGL2 support has been widely available for years, and usage of the legacy path had plateaued, so maintaining it no longer justified the added complexity.
+  To ease this breaking change, we have aloso refactored how we handle the case that no webgl is avaliableWebGL (e.g. due to browser restrictions).
+  The fallback message can be customized via our i18n system, the `.on("error")` handler, or by overriding `Map._showWebGL2Error`.
```
```suggestion
  The fallback message can be customized via the `.on("error")` handler, or by overriding `Map._showWebGL2Error`.
```

src/ui/map.ts
CommanderStorm: I don't think this is a good idea, but if this is what it takes to get some progress..

This will make it much more confusing for the regular users why the map is not working...

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
...
+- ⚠️ WebGL (v1) support has been removed; WebGL2 is now required.
+  In practical terms, this will not change how you interact with the map.
+  This enables performance improvements (e.g. line opacity), Terrain3D enhancements, and several bug fixes.
+  WebGL2 support has been widely available for years, and usage of the legacy path had plateaued, so maintaining it no longer justified the added complexity.
+  To ease this breaking change, we have aloso refactored how we handle the case that no webgl is avaliableWebGL (e.g. due to browser restrictions).
+  The fallback message can be customized via the `.on("error")` handler, or by overriding `Map._showWebGL2Error`.
```
```suggestion
  To ease this breaking change, we have aloso refactored how we handle the case that no webgl is avaliableWebGL (e.g. due to browser restrictions).
  You can now listen to the webgl error via `.on("error")`, or by overriding `Map._showWebGL2Error`.
```

src/util/test/null_gl.ts
HarelM: Isn't this part only used in unit tests? I'm confused now...
I meant this file (sorry for the wrong order of words): `test/unit/lib/webgl2_mock_setup.ts`

src/render/painter.test.ts
```diff
@@ -21,7 +22,7 @@ describe('render', () => {
     };
 
     beforeEach(() => {
-        const gl = document.createElement('canvas').getContext('webgl');
+        const gl = createNullGL();
```
HarelM, line 25: wouldn't changing this to `document.createElement('canvas').getContext('webgl2')` will use webgl2_mock_setup.ts so that there's no need to do anything "special" here and simply use "regular code"?

src/ui/map.ts
```diff
@@ -3466,26 +3469,29 @@ export class Map extends Camera {
...
         this.painter = new Painter(gl, this.transform);
     }
 
+    _showWebGL2Error(webglcontextcreationerrorDetailObject: any) {
+        const msg = 'Failed to initialize WebGL';
+        if (webglcontextcreationerrorDetailObject) {
+            webglcontextcreationerrorDetailObject.message = msg;
+        }
+
+        this.fire(new ErrorEvent(new Error(
```
HarelM: The previous code threw an exception instead of fire an error event, is this change intentional? Not that I mind, I think both are good options, just wanted to make sure this wasn't overlooked.

src/ui/map.ts
```diff
@@ -3466,26 +3469,29 @@ export class Map extends Camera {
...
-            }
+            this._showWebGL2Error(webglcontextcreationerrorDetailObject);
+            return;
         }
 
         this.painter = new Painter(gl, this.transform);
     }
 
+    _showWebGL2Error(webglcontextcreationerrorDetailObject: any) {
+        const msg = 'Failed to initialize WebGL';
```
HarelM: There's no harm in improving this message string if you think more information might be helpful here. Your call, I'm good either way.

src/render/painter.test.ts
CommanderStorm: I would prefer to not rely on magic mocks if possible.
If I do `document.createElement('canvas').getContext('webgl2')` it is less clear what code actually runs and why it does not behave the same way as the real thing (all methods are that of an null-device)

HarelM: Fair enough, can we avoid using `webgl2_mock_setup.ts` completely? I would prefer to avoid using it altogether since it will allow running tests with `npx vitest some-parameters`, which is currently not possible due to the extra vitest configuration.
This can be out of scope for this PR, but it would be nice to simplify this.
I agree about the magic stuff, but I prefer to have a consistent solution across the tests...

CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
...
+  WebGL2 support has been widely available for years, and usage of the legacy path had plateaued, so maintaining it no longer justified the added complexity.
+  To ease this breaking change, we have aloso refactored how we handle the case that no webgl is avaliableWebGL (e.g. due to browser restrictions).
+  You can now listen to the webgl error via `.on("error")`, or by overriding `Map._showWebGL2Error`.
+  See [caniuse.com/webgl2](https://caniuse.com/webgl2) for ecosystem support and [our RFC for details](https://github.com/maplibre/maplibre-gl-js/discussions/6017). ([#7453](https://github.com/maplibre/maplibre-gl-js/pull/7453)) (by [@CommanderStorm](https://github.com/CommanderStorm))
+- Expose `getProjectionData` function in custom layer args objects ([#7471](https://github.com/maplibre/maplibre-gl-js/pull/7471)) (by [@kubapelc](https://github.com/kubapelc))
+- Marked package `sideEffects` as CSS-only in package metadata, which may improve tree-shaking and reduce bundle size in some bundlers ([#7258](https://github.com/maplibre/maplibre-gl-js/pull/7258)) (by [@CommanderStorm](https://github.com/CommanderStorm))
```
```suggestion
- ⚠️ WebGL (v1) support has been removed; WebGL2 is now required.
  In practical terms, this will not change how you interact with the map.
  This enables performance improvements (e.g. line opacity), Terrain3D enhancements, and several bug fixes.
  WebGL2 support has been widely available for years, and usage of the legacy path had plateaued, so maintaining it no longer justified the added complexity.
  To ease this breaking change, we have aloso refactored how we handle the case that no webgl is avaliableWebGL (e.g. due to browser restrictions).
  You can now listen to the webgl error via `.on("error")`, or by overriding `Map._showWebGL2Error`.
  See [caniuse.com/webgl2](https://caniuse.com/webgl2) for ecosystem support and [our RFC for details](https://github.com/maplibre/maplibre-gl-js/discussions/6017). ([#7453](https://github.com/maplibre/maplibre-gl-js/pull/7453)) (by [@CommanderStorm](https://github.com/CommanderStorm))
```

src/webgl/webgl2.ts
HarelM, line 1: GREAT! Thanks!!

test/examples/check-if-webgl-is-supported.html
```diff
@@ -17,28 +17,28 @@
 <div id="map"></div>
 
 <script>
-    function isWebglSupported() {
-        if (window.WebGLRenderingContext) {
+    function isWebgl2Supported() {
```
HarelM: Should we change this example to use the new error event when creating the map instead?
Or have both in this example to show how one could decide which one to use?

HarelM:
I've added a few last comments, assuming the tests pass I think this can be merged.
I've left only the relevant comments unresolved and resolved the rest.

src/ui/map.ts
CommanderStorm: firing an exception is basically the same thing as throwing, just that firigng can be caught by the users -> better ergonomic.

If the user does not catch this, we raise anyhow

test/examples/check-if-webgl-is-supported.html
CommanderStorm: likely, done

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 ## main
 ### ✨ Features and improvements
+- ⚠️ WebGL (v1) support has been removed; WebGL2 is now required.
+  In practical terms, this will not change how you interact with the map.
+  This enables performance improvements (e.g. line opacity), Terrain3D enhancements, and several bug fixes.
+  WebGL2 support has been widely available for years, and usage of the legacy path had plateaued, so maintaining it no longer justified the added complexity.
+  To ease this breaking change, we have aloso refactored how we handle the case that no webgl is avaliableWebGL (e.g. due to browser restrictions).
```
```suggestion
  To ease this breaking change, we have also refactored how we handle the case that no webgl is avaliableWebGL (e.g. due to browser restrictions).
```

HarelM (approved):

PR #7468, opened 2026-04-13
fix: make sure that terrain and globe have correct Popup placement

### LINKED ISSUE #7467: Popup and map switch terrain -> globe and viceversa
**maplibre-gl-js version**: 5.23

**browser**: chrome, firefox, edge

When a popup is opened on a map, switching map type doesn't keep the popup at the correct latlng

https://codepen.io/maxargy/pen/emdQLBW

 1. Open the pen sample
 2. click the marker to open the popup
 3. change map type and the popup is far from the marker

Error seems to be bigger when the marker and popup is far from map center-view

CommanderStorm:
"Just works" now.
Resolves https://github.com/maplibre/maplibre-gl-js/issues/7467

image

### DIFF
--- a/src/ui/popup.test.ts
+++ b/src/ui/popup.test.ts
@@ -1,5 +1,5 @@
 import {describe, beforeEach, test, expect, vi} from 'vitest';
-import {createMap as globalCreateMap, beforeMapTest} from '../util/test/util';
+import {createMap as globalCreateMap, beforeMapTest, createTerrain} from '../util/test/util';
 import {Popup, type Offset} from './popup';
 import {LngLat} from '../geo/lng_lat';
 import Point from '@mapbox/point-geometry';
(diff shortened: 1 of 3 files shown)


src/ui/popup.test.ts
```diff
@@ -1069,4 +1069,43 @@ describe('popup', () => {
...
+        await map.once('load');
+
+        const popup = new Popup()
+            .setLngLat(new LngLat(20, 30))
+            .setText('Test')
+            .addTo(map);
+
+        expect(popup.getElement().style.transform).toBe('translate(-100%,0) translate(1075px,-187px)');
+
+        map.addSource('terrain', {type: 'raster-dem', tiles: ['http://example.com/{z}/{x}/{y}.png']});
```
HarelM, line 1104: The marker tests are simply calling `map.fire('terrain')`, which I think is simpler than all this, isn't it?

HarelM: That might not more the popup. Nevermind.

HarelM (approved):

src/ui/popup.test.ts
CommanderStorm: this needs the kind of setup because we need this to have a height -> change the location.

PR #7471, opened 2026-04-14
fix: ProjectionDataParams no longer requires OverscaledTileID instance

kubapelc:
`transform.getProjectionData()` is part of the public API that is useful when rendering completely custom tiles, but the `ProjectionDataParams` object it accepts as parameter requires an instance of `OverscaledTileID`, which is only exported as a type by MapLibre. This makes it cumbersome to use, since there is no easy way of creating an `OverscaledTileID` instance.

This PR modifies the `ProjectionDataParams` object to also accepts an object literal, in addition to an `OverscaledTileID` instance, making it easy to use by external code. `getProjectionData` implementations are also modified to accept this object literal.

There doesn't seem to be a preexisting issue for this, and it was a simple fix, so I went ahead and implemented it.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
Transform is not really a part of the public API.
Can you elaborate on what you would like to achieve and maybe expose something else?

kubapelc:
Hi! My original intention when implementing globe was for `getProjectionData` to be available to users. The reason is that this function is essential for making complex custom layers that work both on mercator and globe and also work seamlessly with the rest of MapLibre. There is even an example on how it can be used to make such a layer:

https://maplibre.org/maplibre-gl-js/docs/examples/add-a-custom-layer-with-tiles-to-a-globe/

Scope of this PR is only to fix the function's parameter types to make sure it can be used properly without creating `OverscaledTileID` instances. I think `OverscaledTileID` class was fully exported back when globe was implemented, but since has changed to only be exported as a type.

HarelM:
I don't think it was ever exposed as is, but that's besides the point.
Custom layer has an interface that we can add data to if needed, like the matrices. 
If you need this as a public API you need to expose it from the map object. 
But if it can only really be used in a custom layer I think it will be better to expose it there only.

kubapelc:
Okay, I see no harm in exposing `getProjectionData` properly, though my intention was only to fix a minor bug. I've added it to custom layer args in the latest commit, and tweaked the example I linked earlier to use it. The function is only useful for custom layers, and is only needed during rendering of tiles, so I think that custom layer args is a better fit than expanding the map's API. Let me know what you think!

src/webgl/draw/draw_custom.ts
```diff
@@ -27,6 +27,7 @@ export function drawCustom(painter: Painter, tileManager: TileManager, layer: Cu
             define: projection.shaderDefine,
         },
         defaultProjectionData: projectionData,
+        getProjectionData: transform.getProjectionData.bind(transform),
```
HarelM: Use lambda instead of `bind` please.

src/style/style_layer/custom_style_layer.ts
```diff
@@ -105,6 +105,15 @@ export type CustomRenderMethodInput = {
      * or more accurately for globe, elevation above the surface of the perfect sphere used to render the planet.
      */
     defaultProjectionData: ProjectionData;
+
+    /**
+     * Generates a `ProjectionData` instance to be used while rendering a given tile.
+     * In custom layers, this function is only needed when rendering tiles in a completely custom way and with shaders that are compatible with both projections.
+     *
+     * Please see the example "Add a custom layer with tiles to a globe" for more details.
```
HarelM: Use a link here and `@see` comment syntax.

src/style/style_layer/custom_style_layer.ts
```diff
@@ -105,6 +105,15 @@ export type CustomRenderMethodInput = {
...
+     * @param params - Parameters for the projection data generation.
+     */
+    getProjectionData: (params: ProjectionDataParams) => ProjectionData;
```
HarelM: I think a better approach would be to expose something that isn't "familiar" with `OverscaleTileId` at all and do the conversion when this method is used instead of make `getProjectionData in the transform class be aware of this new type of input.
I think it will be cleaner.

src/webgl/draw/draw_custom.ts
HarelM: Given my other method, this can be a small conversion method from `OverscaledTileIDLike` to `OverscaledTileID`, which will remove the `bind`.

src/geo/projection/projection_data.ts
```diff
@@ -54,7 +54,15 @@ export type ProjectionDataParams = {
     /**
      * The ID of the current tile
      */
-    overscaledTileID: OverscaledTileID | null;
+    overscaledTileID: OverscaledTileID | null | {
```
HarelM: Use `overscaledTileIDLike` type defined as the below parameters.

src/geo/projection/mercator_transform.ts
```diff
@@ -728,12 +728,26 @@ export class MercatorTransform implements ITransform {
 
     getProjectionData(params: ProjectionDataParams): ProjectionData {
         const {overscaledTileID, aligned, applyTerrainMatrix} = params;
-        const mercatorTileCoordinates = this._helper.getMercatorTileCoordinates(overscaledTileID);
-        const tilePosMatrix = overscaledTileID ? this.calculatePosMatrix(overscaledTileID, aligned, true) : null;
+        let tile: OverscaledTileID;
+        if (!overscaledTileID) {
+            tile = null;
+        } else if (overscaledTileID instanceof OverscaledTileID) {
```
HarelM: Revert this and do the conversion only for custom layer.

HarelM:
Sorry for the delay in review.
I think this approach is good, I've added some comments around where to do the conversion.

src/webgl/draw/draw_custom.ts
kubapelc: Done

src/style/style_layer/custom_style_layer.ts
kubapelc: Done. I'm not entirely sure I used the `@see` syntax correctly, but it seems to line up with examples I found online.

src/style/style_layer/custom_style_layer.ts
kubapelc: I agree, done.

src/geo/projection/projection_data.ts
kubapelc: I've opted for a slightly reduced type in the end without the `overscaledZ` field, since it does not affect the output matrices.

src/geo/projection/mercator_transform.ts
kubapelc: Done

src/geo/projection/projection_data.ts
```diff
@@ -68,3 +68,4 @@ export type ProjectionDataParams = {
      */
     applyGlobeMatrix?: boolean;
 };
+
```
HarelM: Probably better to revert this change, to reduce changes that are part of the PR...

src/style/style_layer/custom_style_layer.ts
```diff
@@ -105,6 +149,15 @@ export type CustomRenderMethodInput = {
      * or more accurately for globe, elevation above the surface of the perfect sphere used to render the planet.
      */
     defaultProjectionData: ProjectionData;
+
+    /**
+     * Generates a {@link ProjectionData} instance to be used while rendering a given tile.
+     * In custom layers, this function is only needed when rendering tiles in a completely custom way and with shaders that are compatible with both projections.
+     *
+     * @see {@link https://maplibre.org/maplibre-gl-js/docs/examples/add-a-custom-layer-with-tiles-to-a-globe/ | the example "Add a custom layer with tiles to a globe"} for more details
```
HarelM:
```suggestion
     * @see [Add a custom layer with tiles to a globe](https://maplibre.org/maplibre-gl-js/docs/examples/add-a-custom-layer-with-tiles-to-a-globe)
```

src/geo/projection/mercator_transform.test.ts
```diff
@@ -616,4 +617,21 @@ describe('transform', () => {
         expect(transform.getCameraLngLat().lng).toBeCloseTo(camLngLat.lng, 10);
         expect(transform.getCameraLngLat().lat).toBeCloseTo(camLngLat.lat, 10);
     });
+
+    describe('getProjectionData', () => {
```
HarelM, line 621: Is this test still relevant?

HarelM:
Added a few last comments, looks good otherwise.

src/geo/projection/mercator_transform.test.ts
kubapelc: It tests getProjectionData in general. I think there is no harm in keeping it.

src/geo/projection/projection_data.ts
kubapelc: Fixed

src/geo/projection/mercator_transform.test.ts
HarelM: Fair point.

HarelM:
Seems like the coverage is missing for the getProjectionData you added for the custom layer.
How complicated it would be to add a test for that method/usage?
If it's too complicated consider avoiding it.

HarelM (approved):

kubapelc:
Hi, I've added some tests for getProjectionData in custom layer args, and rebased the branch on latest main.

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## main
 ### ✨ Features and improvements
+- Expose `getProjectionData` function in custom layer args objects ([#7471](https://github.com/maplibre/maplibre-gl-js/pull/7471))
```
```suggestion
- Expose `getProjectionData` function in custom layer args objects ([#7471](https://github.com/maplibre/maplibre-gl-js/pull/7471)) (by [@kubapelc]((https://github.com/kubapelc))
```

HarelM:
One last comment before I merge. Looks good, thanks!

HarelM:
I was thinking about starting version 6 pre releases just now. Is this urgent so you'd need it in version 5 or can it wait for version 6?

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## main
 ### ✨ Features and improvements
+- Expose `getProjectionData` function in custom layer args objects ([#7471](https://github.com/maplibre/maplibre-gl-js/pull/7471)) (by [@kubapelc]((https://github.com/kubapelc))
```
```suggestion
- Expose `getProjectionData` function in custom layer args objects ([#7471](https://github.com/maplibre/maplibre-gl-js/pull/7471)) (by [@kubapelc](https://github.com/kubapelc))
```

HarelM: extra "(" removed...

kubapelc:
It is not urgent, it can wait for version 6.

HarelM:
Ok, cool, thanks for the update. I'll merge this now nevertheless...

PR #7475, opened 2026-04-15
chore: migrate benchmarks to Mapterhorn and OFM

CommanderStorm:
I am having weird, semi-intermitent issues with maptiler and auth, particularly in the terrain3d parts.

This migrates this parts of our benchmarks away since this seems less reliable.
Not sure if there are better options to this


HarelM (approved):
Thanks!

Issue #7419, opened 2026-04-06
Latitude precision issue in Globe Projection with Mali GPUs


tavimori:
**maplibre-gl-js version**: 5.22.0

**browser**: any phone with Mali GPU?

### Steps to Trigger Behavior

 1. Mali GPU, zoom around 11, globe projection, latitude around 30.9 N.
 2. Open the [demo links](https://jsbin.com/zenatuk/edit?html,output) on phones with Mali GPU.
 3. The y-direction projection has a precision issue at zoom around 11 under the globe projection. Some distinct latitude values will overlap (see the image below); this applies to both custom layers and map tiles. 

BTW, I noticed that there is a discussion on GPU "atan" error correction in https://github.com/maplibre/maplibre-gl-js/blob/main/developer-guides/globe.md, not sure if related.

### Link to Demonstration

https://jsbin.com/zenatuk/edit?html,output
(modified based on the official example https://maplibre.org/maplibre-gl-js/docs/examples/add-a-custom-layer-with-tiles-to-a-globe/ )

### Expected Behavior

(Screenshot on Xiaomi 13, Adreno 740)
<img width="300" alt="Image" src="https://github.com/user-attachments/assets/99bfaf9f-bab5-4620-a339-76b2cb79168b" />

(Screenshot on iPhone 16e, Apple A18)
<img width="300" alt="Image" src="https://github.com/user-attachments/assets/b652305b-0426-475e-a761-a8b03d06640b" />

### Actual Behavior

(Screenshot on Huawei Nova 13, Mali-G610?)
<img width="300" alt="Image" src="https://github.com/user-attachments/assets/b97df8c7-0dc5-42aa-90d3-312da8d18777" />

We also tested on the Oppo Find N2 Flip (PGT110) with Mali-G710, which has the same issue.

tavimori:
## The root cause: the precision of `sine` and `cosine`.

I can confirm this is an issue with `sin` precision on certain devices.
A device test page has been created: https://jsbin.com/yamahed
Most GPUs handle `sin` and `cos` with `highp` precision correctly, which uses 32 bits (the Earth's surface-level resolution is just a few meters).

However, for Mali-G710 and Mali-G610, the error can vary by up to 200m. Which is significantly above the 32bits precision.

## Potential solution

Having said that above, I also want to discuss a potential fix that can skip all trigonometric computation when handling latitudes. (and I noticed that the existing `atan` fix may also not be necessary anymore after the fix). 

[...]

PR #7478, opened 2026-04-16
perf(terrain): use hardware based interpolation for terrain height (-22/33%)

CommanderStorm:
## Launch Checklist

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!n.

Seems to be 22-33% (22% when run under vulcan on an 1660TI and 33% under Switftshader) faster in the benchmark that I submitted in the other PR.

When I added GPU probes around this, it was 

Not sure if there is a simpler way to do this

Opus helped me..
Needs graphics review if there is a better way to do this thing with the textures.. still seems a bit hacky

<details><summary>Terrain3D Performance Model as analaysed by Opus</summary>

## Architecture Overview

Terrain3D works by rendering all 2D map layers (fill, line, background, raster, hillshade) into per-tile **Render-to-Texture (RTT)** framebuffers, then draping those textures onto displaced terrain meshes.
Layers that can't be draped (symbols, circles) render directly to screen with depth-based occlusion against the terrain.

### Render Passes (per frame)

```
1. Depth Pre-Pass         → terrain mesh → RGBA8 depth FBO (shared with coords)
2. RTT Layer Rendering    → style layers → per-tile FBO pool (RGBA, tileSize × qualityFactor)
3. Terrain Composite      → terrain mesh + RTT textures → screen
4. Translucent Pass       → symbols, circles → screen (samples depth texture for occlusion)
```

The depth and coords passes share a single FBO and swap RGBA8 color attachments.
The coords pass is lazy (only runs on mouse interaction).

### Terrain Mesh

- **Resolution:** 128×128 grid = 32,768 triangles per tile (hardcoded `meshSize: 128`)
- **Vertex format:** `vec3 a_pos3d` (xy = tile position, z = skirt flag)
- **Mesh reuse:** One mesh instance shared across all tiles (vertex positions are tile-local, projection handled by uniforms)

### RTT Pool

- **Pool size:** 30 FBO objects (RGBA, `tileSize × qualityFactor` resolution, default 512×2=1024px)
- **Caching:** Tiles have a fingerprint stamp; cached tiles skip re-rendering
- **Eviction:** When pool is full, all current RTT tiles are composited to screen and the pool is freed
- **Layers rendered to texture:** background, fill, line, raster, hillshade, color-relief
- **Layers NOT rendered to texture:** symbol, circle, fill-extrusion, heatmap, custom

### Elevation Sampling (before optimization)

Every terrain vertex shader called `get_elevation()` which performed **manual bilinear interpolation** of the DEM texture — 4 dependent texture fetches per vertex, each decoding RGB to elevation:

```glsl
// OLD: 4 texture fetches + decode per vertex
float get_elevation(vec2 pos) {
    vec2 coord = (u_terrain_matrix * vec4(pos, 0.0, 1.0)).xy * u_terrain_dim + 1.0;
    vec2 f = fract(coord);
    vec2 c = (floor(coord) + 0.5) / (u_terrain_dim + 2.0);
    float d = 1.0 / (u_terrain_dim + 2.0);
    float tl = ele(c);                    // texture fetch 1 + RGB decode
    float tr = ele(c + vec2(d, 0.0));     // texture fetch 2 + RGB decode
    float bl = ele(c + vec2(0.0, d));     // texture fetch 3 + RGB decode
    float br = ele(c + vec2(d, d));       // texture fetch 4 + RGB decode
    return mix(mix(tl, tr, f.x), mix(bl, br, f.x), f.y) * u_terrain_exaggeration;
}
```

This manual bilinear existed because the DEM was stored as RGB-encoded elevation (mapbox/terrarium format). Hardware `gl.LINEAR` would interpolate the raw encoded RGB channels before decoding, producing incorrect elevation values at encoding boundaries.

### Depth Encoding

The depth pass stores `gl_Position.z / gl_Position.w` (NDC) packed into RGBA8 via:
```glsl
vec4 pack(float value) {
    vec4 comp = fract(value * vec4(256³, 256², 256, 1));
    comp -= comp.xxyz * vec4(0, 1/256, 1/256, 1/256);
    return comp;
}
```

Consumers (symbol/circle vertex shaders) unpack via `dot(color, 1.0 / vec4(256³, 256², 256, 1))`.

### What changed

Decode the RGB-encoded DEM elevation to float on the CPU at tile upload time. Store as an **R16F texture with `gl.LINEAR` filtering**. The vertex shader becomes a single hardware-interpolated fetch:

```glsl
// NEW: 1 hardware-bilinear texture fetch, zero decode math
float get_elevation(vec2 pos) {
    vec2 coord = (u_terrain_matrix * vec4(pos, 0.0, 1.0)).xy * u_terrain_dim + 1.0;
    vec2 uv = (coord + 0.5) / (u_terrain_dim + 2.0);
    return texture(u_terrain, uv).r * u_terrain_exaggeration;
}
```

The `+0.5` offset is critical: `gl.LINEAR` maps UV to texel-space as `uv * textureSize - 0.5`, so `(coord + 0.5) / stride` produces the correct texel-center alignment matching the original manual bilinear.

### Files changed

- **`src/render/terrain.ts`** — `_updateElevationTexture()`: decodes DEM RGBA8 to Float32Array, uploads as R16F. `_elevationTextures` cache keyed by DEM tile UID. Empty fallback is also R16F. `getTerrainData()` returns R16F texture instead of RGBA8.
- **`src/shaders/glsl/_prelude.vertex.glsl`** — `get_elevation()` simplified to single `texture()` call. `ele()` function removed. `u_terrain_unpack` uniform no longer used in shader (kept in type system for compatibility).

### Why R16F

- R16F is **filterable by default in WebGL2** (no extension needed for `gl.LINEAR`)
- `EXT_color_buffer_half_float` already enabled in `context.ts`
- R16F precision (~3 decimal digits) is sufficient for visual terrain rendering
- Upload uses `Float32Array` with `gl.FLOAT` type — driver converts to half-float

### Precision

R16F has ~10-bit mantissa.
For elevation range ±32768m, worst-case precision is ~32m at the extremes, ~0.1m near sea level.
The original RGBA8 encoding had ~24-bit effective precision.
For visual rendering at typical map zoom levels, R16F artifacts are not visible.
CPU-side elevation queries (`DEMData.get()`) still use full precision from the original data.

### Setup

- **GPU:** NVIDIA GeForce GTX 1660 SUPER, driver 590.48.01
- **Renderer:** Headless Chrome, Vulkan/ANGLE backend
- **Scene:** OpenFreeMap Liberty style + Mapterhorn DEM, Alps (10.5°E, 46.9°N), zoom 12, pitch 60°, 1024×768, camera rotating continuously
- **Tiles:** 9-11 renderable terrain tiles, 79 style layers per tile
- **Instrumentation:** `EXT_disjoint_timer_query_webgl2` per-pass and per-tile

### GPU Time Breakdown (before optimization)

| Pass                                       | Per Tile | × Tiles | Total        | % of Frame |
|--------------------------------------------|----------|---------|--------------|------------|
| **Composite** (terrain mesh → screen)      | 0.171 ms | × 9     | **1.54 ms**  | 46%        |
| **Depth** (terrain mesh → depth FBO)       | 0.076 ms | × 9     | **0.68 ms**  | 20%        |
| **RTT layers** (style layers → tile FBOs)  | —        | —       | **~0.37 ms** | 11%        |
| **Translucent** (symbols/circles → screen) | —        | —       | **0.17 ms**  | 5%         |
| **Overhead** (FBO setup, clears, state)    | —        | —       | **~0.6 ms**  | 18%        |
| **Total**                                  |          |         | **~3.4 ms**  | 100%       |

### Per-Tile Improvement (GPU timer queries)

| Pass               | Before (4-tap RGBA8) | After (R16F LINEAR) | Speedup  |
|--------------------|----------------------|---------------------|----------|
| **composite_tile** | 0.171 ms             | 0.031 ms            | **5.5×** |
| **depth_tile**     | 0.076 ms             | 0.031 ms            | **2.5×** |

### E2E Benchmark (same benchmark on both branches)

| Backend                               | Before  | After   | Improvement |
|---------------------------------------|---------|---------|-------------|
| **SwiftShader** (software rasterizer) | 59.2 ms | 41.0 ms | **-31%**    |
| **GPU** (Vulkan, GTX 1660 SUPER)      | 3.50 ms | 2.72 ms | **-22%**    |

### What drives the improvement

- **4 vertex texture fetches → 1**: Eliminates 75% of vertex texture lookups
- **Zero decode math**: No more `rgb * 255 * unpack` per fetch
- **Hardware bilinear**: GPU texture units do interpolation in fixed-function hardware, faster than shader ALU
- **Better cache behavior**: 1 coherent fetch vs 4 scattered fetches per vertex

### RTT Caching Effectiveness

In steady-state (camera rotating), the RTT cache hit rate is **100%** (9/9 tiles cached). Layer rendering cost drops to near zero. The RTT cache only invalidates when:
- Source tile data changes (new vector tiles loaded)
- Zoom level changes
- Pool overflow (>30 uncached tiles forces flush)

### 1. Reduce mesh resolution (128×128 → 64×64)

Reduce `meshSize` from 128 to 64: 32K → 8K triangles per tile. 4× fewer vertices, 4× fewer texture fetches.

**Impact:** ~4× reduction in vertex processing. May show tessellation artifacts on steep terrain at high zoom levels.

**Trade-off:** Visual quality on steep slopes. Could use adaptive LOD (lower resolution for distant/flat tiles, higher for near/steep).

### 2. Skip depth pass when no occludable layers are visible

The depth pass only exists for `calculate_visibility()` used by symbol and circle layers. If the style has no visible symbol/circle layers, the depth pass is pure waste.

**Impact:** Eliminates ~20% of frame time for styles without symbols on terrain.

**Implementation:** Check layer visibility before `maybeDrawDepth()`.

### 3. Instanced rendering for terrain tiles

Currently: 9 separate `drawElements()` calls per pass, each with uniform uploads and state changes. With instancing: 1 draw call with per-instance tile matrices in a UBO or SSBO.

**Impact:** Reduces draw call overhead and CPU-side uniform upload. Benefit scales with tile count. Now that per-tile vertex cost is 5× lower, draw call overhead is a larger fraction of total time.

### 4. RTT texture atlas

Instead of 30 separate FBOs (one per pool slot), render all tile layers into regions of a single large atlas texture. Reduces FBO bind/unbind overhead from O(tiles) to O(1).

**Impact:** Reduces state change overhead. Primarily helps mobile GPUs which have expensive FBO switches.

### Native WebGL2 depth texture (DEPTH_COMPONENT32F + R32F)

**Attempted:** Replace RGBA8 depth FBO with native depth texture for shader sampling + R32F for CPU readback.

**Result:** 4× regression on GPU, 26% regression on SwiftShader. The separate FBO with float textures is far more expensive than sharing a single FBO with RGBA8 texture swaps. FBO format heterogeneity (RGBA8 for coords, float for depth) forces driver revalidation.

**Lesson:** The shared RGBA8 FBO approach is well-optimized. Format changes that break FBO sharing cause disproportionate overhead.

### R16F depth texture (shared FBO)

**Attempted:** Keep shared FBO but swap depth texture from RGBA8 to R16F.

**Result:** 34% regression (SwiftShader), 4× regression (GPU). Mixing RGBA8 and R16F color attachments on the same FBO triggers expensive format revalidation on every swap.

**Lesson:** FBO color attachment format must be uniform across all textures that share it.

### blitFramebuffer depth capture

**Attempted:** Eliminate depth pre-pass by blitting the main FBO's depth buffer to the depth texture after terrain composite.

**Result:** Resolution mismatch bug (main FBO at full res vs depth texture at dpr-scaled). `blitFramebuffer` with `DEPTH_BUFFER_BIT` requires equal src/dst sizes. Would require full-res depth texture, breaking UV mapping assumptions in the depth sampling shader.

**Lesson:** The depth pre-pass and main composite render at different resolutions. Sharing depth between them requires a resolution-aware approach.

</details>


CommanderStorm:
(should be clear, but defitively needs v6.0)

src/render/terrain.ts
CommanderStorm:
```suggestion
import {RGBAImage} from '../util/image';
import {warnOnce} from '../util/util';
import {Pos3dArray, TriangleIndexArray} from '../data/array_types.g';
import pos3dAttributes from '../data/pos3d_attributes';
import {SegmentVector} from '../data/segment';
import {Texture} from '../webgl/texture';
import {MercatorCoordinate} from '../geo/mercator_coordinate';
import {TerrainTileManager} from '../tile/terrain_tile_manager';
import {EXTENT} from '../data/extent';
import {earthRadius, type LngLat} from '../geo/lng_lat';
import {Mesh} from './mesh';
import {isInBoundsForZoomLngLat} from '../util/world_bounds';
import {NORTH_POLE_Y, SOUTH_POLE_Y} from './subdivision';
import {coveringTiles} from '../geo/projection/covering_tiles';
import type {DEMData} from '../data/dem_data';
```

src/render/terrain.ts
```diff
@@ -14,6 +28,7 @@ import {Mesh} from './mesh';
 import {isInBoundsForZoomLngLat} from '../util/world_bounds';
 import {NORTH_POLE_Y, SOUTH_POLE_Y} from './subdivision';
 import {coveringTiles} from '../geo/projection/covering_tiles';
+import type {DEMData} from '../data/dem_data';
```
HarelM, line 17: Isn't this already imported above?

CommanderStorm: yea, github is not a great place to edit this

src/render/terrain.ts
```diff
@@ -266,21 +279,32 @@ export class Terrain {
...
         if (!this._emptyDemTexture) {
             const context = this.painter.context;
+            const gl2 = context.gl as WebGL2RenderingContext;
             const image = new RGBAImage({width: 1, height: 1}, new Uint8Array(1 * 4));
             this._emptyDepthTexture = new Texture(context, image, context.gl.RGBA, {premultiply: false});
             this._emptyDemUnpack = [0, 0, 0, 0];
-            this._emptyDemTexture = new Texture(context, new RGBAImage({width: 1, height: 1}), context.gl.RGBA, {premultiply: false});
-            this._emptyDemTexture.bind(context.gl.NEAREST, context.gl.CLAMP_TO_EDGE);
+            // 1x1 R16F texture with elevation=0 as fallback (matches decoded elevation texture format)
+            const emptyTex = gl2.createTexture();
```
HarelM: It's better to write the full name - `emptyTexture`.

src/render/terrain.ts
```diff
@@ -266,21 +279,32 @@ export class Terrain {
...
+            gl2.bindTexture(gl2.TEXTURE_2D, emptyTex);
+            gl2.texParameteri(gl2.TEXTURE_2D, gl2.TEXTURE_MIN_FILTER, gl2.LINEAR);
+            gl2.texParameteri(gl2.TEXTURE_2D, gl2.TEXTURE_MAG_FILTER, gl2.LINEAR);
+            gl2.texParameteri(gl2.TEXTURE_2D, gl2.TEXTURE_WRAP_S, gl2.CLAMP_TO_EDGE);
+            gl2.texParameteri(gl2.TEXTURE_2D, gl2.TEXTURE_WRAP_T, gl2.CLAMP_TO_EDGE);
+            gl2.texImage2D(gl2.TEXTURE_2D, 0, gl2.R16F, 1, 1, 0, gl2.RED, gl2.FLOAT, new Float32Array([0]));
+            this._emptyDemTexture = {texture: emptyTex, destroy: () => gl2.deleteTexture(emptyTex)} as any as Texture;
```
HarelM: Why is this casting needed?

src/render/terrain.ts
```diff
@@ -299,19 +323,53 @@ export class Terrain {
             this._demMatrixCache[matrixKey] = {matrix: demMatrix, coord: tileID};
         }
         // return uniform values & textures
+        const elevTex = sourceTile?.dem ? this._elevationTextures[sourceTile.dem.uid] : null;
```
HarelM: Looks like this is cached per tile, on the tile object, in the original approach, can we use the same mechanism?

src/render/terrain.ts
```diff
@@ -299,19 +323,53 @@ export class Terrain {
...
+            const b = pixels[i * 4 + 2];
+            elevations[i] = r * dem.redFactor + g * dem.greenFactor + b * dem.blueFactor - dem.baseShift;
+        }
+
+        if (this._elevationTextures[key]) {
+            gl.deleteTexture(this._elevationTextures[key]);
+        }
+
+        const tex = gl.createTexture();
+        gl.bindTexture(gl.TEXTURE_2D, tex);
```
HarelM: This part looks very similar to the empty texture flow, consider moving this to a different method and reuse it.

src/render/terrain.ts
```diff
@@ -299,19 +323,53 @@ export class Terrain {
...
+            texture: elevTex || (sourceTile?.demTexture || this._emptyDemTexture).texture,
             depthTexture: (this._fboDepthTexture || this._emptyDepthTexture).texture,
             tile: sourceTile
         };
     }
 
+    /**
+     * Decode DEM pixel data to a float elevation texture (R16F) with hardware accelaerted gl.LINEAR filtering
+     */
+    private _updateElevationTexture(dem: DEMData): void {
```
HarelM: I think it might be worth splitting this method into create and update.
i.e. `getElevationTextureFromDem(dem: DEMData)` and move the assignment outside this method.

src/render/terrain.ts
```diff
@@ -299,19 +323,53 @@ export class Terrain {
...
+            const g = pixels[i * 4 + 1];
+            const b = pixels[i * 4 + 2];
+            elevations[i] = r * dem.redFactor + g * dem.greenFactor + b * dem.blueFactor - dem.baseShift;
+        }
+
+        if (this._elevationTextures[key]) {
+            gl.deleteTexture(this._elevationTextures[key]);
+        }
+
+        const tex = gl.createTexture();
```
HarelM: don't be stingy on characters for variable names :-)

src/render/terrain.ts
```diff
@@ -306,12 +315,64 @@ export class Terrain {
...
             depthTexture: (this._fboDepthTexture || this._emptyDepthTexture).texture,
             tile: sourceTile
         };
     }
 
+    /**
+     * Create a new R16F texture with gl.LINEAR filtering for hardware bilinear interpolation.
+     * Used for both the per-tile decoded elevation textures and the empty fallback.
+     */
+    private _createR16FTexture(width: number, height: number, data: Float32Array): WebGLTexture {
```
lucaswoj, line 327: I'd recommend testing this on some devices with Mali and Adreno GPUs. Mali and Adreno do FP16 texture filtering at reduced precision. I've been bit by this recently. Look for a "shimmer" effect during slow map movements.

See https://keaukraine.medium.com/filtering-of-half-float-textures-on-different-mobile-gpus-6be3c277f55e 

Related issue: #7419

src/render/terrain.ts
```diff
@@ -266,21 +271,25 @@ export class Terrain {
         // creates an empty depth-buffer texture which is needed, during the initialization process of the 3d mesh..
         if (!this._emptyDemTexture) {
             const context = this.painter.context;
-            const image = new RGBAImage({width: 1, height: 1}, new Uint8Array(1 * 4));
+            const gl2 = context.gl as WebGL2RenderingContext;
```
CommanderStorm:
```suggestion
```

src/render/terrain.ts
```diff
@@ -266,21 +271,25 @@ export class Terrain {
...
+            this._emptyDemTexture = this._createR16FTexture(1, 1, new Float32Array([0]));
             this._emptyDemMatrix = mat4.identity([] as any);
         }
-        // find covering dem tile and prepare demTexture
+        // find covering dem tile and prepare textures
         const sourceTile = this.tileManager.getSourceTile(tileID, true);
         if (sourceTile?.dem && (!sourceTile.demTexture || sourceTile.needsTerrainPrepare)) {
             const context = this.painter.context;
+            // Keep RGBA8 demTexture for hillshade/color-relief compatibility
             sourceTile.demTexture = this.painter.getTileTexture(sourceTile.dem.stride);
```
lucaswoj, line 287: We are allocating a brand new R16F texture for every tile:
```
sourceTile.demElevationTexture = this._getElevationTexture(sourceTile.dem, sourceTile.demElevationTexture);
```

Whereas the RGBA8 textures are pooled:
```
sourceTile.demTexture = this.painter.getTileTexture(sourceTile.dem.stride);
```

On fast pans through a DEM source this could result in 100+ new texture allocations per second which is significantly slow.

src/render/terrain.ts
```diff
@@ -306,12 +315,64 @@ export class Terrain {
             'u_terrain_matrix': matrixKey ? this._demMatrixCache[matrixKey].matrix : this._emptyDemMatrix,
             'u_terrain_unpack': sourceTile?.dem?.getUnpackVector() || this._emptyDemUnpack,
```
lucaswoj, line 316: Are we still using this uniform?

lucaswoj (commented):
This PR moves some work from the GPU to the CPU (on the main UI thread). That's a reasonable tradeoff in a lot of cases. But I want to make sure we've looked at it from all angles before we landing this PR because the `terrain_render` benchmark doesn't capture the full picture.

The `terrain_render` benchmark measures steady-state frame render time with pre-loaded DEM tiles. It doesn't measure the new CPU decode cost, texture allocation/upload cost, or memory usage. These could add up to noticeable jank during pan/zoom as new DEM tiles come in.

Some metrics we should test before declaring this new system is "faster": 

 1. Total time to process a terrain tile, from "data fetched" to "uploaded to GPU".
 2. UI jank or dropped frames during pan/zoom.
 3. RAM memory usage.
 4. GPU memory usage.
 5. Time to render one frame (what we have today)

My hunch is that 1, 2, 3, and 4 have gotten worse in this PR. If that's true, we might be able to claw back some performance by

 - moving the CPU-side work to a worker thread 
 - completely eliminating the RGBA8 texture 
 - pooling R16F textures to reduce the # of calls to `gl.createTexture`

CommanderStorm:
Yea, likely not the best of ideas in this PR.. on second thought moving to the main thread does not sound like the best idea. Let's regroup

lucaswoj:
To be clear, I do think there's a good chance we can mitigate all of the CPU perf downside, reduce overall memory usage, and keep the render perf improvement with this approach!

PR #7481, opened 2026-04-16
fix!: improve types for `{set,set}LayoutProperty`, `{set,set}PaintProperty` to be the actual type instead of `string`/`any`

CommanderStorm:
Should remove a bunch of incorrect code looking at our codebase alone.

Definitively breaking given how many issues we had with this in our codebase too though.

`tsc` slows down a bit because of this, but they are working on this part, so likely fine.


src/style/style_layer.ts
```diff
@@ -265,22 +269,22 @@ export abstract class StyleLayer extends Evented {
...
         if (name.endsWith(TRANSITION_SUFFIX)) {
             const baseName = name.slice(0, -TRANSITION_SUFFIX.length);
             if (baseName === 'visibility' || this._unevaluatedLayout?.hasProperty(baseName)) {
                 throw new Error(name + ERROR_LAYOUT_NOT_PAINT);
             }
-            return this._transitionablePaint.getTransition(baseName);
+            return this._transitionablePaint.getTransition(baseName) as AllPaintProperties[K];
         } else {
-            if (name === 'visibility' || this._unevaluatedLayout?.hasProperty(name)) {
+            if (this._unevaluatedLayout?.hasProperty(name)) {
```
CommanderStorm: now a ts error.. not sure if we want to keep these errors, this will error in TS code, but won't in js

src/style/style_layer.ts
```diff
@@ -153,10 +157,10 @@ export abstract class StyleLayer extends Evented {
             this._transitionablePaint = new Transitionable(properties.paint, globalState);
 
             for (const property in layer.paint) {
-                this.setPaintProperty(property, layer.paint[property], {validate: false});
+                this.setPaintProperty(property as keyof AllPaintProperties, layer.paint[property as keyof typeof layer.paint], {validate: false});
```
CommanderStorm, line 160: all of these are nessary because for..in does not narrow the type for some reason

src/style/style_layer.test.ts
```diff
@@ -196,7 +196,7 @@ describe('StyleLayer.setPaintProperty', () => {
         } as LayerSpecification, {});
 
         const errorPromise = layer.once('error');
-        layer.setPaintProperty('visibility', 'visible');
+        layer.setPaintProperty('visibility' as any, 'visible');
```
HarelM, line 199: This seems strange that it requires "as any" doesn't it?

HarelM: Nevermind, this is an error case test.

src/style/style_layer.test.ts
```diff
@@ -333,7 +333,7 @@ describe('StyleLayer.setLayoutProperty', () => {
         } as LayerSpecification, {});
 
         const errorPromise = layer.once('error');
-        layer.setLayoutProperty('text-color', 'blue');
+        layer.setPaintProperty('text-color', 'blue');
```
HarelM: Are you sure about this change?

src/style/style_layer.ts
HarelM: Yeah, this sucks,not sure why typescript traspiler doesnt do that...

src/style/style_layer.ts
```diff
@@ -265,23 +269,23 @@ export abstract class StyleLayer extends Evented {
...
         if (name.endsWith(TRANSITION_SUFFIX)) {
             const baseName = name.slice(0, -TRANSITION_SUFFIX.length);
             if (baseName === 'visibility' || this._unevaluatedLayout?.hasProperty(baseName)) {
                 throw new Error(name + ERROR_LAYOUT_NOT_PAINT);
             }
-            return this._transitionablePaint.getTransition(baseName);
+            return this._transitionablePaint.getTransition(baseName) as AllPaintProperties[K];
         } else {
-            if (name === 'visibility' || this._unevaluatedLayout?.hasProperty(name)) {
+            if (this._unevaluatedLayout?.hasProperty(name)) {
```
HarelM: Are you sure about this condition change? I think it was there delibery, wasn't it?

HarelM:
I'm not really sure this is a breaking change to be honest, if your are passing incorrect values this should help you. 
But I understand the concern, since we are planning a breaking change version this should be part of it.
I'll think about starting a first pre-release next week maybe.

src/style/style_layer.test.ts
CommanderStorm:
```suggestion
        layer.setLayoutProperty('text-color' as any, 'blue' as any);
```

CommanderStorm: Yea, that was the wrong change on second look..
I thought that because text-color is paint-only, this is a typo and needs to be paint too

src/style/style_layer.ts
CommanderStorm: there is a long issue. Marked as not planned.
For some reason, objects in JS don't have fixed keys or something and somehow keys may apear??

So string is the only valid key apparently??

src/style/style_layer.ts
CommanderStorm: The other code now also triggers a TS error.
So not sure.. We can keep it if I `as any`, but that sounds hacky..

I added this because TS was "string".

CommanderStorm:
```suggestion
            if (name as any === 'visibility' || this._unevaluatedLayout?.hasProperty(name)) {
```

src/style/style_layer.ts
```diff
@@ -297,7 +301,8 @@ export abstract class StyleLayer extends Evented {
             const wasDataDriven = transitionable.value.isDataDriven();
             const oldValue = transitionable.value;
 
-            this._transitionablePaint.setValue(name, value);
+            // FIXME: Transitionable.setValue uses a free-floating T that can't unify with the AllPaintProperties union
```
CommanderStorm:
```suggestion
            // Transitionable.setValue uses a free-floating T that can't unify with the AllPaintProperties union -> better types downstream of this code needed
```

CommanderStorm:
> I'm not really sure this is a breaking change to be honest

technically not, but it will break lots of people who are using this wrong.
The previous warnings was simple to ignore...

src/style/style_layer.ts
```diff
@@ -32,6 +34,8 @@ import {type mat4} from 'gl-matrix';
 import type {UnwrappedTileID} from '../tile/tile_id';
 import type {VectorTileFeatureLike} from '@maplibre/vt-pbf';
 
+type PaintPropertyEntry = { [K in keyof AllPaintProperties]: {name: K; value: AllPaintProperties[K]} }[keyof AllPaintProperties];
```
CommanderStorm:
```suggestion
export type PaintPropertyEntry = { [K in keyof AllPaintProperties]: {name: K; value: AllPaintProperties[K]} }[keyof AllPaintProperties];
```

src/ui/map.ts
CommanderStorm, line 26:
```suggestion
import type {StyleLayer, PaintPropertyEntry} from '../style/style_layer';
```

HarelM (approved):

PR #7483, opened 2026-04-17
fix: add accept headers for MLT, MVT and raster sources according to the current capabilities

CommanderStorm:
As requested in https://github.com/maplibre/maplibre-style-spec/pull/1251#discussion_r2289493068 we should include the Accept header.

This is important for us as a tile server, given that we otherwise have no clue if we should transcode an MLT <-> MVT source.


nyurik (dismissed):
lgtm

CommanderStorm:
just to be save, lets put this under v6.
I don't think this is breaking, but we never know what people are doing with headers... maybe WAF filtering for incoming requests or something bad.. Not sure

HarelM:
Looks good, thanks, let's wait for version 6 with this.

HarelM:
See discussion here:
- https://github.com/maplibre/maplibre-native/issues/4236

And PR here:
- https://github.com/maplibre/maplibre-native/pull/4231

Let's finish the discussion there first before merging this.

test/build/bundle_size.json
```diff
@@ -1,7 +1,7 @@
 {
     "dist/maplibre-gl.mjs": {
-        "raw": 964777,
-        "gzip": 247855
+        "raw": 960710,
+        "gzip": 247310
```
```suggestion
        "raw": 963952,
        "gzip": 247901
```

CommanderStorm:
@HarelM seems that everyone in that discussion agrees that this is the right way to handle this.
Can we merge this or are there any concerns ot address that I missed?

https://github.com/maplibre/maplibre-native/issues/4236

CommanderStorm:
> Let's finish the discussion there first before merging this

PR is finally moving forward on native, so can also move forward here 😉

test/build/bundle_size.json
```diff
@@ -1,7 +1,7 @@
 {
     "dist/maplibre-gl.mjs": {
-        "raw": 585205,
-        "gzip": 148827
+        "raw": 585280,
+        "gzip": 14885
```
```suggestion
        "raw": 585062,
        "gzip": 148825
```

Issue #3062, opened 2023-09-04
Add a per-layer opacity property


FloChehab:
Hello,

## User Story

As a user I would like to have an opacity setting at the layer level to avoid "additive opacity" inside the same layer.

## Rationale

Here is an example with the current situation:
![image](https://github.com/maplibre/maplibre-gl-js/assets/20242665/7480b9fa-6a90-4476-93c1-7de149830b4e)
(Screenshot from https://jsfiddle.net/wq8gk2z5/22/)

We can see that the resulting line has actually ~4 colors due to the way it's rendered. I would like the rendered line to be a single color / opacity no matter how many times it crosses itself.

## Impact

Would be a nice addition.

## Related

* The matching (old) issue on MapboxGlJS side seems to be https://github.com/mapbox/mapbox-gl-js/issues/4090 
* This might also be related to https://github.com/maplibre/maplibre-gl-js/issues/48 though I am not 100% sure.

HarelM:
Might be related to the following: #1191
This PR is out of sync and very old unfortunately...

FloChehab:
Hello @HarelM,

I have been chatting with a colleague of mine @philae-ael in the recent days as I know his general interest with the matter being discussed here. And... he has done few POCs already.
As happy user of Maplibre, we are willing to allocate a bit of his professional time to contribute back :rocket: 

I'll let the mic to him but I'll keep an interested eye on what's happening here :nerd_face: 

I guess the next step would be to discuss the actual objective(s) and the different approaches possible and the one that are the most likely to get merged ; what we should be careful, etc. On my side I can help track / followup things if welcomed.

All the best !

FloChehab:
Great, we will try to kick-off a new discussion there in the coming days.

FloChehab:
Digging a bit more we found https://github.com/mapbox/mapbox-gl-js/pull/11082, @HarelM any ideas on if we could bring something that resembles to Maplibre (this has no impact on the spec) or should we try to give another shot to https://github.com/maplibre/maplibre-gl-js/pull/1191 ?

HarelM:
If we could fix it without a style spec change it would be great.
We can't copy code from mapbox as of version 2.0 and I think the above PR is dated after the fork of this repo and the license change.
In order not to infringe copyright laws we should avoid looking at the linked code.

pnorman:
[This](https://jsbin.com/puyihexazu/2/edit?js,output) is an example of the problem with real-world data and inset lines.

[...]

PR #7490, opened 2026-04-18
fix: Line opacity shading during overlap

### LINKED ISSUE #6334: Overlap of semi-transparent lines renders darker (JSFiddle included)
### Description
When two or more line features with partial transparency are drawn and overlap, the overlapping region appears noticeably darker in MapLibre GL JS.

I tested this behavior with a simple JSFiddle example (see link below), and I confirmed that overlapping transparent lines accumulate opacity in a way that isn't desired for uniform styling.

### Expected Behavior
Overlapping lines with the same styling (color/opacity) should visually appear uniform, with no darker overlap—behaving similarly to Mapbox GL JS.

Image

### Actual behavior

When two or more line features with partial transparency overlap, the overlapping region appears noticeably darker in MapLibre GL JS.

Image

#### Reproduction URL
https://jsfiddle.net/Samane/db97kq51/

I’m eager to address this issue personally and submit a PR.

Thank you for your excellent work on MapLibre!

CommanderStorm:
This is nessary for the OSM main page migration (blocker for navigation) because it looks ugly as hell ^^

Benchmarks as 3x slower, but the visual output is imo worth it.
Could likely be improved by someone with graphics experience

image

Or more clearer

| new, `line-oppacity` handling | old handling `line-oppacity` or unchanged transparent `line-color`) |
|--------|--------|
| image | image | 

Resolves #6334

CC @SamaneYaghoobi
CC @1ec5


HarelM:
Thanks for taking the time to solve this one! 
Does it affect line dashes? Line pattern?
How does this affect performance? 
The changes in performance is only when there is a problem, right? I.e only when the line is semi transparent, right? 
If this is the case this is good since it will affect only places where there's a problem.

HarelM:
I'll do a deeper review later this week.

CommanderStorm:
I ran the benchmarks where I suspected this might have affect and no, does not seem to be a problem.
Would also be a bit supprising given that this surgically only affects lines where this is actually really really really nessary.

Render tests are farly solid on lines imo, if there is something missing I don't know.

src/style/style_layer/line_style_layer.ts
```diff
@@ -117,6 +120,23 @@ export class LineStyleLayer extends StyleLayer {
...
+        const opacity = this.paint.get('line-opacity');
+        const constantOpacity = opacity.constantOr(-1);
+        return constantOpacity > 0 && constantOpacity < 1 && !this.isHidden();
+    }
+
+    onRemove = () => {
+        this.resize();
+    };
+
+    resize() {
```
HarelM, line 134: Why is this called resize?

src/webgl/draw/draw_line.ts
```diff
@@ -139,14 +145,88 @@ function bindGradientAndDashTextures(
...
-
-    const {isRenderingToTexture} = renderOptions;
+    if (painter.renderPass !== 'offscreen' && painter.renderPass !== 'translucent') return;
 
     const opacity = layer.paint.get('line-opacity');
     const width = layer.paint.get('line-width');
     if (opacity.constantOr(1) === 0 || width.constantOr(1) === 0) return;
 
+    const constantOpacity = opacity.constantOr(-1);
+    const useOffscreen = constantOpacity > 0 && constantOpacity < 1 && !painter.style.map.terrain;
```
HarelM: Can't this use `layer.hasOffscreenPass()` instead of duplicating the logic?

src/webgl/draw/draw_line.ts
```diff
@@ -139,14 +145,88 @@ function bindGradientAndDashTextures(
...
+    // Free FBO when offscreen path is no longer needed
+    if (!useOffscreen && layer.lineFbo) {
+        layer.lineFbo.destroy();
+        layer.lineFbo = null;
+    }
+
+    // if the opacity is not constant (full/tansparent), we need to render to an offscreen FBO at full opacity and the 
+    // composite pass will apply the opacity.
+    // If we do this any other way, there will be really ugly artifacts
+    if (useOffscreen) {
```
HarelM, line 168: Consider early return here instead of this nested if - i.e `if (!useOffscreen && painter.renderPass === 'offscreen') return.
This matches the previous logic I believe, so I think it's worth adding it for readability and resemblance.

src/webgl/draw/draw_line.ts
```diff
@@ -139,14 +145,88 @@ function bindGradientAndDashTextures(
...
+    } else if (painter.renderPass === 'translucent') {
+        drawLineTiles(painter, tileManager, layer, coords, renderOptions, layer.paint, true);
+    }
+}
+
+function drawLineOffscreen(painter: Painter, tileManager: TileManager, layer: LineStyleLayer, coords: OverscaledTileID[], renderOptions: RenderOptions) {
+    const context = painter.context;
+
+    if (!layer.lineFbo) {
+        layer.lineFbo = createLineFbo(context, painter.width, painter.height);
```
HarelM: Would it be more performant to keep it instead of deleting and creating it between render calls?

src/webgl/draw/draw_line.ts
```diff
@@ -139,14 +145,88 @@ function bindGradientAndDashTextures(
...
+
+    context.bindFramebuffer.set(layer.lineFbo.framebuffer);
+    context.viewport.set([0, 0, painter.width, painter.height]);
+    context.clear({color: Color.transparent, depth: 1, stencil: 0});
+
+    // Force stencil masks to render into the FBO
+    painter.currentStencilSource = undefined;
+    painter._renderTileClippingMasks(layer, coords, false);
+
+    const paintOverride = {
```
HarelM: Can you elaborate on this override? it looks more like a hack than a real solution...?

src/webgl/draw/draw_line.ts
```diff
@@ -139,14 +145,88 @@ function bindGradientAndDashTextures(
...
+    if (painter.renderPass !== 'offscreen' && painter.renderPass !== 'translucent') return;
 
     const opacity = layer.paint.get('line-opacity');
     const width = layer.paint.get('line-width');
     if (opacity.constantOr(1) === 0 || width.constantOr(1) === 0) return;
 
+    const constantOpacity = opacity.constantOr(-1);
+    const useOffscreen = constantOpacity > 0 && constantOpacity < 1 && !painter.style.map.terrain;
+
+    // Free FBO when offscreen path is no longer needed
```
HarelM: I think some of these comment can be removed...

HarelM:
Thanks for looking into fixing this! This is a long standing issue and I'm glad you were able to tackle it.
I've added some comments, but looks good otherwise, THANKS!!

HarelM:
@lucaswoj feel free to take a look at this solution and let us know what you think.

src/webgl/draw/draw_line.ts
lucaswoj: Or, better yet, can we use the existing texture pool at `painter.getTexture`?

lucaswoj:
This approach will certainly produce the correct visual result.

Could you upload a PDF of the bench results from the web runner? That should give us enough info to tell whether this causes a perf regression or is "within the noise floor."

This exact approach was implemented in GL JS long ago but was removed due to perf concerns. See discussion at https://github.com/mapbox/mapbox-gl-js/issues/523 and https://github.com/mapbox/mapbox-gl-js/issues/4090. 

It's possible that the performance landscape has changed since then but let's proceed with extreme caution.

EDIT: All significant issue comments are from _before_ the project went close source. Hope it's okay to link to the issues here.

src/style/style_layer/line_style_layer.ts
CommanderStorm: Because that is the interface?

I want to invalidate the FBO on resize..
Same as heatmaps..

src/webgl/draw/draw_line.ts
CommanderStorm: That sounds incorrect.

Where did we early return? Why did we early return?
That would mean we are not rendering if you change your style from having oppacity to not having it.

HarelM: In line 148, the original condition retuned.
This code flow can be similar in that aspect that it would return as well for the original condition + "integer" opacity.

src/style/style_layer/line_style_layer.ts
HarelM: Oh, ok, thanks for the extra info.

src/webgl/draw/draw_line.ts
CommanderStorm: I have comments when I would scratch my head why something exists. I am going to leave those

src/webgl/draw/draw_line.ts
CommanderStorm: I think this splits the logic and obvuscates rather than making it clearer.
Lets not.

```diff
+    if (!useOffscreen && painter.renderPass === 'offscreen') return
    if (useOffscreen) {
        if (painter.renderPass === 'offscreen') {
            drawLineOffscreen(painter, tileManager, layer, coords, renderOptions);
        } else {
            drawLineComposite(painter, layer);
        }
-    } else if (painter.renderPass === 'translucent') {
+    } else {
        drawLineTiles(painter, tileManager, layer, coords, renderOptions, layer.paint, true);
    }
```

src/webgl/draw/draw_line.ts
CommanderStorm: `painter.getTexture` is for square textures and is cleared every frame.
We need a paint-sized texture, zero/one * per layer (zero if non-transparent, one if transparent).

This cannot be shared between layers because of how offscreen and non-offscreen is executed after each other.

Also, we are not deleting and creating the texture, we are keeping it and only allocating it if we need it..
We are de-allocating it if the size parameters change or if removed..
This is consistent with the heatmap.

I could move this into the translucent pass and reuse one, static, FBO... but not sure if this is a good approach.
There must be a (graphics) reason why we are splittiing offscreen and non-offscreen passes.. 🤔

src/webgl/draw/draw_line.ts
CommanderStorm: The alternative would be to move this down one layer into the shader. I can add a commit that shows what this would look like.

Not sure what is better.

src/webgl/draw/draw_line.ts
```diff
@@ -139,14 +143,83 @@ function bindGradientAndDashTextures(
...
+        drawLineTiles(painter, tileManager, layer, coords, renderOptions, false, true);
+    }
+}
+
+function drawLineOffscreen(painter: Painter, tileManager: TileManager, layer: LineStyleLayer, coords: OverscaledTileID[], renderOptions: RenderOptions) {
+    const context = painter.context;
+
+    if (!layer.lineFbo) {
+        layer.lineFbo = createLineFbo(context, painter.width, painter.height);
+    }
```
CommanderStorm:
```suggestion
    layer.lineFbo ??= createLineFbo(context, painter.width, painter.height);
```

CommanderStorm:
image

Not sure how to generate these PDFs, but this is what I get when I compare the performance against main..

CommanderStorm:
Also, generating the entire benchmark set seems a bit long, so I only incldued the few that actually can be impacted perf wise..

The others are simpliy not containing transparent line layers..

lucaswoj:
You can learn more about the PDFs generated by `npm run benchmark` by reading https://github.com/maplibre/maplibre-gl-js/blob/50d9756cbbe9df104ae8e8692b1719b7b04099ec/test/bench/README.md#L80

More importantly, neither `Paint` nor `LayerLine` exercises the new code path. `hasOffscreenPass()` triggers only when `line-opacity.constantOr(-1)` is strictly between 0 and 1. `LayerLine` doesn't set `line-opacity`, and Liberty's four `line-opacity` users are all zoom-interpolated and evaluate to either 0 or 1.

I modified the `LayerLine` to render a 0.5 opacity line and found it be 3x slower on this branch vs `main`

```ts
import Benchmark from '../lib/benchmark';
import createMap from '../lib/create_map';
import style from '../data/empty.json' with {type: 'json'};

export default class LineOpacityTranslucent extends Benchmark {
    map: any;

    async setup() {
        const layers = [];
        for (let i = 0; i < 50; i++) {
            layers.push({
                id: `linelayer${i}`,
                type: 'line',
                source: 'openmaptiles',
                'source-layer': 'transportation',
                paint: {'line-opacity': 0.5},
            });
        }
        this.map = await createMap({
            zoom: 16,
            width: 1024,
            height: 768,
            center: [-77.032194, 38.912753],
            style: {...style, layers},
        });
    }

    bench() { Benchmark.renderMap(this.map); }
    teardown() { this.map.remove(); }
}
```

| | main | this branch |
| --- | --- | --- |
| `LineOpacityTranslucent` | 0.62 ms | 1.94 ms |

I hope this perf cost motivates us to consider some ways to mitigate the impact of this feature, and extend support for data-driven styles, before shipping.

Screenshot 2026-04-18 at 22 19 41

HarelM:
Not that I argue that we should see if there's a better way to implement this from a performance perspective, but it's worth noting that people are not using this feature due to this issue. I've seen a few issues opened on that which resulted in people saying: "never mind, I'll just use a solid color line". (which won't be impacted by this change).
So if I needed to choose between worse performance and the current "bug", I would probably favor solving the bug.
If we can't find a solution that has better performance I would be inclined to add a note in regards to `line-opacity` that it can impact performance.
That's my 2 cents I guess...

CommanderStorm:
Will have a look what I can do. I have a few ideas, but the big ones (which also would solve data driven styling) would require MRTs, so is not curently possible on v5.x.

lucaswoj:
> people are not using this feature 

Many real world styles _are_ using `line-opacity` and some depend on the per-feature opacity behavior. 

It's common to use transparent lines to indicate administrative boundaries. Mapeak, for example, includes about 5 such layers. Some more complex basemaps have dozens of such layers. Are we going to consider the perf impact on these users?

Screenshot 2026-04-19 at 07 51 17


Some styles overlay lots of transparent lines and use the overlapping opacity to show patterns, like the Gaia GPS public tracks layer. Are we going to leave these users no way to recreate this effect? 

image

CommanderStorm:
cannot reproduce the render test creating different results on my device, not sure why bankok/SF/nepal is failing..

CommanderStorm:
> I hope this perf cost motivates us to consider some ways to mitigate the impact of this feature, and extend support for data-driven styles, before shipping.

Perf impact is now down to 1.95x.
Still some impact, but only for layers where this feature is used.
Maybe the work that will be possible once we are on webgl2 alliteratives this, not sure.

test/integration/render/tests/line-triangulation/round/expected.png
HarelM, line 1: I'm not sure if this was missed or not, but this is not a good result for the render test - the + in the middle needs to be removed.

CommanderStorm: ok, did not notice this. Reverted that commit.
Now back to 3x

mizmay:
Grabbed these from real-world stylesheet (Positron, as currently written for OpenFreeMap)

`"line-opacity": ["interpolate",["linear"],["zoom"],0,0.4,4,1]`
`"line-color": "#d9d9d987"`
`"line-color":  [ "interpolate", ["linear"],["zoom"],5.8, "#d9d9d987", 6, "#ffffff"]`

My understanding is that this change would:
- not correct artifacts if `line-opacity` has a zoom expression
- ignore alpha channel in `line-color`

On the + side, I do think Gaia GPS could still render their singular "track" style, for whatever that is worth (though I am not advocating for this...).

On the - side, it is reasonable to expect that future users would expect the same behavior from an interpolated expression.

I still think this could be a great improvement, but I suspect supporting interpolated expressions will make the performance picture worse, not better?

EDIT: I can't prove this, but ironically, the interpolate/zoom expressions could exist because the overlaid opacities look so bad when you zoom in. If true, this could be evidence in support of fixing this....

EDIT2: removed my comment about breaking change, distracting. Recognizing instead: these two cases (line-color alpha channel, not fixing expressions) would make communicating about this change challenging.

HarelM:
I'm not sure fixing a bug (the way I see it) should be a breaking change.
Having said that, there were places where we introduced a bug fix as a breaking change in the past (and even for v6 there's one related to geojson).
I think using the alpha channel in `line-color` instead of using `line-opacity` would allow the use case where you would like to overlay lines to get a heatmap or something like that, so fixing line-opacity and allowing line-color "overdraw" should solve most problems.

mizmay:
@HarelM just noting that you missed my point here, which was about interpolation expressions.

HarelM:
I don't see any problem interpolating an expression in general, expression evaluation in maplibre works well, if there's a bug with this implementation it's not clear from what you wrote.

CommanderStorm:
We are going in circles..
This implemntation is not targeting expression functions, it is only targeting constant functions as a first step.
As I said: expressions need Webgl2 support as far as I can tell because I think that needs an MRT..

=> Fixing this bug for zoom or property Expressions are a different PR.

mizmay:
@CommanderStorm is there a plan for expressions?

1ec5:
Does this PR also mitigate the self-overlaps in https://github.com/maplibre/maplibre-gl-js/issues/3062#issuecomment-1928007171? This would be enough of an improvement that I’d imagine some designers would be open to more of a performance hit than usual. The issue is even apparent in one of the counterexamples in https://github.com/maplibre/maplibre-gl-js/pull/7490#issuecomment-4276121744.

CommanderStorm:
Sure, works as intended.
I added a render test to show this.

Do note that the example that @pnorman shared will NOT show a diff because he used a transparent color.

| new, `line-oppacity` handling | old handling `line-oppacity` or unchanged trasparent `line-color`) |
|--------|--------|
| image | image |

test/integration/render/tests/line-offset/line-opacity/style.json
HarelM, line 1: This example is a bit big, I would advocate to simplify it.

test/integration/render/tests/line-offset/line-opacity/expected.png
1ec5, line 1: Thanks, this is a welcome improvement. The spikes and bleeding are still unfortunate, but it’s progress for sure.

test/integration/render/tests/line-offset/line-opacity/style.json
CommanderStorm: it is 58 lines, I don't see how I can simplify it further.

The geometry here is an US state I think and given how many interesting edgecases it shows at once, I think we should keep it.

HarelM:
@CommanderStorm since we are targeting the removal of webgl1 for version 6, does it make sense to target this feature to version 6 as well and use webgl2 stuff?

CommanderStorm:
That is a different feature, I would like to move that to a different PR

src/webgl/draw/draw_line.ts
```diff
@@ -139,14 +143,81 @@ function bindGradientAndDashTextures(
...
+        layer.lineFbo = null;
+    }
+
+    // if the opacity is not constant (full/tansparent), we need to render to an offscreen FBO at full opacity and the
+    // composite pass will apply the opacity.
+    // If we do this any other way, there will be hideous artifacts
+    //
+    // Because terrain may have cases where we snake and thus need the opacity, we are currently accepting the artefacts.
+    // Needs more looking into how to solve this.
+    if (useOffscreen) {
```
HarelM, line 168: I still have issues with this nested if part.
Can you change it to the following maybe?
```ts
if (painter.renderPass === 'translucent') {
    // maybe rename that to drawLineTranslucentPass(...)? or drawLineTranslucent(...)
    drawLineTiles(painter, tileManager, layer, coords, renderOptions, false, true); 
    return;
}

// all the long comment about offscreen here...
if (useOffscreen && painter.renderPass === 'offscreen') { 
// I think render pass has to be offscreen at this point in the code, so you might want to omit this check as well.
    drawLineOffscreen(painter, tileManager, layer, coords, renderOptions);
    return;
}
drawLineComposite(painter, layer);

File: src/style/style_layer/line_style_layer.ts [hunk 9]
```diff
@@ -117,6 +120,23 @@ export class LineStyleLayer extends StyleLayer {
...
+        if (this.lineFbo) {
+            this.lineFbo.destroy();
```
HarelM: Optional:
```ts
resize() {
    this.lineFbo?.destory();
    this.lineFbo = null;
}
```

src/webgl/draw/draw_line.ts
HarelM: This is resolved, right?
If not let me know...

HarelM:
@CommanderStorm would you like to merge this to version 5 or wait with this to version 6 (regardless of webgl2)?
Please also look into the coverage report as it seems there are some coverage missing (I'm guessing it's minor, but might be worth trying to see what could cover it, don't kill yourself to cover those line though).
I would also add a test with line-color with alpha channel if there is not such test to make sure we allow using that for "line heatmap" like feature as suggested in this thread.
Maybe even consider updating the style-spec docs to indicate that in order to "overdraw" lines one should be using `line-color` and not `line-opacity`.

CommanderStorm:
Lets put this inot v6 and call out how this is breaking and how to migrate in the changelog

src/webgl/draw/draw_line.ts
```diff
@@ -139,14 +143,85 @@ function bindGradientAndDashTextures(
...
+    const useOffscreen = layer.hasOffscreenPass() && !painter.style.map.terrain;
+
+    // GC the FBO if style is transtitoned from having opacity to not having opacity
+    if (!useOffscreen && layer.lineFbo) {
+        layer.lineFbo.destroy();
+        layer.lineFbo = null;
+    }
+
+    if (painter.renderPass === 'translucent') {
+        // maybe rename that to drawLineTranslucentPass(...)? or drawLineTranslucent(...)
```
HarelM: Better not keep this comment in the code...

CHANGELOG.md
```diff
@@ -3,6 +3,8 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
+
+- Fix transparent, overlapping lines creating weird artefacts ([#7490](https://github.com/maplibre/maplibre-gl-js/pull/7490)) (by [@CommanderStorm](https://github.com/CommanderStorm))
```
```suggestion
- Fix transparent, overlapping lines creating artefacts. This is fixed for `line-oppacity`, but purposefully not for transparent `line-color` propertys, thus still allowing transparent colors to stack their effect. ([#7490](https://github.com/maplibre/maplibre-gl-js/pull/7490)) (by [@CommanderStorm](https://github.com/CommanderStorm))
```

src/style/style_layer/line_style_layer.test.ts
```diff
@@ -48,4 +49,41 @@ describe('LineStyleLayer', () => {
         lineLayer.setPaintProperty('line-gradient', null);
         expect(lineLayer.gradientVersion).toBeGreaterThan(gradientVersion);
     });
+
+    describe('lineFbo', () => {
```
CommanderStorm, line 53: fairly meaningless tests, only there because of codecov

src/webgl/draw/draw_line.ts
CommanderStorm:
```suggestion
```

src/webgl/draw/draw_line.ts
CommanderStorm: We cannot, that would be incorrect.
Your logic is not 1:1 identical, nor does it pass the render tests.
I reverted this

src/style/style_layer/line_style_layer.ts
```diff
@@ -117,6 +120,23 @@ export class LineStyleLayer extends StyleLayer {
...
+            this.lineFbo = null;
+        }
```
CommanderStorm:
```suggestion
        this.lineFbo?.destroy();
        this.lineFbo = null;
```

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 - Add a new map creation option, `terrainSkirtLength`, which allows the removal of visually unappealing vertical artifacts when using a terrain along with a transparent background ([#7523](https://github.com/maplibre/maplibre-gl-js/pull/7523)) (by [@safwat-halaby](https://github.com/safwat-halaby))
 
 ### 🐞 Bug fixes
+
+- Fix transparent, overlapping lines creating artefacts. This is fixed for `line-opacity`, but purposefully not for transparent `line-color` properties, thus still allowing transparent colors to stack their effect. ([#7490](https://github.com/maplibre/maplibre-gl-js/pull/7490)) (by [@CommanderStorm](https://github.com/CommanderStorm))
```
```suggestion
- ⚠️ Fix transparent, overlapping lines creating artefacts. This is fixed for `line-opacity`, but purposefully not for transparent `line-color` properties, thus still allowing transparent colors to stack their effect. ([#7490](https://github.com/maplibre/maplibre-gl-js/pull/7490)) (by [@CommanderStorm](https://github.com/CommanderStorm))
```

HarelM (approved):

src/webgl/draw/draw_line.ts
HarelM: That's exactly what's bothering me, but I'll open a different PR to address this as this shouldn't prevent this fix.

xavierjs:
Can we use the stencil buffer to avoid overdraw instead?
For example it is used at this purpose in projected shadows.
It would avoid adding a rendering pass.

CommanderStorm:
I tried removing the stencil masks in https://github.com/maplibre/maplibre-gl-js/pull/7490/changes/d7fe41d47682934b03e0eb96a3e58e4a71b3c175 if this is what you are talking.
Yes, issues at tile edges with artefacts killed that approach.
If we can remove stenceling, this would improve perf  a lot on this.

Not sure how stencil buffer and overdraw can be used here.
As I said, I am not a graphics engineer, so if you think this impl is bad and you can improve upon https://github.com/maplibre/maplibre-gl-js/pull/7570/ this would be appreciated.
That PR is the plift-and-shift revert of this PR.

xavierjs:
Currently we use the 8 bits of the stencil buffer to fix the tiles edge artifacts (the stencil mask is always `0xFF`).
What about using 7 bits of the stencil buffer to fix the tile edges artifacts (`0xFE` mask), and 1 bit to avoid transparent lines overdraw (`0x01` mask)?

But this approach could not be used for general layer opacity, for 3D raster tiles for example we need cannot just discard the rendering since 1 transparent stuff is already rendered at this place. Except if we render the triangles from nearest to farthest.

xavierjs:
I tried solving the issue using the stencil buffer. It works for simple examples, but there are some significant drawbacks:

* When drawing a collection of features, it fixes overdraw. However, for overlapping features, it displays the feature drawn first instead of the feature drawn last. This could be addressed by reversing the rendering order, but that would be counter-intuitive.
* If the line is filled with a semi-transparent pattern texture, the stencil buffer is still updated in the transparent areas, and the color is not blended correctly in the overlapping regions.
* There are artifacts along the line border, where anti-aliasing is computed using SDF in the shaders.

Given these issues, I don’t currently see a solution based on the stencil buffer that would address all of these concerns.

test/bench/benchmarks/layer_line_opacity.ts
```diff
@@ -0,0 +1,30 @@
+import Benchmark from '../lib/benchmark';
+import createMap from '../lib/create_map';
+import style from '../data/empty.json' with {type: 'json'};
```
mizmay, line 3: This test isn't loading any tiles, so the benchmark values are meaningless.

CommanderStorm: Well..
it is precisely showing the impact of this PR though, so the FBO creation overhead..
I don't think we are allowed to skip that anywere, which is why 3x slower.

It is not a realistic style, but macro-benchmarks rarely are.
https://github.com/maplibre/maplibre-gl-js/pull/7490#issuecomment-4275211265

Issue #4492, opened 2024-08-02
Per-feature fill-extrusion-opacity


AbelVM:
As of today, `fill-extrusion-opacity` is applied to the whole layer, not per feature

opacity: 1
![image](https://github.com/user-attachments/assets/25db4d56-203f-4f51-b6dc-07db93b8b202)

opacity: 0.5
![image](https://github.com/user-attachments/assets/e3d83671-54c8-4278-aff4-72eb03db2acf)

The user may want the entire extrusion layer to be transparent to show the 2D layers underneath, but other users may want to get a sense of the extruded features behind those in the foreground.

Maybe a flag or maybe decoupling this property in `fill-extrusion-layer-opacity` and `fill-extrusion-feature-opacity` would do the trick

AbelVM:
Might be related to https://github.com/maplibre/maplibre-gl-js/issues/2559

PR #1191, opened 2022-04-19
New paint property fill-comp-op for fill type layer

tongust:
## Description
This PR is to add new paint property opacity-per-layer to draw fill type layer.
By setting *fill-opacity*, multiple polygons can be drawn with transparency for data visualization. But in some scenarios, the *fill-opacity* is misleading.

- The opacity on overlapped/stacked polygons is increased. It can be opaque when stacking more polygons.
- The color of overlapped area is mixed too. In the left below image , the overlaped region is not red but dark red.

|Current fill opacity| This new feature(by setting *fill-per-layer-opacity* as true) |
:-----------------------------:|:--------------------------------:
image | image

With this change, the opacity of whole layer is fixed and the color will always be same to one of the top most shape.

**Note**:

- To open this feature, the option *fill-per-layer-opacity* need to be set as true explicitly
- Currently it doesn't support line per-layer-opacity, which is one of my TODOs.

## Launch Checklist

 - [x] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [x] Briefly describe the changes in this PR.
 - [x] Include before/after visuals or gifs if this PR includes visual changes.
 - [x] Write tests for all new functionality.
 - [x] Document any changes to public APIs.
 - [x] Post benchmark scores.
 - [x] Manually test the debug page.
 - [x] Suggest a changelog category: bug/feature/docs/etc. or "skip changelog".
 - [x] Add an entry inside this element for inclusion in the `maplibre-gl-js` changelog: `<changelog></changelog>`.

PR #7514, opened 2026-04-24
feat: data driven line opacity shading fix

CommanderStorm:
Second part of the line-oppacity during overlap fix, this time for data-driven styling.
There are more tradeoffs here of "how should this work", here is the one I came up with:

image

image

Opus helped a bit with finding all the places that needed tweaking.


src/webgl/framebuffer.ts
```diff
@@ -1,4 +1,4 @@
-import {ColorAttachment, DepthAttachment, DepthStencilAttachment} from './value';
+import {ColorAttachment, ColorAttachment1, DepthAttachment, DepthStencilAttachment} from './value';
```
CommanderStorm:
```suggestion
import {ColorAttachment, type ColorAttachment1, DepthAttachment, DepthStencilAttachment} from './value';
```

src/webgl/framebuffer.ts
```diff
@@ -39,6 +41,11 @@ export class Framebuffer {
         const texture = this.colorAttachment.get();
         if (texture) gl.deleteTexture(texture);
 
+        if (this.colorAttachment1) {
+            const texture1 = this.colorAttachment1.get();
+            if (texture1) gl.deleteTexture(texture1);
+        }
```
CommanderStorm:
```suggestion
        const texture1 = this.colorAttachment1?.get();
        if (texture1) gl.deleteTexture(texture1);
```

src/webgl/value.ts
```diff
@@ -502,6 +502,20 @@ export class ColorAttachment extends FramebufferAttachment<WebGLTexture> {
...
 }
 
+export class ColorAttachment1 extends FramebufferAttachment<WebGLTexture> {
+    setDirty() {
+        this.dirty = true;
+    }
+    set(v?: WebGLTexture | null): void {
+        if (v === this.current && !this.dirty) return;
+        this.context.bindFramebuffer.set(this.parent);
+        const gl = this.gl as WebGL2RenderingContext;
```
CommanderStorm:
```suggestion
        const gl = this.gl;
```

src/webgl/draw/draw_line.ts
```diff
@@ -183,18 +184,40 @@ export function drawLine(painter: Painter, tileManager: TileManager, layer: Line
...
+    if (isDataDriven) {
+        // RT1 (opacity) uses MAX so self-overlap and feature crossings don't accumulate.
+        // Falls back to last-write-wins if OES_draw_buffers_indexed is unavailable.
+        const ext = context.extDrawBuffersIndexed;
+        ext?.blendEquationiOES(1, gl.MAX);
+        drawLineTiles(painter, tileManager, layer, coords, renderOptions, true, false, true);
+        ext?.blendEquationiOES(1, gl.FUNC_ADD);
+    } else {
+        drawLineTiles(painter, tileManager, layer, coords, renderOptions, true, false, false);
+    }
```
CommanderStorm, line 220: Sort of also has artefacts if pixel peeing on the borders of lines and the max-self-oppacity prevention has a bit of the "not quite, but mostly correct look", not sure if there is "more correct" color blending that we could do here.
Maybe, but this is a big step forward also.

package-lock.json
HarelM, line 1: Have you changed any package? Please revert this if not.

src/webgl/value.ts
```diff
@@ -502,6 +502,20 @@ export class ColorAttachment extends FramebufferAttachment<WebGLTexture> {
     }
 }
 
+export class ColorAttachment1 extends FramebufferAttachment<WebGLTexture> {
```
HarelM, line 505: Can `ColorAttachment` be renamed to `ColorAttachment0` to match the `framebufferTexture2D` argument name?

src/webgl/draw/draw_line.ts
```diff
@@ -20,6 +20,7 @@ import type {OverscaledTileID} from '../../tile/tile_id.ts';
 import type {Tile} from '../../tile/tile.ts';
 import type {Context} from '../context.ts';
 import type {Framebuffer} from '../framebuffer.ts';
+import {ColorAttachment1} from '../value.ts';
```
HarelM, line 23: Nit: move this import up to where non-type only are imported.

HarelM:
In general, this looks good.
@lucaswoj can you take a look at the webgl related code, I don't think I have enough expertise to add meaningful comments.
The render tests looks good.

src/webgl/draw/draw_line.ts
CommanderStorm: this should be a formating rule, pass.
I is inconsistent previously too.

src/webgl/value.ts
CommanderStorm: It can, but that would mean that the color attachement 1 is now named 0..
pass.

HarelM:
Let's wait for @lucaswoj to review this.
I don't have other comments. If this adds the relevant functionality and we get good performance this can be shipped.

lucaswoj:
@HarelM Thanks for the ping. Before I dig into the WebGL code, I want to flag a bigger concern: I'm not sure this and #7490 are the right shape of fix for v6.

Together they're a significant breaking change to the style spec and they only change behavior for one layer type. This breaking change is especially odd for lines with a data-driven opacity. If two intersecting lines have different data-driven opacities, what's the opacity at the intersection? I think it's possible that a line with 99% opacity intersects a line with 1% opacity and is rendered at 1% opacity. That's going to break a huge number of data visualizations. 

A cleaner model is the `layer-opacity` concept in https://github.com/maplibre/maplibre-gl-js/issues/3062. It gives the same visual result, but allows users to opt-into the breaking behavior, makes the perf degradation opt-in, avoids breaking existing styles, the semantics stay clear for data-driven cases, and an easy path forward is established for other layer types.

I know I'm coming in late on a v6 milestone item and I don't want to derail momentum. But I think it's worth pausing to talk through alternatives before we lock this in.

HarelM:
It's a fair point, we are in the middle of v6, so we can revert changes still.
I'm not sure I'm fully align with the fear of breaking things that didn't work well before, but we can discuss this in the monthly meeting next week if you'd like.

lucaswoj:
I think there's a meaningful distinction between breaking things that *did* work and fixing things that *didn't*. The current behavior of overlapping translucent lines is, for some users, exactly the effect they want. Gaia GPS uses it to show density of public tracks. Mapeak stacks transparent lines to compose administrative boundary styles. Many basemaps rely on the additive feel of stacked transparent road casings. I don't think these users would describe today's behavior as broken.

For other users, the current behavior is genuinely unwanted, and the OSM main page is a great motivating example. I'm not arguing we should ignore them. But I'd love to find a path that gives them what they want without taking away what the first group depends on. That's what makes `layer-opacity` (or some opt-in equivalent) feel like the right shape to me.

HarelM:
They can use alpha channel for that. We're not talking away, we are refining. 
Also I've seen a few posts that talk about how they can't use line opacity due to this issue, and simply choose a different design, which is not great...

lucaswoj:
> They can use alpha channel for that

Users may be surprised to learn that `line-opacity` suddenly has different behavior from `line-color` and that they need to edit their styles to eliminate HSLA strings in `line-color` after migrating to v6 if they depend on the existing behavior.

In some cases, such as if users have `hsla` strings stored within their feature properties, that migration may be impossible without more significant engineering work.

And please don't gloss over the performance regression caused by this change or the undefined behavior where two lines intersect.

CommanderStorm:
The alternative approach of adding a top level oppaxity property which then is applied to each rendered layer sees rather hard again to explain though. How does top level oppaxity interact with line level oppaxity Vs color level oppaxity and what would be the difference?
I think that with this change we have a difference why there are two style properties..

We currently have two properties for this already
The 30% performance impact is here specifically only when the feature of transparent lines are used. There exists a fast path for each of the cases where we don't need the MRT or FBO.

lucaswoj:
> The alternative approach of adding a top level opacity property which then is applied to each rendered layer sees rather hard again to explain though. How does top level oppaxity interact with line level oppaxity Vs color level oppaxity and what would be the difference?

My proposal is that `line-color` and `line-opacity` keep their existing semantics (no breaking change, nothing to explain). A new `layer-opacity` opts into the new compositing behavior, layer-level. The diagram in https://github.com/mapbox/mapbox-gl-js/issues/4090#issuecomment-276799362 covers the cases. 

25a3f698-3d23-11e6-965b-2fe82a9ab5f3


> The 30% performance impact is here specifically only when the feature of transparent lines are used.

Two pieces here. First, per the #7490 description, the slowdown is 3x, not 30%. Second, "only when transparent lines are used" understates the blast radius. The fast path requires a constant `line-opacity` of exactly `1`. A quick survey of the OpenFreeMap public styles:

- Liberty: 4 line layers with data-driven `line-opacity`
- Bright: 9 line layers with data-driven `line-opacity`
- Positron: 1 line layer with data-driven `line-opacity`

Every one of those layers hits the offscreen MRT path, on every frame, by default. Users who never wanted the new behavior pay the cost unless they read the v6 release notes, understand this change, and manually rewrite their styles.

lucaswoj:
_EDIT: Switched to using the standard `Paint` benchmark instead of a one-off benchmark designed to stress line FBOs._

I tried OSM Americana against this PR. Paint time per frame in a typical z14 view goes from **2.3 ms → 8.7 ms**, a **3.8x slowdown**. 277 of its 301 line layers use feature-driven `line-opacity` expressions, which hit the offscreen MRT path.

| Bench | Pre-#7490 | Post-#7514 | Slowdown |
|---|---|---|---|
| `BasemapPaintAmericanaZ14` | 2.31 ms | 8.71 ms | **3.8x** |

Here's the benchmark I used:

```ts
import Paint from './paint.ts';

export class BasemapPaintAmericanaZ14 extends Paint {
    constructor() {
        super('https://americanamap.org/style.json', [{center: [-73.9857, 40.7484], zoom: 14}]);
    }
}
```

lucaswoj:
Tested across some more zoom levels:

Zoom | Pre-#7490 | Post-#7514 | Slowdown |
|---|---|---|---|
| 4 | 1.41 ms | 4.01 ms | 2.8x |
| 6 | 1.58 ms | 4.17 ms | 2.6x |
| 8 | 1.60 ms | 4.93 ms | 3.1x |
| 10 | 2.66 ms | 6.93 ms | 2.6x |
| 12 | 3.93 ms | 9.19 ms | 2.3x |
| 14 | 2.37 ms | 9.07 ms | 3.8x |
| 16 | 2.27 ms | 9.13 ms | 4.0x |
| 18 | 1.21 ms | 8.38 ms | 6.9x |
| 20 | 0.87 ms | 8.26 ms | 9.5x |

CommanderStorm:
that is interesting, because running the modified benchmark above only shows 30% (20ms -> 30ms).

I might not entirely understand how to run the benchmarks maybe.

CommanderStorm:
> americana

Well.. that seems to be an outlier in terms of how the style is written. CC @1ec5

Of 356 layers, 285 of 301 line layers have zoom-transitionable oppacity.
That is basically the benchmark you wrote..
So the performance impact is because they always do data driven styling..

For the other styles, I would not expect such a large impact given that americana is a bit out there in many metrics (complexity, ...)

lucaswoj:
Running this benchmark is a little tricky. Since it's not already included on `main`, we can't use `--compare`. Here's a little AI slop script that should walk you through it:

```
#!/usr/bin/env bash
# bench-americana.sh
# Run from inside any maplibre-gl-js git checkout. Stash any uncommitted work
# under test/bench/ first.

set -euo pipefail
cd "$(git rev-parse --show-toplevel)"

ZOOMS=(4 6 8 10 12 14 16 18 20)

if [ -n "$(git status --porcelain -- test/bench/)" ]; then
    echo "Uncommitted changes under test/bench/. Stash them first:"
    echo "    git stash push --include-untracked test/bench/"
    exit 1
fi

ORIG_REF=$(git symbolic-ref -q --short HEAD || git rev-parse HEAD)

echo "Fetching maplibre/maplibre-gl-js@main..."
git fetch --quiet https://github.com/maplibre/maplibre-gl-js.git main
MAIN_REF=$(git rev-parse FETCH_HEAD)

echo "Fetching CommanderStorm/maplibre-gl-js@line-oppacity-shading2 (#7514)..."
git fetch --quiet https://github.com/CommanderStorm/maplibre-gl-js.git line-oppacity-shading2
PR_REF=$(git rev-parse FETCH_HEAD)

echo "  main: $MAIN_REF"
echo "  pr:   $PR_REF"

SERVER_PID=""
if curl -fs --max-time 1 http://localhost:9966 >/dev/null 2>&1; then
    echo "Reusing existing server on :9966"
else
    echo "Starting static file server on :9966..."
    npm run start-server >/tmp/bench-americana-server.log 2>&1 &
    SERVER_PID=$!
    for _ in $(seq 1 60); do
        if curl -fs --max-time 1 http://localhost:9966 >/dev/null 2>&1; then break; fi
        sleep 0.5
    done
fi

cleanup() {
    set +e
    rm -f test/bench/benchmarks/basemap_paint.ts
    git checkout HEAD -- test/bench/versions/index.ts 2>/dev/null
    git checkout --quiet "$ORIG_REF" 2>/dev/null
    [ -n "$SERVER_PID" ] && kill "$SERVER_PID" 2>/dev/null && wait "$SERVER_PID" 2>/dev/null
}
trap cleanup EXIT

run_one() {
    local REF="$1" LABEL="$2"
    echo
    echo "===== $LABEL ====="
    git checkout --quiet "$REF"

    if grep -q "from '../lib/benchmark.ts'" test/bench/benchmarks/paint.ts; then EXT=".ts"; else EXT=""; fi

    {
        echo "import Paint from './paint${EXT}';"
        echo "const STYLE = 'https://americanamap.org/style.json';"
        echo "const CENTER: [number, number] = [-73.9857, 40.7484];"
        echo "class BPA extends Paint { constructor(zoom: number) { super(STYLE, [{center: CENTER, zoom}]); } }"
        for Z in "${ZOOMS[@]}"; do
            echo "export class BasemapPaintAmericanaZ${Z} extends BPA { constructor() { super(${Z}); } }"
        done
    } > test/bench/benchmarks/basemap_paint.ts

    EXT="$EXT" ZOOMS="${ZOOMS[*]}" python3 <<'PY'
import os, re
ext = os.environ['EXT']
zooms = os.environ['ZOOMS'].split()
classes = [f'BasemapPaintAmericanaZ{z}' for z in zooms]
imp = f"import {{{', '.join(classes)}}} from '../benchmarks/basemap_paint{ext}';"
regs = '\n'.join(f"register('{c}', new {c}());" for c in classes)
with open('test/bench/versions/index.ts') as f: s = f.read()
s = re.sub(r"(import \{Terrain3DGlobe[^\n]*\n)", lambda m: m.group(1) + imp + '\n', s, count=1)
s = re.sub(r"(register\('Terrain2DMercator'[^\n]*\n)", lambda m: m.group(1) + regs + '\n', s, count=1)
with open('test/bench/versions/index.ts', 'w') as f: f.write(s)
PY

    echo "Building..."
    npm run build-benchmarks >/dev/null 2>&1

    NAMES=()
    for Z in "${ZOOMS[@]}"; do NAMES+=("BasemapPaintAmericanaZ${Z}"); done
    npm run benchmark -- --compare=main "${NAMES[@]}" 2>&1 | tail -$((${#ZOOMS[@]} + 2))

    rm -f test/bench/benchmarks/basemap_paint.ts
    git checkout --quiet HEAD -- test/bench/versions/index.ts
}

run_one "$MAIN_REF" "main"
run_one "$PR_REF" "CommanderStorm:line-oppacity-shading2 (#7514)"

echo
echo "Done."
```

CommanderStorm:
> A new `layer-opacity` opts into the new compositing behavior, layer-level

Migrating to lumagl is a large body of work and something that I don't think I have the time for.
I think this is what you are asking..
Maybe this is simpler, but as far as I read migrating to the different blending modes like adding or subtracting is a loot of work..

So to maybe ask:
What is the difference between `line-opacity` and a color with an oppcicity? How SHOULD this work together?
3 different settigns to control the same thing seems very complicated..

lucaswoj:
> Migrating to lumagl is a large body of work and something that I don't think I have the time for.

Sorry for the confusion, `layer-opacity` (#3062, mapbox/mapbox-gl-js#4090) doesn't require luma.gl. It can be done with WebGL 1 features alone. The approach is the same as today's `raster-opacity`: render the layer to an offscreen FBO, then composite onto the main framebuffer at the requested opacity. It's also roughly analogous to CSS's `opacity`.

> What is the difference between line-opacity and a color with an opacity? 3 different settings to control the same thing seems very complicated.

There's no difference between `line-opacity` and the alpha channel of `line-color`. They've always been equivalent. It's a little redundant, but they're core features nearly every user of GL JS relies on, so we can't change them without breaking a lot of existing styles.

`layer-opacity` would be a new, opt-in property that works across all layer types. The layer renders as it does today, then the full result gets composited at the requested opacity. Users who want the OSM Carto effect set `layer-opacity` and pay the offscreen-FBO cost. Users who don't get no behavior change and no perf regression.

One more alternative worth weighing: rewrite the line-join geometry to eliminate overlapping triangles _within_ a feature, while keeping overlap _across_ features identical. That'd address the motivating OSM Carto case directly, with no breaking change, no opt-in, and likely minimal perf impact. I haven't dug into the join code yet so I can't say how hard it'd be.

lucaswoj:
> americana is a bit out there in many metrics

There are other styles that use enough data-driven `line-opacity` to experience measurable regressions. I've been telling folks for years to use data-driven styling to reduce their layer count for better performance! Another style that comes to mind is [Swisstopo](https://cms.geo.admin.ch/fmc/bmvar.html) (the official Swiss federal cartography product) which has 27 feature-driven `line-opacity` layers. I'm measuring up to a 2x regression on that style.

HarelM:
To be honest, I would like to avoid adding a new style spec property for this if possible.
@lucaswoj do you think there might be a way to solve this using smart stencil manipulation to improve the performance?

lucaswoj:
The more fundamental fix I mentioned doesn't involve the stencil buffer. The source of self-intersections is in how we build line geometry: at every corner and miter, we generate overlapping triangles. If we change the way line corners are constructed so the pieces don't overlap, the artifacts disappear with no extra rendering pass and minimal perf cost. I think that's worth investigating, but **I'd set expectations: this is a ~10-year-old known issue and nobody's tackled it, so it's probably harder than it sounds**.

Even if we fix the geometry, we'd only solve overlap within a single line feature. Overlap between different lines, and the broader "fade a whole layer" effect, still need something else. That's where I think `layer-opacity` is the right shape:

- It's a new feature, not a workaround. It's been on the wishlist for years (#3062, mapbox/mapbox-gl-js#4090) and would enable hover fades, layer toggle animations, atmospheric depth, and the OSM Carto flat-road effect with a single property.
- It's small to build. We already do the same off-screen rendering for `raster-opacity`. Extending it to other layer types is mostly plumbing.
- The perf cost only applies to users who opt in, the same way CSS `opacity` works.

I hear you on not wanting to grow the spec. But this isn't growing it to patch one bug. It's adding one well-understood property that covers this case plus a long wishlist of others, without breaking existing styles or imposing cost on users who don't opt in.

CommanderStorm:
> would be a new, opt-in property that works across all layer types

I can only introduce this one layer at a time, when I need it for capacity reasons.
Sorry.

lucaswoj:
> > would be a new, opt-in property that works across all layer types
> 
> I can only introduce this one layer at a time, when I need it for capacity reasons. Sorry.

No problem! It's not all on your shoulders. I'm happy to chip in some dev time once we agree on a path forward.

HarelM:
I believe we should start a discussion in the style-spec repo if we want to introduce a new feature to the spec.
From a spec perspective, I think it's a confusion addition. The fact that you need an image to explain the different possibilities is a smell that it's not straight forward. So I'll probably advocate against it. But maybe the community would think otherwise, and that's the beauty of it 😀
I'm sure you know that without a spec change this can't be introduced (it doesn't mean it can't be implemented and tested, just not merged).

HarelM:
As a side-note, while the geometry approach might solve the places where lines change direction and overlap itself I don't think it can solve places where the line "crosses itself".
The main two example in the first PR to address this will not be solved by geometry changes as far as I understand (correct me if I'm wrong):
- https://github.com/maplibre/maplibre-gl-js/pull/7490

Smart stencil logic might be an interesting approach to pursue, according to AI.

lucaswoj:
## On the Diagram

image

We're shipping this diagram no matter what!

The diagram shows two "flavors" of opacity: 

 - flavor A: (per-feature) opacity, each feature has its own opacity, then they blend
 - flavor B: per-layer opacity, composite the whole layer at one opacity

In this PR, `line-color` becomes flavor B, `line-opacity` becomes flavor A. Existing styles will, without warning, find themselves randomly, without warning, in some row of this diagram. Now that's confusing!

If we add `layer-opacity`, existing styles will be backwards compatible. They can opt into flavor B using a new property.

## On Community Signal

mapbox/mapbox-gl-js#4090 is the 5th most-commented issue on the mapbox-gl-js repo. That's the strongest data we have and it's strong. Our community isn't as old or large, so without some maintainer support behind a proposal here I don't think we'd get comparable independent data in our repo before v6 ships.

## On Fixing Tessellation 

You're right that re-tessellation only fixes overlap within a single feature (corners, miters, U-turns). It doesn't fix a line that crosses itself, and it doesn't fix two different features overlapping. Both of those still need flavor B opacity. I want to push back on the idea that tessellation fixes aren't worth pursuing though. It would fix many real-world cases (the OSM Carto motivating example, road casings with sharp bends) with zero spec change and lower perf cost.

## On Stencils

We could use a stencil if and only if we drop support for data-driven "flavor B" opacity. Stencils can make uniform opacity work but it can't compose two different alpha values at one pixel. A stencil-based #7490 would might have somewhat better perf but would stricter (`line-opacity` must be a constant).

lucaswoj:
## On Community Signal

Per-layer opacity has come up within the Maplibre issue tracker rather often

 - https://github.com/maplibre/maplibre-gl-js/issues/48 per-layer opacity is one case of the request for general bending mode support, but many comments in here are simply about per-layer opacity
 - https://github.com/maplibre/maplibre-gl-js/issues/4492 is a direct request for per-layer opacity
  - https://github.com/maplibre/maplibre-gl-js/issues/3062 is a direct request for per-layer opacity
  - https://github.com/maplibre/maplibre-gl-js/pull/1191 is a direct request for per-layer opacity

HarelM:
For upgrading between version 5 and version 6 we started a migration guide so people who are upgrading can read it and do the relevant changes if need be.
For the long run, I think it's better to keep the spec simpler and treat this as a bug fix instead of introducing a new property.

The main issue that I would like see solved is the following:
- #6334

This is, from my perspective, what users expect when they set line-opacity.

Having said all that, a performance degradation is a problem we can't dismiss.

CommanderStorm:
@HarelM lets just discuss this on our TSC.
No spec proposal nessary, the great graphic is fine to have a discussion on.

1ec5:
> For the other styles, I would not expect such a large impact given that americana is a bit out there in many metrics (complexity, ...)

osm-americana/openstreetmap-americana#995 (based on [this fork](https://github.com/aaroads-wiki/openstreetmap-americana/)) is much more representative of a typical production style, making heavier use of data-driven styling on the road layers for less complexity. We still haven’t optimized the railway layers to take advantage of #5812: osm-americana/openstreetmap-americana#526. But I suspect a significant performance hit there would be indicative of a broader problem.

I had @lucaswoj run the same benchmarks on this branch. He found that the impact was less severe but still notable. Here are the per-frame stats he posted [in Slack](https://osmus.slack.com/archives/C01V02K52UX/p1777838828136969):

main | #7514 | slow
----|----|----
Production Americana, z8 | 1.40 ms | 4.86 ms | 3.5x
Production Americana, z14 | 2.13 ms | 8.69 ms | 4.1x
Production Americana, z20 | 0.90 ms | 8.26 ms | 9.2x
Road rewrite (osm-americana/openstreetmap-americana#995), z14 | 1.68 ms | 4.87 ms | 2.9x
Road rewrite (osm-americana/openstreetmap-americana#995), z20 | 0.73 ms | 4.08 ms | 5.6x

Even though OSM Americana’s complexity seems comical, I’ve definitely seen more complex Mapbox styles in the wild and have no reason to believe it doesn’t exist among MapLibre styles too.

Another good data point would be the three MapLibre styles currently deployed on the osm.org homepage: Transport Map, Shortbread ([VersaTiles](https://github.com/versatiles-org/versatiles-style/) Colorful), and MapTiler OMT ([OSM OpenMapTiles](https://github.com/openmaptiles/openmaptiles/tree/master/style/)). OpenHistoricalMap also has [several styles](https://github.com/OpenHistoricalMap/map-styles/) [in production](https://embed.openhistoricalmap.org/). Some of these styles still have a lot of gaps that developers will want to fill in the future, so it’s important to have some slack, performance-wise.

Taking a step back, I hope my comments in https://github.com/maplibre/maplibre-gl-js/pull/7490#issuecomment-4299688041 aren’t taken as a carte blanche on the performance side. I think designers are desperate enough for a solution to that problem to be open to some performance degradation, but the tradeoff will depend on the particular stylesheet and application. Also, note that the problem isn’t just about overlaps; designers care just as much about the spillover effect, which is not addressed by this change.

xabbu42:
I'd like to support @lucaswoj that a layer-opacity would be the better fix (but I do not currently have the bandwidth to push this forward myself, so I just add my 2 cents here instead of opening an issue maplibre-style-spec). Consider the following use-case which uses all three properties in a natural way:

Visualizing a large number of gps tracks from different sport activities so that hotspots and mixed-use ways are clearly recognizable but the base map is never completely covered by the data. 

A very straightforward style to do that would use:
- a data-driven line-color without opacity to differentiate between the activities
- a constant line-opacity to control the range of opacity between rarely used ways and hotspots 
- a constant layer-opacity to control the maximum opacity and therefore how visible the basemap is behind hotspots

sjg-wdw:
Rather than slipping in an offscreen render step behind the scenes, why not make it explicit?
If you introduced the idea of render targets into the style spec, you could get this functionality very cleanly... and a lot more.

In response to a question of @nyurik's a couple years ago, I [sketched out a design](https://www.wetdogweather.com/2024/06/12/how-to-make-stamen-watercolor-in-maplibre/) for doing more complex multi-pass styles.  That's basically what you're doing here, but just for transparent lines.

Muti-pass rendering is how most games work, helpful for functionality like shadows, data rendering and definitely where things are going in the future.  You could get this improvement and open up a whole new area of cartographic experimentation.

CommanderStorm:
I have no clue what this would mean and I don't want to do render targets.
That seems fairly complicated and I have no clue how to implement that.

I don't want to add targets and operations and this sort of thing for a simple fix like this

sjg-wdw:
> I have no clue what this would mean and I don't want to do render targets. That seems fairly complicated and I have no clue how to implement that.
> 
> I don't want to add targets and operations and this sort of thing for a simple fix like this

You're already doing them.  That's what an offscreen buffer is.
The difference is you're hiding it from the developer.

Ignore the parts of my Stamen Watercolor post that discuss operations on the buffer.  That's just what would be necessary for effects.  The core idea is simple:
- Render a layer to an offscreen buffer, as you already are
- Take that buffer and render it to the screen with a color and opacity

The extra part is making it explicit in the style sheet.

lucaswoj:
+1 to render targets. They're effectively the same as what we've been calling "blending" or "comp-ops," which is a much larger discussion than `line-opacity`, but undoubtedly worth considering long term. See https://github.com/maplibre/maplibre-gl-js/issues/48 https://github.com/maplibre/maplibre-gl-js/pull/1191 https://github.com/mapbox/mapbox-gl-js/pull/380

sjg-wdw:
> +1 to render targets. They're effectively the same as what we've been calling "comp-ops," which is a much larger discussion than `line-opacity`, but undoubtedly worth considering long term. See #1191 [mapbox/mapbox-gl-js#380](https://github.com/mapbox/mapbox-gl-js/pull/380)

One of the reasons I run the Native Development Group for Amazon is I have a non-M*pbox perspective on map rendering.  I wasn't aware they'd discussed doing something similar and I'm going to avoid looking at it.  

The concept of render targets is something we use for weather rendering, but we didn't invent it.  It's a pretty basic concept, not likely to infringe anyone's IP.

lucaswoj:
Fair to keep your perspective independent of Mapbox. That said, there's a substantial MapLibre-side discussion on render targets and per-layer compositing that's worth engaging with before sketching new designs, none of which touches Mapbox prior art:

- maplibre/maplibre-gl-js#48
- maplibre/maplibre-gl-js#1191
- maplibre/maplibre#269
- maplibre/maplibre/discussions/170

Whatever shape we land on will be stronger if it builds on the tradeoffs the community has already worked through here.

CommanderStorm:
I still don't understand what a render target is for a map designer.
I undestand blending and that sort of stuff though.

I think that this PR seems to not have the support it would need to get merged.
So lets close this PR

sjg-wdw:
> I still don't understand what a render target is for a map designer.

Well it's a compromise for real-time rendering.  A cartographer might not give it much thought unless they were doing something very experimental (e.g. Stamen Watercolor).  It might normally just be an implied operation on a layer that requires a separate render step.  Mapnik did stuff like this under the hood if you picked the right operations.

The problem with just doing it by implication is the cost.  Setting up a separate rendering step is expensive and you really need to let the map designer control its resolution, how it's blended back in and what goes in to it.  If they care about performance at all, that is.

For image based maps, they did things like this quite a bit, I believe.  They'd form layers with a variety of content and then composite them after the fact.  In a sense, it's a return to older, richer techniques.

HarelM:
I don't think render target is the right approach. You, @sjg-wdw and @lucaswoj, are looking at render target like it's a second nature. Most users won't. I don't think the style spec should be overloaded with graphic engineering concepts. 
The target audience are not graphics engineers, the target audience are not even software engineers, so I'm finding this concept far from something that should be embedded in the style spec language.

1ec5:
For what it’s worth, I’ve mainly seen a desire for compositing operations from non-engineers used to similar operations in graphic _design_. The operations fit more naturally in a raster graphics environment than a vector one, so the requests will naturally come out of things like wanting to implement Watercolor with vector tiles.

Anyways, a general solution is much more substantial than what @CommanderStorm started out with. As far as I could tell from the monthly meeting this morning, a lot of the unease with the proposed change would be addressed by making behavior opt-in somehow. Is there a way to express the one case that he’s improving without necessarily having to implement the whole feature? Or is the whole feature ultimately easier to implement than making the one case intuitive to developers and designers?

HarelM:
Let's keep the conversation in the newly created discussion.

Issue #1602, opened 2022-09-01
Terrain elevation tile stiching


birkskyum:
https://stackblitz.com/edit/typescript-oepgsb?file=index.html

There appears to be an issue where the tiles further away are on a different zoom level, so when zooming out the tiles in the viewport doesn't update at the same time leading to inconsistencies in the mesh between tiles. 

In the example, deck.gl can be toggled to render the same terrain, and the issue is less pronounced there, so we might want a better strategy for requesting the elevation rgb tiles.

<img width="388" alt="Screen Shot 2022-09-01 at 23 28 04" src="https://user-images.githubusercontent.com/74932975/188016264-c6a56896-3bc3-453a-b972-b2c4c02245c9.png">

HarelM:
Are you sure this is not a data issue? The problem can be seen in both deck.gl and maplibre...

birkskyum:
I believe it's because they both suffer from the same issue with 3d tiles at different zoom levels not aligning.

prozessor13:
Yep, this happens on borders of different zoomlevels. This could be easy fixed, creating a small border in the 3d-mesh down to sea-level on every render-to-texture tile.

current mesh:
![IMG-2535](https://user-images.githubusercontent.com/45103/188307692-c61d1edc-e2c9-4f93-b134-037b3220e9ce.JPG)

mesh with border to sea-level
![IMG-2536](https://user-images.githubusercontent.com/45103/188307708-d80e5e44-a052-4443-a935-4b5b0e529a81.JPG)

HarelM:
I wouldn't go to sea level due to dead sea problem being under the sea level, but sure, we could create a surface with a few (hunderd) meters at the edge I guess...
I have no clue how to do it, but this sounds like it should solve the problem :-)

birkskyum:
because of the elevation offset hack the dead sea is actually rendered above sea level

prozessor13:
I sayed down to *sea-level* but i am wrong, something like 10 pixels should be enough, every pixel more may costs performance.

birkskyum:
There's also the option to request more tiles of the same zoom level, depending on how expensive that would be.

HarelM:
When the pitch is allowing you to see many tiles this becomes problematic... The original reason for different zoom level tiles is performance :-)

PR #7523, opened 2026-04-25
Add option to disable terrain skirts

### LINKED ISSUE #2967: artifacts at tile borders wtih terrain 3D
When vizualizing a map with terrain there are artifacts at tile borders. It looks like the terrain lays at lower elevations below the map. I guess this is just visible if we use basemaps which are not covering the entire viewport where we can look "below" the map. 

While I experience this with the latest release, it seems not to be the case with maplibre gl js 2.4. (these are the only versions I tested).

Example:

With maplibre latest
maplibre_error

With maplibre 2.4
maplibre_24

The purple areas in both images are due to tiny differences between terrain and vector tiles at the border I guess.

```html
<html>
    <head>
     <meta charset="utf-8"/>
     <title> terrain test </title>
     <meta name="viewport" content="initial-scale=1,maximum-scale=1,user-scalable=no" />
    
     <script src="https://unpkg.com/maplibre-gl@2.4.0/dist/maplibre-gl.js"></script>
     <link href="https://unpkg.com/maplibre-gl@2.4.0/dist/maplibre-gl.css" rel="stylesheet"/>

     <!-- 
     <script src='https://unpkg.com/maplibre-gl@latest/dist/maplibre-gl.js'></script>
     <link href='https://unpkg.com/maplibre-gl@latest/dist/maplibre-gl.css' rel='stylesheet' />
       -->

     <style>
         body { margin:0; padding:0; }
         #map { position:absolute; top:0; bottom:0; width:100%; }
     </style>
   </head>
  
   <body>
      <div id="map"></div> 
      <script>
          var map = (window.map = new maplibregl.Map({
            container: 'map',
            zoom: 11,
            center: [9.61,47.5761],
            pitch: 85,
            bearing: -67,
            hash: true,
            style: 'https://web3d.basemap.de/maplibre/bm_web_col.json', 
            maxZoom: 18,
            maxPitch: 85
         }));
            
            map.addControl(
               new maplibregl.NavigationControl({
                  visualizePitch: true,
                  showZoom: true,
                  showCompass: true
               })
            );
            
            map.addControl(
               new maplibregl.TerrainControl({
                  source: 'terrainSource',
                  exaggeration: 1
               })
            );
      </script>
   </body>
  </html>
```

### LINKED ISSUE #4991: Incorrect semi-transparent tile rendering at terrain tile borders
**maplibre-gl-js version**: 4.7.1, 5.0.0-pre.5

**browser**: Chrome

### Steps to Trigger Behavior

1. Create a map with semi-transparent raster tiles
2. Enable terrain (may be zero-elevation tiles)
3. Pitch the map

### Link to Demonstration

Raster tiles:
https://jsfiddle.net/gbc64wm5/

Background layer:
https://jsfiddle.net/hb5e1rj6/

### Expected Behavior

Rendered image is seamless across the border between terrain tiles.

### Actual Behavior

It looks like a "cliff" at the edge of the terrain tile is being rendered in the imagery.

image

### LINKED ISSUE #6920: Vertical “edge wall” artifact at tile boundary when adjacent tile is not loaded (appears when pitch != 0)
Hi,
I’m using **MapLibre GL JS v5.15.0** inside a **Flutter WebView** environment. In some cases I’m seeing a rendering artifact on the border between a **loaded terrain tile** and a **not-yet-loaded tile** - only when the map is tilted..

### Problem description

At the boundary between tiles, when the neighboring tile is not loaded yet, a visible **vertical “wall” / border** appears — **as if the last row of pixels is extruded downwards to zero elevation**. This artifact occurs when pitch != 0. It looks like a rendering glitch.
Once the missing tile is loaded, the wall disappears.
Screenshot example:
Image

### Steps to Trigger Behavior
1. Tilt the map
2. Pan the map toward an area where neighboring tiles are not yet loaded.
3. Observe the vertical “edge wall” at the tile boundary.

### Expected behavior
If a neighboring tile is not yet available, the map should ideally keep the previous geometry/height
instead of rendering a vertical drop-off

### Question
Is there any way to disable this “edge wall” effect?

Any hints would be greatly appreciated. Thanks in advance!

safwat-halaby:
Allow maps with transparent backgrounds and terrain enabled to avoid seam-line artifacts by disabling terrain tile edge extensions. The new option is called `terrainSkirtLength` (formerly `disableTerrainVerticalExtensions`/`disableTerrainSkirts` in outdated commits). See issue #4991.

Docs description:

> Controls the length of the vertical extensions which are added to the edges of terrain tiles.
> It is recommended to set this to "none" only when using a transparent background along with an elevation terrain.
>
> - `"none"` disables skirts entirely, eliminating vertical artifacts that occur on tile boundaries when a transparent background is used along with a terrain, but might introduce some horizontal hairline gaps (stitches).
> - `"auto"` renders skirts to eliminate horizontal hairline gaps (stitches) that might occur on tile boundaries with different zoomlevels for some terrain datasets, but introduces vertical artifacts if using a transparent background along with a terrain.
>
> Currently, this parameter is a tradeoff. Future rendering techniques may eliminate the need for it.
> @defaultValue "auto"

**Before (MapLibre 5.24.0)**  - Try it out [live](https://safwat-halaby.com/apps/vertical-artifacts/release.html) - Screenshot:

before-5 24 0

**After (Commit 0d3f078f5)**  - Try it out [live](https://safwat-halaby.com/apps/vertical-artifacts/0d3f078f5.html) - Screenshot:

after-0d3f078f5

## Launch Checklist

 - [x] Link to related issues.


safwat-halaby:
I have not included any tests. Does this minor change warrant writing tests?

@HarelM - I changed the name of the parameter as [discussed](https://github.com/maplibre/maplibre-gl-js/issues/4991#issuecomment-4231763384). Is the new name (`disableTerrainVerticalExtensions`) ok?

src/render/terrain.ts
```diff
@@ -452,39 +457,41 @@ export class Terrain {
...
-        const offsetRight = offsetLeft + (meshSize + 1) * 2;
-        for (const x of [0, 1]) for (let y = 0; y <= meshSize; y++) for (const z of [0, 1]) {
-            vertexArray.emplaceBack(x * EXTENT, y * delta, z);
-        }
-        for (let y = 0; y < meshSize * 2; y += 2) {
-            indexArray.emplaceBack(offsetLeft + y, offsetLeft + y + 1, offsetLeft + y + 3);
-            indexArray.emplaceBack(offsetLeft + y, offsetLeft + y + 3, offsetLeft + y + 2);
-            indexArray.emplaceBack(offsetRight + y, offsetRight + y + 3, offsetRight + y + 1);
-            indexArray.emplaceBack(offsetRight + y, offsetRight + y + 2, offsetRight + y + 3);
+        if (!this._disableTerrainVerticalExtensions) {
```
HarelM: Can you move this part to a separate method?

HarelM:
Can you either add a render test to cover this functionality or a unit test please? 

Thanks for the work around this!

test/integration/assets/geojson/geometry-on-terrain.json
```diff
@@ -0,0 +1,8 @@
+{
+"type": "FeatureCollection",
+"description": "A geometry and a line geographically placed at the same location as the 'assets/tiles/terrain/10-189-401.terrain' tile and its neighbors",
```
safwat-halaby: typo. Should be a "polygon" and a line

safwat-halaby:
I added render tests. The tests fail if either the line or the polygon draw/don't draw a vertical extension when expected. 

Nitpick: Are you sure you don't want to go with `skirt`? It's much shorter than `vertical extension` and it's more descriptive in my opinion! It really does look like each tile wears a skirt. I am rather indifferent on this though, and `verticalExtension` is also fine.

HarelM:
I've consulted with AI and skirt is a known graphics term, apparently.
Here's an example from cesium docs:
https://cesium.com/learn/cesiumjs/ref-doc/Cesium3DTilesTerrainData.html

Feel free to use that if you like it better.

src/ui/map.ts
```diff
@@ -80,7 +80,7 @@ export type MapOptions = {
     /**
      * If `true`, the map's position (zoom, center latitude, center longitude, bearing, and pitch) will be synced with the hash fragment of the page's URL.
      * For example, `https://example.com#2.59/39.26/53.07/-24.1/60`.
-     *
```
safwat-halaby, line 88: stray whitespace

safwat-halaby:
Renamed to skirt. Moved the skirt construction into a separate function. Awaiting feedback!

test/integration/render/tests/terrain/skirts-disabled/style.json
```diff
@@ -0,0 +1,72 @@
+{
+  "version": 8,
+  "metadata": {
+    "test": {
+      "description": "Terrain with skirts disabled to avoid vertical artifacts when using transparent terrains.",
+      "height": 512,
+      "width": 512,
+      "disableTerrainSkirts": true,
+      "threshold": 0.05,
+      "note_on_threshold": "Since we are dealing with transparent lines and polygons, we need a sensitive threshold. Somewhere between 0.11 and 0.15 overlaping polygon parts no longer qualify as different from non overlaps. So ~0.11 is the safest maximum threshold."
```
safwat-halaby: Typo. Should be "Maximum safe threshold"

src/render/terrain.ts
```diff
@@ -544,4 +518,41 @@ export class Terrain {
             mercatorY
         };
     }
+
+    _buildSkirts(vertexArray: Pos3dArray, indexArray: TriangleIndexArray, meshSize: number, delta: number, northPole: boolean, southPole: boolean) {
+        // add an extra frame around the mesh to avoid hairline gaps (stitching) on tile boundaries with different zoomlevels
```
HarelM: Move this comment to be a tsdoc comment, I think it would be better.

src/ui/map.ts
```diff
@@ -392,6 +392,16 @@ export type MapOptions = {
...
+    /**
+     * Removes vertical artifacts that occur at tile seam lines when a transparent background is used along with an elevation terrain.
+     * Currently, this parameter is a tradeoff; Setting it to true disables the vertical extensions which are added to the edges of terrain tiles.
+     * This removes the visually unappealing vertical artifacts, but might introduce horizontal hairline gaps (stitches) that occur when viewing some terrain datasets from a high-pitch angle.
+     * It is recommended to set this to true only when one is using a transparent background along with a terrain.
+     * There is no reason to set this to true if the background tiles are opaque or if no terrain is being used.
+     * Future rendering techniques may eliminate the need for this parameter.
+     * @defaultValue false
+     */
+    disableTerrainSkirts?: boolean;
```
HarelM: Can we make this parameter more genetic maybe, like in Cesium?
i.e. have it receive a number, and in case of 0 it will not render the skirt.
I'm not sure what the current size is, but it's hard coded, and maybe someone will need it to be "deeper".

test/integration/assets/geojson/geometry-on-terrain.json
```diff
@@ -0,0 +1,8 @@
+{
```
HarelM, line 1: Can this be formatted to look like a json? It's hard to read it this way...

src/ui/map.ts
safwat-halaby: It's currently dictated by `getMeshFrameDelta` in `terrain.ts`:

```ts
getMeshFrameDelta(zoom: number) {
        // divide by 5 is evaluated by trial & error to get a frame in the right height
        return 2 * Math.PI * earthRadius / Math.pow(2, zoom) / 5;
    }
```

I am not entirely sure what the use case is (a dataset with severe terrain discrepancies between zoom levels or extreme quantization?) . Can you think of reasonable use cases?

But addition is trivial and we could go with that.

HarelM: I can be creative in thinking about a use cases :-) for example a game similar to minecraft that will utilize the terrain features of maplibre, but inject a DEM that contains boxes and not a "regular" surface.
Also someone is using terrain to show painting's surface or something, so terrain might not be something that only represents real world terrain data.
I think that making this parameter a number instead of boolean, assuming as you said it's trivial, will future-proof it better.

safwat-halaby: There is a full sprite on each face of a Minecraft cube, but the skirts would be made of vertical stripes, each column of pixels having only one color, which would be aesthetically jarring! I think the fact that the skirts are aesthetically unappealing limits their practical use to be short "cliffs" in unaligned datasets, but maybe I am not being imaginative enough.

For executing this, we'd need to rename `disableTerrainSkirts` to `terrainSkirtLength`. I'm wondering which units we should use.

Meters would be awkward; A good skirt on high zoom would be barely visible on low zoom. A good skirt on low zoom would be huge on high zoom.

I think the most natural would be a unit relative to the tile sizes at the current zoom level. It could be a ratio. For example, `terrainSkirtLength: 1` would mean that the skirt is as big as a tile at the current zoom level, and `0.5` means it is half the size of the tile at the current zoom level.

_What follows are preliminarily findings that might turn out different once I investigate more:_

I think the above unit is very easy to implement with one caveat. In fact, if I am understanding the code right, this is precisely what `getMeshFrameDelta` currently does, except that the ratio is currently hardcoded to be 1/5 of the tile size. 

And here's the caveat:  I think that when zooming out to a partial zoom (say, from 12 to 11.9), the skirt will temporarily outgrow its 1/5 ratio until the lower zoom tiles are loaded, because Math.pow(2, zoom) assumes that 11.9 tiles are slightly bigger tiles but the map is just showing 12 tiles. This can be mitigated by playing around with ceiling/flooring the zooms in the formula. But it will always be imperfect, because as far as I am aware, there may be tiles from DIFFERENT zoom levels at the map at the same time. So I think there are 3 options:
1. Accepting the fact that when the user specifies `terrainSkirtLength`, the actual length will wobble around the specified value.
2. Rounding/ceiling/flooring the zoom level, making the terrain skirt length follow `terrainSkirtLength` in more situations, but when there are a low-level and higher-level zoom tiles on the map, the low level one would have the skirt length of the higher level one (shorter than its usual)
3. Refactoring such that skirt length can be tweaked per tile. This is not worth the effort and code complexity in my opinion!

I don't see drawbacks to `2` currently. But maybe I missed something?

safwat-halaby: The more I think about this, the more I think it is unnecessary.

In Cesium it is extremely common to reach the edge of a 3dTiles model. Furthermore, it's not uncommon to have something rendered below the 3dTiles model (e.g. a 2d background). For these workloads, tweaking the skirt length of the 3dTiles model makes more sense in my mind.

But it's different for MapLibre

* When the background map is semi-transparent, the skirts are aesthetically horrible and the only sensible length is 0.
* When the background map is opaque, their length doesn't really matter as long as it's long enough (and not too long to look weird while tiles are loading) - there is very little value in letting the user tweak that. Give them a good builtin and they're good to go.
* Reaching the edge of the background map is not as common, and even then, there's nothing beneath to justify a play with the skirt length.
* It is always trivial to **add** a `skirtLength` parameter later but removing would be a challenge. Once we add it, we're married to the current implementation of constant-length skirt, and a future two-pass renderer or a smarter skirt mechanism would be API breaking.

Therefore I suggest postponing the decision to expose a skirt length parameter unless there is clear justification in terms of use cases.

HarelM: Having two parameters for the same thing in the future will also not be great.
But since I don't see anyone actually using this length parameter it might be fine to keep it as a boolean for now.
Can you negate the definition please - i.e. `enableTerrainSkirts` which is `true` by default?

safwat-halaby: How about an optional `skirtLength` which can currently only be 0 or the string `default` (other values are ignored and considered 'default'). Type is `number | "default"` or even `0 | "default"`

It'll likely stay like this forever but is extensible without API breakage if ever needed, no two parameters, no marriage to current implementation.

safwat-halaby: Maybe `number | "dynamic"` or `number | "auto"` is slightly more descriptive as "default" implies a constant.

HarelM: If we add another parameter it misses the point of having just one parameter for this settings. 
In any case, it's not part of the style spec, so we can decide to break it in a breaking change version if we ever will need to. 
Boolean parameter is probably enough for now...

safwat-halaby: I am suggesting only ONE optional parameter. No `enableTerrainSkirts` and no `disableTerrainSkirts`, only `skirtLength`.

Its value is either `0` or `default/auto/dynamic`. It currently plays the role of `disableTerrainSkirts`. It allows us to open it up for arbitrary numbers like 3 or 10 if we ever decide this is needed.

HarelM: Making it a number but not supporting anything other than 0 might be a confusing DX.
what about making is a string like css style width where you could write "20% for example. 
Keeping it as a string but currently only support "none" and "auto" will solve your requirement and future proof it, right?

safwat-halaby: I agree. That sounds better.

It future-proofs for multiple units e.g. `pixels/meters/tile ratios`, and it also future-proofs non-numeric strategies other than the current one, similar to how CSS supports e.g. `16px / auto / inherit` in the same field.

Let's agree on the specific wording.

```
terrainSkirtLength: "none" | "auto"
```

Sound good?

HarelM: Yup.

safwat-halaby:
I pushed some changes.

- `terrainSkirtLength` changes performed as discussed.

Smaller touches:

- Renamed `getMeshFrameDelta` to `getSkirtLength`.
- Formatted the geojson.
- Moved the `_buildSkirts` comment into tsdoc.
- Fixed a typo in the threshold commentary in unit test.

src/ui/map.ts
safwat-halaby: Done! See latest commits.

safwat-halaby:
I consider the PR ready for merge.

Let me know if there are more comments!

HarelM (approved):

HarelM:
Can you resolve the conflicts and I'll merge this?

HarelM:
Sorry, refactored the render tests to use vitest instead of the code that was written here...

safwat-halaby:
I am getting an error when running render tests, even on the main branch.

```
xvfb-run -a npm run test-render

> maplibre-gl@6.0.0-5 test-render
> vitest test/integration/render/render.test.ts


 DEV  v4.1.5 <REDACTED>/maplibre-gl-js


⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Unhandled Rejection ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
Error: ENOSPC: System limit for number of file watchers reached, watch '<REDACTED>/maplibre-gl-js/.cache/15021683862997205462'
 ❯ FSWatcher.<computed> node:internal/fs/watchers:254:19
 ❯ Object.watch node:fs:2539:36
 ❯ createFsWatchInstance node_modules/vite/dist/node/chunks/node.js:8497:16
 ❯ setFsWatchListener node_modules/vite/dist/node/chunks/node.js:8539:14
 ❯ NodeFsHandler._watchWithNodeFs node_modules/vite/dist/node/chunks/node.js:8653:20
 ❯ NodeFsHandler._handleFile node_modules/vite/dist/node/chunks/node.js:8697:24
 ❯ NodeFsHandler._addToNodeFs node_modules/vite/dist/node/chunks/node.js:8871:26

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯
Serialized Error: { errno: -28, syscall: 'watch', code: 'ENOSPC', path: '<REDACTED>/maplibre-gl-js/.cache/15021683862997205462', filename: '<REDACTED>/maplibre-gl-js/.cache/15021683862997205462' }




 Test Files 0 passed (0)
      Tests 0 passed (0)
   Start at 
   Duration 0ms
```

safwat-halaby:
Conflicts resolved and CI tests passed (though I am no longer able to run local tests after the changes to the tests on main)

HarelM:
Make sure you run npm install, npm run build-dist etc, as you can see, CI is able to run.

safwat-halaby:
For the record, this PR resolves at least:

#2967 #4991 #6920

Highly relevant issue is #1602

HarelM:
Thanks! I linked all the relevant issue to this PR.

PR #7528, opened 2026-04-27
fix: prevent double rendering of features crossing the antimeridian

### LINKED ISSUE #6248: GeoJSON layers partially rendered twice with globe projection

pabueco:
Together with https://github.com/maplibre/geojson-vt/pull/112 this PR fixes https://github.com/maplibre/maplibre-gl-js/issues/6248, by disabling the wrapping of features (and splitting them instead) in `geojson-vt` via `worldCopies:false` and filtering out the edges created by the splitting when preparing the geometry in the worker.

As mentioned in the other PRs, this is not my area of expertise and this PR was heavily assisted by AI (although I did my best to explore alternatives and verify decisions during the past weeks), so let me know if there are problems with this approach or anything to rethink. 
While this works great in my limited testing, there might be cases I'm not aware of, so I hope this can at the very least serve as a starting point to getting the double rendering issue solved.

To test this you can clone [this repo](https://github.com/pabueco/maplibre-antimeridian-test), update the link to your local maplibre instance, serve the HTML file and play around with it. Notice that features are never rendered double or overlap at the antimeridian; also there are no visible artefacts at the antimeridian. You can also switch to mercator projection to verify nothing has changed there.

Note, checks are failing, because we rely on the `geojson-vt` changes made [here](https://github.com/maplibre/geojson-vt/pull/112).

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

Assisted-By: Claude Opus 4.7; Used for the core logic (edge detection, splitting, filter), but everything was carefully reviewed (as least as far as I was able to) and iterated on many times.

### DIFF
--- a/src/data/bucket.ts
+++ b/src/data/bucket.ts
@@ -21,13 +21,6 @@ export type BucketParameters<Layer extends TypedStyleLayer> = {
     collisionBoxArray: CollisionBoxArray;
     sourceLayerIndex: number;
     sourceID: string;
-    /**
-     * Whether the source wraps features into duplicate world copies. Only
-     * `false` when the active projection disallows world copies (globe,
-     * vertical perspective) AND the source is GeoJSON — in which case features
-     * are clipped at the antimeridian and buckets must suppress the resulting
-     * synthetic outline edges. Defaults to `true` when omitted.
-     */
     worldCopies?: boolean;
 };
 
--- a/src/data/bucket.ts
+++ b/src/data/bucket.ts
@@ -21,7 +21,6 @@ export type BucketParameters<Layer extends TypedStyleLayer> = {
     collisionBoxArray: CollisionBoxArray;
     sourceLayerIndex: number;
     sourceID: string;
-    worldCopies?: boolean;
 };
 
 export type PopulateParameters = {
--- a/src/data/bucket.ts
+++ b/src/data/bucket.ts
@@ -21,6 +21,14 @@ export type BucketParameters<Layer extends TypedStyleLayer> = {
     collisionBoxArray: CollisionBoxArray;
     sourceLayerIndex: number;
     sourceID: string;
+    /**
+     * Whether the source wraps features into duplicate world copies. Only
+     * `false` when the active projection disallows world copies (globe,
+     * vertical perspective) AND the source is GeoJSON — in which case features
+     * are clipped at the antimeridian and buckets must suppress the resulting
+     * synthetic outline edges. Defaults to `true` when omitted.
+     */
+    worldCopies?: boolean;
 };
 
 export type PopulateParameters = {
--- a/src/data/bucket/antimeridian_bucket_features.test.ts
+++ b/src/data/bucket/antimeridian_bucket_features.test.ts
@@ -10,27 +10,27 @@ describe('getAntimeridianEdgePredicate', () => {
     });
 
     test('left-edge tile (x === 0) flags x=0 edges only', () => {
-        const pred = getAntimeridianEdgePredicate(new CanonicalTileID(2, 0, 1));
-        expect(pred).not.toBeNull();
-        expect(pred(0, 0)).toBe(true);
-        expect(pred(0, 10)).toBe(false);
-        expect(pred(10, 0)).toBe(false);
-        expect(pred(EXTENT, EXTENT)).toBe(false);
+        const isEdge = getAntimeridianEdgePredicate(new CanonicalTileID(2, 0, 1));
+        expect(isEdge).not.toBeNull();
+        expect(isEdge(0, 0)).toBe(true);
+        expect(isEdge(0, 10)).toBe(false);
+        expect(isEdge(10, 0)).toBe(false);
+        expect(isEdge(EXTENT, EXTENT)).toBe(false);
     });
 
     test('right-edge tile (x === (1<<z) - 1) flags x=EXTENT edges only', () => {
-        const pred = getAntimeridianEdgePredicate(new CanonicalTileID(2, 3, 1));
-        expect(pred).not.toBeNull();
-        expect(pred(EXTENT, EXTENT)).toBe(true);
-        expect(pred(EXTENT, 10)).toBe(false);
-        expect(pred(0, 0)).toBe(false);
+        const isEdge = getAntimeridianEdgePredicate(new CanonicalTileID(2, 3, 1));
+        expect(isEdge).not.toBeNull();
+        expect(isEdge(EXTENT, EXTENT)).toBe(true);
+        expect(isEdge(EXTENT, 10)).toBe(false);
+        expect(isEdge(0, 0)).toBe(false);
     });
 
     test('zoom 0 tile is both the left and right edge', () => {
-        const pred = getAntimeridianEdgePredicate(new CanonicalTileID(0, 0, 0));
-        expect(pred).not.toBeNull();
-        expect(pred(0, 0)).toBe(true);
-        expect(pred(EXTENT, EXTENT)).toBe(true);
-        expect(pred(0, EXTENT)).toBe(false);
+        const isEdge = getAntimeridianEdgePredicate(new CanonicalTileID(0, 0, 0));
+        expect(isEdge).not.toBeNull();
+        expect(isEdge(0, 0)).toBe(true);
+        expect(isEdge(EXTENT, EXTENT)).toBe(true);
+        expect(isEdge(0, EXTENT)).toBe(false);
     });
 });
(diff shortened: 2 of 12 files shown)


HarelM:
This PR is confusing... Should this one be kept open and the other one closed?
- #6970

Also see my comment here about tile buffer:
https://github.com/maplibre/maplibre-gl-js/pull/6970#issuecomment-4329995986

pabueco:
@HarelM Yes, closed the other one now. 

What exactly do you find confusing about this PR? I realized that I didn't give much context around the changes here, so I'm happy to elaborate if that is the reason. Or do you have a general concern about the approach chosen here?

HarelM:
Confusing was the fact that there was a previous PR that should've solved the same issue, that's all, wasn't sure which one is the relevant one.

pabueco:
Regarding tour comment on the tile buffer:

> Is it possible the the clipping is done in geojson-vt on the tile border instead of taking into account the buffer around the tile? If so, this might solve this...?
geojson-vt should mimic the regular vector tiles behavior, does this bug happens on regular vector tiles?

I'm not sure I entirely understand what you mean, but I tried integrating the buffer when clipping the features in `geojson-vt` (just like it is when wrapping), however that just caused the exact visual result we had initially with the wrapping: the features being duplicated left and right of the antimeridian, with the size of the overlap depending on the buffer size. Or did you mean something else with 'taking into account the buffer around the tile'?

When it comes to vector tiles I haven't seen this behaviour and I would imagine a lot more bug reports if that were the case, or maybe I'm misunderstanding.

pabueco:
> Confusing was the fact that there was a previous PR that should've solved the same issue, that's all, wasn't sure which one is the relevant one.

Ah right, yea sorry for that. Just to make sure we're on the same page here: The previous (now closed) PR solved the issue within the constraint of not being able to make changes to `geojson-vt` by essentially just preventing the same pixel being drawn on twice. With the wrapping fix in `geojson-vt` we don't need that, we just need to remove the artefacts caused by now splitting the features.

HarelM:
> I'm not sure I entirely understand what you mean, but I tried integrating the buffer when clipping the features in geojson-vt (just like it is when wrapping), however that just caused the exact visual result we had initially with the wrapping: the features being duplicated left and right of the antimeridian, with the size of the overlap depending on the buffer size. Or did you mean something else with 'taking into account the buffer around the tile'?

I think you got it, this is what I meant with the buffer.
The way to avoid the border of a polygon/area from appearing inside a tile it shouldn't appear in is by using a tile buffer - this allows the border to be drawn outside the tile and then getting clipped when the tile is rendered.

So I would expect this problem to be solved the same way for "regular" vector tiles and for geojson as well assuming we do the same operation as one would do when generating tiles.

I know this PR solves this issue, but I would like to avoid this kind of solution where we assume there's a GEOJSONVT_ANTIMERIDIAN_CLIP property, if possible.

I know that I'm asking a lot, but I would generally like to just make sure we create tiles from geojson similar to what's done for regular vector tiles, this way we avoid all kind of bugs that were already solved for vector tiles.

pabueco:
Okay, I think I understand what you're getting at, also of course makes sense to prevent workarounds like this if possible.

What I struggle with when it comes to the buffer and tile clipping, is why this double rendering artefact only appears when turning the globe a certain way (zooming out quite far or viewing the poles). Like everything seems to generally work as expected, it's not like a part of the pipeline is just missing for the globe or geojson sources. 
I assume this is has something to do with the tile 'layout' or how the world is split up, but enabling `showTileBoundaries` doesn't really reveal anything, at least to me.
Do you have any ideas or insights here why this might be? I'd be happy to look into better solution or figure out what the differences are for 'regular' vector tiles and geojson based ones, but feel like it take less time with some pointers (even if they are just guesses). 

Regarding the `GEOJSONVT_ANTIMERIDIAN_CLIP` property: I initially didn't do this and just removed all edges that are exactly on the tile border at the antimeridian, but figured it would be safer to only do this for tagged features. Just for context, I assume that wouldn't change much about this PR in general.

HarelM:
Is this happening only on globe? Because I think you showed it on mercator as well.
I would focus on mercator to make sure everything runs well and only then look into globe and related "edge" cases.
I would also try and take the geojson and serve it using tippecanoe or a similar tile server that can convert geojson into tiles. This way you can reduce some noise and compare if the issue is with the geojson-vt code or maplibre vector rendering.
If the problem is in maplibre, and not geojson-vt then it needs to be solved, but you can rest assure the code in geojson-vt is not the root cause of it, and you can then just use the tiles generated from tippecnoe to debug/and reproduce.
That's what I would try and do - make a minimal example, and try to see where things differ between the two code flows. The simpler the example and faster to reproduce the easier it will be for you to debug it.
You can also write some render tests to speed up the feedback cycle.
I hope that helps.

pabueco:
It almost definitely only happens on globe, and only at certain angles. I haven't seen it on a mercator maps yet, also wasn't able to reproduce it there during testing. 

Thanks for the tippecanoe idea, I'll take a look.

pabueco:
@HarelM I followed your suggestions and did both experiments (mercator-only reproduction and serving the same GeoJSON through tippecanoe) and you were right on both counts. I'm pretty sure I now understand the actual root cause. 
Summarizing the findings here, because they change the direction quite a bit:

1. The bug is globe-only and, more specifically, zoom-0-only. With the same data and camera the doubling only ever appears when the covering tiles includethe canonical z0 tile (zoomed far out, or when the LOD picks z0). At z1+ everything renders correctly. I could not reproduce anything on mercator.

2. Regular vector tiles have exactly the same problem, so geojson-vt is not the root cause. I ran the repro polygons through tippecanoe and decoded the z0 tile, which shows the intentionally duplicated geometry near the antimeridian into the z0 tile buffer, shifted by +/-360 degrees (just like geojson-vt's wrap). Serving these tippecanoe tiles to maplibre produces the identical double rendering on globe, without geojson-vt being involved.

**What and why**: the buffer duplication is harmless under mercator, because buffer geometry either lands outside the world copy (where the per-tile stencil clipping mask discards it) or on top of an adjacent world copy. 
Under globe the projection of in-tile coordinates to the sphere is periodic in X, so buffer geometry past +/-180 degrees wraps around the planet. At z1+ it lands inside a different tile's stencil mask and is clipped, as intended. But the z0 tile covers the whole world, so its own buffer wraps around and lands back on the tile itself, inside its own stencil mask. The original and the +/-360 degrees copy end up on the same pixels within the same tile, so the screen-space stencil can't tell them apart. That's also why my earlier stencil approach in https://github.com/maplibre/maplibre-gl-js/pull/6970 had to resort to "draw each pixel once per layer", which had other downsides.

**Consequence for this PR**: the `worldCopies: false` direction is the wrong level, as you suspected, as it only fixes geojson sources (vector tiles stay 'broken').

I've opened https://github.com/maplibre/maplibre-gl-js/pull/7896 which fixes this at the renderer level instead: the z0 tile is clipped to its extent in the fragment shader (geometry-space, since screen-space can't distinguish the overlapping fragments), mirroring what the stencil masks do at every other tile border. It fixes geojson and vector tile sources, needs no changes to geojson-vt, and no coupling like `GEOJSONVT_ANTIMERIDIAN_CLIP`. It also comes with render tests, including one that renders the tippecanoe-generated tiles directly.

Closing this PR and https://github.com/maplibre/geojson-vt/pull/112 in favor of https://github.com/maplibre/maplibre-gl-js/pull/7896. I hope this is the right direction and approach now (it's also generally much smaller) 🤞

Issue #7363, opened 2026-03-31
Terrain performance needs improvement, especially during rotation/pitch and close-range interaction


NeluQi:
## Description

I am seeing significant and unstable performance degradation when using terrain in MapLibre GL JS.

The problem is most visible when the camera is close to the terrain surface and the view is rotated or strongly pitched toward the horizon. In that situation FPS drops sharply and frame times become unstable.

If the camera moves only slightly farther away, performance improves noticeably. So the issue is not only low terrain performance in general, but also that performance changes too dramatically depending on camera distance and pitch angle.

## Why this matters

Performance is very important for my use case, and overall terrain performance is still far from ideal in practical 3D map scenarios.

High pitch / strongly tilted views are not an edge case for me. They are a normal interaction pattern, because users often need oblique views and sometimes need to inspect 3D objects from below or from a very low-angle perspective. Avoiding pitch is therefore not a practical workaround.

Because of that, both general terrain performance improvements and specific optimizations for high-pitch interaction would be very valuable.

## Terrain setup

The terrain is not a tiny demo dataset. It is a relatively extended / long terrain dataset covering a large area. It is not the most extreme possible case, but it is a realistic terrain-heavy scene where users navigate over a long surface.

## Observed behavior

When the camera is close to terrain and rotated with strong pitch:

- FPS drops heavily
- frame times become unstable
- interaction feels stuttery

When the camera moves slightly farther away:

- FPS improves significantly
- frame times become much more stable

More generally, terrain performance becomes much worse when the camera is strongly tilted toward the horizon.

## What I observed in profiling

In the Chrome performance trace, recurring hotspots repeatedly go through terrain-related internal paths such as:

- `getRenderableTiles`
- `getTileBoundingVolume`
- `getTerrainData`
- `getDEMElevation`
- `getElevationForLngLatZoom`
- `unproject`
- `pointCoordinate`
- `readPixels`

This suggests that close-range terrain interaction and strongly pitched views may be causing more expensive terrain work than expected.

The trace also shows noticeable memory pressure / GC activity during interaction, which may further worsen frame-time spikes.

## Possible optimization directions

It would be very useful if MapLibre could provide additional optimization mechanisms for terrain-heavy scenes, for example:

- more aggressive culling for distant terrain when pitch is high
- configurable terrain view distance / horizon distance
- pitch-aware terrain quality reduction
- lower-cost handling of terrain elevation queries / readback paths
- better stability of frame times during close-range terrain interaction

[...]

Issue #4487, opened 2024-08-01
3D Terrain significantly decrease performance of the map


AEiosApp:
**maplibre-gl-js version**:
4.5.0
**browser**:
Chrome, Android WebView with Chrome version 60

### Steps to Trigger Behavior

 1. Define terrain source, type of 'raster-dem' in the style (do not add 'terrain' configuration to the style)
 2. Create new map in your app, on map 'load' event - add terrain control with a terrain source defined in the style
 3. Check the map working with terrain disabled - should work fast & smooth
 4. Click on the 'terrain' control button to enable terrain - check the map working with terrain - should work slow
 5. Click again on the 'terrain' control button to disable terrain - check the map again - should work fast & smooth

### Link to Demonstration

Hey guys, thanks a lot for the 3D terrain support, it is really great!

However, I'm facing a performance issue when I use it. Please, see the video below:
- it's an Android hybrid app, so npm maplibre-gl is used here with an Android WebView browser
- video made with real Android device (not a simulator)
- when terrain is disabled (only vector map with hillshades) - everything works smooth and fast
- when I turn on terrain (with a terrain control button at timestamp 0:07):
- - I make same gestures to move / zoom map, but responce time is almost 1 second
- - cpu & memory usage is growing (see performance monitor on the video)
- - my terrain tiles are pretty small, about 1 MB of data in total
- when I turn off terrian (at timestamp 0:30), immediately:
- - I make same gestures to move / zoom map and everything works smooth and fast again
- - cpu & memory usage decreases

Is there a chanse to figure out with the performance?
**Because of this issue, we can't use this great feature in production.**

Let me know if you'd like me to provide more information or make more tests.

https://github.com/user-attachments/assets/67df8c6a-6aa0-445d-bee7-b3c2c1f391ad

### Style sample

```
{
  "version": 8,
  "name": "App Raster",
  "sources": {
    "raster-tiles": {
      "type": "raster",
      "tiles": ["https://maps.wikimedia.org/osm-intl/{z}/{x}/{y}.png"],
      "tileSize": 256
    },
    "terrainSource": {
      "type": "raster-dem",
      "url": "./hillshades/budapest/terrainTiles.json",
      "tileSize": 256
    },
    "hillshadeSource": {
      "type": "raster-dem",
      "url": "./hillshades/budapest/terrainTiles.json",
      "tileSize": 256
    }
  },
  "layers": [
    {
      "id": "simple-tiles",
      "type": "raster",
[...]

PR #7529, opened 2026-04-27
Improve terrain benchmarks

lucaswoj:
Related to #7363
Related to #4487
Continuation of @CommanderStorm's work in #7400.

This PR replaces the `TerrainRender` benchmark with four new ones that aim to answer the question: does turning on 3D terrain make the map slower, and by how much?

```
$ npm run benchmark -- Terrain3DGlobe Terrain3DMercator Terrain2DGlobe Terrain2DMercator
                   terrain-dropped-frames-bench
    Terrain3DGlobe                  414.1778 ms
 Terrain3DMercator                  425.1833 ms
    Terrain2DGlobe                    1.4444 ms
 Terrain2DMercator                    1.4833 ms
```

Each benchmark reports a "jank-sum" score (in ms) over a `flyTo` animation across a complex basemap, at pitch 85, with terrain either on or off, and projection either `globe` or `mercator`. Jank-sum is the total time spent past the 16.67ms frame budget across all frames in the flight: each frame contributes `max(0, frameTime - 16.67)`, summed over the animation. Conceptually, it is how many milliseconds the user spent looking at a frozen map canvas.

I picked jank-sum over the alternatives because it is much more stable run-to-run. Max frame time is dominated by a single outlier (one GC pause swings the whole result). p99 frame time is better but still picks one frame out of ~60, which leaves it noisy at this measurement length. Jank-sum aggregates the entire tail of the distribution, so it captures both how often frames are dropped and how badly, and a single bad frame doesn't dominate. It also has a nice property that 0 means every frame hit 60fps.

The new bench improves on the existing `TerrainRender` benchmark (added in #7400) by exercising tile loading, processing, and GPU upload, which bottleneck 3d terrain performance in my testing.

Once this lands I'm planning to use it to drive a deep-dive on improving terrain perf.

cc @CommanderStorm @HarelM

## Launch Checklist

 - [x] Link to related issues.
 - [x] Post benchmark scores.

 ## AI Disclosure

Assisted-By: Claude Opus 4.7


test/bench/benchmarks/terrain.ts
```diff
@@ -0,0 +1,275 @@
...
+        }
+    }
+
+    return {
+        version: 8,
+        glyphs: '/test/integration/assets/glyphs/{fontstack}/{range}.pbf',
+        sources: {
+            vector: {
+                type: 'vector',
+                tiles: [`${location.origin}/test/bench/data/785.vector.pbf?id={z}/{x}/{y}`],
```
CommanderStorm, line 267: 🤔
I think if we re-use the data, this will mean that caches are also reused, or?
Would another bench where this is not the case make sense?

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## main
 ### ✨ Features and improvements
+- Added `Terrain3DGlobe`, `Terrain3DMercator`, `Terrain2DGlobe` and `Terrain2DMercator` benchmarks. (by [@lucaswoj](https://github.com/lucaswoj))
```
CommanderStorm: I'd somewhat consider this internal -> no changelog entry, but if you think this is usefull we can keep it ^^

CommanderStorm (approved):
A much better benchmark!
I left one comment, you know better if that means that these benchmarks are one-sided on this axis.

test/bench/benchmarks/terrain.ts
```diff
@@ -0,0 +1,275 @@
...
+    constructor() { super('Terrain3DMercator', true, 'mercator'); }
+}
+export class Terrain2DGlobe extends TerrainBase {
+    constructor() { super('Terrain2DGlobe', false, 'globe'); }
+}
+export class Terrain2DMercator extends TerrainBase {
+    constructor() { super('Terrain2DMercator', false, 'mercator'); }
+}
+
+// Create a basemap style that uses the features from the bundled
```
HarelM: Can you convert this to tsdocs style please?

test/bench/benchmarks/terrain.ts
```diff
@@ -0,0 +1,275 @@
...
+            center: CENTER_END,
+            zoom: ZOOM,
+            pitch: PITCH,
+            bearing: BEARING,
+            duration: DURATION,
+            curve: 1,
+            minZoom: ZOOM,
+            easing: t => t,
+        });
+        await new Promise(resolve => this.map.once('moveend', resolve));
```
HarelM: you can simply call `await this.map.once('moveend')`

test/bench/benchmarks/terrain.ts
```diff
@@ -0,0 +1,275 @@
...
+            this.map.style.tileManagers[id].clearTiles();
+        }
+        if (this.terrain) {
+            this.map.setTerrain(null);
+            this.map.setTerrain({source: 'dem', exaggeration: 1});
+        }
+
+        // Reset the map viewport
+        this.map.jumpTo({center: CENTER_START, zoom: ZOOM, pitch: PITCH, bearing: BEARING});
+        await new Promise(resolve => this.map.once('idle', resolve));
```
HarelM: You can use `await this.map.once('idle')`

test/bench/benchmarks/terrain.ts
```diff
@@ -0,0 +1,275 @@
...
+            style,
+            fadeDuration: 0,
+            stubRender: false,
+            showMap: true,
+            idle: true,
+        });
+
+        if (this.terrain) {
+            this.map.setTerrain({source: 'dem', exaggeration: 1});
+            await new Promise(resolve => this.map.once('idle', resolve));
```
HarelM: Same as others...

test/bench/benchmarks/terrain.ts
```diff
@@ -0,0 +1,275 @@
...
+            easing: t => t,
+        });
+        await new Promise(resolve => this.map.once('moveend', resolve));
+        running = false;
+
+        // Return the slowest frame time.
+        times.shift();
+        let maxTime = 0;
+        for (const time of times) if (time > maxTime) maxTime = time;
+        return maxTime;
```
HarelM: Is this return value used anywhere?

test/bench/benchmarks/terrain.ts
HarelM: Oh, sorry, missed the second time its being used...

test/bench/benchmarks/terrain.ts
```diff
@@ -0,0 +1,275 @@
...
+            await new Promise(resolve => this.map.once('idle', resolve));
+        }
+    }
+
+    async run(): Promise<Measurement[]> {
+        try {
+            await this.setup();
+
+            // Warmup
+            await this.runInner();
```
HarelM: If this is part of the warmup, shouldn't it be ran inside the setup method?

test/bench/benchmarks/terrain.ts
```diff
@@ -0,0 +1,275 @@
...
+
+        if (this.terrain) {
+            this.map.setTerrain({source: 'dem', exaggeration: 1});
+            await new Promise(resolve => this.map.once('idle', resolve));
+        }
+    }
+
+    async run(): Promise<Measurement[]> {
+        try {
+            await this.setup();
```
HarelM, line 61: Doesn't the benchmark framework calls setup? Why is it being called inside `run` method?

test/bench/benchmarks/terrain.ts
```diff
@@ -0,0 +1,275 @@
...
+
+            // Warmup
+            await this.runInner();
+
+            const measurements: Measurement[] = [];
+            for (let i = 0; i < ITERATIONS; i++) {
+                measurements.push({time: await this.runInner(), iterations: 1});
+            }
+
+            this.teardown();
```
HarelM, line 68: Same as previous comment, isn't this ran as part of the framework?

test/bench/benchmarks/terrain.ts
```diff
@@ -0,0 +1,275 @@
+import Benchmark, {type Measurement} from '../lib/benchmark';
+import createMap from '../lib/create_map';
+import type {Map} from '../../../src/ui/map';
+import type {StyleSpecification} from '@maplibre/maplibre-gl-style-spec';
+
+const STYLE_COMPLEXITY = 64;
+const ITERATIONS = 5;
```
HarelM: Should this be replaced by `minimumMeasurements`?

test/bench/benchmarks/terrain.ts
lucaswoj: We are overriding the default `Bench#run` method, which is where the calls to `setup` and `teardown` live. 

The default `Bench#run` method can only report the overall execution time of the `Bench#bench` method. We want to report the max frame time of the operation, not the overall operation execution time.

test/bench/benchmarks/terrain.ts
lucaswoj: See https://github.com/maplibre/maplibre-gl-js/pull/7529/changes#r3155261223

test/bench/benchmarks/terrain.ts
lucaswoj: See https://github.com/maplibre/maplibre-gl-js/pull/7529/changes#r3155261223

test/bench/benchmarks/terrain.ts
lucaswoj: Yes. This return value is fed into the visualization of the bench results at `test/bench/results/Terrain2DMercator.pdf`

Image

test/bench/benchmarks/terrain.ts
lucaswoj: My goal is to have this benchmark work off a **warm** HTTP cache (because that's completely outside of the terrain system we are targeting with his benchmark) and **cold** tile and terrain caches (because that's a significant bottleneck for the terrain system).

We clear the tile and terrain caches before every run at https://github.com/maplibre/maplibre-gl-js/pull/7529/changes#diff-95ba005a0a24285a87c2740ae4774325f103120791215cc9281cfec275d21e34R78

Only benchmarking the case where the tile and terrain caches are cold is a reasonable design for the time being. If we can make the cold-cache-case fast, it follows that the warm-cache-case will also be fast. Each additional benchmark slows down the full benchmark suite runtime and adds maintenance burden.

test/bench/benchmarks/terrain.ts
lucaswoj: Done in the latest push

test/bench/benchmarks/terrain.ts
lucaswoj: Done

test/bench/benchmarks/terrain.ts
lucaswoj: Done

test/bench/benchmarks/terrain.ts
lucaswoj: Done

test/bench/benchmarks/terrain.ts
lucaswoj: Done

CHANGELOG.md
lucaswoj: Done

lucaswoj:
I have addressed all code review comments and confirmed that the benchmarks still produce meaningful results after the latest round of changes. Ready for another round of review.

lucaswoj:
Results on an iPhone 16 Pro:

image

test/bench/benchmarks/terrain.ts
```diff
@@ -0,0 +1,278 @@
...
+        } catch (e) {
+            console.error(e);
+        }
+    }
+
+    teardown() {
+        this.map.remove();
+    }
+
+    private async runInner(): Promise<number> {
```
HarelM: Can this be renamed to `bench` in order to match the other benchmarks?

test/bench/benchmarks/terrain.ts
HarelM: Is there a different way to solve this maybe? This feels like a rewrite of the baseclass, which I find odd.
Maybe we can define an interface for the benchmark class and implement it here instead of using a base class and not using its methods?

test/bench/benchmarks/terrain.ts
lucaswoj: Done in https://github.com/maplibre/maplibre-gl-js/pull/7529/commits/4cc10e6050dc94e9c76fc231ff57263cbfdbed50

test/bench/benchmarks/terrain.ts
lucaswoj: Done in https://github.com/maplibre/maplibre-gl-js/pull/7529/commits/4cc10e6050dc94e9c76fc231ff57263cbfdbed50

lucaswoj:
I have defined a new `BenchmarkLike` interface, done some parameter tuning, and switched from a max-frame-time to a jank-sum metric, to improve the stability of the outputs. This is ready for another round of review.

HarelM (approved):

PR #7537, opened 2026-04-28
feat!: change default value of zoom levels to overscale

HarelM:
## Launch Checklist

- This feature was introduced here: #6521

This changes the name of `experimentalZoomLevelsToOverscale` to `zoomLevelsToOverscale` and sets the default to 4.
This is a breaking change as it changes rendering and query rendered features results.
It improves performance and accuracy of high zoom labels.

 - [x] Link to related issues.

No AI.


src/ui/map.ts
CommanderStorm: wait a minute.. what does this do?
Is this a typo?

HarelM: The functionality didn't change, just the removal of "experimental" and the default set to 4.
It splits vector tiles instead of overscale and clip them.
Which part is a typo?
You can check the original PR that added this functionality for more information I guess...

CommanderStorm: so if 20 is max and this is 3 and we start splitting at 20 and all values below this get overscaled..
What does this do actually?

The description is failry uncear.

HarelM: If source max zoom is 10, map max zoom is 20 and this parameter is 3 then until zoom 10 you get tiles from server, from zoom 10 to zoom 17 the vector tiles are split and then rendered using the vector information in the tiles, from zoom 17 to zoom 20 tiles from zoom 17 are used and rendered larger than the original size by overscaling the tile and then clipped, which has precision issues when overscaling is 10 zoom levels for example. 
If you have a good idea how to improve the description let me know.

src/ui/map.ts
```diff
@@ -403,10 +403,9 @@ export type MapOptions = {
      * When undefined, all zoom levels after source's max zoom will be overscaled.
      * This can help in reducing the size of the overscaling and improve performance in high zoom levels.
      * The drawback is that it changes rendering for polygon centered labels and changes the results of query rendered features.
```
```suggestion
    * Defines the number of zoom level that will overscale from map's max zoom and below.
    *
    * This can help in reducing the size of the overscaling and improve performance in high zoom levels.
    * The drawback is that it changes rendering for polygon centered labels and changes the results of query rendered features.
    * When `undefined`, all zoom levels after source's max zoom will be overscaled.
    * 
    * For example if map's `maxZoom` is 20, the source's `maxzoom` is 10 (tiles are avaliable until zoom 10) and `zoomLevelsToOverscale` is set to 3:
    * - The zoom levels of 20, 19, 18 will be overscaled.
    * - The zoom levels 11 to 17 will be split.
```

HarelM: I changed it a bit, let me know if it's still clear.

src/ui/map.ts
```diff
@@ -402,11 +402,18 @@ export type MapOptions = {
      * and the rest will be split.
      * When undefined, all zoom levels after source's max zoom will be overscaled.
      * This can help in reducing the size of the overscaling and improve performance in high zoom levels.
```
```suggestion
```

src/ui/map.ts
```diff
@@ -402,11 +402,18 @@ export type MapOptions = {
...
-     * The drawback is that it changes rendering for polygon centered labels and changes the results of query rendered features.
-     * @defaultValue undefined
-     * @experimental
+    * Defines the number of zoom level that will overscale from map's max zoom and below.
```
CommanderStorm: how about this

```suggestion
    * Defines the number of zoom level that will overscale instead of split tiles below (inclusive) a map's `maxZoom`.
```

src/ui/map.ts
```diff
@@ -401,12 +401,18 @@ export type MapOptions = {
      * For example if the map's max zoom is 20 and this is set to 3, the zoom levels of 20, 19 and 18 will be overscaled
      * and the rest will be split.
      * When undefined, all zoom levels after source's max zoom will be overscaled.
```
CommanderStorm: github does not seem to want to delete this via code comments, but delete this duplciate part

CommanderStorm (approved):
assuming the code comments are resolved

HarelM:
Can you check the description now? The githug editing did not went well for this part, I changed it manually, but I'm not sure this is exactly what you had in mind...

CommanderStorm (approved):

derEremit:
Hit a regression with this default while testing v6 — filing here since the change just merged. Happy to open a separate issue or a PR if you'd prefer.

**Symptom:** when `Map.maxZoom < source.maxzoom + zoomLevelsToOverscale`, every source-layer that only appears at the upper end of the source's zoom range silently disappears. No console error, tiles just stop being requested below source.maxzoom.

In our case: `maxZoom: 17` on the Map + OpenFreeMap planet's `maxzoom: 14` ⇒ `building`, `housenumber`, fine `poi` all gone. Visually obvious — central Prague at z14.5 pitch 60° went from ~270 rendered building features down to 0. Setting `map._zoomLevelsToOverscale = undefined` at runtime restored 1498 rendered buildings (and reload is automatic).

**Bisect:**

| version | published | default | buildings render? |
|---|---|---|---|
| 6.0.0-3 | 2026-04-29 06:36Z | `experimentalZoomLevelsToOverscale: undefined` | ✅ |
| **6.0.0-4** | **2026-04-29 07:59Z** | **`zoomLevelsToOverscale: 4`** | ❌ |
| 6.0.0-5 | 2026-04-29 11:36Z | `zoomLevelsToOverscale: 4` | ❌ |

PR #7537 merged 07:22Z, v6.0.0-4 published 37 min later — first prerelease with the new default.

**Root cause:** the `maxzoom` argument passed into `coveringTiles` at `src/tile/tile_manager.ts:545-547` (introduced by #6521) **replaces** the source's maxzoom rather than clamping it:

```ts
maxzoom: this._source.type === 'vector' && this.map._zoomLevelsToOverscale !== undefined
    ? transform.maxZoom - this.map._zoomLevelsToOverscale
    : this._source.maxzoom,
```

When `transform.maxZoom - zoomLevelsToOverscale < source.maxzoom` (e.g. `17 − 4 = 13 < 14`), `coveringTiles` caps tile fetches at z13 and the source never returns its z14 data. The slicing path in `vector_tile_source.ts:233-235` is a no-op anyway when `canonical.z <= source.maxzoom`, so under-fetching here gains nothing.

**Repro shape** (any vector source with `maxzoom: 14` plus a `building` source-layer — OpenMapTiles, OpenFreeMap, MapTiler, Planetiler outputs):

```js
  container: 'map',
  style: '<style with vector source maxzoom: 14>',
  center: [-74.006, 40.7128],
  zoom: 14,
  maxZoom: 17,
});
map.on('load', () => {
  const n = () => map.querySourceFeatures(SOURCE, { sourceLayer: 'building' }).length;
  console.log('buildings in source:', n());
  // v6.0.0-3:           hundreds
  // v6.0.0-4 / -5:      ~0–6
  map._zoomLevelsToOverscale = undefined;       // workaround
  map.style.tileManagers[SOURCE].reload();
  setTimeout(() => console.log('after override:', n()), 5000);
});
```

**Suggested fix** — clamp instead of replace:

```ts
maxzoom: this._source.type === 'vector' && this.map._zoomLevelsToOverscale !== undefined
    ? Math.max(this._source.maxzoom, transform.maxZoom - this.map._zoomLevelsToOverscale)
    : this._source.maxzoom,
```

Rationale: the slicing optimization only does work when `canonical.z > source.maxzoom`. When `source.maxzoom > transform.maxZoom − zoomLevelsToOverscale`, slicing is intended to be a no-op — but the current code instead under-fetches the source. `Math.max` preserves the slicing optimization when it's actually applicable (sources with low maxzoom + high `Map.maxZoom`) while letting deeper sources use their full data.

Cases:

| `source.maxzoom` | `Map.maxZoom` | `Map.maxZoom − 4` | proposed cap | behavior |
|---|---|---|---|---|
| 14 | 17 | 13 | **14** | fetch real z14 (was: capped at z13, data missing) |
| 14 | 22 | 18 | 18 | fetch up to z14 + slice z15–z18 (unchanged) |
| 10 | 22 | 18 | 18 | fetch up to z10 + slice z11–z18 (unchanged) |

**Affected configurations:** any app with `Map.maxZoom < source.maxzoom + zoomLevelsToOverscale`. The 14-maxzoom OpenMapTiles/OpenFreeMap convention plus a typical UX-cap of `maxZoom: 17` or `18` puts most disc-golf / hiking / city-explorer apps in this bucket.

Workaround until fixed: pass `zoomLevelsToOverscale: undefined` to the `Map` constructor.

HarelM:
Thanks for testing!! Sounds like a clamping issue. Feel free to submit a PR to fix this.

PR #7545, opened 2026-04-29
Migrate render tests to vitest

HarelM:
## Launch Checklist

This is a refactoring of the render test to use vitest.
Looks like it takes about the same time and it has all the capabilities of vitests, which is important.


HarelM:
Looks like the other PR (#7546) is taking the same time, so I don't think introducing vitest here makes things slower.
For running locally, the reporting is a bit better, but running locally all the test is too long now, so it's mostly for CI, which is good, and you get all the vitest capabilities for running a subset etc, which is better.

CommanderStorm (approved):

PR #7549, opened 2026-04-29
Improve 3d terrain performance by redesigning RTT cache to increase hitrate

### LINKED ISSUE #4487: 3D Terrain significantly decrease performance of the map

### LINKED ISSUE #7363: Terrain performance needs improvement, especially during rotation/pitch and close-range interaction

lucaswoj:
When terrain is on, MapLibre renders each tile into an offscreen WebGL texture (internally called a "render to texture" object, abbreviated as RTT), then drapes those textures onto the 3D terrain mesh. In the real world, a mobile device might render 20 tiles, each with 10 stacks. That's 200 of texture renders per frame. We improve 3D terrain performance by caching these RTT objects across frames. 

Today the RTT cache lives in `RenderPool` ([`src/webgl/render_pool.ts`](https://github.com/maplibre/maplibre-gl-js/blob/main/src/webgl/render_pool.ts) on `main`) and supports a fixed-size pool of 30 FBO+texture slots. Each tile remembers an `(id, stamp)` reference to "its" slot, and the cache "hits" when the slot's stamp still matches.

The hit rate of this cache is ~0% on real-world workloads. I measured 0.2% on MapTiler's topo-v2 style at zoom 12, pitch 80°, on a 393×852 viewport. The cache is doing essentially nothing.

This PR fixes that, getting a 94.6% hit rate at ~2× the GPU memory, with the same cleanup behavior we have today.

Fix #7363
Fix #4487

### Why the cache hit rate is so low

1. `freeAllObjects()` runs at the end of every layer stack, marking all slots `inUse=false`. The next miss in the next stack calls `getOrCreateFreeObject`, which walks `_recentlyUsed` LRU-first and returns one of those just-freed slots. `stampObject` bumps the slot's stamp, which invalidates the previous tile's cache reference instantly.
2. Pool size 30 = working set 30 already touches every slot on every frame, so step 1 fires every frame for every tile.

### Why simpler fixes don't work

I tested two before settling on this design:

- **Bigger pool alone**: pool size 400 still gets 0.1% hit rate. `getOrCreateFreeObject` aggressively reuses freed slots before creating new ones, so the pool can be 99% empty and the cache still misses every time.
- **Bigger pool + remove `freeAllObjects`**: 95% hit rate, but GPU memory grows to 3.36 GB and the page crashes under sustained panning with `No free RenderPool available` once new tiles enter the view. 
- Adding retain/release to `RenderPool` would work, but adds a lot more complexity. Tiles already manage the lifetime of their other GPU resources (textures, buffers), so it's simpler to let tiles hold the slot references directly and release them on unload.

### What this PR does

Each `Tile` owns its RTT objects directly via `tile.rttObjects: Map<stack, RTTObject>`. Slots return to a painter-level recycle pool when the tile unloads (`tile.releaseRTT`). Cache lookup is `tile.rttObjects.has(stackIndex)`. No stamps, no LRU heuristic.

RTT lifetime is tied to tile lifetime, so eviction happens naturally when `TerrainTileManager.update()` drops tiles that left the view.

| | main | main + pool 400 + no `freeAllObjects` | this PR |
|---|---|---|---|
| Hit rate | 0.2% | 95.2% | 94.6% |
| GPU memory | 151 MB | 3.36 GB | **302 MB** |
| Sustained panning | works | crashes | works |

Same hit rate as the hacked-up version, ~1/11 the memory, no crash.

### Benchmarks

```
$ npm run build-benchmarks && time npm run benchmark -- --compare='' Terrain3DGlobe Terrain3DMercator Terrain2DGlobe Terrain2DMercator

                   rtt-cache-dirty cd14f21
    Terrain3DGlobe             37.4944 ms
 Terrain3DMercator             37.4833 ms
    Terrain2DGlobe              1.4833 ms⚠️
 Terrain2DMercator              1.4444 ms
```

This branch consistently returns around 37ms for the `Terrain3DGlobe` benchmark vs ~410ms for the same benchmark on `main`. About 91% jank-sum reduction.

## Launch Checklist

 - [x] Link to related issues.
 - [x] Post benchmark scores.

 ## AI Disclosure

Assisted By: Claude Opus 4.7


src/render/painter.ts
```diff
@@ -80,6 +81,9 @@ export class Painter {
     _tileTextures: {
         [_: number]: Texture[];
     };
+    _rttObjects: {
+        [_: number]: RTTObject[];
+    };
```
CommanderStorm: nit: Have you tested using `Map` for this cache? Given how it is used, this "feels" slightly better, but might not actually be more performant.

src/render/painter.ts
```diff
@@ -704,6 +710,30 @@ export class Painter {
         return textures && textures.length > 0 ? textures.pop() : null;
     }
 
+    acquireRTT(size: number): RTTObject {
+        const obj = this._rttObjects[size];
+        if (obj && obj.length > 0) return obj.pop();
```
CommanderStorm: nit:

```suggestion
        if (obj?.length > 0) return obj.pop();
```

src/render/painter.ts
```diff
@@ -687,6 +692,7 @@ export class Painter {
     }
 
     static readonly MAX_TEXTURE_POOL_SIZE_PER_BUCKET = 50;
+    static readonly MAX_RTT_SLOT_POOL_SIZE_PER_BUCKET = 400;
```
CommanderStorm: How did you come up with this number? 🤔

(It looks fine, I just would like to know how you sized this 😉)
Is there maybe a nice way to make this semi-dynamic based on the avaliable GPU memory?

CommanderStorm (dismissed):
Nice work!

src/render/painter.ts
lucaswoj: Using `Map` wouldn't be any more performant than a plain object. My intent is to follow the conventions established by the `_tileTextures` system above.

src/render/painter.ts
lucaswoj: Fixed

src/render/painter.ts
lucaswoj: I have removed this constant in the latest commit. the RTT pool size is naturally bounded by: 
```
(max # of tiles visible) * (# of stacks in style)
```

Devices with smaller screens and simpler styles will naturally use less memory. More complex styles will require more memory, but that's probably better than just re-rendering every RTT on every frame.

lucaswoj:
This is ready for review.

The failing [Tests / Code Hygiene (pull_request)](https://github.com/maplibre/maplibre-gl-js/actions/runs/25186515179/job/73845123765?pr=7549) seems to be an upstream issue.

src/tile/terrain_tile_manager.ts
```diff
@@ -113,19 +119,24 @@ export class TerrainTileManager extends Evented {
         }
         // free unused tiles
```
lucaswoj, line 115: If we free these before allocating new RTTs, we can reduce peak usage

lucaswoj: nevermind, the new RTTs are acquired during the render step, so this is a moot point

lucaswoj:
> This is ready for review.
> 
> The failing [Tests / Code Hygiene (pull_request)](https://github.com/maplibre/maplibre-gl-js/actions/runs/25186515179/job/73845123765?pr=7549) seems to be an upstream issue.

Put together a fix for the upstream issue at https://github.com/maplibre/maplibre-gl-js/pull/7557

src/render/painter.ts
```diff
@@ -80,6 +81,9 @@ export class Painter {
     _tileTextures: {
         [_: number]: Texture[];
     };
+    _rttObjects: {
```
HarelM: Can we use `Record`? I'm trying to migrate the code to use it instead of these object types...

src/render/terrain.ts
```diff
@@ -144,6 +144,7 @@ export class Terrain {
     constructor(painter: Painter, tileManager: TileManager, options: TerrainSpecification, terrainSkirtLength: 'none' | 'auto' = 'auto') {
         this.painter = painter;
         this.tileManager = new TerrainTileManager(tileManager);
+        this.tileManager.painter = painter;
```
HarelM: Why not pass it in the constructor?

src/render/painter.ts
```diff
@@ -704,6 +709,24 @@ export class Painter {
         return textures && textures.length > 0 ? textures.pop() : null;
     }
 
+    acquireRTT(size: number): RTTObject {
```
HarelM, line 725: Passing a size parameter as an indexer is confusing at first read, maybe add a tsdoc comment to explain this a bit? Either on this method, or over the `_rttObjects` field at the top?

src/render/painter.ts
```diff
@@ -809,6 +832,16 @@ export class Painter {
             this._tileTextures = {};
         }
 
+        if (this._rttObjects) {
```
HarelM: Can we initialize `_rttObjects` to `{}` and avoid this if?

src/webgl/render_to_texture.ts
```diff
@@ -67,15 +70,17 @@ export class RenderToTexture {
     constructor(painter: Painter, terrain: Terrain) {
         this.painter = painter;
         this.terrain = terrain;
-        this.pool = new RenderPool(painter.context, 30, terrain.tileManager.tileSize * terrain.qualityFactor);
+        this._rttSize = terrain.tileManager.tileSize * terrain.qualityFactor;
     }
 
     destruct() {
-        this.pool.destruct();
+        for (const tile of Object.values(this.terrain.tileManager._tiles)) {
```
HarelM: I would prefer if we don't use `_tiles` as this is an internal field.
Is there a different way to do it?

src/tile/tile.ts
```diff
@@ -197,6 +208,21 @@ export class Tile {
         this.demTexture = null;
     }
 
+    /**
+     * Return all cached RTT slots to the painter's pool. If no painter is
+     * available (e.g. tile created in a test without a render context), the
+     * slots are simply dropped — they were never rendered into.
+     */
+    releaseRTT(painter: Painter | undefined) {
```
HarelM: I couldn't find places in the code where this is called without a painter.
Can you link to those maybe?
Generally speaking, I find it a bit hard to understand when reading this PR where things are stored and who's responsible for releasing, and which part is released where.

If there's a way to simplify it while keeping the performance gain it would be great.
Even having something like the render pool but written/managed differently it might make reading this a bit easier.

HarelM: Another approach would maybe remove the painter input from this method and return a list of objects that needs to be released to the caller, so the caller can call painter's release method with the relevant input.
I think that the fact the a tile knows about painter is probably a circular dependency of some sort, so it would be good to avoid making it more coupled, if possible.

src/webgl/render_to_texture.test.ts
```diff
@@ -214,14 +219,64 @@ describe('render to texture', () => {
...
+        rtt.prepareForRender(style, 0);
+
+        const acquireSpy = vi.spyOn(painter, 'acquireRTT');
+        acquireSpy.mockClear();
+
+        const renderOptions = {isRenderingToTexture: false, isRenderingGlobe: false};
+        rtt.renderLayer(fillLayer, renderOptions);
+        rtt.renderLayer(symbolLayer, renderOptions);
+
+        expect(acquireSpy).toHaveBeenCalledWith(rtt._rttSize);
```
HarelM: Would be great if we could avoid using `_...` fields in the tests, not sure I know how to do that, but it will make the test more future proof if we don't use "internal" fields.

src/render/painter.ts
```diff
@@ -848,3 +881,15 @@ export class Painter {
...
     }
 }
+
+/**
+ * Holds the GPU resources used to render a 2D tile into a texture so it can
+ * be draped over 3D terrain: an FBO with a color texture and depth-stencil
+ * renderbuffer attached. Owned by a `Tile` while in use, recycled via
+ * `Painter.releaseRTT` when the tile no longer needs it.
+ */
+export type RTTObject = {
```
HarelM: Types are defined in the beginning of the file, can you move it upwards?

src/tile/terrain_tile_manager.test.ts
```diff
@@ -215,4 +216,69 @@ describe('TerrainTileManager', () => {
...
+
+            return {parent, same, child, sibling, tiles, rttObjects, painter};
+        }
+
+        test('with no tileID releases every cached tile', () => {
+            const {tiles, rttObjects, painter} = setupTilesWithRttObjects();
+
+            tsc.releaseRTT();
+
+            expect((painter.releaseRTT as any).mock.calls.length).toBe(Object.keys(tiles).length);
```
HarelM: Can you type it a bit better? `as any as Spy` or something?

src/tile/terrain_tile_manager.ts
```diff
@@ -60,6 +61,11 @@ export class TerrainTileManager extends Evented {
      */
     _lastTilesetChange: number = now();
 
+    /**
+     * the painter that owns the RTT pool, used to release RTT objects on tile unload.
+     */
+    painter: Painter | undefined;
```
HarelM: Can we avoid this somehow? Maybe use `this.tilemanager.map.painter`?
While both options are probably bad from a coupling and circular dependencies issue, I would like at some point in time to solve this, and I think adding painter here makes the problem worse, I think.
Maybe use a smaller interface to narrow down what's being using in this class and not require the entire painter object?

HarelM:
@lucaswoj thanks for looking into this! I'll be happy to try it out on a mobile device once this lands to see if I can enable 3D terrain in my app.

src/render/painter.ts
lucaswoj: Done in d9e8bc6ac0c0048becbbbf624a68cb840544a118

src/render/terrain.ts
lucaswoj: Done in 46b0ffadb

src/render/painter.ts
lucaswoj: Done in 388f29042

src/webgl/render_to_texture.ts
lucaswoj: Good point. Since the `TerrainTileManager` owns the tiles, we can make it directly responsible for the lifecycle of those objects, and eliminate this destruct method entirely. Done in 3d92c1abd

src/tile/tile.ts
lucaswoj: Made this parameter required in 897c1f1ca

lucaswoj: For the broader question, the mental model is:

1. `Painter` is a facade over the WebGL API. It's global within a map instance and acts as the single nexus for WebGL calls. That keeps GL usage easy to audit and lets us encapsulate WebGL-specific quirks in one place. The recycle pool of RTT objects is an internal implementation detail of `Painter`. No other part of the system needs to know it exists.

2. `Tile` objects own RTT objects. The lifecycle of an RTT object is tied to the lifecycle of its tile. When the tile is destroyed, the RTT is destroyed. While the tile lives, the RTT lives. This is the main thrust of the PR. Before, RTT lifecycles were tracked separately from tiles, and it was hard to come up with a caching system that allowed them to be reused across frames.

3. `TerrainTileManager` owns tiles. It can create and remove them in bulk. I agree it's weird that `TerrainTileManager` calls `Tile#freeRTT` directly instead of a more general `Tile#destroy`. That's how the codebase is currently structured, and I'd rather not expand the scope of this PR to refactor it.

lucaswoj: Also worth mentioning again: the system for managing RTT objects was meant to be as similar as possible to the existing system we use for managing "tile textures." There's a lot of value in consistency here. The architecture of the tile texture system has stood the test of time and is understood by developers who work on this part of the codebase.

src/webgl/render_to_texture.test.ts
lucaswoj: Removed the _ from this property in 7cea38139

src/render/painter.ts
lucaswoj: Done in 73a965eae

src/tile/terrain_tile_manager.test.ts
lucaswoj: Done in d9fcf808a

src/tile/terrain_tile_manager.ts
lucaswoj: Removed this property in 46b0ffadba6d91044aba70e8cbf6b30f7f06287d

src/render/painter.ts
lucaswoj: Added a comment in 9fd6fd44c

lucaswoj: Ended up removing this in 6c92e57728e37c809f3d963201e658a57739f86d. The RTT size doesn't change often and only one is in use at a time. When it does, we can resize the existing RTTs.

lucaswoj:
This is ready for another round of review.

HarelM:
I'll take a look tomorrow or on Sunday. 
Can you revert the changes in package lock file meanwhile?

lucaswoj:
> Can you revert the changes in package lock file meanwhile?

Done in https://github.com/maplibre/maplibre-gl-js/pull/7549/commits/ace2b3b451ea5637ce5e93a69b62a965bee12aa1

src/render/painter.ts
```diff
@@ -704,6 +722,38 @@ export class Painter {
...
+        if (this.context.extTextureFilterAnisotropic) {
+            gl.texParameterf(gl.TEXTURE_2D, this.context.extTextureFilterAnisotropic.TEXTURE_MAX_ANISOTROPY_EXT, this.context.extTextureFilterAnisotropicMax);
+        }
+        fbo.depthAttachment.set(this.context.createRenderbuffer(gl.DEPTH_STENCIL, size, size));
+        fbo.colorAttachment.set(texture.texture);
+        return {fbo, texture, size};
+    }
+
+    releaseRTT(obj: RTTObject) {
+        this._rttObjects.push(obj);
```
HarelM: I'm confused, why release pushes an object?

src/tile/tile.ts
```diff
@@ -197,6 +208,19 @@ export class Tile {
...
 
+    /**
+     * Return all cached RTT slots to the painter's pool. If no painter is
+     * available (e.g. tile created in a test without a render context), the
+     * slots are simply dropped — they were never rendered into.
+     */
+    releaseRTT(painter: Painter) {
+        if (this.rttObjects.length === 0) return;
+        for (const obj of this.rttObjects) {
+            if (obj) painter.releaseRTT(obj);
```
HarelM: Why would `obj` be null here?

src/tile/tile.ts
```diff
@@ -197,6 +208,19 @@ export class Tile {
...
+        }
+        this.rttObjects.length = 0;
```
HarelM, line 238: Shouldn't this be `this.rttObjects = []`?

src/render/painter.test.ts
```diff
@@ -82,3 +82,58 @@ describe('tile texture pool', () => {
         painter.destroy();
     });
 });
+
+describe('RTT pool', () => {
+    function createPainter() {
```
HarelM: Use beforeEach maybe?
And then painter.destroy(); in afterEach?

src/render/painter.ts
```diff
@@ -704,6 +722,38 @@ export class Painter {
...
+        }
+        const fbo = this.context.createFramebuffer(size, size, true, true);
+        const texture = new Texture(this.context, {width: size, height: size, data: null}, gl.RGBA);
+        texture.bind(gl.LINEAR, gl.CLAMP_TO_EDGE);
+        if (this.context.extTextureFilterAnisotropic) {
+            gl.texParameterf(gl.TEXTURE_2D, this.context.extTextureFilterAnisotropic.TEXTURE_MAX_ANISOTROPY_EXT, this.context.extTextureFilterAnisotropicMax);
+        }
+        fbo.depthAttachment.set(this.context.createRenderbuffer(gl.DEPTH_STENCIL, size, size));
+        fbo.colorAttachment.set(texture.texture);
+        return {fbo, texture, size};
```
HarelM, line 750: Shouldn't this be added to _rttObjects before returning?

HarelM:
Added a few last comments, looks a lot better now, thanks!!

src/render/painter.ts
lucaswoj: `Painter#_rttObjects` is a pool of RTT objects that aren't currently in use by any tile. `acquireRTT` pulls one out, `releaseRTT` puts one back so a future `acquireRTT` call can reuse it instead of allocating a fresh framebuffer + texture.

I tried to sketch the full RTT object lifecycle in [this earlier comment](https://github.com/maplibre/maplibre-gl-js/pull/7549#discussion_r3173424411). Happy to expand on any part of that mental model that's still murky.

src/render/painter.ts
lucaswoj: See https://github.com/maplibre/maplibre-gl-js/pull/7549#discussion_r3177903769

src/tile/tile.ts
lucaswoj: `array.length = 0` mutates the existing array in place rather than allocating a new one.

src/tile/tile.ts
lucaswoj: I can't think of a case where it would be. Changed in ce036287c

src/render/painter.test.ts
lucaswoj: done in d04d58018

src/tile/tile.ts
lucaswoj: This method may be called rather rapidly during style transitions, so avoiding an unnecessary array allocation here may be meaningful.

src/render/painter.ts
HarelM: So this is similar to "outOfView" cache we have in the source objects?
If so, then please rename this field to "_outOfViewRTTCache" or something that will better explain why a call to a release method is pushing to an array.

HarelM: or something with the name "pool" I guess...?

src/tile/tile.ts
HarelM: Ok, I didn't know that. Thanks for the info.

src/tile/tile.ts
```diff
@@ -197,6 +208,19 @@ export class Tile {
         this.demTexture = null;
     }
 
+    /**
+     * Return all cached RTT slots to the painter's pool. If no painter is
+     * available (e.g. tile created in a test without a render context), the
+     * slots are simply dropped — they were never rendered into.
+     */
+    releaseRTT(painter: Painter) {
```
HarelM, line 233: I would expect some symmetry here, I think:
i.e. if tile is the class responsible for calling painter.releasetRTT I would expect to see here a call to aquireRTT as well.
Or am I reading it wrong?

CHANGELOG.md
HarelM, line 1: Can you check if this was merged correctly?

src/webgl/render_to_texture.ts
```diff
@@ -161,27 +160,11 @@ export class RenderToTexture {
...
-                }
-                // get free PoolObject
-                const obj = this.pool.getOrCreateFreeObject();
-                this.pool.useObject(obj);
-                this.pool.stampObject(obj);
-                tile.rtt[stack] = {id: obj.id, stamp: obj.stamp};
-                // prepare PoolObject for rendering
+                // Cache hit: this tile already has a RTT object for this stack from a previous frame.
+                if (tile.rttObjects[stack]) continue;
+                const obj = painter.acquireRTT(this.rttSize);
```
HarelM: Probably these two lines should be encapsulated inside tile to make the symmetry I'm suggesting...?

CHANGELOG.md
lucaswoj: fixed in a77195fc5

src/tile/terrain_tile_manager.ts
```diff
@@ -113,19 +114,24 @@ export class TerrainTileManager extends Evented {
...
     /**
-     * Free render to texture cache
-     * @param tileID - optional, free only corresponding to tileID.
+     * Release the RTT objects for `tileID` (and its ancestors/descendants),
+     * or for all tiles if `tileID` is omitted. RTT objects return to the
+     * painter's pool for recycling. Called when source data, image sources, or
+     * the style itself change in ways the per-tile fingerprint doesn't capture.
      */
-    freeRtt(tileID?: OverscaledTileID) {
+    releaseRTT(tileID?: OverscaledTileID) {
```
HarelM: It might be better to split this one into two methods for readability?
`releaseAllRtt()` and `releaseRttByTileId`?
I've seen that they are used separately, and I'm not sure it's worth combining them together.
This is optional.

src/render/painter.ts
lucaswoj: The closer parallel is the tile texture cache on this same class: `_tileTextures` + `getTileTexture` / `saveTileTexture`. `_rttObjects` mirrors `_tileTextures` (also a recycle structure whose name doesn't advertise the mechanism), and `acquireRTT` / `releaseRTT` mirror the get/save pair.

Today the painter satisfies `acquireRTT` from a recycle pool. Tomorrow we might allocate fresh, or share storage with `_tileTextures`, or store them in an atlas. The names capture the caller's intent ("I'm done with this RTT"), not how the painter fulfills it.

src/tile/tile.ts
lucaswoj: We could add another method to `Tile` to acquire an RTT but I can't see how that additional layer of indirection would improve the safety or clarity of our codebase.

src/tile/terrain_tile_manager.ts
lucaswoj: As you wish 8458d1df4

src/webgl/render_to_texture.ts
lucaswoj: Noted. Let's keep this discussion within https://github.com/maplibre/maplibre-gl-js/pull/7549#discussion_r3178220990

src/tile/tile.ts
HarelM: It help defines the responsibility, if you aquire in one class and release in another it creates discrepancy that's later hard to fix and understand. 
This symmetry is helpful. 
Also from a resource management perspective and OOP having this object "internal" to the tile class is better that manipulating it from the outside.

src/render/painter.ts
HarelM: The problem is that you have an object with a similar name that acts as a cache while this one acts as a pool. When reading the code it is confusing. So having names with meaning, even if they are internal to the class will be helpful to future readers.

src/tile/tile.ts
lucaswoj: Done 93cdf8806

package-lock.json
HarelM, line 1: This got accidentally pushed again...

src/render/painter.ts
lucaswoj: I renamed the internal property in https://github.com/maplibre/maplibre-gl-js/pull/7549/commits/3e77516a9dd90f1e961d00d739eb1a282eea2a29

src/tile/tile.ts
```diff
@@ -117,7 +117,18 @@ export class Tile {
...
+     * Caches the result of rendering this 2D tile into a texture so it can be
+     * draped over 3D terrain. Indexed by stack index, where a stack is a
+     * contiguous run of RTT-eligible layers (fill, line, raster, etc.) split
+     * by non-RTT layers like symbols. So `_rttObjects[0]` holds the texture
+     * for layers below the first symbol break, `_rttObjects[1]` for layers
+     * between the first and second symbol breaks, and so on. Each entry
+     * survives across frames until the tile is unloaded or its source data
+     * changes.
+     */
+    rttObjects: Array<RTTObject | undefined>;
```
HarelM, line 131: Is this field used from the outside give all the added methods? Maybe it can be marked as private, or with `_`?

package-lock.json
lucaswoj: fixed in 2d7a01e2c

src/tile/tile.ts
HarelM: This is optinal.

HarelM (approved):

src/tile/tile.ts
lucaswoj: We need to directly mutate this field in our unit tests. We cannot avoid accessing `Tile#_rttObjects` in the unit tests without adding even more methods to the Tile class.

HarelM:
Thanks for fixing all the code review comments!!
I left a final one for you to decide if to act on it or not.
This can me merged now.

lucaswoj:
Just did a final round of manual testing and ran the relevant benchmarks. Ready to merge when you are.

HarelM:
THANKS!

PR #7550, opened 2026-04-29
Optimization for Feature State: Replace String-Indexed Object with Array

xavierjs:
Hi,

## Issue

Managing large feature states is often slow. I can provide specific tests that demonstrate lag when feature state is used.

This PR aims to improve performance. While it does not fully resolve the slow feature state issue, it does offer some improvements.

Feature state is internally managed by an object indexed by strings (the IDs).  
If there are many IDs, iterating over such an object with a `for` loop is slow. Additionally, measuring the number of IDs requires a call to `Object.keys(state).length`, which allocates a new array.

## Proposed solution

We store the feature state using an array.
Note: this PR does not fix the *feature state not removed* issue (I hope I will find a solution fo this one). It is just an optimization.

## Benchmarks

Here are the results for the `PaintStates` benchmark:

**Before:**
Google Chrome, Macbook Pro M4 Max: `40.6ms`
Google Chrome, Macbook Pro M4 Max with swiftshaders (hardware acceleration disabled): `232ms`

**After:**
Google Chrome, Macbook Pro M4 Max: `11.9ms` (71% time reduction)
Google Chrome, Macbook Pro M4 Max with swiftshaders (hardware acceleration disabled): `209ms` (10% time reduction. It is normal that the time reduction% is smaller than with hardware acceleration since this optimization does not affect the rendering)

## Tests

## Misc

No related issues.

## Launch Checklist
 - [X] Link to related issues.
 - [X] Post benchmark scores.


HarelM:
The previous code was "immuned" to duplicate feature ids, is this a possible issue with the new code? 

Also there's a object that is being used like a "HashMap", should hashmap be used there instead?

src/source/source_state.ts
```diff
@@ -3,9 +3,21 @@ import type {Tile} from '../tile/tile';
 import type {FeatureState} from '@maplibre/maplibre-gl-style-spec';
 import type {InViewTiles} from '../tile/tile_manager_in_view_tiles';
 
-export type FeatureStates = {[featureId: string]: FeatureState};
+export type FeatureStateEntry = {id: string; state: FeatureState};
+export type FeatureStates = FeatureStateEntry[];
 export type LayerFeatureStates = {[layer: string]: FeatureStates};
 
+type FeatureStatesMap = {[featureId: string]: FeatureState};
```
HarelM: Use `Record<...>`.

src/source/source_state.ts
```diff
@@ -3,9 +3,21 @@ import type {Tile} from '../tile/tile';
...
+type LayerFeatureStatesMap = {[layer: string]: FeatureStatesMap};
+
+function featureStatesMapToArray(map: FeatureStatesMap): FeatureStates {
```
HarelM, line 13: Doesn't `Object.entries` does this functionality?

src/webgl/draw/draw_custom.test.ts
```diff
@@ -8,7 +8,7 @@ import {drawCustom} from './draw_custom';
 import {CustomStyleLayer} from '../../style/style_layer/custom_style_layer';
 import {MercatorTransform} from '../../geo/projection/mercator_transform';
 import {MercatorProjection} from '../../geo/projection/mercator_projection';
-import {type CustomRenderMethodInput} from '../../../dist/maplibre-gl';
+import {type CustomRenderMethodInput} from '../../style/style_layer/custom_style_layer';
```
HarelM, line 11: THANKS!

HarelM:
I've added some minor comments, but otherwise looks good.

src/source/source_state.ts
xavierjs: /done in the latest commit

src/source/source_state.ts
xavierjs: Good point!
If we use `Object.entries` and we keep the same output type ( `{id: string; state: FeatureState}[]` ), then we need to call a `.map`after then `Object.entries`, something like:

```
Object.entries(map).map(([id, state]) => ({id, state}));
```

It is shorter but less optimized since we allocate an intermediary array.

I tried to replace the output type by `[string, FeatureState][]` so that we can directly output `Object.entries(map)` in this function. But the result in the `PaintStates` benchmark is not better, it is exactly the same, and the code is less readable where the id and the state are read ( `fs.id` -> `fs[0]`and `fs.state`-> `fs[1]` ). So I think that we should keep this function as is.

xavierjs:
> The previous code was "immuned" to duplicate feature ids, is this a possible issue with the new code?

Good catch! There was indeed a possible feature id duplication issue in `source_state.ts`, function `coalesceChanges`.
It is solved by the latest commit using a temporary object.
There is a slight performance degradation: the `PaintStates` benchmark takes `13.5ms`instead of `11.9`. Fortunately we are still well below `40.6ms`.

> Also there's a object that is being used like a "HashMap", should hashmap be used there instead?

For the hashmap, I don't think it will bring a lot of improvement since:
* The main issue comes from reading, not inserting values,
* Hashmap will have more memory overhead

lucaswoj:
I have a fix for the failing Code Hygiene job at https://github.com/maplibre/maplibre-gl-js/pull/7557

HarelM (approved):

Issue #6633, opened 2025-10-28
Large number of feature state keys leads to zoom lag when loading cached tiles


bradymadden97:
**maplibre-gl-js version**: 5.10.0

**browser**: Chrome

### Steps to Trigger Behavior

1. Have a large number (20k is noticeable) of GeoJSON features on a source
2. For each feature, have a significantly large feature property object (my repro has ~40 properties per feature)
  2a. (Note: this scale is representative of many real cases we see in our production app)
3. Set and unset a piece of feature state for each feature. Since `maplibre-gl` does not fully clear the key from the feature state after it's been set (rather just setting it to `{[key]: {}}`, you should be able to see the large quantity of feature state by running (using the below JSBin):
```javascript
> Object.keys(map.style.sourceCaches.airports._state.state._geojsonTileLayer).length;
< 20180
```
4. Zoom to a certain level, allow tiles to load. Zoom to a different level, allow tiles to load. Zoom back to the first level - notice _lagginess_ during zooming while returning to the first level, and _any_ subsequent zoom level that you had previously loaded tiles at.

### Link to Demonstration

Simplest repro is on JSBin: https://jsbin.com/lowojotomu/edit?html,console,output

Source instructions, if interested:
 1. Clone `https://github.com/bradymadden97/maplibre-tools`
 2. `pnpm i` and run repro `pnpm dev:repro slow-zoom-on-large-feature-state`
 3. Run a performance profile while the map is automatically zooming in and out

### Expected Behavior

Zooming does not hang when loading tiles from the tile cache. This will probably involve some refactor that doesn't re-process feature state-dependent expressions from every tile read from the tile cache every time. It seems like we should be able to skip some duplicative/unnecessary work here. Especially highlighted that repeated zooms in / out (with no changes to feature state in between) duplicate expensive work.

<img width="867" height="996" alt="Image" src="https://github.com/user-attachments/assets/9c3ce4ed-c114-47d2-9c04-391f72f59fd2" />

When you comment out the following lines from the JSBin:
```javascript
    for (const i of data.features) {
      map.setFeatureState(
        { source: "airports", id: i.properties.pk },
        { hover: true }
      );
      map.removeFeatureState({ source: "airports", id: i.properties.pk });
    }
```

The stalling behavior does not happen:

<img width="847" height="827" alt="Image" src="https://github.com/user-attachments/assets/589d0253-803a-4158-bdd6-4a27d3e47c28" />

[...]

PR #7554, opened 2026-04-30
FIX Feature state bulk remove + feature set per-id set does not remove state of first feature #7535

### LINKED ISSUE #7535: Feature state bulk remove + feature set per-id set does not remove state of first feature

**browser**: Firefox 150.0

### Steps to Trigger Behavior

Note: The following is assuming a style rule where the fill color is set to `["coalesce", ["feature-state", "color"], "blue"]`, i.e. the fill color is determined by the `"color"` feature state and defaults to blue if feature state is not set.

 1. Set feature state `color: "red"` for all features:
  ```js
  for (const id of ids) {
    map.setFeatureState({ source: "polygons", id }, { color: "red" });
  }
```
 2. Bulk remove feature state:
  ```js
  map.removeFeatureState({ source: "polygons" });
  ```
 3. Set feature state with another key (e.g. `visible`) for all features (can be unused):
  ```js
  for (const id of ids) {
    map.setFeatureState({ source: "polygons", id }, { visible: true });
  }
  ```

### Link to Demonstration

https://jsfiddle.net/6k7wztrq/

_Note: jsbin.com was down when I was reporting this issue, I hope jsfiddle is OK as an alternative._

### Expected Behavior

- Feature state is removed, all polygons are blue

### Actual Behavior

- Feature state of first feature is not removed, first polygon is still red

### AI tool disclosure

Claude Code was used to assist in creating the minimal reproduction code. I have verified myself that a similar issue has not already been reported and have written the issue description myself.

### PR DESCRIPTION
**Pull request by xavierjs**
- Fixes #7535

## Launch Checklist

 - [X] Link to related issues.
 - [X] Post benchmark scores.

### CODE REVIEW

**Comment by xavierjs** [comment]
I took a look at the failing tests.
* The Render tests with windows fail because of a timeout issue ( 40seconds ). 
* The Tests/Code Hygiene fail because unescaped [ ] are in comments, and the doc generation threat them as Markdown stuff. We can fix that by escaping these characters but it will make the comments less readable. In any case it would be better to do that in another PR.

File: src/source/source_state.ts [hunk 1]
```diff
@@ -37,14 +37,21 @@
         if (this.deletedStates[sourceLayer] === null) {
             this.deletedStates[sourceLayer] = {};
             for (const ft in this.state[sourceLayer]) {
-                if (ft !== feature) this.deletedStates[sourceLayer][ft] = null;
+                if (ft === feature){
+                    this.deletedStates[sourceLayer][feature] = {};
```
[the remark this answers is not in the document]
xavierjs: It is not really a security concern but annoying if a user uses some reserved property names like `__proto__` for feature states property names.  I have added some code to trigger an error in this case: https://github.com/xavierjs/maplibre-gl-js/blob/2e51fc5f21af0a8a89be3ef968b73062ad6ac712/src/style/style.ts#L1408

src/source/source_state.ts
```diff
@@ -37,14 +37,21 @@
...
+                    for (const prop in this.state[sourceLayer][ft]){
+                        if (newState[prop] === undefined) this.deletedStates[sourceLayer][feature][prop] = null;
+                    }
+                } else {
+                    this.deletedStates[sourceLayer][ft] = null;
```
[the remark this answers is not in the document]
xavierjs: Same as previous comment reply

src/source/source_state.ts
```diff
@@ -37,14 +37,21 @@
...
+                }
             }
         } else {
             const featureDeletionQueued = this.deletedStates[sourceLayer]?.[feature] === null;
             if (featureDeletionQueued) {
                 this.deletedStates[sourceLayer][feature] = {};
                 for (const prop in this.state[sourceLayer][feature]) {
-                    if (!newState[prop]) this.deletedStates[sourceLayer][feature][prop] = null;
+                    if (newState[prop] === undefined) this.deletedStates[sourceLayer][feature][prop] = null;
```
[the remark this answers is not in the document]
xavierjs: if we update the state with a `false` value, it may consider that this value should be deleted, so we need to compare `newState[prop]` with `undefined`

xavierjs:
Test coverage issues should be solved by the latest commit.

xavierjs:
This PR also fixes This issue Large number of feature state keys leads to zoom lag when loading cached tiles - https://github.com/maplibre/maplibre-gl-js/issues/6633

Here is the performance profiling for @bradymadden97 repro.

**Before** (notice the red spots in the main trains for CPU stalling):
Screenshot 2026-04-30 at 4 40 57 PM

**After**:
Screenshot 2026-04-30 at 4 41 08 PM

src/source/source_state.test.ts
```diff
@@ -20,4 +20,52 @@ describe('SourceFeatureState', () => {
...
+        sourceState.coalesceChanges(inViewTilesMock, {});
+
+        // Delete the entire source layer (sets deletedStates['layer1'] = null)
+        sourceState.removeFeatureState('layer1');
+
+        // Now update feature1 with a partial state — this triggers the branch
+        // where deletedStates[sourceLayer] === null and we iterate existing features
+        sourceState.updateState('layer1', 'feature1', {a: 10});
+
+        // feature1 should have prop 'b' queued for deletion (not in newState),
```
HarelM: This is a bit heavy on the comments, can we keep only the important ones maybe?

src/source/source_state.test.ts
```diff
@@ -20,4 +20,52 @@ describe('SourceFeatureState', () => {
         sourceState.coalesceChanges(inViewTilesMock, painterMock);
         expect(sourceState.revision).toBe(1);
     });
+
+    test('updateState after layer deletion rebuilds deletedStates correctly', () => {
```
HarelM: I think the test name should be using more "user oriented perspective", similar to the bug that this solves - i.e. describe the flow and not the internal implementation, if possible.

HarelM:
Which PR fixes #6633?
This one, or did #7550 fixed it?

I want to link the issue with the relevant PR.
Note that I have updated your initial thread post so that once merged it will close the relevant related issues and link to them. Mainly for bookkeeping and future PRs.
Thanks for all your help!

xavierjs:
> Which PR fixes #6633? This one, or did #7550 fixed it?
> 
> I want to link the issue with the relevant PR. Note that I have updated your initial thread post so that once merged it will close the relevant related issues and link to them. Mainly for bookkeeping and future PRs. Thanks for all your help!

This PR does NOT fix [#6633](https://github.com/maplibre/maplibre-gl-js/issues/6633)

src/source/source_state.test.ts
xavierjs: I have renamed it in the latest commit

src/source/source_state.test.ts
xavierjs: I have removed the less important comments in the latest commit

bradymadden97:
Sent this to Xavier offline, but just copying my note here. If possible it would be great to confirm no regressions via this test case specifically:

> It's possible your code is different and it worked, but it looks similar to something I tried. I can definitely look more closely later! In the meantime, the test case I'd suggest trying: (this is a good starter state if you need one https://github.com/bradymadden97/maplibre-tools/tree/main/repros/state-hover-state) Assuming your feature state depends on hover (like, for example, changing background color of a state when you hover over it)
> 1. Hover over a state (VA) at zoom level X (lets say 5), so that the state turns that color
> 2. While still hovered over the same state, zoom out to level Y (say 7)
> 3. Now unhover that state and go hover some other state
> 4. Re-zoom in to level X (5) without hovering over the original state
> 
> When you re-enter z5, you would expect VA to not be filled, as you're currently not hovering it. But since we now exit early, it uses the cached tile renderered with the old feature state, and VA does have the old filled color again

xavierjs:
I made a mistake for the 2nd repro, I don't manage to get the same result a second time.
I think the lines changing the feature state in the repro code were still commented (it sometimes takes time to refresh).
I need to find something else for this second issue. So this PR does NOT fix 6633. I am sorry for that, next time I will double check.

I checked the test for https://github.com/bradymadden97/maplibre-tools/tree/main/repros/state-hover-state
and this PR does not introduce a regression.

HarelM:
Can you update the first post here with the relevant issue that this PR fixes? Or are you planning to solve both as part of this PR?

xavierjs:
> Can you update the first post here with the relevant issue that this PR fixes? Or are you planning to solve both as part of this PR?

I have updated the post to only mention the issue #7535
The other one is more tricky to fix. I will investigate on it and post a new PR if I found something.

CHANGELOG.md
```diff
@@ -22,7 +22,9 @@
 
 ### 🐞 Bug fixes
 
+- _...Add new stuff here..._
 - ⚠️ Fix transparent, overlapping lines creating artefacts. This is fixed for `line-opacity`, but purposefully not for transparent `line-color` properties, thus still allowing transparent colors to stack their effect. ([#7490](https://github.com/maplibre/maplibre-gl-js/pull/7490)) (by [@CommanderStorm](https://github.com/CommanderStorm))
+- Fix feature state bulk remove + feature set per-id set does not remove state of first feature ([#7554](https://github.com/maplibre/maplibre-gl-js/pull/7554)) (by [@xavierjs](https://github.com/xavierjs))
```
HarelM: Make sure to move this up to the beginning of the changelog.

HarelM (approved):

PR #7563, opened 2026-05-01
(chore): reduce if complexity in draw line

HarelM:
This reduces the complexity of the if so that it's easier to understand what drawing is done for each case.

## Launch Checklist

### DIFF
--- a/src/webgl/draw/draw_line.ts
+++ b/src/webgl/draw/draw_line.ts
@@ -151,28 +151,33 @@ export function drawLine(painter: Painter, tileManager: TileManager, layer: Line
 
     const useOffscreen = layer.hasOffscreenPass() && !painter.style.map.terrain;
 
-    // GC the FBO if style is transtitoned from having opacity to not having opacity
     if (!useOffscreen && layer.lineFbo) {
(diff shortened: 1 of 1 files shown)


CommanderStorm:
I understand the previous version better. There it was more clear that if we need an offscreen pass, what happens.

CommanderStorm:
So I would like to stick with the previous version that I understand better than the lengthy if-return where it is much simpler to make mistakes.

CommanderStorm (approved):
if you insist to push this through here is the +1, I would prefer the previous code though and would not write it out in such an dis-aggregatied way.
Much harder to understand overall.

HarelM:
When reading this part it was not clear to me what happens in each pass, the `useOffscreen` is secondary to which pass is currently applied.
And in general, more indentation is considered more code complexity.
This explicitly states which draw call is executed.
If you'd like to switch between the `useOffscreen` and rednderpass that could be the middle ground.

CommanderStorm:
I somewhat disagree, `useOffscreen` is the primary thing that this code cares about, not the internal render passes we use to achive this.
So the user-facing issue of why this actually is 4 different cases..

As I said: if you insist, merge.
I would prefer the current version.

PR #7564, opened 2026-05-01
Use rolldown-plugin-dts+tsgo for 78x faster .d.ts generation

birkskyum:
Use rolldown-plugin-dts+tsgo for .d.ts generation. 

generate-typings-time-by-tool-with-tsgo-horizontal

Rolldown is [already](https://npmtrends.com/dts-bundle-generator-vs-rolldown-plugin-dts-vs-rollup-plugin-dts) the most common way to do this

Screenshot 2026-05-01 at 16 43 22

**Test**

- `npm run generate-typings` - produces dist/maplibre-gl.d.ts
- `tsgo --project tsconfig.dist.json` - checks the bundled .d.ts is 

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
Thanks! I was looking for alternative to this one as it seems like the next bottleneck in the build. You can take a look at my branch, but it was slow too...

HarelM:
The branch is here for reference:
https://github.com/maplibre/maplibre-gl-js/tree/replace-dts-bundle-generator

If you needs to solution to the integration tests stuff.

rolldown.config.dts.ts
HarelM, line 1: Can we move it inside the build itself, maybe for either production or dev, or both? See my branch for reference.

HarelM:
There's a note in the docs about using tsgo, maybe we can leverage this as well?

rolldown.config.dts.ts
birkskyum: Yes, I've now added it to rolldown.config.ts. It means that the modes are:

- generate-typings (only .d.ts) ~ 650ms
- **build-prod ~ now 850 ms instead of 200ms because it includes .d.ts**
- build-dev ~ 200ms
- build-dist ~4s

birkskyum:
> There's a note in the docs about using tsgo, maybe we can leverage this as well?

At this time we can't. They are still working on a declaration flag for tsgo, but i don't think it's in yet. I can keep looking for ways to leverage it, but I think this setup already improves so much on the .d.ts generation time that the gains would be small.

rolldown.config.dts.ts
HarelM: I think you can remove generate types...

HarelM:
Yeah, generating this in 1 sec is definitely good enough.

birkskyum:
@HarelM , actually, we can use tsgo with rolldown - it jsut took the .d.ts from 1.4s to ~650ms 

so tsgo makes a .d.ts for each file, and rolldown bundles it.

test/build/es2022-compat.test.ts
CommanderStorm, line 1: Can we bump this file to es2022?
our tsconfig does only have es2022 and I did not notice this test

birkskyum: Yes, i've bumped it now

rolldown.config.dts.ts
birkskyum: I am using that command a lot currently, for tinkering with isolated declarations, so if it can stay for just a bit longer that would be great

HarelM:
I don't mind keeping it if you remember to remove it in the near future.

CommanderStorm (approved):
LGTM from my side

HarelM (approved):

PR #7565, opened 2026-05-01
chore: replace ts-node with native Node 24 typescript support

birkskyum:
Node 24 can run typescript ( with [Amaro](https://github.com/nodejs/amaro) ) so we don't need ts-node anymore. The requirement is that we add .ts/.tsx on imports, which is a bit more strict, but more correct and follow the tooling-trend towards less ambiguity. Apart from one less dependency we also get a speedup:

codegen-scripts-ts-node-vs-node24-native-clean

Assisted-By: Opus 4.7

## Launch Checklist

 - [ ] Link to related issues.
 - [x] Post benchmark scores.


HarelM:
Why do we need to change the import extension on all the source code files when those are mostly not in use for the generation scripts...?

HarelM:
I'll rephrase the question, can we reduce the scope of this PR to only the generation scripts?

birkskyum:
I can't reduce it much. It's because generate-struct-arrays.ts is pulling from ./src which recursively touches > 370 files, and at that point might as well do them all ~500. So, the 7 other generation scripts can avoid this, but generate-struct-arrays.ts requires the extensions, and it's the script where there is most to win.

HarelM:
No that I mind changing all the file extensions as it's a very technical and non risky charge, but in the past this was not the correct way to include typescript files. 
If this is now broadly supported and doesn't cause any issues, I'm fine with removing ts-node dependecy.

HarelM (approved):
I didn't inspect all files, but looks good in general

PR #7570, opened 2026-05-03
feat: implement `{line,fill}-layer-opacity`

CommanderStorm:
This basically does the same thing as previously.

Changes:
- [x] lines
- [x] fills

image

Currently blocked by:
- https://github.com/maplibre/maplibre-style-spec/pull/1650


src/style/style_layer/fill_style_layer.ts
```diff
@@ -54,7 +58,21 @@ export class FillStyleLayer extends StyleLayer {
...
     }
 
-    isTileClipped(): boolean {
+    isTileClipped() {
         return true;
     }
+
+    hasOffscreenPass() {
+        const layerOpacity = this.paint.get('layer-opacity' as any) as number;
+        return layerOpacity > 0 && layerOpacity < 1 && !this.isHidden();
```
lucaswoj: We'll need to create a PR against the style spec too for this property. Will we support any data driven styling out of the gate? Zoom functions? Global state?

src/webgl/draw/draw_line.ts
```diff
@@ -152,21 +152,18 @@ export function drawLine(painter: Painter, tileManager: TileManager, layer: Line
     const useOffscreen = layer.hasOffscreenPass() && !painter.style.map.terrain;
```
lucaswoj: Does `layer-opacity` work with 3D terrain? This expression makes it look like users can't use these features at the same time.

src/webgl/draw/draw_fill.ts
```diff
@@ -42,21 +62,57 @@ export function drawFill(painter: Painter, tileManager: TileManager, layer: Fill
...
+    if (!fbo) return;
+
+    const context = painter.context;
+    const gl = context.gl;
+
+    context.activeTexture.set(gl.TEXTURE0);
+    gl.bindTexture(gl.TEXTURE_2D, fbo.colorAttachment.get());
+
+    const opacity = layer.paint.get('layer-opacity' as any) as number;
+    painter.useProgram('lineTexture').draw(context, gl.TRIANGLES,
```
lucaswoj: Why does this use the `lineTexture` program?

lucaswoj (commented):
This is on the right track!

src/webgl/draw/draw_line.ts
CommanderStorm: Current `layer-opacity` change also ignores this because I honestly don't quite know what oppacity with 3d terrain should do.

There is some code there which would need more time to understand than I timeboxed on this.


| terrain disallows offscreen | terrain allows offscreen |
|--------|--------|
| image | image |

src/webgl/draw/draw_fill.ts
CommanderStorm: because copy-paste fail I think 😅

CommanderStorm: Actually `lineTexture` program is a generic "draw a textured fullscreen quad with opacity" shader -> it doesn't do anything line-specific.
Maybe something like `compositeTexture` might be a better name, not sure.

CommanderStorm: lets actually name the program `layerOpacity`..

test/integration/render/tests/line-layer-opacity/opacity-zero/expected.png
HarelM, line 1: Is `line-layer-opacity: 0` ignored? as I would've expected this not to show the line at all, am I missing anything?

test/integration/render/tests/fill-layer-opacity/opacity-zero/expected.png
HarelM, line 1: Same as previous comment about opacity 0, but for fill layer.

HarelM:
This is a bit hard to review because half the PR here is a revert of the previous feature. I think it would be better to split this into two PRs.
Sorry for the extra work, but I'm having a hard time understanding which part is related to the revert and which is related to the new code...

src/webgl/program/layer_opacity_program.ts
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Sub-pass for `{line,fill}-layer-opacity`.
+ * Renders the layer image into the shared scratch FBO at the target framebuffer's resolution via `renderToScratch`.
+ * Then it composites it back into the previously bound framebuffer with `opacity` as a multiplier.
+ * Canvas in the flat case, a terrain RTT tile texture in the terrain case.
+ *
+ * Compositing this way (single quad per layer) applies opacity uniformly to the
+ * whole layer output instead of accumulating alpha across overlapping geometry.
+ */
+const drawLayerOpacitySubpass = (
```
HarelM: Is it possible to split this method to "before and after" instead of passing a lambda to it?
I understand the desire to reduce code duplication, but I think the readability of this is compromised...

src/render/painter.ts
```diff
@@ -107,6 +107,17 @@ export class Painter {
         depthRenderbuffer: WebGLRenderbuffer;
         size: number;
     } | null;
+    /**
+     * Shared scratch FBO (color + depth-stencil) used by `{line,fill}-layer-opacity`.
```
HarelM, line 110: What's the meaning of "scratch" in this context? The fact that it is "disposable"?

src/render/painter.ts
```diff
@@ -910,6 +958,11 @@ export class Painter {
             this._rttSharedFbo = null;
         }
 
+        if (this._layerOpacityScratchFbo) {
+            this._layerOpacityScratchFbo.fbo.destroy();
```
HarelM: Can we use `.?` instead of the if?

src/webgl/draw/draw_fill.ts
```diff
@@ -57,6 +70,26 @@ export function drawFill(painter: Painter, tileManager: TileManager, layer: Fill
     }
 }
 
+function drawFillAndOutline(
```
HarelM, line 67: This is probably related to my previous comment about splitting this PR, but I'm not sure I understand where the code for this method was before this change, as I would expect it to exists somewhere? I guess it's not new, and overall looks good, but I'm just confused...

src/render/painter.ts
CommanderStorm: Scratch in the meaning of https://en.wikipedia.org/wiki/Scratch_space
Not disposable, but reusable and temporary.

src/webgl/program/layer_opacity_program.ts
CommanderStorm: Would add the possibility to misuse the api, but sure.

src/render/painter.ts
HarelM: I think we can remove the word scratch then...

test/integration/render/tests/line-layer-opacity/globe/expected.png
CommanderStorm, line 1: non-layer variant as a reference.
Is correct.

Image

test/integration/render/tests/fill-layer-opacity/globe/expected.png
CommanderStorm, line 1: non-layer variant as a reference, is correct

Image

test/integration/render/tests/layer-opacity/mixed-layer-opacity/expected.png
CommanderStorm, line 1: regarding test folder structure, if this were not a test that is kind of "unique", both `*-layer-opacity` would be top level constructs, like the rest.

Not sure if this is actually a case that is worth keeping.
I highly doubt that this will ever catch a regression, someone would have disregarded the things that we semi-structurally require (begin->end, no-parallel layer-begins, ...)
I am pretty sure at that point, other testcases would likely also break, given that the only thing that they don't test is layer-mix-overlay...

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## main
 ### ✨ Features and improvements
+- Add `fill-layer-opacity` and `line-layer-opacity` paint properties, which apply opacity to the entire layer output uniformly.
+- Revert the `line-opacity`-driven offscreen rendering introduced in [#7490](https://github.com/maplibre/maplibre-gl-js/pull/7490). The overlap-artefact fix is now driven by `line-layer-opacity` instead.
```
HarelM: Proabably a bad merge...

test/integration/render/tests/layer-opacity/mixed-layer-opacity/expected.png
HarelM: Your call, I'm OK eitherway.

test/integration/render/tests/fill-layer-opacity/terrain/style.json
HarelM, line 1: Does this really capture this feature? This style has only one polygon, to see the effect there should be at least two polygons, right?

test/integration/render/tests/fill-layer-opacity/function/style.json
HarelM, line 1: Shouldn't there be 4 triangles in this test?

src/render/painter.ts
```diff
@@ -107,6 +108,24 @@ export class Painter {
...
+     * The canvas in the flat case, the per-terrain-tile RTT texture in the terrain case.
+     * Resized in place to match the target dimensions.
+     */
+    private _layerOpacityFbo: Framebuffer | null;
+    /**
+     * Transient state for an in-flight `{line,fill}-layer-opacity` subpass.
+     * Set by {@link beginLayerOpacitySubpass}, cleared by {@link endLayerOpacitySubpass}.
+     * Non-null exactly while a subpass is in flight.
+     */
+    private _layerOpacitySubpass: {
```
HarelM: Can this object be returned from the begin method and passes to the end method instead of keeping the state on the object?

src/render/painter.ts
```diff
@@ -789,6 +810,87 @@ export class Painter {
         this._rttObjectRecyclePool.push(obj);
     }
 
+    /**
+     * Begin a `{line,fill}-layer-opacity` subpass. Saves the currently bound framebuffer / viewport,
+     * binds the shared scratch FBO at the target's resolution, clears it, and writes the layer's clipping masks.
+     * The caller then renders the layer into the scratch FBO and finishes with {@link endLayerOpacitySubpass}.
+     */
+    beginLayerOpacitySubpass(layer: LineStyleLayer | FillStyleLayer, coords: OverscaledTileID[], terrain: boolean): void {
```
HarelM: I'm looking at a similar code here:
https://github.com/maplibre/maplibre-gl-js/pull/7635

I'm wondering if this logic shouldn't be placed in the `draw` folder and expose the relevant methods. What do you think?

HarelM:
I've added a few comments.
Thanks for the revert PR, it allowed me to see the added part here, which is a lot smaller than I thought initially.
The render tests you added are great, thanks!
Overall, looks good.

src/render/painter.ts
CommanderStorm: I mean yea, but that would likely be a lot of GC pressure to add.
That object would be created every frame and layer which has this style prop..

HarelM: The code is assigning it as part of every call to `beginLayerOpacitySubpass` how would that be any different in terms of GC pressure?

CommanderStorm: Well `_layerOpacitySubpass` still exists, so I doubt that JS would remove the object, or?

test/integration/render/tests/fill-layer-opacity/function/style.json
CommanderStorm: was wrong zoom and really does not need 4, only two

test/integration/render/tests/fill-layer-opacity/terrain/expected.png
HarelM, line 1: A second fill geometry is still missing in this test...

test/integration/render/tests/fill-layer-opacity/opacity-one/expected.png
HarelM, line 1: A second object is missing here too.

test/integration/render/tests/layer-opacity/mixed-layer-opacity/expected.png
CommanderStorm: removed the testcase. I don't think it is an imprtant test and I think all potential breakages would really be caught already..

test/integration/render/tests/fill-layer-opacity/function/expected.png
HarelM, line 1: Triangles should be overlapping I believe.

test/integration/render/tests/fill-layer-opacity/fill-pattern/expected.png
HarelM, line 1: It would be interesting to see how two objects with pattern are behaving instead of just one.

test/integration/render/tests/fill-layer-opacity/terrain/style.json
CommanderStorm: not really.
With this test I am testing that the terrain does not weirdly warp the geometry.
So the NEC math error I did that lead to this funny artefact: https://github.com/maplibre/maplibre-gl-js/pull/7570#discussion_r3178945632

test/integration/render/tests/fill-layer-opacity/fill-outline-color/expected.png
HarelM, line 1: What would happen in this case for two objects?

test/integration/render/tests/fill-layer-opacity/fill-outline-color-translucent/expected.png
HarelM, line 1: Same here - two objects.

src/render/painter.ts
HarelM: It would, as the previous object is not pointed to anymore.

HarelM:
Code review comments were marked as resolved although they were not addressed. Please link to a commit if you decide to mark as review as resolved to avoid this in the future.

test/integration/render/tests/fill-layer-opacity/terrain/style.json
HarelM: The lines in the render test for terrain show that even though they overlap they don't accumulate opacity, I think it should be added here as well.

test/integration/render/tests/fill-layer-opacity/terrain/expected.png
HarelM: Sorry, I didn't see my previous comment for some reason. this is a dup...

test/integration/render/tests/fill-layer-opacity/fill-outline-color-translucent/expected.png
CommanderStorm: https://github.com/maplibre/maplibre-gl-js/pull/7570/commits/4070ccfff9a10c9ed8b564204af20f38c985ea13

test/integration/render/tests/fill-layer-opacity/fill-outline-color/expected.png
CommanderStorm: 🤷🏻 not sure if the output here is correct

https://github.com/maplibre/maplibre-gl-js/pull/7570/commits/4070ccfff9a10c9ed8b564204af20f38c985ea13

test/integration/render/tests/fill-layer-opacity/fill-pattern/expected.png
CommanderStorm: works fine I think https://github.com/maplibre/maplibre-gl-js/pull/7570/commits/4070ccfff9a10c9ed8b564204af20f38c985ea13

test/integration/render/tests/fill-layer-opacity/fill-outline-color-translucent/expected.png
CommanderStorm: or more rather https://github.com/maplibre/maplibre-gl-js/pull/7570/commits/138e492b5b927cefc6645a96376d9df39c93462b

test/integration/render/tests/fill-layer-opacity/function/expected.png
CommanderStorm: Between layer overlap is tested in the others already... don't think that is an interesting case..

CommanderStorm: this is testing that functions work which by itsself is something that I also don't entirely find an interesting case to test (i don't think this is adding much coverage if any), but that is what we do here..

test/integration/render/tests/fill-layer-opacity/opacity-one/expected.png
CommanderStorm: Don't think so, opacity=1 means we don't run the code I added..
So would not add coverage and thus not interesting..

CommanderStorm:
> Code review comments were marked as resolved although they were not addressed. Please link to a commit if you decide to mark as review as resolved to avoid this in the future.

Was just a race condition of me working through your comments

test/integration/render/tests/fill-layer-opacity/terrain/style.json
CommanderStorm: does not really add any value or coverage, but sure

CommanderStorm: expected

In case you are wondering what kind of a cursed geometry this is now

src/render/painter.ts
CommanderStorm: we need access to `context` which is a property of `Painter`..
So we need to be in Painter I think.

test/integration/render/tests/fill-layer-opacity/fill-outline-color/expected.png
HarelM: At least we know how it looks now :-)

test/integration/render/tests/fill-layer-opacity/fill-pattern/expected.png
HarelM: Looks good, yes.

test/integration/render/tests/fill-layer-opacity/terrain/style.json
HarelM: But this is a single polygon... I think that this should be 2 squares instead of this weird form.
While it might not add to coverage, it does make sure that we can see that opacity is not accumulated for fill over terrain.

src/render/painter.ts
HarelM: In other draw methods and files painter is passes as a parameter.

test/integration/render/tests/fill-layer-opacity/opacity-one/expected.png
HarelM: It not about code coverage, it's about making sure we are testing this also for future changes. If someone will change the code flow for opacity in the future this might catch it.

test/integration/render/tests/fill-layer-opacity/function/expected.png
HarelM: While not adding coverage it better to test the function don't break the code somehow. copying the shapes from another test should be fairly easy.

test/integration/render/tests/fill-layer-opacity/opacity-one/expected.png
CommanderStorm: if you insist.. https://github.com/maplibre/maplibre-gl-js/pull/7570/commits/e35da1329b6980c35a8d6e64f4ef91d33ed57550

test/integration/render/tests/fill-layer-opacity/terrain/style.json
CommanderStorm: added

test/integration/render/tests/fill-layer-opacity/function/expected.png
CommanderStorm: added two squars

src/render/painter.ts
CommanderStorm: If you want `begin` and `end` to be more clearly separated, I'm happy to do that — I've already pushed a commit implementing the change.

That said, I think this moves us in the wrong direction from a design perspective. The fact that every `begin` must have a matching `end` is a structural invariant. Originally, the closure-based API made violating that invariant impossible. The earlier `begin`/`end` split weakened that guarantee, and splitting the implementation across files weakens it further.

Personally, I prefer invariants to be enforced by the structure of the code whenever possible rather than relying on callers to remember the correct usage pattern.

In this case, `begin` also needs to originate from `Painter`, since it requires access to the shared per-map FBO as well as several internal painter methods such as `_renderTileClippingMasks`. It therefore has to pass through `context`, `gl`, and other private painter state.

Regarding the naming, `draw` did not feel quite right to me because this code is not actually performing any drawing. However, I tried to align the implementation with your request.

HarelM:
Damn, I broke the tests. I'll open another PR to fix it unless you'll beat me to it.

HarelM (approved):

PR #7590, opened 2026-05-05
[FIX] Fix Large number of feature state keys leads to zoom lag when loading cached tiles - 6633

### LINKED ISSUE #6633: Large number of feature state keys leads to zoom lag when loading cached tiles

xavierjs:
## Launch Checklist

* Fixes https://github.com/maplibre/maplibre-gl-js/issues/6633

Cache the feature state revision in the tiles and early exits if the feature state revision is the same.

This PR does not affect the `PaintStates` benchmark.
`npm run test-render` runs as before.

## Benchmark
Copy/paste this file: https://github.com/xavierjs/maplibre-gl-js/blob/debug/reproIssue6633/test/examples/0_repro_brady.html to `/test/examples/repro.html`, then run `npm run start`and open http://localhost:9966/test/examples/repro.html

Wait about 10 seconds till the button *ZOOM AND DEZOOM* appear on the top left (it will load the map and zoom at different zoom levels to load the tiles).
Then click on *ZOOM AND DEZOOM* and start a performance benchmark for 10 seconds in the web console.
 
Before this PR (main branch) - some red dots show that the browser is not responsive during the layout changes (the tiles are reloaded even if the feature state is not updated):
Screenshot 2026-05-06 at 1 58 42 PM

After this PR - the browser is responsive all the time:
Screenshot 2026-05-06 at 1 56 50 PM

 - [X] Link to related issues.
 - [X] Post benchmark scores.


HarelM:
@xavierjs ping me when this is ready for review, I see that you are still fiddling with it... (maybe convert it to draft until you finish playing around with it)?

xavierjs:
@HarelM It is ready since the latest commit, all the render tests are passing now :)

This PR tile cache granularity is fairly coarse: we only cache the feature state revision. We could potentially cache more per-tile data to update only the tiles whose used feature state has changed. However, that would increase memory usage and add complexity.

This PR takes a lightweight approach and addresses the main issue: slow interactivity after setting feature state on many elements.

HarelM:
This looks fairly simple and good overall.
Can you think of a test or two worth adding to cover this new functionality?
I'm mainly concerned that this might break some set feature state scenarios as I believe feature state is not well tested in general...

HarelM (approved):
Thanks!!

PR #7595, opened 2026-05-07
Fix CDN usage

birkskyum:
This auto loads the worker module so that we can keep the unpkg.com install path.

For usage with bundlers, the worker url should still be set explicitly.

AI-Assisted: Claude 4.7

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


src/util/web_worker.ts
```diff
@@ -12,13 +13,114 @@ export interface WorkerGlobalScopeInterface {
...
+    const response = await fetch(url);
+    if (!response.ok) {
+        throw new Error(`Failed to fetch worker script (${response.status}): ${url}`);
+    }
+    const code = await response.text();
+    const blob = new Blob([code], {type: 'text/javascript'});
+    return URL.createObjectURL(blob);
+}
+
+class DeferredWorker implements ActorTarget {
```
HarelM: I'm a bit concerned about this code.
Can we maybe change the solution to make `WorkerPool.aquire` async, and make the `createWorker` async as well? So that we create the worker/actor only after the fetch is complete maybe? Would that work?

HarelM:
Thanks for looking into this BTW! I truly appreciate it.

src/index.ts
```diff
@@ -17,6 +17,9 @@ import Point from '@mapbox/point-geometry';
 import {MercatorCoordinate} from './geo/mercator_coordinate.ts';
 import {Evented, type ErrorEvent, Event, type Listener} from './util/evented.ts';
 import {type AddProtocolAction, config} from './util/config.ts';
+import {defaultWorkerUrl} from './util/default_worker_url.ts';
+
+config.WORKER_URL = defaultWorkerUrl;
```
CommanderStorm: not great for side-effect-free-ness, but not sure of there is a bettter way.

Also, did you mean this?


```suggestion
import {defaultWorkerUrl} from './util/default_worker_url.ts';

config.WORKER_URL = defaultWorkerUrl();
```

CommanderStorm (changes requested):
There are a few cases where `setWorkerUrl`
is still used the "old" way.

- [docs/guides/mapbox-migration-guide.md](https://github.com/maplibre/maplibre-gl-js/blob/4412ee1bd912f669534fc801aaae87aaf8066274/docs/guides/mapbox-migration-guide.md?plain=1#L37)
- [docs/guides/v5-to-v6-migration-guide.md](https://github.com/maplibre/maplibre-gl-js/blob/4412ee1bd912f669534fc801aaae87aaf8066274/docs/guides/v5-to-v6-migration-guide.md?plain=1#L18)
- [EADME.md](https://github.com/maplibre/maplibre-gl-js/blob/4412ee1bd912f669534fc801aaae87aaf8066274/README.md?plain=1#L32)
- ... -> https://github.com/search?q=repo%3Amaplibre%2Fmaplibre-gl-js+setWorkerUrl+-path%3A%2F%5Etest%5C%2Fexamples%5C%2F%2F&type=code

Could you update the rest too? ^^

src/tile/tile.ts
```diff
@@ -95,7 +95,7 @@ export class Tile {
     redoWhenDone: boolean;
     showCollisionBoxes: boolean;
     placementSource: any;
-    actor: Actor;
+    actor: Promise<Actor>;
```
HarelM: I think you can keep this one without a promise, right?

src/util/dispatcher.ts
```diff
@@ -6,67 +6,95 @@ import type {WorkerPool} from './worker_pool.ts';
...
 /**
  * Responsible for sending messages from a {@link Source} to an associated worker source (usually with the same name).
  */
 export class Dispatcher {
     workerPool: WorkerPool;
     actors: Actor[];
+    actorsPromise: Promise<Actor[]>;
     currentActor: number;
     id: string | number;
+    private pendingHandlers: PendingHandler[];
```
HarelM: I'm not sure this is needed. Can we simply make the register async as well?

src/util/dispatcher.ts
```diff
@@ -6,67 +6,95 @@ import type {WorkerPool} from './worker_pool.ts';
...
-        for (let i = 0; i < workers.length; i++) {
-            const worker = workers[i];
-            const actor = new Actor(worker, mapId);
-            actor.name = `Worker ${i}`;
-            this.actors.push(actor);
-        }
-        if (!this.actors.length) throw new Error('No actors found');
+        this.pendingHandlers = [];
+        this.removed = false;
+        this.actorsPromise = workerPool.acquire(mapId).then(workers => {
```
HarelM: Consider moving this code to a method - something like `initActors` or something.

build/generate-docs.ts
```diff
@@ -95,7 +95,7 @@ function generateMarkdownForExample(title: string, description: string, file: st
 
 ${description}
 
-<iframe src="../${file}" width="100%" style="border:none; height:400px"></iframe>
+<iframe src="${file}" width="100%" style="border:none; height:400px"></iframe>
```
HarelM, line 98: Are you sure about this change? I don't think we changed the folder structure...

src/source/geojson_source.test.ts
```diff
@@ -143,7 +143,8 @@ describe('GeoJSONSource.setData', () => {
             } as any as RequestManager
         } as any;
         const spy = vi.fn();
-        source.actor.sendAsync = (message: ActorMessage<MessageType>) => {
+        const actor = await source.actor;
+        actor.sendAsync = (message: ActorMessage<MessageType>) => {
```
HarelM: I've looked into these tests, I think it will be better to move this part to the initialization of the source:
```ts
const source = new GeoJSONSource('id', {collectResourceTiming: true} as any, wrapDispatcher({
            sendAsync: (message: ActorMessage<MessageType>) => {
                return new Promise((resolve, reject) => {
                    if (message.type === MessageType.loadData) {
                        setTimeout(() => resolve({} as any), 0);
                        spy(message);
                    } else {
                        reject(new Error(`MessageType.loadData is expected but got ${message.type}`));
                    }
                });
            }
        }), undefined);
        source.map = {
            _requestManager: {
                transformRequest: (url:string) => ({url})
            } as any as RequestManager
        } as any;
...
```

src/source/geojson_source.test.ts
```diff
@@ -167,7 +168,8 @@ describe('GeoJSONSource.setData', () => {
...
```
HarelM: Same here:
```ts
const spy = vi.fn();
        const source = new GeoJSONSource('id', {collectResourceTiming: true} as any, wrapDispatcher({
            sendAsync(message: ActorMessage<MessageType>) {
                return new Promise((resolve) => {
                    if (message.type === MessageType.loadData) {
                        spy(message);
                        resolve({});
                    }
                });
            }
        }), undefined);
        source.map = {
            _requestManager: {
                transformRequest: async (url: string) => ({url})
            } as any as RequestManager
        } as any;
```

src/source/geojson_source.ts
```diff
@@ -140,7 +140,7 @@ export class GeoJSONSource extends Evented implements Source {
     _options: GeoJSONSourceInternalOptions;
     workerOptions: GeoJSONWorkerOptions;
     map: Map;
-    actor: Actor;
+    actor: Promise<Actor>;
```
HarelM: Should this be called `actorPromise` maybe?
Other sources are getting an actor only when they need instead of storing it in the class.
Does it make sense to store the dispatcher instead of the actor as a field memeber?

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 ### 🐞 Bug fixes
 - Fix feature state bulk remove + feature set per-id set does not remove state of first feature ([#7554](https://github.com/maplibre/maplibre-gl-js/pull/7554)) (by [@xavierjs](https://github.com/xavierjs))
 - Remove error when actor doesn't have a registered message type for better usability of custom messages in workers ([#7589](https://github.com/maplibre/maplibre-gl-js/issues/7589)) (by [@HarelM](https://github.com/HarelM))
+- Auto-load worker module for CDN usage
```
```suggestion
- Auto-load worker module for CDN usage ([#7595](https://github.com/maplibre/maplibre-gl-js/pull/7595)) (by [@birkskyum](https://github.com/birkskyum))
```

test/unit/lib/web_worker_mock.ts
```diff
@@ -1,7 +1,10 @@
 import MapLibreWorker from '../../../src/source/worker.ts';
+import {config} from '../../../src/util/config.ts';
 import type {WorkerGlobalScopeInterface} from '../../../src/util/web_worker.ts';
 import type {ActorTarget} from '../../../src/util/actor.ts';
 
+config.WORKER_URL = '';
```
HarelM: Doesn't the factory handles undefined/null. Are you sure this is needed?

src/index.ts
HarelM: I think this logic can happen in the workerFactory, when the `WORKER_URL` is empty, right?

build/generate-docs.ts
birkskyum: It's a separate issue that I just included in this PR. The latest zensical does some file rewriting (use_directory_urls: true), which means we have to update this.

src/util/worker_pool.ts
```diff
@@ -15,31 +15,44 @@ export class WorkerPool {
...
-        return this.workers.slice();
+        return (await this.workersPromise).slice();
     }
 
     release(mapId: number | string): void {
         delete this.active[mapId];
-        if (this.numActive() === 0) {
-            for (const w of this.workers) {
-                w.terminate();
+        if (this.numActive() === 0 && this.workersPromise) {
```
HarelM, line 38: I think an `await` or `then` are missing here...

HarelM: Nevermind...

HarelM: No, wait, this code still looks a bit off.
See if you can simplify this somehow, the active and workersPromise is confusing...

src/util/worker_pool.ts
```diff
@@ -15,32 +15,40 @@ export class WorkerPool {
         [_ in number | string]: boolean;
     };
     workers: ActorTarget[];
+    private workersPromise: Promise<ActorTarget[]> | null;
```
HarelM: Are both workers and workersPromise needed? Maybe the promise is enough?

src/tile/tile.ts
birkskyum: I think we need this. I got a race condition in vector_tile_source because of two loadTile calls, and this solves it.

HarelM: Are you sure? Why would that solve it?

HarelM:
The only open question is if we can avoid changing the tile actor field. Otherwise this looks great. Thanks for pushing this through.

src/tile/tile.ts
birkskyum: I'll try to remove teh promise

birkskyum: it's now `actor: Actor;`

CommanderStorm (approved):

src/source/raster_dem_tile_source.ts
```diff
@@ -87,8 +87,11 @@ export class RasterDEMTileSource extends RasterTileSource implements Source {
                 if (tile.actor && tile.state !== 'expired' && tile.state !== 'reloading') {
                     return;
                 }
+                if (!this.dispatcher.actors.length) {
+                    await this.dispatcher.actorsPromise;
```
HarelM: I'm not fully aligned with this pattern, can you elaborate on it a bit?
Also it might be better to expose a method like `waitForInitComplete` instead of exposing internal fields...
Also how is this change - wait for this.dispatcher.actorsPromise and then this.dispatcher.getReadyActor different than `await this.dispatcher.getActor()`?

src/util/dispatcher.ts
```diff
@@ -12,60 +12,74 @@ import {MessageType} from './actor_messages.ts';
 export class Dispatcher {
     workerPool: WorkerPool;
     actors: Actor[];
+    actorsPromise: Promise<Actor[]>;
```
HarelM, line 15: It might be that `actors` filed is no longer in use and is fully replaced with the actorsPromise?

src/source/raster_dem_tile_source.ts
birkskyum: I've added the waitForInitComplete

src/util/dispatcher.ts
birkskyum: I think we still need actors, since it's the sync cache behind getReadyActor(). That has to stay synchronous to avoid a concurrent-loadTile race, and Promises don't expose a sync read.

Source files no longer touch it directly. They call waitForInitComplete() then getReadyActor(). The only readers of actors now are Dispatcher's own methods, so it could technically be made private, but then we'd have to cast "( .. as any).actors" in unit tests for inspection.

HarelM: If the unit tests are touching something that is internal then they are probably doing something wrong. 
Having said that, fixing it should not prevent merging the PR.

src/source/geojson_source.test.ts
```diff
@@ -16,12 +16,19 @@ import type {IActor} from '../util/actor.ts';
 import type {MapSourceDataEvent} from '../ui/events.ts';
 import type {GeoJSONSourceDiff, UpdateableGeoJSON} from './geojson_source_diff.ts';
 
-const wrapDispatcher = (dispatcher: IActor) => {
+const wrapDispatcher = (actor: IActor) => {
```
HarelM: I think this is a duplicate of the method you changed in utils, it's it? Maybe it can be removed?
getWrapDispatcher.

src/util/dispatcher.ts
```diff
@@ -12,60 +12,80 @@ import {MessageType} from './actor_messages.ts';
...
+        }
+    }
+
+    getReadyActor(): Actor {
         this.currentActor = (this.currentActor + 1) % this.actors.length;
         return this.actors[this.currentActor];
     }
 
     remove(mapRemoved: boolean = true): void {
+        this.removed = true;
```
HarelM, line 71: Where is remove checked? I see only assignment, or did I miss anything thing?

src/source/raster_dem_tile_source.ts
HarelM: This is better, but I still don't understand how a race condition can occur in this case. What will happen if two load tile requests get the same actor? Or what race condition does this solve?

birkskyum: The problem is the `if (!tile.actor)` check. If `loadTile` is called twice, both reach `if (!tile.actor)` (true for both) and both set `tile.actor`.

Moving `await dispatcher.waitForInitComplete()` above the check fixes it: after the await, the first call sets `tile.actor` synchronously (no `await` between check and assignment), so when the second call's continuation runs, `!tile.actor` is false and it takes the reload branch.

src/util/dispatcher.ts
birkskyum: It's checked in `initActors` (same file, on line 31). If `remove()` is called while worker acquisition is still in flight, the `.then` callback sees `this.removed === true` and bails out instead of creating actors that would never be used.

HarelM (approved):

birkskyum:
Ran docs locally. This seems to work with 6.0.0-9

PR #7606, opened 2026-05-08
Fix benchmark stall issue

xavierjs:
Hi,

## Issue

Some benchmark take forever on my laptop (Macbook M4 pro max), for example this one: http://localhost:9966/test/bench/versions/index.html?compare=#ExpressionCreate
The issue is with the calibration stage, in `benchmark.ts`:

```javascript
 while (true) {
    const time = performance.now() - this._start;
    this._elapsed += time;
    if (time < minTimeForMeasurement) {
        this._iterationsPerMeasurement++;
    } else {
        this._measurements.push({time, iterations: this._iterationsPerMeasurement});
    }
    if (this._done()) {
        return this._end();
    }
    this._start = performance.now();
    for (let i = this._iterationsPerMeasurement; i > 0; --i) {
        this.bench();
    }
}
```

Since my computer is fairly fast, `time < minTimeForMeasurement` is always `true`. As a result, we increment `this._iterationsPerMeasurement` and rerun the benchmark without ever reaching the minimum duration. This accumulates over time and slows down the process. Eventually, the browser displays an alert warning that a script is taking too long to run.

## Solution

We don't need an exact value for `this._iterationsPerMeasurement`; we only need a value that makes the task run long enough for the measurement to be meaningful.
So, instead of increasing it linearly, using a mild exponential increase avoids this issue.

After that the full benchmark suite passes (http://localhost:9966/test/bench/versions/index.html?compare=).

I have also increased the `minTimeForMeasurement` from 5ms to 20ms so that the measurement are more meaningful (There are less benchmarks with low R values).

## Misc

No related issues.

## Launch Checklist
 - [X] Link to related issues.
 - [X] Post benchmark scores.


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 - Fix feature state bulk remove + feature set per-id set does not remove state of first feature ([#7554](https://github.com/maplibre/maplibre-gl-js/pull/7554)) (by [@xavierjs](https://github.com/xavierjs))
 - Remove error when actor doesn't have a registered message type for better usability of custom messages in workers ([#7589](https://github.com/maplibre/maplibre-gl-js/issues/7589)) (by [@HarelM](https://github.com/HarelM))
 - Fix Large number of feature state keys leads to zoom lag when loading cached tiles ([#6633](https://github.com/maplibre/maplibre-gl-js/pull/7590)) (by [@xavierjs](https://github.com/xavierjs))
+- Fix Benchmark stall issue ([#7606](https://github.com/maplibre/maplibre-gl-js/pull//7606)) (by [@xavierjs](https://github.com/xavierjs))
```
HarelM: No need for a changlog entey for this I believe.

HarelM:
Looks good I guess, I think the benchmarks should use an external benchmark library, but that's outside the scope of this PR.

xavierjs:
IMHO They are pretty good, we just need to add some quickfixes for now to make them more usable.

HarelM (approved):

PR #7612, opened 2026-05-09
Fix benchmark display names when compare refs share a commit

### LINKED ISSUE #4762: Problems running bench, since 4.7.1 was released
When running the bench, it only shows `4.7.1` and `main` and not the current branch, but in fact what really happens is that it correct loads `main` (displayed as `4.7.1`) and the local branch (displayed as `main`), and fails to load the latest tag `4.7.1`.

Screenshot 2024-09-25 at 12 18 16

Changing this line:
https://github.com/maplibre/maplibre-gl-js/blob/3160299a30ee54ea836d5cb371578ab25d251aa0/test/bench/versions/index.html#L24

To use 4.7.0 instead:
`window.versionsDisplayName = ['v4.7.0', 'main'];`

Resolves the problem:
Screenshot 2024-09-25 at 12 19 30

shreeharshshinde:
## Description

This fixes an issue where the benchmark page can show incorrect labels when two compare refs resolve to the same generated benchmark version key.

Previously, the page stored requested labels in `window.versionsDisplayName` and later matched them to loaded benchmark versions by array position. If two refs, for example `v4.7.1` and `main`, point to the same generated benchmark key, loading the second script does not add a new benchmark version. However, its label still remains in the display-name array, causing later labels to shift. As a result, the local branch can be displayed with the wrong label.

To fix this, the benchmark loader now checks which benchmark version keys exist before and after each compare script is loaded. It only maps a requested display label to version keys that were newly added by that script. If no display label is mapped, it falls back to the actual benchmark version key.

## Screenshot

Before, duplicate refs could shift labels and show the local branch as `main`.

After this change, labels stay aligned with the loaded benchmark versions:

image

This keeps labels aligned with the benchmark data that was actually loaded, including cases where multiple refs point to the same commit.

Fixes #4762

### DIFF
--- a/test/bench/versions/index.html
+++ b/test/bench/versions/index.html
@@ -41,20 +41,28 @@
         async function runComparison() {
             try {
                 const params = new URL(location).searchParams;
-                window.versionsDisplayName = params.getAll('compare').filter(Boolean);
+                const requestedVersionsDisplayName = params.getAll('compare').filter(Boolean);
                 if (!params.has('compare')) {
-                    const response = await fetch('https://registry.npmjs.org/maplibre-gl/latest', {});
+                    const response = await fetch('https://api.github.com/repos/maplibre/maplibre-gl-js/releases/latest');
                     const responseJson = await response.json();
-                    window.versionsDisplayName = [`v${responseJson['version']}`, 'main'];
+                    requestedVersionsDisplayName.push(responseJson['tag_name'], 'main');
                 }
-                console.log(`Comparing versions: ${window.versionsDisplayName}`);
+                console.log(`Comparing versions: ${requestedVersionsDisplayName}`);
 
-                for (const v of window.versionsDisplayName) {
+                const versionDisplayNameByVersion = {};
+                for (const v of requestedVersionsDisplayName) {
+                    const beforeVersions = getBenchmarkVersions();
                     await loadVersion(v);
+                    const afterVersions = getBenchmarkVersions();
+                    const newVersions = afterVersions.filter(version => !beforeVersions.includes(version));
+                    for (const version of newVersions) {
+                        versionDisplayNameByVersion[version] = v;
+                    }
                 }
                 await import('/test/bench/versions/benchmarks_generated.mjs');
 
                 const benchmarks = [];
+                window.versionsDisplayName = [];
                 const benchmarkNameList = Object.keys(window.maplibreglBenchmarks).sort();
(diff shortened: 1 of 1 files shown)


HarelM:
Looks good overall, what tests did you ran to make sure this is working as expected?

shreeharshshinde:
> Looks good overall, what tests did you ran to make sure this is working as expected?

I think to make a test file for that would be good option

HarelM:
No need, this is a testing tool. I was curious to know which tests you ran manually to make sure this is working as expected.

shreeharshshinde:
I manually tested the benchmark comparison flow with:

- Terminal 1: `npm run start-bench`
- Terminal 2: `npm run benchmark -- --compare v4.7.0 --compare v4.6.0 --compare v4.7.1`

The important part I verified was that the benchmark output/header now keeps the labels aligned and shows the current branch separately, e.g. `v4.7.0`, `v4.6.0`, `v4.7.1`, and the local branch label, instead of shifting the local branch to `main`.

I also ran:

- `npm run lint -- test/bench/versions/index.html`
- `git diff --check`
- `npm run build-benchmarks`

HarelM:
Cool! Thanks for the update, I'll look into fixing the docs build to due recent zensical changes, but that's not relevant to this PR.

shreeharshshinde:
Sure !! Happy to contribute

HarelM (approved):

HarelM:
Can you redo the fix? There was another PR which changed this file a bit...

shreeharshshinde:
I hope failing render tests are not related to my changes

test/bench/versions/index.html
```diff
@@ -41,20 +41,28 @@
         async function runComparison() {
             try {
                 const params = new URL(location).searchParams;
-                window.versionsDisplayName = params.getAll('compare').filter(Boolean);
+                const requestedVersionsDisplayName = params.getAll('compare').filter(Boolean);
                 if (!params.has('compare')) {
-                    const response = await fetch('https://registry.npmjs.org/maplibre-gl/latest', {});
+                    const response = await fetch('https://api.github.com/repos/maplibre/maplibre-gl-js/releases/latest');
```
HarelM: This looks like a bad merge... It should use npm as it was fixed in a recent PR

shreeharshshinde: Updated @HarelM

HarelM (approved):

PR #7614, opened 2026-05-11
Debounce `setImages` broadcast to once per frame

bradymadden97:
## Background

Each call to `addImage`/`removeImage` triggers `dispatcher.broadcast(setImages)` with the full available image name list. With a large number of images (I noticed this with 10,000 images), we would often hit strange OOMs when many of the images were loaded in quick succession. 

Since there is no plural `addImage(s)` endpoint, there is no way to avoid having every image you add result in a `setImages` broadcast `postMessage` to the worker. For the Nth image, that means a structuredClone of an array with N image ids in it. This is a total cost of O(n^2), as every image re-clones the entire array of all image ids before it. 

We had image ids with >250 characters, so the incremental Nth message was costing up to 5mb to serialize and send (10k * 250 * 2 bytes per char). Claude calculated a total of 14GB of cumulative string data cloned through postMessage for 10k images with ~280 chars, since each of the 10k images triggered its own ever-increasing `setImages` payload.

Oftentimes, we'd see `setImages` messages queue up in the renderer process message buffer faster than the worker could drain them, which would lead to OOM crashes (DataCloneError / V8 MarkCompactCollector failure).

## Solution

Move the broadcast into style.update(), which runs once per animation frame. The worker only reads availableImages at the start of WorkerTile.parse(), which is triggered by reloadTile messages also sent from style.update() — so the worker always has the current list before any tile parsing begins. 

This matches the existing pattern used by _updateWorkerLayers, which batches layer changes and broadcasts once per frame. But the benefit of this is far fewer messages sent to the worker within a single animation frame.

## Alternatives considered
- Plural `addImages` API: just a higher LOE as that's part of the public API. But a reasonable solution that would allow us to debounce in user-land (as we're already doing for our image loading - just ultimately you have to call `addImage` serially at some point so there's no way currently to avoid the 10k postMessages).
- Update API to send a diff of added/removed images - higher LOE and possibility of unintended regression from state getting out of sync.

### Automated
Ran the entire test suite

### Manual
Had a map with 10k points, each with an different icon image and a large number (100) of feature properties. I loaded the icon images incrementally, and called `addImage` after the image loaded, as well as updated the associated geojson feature with the icon image url. Previously this map would typically OOM. After this change, the map would not OOM.

Importantly, the combination of large _feature properties_ AND large number of images was important to the repro here. The large number of feature properties meant operations like `loadData` would take a little while to deserialize the large feature property objects and saturate the worker. During this time, the many calls to `setImages` would fill up the v8 renderer process memory and eventually OOM, because the worker wasn't able to process them yet.

Running the repro here https://github.com/bradymadden97/maplibre-tools/tree/main/repros/addimage-oom before vs. after was very clear:

```
# before
> pnpm dev:repro addimage-oom
```

Crashed:

https://github.com/user-attachments/assets/5f7fbf06-d974-48bf-8667-64bb0a03df32

```
# after
> pnpm dev:repro:local addimage-oom
```

Did not crash and loaded all 10k, despite only making 40 `setImages` calls

https://github.com/user-attachments/assets/1d24bedf-9772-4ebe-82a8-2208a2f12248

One final example - if I tone down the number of properties on each feature from 100 to 10 we can get by without a crash, but you will see 10,000 `setImages` calls, and a higher `postMessage` cumulative clone time than you saw in the video above:

You can see the final cumulative postMessage times are almost 50x higher than they were in the video above.

https://github.com/user-attachments/assets/918951c8-e79f-4c1f-a256-a13edb387fa5

## Launch Checklist

 - [x] Post benchmark scores.


HarelM:
Out of curiosity, why are you updating 10k images?

src/style/style.ts
```diff
@@ -733,6 +732,14 @@ export class Style extends Evented {
 
         const changed = this._changed;
         if (changed) {
+            // Broadcast available images once per frame, debounced from
+            // _afterImageUpdated. This must happen before any other worker
+            // messages (updateLayers, reloadSource, reloadTile) so the worker
+            // has the current list before any tile parsing begins.
+            if (Object.keys(this._changedImages).length) {
+                this.dispatcher.broadcast(MessageType.setImages, this._availableImages);
```
HarelM, line 741: Looks reasonable overall, are you sure that there's no way for this to be "missed" in an initialization or something like that?

bradymadden97:
> Out of curiosity, why are you updating 10k images?

Valid question. This happens even when just loading and adding 10k images for the first time. They don't even need to be updated after they've been added to the map. 

We discovered this as part of a scale test we were running. But our product does support a high level of configurability for icon images, which can result in a very large number of unique images being loaded onto a map.

HarelM:
Any plans to take it out of draft?

src/style/style.ts
bradymadden97: The three paths where the broadcast is removed (`_afterImageUpdated`, `_unloadSprite`, `removeSprite`) all set both `_changedImages[id] = true` and `_changed = true`, so the new check in update() (`Object.keys(this._changedImages).length`) will always pick them up.

The initial sprite load path (`_loadSprite`) is the one case that doesn't set `_changedImages` (when `isUpdate === false`), but the broadcast there (line 585) is intentionally left untouched — it still fires unconditionally in the .finally() block.

`updateImage()` also doesn't go through this path, but that's pre-existing — it never broadcast `setImages` since it doesn't change the list of available images, only the image content.

bradymadden97:
Yep was just trying to find some time to respond to your question. Undrafting now.

bradymadden97:
Curious if there are any other objections here or if we can move forward with this PR?

HarelM:
Checkout the checklist - this needs some tests and a changelog entry.
This looks like a good improvement to centralize image updates. thanks!

bradymadden97:
Updated with test + changelog entry.

HarelM (approved):

HarelM:
The only last concern I have is with `styleimagemissing` event. There's a bug that was recently reported about it:
- #7752

I don't think this bug is related to this fix, but it did got me to think that this change must not regress the `stylemissingimage` handling. I don't think it does, but to be on the safe side please let me know that you've looked into it and that it's safe to merge this.

THANKS!

bradymadden97:
I think that would already be covered by this style of test pretty well, yes? https://github.com/maplibre/maplibre-gl-js/blob/ae782be467a44bafa89a5f67c67451e5b422c981/src/ui/map_tests/map_images.test.ts#L30

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## main
 ### ✨ Features and improvements
-- _...Add new stuff here..._
+- _Add new stuff here..._
```
bradymadden97:
```suggestion
- _...Add new stuff here..._
```

HarelM:
Can the comments be toned down a bit please? 
Also I think there's a test for style image missing as you said, no need for a new file just for that.

bradymadden97:
Updated for latest comments

PR #7635, opened 2026-05-13
Optimization: Cache stable translucent layers to avoid redundant re-rendering 

xavierjs:
Hi,

## Problem

In MapLibre GL JS, every translucent layer is repainted every frame, even when most layers haven't changed. This is wasteful in scenarios with many static base layers and only one or few frequently updating layers on top (e.g., an animated data overlay on a basemap).  

## Solution

After rendering the translucent pass normally, the screen content is snapshotted into a cached texture once the bottom N layers have been stable for a configurable number of frames. On subsequent frames, the cached texture is blitted to screen in a single draw call, and only the remaining (changing) layers are rendered on top.

## Implementation notes

* Screen-space snapshot, not a separate FBO: layers are rendered to the main framebuffer as usual, then the result is copied to a texture. This guarantees pixel-identical blending with the opaque pass — no double-premultiplication artifacts.
* Per-layer stability tracking: each StyleLayer has an `_unchangedFrameCount` counter, incremented every frame and reset to 0 on style changes, paint property changes, source reloads, transitions, or tile data arriving from workers.
* Cache invalidation: the cache is invalidated on camera movement (pan/zoom/rotate), canvas resize, or when any cached layer's content changes.
* Configurable: `translucentCacheStabilityThreshold` (default 3 frames) and `translucentCacheMinLayers` (default `5`) control when caching kicks in. The entire optimization can be disabled via `optimizationFlagCacheFirstLayersRendering: false` at map creation.

## Benchmarks

This improvement is especially important for layer rendering benchmarks.
Resultas are given for a Macbook Pro M4 Max, Google Chrome browser:

| Benchmark | main duration (ms) | with this PR duration (ms) | Speedup |
|---|---|---|---|
| LayerCircle | 0.821 | 0.311 | 2.64X |
| LayerColorRelief256Colors | 0.508 | 0.00407 | 124.82X |
| LayerColorRelief2Colors | 0.311 | 0.00330 | 94.24X |
| LayerFill | 1.77 | 1.03 | 1.72X |
| LayerFillExtrusion | 2.17 | 0.313 | 6.93X |
| LayerHeatmap | 2.44 | 1.18 | 2.07X |
| LayerHillshader | 3.85 | 0.326 | 11.81X |
| LayerLine | 1.07 | 0.335 | 3.19X |
| LayerRaster | 3.02 | 0.313 | 9.65X |
| LayerSymbol | 2.83 | 0.332 | 8.52X |
| LayerSymbolWithIcons | 2.91 | 0.332 | 8.77X |
| LayerSymbolWithSortKey | 0.448 | 0.370 | 1.21X |
| LayerTextWithVariableAnchor | 9.72 | 0.330 | 29.45X |

## Misc

No related issues.

## Launch Checklist
 - [X] Link to related issues.
 - [X] Post benchmark scores.

Assisted-By: Claude-Code


HarelM:
Recent changes related to caching caused OOM on a mobile device I checked. How are you planning to address this concern?

xavierjs:
> Recent changes related to caching caused OOM on a mobile device I checked. How are you planning to address this concern?

Is it related to any submitted issue? I would be happy to take a look.
This PR does not involve heavy caching. The increase of memory due to this PR only consists in a new shader program and a texture at the viewport resolution. It is more reusing a previous rendering than real caching. There is no per-tile or per-layer cached object.

HarelM:
The following is the relevant PR I was referring to:
- https://github.com/maplibre/maplibre-gl-js/pull/7549

It's worth mentioning that I haven't tried activating terrain recently on a mobile device, so it might not be just this PR.
But version 6 on my Samsung s24 with mapeak app crashes - a testing version since I disabled terrain in mapeak due to problematic performance a while back.

xavierjs:
> The following is the relevant PR I was referring to:
> 
> * [Improve 3d terrain performance by redesigning RTT cache to increase hitrate #7549](https://github.com/maplibre/maplibre-gl-js/pull/7549)
> 
> It's worth mentioning that I haven't tried activating terrain recently on a mobile device, so it might not be just this PR. But version 6 on my Samsung s24 with mapeak app crashes - a testing version since I disabled terrain in mapeak due to problematic performance a while back.

I have noticed some small issues with this PR that could explain your issue. I have submitted another PR for this: https://github.com/maplibre/maplibre-gl-js/pull/7637

This PR uses cache at a viewport level, so the cache size is very controlled (just a texture having the same resolution as the viewport). It is more a question of prerendering & reuse than real caching.

src/ui/map.ts
```diff
@@ -430,6 +430,13 @@ export type MapOptions = {
      *   to bearing change regardless of cursor position.
      */
     aroundCenter?: boolean;
+    /**
+     * If `true`, the first N stable translucent layers are cached into a texture
+     * and reused across frames, avoiding redundant re-rendering when only upper
+     * layers change and the camera is static.
+     * @defaultValue true
+     */
+    optimizationFlagCacheFirstLayersRendering?: boolean;
```
HarelM: This probably needs a better name. No need to include the word "flag* in the name for example.
It's also not clear what the trade-offs are. When should I set to true and when not?

src/tile/tile.ts
```diff
@@ -287,6 +287,17 @@ export class Tile {
         this.collisionBoxArray = data.collisionBoxArray;
         this.buckets = deserializeBucket(data.buckets, painter?.style);
 
+        // Reset unchanged frame counters for layers whose buckets just loaded,
+        // so the translucent layer cache knows this layer's content has changed.
+        if (painter?.style) {
```
HarelM: This will cause a very circular dependency, please see if there's a different place this can be done...

HarelM:
I haven't completed the review, I'll continue later. 
Since this doesn't "help" the case of moving the map, what problem does it solve? Battery usage on mobile devices?

xavierjs:
> I haven't completed the review, I'll continue later. Since this doesn't "help" the case of moving the map, what problem does it solve? Battery usage on mobile devices?

In many cases we have a dynamic layer atop a stack of static layers. I can be:
* a map displaying a Lyft car arriving (they use Maplibre when you share your ride with somebody). The car is moving but the background layer does not change
* a map displaying another kind of animation done with a dynamic geojson or a symbol with changing properties
* a map in editing mode

In these cases, we will update the top layer more often than the background layers, even if the user does not interact with the map (fixed camera). And without this cache optimization we rerender all the layer stack at each small change of the top layer.

src/ui/map.ts
xavierjs: > This probably needs a better name. No need to include the word "flag* in the name for example. It's also not clear what the trade-offs are. When should I set to true and when not?

With this optimization we allocate a texture that has the same resolution as the viewport. It is not a big deal though. If we have an animated or dynamic layer displayed atop the other layers (an application to annotate a map where we change a geojson layer, if we display a Lyft/Uber driver arriving, if we show symbols depending on a geographical filter like AirBnB), then this optimization will bring a lot. If the map is only displayed for dynamic interaction then not that much. What about `optimizeForDynamicContent`name? It is more usage-oriented and this flag could trigger other kind of optimizations in the future.

src/tile/tile.ts
xavierjs: I do this in the source with the latest commit.

src/render/painter.ts
```diff
@@ -142,6 +143,16 @@ export class Painter {
     // every time the camera-matrix changes the terrain-facilitators will be redrawn.
     terrainFacilitator: {depthDirty: boolean; coordsDirty: boolean; matrix: mat4; renderTime: number};
 
+    // Translucent layer cache: snapshots the screen after rendering the first
+    // N stable translucent layers, reuses that snapshot on subsequent frames.
+    _translucentCacheTexture: Texture | null;
```
HarelM: I would recommend moving all this code to a separate class and handle this kind of caching there instead of adding all those fields.

src/source/geojson_source.ts
```diff
@@ -597,6 +597,18 @@ export class GeoJSONSource extends Evented implements Source {
 
         if (!tile.aborted) {
             tile.loadVectorData(data, this.map.painter, message ===  MessageType.reloadTile);
+
+            // Reset unchanged frame counters for layers whose buckets just loaded,
+            // so the translucent layer cache knows this layer's content has changed.
+            const style = this.map.painter?.style;
```
HarelM: This creates some circular dependencies I'd like to avoid id possible, it also sets an "internal" layer field.
Please see if there's a better way to achieve this somehow, maybe using some events.

src/ui/map.ts
HarelM: It's better, but I'm not sure I'd know what `dynamicContent` would mean in regards to this flag. Maybe `optimizeStaticMaps`? I'm not sure it's better to be honest...

src/source/vector_tile_source.ts
```diff
@@ -288,6 +288,18 @@ export class VectorTileSource extends Evented implements Source {
 
         tile.loadVectorData(data, this.map.painter);
 
+        // Reset unchanged frame counters for layers whose buckets just loaded,
+        // so the translucent layer cache knows this layer's content has changed.
+        const style = this.map.painter?.style;
```
HarelM: Same as previous comment, also note that this is repeating the same code, so a better home for this would be better I guess...

HarelM:
Cache is always tricky to make sure we invalidate it correctly.
Can you link to a site where this can be observed? I understand the use case, but I feel like it might be a super edge case. I think that in most cases where you show a map you also allow the user to interact with it to some degree, so this would not be used.
The solution is clever, don't get me wrong, but it's not trivial or simple.
If there's a way to contain it as much as possible in a few files it would be better I think.
The fact that this is opt-in is also a good approach, as it will probably won't affect most users if they don't really need it.

src/ui/map.ts
xavierjs: I would go for `optimizeStaticBase`

src/render/painter.ts
xavierjs: It is done in the latest commit. I also renamed the shader program

src/source/vector_tile_source.ts
xavierjs: It is cleaner in the latest version, data change is caught directly in the on data event in `style.ts`.

src/source/geojson_source.ts
xavierjs: It is cleaner in the current version, source change is caught directly in the on data event in `style.ts`.

xavierjs:
To showcase the utility of this PR I have added in `src/shaders/glsl/fullscreen_texture.fragment.glsl`: `fragColor *= vec4(1., 0.5, 0.5, 1.);`.
Then I run `npm run generate_shaders`. Now, when the texture cache is used for rendering instead of rerendering all base layers, it will display it in reddish tones.

To make more sure that this PR has no bad side effect, I checked manually all the examples (and found some bugs not caught by the tests - fixed). Here are the examples where this PR is particularly effective:

* animate-a-line.html
* animate-a-point.html
* animate-a-series-of-images.html
* add-an-animated-icon-to-the-map.html 
* animate-a-point-along-a-route.html
* update-geojson-polygons.html
* measure-distances.html

You can see that we render the base layers using our cache, and if you run the performance benchmark in the web console you will see that the CPU usage is really lower.
It does not slow down the rendering in interactive mode since we wait for the rendering to be static for a few frames before updating and using the cache.

Interactive mode is great and not bothered (nor improved) by this PR. But many interactive maps have animated feature (a Uber car arriving, the current GPS position during a hike, an important stuff blinking...) . So without this PR we are rerendering everything all the time, even if the user does not interact with the map most of the time (I think in most cases he will interact 1/5 of the time or something like that).

Here is a small screenshot video from the example `animate-a-point-along-a-route.html` where the output color of the cached base layers is displayed in reddish tones:

https://github.com/user-attachments/assets/fd0ed26c-4696-4ffd-b75c-d68b692e6725

When the planes move, for each frame we render once the cached base layers instead of rerendering all layers.
When I panned a bit the map, it invalidated the cache, then the cache was used again when the panning was released.

HarelM:
@xavierjs this is still in draft, if you wish to push this forward please move this out of draft.

HarelM:
Overall the code looks good, but I can't help but wonder, given the amount of cache invalidation in `map.ts` file, there's might a lot of code path that have a missing cache invalidation and I'm concerned this is a honest's nest... 
Moreover, these might be hard to fix or reproduce bugs due to how the cache works.
So I'm mainly looking for a way to de-riskify this PR.
I think it shows good performance improvements, but I'm looking for a way to feel comfortable merging it.

src/render/painter.ts
```diff
@@ -181,6 +187,9 @@ export class Painter {
         this.pixelRatio = pixelRatio;
         this.context.viewport.set([0, 0, this.width, this.height]);
 
+        // Invalidate translucent layer cache on resize
```
HarelM: Comment is not needed, we are inside the resize method...

src/render/painter.ts
```diff
@@ -623,6 +644,11 @@ export class Painter {
 
             this._renderTileClippingMasks(layer, coordsAscending[layer.source], !!this.renderToTexture);
             this.renderLayer(this, tileManager, layer, coords, renderOptions);
+
+            // After rendering the last stable layer, snapshot the screen
+            if (needsSnapshot && this.currentLayer === stableLayerCount - 1) {
+                this.staticBaseCache.snapshot(this, stableLayerCount);
```
HarelM: Is there a better way besides passing `this`? Passing `this` is usually a sign of circular dependencies...

src/render/static_base_cache.ts
```diff
@@ -0,0 +1,221 @@
...
+     */
+    prepareTranslucentPass(
+        painter: Painter,
+        layerIds: string[],
+        layers: {[_: string]: StyleLayer},
+        zoom: number,
+        options: PainterOptions,
+        imageManager: ImageManager,
+        tileManagers: {[_: string]: TileManager},
+    ): {cacheStartLayer: number; needsSnapshot: boolean; stableLayerCount: number} {
```
HarelM: Probably worth defining a type for the return value.

src/render/static_base_cache.ts
```diff
@@ -0,0 +1,221 @@
...
+    /**
+     * Determine how the translucent pass should be rendered: from cache,
+     * with a snapshot to build the cache, or normally.
+     *
+     * @returns An object with:
+     *   - `cacheStartLayer`: index at which the main rendering loop should start
+     *   - `needsSnapshot`: whether to call {@link snapshot} after rendering layer `stableLayerCount - 1`
+     *   - `stableLayerCount`: number of consecutive stable layers from the bottom
+     */
+    prepareTranslucentPass(
```
HarelM: This method is too long. Try splitting it to more smaller parts.

src/render/static_base_cache.ts
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        // The snapshot captures the opaque pass content for ALL layers.
+        // If any layer has an active paint transition (e.g. from setPaintProperty),
+        // the opaque pass is mid-transition and the snapshot would be stale.
+        // Don't build or reuse a cache while transitions are active.
+        // Uses hasActiveTransition(now) instead of hasTransition() to ignore
+        // expired transitions whose prior reference hasn't been cleared yet.
+        const currentTime = now();
+        let hasActiveTransition = false;
```
HarelM: nit: this might be converted to `map(...).some(...)`?

src/render/static_base_cache.ts
```diff
@@ -0,0 +1,221 @@
...
+            painter.viewportSegments,
+        );
+    }
+
+    invalidate(): void {
+        this._cachedLayerCount = 0;
+    }
+
+    destroy(): void {
+        if (this._texture) {
```
HarelM: Can we remove the if here?
```js
this._texture?.destroy();
this._texture = null;
```

src/render/static_base_cache.ts
xavierjs: Sure, I will name it `TranslucentPassPlan`

lucaswoj:
Is there any way we can re-use the RTT system that powers terrain rendering to power this cache layer as well? There may be opportunities to share a lot of code between these two systems & a tile-based caching layer may have better performance characteristics than a viewport-based one.

HarelM:
@lucaswoj is making a good point, I was thinking about it but forgot to ask about it.

xavierjs:
This is an interesting idea. However, I have some concerns about a tile based approach:                                                                                                                       
                                                                                                                                                                                                                                            
 1. **Cache overhead when inactive**. With a screen space cache, the worst case is a single `copyTexSubImage2D` call and when the cache cannot be used (camera moving, layers transition, ...), the overhead is neglictible. A tile-based cache requires textures to FBO binds, FBO clear, and render pass per visible tile. With 20–50 tiles on screen, that is a significant overhead even when every tile needs re-rendering and the cache provides no benefit. 
                                                                                                                                                                                                                                
 2. **Cache overhead when active**. When the cache IS used, a screen-space blit is a single fullscreen quad draw call. A tile-based cache requires compositing N tile textures individually, each with its own draw call and texture bind. The
 screen-space approach is strictly cheaper on the reuse path. For GPU memory it should also be cheaper.
                                                                                                                                                                                                                                                  
 3. **Visual fidelity with perspective**. A tile-based cache renders layers into textures in tile coordinates (orthographic). When these textures are projected to screen at high pitch angles, texture filtering will likely degrade the result, particularly for sharp vector geometry like fill edges, line strokes, and SDF text/icons. Even with anisotropic filtering, I doubt that the quality will match direct vector rasterization. The screen-space snapshot avoids this entirely since it captures the exact screen pixels.
                                                                                 
 A tile-based approach would have the advantage of surviving camera pans, which the screen-space cache does not. But given the overhead, complexity, and fidelity trade-offs, I believe that a screen-space approach is a good choice for this optimization. It's simple ( about 200 lines tests excluded), zero-cost when inactive, and pixel accurate.

lucaswoj:
Thanks for laying these points out clearly. I'm a lowly bystander on this PR so nothing blocking here, but I do want to share some follow up thoughts... 

1. **Overhead when inactive.** The PR already short-circuits caching during camera motion, symbol fading, transitions, and unstable layers (`StaticBaseCache.prepareTranslucentPass`). A shared RTT path could gate itself the same way and pay nothing in those cases.

2. **Overhead when active.** Drawing N quads is more work than drawing one, agreed, but the layer re-renders saved should swamp that difference on the reuse path. I'd guess a tile cache wins for many real scenes, especially since the RTT pipeline has already been tested in production and undergone some careful performance tuning. 

3. **Fidelity with perspective.** Hard to say without seeing it. Pitched cameras without 3D terrain are probably uncommon in GL JS today, so any quality hit may not affect many users.

My biggest concern is the cache invalidation system that @HarelM flagged. Two caching systems means two places to keep correct as the renderer evolves. Funneling both through one path lets fixes and perf work benefit both features at once.

src/render/image_manager.ts
```diff
@@ -328,8 +328,27 @@ export class ImageManager extends Evented {
         this.dirty = true;
     }
 
+    /**
+     * Whether any dynamic image (with a render callback) was updated this frame.
+     * Used by the static base cache to invalidate layers that depend on dynamic images.
+     */
+    dynamicImageUpdatedThisFrame: boolean;
```
HarelM: Please move this to the upper part of the class.
[Optinal] Make it private and allow setting it from outside if needed?

src/render/painter.ts
```diff
@@ -601,7 +611,17 @@ export class Painter {
 
         let globeDepthRendered = false;
 
-        for (this.currentLayer = 0; this.currentLayer < layerIds.length; this.currentLayer++) {
+        // Disable static base cache when render-to-texture is active (terrain/globe) —
```
HarelM, line 614: While not your fault, maybe it's worth moving each pass to its own method instead of these `=======` comment?
This method is already too long...

src/render/static_base_cache.ts
```diff
@@ -0,0 +1,235 @@
...
+    /** Number of consecutive stable layers from the bottom. */
+    stableLayerCount: number;
+};
+
+/**
+ * Caches the screen content after rendering the first N stable translucent
+ * layers, and reuses that snapshot on subsequent frames to avoid redundant
+ * re-rendering when only upper layers change and the camera is static.
+ */
+export class StaticBaseCache {
```
HarelM, line 14: This class is not really a cache, it has a cache in it, but it does a lot of things in order to decide what to cache and also does the draw call.
I'm not sure it makes sense, but maybe this needs to be split a bit in order not to be too general of an object.
The caching part is usually the smallest part of the code, the cache managing and logic is where most of the code is, and it's usually not part of the cache.

I would recommend trying to get a bird view on this and decide on each method of this class where it should be according to the current code architecture.

A class diagram might be helpful here, maybe, to help with this.

HarelM:
I've added a few more comments, the major one is related to the design of this new class.
I'm also concerned about mobile performance. While the RTT cache looks like a good approach I'm still unsure how good it works on mobile devices. The latest version of maplibre doesn't crash on terrain, but it's still heavy and looks like dropping FPS significantly.
Bottom line, there's more work needed here, sorry to drag this...

xavierjs:
> Thanks for laying these points out clearly. I'm a lowly bystander on this PR so nothing blocking here, but I do want to share some follow up thoughts...
> 
> 1. **Overhead when inactive.** The PR already short-circuits caching during camera motion, symbol fading, transitions, and unstable layers (`StaticBaseCache.prepareTranslucentPass`). A shared RTT path could gate itself the same way and pay nothing in those cases.
> 2. **Overhead when active.** Drawing N quads is more work than drawing one, agreed, but the layer re-renders saved should swamp that difference on the reuse path. I'd guess a tile cache wins for many real scenes, especially since the RTT pipeline has already been tested in production and undergone some careful performance tuning.
> 3. **Fidelity with perspective.** Hard to say without seeing it. Pitched cameras without 3D terrain are probably uncommon in GL JS today, so any quality hit may not affect many users.
> 
> My biggest concern is the cache invalidation system that @HarelM flagged. Two caching systems means two places to keep correct as the renderer evolves. Funneling both through one path lets fixes and perf work benefit both features at once.

  You are very welcome to discuss this PR, your points are valid and really deserve the debate. It is a complicated issue since there are pros and cons for each solution and at the end, it depends on the real use cases.                         
                                                                                                                                                                                                                                                    
  1. **Overhead when inactive**. You are right if the cache is completely inactive (not computed). However, if it is computed but not used or not used a lot (we repeatedly have a few stable frames → cache update, then a cache invalidation), for screen-based RTT the overhead is minimal (memory: one screen-sized texture, just 1 RTT) whereas it is significant for tile-based (need to maintain a texture pool and do 20 or 50 RTTs each time the cache is rebuilt).                                                
 
 2. **Overhead when active.** With a tile-based cache we can expect to save work during user interactions (pan, zoom without layout recomputing, ...), whereas this current PR purposefully excludes user interactions. The use case here is a static camera with dynamic overlays, not general interaction caching. I think it makes the cache invalidation easier since its use is more scoped. Also many maps have hovering effects that will break the cache each time the cursor moves in both approaches. The resource usage of this PR is very predictive so it does not need too many tuning. There are only 2 parameters and I put quite conservative values (we cache the layers only if there are at least 5 stable layers and we consider a layer as stable if it does not change for 3 renderings).
                                                                                                                          
 3. **Fidelity with perspective**. We would need to test. This issue may also happen if the user zooms with no pitch without changing the tile level. I think the texture filtering would degrade the rendering, but I'm not sure how much in practice. For this point I think we need to draft a quick test.
                                                                                                                                                                                                                                                    
  As a separate note, I think this optimization would be especially useful on mobile. In many cases we observe a map updating without the user interacting much. For example:                                                                       
  - While waiting for an Uber car, we see the car arriving. We pan from time to time but most of the time we only watch the car moving on a static base map.                                                                                        
  - On a hiking app, we see the position updating from the GPS while the map stays still.                                                                                                                                                           

On desktop, a common case is a screen shared between a map view and a search form updated in real time (for example if we are looking for a rental house )

xavierjs:
@HarelM Do you think it would be helpful to add a debug flag to the map named `showStaticBaseCache` as we currently have `showOverdrawInspector`? It would display the static cache in reddish tones (same change as https://github.com/maplibre/maplibre-gl-js/pull/7635#issuecomment-4490636412 ) to make cache cache utility easier to evaluate and cache coherency issues easier to debug.

HarelM:
Maybe we should add it as experimental flag like I did for the worker tile slicing? This way if we decide to change its name or remove it, we can do it without a breaking change version?

src/render/painter.test.ts
```diff
@@ -83,6 +83,269 @@ describe('tile texture pool', () => {
...
+        painter = new Painter(gl, transform);
+        painter.width = 512;
+        painter.height = 512;
+    });
+
+    afterEach(() => {
+        painter.destroy();
+    });
+
+    test('captureCache creates texture and copies framebuffer', () => {
```
HarelM: This test is asserting on "private" members of the cache. If this needs to move to the cache tests file let's do that instead.

src/render/painter.test.ts
HarelM, line 1: I've made a comment about one test, but I think it applies to all the newly added tests in this file.

src/ui/map.ts
```diff
@@ -3297,6 +3319,7 @@ export class Map extends Camera {
      */
     removeFeatureState(target: FeatureIdentifier, key?: string): this {
         this.style.removeFeatureState(target, key);
+        this.painter?.staticBaseCache.invalidate();
```
HarelM: Should this be inside removeFeatureState maybe? and avoid the call from map straight to painter?
The pattern here of calling the style and then calling the painter for every method here seems like a smell...

src/ui/map.ts
```diff
@@ -526,6 +533,7 @@ const defaultOptions: Readonly<Partial<MapOptions>> = {
     terrainSkirtLength: 'auto',
     zoomLevelsToOverscale: 4,
     anisotropicFilterPitch: defaultAnisotropicFilterPitch,
+    optimizeStaticBase: true,
```
HarelM: Should be false by default until we gain confidence I think.

src/style/style_layer.ts
```diff
@@ -113,6 +113,8 @@ export abstract class StyleLayer extends Evented {
 
     _visibilityExpression: VisibilityExpression;
 
+    _unchangedFrameCount: number = 0;
```
HarelM: Where is this counter updated? Seems weird that there's no update of it in this class.
If this is the case maybe it shouldn't be here or shouldn't have "_"...?

src/ui/map.ts
```diff
@@ -3582,12 +3593,16 @@ export class Map extends Camera {
      *
      * @param updateStyle - mark the map's style for reprocessing as
      * well as its sources
+     * @param invalidateStaticBaseCache - force the static base cache to be invalidated
      */
-    _update(updateStyle?: boolean): this {
+    _update(updateStyle?: boolean, invalidateStaticBaseCache?: boolean): this {
```
HarelM: Can we make these parameters non optional so that it's clear with method is using it and how?

src/ui/map.ts
```diff
@@ -3491,6 +3501,7 @@ export class Map extends Camera {
         }
 
         this.painter = new Painter(gl, this.transform);
+        this.painter.staticBaseCache.enabled = this._optimizeStaticBase;
```
HarelM: Can we move this line to be inside the constructor of painter?

src/ui/map.ts
```diff
@@ -429,6 +429,13 @@ export type MapOptions = {
      *   to bearing change regardless of cursor position.
      */
     aroundCenter?: boolean;
+    /**
+     * If `true`, the first N stable translucent layers are cached into a texture
+     * and reused across frames, avoiding redundant re-rendering when only upper
+     * layers change and the camera is static.
+     * @defaultValue false
+     */
+    optimizeStaticBase?: boolean;
```
HarelM: `cacheStaticTranlucentLayersOptimization` instead?
"static base" remind me of inheritance and OOP too much...

src/ui/map.ts
```diff
@@ -3582,12 +3592,16 @@ export class Map extends Camera {
      *
      * @param updateStyle - mark the map's style for reprocessing as
      * well as its sources
+     * @param invalidateStaticBaseCache - force the static base cache to be invalidated
      */
-    _update(updateStyle?: boolean): this {
+    _update(updateStyle: boolean, invalidateStaticBaseCache: boolean): this {
```
HarelM, line 3597: Use the same parameter name as the one passed (from the class field)?

src/ui/map.ts
```diff
@@ -3582,12 +3592,16 @@ export class Map extends Camera {
...
         if (!this.style?._loaded) return this;
 
         this._styleDirty ||= updateStyle;
         this._sourcesDirty = true;
+        if (invalidateStaticBaseCache) {
+            this.painter?.staticBaseCache?.invalidate();
```
HarelM, line 3603: Is there a way to move this further inside into the render method since we use painter object there mainly? I'm not sure it's better to cascade this inside, but in this context we mainly check style related things and less "painter related things".
I don't know, something feels a bit off here...

src/render/static_base_cache_manager.ts
```diff
@@ -0,0 +1,171 @@
...
+    /**
+     * Check whether a visible layer is cacheable.
+     * Returns false if the layer is custom, recently changed, has a source
+     * with active transitions or incomplete tiles, or uses dynamic images.
+     */
+    _isLayerCacheable(layer: StyleLayer, imageManager: ImageManager, tileManagers: {[_: string]: TileManager}): boolean {
+        if (layer.type === 'custom' || layer.unchangedFrameCount < this.stabilityThreshold) {
+            return false;
+        }
+        if (layer.source && tileManagers[layer.source]) {
```
HarelM, line 43: This if can be negated and allow reducing complexity and indentation of this method.

src/render/static_base_cache_manager.ts
```diff
@@ -0,0 +1,171 @@
...
+                stableLayerCount++;
+                continue;
+            }
+            if (!this._isLayerCacheable(layer, imageManager, tileManagers)) {
+                break;
+            }
+            stableLayerCount++;
+        }
+
+        // Invalidate cache during camera movement.
```
HarelM, line 118: Can you reduce the comments throughout this PR?
Use small methods with tsdocs comment instead if needed.

src/render/static_base_cache_manager.ts
```diff
@@ -0,0 +1,171 @@
...
+        // Note: dynamic image updates don't invalidate the cache because the
+        // per-layer check above already excludes layers with dynamic images
+        // from the stable set — the cached layers are guaranteed static.
+        const cameraMoving = options.moving || options.zooming || options.rotating;
```
HarelM, line 122: Is `pitching` missing here?

src/render/static_base_cache_manager.ts
```diff
@@ -0,0 +1,171 @@
...
+        if (cameraMoving || painter.symbolsAreFading) {
+            cache._cachedLayerCount = 0;
```
HarelM, line 124: This looks like a cache internal field that is being modified here, maybe a setter method is needed?

src/render/static_base_cache_manager.ts
```diff
@@ -0,0 +1,171 @@
...
+        }
+
+        // The cache captures the opaque pass content for ALL layers.
+        // If any layer has an active paint transition (e.g. from setPaintProperty),
+        // the opaque pass is mid-transition and the cache would be stale.
+        // Don't build or reuse a cache while transitions are active.
+        const hasActiveTransition = this._hasActiveTransitions(layerIds, layers, zoom);
+
+        // Check if existing cache is still valid
+        const cacheValid = cache._cachedLayerCount > 0 &&
```
HarelM, line 134: This should probably be a method on the cache object...?

src/render/static_base_cache.ts
```diff
@@ -0,0 +1,85 @@
...
+
+        this._cachedLayerCount = layerCount;
+    }
+
+    /**
+     * Draw the cached texture as a fullscreen quad.
+     * Uses unblended (replace) mode since the cache includes the full
+     * screen content (opaque pass + cached translucent layers).
+     */
+    blitCacheToScreen(painter: Painter): void {
```
HarelM, line 51: can this be named drawCacheToScreen? I'm not sure I know what blit is...

src/render/static_base_cache_manager.ts
HarelM: Probably calling `invalidate` is the equivalent...

src/render/static_base_cache_manager.ts
```diff
@@ -0,0 +1,171 @@
...
+
+        let cacheStartLayer = 0;
+        let needsCapture = false;
+
+        if (cacheValid) {
+            // Reuse existing cache — blit cached texture then skip to remaining layers
+            cache.blitCacheToScreen(painter);
+            painter.clearStencil();
+            cacheStartLayer = cache._cachedLayerCount;
+        } else if (!cameraMoving && !hasActiveTransition && stableLayerCount >= this.minLayers) {
```
HarelM, line 148: Use `return` instead of "else if"ing, as I'm not sure I understand the conditions here...

src/render/painter.ts
HarelM: Still relevant...

src/render/painter.test.ts
```diff
@@ -83,6 +83,129 @@ describe('tile texture pool', () => {
...
+
+        const useProgramSpy = vi.spyOn(painter, 'useProgram');
+
+        painter.style = createMockStyle(0);
+
+        painter.render(painter.style, defaultRenderOptions);
+
+        // fullscreenTexture program should NOT have been used
+        const calls = useProgramSpy.mock.calls.filter(c => c[0] === 'fullscreenTexture');
+        expect(calls).toHaveLength(0);
```
HarelM, line 176: `not.toHaveBeCalled()`.

src/render/painter.test.ts
```diff
@@ -83,6 +83,129 @@ describe('tile texture pool', () => {
...
+    });
+
+    test('cache is not reused when height changes', () => {
```
HarelM, line 179: This test is similar to width, I think it's redundant.

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 ## main
 ### ✨ Features and improvements
 - ⚠️ Removed the remaining mapbox references in the code and in the tests. This changes the `#pragma mapbox` to `#pragma maplibre` in case you have shader code that relied on it. ([#7761](https://github.com/maplibre/maplibre-gl-js/issues/7761)) (by [@HarelM](https://github.com/HarelM))
-- _...Add new stuff here..._
+- Optimization: Cache stable translucent layers to avoid redundant re-rendering ([#7607](https://github.com/maplibre/maplibre-gl-js/pull/7635)) (by [@xavierjs](https://- - - _...Add new stuff here..._
```
HarelM, line 4: Please fix the comment...

src/render/static_base_cache_manager.ts
```diff
@@ -0,0 +1,171 @@
...
+    needsCapture: boolean;
+    /** Number of consecutive stable layers from the bottom. */
+    stableLayerCount: number;
+};
+
+/**
+ * Manages the static base cache by deciding which layers are cacheable,
+ * when to capture or reuse cached content, and when to invalidate.
+ */
+export class StaticBaseCacheManager {
```
HarelM, line 29: Might be worth aligning the name with the map flag name, not sure, your call.

src/render/static_base_cache.ts
```diff
@@ -0,0 +1,85 @@
...
+import {CullFaceMode} from '../webgl/cull_face_mode.ts';
+import {fullscreenTextureUniformValues} from '../webgl/program/fullscreen_texture_program.ts';
+
+import type {Context} from '../webgl/context.ts';
+import type {Painter} from './painter.ts';
+
+/**
+ * Stores and retrieves a cached framebuffer snapshot as a fullscreen texture.
+ */
+export class StaticBaseCache {
```
HarelM, line 14: Might be worth aligning the name with the map flag name, not sure, your call.

src/render/static_base_cache.ts
```diff
@@ -0,0 +1,85 @@
...
+    _width: number = 0;
+    _height: number = 0;
+
+    /**
+     * Copy the current screen framebuffer into the cache texture.
+     * Call this after rendering the first N stable translucent layers to
+     * the main framebuffer.
+     */
+    captureCache(context: Context, width: number, height: number, layerCount: number): void {
+        const gl = context.gl;
```
HarelM, line 26: I wonder if the gl stuff should be here or not.
Other gl stuff are located mainly in the `draw` folder exposed as methods to allow drawing.
Does it make sense to move this logic there and only cache the texture itself?
This might make this cache class really small, but usually cache logic is fairly small...

HarelM: @birkskyum can you take a look at this from a webgpu perspective and let me know where you think things should be placed?

HarelM:
I've added a few more comments, something about the solution here feels a bit hackey. I'm guessing there's some LLM involved in the implementation process here, which is fine, but I found it to be able to add fixes but lacks a bit in terms of adding them in the right place according to current architecture. So I think a high level review from your end about where to place things so that they are in the right place is in order here.
@lucaswoj @birkskyum I would appreciate your input here as well. Thanks!

xavierjs:
I’ll get back to this PR in a few weeks. I’m setting it as a draft in the meantime.

HarelM:
I'm closing this in order to clear up my pull request review queue, feel free to reopen or create a new one.

PR #209, opened 2021-07-14
Migrate to typescript

HarelM:
## Launch Checklist

 - [x] confirm your changes do not include backports from Mapbox projects (unless with compliant license)
 - [x] document any changes to public APIs: 
     - Removed `version` from library API.
     - Remove support of IE11
 - [ ] post benchmark scores
 - [x] manually test the debug page
 - [x] apply changelog label ('bug', 'feature', 'docs', etc) or use the label 'skip changelog'
 - [x] add an entry inside this element for inclusion in the `maplibre-gl-js` changelog: `<changelog></changelog>`

### Motivation and some background:
The current code run on node 10, the migration to node 14 is difficult since some of the libraries are no longer supported.
This is also true regarding documentation of essential libraries such as flow (for typings) and tap (for testing).
The goal of this PR is to update the stack to a more widely used technologies and differentiate this library from mapbox-gl to avoid license breach.

### Techonologies
Typescript for type checking
Jest for testing
rollup for packaging
eslint for code checking

The following boilerplate code would be used to facilitate the above widely used technologies.
https://github.com/VitorLuizC/typescript-library-boilerplate

This is a very large task and we'd appreciate any help from anyone :-)

Task by priority:
- [x] Convert src file from Flow to Typescript
- [x] Upgrade CI and references script to esm and node 14
- [x] Make lint pass on all source files
- [x] Make build pass on tsc including rollup required changes
- [x] Remove usage of `@mapbox/geojson-types` and use `@types/geosjon`.
- [x] Remove usage of `@mapbox/point-geometry` and bring it in to this library (after converting it to typescript as this is only one file).
- [x] ~~Convert test, debug, benchmark code to typescript~~
- [x] ~~Make all linters pass (not just for the main library) #272~~
- [x] ~~Make their tsc pass~~
- [x] ~~Convert tests to use jest #273~~
- [x] Make all tests pass
- [x] ~~Remove `tap` and related code around it  #273~~
- [x] ~~Update dependencies (referenced in `package.json`) #270~~
- [x] ~~Change types only declarations to `.d.ts` files #271~~
- [ ] Make sure this PR doesn't break how the docs repo is using this repo
- [x] Make sure `@maplibre/style-spec` is working
- [x] Make sure everything is working as expected and merge back to main - see if there are feature that need to be added to close the gap from when this branch started.

I have decided to reduce the number of changes introduced in this migration - i.e. keep the tests as they are now, do not convert unit tests to typescript.
[...]

PR #7636, opened 2026-05-14
Fix/tile manager used undefined

### LINKED ISSUE #7634: querySourceFeatures returns empty after setData on GeoJSON source (regression since 5.7.1)
MapLibre GL version: 5.20.0 (regression introduced somewhere after 5.7.1)
Last known working version: 5.7.1 with @maplibre/ngx-maplibre-gl: 19.0.1
Environment: Angular app, clustered GeoJSON source (cluster: true, clusterRadius: 60)

**Description**  
 After calling `source.setData(featureCollection)` on a GeoJSON source, `map.querySourceFeatures(sourceId)` returns [] indefinitely — even with `map.isSourceLoaded(sourceId) === true`. This regression affects only
 5.20.0 (and presumably versions between 5.7.1 and 5.20.0); the same code works correctly on 5.7.1.

isSourceLoaded returning true while querySourceFeatures returning [] is the clearest symptom — it indicates tiles are vacuously "loaded" because none were ever requested.

**Root Cause (traced through source)**
 The regression was introduced by the SourceCache → TileManager refactor. The specific issue is a race between setData's synchronous _dataHandler and style.recalculate().

 Code path in 5.20.0:
 1. source.setData(featureCollection) fires _dataHandler synchronously.
 2. _dataHandler calls tileManager.update() (tile_manager.ts (https://github.com/maplibre/maplibre-gl-js/blob/main/src/tile/tile_manager.ts)).
 3. Inside update() (~line 503): if (!this.used) { idealTileIDs = []; } — because tileManager.used is false at this point, no tile IDs are computed and no tiles are ever requested from the GeoJSON worker.
 4. tileManager.used is only set to true inside style.recalculate() (style.ts (https://github.com/maplibre/maplibre-gl-js/blob/main/src/style/style.ts) ~line 763–774), which only runs when _styleDirty=true in the
 next render frame.
 5. tileManager.used starts at false when the map initializes, and the timing window between setData and the first style.recalculate() that activates the source is where the bug occurs.
 6. Consequence: isSourceLoaded() returns true vacuously (~line 156 of tile_manager.ts: if (!this.used) { return true; } — no tiles pending means "loaded").

In 5.7.1 (SourceCache era): querySourceFeatures queried SourceCache._tiles directly without the used gate, so it always worked after setData.

**Reproduction**  
`// Clustered GeoJSON source
 map.addSource('pins', {
   type: 'geojson',
   cluster: true,
   clusterRadius: 60,
   data: { type: 'FeatureCollection', features: [] },
 });

 map.addLayer({
   id: 'pins-layer',
   type: 'symbol',
   source: 'pins',
   filter: ['!', ['has', 'point_count']],
   layout: { 'icon-image': 'marker' },
 });

 // Later, after map is initialized:
 const source = map.getSource('pins') as GeoJSONSource;
 source.setData({
   type: 'FeatureCollection',
   features: [
     { type: 'Feature', geometry: { type: 'Point', coordinates: [17.1, 48.1] }, properties: { id: '1' } },
     { type: 'Feature', geometry: { type: 'Point', coordinates: [17.2, 48.2] }, properties: { id: '2' } },
   ],
 });

 // Immediately or after a delay:
 console.log(map.isSourceLoaded('pins'));          // true  ← misleading
 console.log(map.querySourceFeatures('pins'));     // []    ← wrong, should have 2 features
`
 **Expected**: querySourceFeatures returns the 2 features (or their cluster representations) once the source reports loaded.

 **Actual**: Returns [] indefinitely. isSourceLoaded returns true vacuously because no tiles were ever requested (the tileManager.used=false gate prevents tile requests).

dojchek-sg:
---
  Launch Checklist
  
  - Confirm your changes do not include backports from Mapbox projects (unless with compliant license)
  - Briefly describe the changes in this PR.
  - Link to related issues.
  - ~~Include before/after visuals or gifs if this PR includes visual changes.~~ (no visual changes)
  - Write tests for all new functionality.
  - ~~Document any changes to public APIs.~~ (no public API changes)
  - ~~Post benchmark scores.~~ (no performance-sensitive changes)
  - Add an entry to CHANGELOG.md under the ## main section.
  - Confirm you have read our AI policy (https://github.com/maplibre/maplibre/blob/main/AI_POLICY.md).

  Assisted-By: Claude Sonnet 4.6

  ---
  Description
  
  Fixes #7634.

  TileManager.used is declared as used: boolean but never initialized, so it starts as undefined. In _dataHandler, when a source content event fires (e.g. from setData()), the method calls this.update() whenever
  this.transform is set — without checking this.used.

  Inside update(), the guard if (!this.used && !this.usedForTerrain) evaluates to true when used is undefined, which resets idealTileIDs to [] and evicts all tiles from InViewTiles. As a result,
  querySourceFeatures() returns [] and isSourceLoaded() returns true vacuously — the source appears loaded but data is inaccessible.

  The race condition is triggered when setData() is called before the first requestAnimationFrame fires (before style.recalculate() sets used = true). This happens for example when setData() is called from a
  microtask (Promise.resolve().then(...)).

  Fix: add && this.used to the condition in _dataHandler. When used is not yet true, the render loop's _updateSources() will call update() in the next frame after style.recalculate() runs — deferring tile loading
  by at most ~16 ms but preventing the destructive eviction.

  This is a regression introduced in the SourceCache → TileManager refactor in 5.20.0. SourceCache did not have this issue because querySourceFeatures read from a flat data cache independent of the used gate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🐞 Bug fixes
 - Fix Benchmarks remote compare issue ([#7607](https://github.com/maplibre/maplibre-gl-js/pull/7607)) (by [@xavierjs](https://github.com/xavierjs))
+- Fix `TileManager` calling `update()` in `_dataHandler` when `used` is `undefined`, causing tile eviction and empty `querySourceFeatures()` results when `setData()` is called before the first render frame ([#7636](https://github.com/maplibre/maplibre-gl-js/pull/7636)) (by [@dojchek-sg](https://github.com/dojchek-sg))
 - _...Add new stuff here..._
 
 ## 6.0.0-9
--- a/src/tile/tile_manager.test.ts
+++ b/src/tile/tile_manager.test.ts
@@ -588,6 +588,35 @@ describe('TileManager / Source lifecycle', () => {
 
     });
 
+    test('does not call update in _dataHandler when used is undefined', () => {
+        const transform = new MercatorTransform();
+        transform.resize(511, 511);
+        transform.setZoom(0);
+
+        const tileManager = createTileManager();
+        // Simulate the race: content event fires before style.recalculate() sets used=true
+        (tileManager as any).used = undefined;
+
+        const updateSpy = vi.spyOn(tileManager, 'update');
+
+        tileManager.on('data', (e) => {
+            if (e.dataType === 'source' && e.sourceDataType === 'metadata') {
+                // Set transform directly so _dataHandler's `if (this.transform)` check passes,
+                // without going through update() which would change used
+                (tileManager as any).transform = transform;
+                // Fire content event while used is still undefined (microtask race scenario)
+                tileManager.getSource().fire(new Event('data', {
+                    dataType: 'source',
+                    sourceDataType: 'content',
+                    sourceDataChanged: true,
+                }));
+            }
+        });
+
+        tileManager.onAdd(undefined);
+        expect(updateSpy).not.toHaveBeenCalled();
+    });
+
 });
 
 describe('TileManager.update', () => {
--- a/src/tile/tile_manager.ts
+++ b/src/tile/tile_manager.ts
@@ -813,7 +813,7 @@ export class TileManager extends Evented {
         }
 
         this.reload(e.sourceDataChanged, e.shouldReloadTileOptions);
-        if (this.transform) {
+        if (this.transform && this.used) {
             this.update(this.transform, this.terrain);
         }
         this._didEmitContent = true;


HarelM:
Thanks for taking the time to open this PR! This is truly appreciated!

I don't see any changes to `used` in the following PR:
- #6635

While I don't argue this PR probably caused the issue, I'll be happy to better understand how it broke the code in the way you describe.
Can you maybe highlight the changes that might have affected this in the linked PR?

CC: @wayofthefuture

wayofthefuture:
I'm pretty sure the block you added this.used never included this.used - when I changed it was refactor only I believe. Also, the test failure did not seem to trigger so there is no failed test verification.

dojchek-sg:
@wayofthefuture @HarelM 

> Thanks for taking the time to open this PR! This is truly appreciated!
> 
> I don't see any changes to `used` in the following PR:
> 
> * [Refactor SourceCache to TileManager #6635](https://github.com/maplibre/maplibre-gl-js/pull/6635)
> 
> While I don't argue this PR probably caused the issue, I'll be happy to better understand how it broke the code in the way you describe. Can you maybe highlight the changes that might have affected this in the linked PR?
> 
> CC: @wayofthefuture

You're right, and I apologize for the misleading PR description. After a deeper investigation, PR #6635 did not introduce this bug — it was a pure refactor and the code was  copied **unchanged** from `source_cache.ts`. The bug was already present there. In `source_cache.ts`, `_dataHandler` also called `update() `with only a `this.transform` guard, never checking  `this.used`. The root cause is that `update()` treats `used === undefined` identically to `used === false` (both result in idealTileIDs = [] and evict all tiles), and `_dataHandler` was  never given the matching guard.

dojchek-sg:
> I'm pretty sure the block you added this.used never included this.used - when I changed it was refactor only I believe. Also, the test failure did not seem to trigger so there is no failed test verification.

**Confirmed** — you're correct that your refactor in #6635 was purely structural. The _dataHandler block in source_cache.ts also had only if (this.transform) without this.used, and PR #6635 copied it faithfully. The bug predates the refactor.

Regarding the test: commit f2b861ee1 in this branch was intentionally written red first — the commit message says "Test is failing at this point." The following commit cc12d8ba2  is the fix that makes it green.  You can verify by checking out f2b861ee1 and running the test — it should fail without the fix applied.

HarelM:
Thanks for the extra info. Are you able to find a version where this was working as expected?

dojchek-sg:
@HarelM 
It seems that the bug was there for a longer time - probably from the very beginning. 
The `used` property was introduced in PR #209 (TypeScript migration, commit e78ad7944). From that moment on, `update() `had the guard if (!this.used) — but `_dataHandler` was calling `update()` unconditionally (only checking this.transform). The two were never wired together.

The thing I don't understand thou is that in my use case the `querySourceFeatures()` used to work in version 5.7 and also stopped working in 5.20 (not stopped, but was returning empty in 1/20 of the cases). The latest version where I  see it working is verion 5.19.
Probably the bug  found here is not directly related to the issue that I have and maybe we need further investigation and unfortunatelly I currently don't have time to address it. I falled back to another way of finding the features on the map which doesn't use the `querySourceFeatures()`.

HarelM:
I'll keep this open if someone would like to push this forward. But I think more time is needed in order to investigate this. Thanks for looking into this!

HarelM:
Closing this as the linked issue is not accurate, see my last comment here:
- #7634

I just fixed a bug that might be similar to this one in the geojson source which is available in latest version 6 prerelease, and there's another PR in progress to fix a different bug in geojson and vector tile workers code.

PR #7637, opened 2026-05-14
Minor optimization for PR7549 (using a single FBO for all RTT)

xavierjs:
Hi,

The PR [7549](https://github.com/maplibre/maplibre-gl-js/pull/7549) brings great improvement in performance for terrain rendering.
This PR consists of 2 minor improvements:

* The cache size is controlled (max 512 elements) to avoid OOM errors
* The render to texture for the cache is done more optimally (it reuses the same FBO and we only switch the texture). It also reduces a lot the GPU memory 

## Benchmarks

The benchmark suite runs with same performance as main.
`npm run test-render` runs as before

## Launch Checklist

 - [X] Link to related issues.
 - [X] Post benchmark scores.

 Assisted by Claude Code


HarelM:
@lucaswoj can you take a lot at this?

lucaswoj:
> great improvement in performance

How did you measure the performance improvement?

xavierjs:
> The PR [7549](https://github.com/maplibre/maplibre-gl-js/pull/7549) brings great improvement in performance

The PR 7549 brings great improvements in performance, not this one, and you put the benchmarks results :)
This one does not change the benchmarks.

lucaswoj:
Thanks for digging in on this @xavierjs 

The FBO sharing change is great. I'd expect this to reduce memory usage around 50% (but don't have a reliable way to measure this today).

I'd recommend removing the RTT recycle pool cap (512 items) though. The pool size is already bounded in practice by the maximum number of visible and cached tiles. The ideal size of this pool is different on every device. A phone might OOM well below 512 objects. A desktop with a big screen and complex style might pass 512 in normal use. Picking one number for all devices doesn't fix OOM on small devices and could hurt performance on large ones (because hitting the cap forces tiles to re-render their RTTs, which is exactly what #7549 was trying to avoid).

xavierjs:
> Thanks for digging in on this @xavierjs
> 
> The FBO sharing change is great. I'd expect this to reduce memory usage around 50% (but don't have a reliable way to measure this today).
> 
> I'd recommend removing the RTT recycle pool cap (512 items) though. The pool size is already bounded in practice by the maximum number of visible and cached tiles. The ideal size of this pool is different on every device. A phone might OOM well below 512 objects. A desktop with a big screen and complex style might pass 512 in normal use. Picking one number for all devices doesn't fix OOM on small devices and could hurt performance on large ones (because hitting the cap forces tiles to re-render their RTTs, which is exactly what #7549 was trying to avoid).

Thank you for your feedback

I will disable the recycle pool cap in the next commit so that we can move forward with this PR.
I understand your point, a hard limit is not ideal, however I am still worried by the "no-limit" cache concept. Maybe we could cap the recycle pool depending on the viewport resolution. But yeah, let's keep the debate for another PR :)

lucaswoj:
I'm delving a little deeper on testing these changes and I want to flag another concern. 

Best practices around FBO attachment swapping have shifted over the years, and I'm just now reading up on the state of the ecosystem. I'm fairly certain you're right that on modern desktop drivers, reattaching textures to a single FBO is competitive with or faster than rotating between pre-built FBOs. 

What worries me is mobile. [MDN's WebGL best practices](https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/WebGL_best_practices) still call out that "almost any change to an FBO's attachment bindings will invalidate its framebuffer completeness" and recommend setting up hot framebuffers ahead of time. On tile-based mobile GPUs (Mali, Adreno, Apple, PowerVR) the re-validation cost and the implicit render-pass break are both more expensive than on desktop, and we've historically hit driver bugs on Adreno around `glFramebufferTexture2D` specifically.

Our terrain optimization work right now is aimed primarily at older mobile devices, where the worst case here matters most. The desktop memory win is real, but I'm increasingly hesitant to merge this until we know what it does to mobile frame times.

Have you done any device testing for this PR, particularly on Android with Mali or Adreno GPUs? Even a rough before/after on one mid-range device would help me get comfortable.

lucaswoj:
I can confirm this reduces peak GPU memory usage in the `test/bench/benchmarks/terrain.ts` benchmark by 50%

xavierjs:
You are right if the framebuffer state is deeply changed (if we add or remove depth buffer, stencil buffer, ...), but if we only swap the color buffer it should be faster.

I had the same dilemma around 2017 for a WebGL deep learning engine running in the web browser (used in https://github.com/jeeliz/jeelizFaceFilter ). It does render to texture all the time, to compute neural network layers (input -> output), everything is stored into textures (weights, inputs, outputs, ...). So I really dig out to make the RTT as fast as possible.

And it was really faster to use only 1 FBO for RTT, and switch the texture (about 50 times per frame). The target was mobile devices so I tested on a bunch of devices (Iphone6, IphoneX, Sony Xperia around 2017, Motorolla XPlay). The lib is used quite a lot, also the Jeeliz Virtual Try-on for eyewear, and it is very mobile oriented. Some customers are in countries where most devices are Samsung A series or equivalent. For the eyewear VTO solution the rendering is done using the same WebGL context, so I use a second FBO for the offline rendering operations since the state should be different than for the computation (use of MRT if available and depth buffer). I don't reuse the compute FBO since it may be indeed more expensive.

lucaswoj:
Your experience with Jeeliz is compelling but the testing you describe sounds like it was from around 2017, and Jeeliz may exercise a different RTT pattern than GL JS. It's essential that we get fresh numbers from the `terrain` benchmark on this PR, on a couple of current mobile devices, ideally including at least one Mali GPU, before we merge. Mali, Adreno, and Apple have all had multiple architecture generations since 2017, and the GL JS terrain path has its own attachment layout that may not match what Jeeliz stresses.

Happy to run the benchmarks myself if you don't have hardware handy, just give me a few days.

lucaswoj:
Just ran a comparison of this branch vs main on a Galaxy S23 via BrowserStack. I had to reduce the [`STYLE_COMPLEXITY`](https://github.com/xavierjs/maplibre-gl-js/blob/main/test/bench/benchmarks/terrain.ts#L8) to 2 to get this bench to run reliably on mobile. 

Screenshot 2026-05-15 at 09 04 34
Screenshot 2026-05-15 at 09 03 27

As best I can tell this is working as advertised. Lower memory usage, no measurable perf regression!

lucaswoj:
(As a small caveat, I'll add that this is measuring 484ms of jank across a 1000ms test duration, so it's possible this benchmark is just too borked to tell us anything)

HarelM:
Do let me know if I can merge this or not, and if there's a need for any code changes. 
I'm OK with merging this, creating a new pre-release, and continue trying to solve other issues.

lucaswoj:
I'm okay with merging this

src/render/painter.ts
```diff
@@ -94,9 +94,17 @@ export class Painter {
...
      */
     _rttObjectRecyclePool: RTTObject[];
+    /**
+     * Shared FBO used by all RTT render passes. The color attachment is
+     * swapped to the target RTTObject's texture via {@link bindRTT}.
+     * Created lazily on first use.
+     */
+    _rttFbo: Framebuffer | null;
+    _rttDepthRenderbuffer: WebGLRenderbuffer | null;
+    _rttFboSize: number;
```
HarelM: Can't we use _rttFbo's width/height instead of storing another parameter? I mainly want to reduce "out of sync" bugs...

HarelM:
Added a minor question about a parameter.
Another approach would be to place all the newly added fields inside an object so that when handling them the code reader would know that they are "related". Current code does that by prefixing with the same name. Both approaches are fine, I don't have a strong preferences either way, just wanted to highlight this option.

src/render/painter.ts
xavierjs: We can use `getRenderbufferParameter`, however it will be another drawcall, and it is not recommended to call `GL.getParameter` or similar at runtime. In most cases it is fine but for some GPU configs it can be badly implemented (forcing the GPU to sync).

xavierjs:
> Added a minor question about a parameter. Another approach would be to place all the newly added fields inside an object so that when handling them the code reader would know that they are "related". Current code does that by prefixing with the same name. Both approaches are fine, I don't have a strong preferences either way, just wanted to highlight this option.

It is done in the last commit

lucaswoj:
I'm really psyched for this to land. Thanks again @xavierjs!

HarelM (approved):

PR #7638, opened 2026-05-15
Fix geometry length check for polygons and lines

widefire:
the  true vertex count is len - first


HarelM:
What does this change solve? Is there a test that can cover this fix?

widefire:
> What does this change solve? Is there a test that can cover this fix?
Fix: guard against degenerate geometry after leading duplicate vertex removal

The existing check `if (len < (isPolygon ? 3 : 2)) return` validates
the total vertex count but does not account for `first`, which is
incremented when consecutive leading duplicate vertices are stripped
from the beginning of the line.

For example, a polygon ring with vertices [P0, P0, P1, P1] results in
len=3 (after trailing duplicate removal) and first=1 (after leading
duplicate removal), leaving only 2 effective unique vertices — which
is insufficient for a valid polygon.

Without this fix, such degenerate geometry could cause the polygon
initialization code (which accesses vertices[len-2]) to produce
incorrect normals, or lead to errors during join computation.

The added check `if (len - first < (isPolygon ? 3 : 2)) return`
ensures that the effective number of vertices (after both trailing and
leading duplicate removal) meets the minimum requirement: 3 for
polygons and 2 for lines.

HarelM:
Can you please add a test to cover this scenario? Thanks!!

widefire:
Added unit tests for degenerate geometry handling

HarelM:
Thanks! Can you add a changelog comment? 
After that this can be merged.

HarelM (approved):

PR #7646, opened 2026-05-17
refactor: basic UBO infrastructure for background layers

CommanderStorm:
This PR implements most of the way how our shader code should likely work.

I am not sure if we want to manually write out every UBO defitittion by hand or if the way that done via codegen.

Both have pros and cons.
Also not sure how ergonomic the exact API is, likely some part of that can be codegen-ed also.

Assisted-By: Claude Opus


src/shaders/glsl/background_pattern.vertex.glsl
```diff
@@ -1,10 +1,28 @@
...
+// u_pixel_coord_upper can reach 2^15 at deep zoom
+// -> beyond mediump's guaranteed exact-integer range
+// The split itself is still required because even highp (FP32, 24-bit mantissa) cannot store the un-split pixel coord
+// exactly once it crosses 2^24 (zoom >= 16 at tileSize=512);
+// get_pattern_pos keeps every intermediate small enough to avoid reassembling the full value.
+layout(std140) uniform DrawableUBO {
+    highp vec2 u_pixel_coord_upper;
+    highp vec2 u_pixel_coord_lower;
+    highp float u_tile_units_to_pixels;
+};
```
CommanderStorm, line 25: Before this change, this was `mediump`...

HarelM:
@birkskyum is doing similar work, is this PR aligned with his work? 
I'll let him review it as he probably has more meaningful inputs.

src/gfx/drawable.ts
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * Self-contained draw call.
+ *
+ * Built per-frame by the per-layer draw function (e.g. `drawBackground`),
+ * carrying everything `draw()` needs to issue GL calls without consulting any
+ * mutable shared state. Resolves its GLSL program lazily via
+ * `painter.useProgram(this.shaderName)`.
+ *
+ * Mid-2.x: only Background uses Drawable; other layers still go through
```
HarelM: This is probably not important as part of the code, it can be highlighted as part of the PR comments/title.

src/gfx/drawable.ts
```diff
@@ -0,0 +1,178 @@
...
+ * `Program.draw()`. Both paths coexist until the migration is complete.
+ */
+export class Drawable {
```
HarelM, line 61: Is this aligned/consistent with the drawable approach in native? Can we leverage the architectural decisions that were made there. As far as I know the architecture change there was a good direction and allowed multiple graphics backgrounds to work, so I'd be happy to adopt it here as well if possible. If this is too big of a lift, I would like at least to make sure we take as much as we can from the native approach.
CC: @louwers, @sjg-wdw

src/shaders/glsl/background_pattern.vertex.glsl
birkskyum: No, it was already `highp`:

https://github.com/maplibre/maplibre-gl-js/blob/c56ba23d54a34b7ee3d69158e74552fc8fc4fb8f/src/shaders/glsl/_prelude.vertex.glsl#L1-L3

CommanderStorm: `u_tile_units_to_pixels` was, but `u_pixel_coord_upper` was not.

birkskyum: The `u_pixel_coord_upper` was vec2 without precision defined.
vec2, being a float vector, will in case of unspecified precision take the default float precision.
The default float precision is highp.

Let me know which of these three statements you don't agree with.

birkskyum: The [spec](https://registry.khronos.org/OpenGL/specs/es/3.0/GLSL_ES_Specification_3.00.pdf) states this in section 4.5.4 about default precision qualifiers:

> Hence for float, _**floating point vector**_ and matrix variable declarations, either the declaration must include a precision qualifier or **_the default float precision_** must have been previously declared.

CommanderStorm: Ah, I was believing that It would need to have been defined by for vec2 as well for the precision to take effect.

birkskyum:
> @birkskyum is doing similar work, is this PR aligned with his work? 
> 
> I'll let him review it as he probably has more meaningful inputs.


I'm working on a more thorough PR for milestone 2.1. This milestone lays the foundation for a lot of follow-up work, both for webgl2 and webgpu, so it deserves a bit more love than a quick prompt.

CommanderStorm:
> I'm working on a more thorough PR for milestone 2.1

Ok

PR #7658, opened 2026-05-18
Optimization: Batch geometry for stencil buffer prepass

xavierjs:
Batch geometry for stencil buffer prepass. For now this is for debating, not for merging :)

Better result for LayerLine benchmark:
12% improvement on Macbook pro M4 Max, 11% with GPU disabled.


xavierjs:
I’m closing this PR because I think the benefit is too small compared to the added complexity.

Issue #7640, opened 2026-05-15
[Tracking] Graphics Modernization (GM)


birkskyum:
After sunsetting WebGL1 as part of GL JS v6 ( #3947 ) we are positioned to move the graphics pipeline forward towards a modern WebGL2 and WebGPU dual renderer. This is a tracking ticket for that modernization.

**Phase 1 - Complete - WebGL2 Pre-Architectural Improvements** - WebGL2 improvements that don't depend on the Drawable architecture from Phase 3. Texture allocation, shader attribute declaration, vertex attribute binding, and modern GLSL ES 3.00 features.

* #7641
* #7643
* #7644
* #7661
* #8130
* #8143
* #8145

**Phase 2 - **In progress** - Shared rendering state** - Move projection, terrain and per-frame shader values into shared std140 UBOs. Bring shared rendering inputs together so they’re clearer to use and don’t need to be set up separately for each draw.

* #8290
* #8304 
* #8306 
* #8313 
* #8318 
* WIP - GM2.6 Centralize tile projection data in `getProjectionDataForTile()`.
* GM2.7 Centralize tile terrain data in `getTerrainDataForTile()`
* TBD

- Closes #4337 

**Phase 3 - Drawable Rendering Architecture** - Each draw call becomes a persistent Drawable that carries its geometry, settings and uniform data, following maplibre-native's drawable / tweaker / uniform buffer design. Per-layer and per-drawable uniform data lives in one arena buffer and is re-uploaded only when it changes. Layers migrate one at a time; each layer's legacy draw path is deleted one release after its drawable path ships.

* GM3.1 Add the Drawable core and render solid background through it
* GM3.2 Move pattern background to drawables, with textures and a per-drawable UBO
* GM3.3 Move fill to drawables, with tile clipping, data-driven paint blocks, outlines and layer opacity
* GM3.4 Move circle to drawables
* GM3.5 Move line and its variants to drawables
* GM3.6 Move raster, color relief and the hillshade render pass to drawables
* GM3.7 Add render targets and move the offscreen passes to drawables
* GM3.8 Move symbol and collision debug to drawables
* GM3.9 Move sky, atmosphere, terrain, debug and clipping masks to drawables
* GM3.10 Remove the legacy draw path

- Closes #4511

**Phase 4 - Instanced Rendering** - Draws repetitive features in bulk with WebGL2 instancing. Circles first, symbols if measurements justify it.

* GM4.1 Add instancing to the Drawable contract
[...]

PR #7661, opened 2026-05-19
GM1.4 Use `flat` to opt out of interpolation for constant shader varyings

birkskyum:
Milestone 1.4 in
- #7640 

In WebGL1, all varyings are smooth-interpolated across a triangle, even when the value is the same at every vertex. WebGL2 adds the flat qualifier, which means "no interpolation", so per-feature constants pass through unchanged, saving unnecessary work.

Main benefit here is arguably correctness, and preparation for WebGPU which has an equivalent `@interpolate(flat)`

## Launch Checklist

 - [X] Link to related issues.
 - [ ] Post benchmark scores.


src/shaders/glsl/symbol_text_and_icon.fragment.glsl
```diff
@@ -11,7 +11,8 @@ uniform highp float u_gamma_scale;
 uniform lowp float u_device_pixel_ratio;
 
 in vec4 v_data0;
-in vec4 v_data1;
+in vec3 v_data1;
+flat in float v_is_sdf;
```
HarelM, line 15: Doesn't this change requires an update in the calling code? I'm confused a bit by this change. Is this a bug fix?

src/shaders/glsl/symbol_text_and_icon.vertex.glsl
```diff
@@ -31,7 +31,8 @@ uniform vec2 u_translation;
 uniform float u_pitched_scale;
 
 out vec4 v_data0;
-out vec4 v_data1;
+out vec3 v_data1;
+flat out float v_is_sdf;
```
HarelM, line 35: Doesn't this change requires an update in the calling code? I'm confused a bit by this change. Is this a bug fix?

HarelM (approved):
I'm approving this although I left two comments. I'll let you decide how to proceed.

src/shaders/glsl/symbol_text_and_icon.vertex.glsl
birkskyum: It's not a bug fix, and there is no typescript calling code to update.

The vertex shaders "out" has to match the fragment shader "in". The values are passed from the vertex shader to the Fragment shader.

The two comments you made are exactly the two places that has to stay in sync.

src/shaders/glsl/symbol_text_and_icon.fragment.glsl
birkskyum: See https://github.com/maplibre/maplibre-gl-js/pull/7661#discussion_r3283614129

PR #7665, opened 2026-05-20
Optimize CrossTileSymbolIndex by leveraging promoteId to automatically pair known feature symbols across tiles

bradymadden97:
Original context: https://github.com/maplibre/maplibre-gl-js/pull/6641

When there are a large number of symbols with the same label (or no label) on the same tile, we currently run a very slow algorithm to try to match symbols to a "cross tile symbol" at the next highest or lowest zoom level as you zoom in or out. This is to prevent symbols from fading out and back in as one zoom level's tile is faded out and another fades in. 

Thanks to the great comment [here](https://github.com/maplibre/maplibre-gl-js/pull/6641#issuecomment-3733418716), #6641 is OBE and this approach can be used, which is much simpler. If a feature has a `promoteId`, just use that for matching across tiles. If not, it continues using the existing (slow) case.

(In my case, I know my features have IDs, so this optimization is perfect). 

Without promoteId - you can clearly see the huge stall when a bunch of symbols share a tile:

https://github.com/user-attachments/assets/dd09a774-0d9a-4d51-be63-1f8663bff04d

With promoteId - the stall completely goes away:

https://github.com/user-attachments/assets/bed0fe52-240c-42f7-9e18-694c829fa624

I wrote a benchmark in https://github.com/maplibre/maplibre-gl-js/pull/7664. Now re-ran with these changes:

Ran:
```
npx st --no-cache -H localhost --port 9966 .
node --no-warnings --experimental-transform-types test/bench/run-benchmarks.ts --compare '' CrossTileSymbolIndex
```
Outputs:
```
Starting headless chrome at: http://localhost:9966/test/bench/versions/index.html?compare=
                      bmadden/cross-tile-id-from-feature-dirty ef4ed4d 
 CrossTileSymbolIndex                                       0.3233 ms 
```

Compare to original benchmark here: https://github.com/maplibre/maplibre-gl-js/pull/7664 this is a 2,500x speedup.

----

Public API changes:
* This PR does not change any public APIs. The promoteId option already exists and is already documented. The optimization is entirely internal — the same inputs produce the same visual output, just faster. No new options, methods, or behavior changes are exposed to users.

## Launch Checklist

 - [x] Link to related issues.
 - [x] Post benchmark scores.


HarelM:
Can you point me to where the cross tile id is assigned? Maybe draw a small diagram of the data flow? I'm having a hard time understanding how this works...
Not that I doubt the excellent work you did, just trying to understand what the limitation of this, maybe memory concerned etc...

bradymadden97:
Will do. One note on memory - these maps are defined at the source level (during one iteration of this PR they were a global map), so I think that's an improvement as now they get gc'd away with the source if it gets removed.

I'll see if I can get to your Q before end of the week.

HarelM:
Thanks! Don't worry about the flakey render tests...

src/symbol/symbol_layout.ts
```diff
@@ -723,7 +728,7 @@ function addSymbol(bucket: SymbolBucket,
         numIconVertices,
         numVerticalIconVertices,
         useRuntimeCollisionCircles,
-        0,
+        crossTileID,
```
HarelM, line 731: Does this has the possibility to overflow? I've seen some nasty bugs related to integer overflow when using this emplaceBack method.

src/symbol/symbol_layout.ts
```diff
@@ -699,6 +702,8 @@ function addSymbol(bucket: SymbolBucket,
     const variableAnchorOffset = getTextVariableAnchorOffset(layer, feature, canonical);
     const [textAnchorOffsetStartIndex, textAnchorOffsetEndIndex] = addTextVariableAnchorOffsets(bucket.textAnchorOffsets, variableAnchorOffset);
 
+    const crossTileID = feature.id != null && layer.id && crossTileIDs ? getCrossTileID(`${feature.id}-${layer.id}`, crossTileIDs) : 0;
```
HarelM, line 705: Does this already use the `promoteId`?

HarelM:
I see, so this is basically assigning ids in case it's possible, the rest is handled as it was before when ids can be used.
So the only concern I have is with the integer overflow and the lifecycle of this map - i.e. when it is getting cleaned.

HarelM:
A new version of KDbush was just released, I'm not sure if it's relevant to this PR but you were trying to improve performance there, so I thought it would be worth mentioning...

bradymadden97:
Here's the data flow for how crossTileID gets pre-assigned in this PR.

### The problem

The existing CrossTileSymbolIndex runs on the main thread after tile parsing. When a new bucket arrives, it tries to match each symbol to symbols from other zoom levels by spatial proximity (anchor position + label key). When many symbols share the same key (e.g. all have the same label or no label), this becomes very slow — it's doing pairwise coordinate comparisons or KDBush range queries across potentially thousands of symbols.

### The optimization

If we know a feature's identity (via promoteId → feature.id), we can assign a stable crossTileID during layout on the worker thread, before the main-thread matching ever runs. The same feature parsed in tile z/x/y and tile z+1/x'/y' gets the same ID, so no spatial matching is needed.

### Data flow

  ```
┌─────────────────────────────────────────────────────────────┐
  │  WORKER THREAD                                              │
  │                                                             │
  │  GeoJSONWorkerSource                                        │
  │    ._crossTileIDs: Map<string, number>  ◄── lives here,     │
  │    │                                        persists across │
  │    │                                        all tile parses │
  │    │                                                        │
  │    ▼                                                        │
  │  _parseWorkerTile()                                         │
  │    │  only passes map if params.promoteId is set            │
  │    ▼                                                        │
  │  WorkerTile.parse()                                         │
  │    │                                                        │
  │    ▼                                                        │
  │  performSymbolLayout({ ..., crossTileIDs })                 │
  │    │                                                        │
  │    ▼                                                        │
  │  addSymbol()                                                │
  │    │                                                        │
  │    │  feature.id != null && crossTileIDs?                   │
  │    │  ┌─YES──────────────────────────┐                      │
  │    │  │ key = `${feature.id}-${layer.id}`                   │
  │    │  │ crossTileID = getCrossTileID(key, map)              │
  │    │  │   └─ map.get(key) ?? (map.size + 1)                 │
  │    │  └──────────────────────────────┘                      │
  │    │  ┌─NO───────────────────────────┐                      │
  │    │  │ crossTileID = 0  (old path)  │                      │
  │    │  └──────────────────────────────┘                      │
  │    │                                                        │
  │    ▼                                                        │
  │  bucket.symbolInstances.emplaceBack(..., crossTileID, ...)  │
  │    (stored as Uint32)                                       │
  │                                                             │
  └──────────────────────────┬──────────────────────────────────┘
                             │  tile result sent to main thread
                             ▼
  ┌─────────────────────────────────────────────────────────────┐
  │  MAIN THREAD                                                │
  │                                                             │
  │  CrossTileSymbolIndex.addBucket()                           │
  │    │                                                        │
  │    ▼                                                        │
  │  symbolInstance.crossTileID ??= 0                           │
  │    (was `= 0` before — this preserves pre-assigned IDs)     │
  │    │                                                        │
  │    ▼                                                        │
  │  findMatches() loop:                                        │
  │    if (symbolInstance.crossTileID) → skip ✓  (fast path)    │
  │    if (!symbolInstance.crossTileID) → spatial match (old)   │
  │    │                                                        │
  │    ▼                                                        │
  │  Any remaining symbols with crossTileID === 0               │
  │  get a new ID from CrossTileIDs.generate()  (old path)      │
  │                                                             │
  └─────────────────────────────────────────────────────────────┘
```

### Why this is correct

The same feature appearing in multiple tiles (e.g. at different zoom levels, or in adjacent tiles) will always produce the same "${feature.id}-${layer.id}" key, so it gets the same crossTileID from the map. This is exactly what the main-thread spatial matching was trying to achieve — just without the expensive coordinate comparisons.

### Memory / lifecycle

The _crossTileIDs map lives on GeoJSONWorkerSource — one per GeoJSON source. It grows with unique featureId × layerId combinations (not tiles), never shrinks, and gets GC'd when the source is removed. For a source with 100K features and a few symbol layers, it's a few MB at most.

The map must be append-only — if we evicted entries, two tiles containing the same feature could end up with different IDs.

### Limitations

- Only works when promoteId is set (features need stable IDs). Without it, falls through to the existing spatial matching.
- Only implemented for GeoJSONWorkerSource in this PR — VectorTileWorkerSource could get the same treatment in a follow-up.
- The worker-side counter and main-thread CrossTileIDs counter are independent ID spaces, but that's fine — symbols with pre-assigned IDs skip findMatches entirely, so the two spaces never interact.

File: src/symbol/symbol_layout.ts (hunk 1 above)
Comment (bradymadden97) [reply src/symbol/symbol_layout.ts:731]: I think only in the same way the existing (global) crossTileID ([here](https://github.com/maplibre/maplibre-gl-js/blob/a93a5d69e7ae9b57b23263c007ecbb4f68d285b0/src/symbol/cross_tile_symbol_index.ts#L173)) has the possibility to overflow. Except this is a counter per feature per source, so you'd need 2^32 features in a source.

**Comment by bradymadden97** [comment]
Found a possible regression with multi-point / line geometries:

For symbol-placement: 'line', a single feature generates multiple anchors, and addSymbol() is called once per anchor. Since the key is "${feature.id}-${layer.id}" (no anchor disambiguation), all anchors from the same feature get the same crossTileID.

In placeLayerBucketPart() at line 463:
```
if (seenCrossTileIDs[symbolInstance.crossTileID]) return;
```

This means only the first anchor would be placed; the rest would be skipped as "duplicates." For line-placement labels (e.g., road names repeated along a highway), this would suppress all but one label per feature. 

Will need to look into testing this a bit more I think.

HarelM:
Thanks for the update, this is not covered by existing render tests, right? Maybe a different PR to add this test would make sense.
Thanks for looking into this.
It's worth updating the PR name to indicate this is only relevant for geojson source some how, as this might be implied by `promoteId`, but it's not obvious from the title.

HarelM:
Version 6 was just released so if you would like to push this forward let me know. Currently it's in draft, when it will be moved to ready for review I'll do another sweep. Thanks.

HarelM:
@johncarmack1984 let me know if you're still free to take this PR and "rebase" it.

johncarmack1984:
Yes, taking it. Rebased onto main and opened as #8470, with Brady's commit kept.

One change to the shape. The id map on the worker is gone: `Placement` keys its state by crossTileID across every source, so a counter per GeoJSON source collided with the main thread's ids and with other sources', and one id per feature is what dropped the line labels Brady found. Instead the worker hashes the promoted id into the symbol key it already computes from the text, so a label is only matched against the same feature's label. Same speedup, no map to grow or clear, no second id space. Numbers, the two-source case and a line-placement render test are in the PR.

HarelM:
- Replaced by #8470

Issue #7672, opened 2026-05-21
Severe FPS drop under globe projection with pitch, zoom > 5, and bearing pointing across the antimeridian


mschmengler:
**maplibre-gl-js version**: 5.24.0

**browser**: Chrome, Windows 11

### Steps to Trigger Behavior

1. Open a map using globe projection with a center near the antimeridian, a pitch around 60°, zoom > 5, and a bearing rotated so the camera looks across the date line. The minimal config below is enough.
2. Observe the frame rate without touching anything — the camera is parked.
3. Optionally drag or zoom further; the freeze gets worse.

Minimal config that reproduces:

```html
<div id="map"></div>
<script>
    const map = new maplibregl.Map({
        container: 'map',
        style: 'https://demotiles.maplibre.org/globe.json',
        center: [175.596, -39.221],
        zoom: 5.5,
        pitch: 60,
        bearing: 270,
    });
</script>
```

The same behavior reproduces in the live quickstart map at <https://maplibre.org/maplibre-gl-js/docs/> — pitch it, rotate the bearing toward the antimeridian, zoom past 5.

### Link to Demonstration

<https://jsbin.com/xiturubusa/edit?html,output>

### Expected Behavior

Steady frame rate similar to other longitudes at the same zoom, pitch, and bearing. Behavior at e.g. `lng=−172` should match behavior at `lng=+175` with appropriate bearing.

### Actual Behavior

- Frame rate collapses to single digits.
- The condition is sustained even with no interaction — the camera is parked, no inputs are happening, and the page is still pathologically slow.
- Page interaction becomes janky as long as the map is in this state.
- Dragging the map or zooming further in while in this state makes the freeze noticeably worse.

**Required conditions (all three):** globe projection, pitch ≥ ~40°, and zoom > 5 with the camera center near the antimeridian and bearing rotated so the view crosses the date line. Removing any one restores normal frame rate. Mercator is unaffected. GeoJSON and custom layers make the symptom dramatically worse but are not required to trigger it.

---

*An AI assistant helped draft this report and investigate the conditions that reproduce it.*

NeluQi:
Even at close zoom, performance drops significantly.
Even without the sphere, just on the surface.

<img width="947" height="451" alt="Image" src="https://github.com/user-attachments/assets/16892f0a-8503-4a73-b0dd-a9c4dd7cbf36" />

[...]

PR #7641, opened 2026-05-15
GM1.1 Enable mipmaps for NPOT tiles

birkskyum:
Part of below as Milestone 1.1 
- #7640  

Enable mipmaps for non-power-of-two raster tiles. Previously, mipmaps were blocked for any texture that wasn't both square and power-of-two in dimension - a WebGL1 restriction. WebGL2 supports mipmaps on all dimensions.

This reduces aliasing artifacts (moire patterns) on non-square raster tiles at high-pitch camera angles.

**Before (without mipmaps)**

actual-before

**After (with mipmaps)**

actual-after

## Launch Checklist

 - [X] Confirm **your changes do not include backports from Mapbox projects** (unless with compliant license) - if you are not sure about this, please ask!
 - [X] Briefly describe the changes in this PR.
 - [X] Link to related issues.
 - [X] Include before/after visuals or gifs if this PR includes visual changes.
 - [X] Write tests for all new functionality.
 - [ ] Document any changes to public APIs.
 - [ ] Post benchmark scores.
 - [x] Add an entry to `CHANGELOG.md` under the `## main` section.
 - [X] Confirm you have read our AI policy [here](https://github.com/maplibre/maplibre/blob/main/AI_POLICY.md).

PR #7673, opened 2026-05-21
Improve terrain RTT texture filtering

AveryanAlex:
I am building UAV ground station software and use MapLibre GL JS for synthetic vision.

When terrain is enabled, far raster textures can look noisy at high pitch, especially when the camera angle constantly changes. In my use case the UAV attitude changes by about ±1°, which made the distant terrain texture artifacts very noticeable.

This PR improves filtering for terrain render-to-texture (RTT) textures.

Changes:
- Enable mipmaps for terrain render-to-texture textures.
- Regenerate mipmaps after rendering into each terrain RTT texture.
- Sample terrain RTT textures with `LINEAR_MIPMAP_LINEAR` instead of just `LINEAR`.

Image comparison:

Before:
image
After:
image

Video comparison:
https://github.com/user-attachments/assets/251749be-9b98-4920-98b8-356e3a62fd5b

(Maybe) related issue: #7640

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

Generated-By: GPT-5.5


HarelM:
I just merged a change to terrain performance, can you resolve conflicts so I can push this forward?
Can you maybe add a render test with the relevant fix, I don't see the existing render tests that changed to have a noticeable change related to the problem you described.

AveryanAlex:
> Thanks for taking the time to open this PR. I just merged a change to terrain performance, can you resolve conflicts so I can push this forward? Can you maybe add a render test with the relevant fix, I don't see the existing render tests that changed to have a noticeable change related to the problem you described.

I merged the main branch, resolved the conflicts, fixed a minFilter caching issue, and added a focused unit test covering mipmap generation and trilinear sampling.

If you meant an integration render test specifically, the existing render-test expectations do change noticeably. For example:
`test/integration/render/tests/projection/globe/terrain/fill-planet-pitched/expected.png` (cropped)
Before:
image
After:
image
I also noticed that my original visual comparison screenshot was inaccurate because I had this in my code:
```javascript
canvasContextAttributes: {
  antialias: true,
}
```

Here is a more accurate comparison:
Before:
image
After:
image

Separately, I think it may be worth considering enabling antialiasing by default, as it noticeably improves terrain edge visuals. Should I open PR for it?

AveryanAlex:
Some of Windows integration tests are failed because of timeout now: `Error: Hook timed out in 10000ms.`, but they passed on my original commit.

src/webgl/render_to_texture.test.ts
```diff
@@ -63,6 +63,10 @@ describe('render to texture', () => {
     } as any as SymbolStyleLayer;
 
     let layersDrawn = 0;
+    const createMockRTTObject = (size: number) => ({
```
HarelM: Can this be replaced with `function ...`? I got confused when reading it for the first time...

HarelM:
@xavierjs @lucaswoj @larsmaxfield can you take a look at this PR?
I think it's OK, but it would be great if you can take a peak.

HarelM:
Can you add the raster image test somehow? I think it's a very good way to show this feature.
I've looked at the other tests you updated, and their changes are very subtle (probably because they use solid colors).
So a new render test with the raster image you provided would be great.
Thanks!!

larsmaxfield:
Does #7641 add this already for v6? (@birkskyum )

HarelM:
This wasn't merged so it's not in version 6 yet.
#7641 was released as part of 6.0.0-10.

xavierjs:
Hi @AveryanAlex thank you for this PR.

For performance sake, it is a good idea to use mipmapping for terrain RTT textures.

If we don't use mipmapping, 2 pixels consecutive pixels of the rendered terrain will fetch texels that are far from each other if the texture is very minified (far away), and this is really bad for GPU texture cache coherency.

I would use texture anisotropic filtering only the pitch is high enough to get even better result (the pitch threshold could be mapped to the map option `anisotropicFilterPitch`). If the pitch is low we can also disable trilinear texture filtering (  use `gl.LINEAR_MIPMAP_NEAREST`as a min filter)
I think it would also be nice to cap the anisotropic texture filtering level to 4, maybe 8 but not more (`this.context.extTextureFilterAnisotropicMax`). I don't think we need crazy anisotropic filtering levels like 16 and some low config may return high levels and then be laggy.

I think that antialiasing should not be enabled by default on the Maplibre canvas. Maplibre runs great on very low end devices or with software rendering because it does not require antialiasing (and is well optimized too). I think it should stay an option of the map disabled by default. We could log a warning in the console if terrain or 3D feature are used without antialiasing.

birkskyum:
> Does #7641 add this already for v6? (@birkskyum )

@larsmaxfield This PR builds on top of the work I did in #7641, it's fully complimentary.

AveryanAlex:
> Can you add the raster image test somehow? I think it's a very good way to show this feature. I've looked at the other tests you updated, and their changes are very subtle (probably because they use solid colors). So a new render test with the raster image you provided would be great. Thanks!!

I added representative render test.

Before:
main-256
After:
my-256

512x512 version showed no significant difference.

src/webgl/render_to_texture.test.ts
AveryanAlex: Okay, replaced

AveryanAlex:
Oh, windows render test timed out again. Prev commit passed tests btw.

zheng655 (approved):

NeluQi:
Could this be related to these issues? (setSourceTileLodParams)

- https://github.com/maplibre/maplibre-gl-js/issues/7672
- https://github.com/maplibre/maplibre-gl-js/issues/7699

src/webgl/render_to_texture.ts
```diff
@@ -175,6 +176,8 @@ export class RenderToTexture {
                     painter.renderLayer(painter, painter.style.tileManagers[layer.source], layer, coords, options);
                     if (layer.source) tile.rttFingerprint[layer.source] = this._rttFingerprints[layer.source][tile.tileID.key];
                 }
+                gl.bindTexture(gl.TEXTURE_2D, obj.texture.texture);
```
HarelM, line 181: This code path doesn't look like it's using gl stuff, should these two method calls be placed in a different place maybe?

test/integration/render/tests/terrain/rtt-mipmap-minification/pitch-70/style.json
```diff
@@ -0,0 +1,49 @@
+{
+  "version": 8,
+  "metadata": {
+    "test": {
+      "height": 256,
+      "width": 256,
+      "allowed": 0.001,
+      "description": "High-frequency raster content draped over pitched terrain. Without mipmaps and trilinear filtering on the terrain render-to-texture result, the distant terrain aliases into moire patterns instead of filtering smoothly."
```
HarelM, line 8: The test should indicate what's currently being tested and not what might have been the case in the past...

HarelM:
I've added a few minor comments, but overall this looks good. I don't think it's related to the severe FPS drop. This PR is mostly about antialiasing...

clement-igonet (commented):
I tested this on a terrain scene at pitch 80 with a continuous pitch oscillation of about ±1 degree, close to your UAV attitude case. The difference is clear in motion: the far field shimmers on main and stays stable with the patch (top is main, bottom is this PR, same synchronized scene):

main vs PR under a small pitch oscillation

In stills the patch trades a little crispness for that stability, which seems like the right trade at grazing angles. Two questions: did you measure the cost of regenerating mipmaps when an RTT tile re-renders, and the extra texture memory (about a third more per RTT tile)? Live page to reproduce: https://maplibre.confinia.io/debug/rtt-filtering-compare-7673.html (run `window.wiggle()` in the console).

clement-igonet:
I measured this more carefully and my first impression was too generous, the effect is mixed. Mean frame-to-frame luma change during the same synchronized ±1 degree pitch oscillation: in the mid-field city band the patch is about 30% calmer (0.42 vs 0.61), but in the far mountain band it changes more than main (0.72 vs 0.56), which looks like trilinear mip-level transitions breathing as the sampling derivative shifts. So the win is real in the mid-distance but there may be a new soft flicker at the horizon. Measured under SwiftShader, worth re-checking on real GPUs where derivative handling differs. Maybe worth trying a small negative LOD bias or NEAREST mip selection to compare.

clement-igonet:
I resolved the conflict with main in AveryanAlex/maplibre-gl-js#1 (CHANGELOG restructure after 6.1.0; your checkerboard golden verified against the merged tree), merging it will update this PR.

HarelM:
@AveryanAlex what's the status of this PR?
If anyone else would like to drive this through and address @xavierjs comment and mine feel free to do so.

AveryanAlex:
> @AveryanAlex what's the status of this PR? If anyone else would like to drive this through and address @xavierjs comment and mine feel free to do so.

Hello! Sorry, I was a bit busy and tired. I will address your comments soon. About performance - looks like it can be noticeable with 1-frame render (random ci timeouts are maybe related), but should be amortized in real use cases.

HarelM:
Note that there are merge conflict and a few open code review comments. Thanks!

HarelM:
- Closing in favor of #8328

PR #7675, opened 2026-05-21
Fix handling cross-origin blob URL in Ajax utils

katemihalikova:
When map runs inside a `blob:` URL that is of different origin than parent, `getReferrer` (from Ajax utils) thinks it's in blob URL worker bundle and tries to access `window.parent.location.href`, which fails with `DOMException` `SecurityError`, preventing the map from loading. This can happen, for example, in some development environments, e.g. in Google Gemini Canvas environment (not to be confused with HTML Canvas) which compiles its AI-generated page into a cross-origin blob for user preview. Example here: https://gemini.google.com/share/1bba61e0f224.

I fixed the `getReferrer` function by adding try-catch block around the parent access to catch the `DOMException`. Alternative would be to use `window.location.origin === window.location.ancestorOrigins[0]` to check that parent is same-origin, however `location.ancestorOrigins` has been added to Firefox only earlier this year so it's still pretty new feature.

I also fixed the other place where `blob:` URLs are handled - `sameOrigin` (also from Ajax utils). This one uses heuristic to determine whether a URL is same-origin. Before, it considered any `blob:` URL to be same-origin, I fixed it to actually check the origin too, also added a check for opaque `blob:` URL.

## Launch Checklist

### DIFF
--- a/src/util/ajax.test.ts
+++ b/src/util/ajax.test.ts
@@ -3,6 +3,7 @@ import {
     getArrayBuffer,
     getJSON,
     type AJAXError,
+    getReferrer,
     sameOrigin
 } from './ajax.ts';
 import {isAbortError} from './abort_error.ts';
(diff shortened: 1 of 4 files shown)


src/util/ajax.ts
```diff
@@ -289,12 +297,13 @@ export function sameOrigin(inComingUrl: string): boolean {
     // A relative URL "/foo" or "./foo" will throw exception in URL's ctor,
     // try-catch is expansive so just use a heuristic check to avoid it
     // also check data URL
-    if (!inComingUrl ||
-        inComingUrl.indexOf('://') <= 0 || // relative URL
-        inComingUrl.startsWith('data:image/') || // data image URL
-        inComingUrl.startsWith('blob:')) { // blob
-        return true;
+    if (!inComingUrl) return true; // empty URL
```
HarelM, line 313: Can you move the comment to be tsdoc comment and elaborate on the added flow if needed please?

HarelM:
Looks good I guess, I'm not sure this check is actually valuable as we keep changing it to fit all kind of use cases, but that's outside the scope of this PR... Let's keep it for now and continue updating it...

src/util/ajax.ts
katemihalikova: I've moved the original top-level comment to TSDoc and added some more info to it, is it sufficient this way?

katemihalikova:
I've fixed the TSDoc documentation according to your comment.

src/util/ajax.test.ts
```diff
@@ -133,6 +134,46 @@ describe('ajax', () => {
         }
     });
 
+    test('getReferrer method (outside Worker)', () => {
```
HarelM: This probably needs to be split to like 3 tests.

HarelM:
Besides splitting the test this looks good. Thanks!

katemihalikova:
I've split the test into three and added a fourth one so all combinations (https/blob, same/different origin) are tested.

HarelM (approved):

PR #7680, opened 2026-05-23
fix: correct linesofar when line-dasharray is used with line-offset

### LINKED ISSUE #7671: line-dasharray rendering broken when line-offset is used
## Description

When combining `line-dasharray` with `line-offset` on a `line` layer, the 
dashes are incorrectly rendered — they appear misaligned or with
inconsistent spacing along the line.                                      
                                                            
Removing `line-offset` restores correct dash rendering.

## Steps to reproduce                                                     
   
Add a line layer with the following configuration:                        
                                                            
```json
  {
    "id": "bbox-layer",
    "type": "line",
    "source": "bbox-source",
    "layout": {
      "line-join": "miter"                                                  
    },
    "paint": {                                                              
      "line-color": "#000000",                              
      "line-width": 1,
      "line-dasharray": [2, 2],
      "line-offset": -32
    }                                                                       
  }
```

## Actual behavior

With line-offset: -32, the dashes are irregularly spaced / broken.

Image
                                                                            
## Expected behavior

Dashes should render correctly regardless of line-offset.

Image                                    
                                                                            
## Environment                                               

- MapLibre GL JS: 5.24.0

### Link to Demonstration

https://jsbin.com/wayefux/edit?html,output

Fatemebookanian:
## Problem
When `line-offset` is combined with `line-dasharray`, the dash pattern 
becomes misaligned at line joins. This happens because `a_linesofar` 
is calculated along the original (non-offset) line path, but the rendered 
vertices are shifted by `offset2`.

Fixes #7671

## Root Cause
In `line_sdf.vertex.glsl`, the texture coordinates `v_tex_a` and `v_tex_b`
used the raw `a_linesofar` value. When `line-offset` moves the vertex 
position, the actual path length changes slightly at joins, but 
`a_linesofar` was not adjusted to reflect this.

## Fix
- Compute `offset_linesofar` that corrects `a_linesofar` based on the
  offset direction (`a_direction`) and the length of the offset vector
- Use `offset_linesofar` instead of `a_linesofar` in `v_tex_a` and `v_tex_b`
- Move `adjustedThickness` calculation before `offset_linesofar` to 
  ensure correct ordering

## Testing
- Added unit test in `line_bucket.test.ts` to verify `linesofar` values
  are consistent and monotonically increasing
- All 2850 existing unit tests pass


src/data/bucket/line_bucket.test.ts
```diff
@@ -224,4 +224,35 @@ describe('LineBucket', () => {
 
         expect(bucket.isEmpty()).toBe(true);
     });
+    test('linesofar values should be consistent with and without offset joins', () => {
```
HarelM, line 227: I would asvise to add a render test instead of this complicated unit test

HarelM:
Please open a new PR and fill the relevant template. Otherwise I'll be forced to assume this is AI generated and I'm conversing straight with an AI agent.

src/data/bucket/line_bucket.test.ts
Fatemebookanian: Hi HarelM, this is my first contribution to MapLibre!

I replaced the unit test with a render test as suggested. The render test is located at:
test/integration/render/tests/line-dasharray/with-offset/

It includes two lines — one with line-offset and one without — both using line-dasharray. The expected.png was generated with the fix applied, showing that dashes render consistently regardless of offset.

Thanks for the feedback and guidance!

PR #7681, opened 2026-05-23
fix: correct linesofar when line-dasharray is used with line-offset

Fatemebookanian:
## What does this PR do?

Fixes a bug where using `line-dasharray` together with `line-offset` 
causes the dashes to render incorrectly.

## How to test?

Render test added at:
`test/integration/render/tests/line-dasharray/with-offset/`


HarelM:
Have you read my previous request to use the PR template?

Fatemebookanian:
Description

Fixes incorrect dash rendering when line-dasharray is used together with line-offset.

The problem was that a_linesofar was based on the original line geometry, while the rendered vertices are shifted by the offset join geometry. At joins this caused the dash spacing to become inconsistent.

This change adjusts the linesofar value used for dash texture coordinates so it better matches the rendered offset geometry.

Fixes #7671.

Changes
adjust linesofar calculation for offset joins
use corrected linesofar value for dash texture coordinates
add render test for dashed lines with offset
Testing

Added render test:

test/integration/render/tests/line-dasharray/with-offset

Verified that:

dash spacing stays consistent with offset
existing tests still pass

HarelM:
I'm now assuming I'm talking to AI given the fact that I'm not getting a straight answer.
Please read our AI policy and act accordingly.

PR #7693, opened 2026-05-25
Introducing `setWorkerFactory` to customize Web Worker initialization

edoardocavazza:
Similar to Mapbox's `workerClass` property, I propose to introduce a `setWorkerFactory` to allow for more advanced management of worker initialization. This should ensure greater interoperability with various bundlers. For example, in Vite it is now possible to import the script as a worker instead of an asset/url.

## Launch Checklist

 - [ ] Link to related issues.  [N/A]
 - [ ] Include before/after visuals or gifs if this PR includes visual changes. [N/A]
 - [ ] Post benchmark scores.  [N/A]


HarelM:
I would like to avoid this if there's no strong use case for this.
If you want to run some code in the worker thread you have `importScriptInWorkers`.
The worker code itself handles most of what maplibre needs, so chaining it is not really feasible...

edoardocavazza:
Here's my use case: when bundling my Map component as IIFE using Vite lib mode, the worker file is being inlined as base64 but not transformed, so the initialization fails because the worker is still using es modules but its url is not ending as `.mjs`.

IMHO, basing the decision to load the Web Worker with type module solely on the extension seems like a very weak solution to me, and it's easy to confuse why it's not working. All the developer needs to do is configure the asset management to use the `js` extension (which would be correct anyway) to break the initialization.

HarelM:
While I don't disagree there's a problem here I think it just moved the problem from one place to the other, as the developer experience around this changes so you need to use the worker factory now.
Can you think of a different solution that might work better somehow and avoid the user need to know that they need to use this worker factory?

HarelM:
A different solution was provided as part of this PR:
- #7745 

So closing this in favor of the other, if we see that's not enough we can reopen/redo this one.

PR #7705, opened 2026-05-27
`line-dasharray` step transition lags one zoom level when the step's branches are data-driven

### LINKED ISSUE #7704: `line-dasharray` step transition lags one zoom level when the step's branches are data-driven
I believe there is some sort of off-by-one error while transitioning between zoom levels when using data-driven `line-dasharray` layers (see #5812).

**browser**: Firefox 150.0.1

### Steps to Trigger Behavior

 1. [View linked demonstration](https://jsbin.com/paqifur/edit?html,output). Note that the upper line should only have either long red dashes or short blue dashes, only. Both dash length and colour should change at z16.
 2. Gradually zoom in (e.g. with mousewheel). Notice that (incorrect) long blue dashes are shown at z >16 and then (correct) short blue dashes appear at z > 17.
 3. Gradually zoom back out. Note that (incorrect) short red dashes are shown at z < 16 and then (correct) long red dashes are shown at z < 15.
 
When zooming in, the incorrect long blue dashes are also quite fuzzy, which might be useful to note.

### Link to Demonstration

[https://jsbin.com/paqifur/edit?html,output](https://jsbin.com/paqifur/edit?html,output)

### Expected Behavior

When transitioning z16, both colour and line-dasharray should change together.

### Actual Behavior

* When zooming in, the colour changes at z16 but the line-dasharray changes at z17.
* When zooming out, the colour changes at z16 but the line-dasharray changes at z15.

<details><summary>Step by step with screenshots</summary>
#### 1. start at z15.5

Everything is as expected, long red dashes:

Image

#### 2. zoom in to z16.5

The colour has correctly changed to blue, but the dashes are incorrect: 

Image

#### 3. zoom in to z17.5

The colours and the dashes are now both correct. The dashes fix themselves at z17.

Image

#### 4. zoom back out to z15.5

The colour has correctly changed, but the dashes are still short when they should be long here. Note that this is different than what was seen on first load.

Image

#### 5. keep zooming out to z14.5

Now the dashes are long again.

Image

</details>

lucaswoj:
fix https://github.com/maplibre/maplibre-gl-js/issues/7704

Data-driven `line-dasharray` and `*-pattern` properties pre-evaluate their expression at `zoom-1`, `zoom`, and `zoom+1`. The bucket packs those into two vertex buffers, one for the zoom-in direction (`mid` paired with `min`) and one for zoom-out (`mid` paired with `max`), and the renderer binds whichever matches the current zoom direction at draw time.

`CrossFadedBinder._setPaintValues` was writing the pair into `(a_dasharray_from, a_dasharray_to)` as `(mid, min/max)`, the opposite of what `CrossFadedConstantBinder` writes for the same property. With `fadeDuration: 0` the shader's `u_mix` snaps to 1 and renders `a_dasharray_to`, so the data-driven path locked onto the value from the integer zoom we'd just crossed away from, lagging the constant path by one zoom level.

The fix swaps the emplace order to `(min/max, mid)` so both paths agree: `from` is the integer zoom we're moving away from, `to` is the one we're moving toward. I also renamed the abstract `emplace` parameters from `midPos` / `minMaxPos` to `fromPos` / `toPos` so they match the attribute layout.

This bug predates #5812. It's been latent in the data-driven `*-pattern` path since [mapbox/mapbox-gl-js#6289](https://github.com/mapbox/mapbox-gl-js/pull/6289) (2018-08-27). I suspect it went unnoticed there because zoom-stepped `*-pattern` is a less common pattern than zoom-stepped `line-dasharray`. #5812 finally surfaced it for `line-dasharray`.

## Tests

Two regression tests under `tests/regressions/`, one per binder:

 - `maplibre-gl-js#7704-dasharray` renders at zoom 10.5 with a top-level `step` on `["zoom"]` whose branches contain `["get"]`. Expected: short `[2,2]` dashes (post-step). Before the fix: long `[8,4]` dashes (pre-step).
 - `maplibre-gl-js#7704-pattern` does the same with `line-pattern`. Expected: `generic_metro`. Before the fix: `generic_icon`.

I confirmed both tests fail without the fix and pass with it.

## Launch Checklist

 - [x] Link to related issues.
 - [x] Post benchmark scores.

Assisted-By: Claude Opus 4.7


HarelM (approved):
Thanks for the super quick turn around!

PR #7707, opened 2026-05-28
querySourceFeatures() throws 'Block overruns tile' on overzoomed MLT …

### LINKED ISSUE #7701: querySourceFeatures() fails for overzoomed MLT tiles
**maplibre-gl-js version**: 6.0.0-X

**browser**: Firefox, Safari latest

### Steps to Trigger Behavior

 1. Add vector layer encoding="mlt"
 2. Overzoom vector layer
 3. Run `map.querySourceFeatures()` on the layer

### Expected Behavior

`map.querySourceFeatures()` should return array of features on the map

### Actual Behavior

`Error: Block overruns tile: 8337 > 8192` etc

### Workaround

Set `zoomLevelsToOverscale` option to `undefined` when instantiating the map

### Analysis

I believe the root cause of this is that with the addition of `zoomLevelsToOverscale` default value of 4, mlt tiles are now overzoomed and the overzoomed `rawTileData` is no longer a valid mlt format and can't be re-parsed the by mlt decoder. 

When an MLT tile is fetched, the `rawTileData` is overzoomed and cached under the `featureIndex`. When `querySourceFeatures()` is called, the overzoomed `rawTileData` is fetched from the `featureIndex` and re-fed to the MLT decoder which can no longer decode the MLT 

1. Decoding succeeds here: https://github.com/maplibre/maplibre-gl-js/blob/ef9b36d2a8b0a843e8d8d04102e1278e206d3a10/src/source/vector_tile_worker_source.ts#L53
2. Raw tile data is overzoomed and overwritten here: https://github.com/maplibre/maplibre-gl-js/blob/ef9b36d2a8b0a843e8d8d04102e1278e206d3a10/src/source/vector_tile_worker_source.ts#L101
3. The overzoomed MLT data fails to decode here: https://github.com/maplibre/maplibre-gl-js/blob/ef9b36d2a8b0a843e8d8d04102e1278e206d3a10/src/data/feature_index.ts#L118

ted-piotrowski:
…tiles because overzoomed tiles are re-encoded in MVT format

With the addition of `zoomLevelsToOverscale: 4`, Maplibre now overzooms tiles by default. This revealed a bug with MLT tiles that are overzoomed because the overzoomed data is re-encoded in MVT format. The tile encoding for these overzoomed tiles was incorrectly stored in the feature index as MLT. This caused a decoding error on subsequent attempts to decode the data during querySourceFeatures() calls

- Fixes https://github.com/maplibre/maplibre-gl-js/issues/7701

Root cause and fix was investigated and validated by @ted-piotrowski

Fix and test was Generated-By: Claude 5.8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🐞 Bug fixes
 - _...Add new stuff here..._
+- querySourceFeatures() throws 'Block overruns tile' on overzoomed MLT tiles because the reported encoding doesn't match the re-encoded MVT data ([#7707](https://github.com/maplibre/maplibre-gl-js/pull/7707)) (by [@ted-piotrowski](https://github.com/ted-piotrowski))
 - Fix geometry length check for polygons and lines in LineBucket after duplicate vertex trimming([#7638](https://github.com/maplibre/maplibre-gl-js/pull/7638)) (by [@widefire](https://github.com/widefire))
 - `line-dasharray` step transition lags one zoom level when the step's branches are data-driven ([#7705](https://github.com/maplibre/maplibre-gl-js/pull/7705)) (by [@lucaswoj](https://github.com/lucaswoj))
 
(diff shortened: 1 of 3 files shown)


HarelM (approved):
Looks great! Thanks!

PR #7711, opened 2026-05-29
Optimization: vertex shader opacity culling for lines and fills

xavierjs:
Hi,

This PR is very similar to this one: [GPU performance optimization: early culling of transparent symbols in vertex shaders 
](https://github.com/maplibre/maplibre-gl-js/pull/7364), but this one is for `line` and `fill` layers instead of `symbols` layer.

## Problem

For fill and line layers, if the opacity is constant (`fill-opacity` or `line-opacity`) in the layer style, then it is forwarded to the shader as a uniform. If the opacity is 0, the rendering is fully discarded on the JavaScript side. In this case everything is good :).

If the opacity is not constant (it typically depends on a feature state value), then it will be forwarded to the vertex shader through a VBO and interleaved with other properties. If the opacity is null it cannot be discarded on the JavaScript side.

In many cases we have many polygons or line objects displayed with a zero opacity, especially:
* For hover effects (the object appears only if the cursor hovers)
* For filtering (we only display some polygons or lines depending on search filters or events)

With the current implementation, those objects are still rendered, but transparently. It means that:
* The vertex shader will project all the points
* The rasterizer will compute all the pixels between the projected points
* The fragment shader will run for all the pixels, even there is an early exit

By enabling the overdraw inspector with `map.showOverdrawInspector=true` we can see the issue. Here is the `create-a-hover-effect.html` example where I changed the opacity when there is no hover effect from 0.5 to 0. My pointer is hovering over the Texas state:
Screenshot 2026-05-29 at 11 53 15 AM

Without this PR, if we enable the overdraw inspector and keep hovering over Texas state, the other states are still rendered, so we cannot see the difference between the hover state and the other states:
Screenshot 2026-05-29 at 11 57 19 AM

With this PR, only the Texas is rendered filled:
Screenshot 2026-05-29 at 11 53 38 AM

## Solution

We propose a very simple fix: If the opacity is zero in the vertex shader, we project the point at the same position outside of the viewport. Then, it should be discarded automatically by the GPU at the clipping test. The rasterizer work is skipped or negligible and the fragment shader is not executed at all.

## Benchmarks

The benchmark suite runs the same before and after this PR. This is expected since no benchmark reproduce this situation (layers with line opacity feature ).
I have added a benchmark showcasing the advantage of this PR, `LayerLineFilteredOpacity`
In this benchmark we display 1/20 geojson objects using feature state bind to the `line-opacity` property.
On my laptop, we observe more than X2 performance improvement (2.21ms vs 4.82ms)

## Misc

No related issues.

## Launch Checklist
 - [X] Link to related issues.
 - [X] Post benchmark scores.

Assisted by Claude Code


HarelM:
I think it might be worth adding a comment similar to other decisions in the shaders inside the added if explaining the -2, -2, -2 position.
CC: @lucaswoj 

The solution seems pretty simple and straight forward, so I don't have other meaningful comments.

HarelM (approved):

HarelM:
Please update bundle size and I'll merge this. Thanks!

PR #7712, opened 2026-05-31
docs: clarify projection matrix backing types

### LINKED ISSUE #6316: defaultProjectionData.fallbackMatrix and .mainMatrix do not reflect the correct types

cat0825:
## Summary

- Add explicit `mat4f32` and `mat4f64` aliases for projection matrices backed by `Float32Array` and `Float64Array`.
- Document that renderer projection matrices are commonly 32-bit for WebGL uniforms, while custom-layer projection matrices can be 64-bit to preserve precision before upload.
- Keep `ProjectionData.mainMatrix` and `fallbackMatrix` typed as `mat4` for compatibility with gl-matrix operations, which do not preserve typed-array backing types.

Fixes #6316

## Verification

- `npx eslint src/geo/projection/projection_data.ts`
- `git diff --check`
- `npx vitest run --config vitest.config.unit.ts src/geo/projection/mercator_utils.test.ts src/geo/projection/mercator_transform.test.ts src/webgl/draw/draw_custom.test.ts`

Note: generated local artifacts were produced for testing with `npx tsx build/generate-unicode-data.ts`, `npx tsx build/generate-style-code.ts`, `npx tsx build/generate-struct-arrays.ts`, and `npx tsx build/generate-shaders.ts` because the current npm `codegen` script uses Node's removed `--experimental-transform-types` flag in this local Node 26 environment.

### DIFF
--- a/src/geo/projection/projection_data.ts
+++ b/src/geo/projection/projection_data.ts
@@ -2,6 +2,16 @@ import type {mat4} from 'gl-matrix';
 import type {OverscaledTileID} from '../../tile/tile_id.ts';
 
 /**
+ * A 4x4 matrix backed by 32-bit floats for direct use with WebGL matrix uniforms.
+ */
+export type mat4f32 = mat4 & Float32Array;
+
+/**
+ * A 4x4 matrix backed by 64-bit floats for higher-precision CPU-side projection math.
+ */
+export type mat4f64 = mat4 & Float64Array;
+
+/**
  * This type contains all data necessary to project a tile to screen in MapLibre's shader system.
  * Contains data used for both mercator and globe projection.
  */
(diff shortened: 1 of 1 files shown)


src/geo/projection/projection_data.ts
```diff
@@ -1,6 +1,16 @@
 import type {mat4} from 'gl-matrix';
 import type {OverscaledTileID} from '../../tile/tile_id.ts';
 
+/**
+ * A 4x4 matrix backed by 32-bit floats for direct use with WebGL matrix uniforms.
+ */
+export type mat4f32 = mat4 & Float32Array;
```
HarelM, line 7: Where are these type used? If they are not exported from the index no one can use them outside and since they are not referenced no one is using them inside. So I'm a bit confused about these types...

HarelM:
This doesn't fix the issue I believe.
Note that there's a PR template to fill, and since it wasn't filled correctly I'm forced to close this issue.
Please open a new PR and fill the template as needed.

PR #7726, opened 2026-06-03
fix: guard style methods after WebGL context loss

### LINKED ISSUE #7710: Uncaught TypeErrors when calling map methods when WebGL context is lost

**browser**:  Any

### Steps to Trigger Behavior

 1. Initialize map
 2. Simulate WebGL context loss
 3. Call map method after context has been lost
 4. See TypeError(s) in console

### Link to Demonstration

https://jsbin.com/yikojivahu/1/edit?html,console,output

### Expected Behavior

No TypeErrors occur

### Actual Behavior

I've seen the following TypeErrors in my application:

* "Uncaught TypeError: Cannot read properties of null (reading 'getLayer')"
* "Uncaught TypeError: Cannot read properties of null (reading 'getSource')"
* "Uncaught TypeError: Cannot read properties of null (reading 'getLayersOrder')"
* "Uncaught TypeError: Cannot read properties of null (reading 'listImages')"
* "Uncaught TypeError: Cannot read properties of null (reading 'setFilter')"
* "Uncaught TypeError: Cannot read properties of null (reading 'setPaintProperty')" 

### Analysis

This issue is similar to #7432 for which a guard has been introduced (#7446), but this `null` style can happen in quite a lot of methods.

cyphercodes:
## Description

Fixes #7710.

Guards the style-backed map methods reported in the issue so calling them while the WebGL context is lost does not throw while `this.style` is temporarily cleared.

## Testing

- `npm run test-unit -- src/ui/map_tests/map_webgl.test.ts`
- `npm run typecheck`
- `git diff --check -- CHANGELOG.md src/ui/map.ts src/ui/map_tests/map_webgl.test.ts`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### 🐞 Bug fixes
 - _...Add new stuff here..._
+- Avoid TypeErrors from style methods while the WebGL context is lost ([#7710](https://github.com/maplibre/maplibre-gl-js/issues/7710)) (by [@cyphercodes](https://github.com/cyphercodes))
 - querySourceFeatures() throws 'Block overruns tile' on overzoomed MLT tiles because the reported encoding doesn't match the re-encoded MVT data ([#7707](https://github.com/maplibre/maplibre-gl-js/pull/7707)) (by [@ted-piotrowski](https://github.com/ted-piotrowski))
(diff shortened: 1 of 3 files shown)


HarelM (approved):
THANKS!

PR #7734, opened 2026-06-04
fix: problem with update returned promise of geojson source

HarelM:
## Launch Checklist

This fixes a bug that was introduced as part of the geojson refactoring.
- https://github.com/maplibre/maplibre-gl-js/pull/7172

The bug is that when you call update right after a source initialization and then immediately call getData you won't get the right data.
This is because of an immediate return instead of waiting for the updates to actually finish.
This bug was reported by one of our sponsors.

 - [x] Link to related issues.

Assisted by: Claude


src/source/geojson_source.ts
```diff
@@ -398,7 +399,9 @@ export class GeoJSONSource extends Evented implements Source {
             this._pendingWorkerUpdate.updateCluster = undefined;
         }
 
-        await this._dispatchWorkerUpdate(params);
+        this._isUpdatingWorker = true;
+        this._updatePromise = this._dispatchWorkerUpdate(params);
+        await this._updatePromise;
```
wayofthefuture, line 403: Should line 448 be moved below this? The set and unset of the Boolean used to be next to each other...

HarelM: I wanted to keep the promise setting and bollean update next to each other, but couldn't keep all of them together, so I compromised on this.

wayofthefuture: Not better to leave it at the top of dispatch function? It's split... what if a future function uses dispatch...

HarelM: That's a valid point. I'll move that then.

HarelM:
Let me know if other fixes are needed, if not please approve so this can be merged. Thanks!

wayofthefuture (approved):
There's a comment on line 391 relating to race condition not sure if it's stale or relevant... fyi

HarelM:
I'm not sure either, so I assume it's still relevant.

wayofthefuture:
Would it have been better to set the promise to undefined and at the top if (updatePromise !== undefined) return updatePromise?

HarelM:
No, I tried that, the boolean can be set in the middle of the method while the rest is still needed to be completed before the promise returns... Having said that, feel free to take a stab at this...

PR #7735, opened 2026-06-04
fix: await document.fonts.load() before TinySDF instantiation (#7307)

### LINKED ISSUE #7307: Load local/web fonts before attempting to render using them
## User Story

I've been playing with the new-ish MapLibre GL ability to [render text using local/web CSS-defined fonts](https://github.com/maplibre/maplibre-gl-js/pull/4564) created by @1ec5 instead of the old glyph PBF solution. It's mostly great! However, one thing that has frustrated me is the requirement that fonts already be loaded before MapLibre fires up TinySDF. Here's an example jsbin, with the `document.font.load()` commented out:  https://jsbin.com/bedunizima/1/edit?html,output

I based the jsbin off of the [web fonts example](https://maplibre.org/maplibre-gl-js/docs/examples/style-labels-with-web-fonts/). I would note that the call to load a font is actually asynchronous though, so in both of these examples it would probably be best to wait for the promise to complete before setting the map style, or the font might still come in after text is (incorrectly) rendered.

## Rationale

I understand why this is needed from a technical perspective, but I am left wondering: why doesn't MapLibre do this itself? It waited for glyphs to download before, so it doesn't seem like a stretch to make sure fonts are downloaded in the new paradigm.

## Impact

The main reason this is a bit of a pain is that it requires code to render a style with text (at least when using a font that is not otherwise already made available). That code either has to have knowledge of what fonts are in the style, or the code can go through the page's stylesheets and wait for anything declared with a @font-face, even if it's not used in the style. Neither of these is ideal. Instead, why not have MapLibre make the call to load the font and wait for it before render?

kodeezabdullah:
## Launch Checklist

- [x] Confirm changes do not include backports from Mapbox projects
- [x] Briefly describe the changes in this PR
- [x] Link to related issues
- [ ] Include before/after visuals (N/A - no visual changes)
- [x] Write tests for all new functionality
- [ ] Document changes to public APIs (N/A - internal only)
- [ ] Post benchmark scores (N/A)
- [x] Add an entry to CHANGELOG.md under ## main
- [x] Confirm I have read the AI policy

## Description

Fixes #7307.

When a glyph needs to be rendered locally, MapLibre currently creates a TinySDF instance immediately and rasterizes the first glyph synchronously. If a web font (@font-face) is still loading at that moment, the browser falls back to a system font and that bitmap gets cached forever — even after the real font finishes downloading.

This change:

- Makes `_createTinySDF` async and awaits `document.fonts.load(...)` for the exact same font descriptor TinySDF is about to use.
- Memoizes the load promise per fontstack via the existing `entry.tinySDF` / `entry.ideographTinySDF` cache (now `Promise<TinySDF>`), so font loading happens exactly once per stack.
- Wraps the load in try/catch and emits a `warnOnce(...)` on rejection so unknown font families fall through to the existing browser fallback rather than crashing glyph rendering.
- Propagates async through `_drawGlyph` and both already-async call sites (`_getAndCacheGlyphsPromise` and `_downloadAndCacheRangePromise`).
- Guards `document.fonts` access (`typeof document !== 'undefined'`) so Node/worker contexts and tests without the FontFaceSet API still work.

Tests added in `src/render/glyph_manager.test.ts`:

- Asserts `document.fonts.load` is called before the TinySDF constructor.
- Asserts a rejected `fonts.load` does not break glyph rendering.
- Asserts the load promise is memoized per fontstack (called once across multiple `getGlyphs` calls).

## Disclosure
This PR was developed with assistance from Claude (Anthropic).
All changes were reviewed and verified before submission.


kodeezabdullah:
Hi, this PR fixes #7307 with full test coverage. Would appreciate a review when you get a chance.

HarelM:
Thanks for this.
Can you add the PR check list back and fill it?
It might be better to replace the tests here with a single render test. Render tests is a better way to showcase these kind of bugs.

kodeezabdullah:
A render test won't exercise this code path — render tests use pre-baked glyph PBFs and never hit TinySDF. The bug is a race condition, not a pixel-diff issue. The existing unit tests mock document.fonts.load and assert it's awaited before TinySDF instantiation, which is the only reliable way to verify this fix.

HarelM:
If you don't provide `glyphs` entry in the style this code path is used I believe, but feel free to make sure I'm not wrong...
I agree about the race condition stuff and the fact that render tests might not fail, but from my experience most of the time the font is not loaded before the map renders.
Having said that, If you strongly believe render test is not a good solution do let me know.

kodeezabdullah:
After digging into the test harness, a render test won't exercise this fix as is. The harness has no @font-face mechanism  the only font knob is localIdeographFontFamily which passes a string directly to TinySDF. There are zero matches for @font-face, FontFace, or document.fonts anywhere in test/integration.
For a system font like sans serif, document.fonts.load resolves on the next microtask, so the bitmap is identical before and after the fix. The bug only manifests with a font that needs actual network/decode time, which requires @font-face support the harness doesn't have.
One option would be to extend the harness with a metadata.test.webFonts hook that injects an @font-face rule  happy to do that as a follow-up PR if that's the direction you'd like. Otherwise, the existing unit tests remain the most reliable way to verify this fix.

src/render/glyph_manager.test.ts
```diff
@@ -239,4 +239,66 @@ describe('GlyphManager', () => {
         await manager.getGlyphs({'Arial Unicode MS': [0x30c6]});
         expect(langSpy).toHaveBeenCalledWith(expect.objectContaining({lang: 'zh'}));
     });
+
+    describe('document.fonts.load integration', () => {
```
HarelM: Let's avoid the describe here, other tests in this file are not using it...

HarelM: Oh, wait, it needs after each and stuff, can you move it to the beginning of the file or outside the other describe?

src/render/glyph_manager.test.ts
```diff
@@ -239,4 +239,66 @@ describe('GlyphManager', () => {
...
+        const installDocumentFontsMock = (impl: (font: string) => Promise<unknown>) => {
```
HarelM: I would consider removing this method and just do it in every test. I know, it's not DRY, but it makes the test readability better...

src/render/glyph_manager.test.ts
```diff
@@ -239,4 +239,63 @@ describe('GlyphManager', () => {
...
+    const installDocumentFontsMock = (impl: (font: string) => Promise<unknown>) => {
+        const loadSpy = vi.fn(impl);
+        Object.defineProperty(document, 'fonts', {
+            configurable: true,
+            value: {load: loadSpy}
+        });
+        return loadSpy;
+    };
+
+    afterEach(() => {
```
HarelM: Move this next to before each, at the top.

kodeezabdullah:
Thanks for the review. All three points addressed:
Removed the describe('document.fonts.load integration') wrapper, tests now sit directly under GlyphManager to match the rest of the file (7ab0534).
Inlined installDocumentFontsMock into each test for better readability (8a823aa).
Moved the delete (document as any).fonts cleanup into the existing top level afterEach (8a823aa).
Ready for another look.

src/render/glyph_manager.test.ts
```diff
@@ -239,4 +240,53 @@ describe('GlyphManager', () => {
         await manager.getGlyphs({'Arial Unicode MS': [0x30c6]});
         expect(langSpy).toHaveBeenCalledWith(expect.objectContaining({lang: 'zh'}));
     });
+
+    test('awaits document.fonts.load before instantiating TinySDF', async () => {
+        const order: string[] = [];
```
HarelM: Sorry for all the nit-picking, but is there a different way to solve this?
Using a single spy maybe, return different results per call invocation and checking the calls array to make sure the order is correct?
I'd like to utilize the spies infra as much as possible instead of this "custom" solution.

kodeezabdullah: No worries at all, that's a fair point! I'll use vitest's built-in mock.invocationCallOrder to check the order between the two spies instead of the custom array. Much cleaner updating it now, thanks!

HarelM (approved):

PR #7736, opened 2026-06-05
fix: camera jump on dragend with globe + terrain at low pitch

### LINKED ISSUE #7025: Terrain / camera jumps upon dragend / movend / mouseup when moving the view on globe at low pitch.
Hello folks, I'm running into a camera / terrain issue.

**maplibre-gl-js version**: 5.16.0

**browser**: Firefox, Chrome

### Steps to Trigger Behavior

 1. Create a map with terrain on globe.
 2. Lower the camera pitch closer to the terrain axis
 3. Drag the position / center of the map
 4. Observe the terrain / camera position jumping on dragend / moveend

Note: this doesn't seem to happen when navigating with the arrow keys. Only when dragging and specifically when the mouse is released. It also occurs on mouse wheel zoom.

My gut feeling is it's a camera issue rather than a terrain issue, I can't confirm without digging deeper.

I _might_ have bandwidth in the coming weeks to look at it, but can't make any promises for now so creating an issue in the meantime.

### Link to Demonstration

https://codesandbox.io/p/sandbox/rkwqtz

### Expected Behavior
Terrain / camera should not jump.

### Actual Behavior
Terrain / camera jumps at moveend / dragend.

kodeezabdullah:
## Launch Checklist

- [x] Confirm changes do not include backports from Mapbox projects
- [x] Briefly describe the changes in this PR
- [x] Link to related issues
- [ ] Include before/after visuals (N/A - no visual changes)
- [x] Write tests for all new functionality
- [ ] Document changes to public APIs (N/A - internal method only)
- [ ] Post benchmark scores (N/A - fix removes work, not adds)
- [x] Add an entry to CHANGELOG.md under ## main
- [x] Confirm I have read the AI policy

## Description
Fixes #7025.

On globe projection with terrain, releasing a drag at low pitch caused
the camera to snap to a wrong position, ignoring terrain elevation.

Root cause: GlobeTransform.recalculateZoomAndCenter was calling both
sub-transforms unconditionally instead of delegating to currentTransform
like every other method on the class.

Fix: delegate to this.currentTransform.recalculateZoomAndCenter(terrain)

Tests: 4 regression tests added - 54/54 passing.

## Disclosure
This PR was developed with assistance from Claude (Anthropic).
All changes were reviewed and verified before submission.


kodeezabdullah:
Note on bundle size failure:

The ~4.7 kB raw bundle increase is not from the camera fix — 
globe_transform.ts and vertical_perspective_transform.ts are 
2 net lines removed.

The increase comes from the pbf package version bump that landed 
in main (the same commit that broke the named PbfReader import, 
fixed in commit 0e36887). That bump enlarges the pbf payload in 
the bundle.

I'll update bundle_size.json to reflect the new baseline so CI 
goes green. Let me know if you'd prefer a different approach.

HarelM:
Can you fill the PR template please?

kodeezabdullah:
Sure! Let me fill that in now.

src/geo/projection/vertical_perspective_transform.ts
```diff
@@ -533,7 +533,11 @@ export class VerticalPerspectiveTransform implements ITransform {
 
     recalculateZoomAndCenter(terrain?: Terrain): void {
         if (terrain) {
-            warnOnce('terrain is not fully supported on vertical perspective projection.');
+            // Terrain is not fully supported on vertical perspective projection.
```
HarelM: Won't this just cause other issues?

src/geo/projection/globe_transform.test.ts
HarelM, line 1: Looks like a lot of tests were added for 1-2 lines of code changes. Are you sure all these tests are actually needed?

HarelM:
Thanks for taking the time to open this PR. I've added a few comments. Let me know what you think.
I'm guessing the current code is better than the previous so we should consider merging this as is, but I do want to make sure there's no regression issues sneaking in...

src/geo/projection/globe_transform.test.ts
kodeezabdullah: Thanks for the review.
You're right that the change itself is small. The tests cover four distinct regression scenarios , globe vs mercator delegation, the terrain early return, and preserving the existing non terrain path.
Do you think any of these are redundant and worth removing?

src/geo/projection/vertical_perspective_transform.ts
kodeezabdullah: Good point. The warnOnce was removed because the early return makes it unreachable but you're right that silently doing nothing could be confusing for users who enable terrain on vertical perspective projection.
I can restore the warnOnce before the early return so the warning still fires, just without the elevation mutation that was causing the jump. Would that work?

kodeezabdullah:
Thanks for the review! Happy to address any concerns before this goes in. The regression tests cover the specific behaviors the fix targets let me know if you'd like any of them trimmed.

src/geo/projection/globe_transform.test.ts
HarelM: To be honest, the tests are testing the code changes that were made, but are not really testing the bug/issue.
I think 1 test that really tests the jumps you aimed to solve would be a lot more meaningful.
Other code flows are tested I believe through other tests... Unless you see a coverage issue in the methods you changed.

src/geo/projection/vertical_perspective_transform.ts
HarelM: Let's keep the warnOnce if we're not really fixing this bug, but instead just making sure elevation is not zeroed...

kodeezabdullah: Agreed I will restore the warnOnce.

src/geo/projection/globe_transform.test.ts
kodeezabdullah: That's a fair point. I'll replace the current tests with one that reproduces the actual camera jump scenario globe + terrain + dragend  and verifies the center position doesn't jump after the fix.

kodeezabdullah: Done I restored the warnOnce and replaced the 4 tests with one regression test that actually reproduces the camera jump scenario. The test sets up a GlobeTransform in globe rendering mode, calls recalculateZoomAndCenter with terrain, and asserts the center doesn't change.

src/geo/projection/globe_transform.test.ts
```diff
@@ -603,4 +603,26 @@ describe('GlobeTransform', () => {
             expect(mercator.renderWorldCopies).toBeTruthy();
         });
     });
+
+    // Regression test for #7025: camera jump on dragend with globe + terrain.
```
HarelM: Better to reduce these comments as much as possible and have the test name be as descriptive as possible. This will be linked to this PR and the relevant issue, so there's no need to rewrite the issue in this comment.

HarelM:
One last code review comment and this can be merged. Make sure to resolve the changelog conflict as I just released a new version.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 - Improve `ProjectionData` matrix backing types for renderer and custom layer projection matrices ([#6316](https://github.com/maplibre/maplibre-gl-js/issues/6316)) (by [@cat0825](https://github.com/cat0825))
 
 ### 🐞 Bug fixes
+- Fix camera jump on dragend with globe + terrain at low pitch ([#7736](https://github.com/maplibre/maplibre-gl-js/pull/7736))
```
HarelM: Add "by ..." part.

kodeezabdullah: Done I removed inline comments from the test, resolved the CHANGELOG conflict, and added the "by @kodeezabdullah" attribution.

HarelM (approved):

PR #7742, opened 2026-06-07
fix!: disable icon scaling if offset is applied

### LINKED ISSUE #5588: icon-offset is incorrect at top and bottom of viewport when map is pitched

HarelM:
## Launch Checklist

This PR replaces:
- #5625

- Fixes #5588

This changes how icons are rendered when they have an offset. even if the offset it zero.
See relevant PR and new render test that was added.

Before (right), After (left):
image

 - [x] Link to related issues.

No AI...


CommanderStorm (approved):

HarelM:
@louwers note that I haven't used the test from the native code.
I didn't want to add a new sprite just for this test and found this "oneway" icon in the existing sprite.
It's worth mentioning that when I used this icon, I needed to scale it (x5) and that showed that you need to add offset to the middle arrow to get the right behavior, other the middle icon gets a different scaling, which is hard to see when the icon-size is set to 1.
The downsize is that it looks blurry at this size, but I think it's a better "edge cade test".
I can open a PR to copy the test over if you want.

birkskyum (approved):

HarelM:
since I changed the "shared" test, I opened the following PR in native to match the style.json that I've updated here:
- https://github.com/maplibre/maplibre-native/pull/4336

Merging the above PR should facilitate for parity tests between native and web.

Issue #6839, opened 2025-12-09
fitBounds problem with minZoom


hheexx:
We have a problem with combination of fitBounds and minZoom in Chrome so that for some fitBounds map flies far away and do crazy stuff. Some time comes back sometimes not.

Here is fiddle:
https://jsfiddle.net/eLf5h8x2/

It has 2 calls to fitBounds in setTimeouts few seconds apart. Bounds are very close geographically, map should do a small move. Instead it does some crazy jumps around. 
If you remove minZoom: 10, everything works as expected.

HarelM:
I've looked to see if I see some references to min zoom in the code path that runs the fit bounds and I didn't see something obvious.
I did look back, and this bug seems to exist from version 2 until today.
Feel free to dig into it.

johncarmack1984:
Fixed in #7743 (v6.0.0). This one is resolved and can be closed.

PR #7743, opened 2026-06-07
fix: correct center calculation in flyTo when minZoom is set

YuChunTsao:
Related issue: https://github.com/maplibre/maplibre-gl-js/issues/6839

When minZoom causes zoom to be clamped by `applyConstrain()`, the theoretical scale from Van Wijk formula becomes inconsistent with `tr.worldSize`. This causes center to jump during the animation.

Fix by computing `actualScale` from the clamped `tr.zoom` after `setZoom`, ensuring the scale and `worldSize` are derived from the same zoom value.

Debug for this change: https://yuchuntsao.github.io/maplibre-flyto-minzoom-bug/

Issue 6839 Case
image

Another case
image

The line is drawn while the map moves. Before the fix, the center point was incorrect, which caused crazy jumping in some cases.

## Launch Checklist

 - [x] Link to related issues.
 - [x] Post benchmark scores.


HarelM:
The provided link has a minzoom of 10 and start zoom of 5, is this a value use case?
Is there a way to solve this in a different way? Updating the easing methods seems a bit like a hack.

YuChunTsao:
> Thanks for taking the time to open this PR! The provided link has a minzoom of 10 and start zoom of 5, is this a value use case? Is there a way to solve this in a different way? Updating the easing methods seems a bit like a hack.

I will dive deeper into the Van Wijk formula implementation in the codebase to find a cleaner solution.

From what I understand so far, the issue is likely triggered when the calculated flight curve peaks at a zoom level that is lower than the user-defined minZoom. When the animation logic tries to apply this out-of-bounds zoom, it gets clamped by `applyConstrain()`, causing the theoretical scale to lose sync with `tr.worldSize`.

image

HarelM:
Yeah, that's my understanding as well, I would expect the curve calculation to take minZoom into consideration. If this is not the case, then a fix there would be less of a hack. I'm not too familiar with this code, so take this with a grain of salt.

YuChunTsao:
@HarelM 

> The provided link has a minzoom of 10 and start zoom of 5, is this a value use case?

Yes, the case is so weird but user can set it.
I've added a test to verify the flight path remains correct and that the `newCenter` is calculated accurately, regardless of the `minZoom` setting.

> Is there a way to solve this in a different way? Updating the easing methods seems a bit like a hack.

Now, I capture `startWorldSize` once before calling `setZoom`, as the world size changes whenever `tr.setZoom()` is invoked. I then use `unprojectFromWorldCoordinates` to recover the `newCenter`.

---

I've encountered an issue with the optimal path calculation. Currently, when the user sets a `minZoom` via `flyTo` options, the code triggers a recalculation of `rho`. The flight path is represented by the blue line.

https://github.com/maplibre/maplibre-gl-js/blob/06ee5a23c5455124f088cdc907e2d1db6516718a/src/ui/camera.ts#L1486-L1491

However, `optionsMinZoom` is only assigned when `flyTo` options are provided. If the user only sets the map's `minZoom` without explicitly defining it in the `flyTo` options, the path defaults to the green line.

https://github.com/maplibre/maplibre-gl-js/blob/06ee5a23c5455124f088cdc907e2d1db6516718a/src/geo/projection/mercator_camera_helper.ts#L163-L171

image

Should I fix this in this PR, or track it in a separate issue?

HarelM:
Can these two paths be joint together? i.e. in case there's a map min zoom defined or that the user passed min zoom to fly to handle it in the same way (preferably like the blue line)?

YuChunTsao:
> Can these two paths be joint together? i.e. in case there's a map min zoom defined or that the user passed min zoom to fly to handle it in the same way (preferably like the blue line)?

Whether the user sets `minZoom` via `map.minzoom` or `flyTo` options, it should follow the blue line path (the flight path calculated based on the minZoom-derived `rho`).

Here are the 4 scenarios:

- Neither is set
- Only map.minzoom is set
- Only flyTo options.minZoom is set
- Both are set
  - **If the values are different in this case, which one should take priority? Should `map.minzoom` have a higher priority?**

HarelM:
I would use the max of both if both are set.

YuChunTsao:
commit: https://github.com/maplibre/maplibre-gl-js/pull/7743/changes/63841551f301c8cec749f0958d3545919bd47799

This fix ensures that `scaleOfMinZoom` is correctly calculated, whether the user configures `minZoom` via `flyTo` options or through the map.

---

commit: https://github.com/maplibre/maplibre-gl-js/pull/7743/changes/c96032d7dfd654d9597077c2addc0e14963ec7a5

Previously, when `minZoom` was set, `rho` was unconditionally overwritten:

`rho = Math.sqrt(wMax / u1 * 2);`

This forced the flight arc to peak exactly at the `minZoom` level, even when the natural arc (at the default `rho = 1.42`) would have stayed well within that boundary. In other words, `minZoom` behaved as a target rather than a constraint.

Conceptually, `minZoom` should act as a ceiling on zoom-out: if the default flight path already respects the limit, there's no reason to deepen the arc. This commit fixes the behavior by replacing the assignment with:

`rho = Math.min(rho, Math.sqrt(wMax / u1 * 2));`

Now `minZoom` only reduces rho when the natural arc would exceed it. If the arc stays within bounds, the original rho (whether the default 1.42 or a user-specified curve) is preserved.

YuChunTsao:
@HarelM

I have tested several scenarios and found that the flight behavior is much better than before. However, the fix for the `FlyToOptions.minZoom` behavior in Case 2 (commit https://github.com/maplibre/maplibre-gl-js/commit/c96032d7dfd654d9597077c2addc0e14963ec7a5) might need your review. You can also adjust the parameters using the debug page I created to verify other cases, just in case I missed any edge cases.

Thanks for your time and feedback.

---

## Case 1

https://yuchuntsao.github.io/maplibre-fly-arc-visualizer/#mode=compare&proj=mercator&fromLng=139.6917&fromLat=35.6895&fromZoom=10&toLng=-0.1276&toLat=51.5074&toZoom=10&mapMinZoom=5&curve=1.42&speed=1.2

Previously, if `map.minzoom` was set, the way `newCenter` was handled could cause an incorrect flight path. (which was fixed in commit https://github.com/maplibre/maplibre-gl-js/pull/7743/changes/6de97273f56c031433af39fa87e867a72dd8d235.

## Case 2

https://yuchuntsao.github.io/maplibre-fly-arc-visualizer/#mode=compare&proj=mercator&fromLng=0.051332780001885285&fromLat=51.50697782256039&fromZoom=15&toLng=-0.1276&toLat=51.5074&toZoom=15&minZoom=7&curve=1.42&speed=1.2

Previously, flyTo's minZoom forced the `rho` calculation to match that specific zoom level, instead of acting merely as a boundary constraint (fixed in https://github.com/maplibre/maplibre-gl-js/commit/c96032d7dfd654d9597077c2addc0e14963ec7a5).

src/geo/projection/mercator_camera_helper.ts
```diff
@@ -153,26 +153,32 @@ export class MercatorCameraHelper implements ICameraHelper {
...
+            const effectiveMinZoom = optionsMinZoom ? +options.minZoom : tr.minZoom;
+            const minZoomPreConstrain = Math.min(effectiveMinZoom, startZoom, targetZoom);
             const minZoom = tr.applyConstrain(targetCenter, minZoomPreConstrain).zoom;
             scaleOfMinZoom = zoomScale(minZoom - startZoom);
         }
 
         const easeFunc = (k: number, scale: number, centerFactor: number, pointAtOffset: Point) => {
             tr.setZoom(k === 1 ? targetZoom : startZoom + scaleZoom(scale));
-            const newCenter = k === 1 ? targetCenter : unprojectFromWorldCoordinates(tr.worldSize, from.add(delta.mult(centerFactor)).mult(scale));
+            const newCenter = k === 1
```
HarelM, line 172: I'm having a hard time seeing if this is a code change or not, can you revert the change if there's no logic change here?

src/geo/projection/mercator_camera_helper.ts
```diff
@@ -153,26 +153,32 @@ export class MercatorCameraHelper implements ICameraHelper {
...
+        const startWorldSize = tr.worldSize;
+        const from = projectToWorldCoordinates(startWorldSize, options.locationAtOffset);
+        const delta = projectToWorldCoordinates(startWorldSize, targetCenter).sub(from);
 
         const pixelPathLength = delta.mag();
 
         const scaleOfZoom = zoomScale(targetZoom - startZoom);
 
         const optionsMinZoom = typeof options.minZoom !== 'undefined';
+        const DEFAULT_MAP_MIN_ZOOM = -2; // See map.ts defaultMinZoom.
```
HarelM: This is a bit of a hack.
Why not always use the min zoom either from the transform or from the options?

src/ui/camera.test.ts
```diff
@@ -1946,20 +1946,20 @@ describe('flyTo', () => {
         expect(leftWorld0).toBeFalsy();
     });
 
-    test('peaks at the specified zoom level', async () => {
+    test('flight arc does not zoom below the minZoom flyTo option', async () => {
```
HarelM, line 1957: Since you are touching those test I would reccomend the following:
1. Use setTimeout with promises instead of indenting the code (`sleep(0)` if needed).
2. Reduce as much logic from the tests - no if (zoom < ...) throw and use spies as much as possible.
3. Don't use the same name for different tests.

Changing the tests here can be a different PR if you want, but the complexity of the tests hare can be greatly reduced...

src/ui/camera.ts
```diff
@@ -1487,7 +1489,9 @@ export abstract class Camera extends Evented {
             // w<sub>m</sub>: Maximum visible span, measured in pixels with respect to the initial
             // scale.
             const wMax = w0 / flyToHandler.scaleOfMinZoom;
-            rho = Math.sqrt(wMax / u1 * 2);
+            // Only reduce rho (limit zoom-out). If the natural arc stays within the minZoom
```
HarelM: I'm not sure I understand this change and this comment. Is this something that can happen given the map min zoom you have added?

HarelM:
I've looked at the demos, they look a lot better. The only concern I have is that they look to be running a lot "slower" - i.e. they end later than the original fly-to operation. If this is expected and falls under the relevant time it should take then I'm completely fine with the new improved behavior.

YuChunTsao:
> I've looked at the demos, they look a lot better. The only concern I have is that they look to be running a lot "slower" - i.e. they end later than the original fly-to operation. If this is expected and falls under the relevant time it should take then I'm completely fine with the new improved behavior.

https://yuchuntsao.github.io/maplibre-fly-arc-visualizer/#mode=compare&proj=mercator&fromLng=139.6917&fromLat=35.6895&fromZoom=10&toLng=-0.1276&toLat=51.5074&toZoom=10&minZoom=5&curve=1.42&speed=1.2
If we change the original Case 1 from `mapMinZoom` to `FlyToOptions` minZoom, you'll find that the flight duration is exactly the same.

---

However, I also noticed that under certain conditions, the flight duration can indeed become significantly longer.

https://yuchuntsao.github.io/maplibre-fly-arc-visualizer/#mode=compare&proj=mercator&fromLng=139.6917&fromLat=35.6895&fromZoom=12&toLng=-0.1276&toLat=51.5074&toZoom=12&mapMinZoom=10&minZoom=10&curve=1.42&speed=1.2

Take this case as an example.
When the user sets `minZoom` to `10`, the flight height will be limited. For a long-distance trip, it ended up taking about 40 seconds to complele.
But this only occurs when `minZoom` is set, resulting in a path that is not optimal.
If `minZoom` is not set, the flight works perfectly.

src/ui/camera.ts
YuChunTsao: The core issue is that the original code unconditionally overwrites `rho` whenever `minZoom` is set:

```js
rho = Math.sqrt(wMax / u1 * 2);
```

This means `minZoom` was directly shaping the flight arc — even when the natural arc (rho = 1.42) would never zoom out past the `minZoom` boundary in the first place.

`minZoom` should act as a ceiling constraint: only kick in when the natural arc would actually violate it.

Therefore, I used `Math.min` to fix the issue, if the natural arc already stays within bounds, `rho` is left unchanged.
I also updated the doc comment about `FlyToOptions.minZoom`.

src/geo/projection/mercator_camera_helper.ts
YuChunTsao: I have refactored this according to your feedback.
I removed `DEFAULT_MAP_MIN_ZOOM` entirely and updated the logic to always compute the effective min zoom directly from either `options.minZoom` or `tr.minZoom`, whichever is more restrictive.

src/geo/projection/mercator_camera_helper.ts
YuChunTsao:
```js
// before
const newCenter = k === 1 ? targetCenter : unprojectFromWorldCoordinates(tr.worldSize, from.add(delta.mult(centerFactor)).mult(scale));
```

```js
// after
const newCenter = k === 1
    ? targetCenter
    : unprojectFromWorldCoordinates(startWorldSize, from.add(delta.mult(centerFactor)));
```


Previously, If `minZoom` was set, `tr.setZoom` would sometimes get clamped.
This will change `tr.worldSize`, casuing it to mismatch with the `scale` and leading to incorrect `newCenter` calculations.
To fix this, I cache `startWorldSize` upfront. Since `project` and `unproject` now share the `startWorldSize` value, they no longer rely on `scale`, eliminating the side effects casued by `tr.setZoom`.

src/ui/camera.test.ts
YuChunTsao: Thanks for the suggestions.
I’ve refactored the test functions.

YuChunTsao:
Hi @HarelM, I have addressed your comments and pushed the updated commits. Please take a look when you have a moment. Thank you!

YuChunTsao:
Hi @HarelM 
I just wanted to check in on this PR. Should I go ahead and resolve the conflicts first, or wait for your review on the newly pushed commits? Thanks!

HarelM:
Yes, please resolve conflicts and have the tests pass and I'll do another review. Thanks!

YuChunTsao:
> Yes, please resolve conflicts and have the tests pass and I'll do another review. Thanks!

Conflicts resolved! The CI tests are all passing now. Ready for your review whenever you're free. Thanks!

HarelM:
> When the user sets minZoom to 10, the flight height will be limited. For a long-distance trip, it ended up taking about 40 seconds to complele.

This sounds like the animation speed should also be involved in case min zoom changes the trajectory, wouldn't you agree? I think the total flyto animation time was bounded before this fix, wasn't it?

src/ui/camera.test.ts
```diff
@@ -2033,6 +2049,43 @@ describe('flyTo', () => {
...
+        camera.on('move', () => {
+            if (midLng === undefined) midLng = camera.getCenter().lng;
+        });
+
+        const promise = camera.once('moveend');
+        const stub = vi.spyOn(timeControl, 'now');
+        stub.mockImplementation(() => 0);
+        camera.flyTo({center: [40, 0], zoom: 10, duration: 10});
+
+        setTimeout(() => {
```
HarelM: This is still complicated... Can this be "unindented" somehow?

src/ui/camera.test.ts
```diff
@@ -3803,43 +3856,57 @@ describe('flyTo globe projection', () => {
...
 
             const promise = camera.once('moveend');
 
             stub.mockImplementation(() => 0);
             camera.flyTo({center: [1, 0], zoom: 20, minZoom, duration: 10});
 
-            setTimeout(() => {
-                stub.mockImplementation(() => 3);
+            // sample every frame across the animation so Math.min reflects the true arc bottom
+            for (let t = 1; t <= 10; t++) {
```
HarelM: Maybe it would make sense to export this for look to a method in this file and reuse it across the tests.

src/ui/camera.ts
HarelM: Doesn't the recent code always sets min zoom now? Can we simplify this "if" condition given that?

HarelM:
I think the code looks good overall, I've added some suggestions to the tests and my main concern is the slowness that we both saw. I think it needs to be addressed before this PR can be merged.

YuChunTsao:
> This sounds like the animation speed should also be involved in case min zoom changes the trajectory, wouldn't you agree? I think the total flyto animation time was bounded before this fix, wasn't it?

The flight time was effectively bounded before, but only implicitly. Without `minZoom`, Van Wijk's cost grows only logarithmically with distance, which naturally keeps the duration within a reasonable range. `minZoom`, however, sits outside that model. To satisfy this constraint, the algorithm overrides `rho` externally to lower the arc, breaking the logarithmic property. Since the geographic distance remains unchanged, what actually grows is the algorithm's internal cost.

Therefore, I'd argue the increased duration isn't actually a bug — it remains consistent with the resulting `rho`, strictly follows the formula, and is essentially an honest reflection of this non-optimal path. In fact, this excessively long duration can already be reproduced today via `options.curve`. `minZoom` simply reaches the same outcome through a different parameter.

If we were to recompute the duration specifically for a `minZoom-constrained rho`, it would conflict with how `options.curve` already works. Both parameters ultimately control the same `rho`: setting `curve` below the default value (1.42) lowers it and lengthens the flight in the exact same way `minZoom` does. Special-casing the `minZoom` path would mean the exact same `rho` produces different durations depending solely on whether it derived from `curve` or `minZoom`.  Furthermore, we couldn't extend this recomputation to curve without overriding the user's explicit choice and breaking its fundamental purpose as an animation knob.

The way I see it, it comes down to which behavior we want flyTo to guarantee:

- **Option A**: The duration always strictly corresponds to the trajectory (the current behavior). Users who want a shorter flight can already limit it explicitly via `options.duration`, flying the same trajectory faster.
- **Option B**: A state constraint like `minZoom` should never pollute the animation timing—meaning we recompute the duration automatically, at the cost of introducing the curve inconsistency mentioned above

I lean toward **Option A**: the current duration is the honest cost of a non-optimal path, and `duration` already gives users an explicit, smooth way to control it without the hidden inconsistency Option B introduces.

What are your thoughts on this? Which behavior do you feel would be more appropriate for MapLibre?

src/ui/camera.ts
YuChunTsao: Done

src/ui/camera.test.ts
YuChunTsao: Done

src/ui/camera.test.ts
YuChunTsao: Done

YuChunTsao:
On a related note — while digging into this, I noticed `maxDuration` resets the `duration` to 0 (instant jump) when exceeded, rather than clamping to the limit. It's covered by tests so I assume it's intentional, but the name reads more like a clamp than a cliff. This behavior directly impacts how effectively users can control the long duration mentioned above. Was the reset-to-0 a deliberate choice? Happy to open a separate issue if it's worth revisiting.

https://github.com/maplibre/maplibre-gl-js/blob/0cd6e6efcefddd3696192dea3738d45e25b9d472/src/ui/camera.ts#L1549-L1551

HarelM (approved):

PR #7745, opened 2026-06-07
Build main and worker in same build context to extract shared chunk

### LINKED ISSUE #7744: Build main and worker in same build context to extract shared chunk
## User Story

This is a follow-up to https://github.com/maplibre/maplibre-gl-js/issues/977#issuecomment-4639941954.

> Here's a minimal example that has main and worker importing a shared chunk. Bundle size down to 500KB main + 19KB worker + 480KB shared.
> 
> https://stackblitz.com/edit/vite-jw7ptvdb?file=main.js
> 
> I import maplibre-gl/src directly to let Vite build from source and use a plugin to load and emit the worker in main build so that they share same build context (plugin code from https://github.com/vitejs/vite/issues/16719#issuecomment-2308170706) to allow code splitting to work.
> 
> How about maplibre exporting main and worker mjs that both import from a shared chunk? That way users can just import from maplibre package instead of src, and configure their bundler as needed. The current dist files are kept for direct browser imports only.
>  

## Rationale

Currently, `dist/maplibre-gl.mjs` and `dist/maplibre-gl-worker.mjs` are standalone bundles that duplicate shared code. The only way to get code splitting is to import from source and configure the bundler with custom plugins.
Building main and worker in a single rolldown context extracts shared code into a common chunk (maplibre-gl-shared.mjs). Total dist JS size should drop from ~1,416 KB to ~1,043 KB.

For bundler users, they can import from the package directly instead of from package source (src), without needing to configure rolldown/oxc options just to build a library from source.

For browser import users, the shared chunk is fetched by the main thread and cached by the time the worker imports it, so the browser serves it from cache instead of making a duplicate request.

## Impact

Without this, bundler users must import from maplibre-gl/src and configure rolldown/oxc to build the library from source, browser import users download the duplicated shared code twice — once for main and once for worker.

## AI Assistance

Drafted with the help of AI assistance.

dangkyokhoang:
Fixes https://github.com/maplibre/maplibre-gl-js/issues/7744

Currently, `dist/maplibre-gl.mjs` and `dist/maplibre-gl-worker.mjs` are standalone bundles that duplicate shared code. The only way to get code splitting is to import from source and configure the bundler with custom plugins.

Building main and worker in a single rolldown context extracts shared code into a common chunk (maplibre-gl-shared.mjs). Total dist JS size should drop from ~1,416 KB to ~1,043 KB.

For bundler users, they can import from the package directly instead of from package source (src), without needing to configure rolldown/oxc options just to build the library from source.

For browser import users, the shared chunk is fetched by the main thread and cached by the time the worker imports it, so the browser serves it from cache instead of making a duplicate request.

image

Notes:
- Though minimally, I've updated some of the current tests, because some expectations should change when we split the shared chunk. However, we're in v6 preview so I think those changes should be ok.
- I haven't updated docs and change logs, just submitting the PR for review/feedback first.
- Assisted-By: MiMo, for automating testing builds.

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
This looks good overall, I'm interested to see how this looks for all the bundlers.
Does it really uses the shared file or just "inline" it into the relevant files...

HarelM:
Take a look at the following PR as well, which is in this same "space":
- https://github.com/maplibre/maplibre-gl-js/pull/7693

Let me know what you think about it.
We are still in 6 pre-release so this is a good time to fix/solve this.

dangkyokhoang:
> I'm interested to see how this looks for all the bundlers.
> Does it really uses the shared file or just "inline" it into the relevant files...

Maybe it depends on the bundler. Vite for example builds main and worker separately so it ends up with the shared chunk being inlined into both main and worker. However, if a user ever needs to keep the shared chunk as a separate file, they can configure their bundler to build main and worker in the same context and have the shared chunk to be emitted again. This is just not possible with the current build setup.

dangkyokhoang:
> Take a look at the following PR as well, which is in this same "space":
> 
>     * [Introducing `setWorkerFactory` to customize Web Worker initialization #7693](https://github.com/maplibre/maplibre-gl-js/pull/7693)
> 
> 
> Let me know what you think about it.

I agree with that PR. The library currently relies on a fragile heuristic check to decide if the worker is an ES module, instead of an explicit flag or an explicit factory function.

With the worker now importing from the shared module, bundler users now need to bundle the worker (Vite `?worker` or `?worker&url`) instead of copying the worker file over (Vite `?url`). Then the app be like:

```ts
import workerUrl from "maplibre-gl/dist/maplibre-gl-worker.mjs?worker&url";
setWorkerUrl(workerUrl + "#.mjs");
```

It's awkward but that's how it is. The `workerFactory` checks `const asModule = url?.endsWith('.mjs') ?? false;` to decide if the worker is an ES module and it just implies a setup where the worker file is copied over, not bundled.

I suggest the heuristic ES module check to default to `true` like `const asModule = url?.endsWith('.cjs') ? false : true;` as we now export esm only, and if the heuristic fails, there's already a fallback in `createWorker`. Then we don't need the `#.mjs` hack.

```ts
import workerUrl from "maplibre-gl/dist/maplibre-gl-worker.mjs?worker&url";
setWorkerUrl(workerUrl);
```

But then, we still use a heuristic check instead of an explicit flag to decide module type. Therefore, the library could expose a `setWorkerFactory` instead. Then:

```ts
import MapLibreWorker from "maplibre-gl/dist/maplibre-gl-worker.mjs?worker";
setWorkerFactory(() => new MapLibreWorker());
```

or

```ts
import workerUrl from "maplibre-gl/dist/maplibre-gl-worker.mjs?worker&url";
setWorkerFactory(() => new Worker(workerUrl, { type: "module" }));
```

This enables bundler users to inline the worker in the main bundle. And, in theory, if they ever want to, they can even replicate the cjs hack in v5 with a custom transform plugin and `setWorkerFactory`.

dangkyokhoang:
With the latest commit, the default worker url should be correctly resolved by bundlers, library's bundler (rolldown) or user's bundler (vite, webpack, etc.), without the need for users to manually specify the worker url.

src/util/web_worker.ts
```diff
@@ -23,7 +23,38 @@ function isCrossOrigin(url: string): boolean {
...
+// Bundler deadlock guard would prevent the shared module from being inlined into main.
+try {
+    let src: string;
+    await import(src);
+} catch {}
+
 function defaultWorkerUrl(): string {
+    try {
+        const moduleUrl = import.meta.url;
+        if (!/^https?:/.test(moduleUrl)) return '';
```
HarelM: These two lines look very similar to the below lines, can't the logic be merged toggether?

test/integration/bundler/webpack/package.json
```diff
@@ -11,6 +11,7 @@
     "maplibre-gl": "file:../../../.."
   },
   "devDependencies": {
+    "copy-webpack-plugin": "^14.0.0",
```
HarelM, line 14: Where is this plugin used in the example? Is it doing it automatically?

HarelM: Nevermind, looks at the files in the wrong order...

HarelM:
Also the webpack test is failing...

HarelM:
@edoardocavazza let me know if this PR is good enough for your use case. If it solves it, I think it will be a better solution as is doesn't change the public API so we won't need to support it for the long term...

HarelM:
To be honest, I'm not sure this actually solves anything in terms of bundle size for bundlers, which I think is the main use case nowadays, but it a small change to improve developer experience, so I think it's a step in the right direction.

dangkyokhoang:
> I'm not sure this actually solves anything in terms of bundle size for bundlers

Actually, for modern bundlers, bundle size is a free win now with automatic code splitting. So with proper caching (correct Cache and Vary headers for example), both bundler users and browser import users now benefit from smaller bundles.

What I'm not sure though is if the code splitting heuristic works reliably because the bundler tests are just minimal apps with a single entry point.

I asked AI agent to checkout main and this branch to build and compare the bundle sizes. Result below:

<table>
  <thead>
    <tr>
      <th>Bundler</th>
      <th colspan="3">main branch</th>
      <th colspan="3">esm-shared-chunk branch</th>
      <th>Delta</th>
    </tr>
    <tr>
      <th></th>
      <th>Main</th>
      <th>Shared</th>
      <th>Worker</th>
      <th>Main</th>
      <th>Shared</th>
      <th>Worker</th>
      <th>Total JS</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>esbuild</td>
      <td>1,205 KB</td>
      <td>—</td>
      <td>456 KB</td>
      <td>629 KB</td>
      <td>654 KB</td>
      <td>24 KB</td>
      <td><b>-354 KB</b> (1,661 → 1,307)</td>
    </tr>
    <tr>
      <td>rollup</td>
      <td>920 KB</td>
      <td>—</td>
      <td>456 KB</td>
      <td>514 KB</td>
      <td>472 KB</td>
      <td>19 KB</td>
      <td><b>-371 KB</b> (1,376 → 1,005)</td>
    </tr>
    <tr>
      <td>vite-rolldown</td>
      <td>914 KB</td>
      <td>—</td>
      <td>456 KB</td>
      <td>509 KB</td>
      <td>469 KB</td>
      <td>19 KB</td>
      <td><b>-373 KB</b> (1,370 → 997)</td>
    </tr>
    <tr>
      <td>vite-rollup</td>
      <td>913 KB</td>
      <td>—</td>
      <td>456 KB</td>
      <td>511 KB</td>
      <td>468 KB</td>
      <td>19 KB</td>
      <td><b>-371 KB</b> (1,369 → 998)</td>
    </tr>
    <tr>
      <td>webpack</td>
      <td>958 KB</td>
      <td>—</td>
      <td>452 KB</td>
      <td>1,032 KB</td>
      <td>464 KB</td>
      <td>19 KB</td>
      <td><b>+103 KB</b> (1,410 → 1,534) *</td>
    </tr>
  </tbody>
</table>

<p>* webpack bundles shared into main AND copies it separately — shared code loaded twice. No auto-split, setWorkerUrl still used.</p>

<h3>Worker chunk reduction</h3>
<table>
  <thead>
    <tr>
      <th>Bundler</th>
      <th>Before</th>
      <th>After</th>
      <th>Reduction</th>
    </tr>
  </thead>
  <tbody>
    <tr><td>esbuild</td><td>456 KB</td><td>24 KB</td><td><b>-95%</b></td></tr>
    <tr><td>rollup</td><td>456 KB</td><td>19 KB</td><td><b>-96%</b></td></tr>
    <tr><td>vite-rolldown</td><td>456 KB</td><td>19 KB</td><td><b>-96%</b></td></tr>
    <tr><td>vite-rollup</td><td>456 KB</td><td>19 KB</td><td><b>-96%</b></td></tr>
    <tr><td>webpack</td><td>452 KB</td><td>19 KB</td><td><b>-96%</b></td></tr>
  </tbody>
</table>

<h3>Main chunk reduction</h3>
<table>
  <thead>
    <tr>
      <th>Bundler</th>
      <th>Before</th>
      <th>After</th>
      <th>Reduction</th>
    </tr>
  </thead>
  <tbody>
    <tr><td>esbuild</td><td>1,205 KB</td><td>629 KB</td><td><b>-48%</b></td></tr>
    <tr><td>rollup</td><td>920 KB</td><td>514 KB</td><td><b>-44%</b></td></tr>
    <tr><td>vite-rolldown</td><td>914 KB</td><td>509 KB</td><td><b>-44%</b></td></tr>
    <tr><td>vite-rollup</td><td>913 KB</td><td>511 KB</td><td><b>-44%</b></td></tr>
    <tr><td>webpack</td><td>958 KB</td><td>1,032 KB</td><td><b>+8%</b> (shared bundled in)</td></tr>
  </tbody>
</table>

dangkyokhoang:
> Also the webpack test is failing...

I'm not sure why it fails as it is not supposed to fail with a `/maplibre-gl-worker.mjs (HTTP 404)` error. For the webpack test, I already use the copy plugin to copy the worker file and the shared file over to the dist dir so it should be there.

Anyway, I think we should still expose the `setWorkerFactory` method. For the webpack case, I could've just used this pattern without the copy plugin, cleaner and producing file names with hash for cache-control: immutable.

```ts
setWorkerFactory(() => new Worker(new URL('maplibre-gl/dist/maplibre-gl-worker.mjs', import.meta.url)));
```

And this serves as an escape hatch for bundler users that, for some reason, the code splitting hack doesn't work for them.

HarelM:
I would like to avoid adding `setWorkerFactory` if it's not strictly needed.
I think that your change with the cjs is probably what would solve most of the issues and setting a workerUrl should be enough.
Let's try this way and if really needed, we'll add that in the future...

edoardocavazza:
@HarelM If you change the default heuristic check to load the worker as a module, it should solve my problem.

As @dangkyokhoang says, I also believe that exporting the `setWorkerFactory` method suggests a better pattern. But obviously, I understand your concerns about maintenance, so I'll just thank you if this PR is accepted.

src/util/web_worker.ts
dangkyokhoang: Maybe by moving it out of the try catch block, this is esm so that should be safe. I'm just not sure if `import.meta.url` ever throws.

HarelM: Don't know about try-catch, but the http check is done twice for example...

dangkyokhoang: I've fixed that.

src/util/web_worker.ts
```diff
@@ -23,10 +23,36 @@ function isCrossOrigin(url: string): boolean {
...
+
 function defaultWorkerUrl(): string {
+    const moduleUrl = import.meta.url;
+    if (!/^https?:/.test(moduleUrl)) return '';
+    try {
+        const workerPath = importWorker.toString().match(/["'`](.+?)["'`]/)?.[1];
+        if (workerPath) {
+            return new URL(workerPath, moduleUrl).href;
+        }
+    } catch {
```
HarelM: Can we consolidate the try-catch into a single block of try-catch?

src/util/web_worker.ts
```diff
@@ -23,10 +23,36 @@ function isCrossOrigin(url: string): boolean {
...
+// in order to prevent the shared module from being inlined into main.
+//
+// Rollup/rolldown by default inlines the shared module into the main bundle
+// if the app has only one entry point.
+// The worker would then import the shared module from the main bundle,
+// which fails because main uses APIs that are not available in workers.
+//
+// To work around this, we make main async with a no-op `await Promise.resolve()` at the top level.
+// Bundler deadlock guard would prevent the shared module from being inlined into main.
+await Promise.resolve();
```
HarelM: Is this the only way to solve this? This looks like a nasty hack... :-/

src/util/web_worker.ts
dangkyokhoang: I don't know the why the try catch is necessary in the first place. My take: import.meta.url should be available in a module, and should be a valid URL, so the code in the try block should not fail.

src/util/web_worker.ts
dangkyokhoang: My thoughts as well. The problem is, if there's only one entry point, a bundler would see that main imports shared, main dynamically imports worker which imports shared, so it doesn't have a good reason to split the shared module, so it emits perfectly valid code as follows:

```ts
// main.js
// inlined shared module
export { ... } // the shared module exports

// worker.js
import { ... } from './main.js';
```

That works efficiently when the worker module runs in the same scope as main, worker simply imports shared from main's live bindings. And the bundler expects this is the case, it doesn't know that the worker script is meant to be run in worker scope, and should not import from main.

To let the bundler know, users would need to make the worker import explicit like with Vite `?worker` or `?worker&url`, then we don't even need the `importWorker` hack but let users configure the worker URL directly, and in that case, by default, Vite inlines shared in both main and worker.

Then this hack, let the bundler think worker runs in the same scope as main but make main an async module. To prevent deadlock, the bundler would bail out of inlining shared into main and split shared into a separate chunk, from which both main and worker can import without importing from each other.

HarelM: I think it's OK to let the user configure the worker and main thread code as part of the bundler configuration, like it is done now for webpack.
If this removes this hack, then I prefer it... I know it requires a bit more work for the developers, but it moves the responsibility of adjusting this on their end.
In version 5 we could not use all kind of new features because the bundler assumed all kind of things on the code, which is also bad.
I think moving the responsibility to the developer alongside relevant examples is the right direction here...

dangkyokhoang: I agree, and I actually think that's the better direction here. Then just the first commit and worker module type check are kept, with the updated examples. Should I proceed?

dangkyokhoang: One more thing to note though, that with bundlers, it is expected to have shared inlined in both main and worker by default. Users need nasty hacks on their own to split the shared chunk. Because typically they should be bundled separately.

HarelM: Yes, please proceed with removing the hack and updating the examples.
For more complicated setups people would need to dive into this, but I would assume it's now possible given the shared file you added.

src/util/web_worker.ts
HarelM: I don't know, I don't mind removing it and worst case add it back if someone complains...

HarelM (approved):

HarelM:
Thanks for all the work around this!

HarelM:
Just thought about this, @dangkyokhoang will this affect the examples? Does this PR will break them when published in the docs (docs are using plain HTML imports)?

dangkyokhoang:
> Just thought about this, @dangkyokhoang will this affect the examples? Does this PR will break them when published in the docs (docs are using plain HTML imports)?

If you mean the examples in `/test/examples`, they should work. Both imports in the main module and the worker are relative imports (`import { } from "./maplibre-gl-shared.mjs"`) so they should be resolved correctly.

HarelM:
The examples in the docs import from unpkg CDN, will it work that way? Jsbin too.

dangkyokhoang:
I think yes, they should work. Those CDNs serve with correct CORS so they should be no different from the browser integration tests.

HarelM:
I've tested the following code, and it seems like it's broken due to this change (which is what I assumed would happen)
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <title>Display a map</title>
    <meta property="og:created" content="2025-06-25" />
    <meta charset='utf-8'>
    <link rel='stylesheet' href='https://unpkg.com/maplibre-gl@5.24.0/dist/maplibre-gl.css' />

    <style>
    </style>
</head>
<body>
<div id="map"></div>
<script type="module">
    import * as maplibregl from 'https://unpkg.com/maplibre-gl@6.0.0-15/dist/maplibre-gl-dev.mjs';

        zoom: 1, // starting zoom
        maplibreLogo: true
    });
</script>
</body>
</html>
```
Here's the relevant jsbin:
https://jsbin.com/qinumarofi/edit?html,output

Can you take a look at this?
I would like to keep the jsbin (and other examples in this repo) as is, and fix the loading code to simplify the case for most users, if possible.

dangkyokhoang:
My bad for assuming it works. The worker script is fetched to create a blob url, not imported directly, because the url must be same-origin. Then the fix would be something like:

```ts
if (asModule && isCrossOrigin(url)) {
    const blob = new Blob([`import ${JSON.stringify(new URL(url, import.meta.url).href)}`], {type: 'text/javascript'});
    const blobUrl = URL.createObjectURL(blob);
    return createWorker(blobUrl, asModule);
}
```

HarelM:
Would be great if you can open a PR to solve that.

indus:
This change also breaks my plugin [maplibre-merge-protocol](https://github.com/indus/maplibre-merge-protocol/tree/main). To me it means there is no longer a [(hacky) way to reuse functions/classes](https://github.com/indus/maplibre-merge-protocol/blob/main/src/index.js#L98-L125) from sharedModule with `addProtocol` in the worker thread. Is there a chance we could re-export things like `PBF` and `makeRequest`?

HarelM:
`makeRequest` is exported to the worker (I've added it relatively recently when I needed to run some code in the worker). PBF is a thirdparty library, you could bundle it with the worker code, can't you?

In any case, if you need something exported in the worker let me know, otherwise things will break in the future, as using undocumented features is the recipe to get things breaking :-)

HarelM:
Here's the relevant PR in case you need a pointer:
- https://github.com/onthegomap/maplibre-contour/pull/431

indus:
> Here's the relevant PR in case you need a pointer:
> 
>     * [Add protocol worker onthegomap/maplibre-contour#431](https://github.com/onthegomap/maplibre-contour/pull/431)

Thanks for the hint. I’ve already [fixed](https://github.com/indus/maplibre-merge-protocol/commit/0996073aa39ce9975eb249470df8a89d9ed35354) the maplibre-merge-protocol by bundling [pbf](https://github.com/mapbox/pbf). It’s just a bit unfortunate that the minified size increased from 4 KB to 13 KB just to include a library that is already in the shared bundle. Nevertheless, thank you for pointing this out.

PR #7764, opened 2026-06-09
refactor: remove line-opacity change and refactor the code to be able to more easily accomodate layer-opacity

CommanderStorm:
mostly just removing logic, except in src/webgl/draw/draw_fill.ts where it de-tangles the pass mangement and pass elibility.


src/webgl/program.ts
```diff
@@ -49,7 +49,7 @@ export class Program<Us extends UniformBindings> {
         configuration: ProgramConfiguration,
         fixedUniforms: (b: Context, a: UniformLocations) => Us,
         showOverdrawInspector: boolean,
-        hasTerrain: boolean,
+        useTerrain: boolean,
```
CommanderStorm, line 52: for consistecy, this is the name used elsewere for this

src/webgl/draw/draw_fill.ts
```diff
@@ -20,41 +20,76 @@ import {translatePosition} from '../../util/util.ts';
...
+        return;
+    }
+    if (fillEligibleForOpaque && painter.renderPass === 'translucent') {
+        // Fill already drew in the opaque pass; just draw the outline here.
+        drawOutline(painter, tileManager, layer, coords, renderOptions);
         return;
     }
+    if (painter.renderPass === 'translucent') {
+        drawFillAndOutline(painter, tileManager, layer, coords, renderOptions);
+    }
```
CommanderStorm, line 47: The function split is now better, but still a bit strange.
This is because I need `drawFillAndOutline` callable from inside the bracketing of the `layerOpacitySubpass` closure in the next PR and I would rather not make the closure span the entire function because otherwise this is really a footgun for the next dev as this messes with our early return paradigm

HarelM (approved):
Great! Tanks!

PR #7765, opened 2026-06-09
fix: only remove parse state once parsing is done

### LINKED ISSUE #7752: `map.queryRenderedFeatures` causes "Out of bounds" error
**maplibre-gl-js version**: 6.0.0-12 (also happens on 5.24.0)

**browser**: Chromium 149 (I have also been able to reproduce it in Firefox 151, but much less reliably)

### Steps to Trigger Behavior

 1. Set a "glyphs" URL using `map.setGlyphs`. It does not seem to matter whether the URL points to an actual glyphs source or just 404's.
 3. Add a `geojson` source. At least two features seem to be needed.
 4. Add a `symbol` layer that renders some text as well as multiple different `icon-image`s.
 5. Listen to the `styleimagemissing` event and *asynchronously* add the required images.
 6. Wait for the map to be fully loaded.
 7. Call `map.queryRenderedFeatures` at a point where it should return a feature.
Probably not all of these steps are strictly necessary, but this is the only way I have been able to reproduce the issue.

### Link to Demonstration

[https://jsbin.com/goxoducowo/edit?html,console,output](https://jsbin.com/goxoducowo/edit?html,console,output)

You might need to refresh the page a few times.

### Expected Behavior

The script should print an array of `GeoJSONFeature`s to the console.

### Actual Behavior

Instead, in about around 10% to 50% of the cases, the following error is thrown:
```
Uncaught Error: Out of bounds. Index requested n=0 can't be >= this._numberToString.length 0
    at DictionaryCoder.decode (dictionary_coder.ts:20:53)
    at FeatureIndex.loadMatchingFeature (feature_index.ts:226:55)
    at FeatureIndex.lookupSymbolFeatures (feature_index.ts:297:18)
    at queryRenderedSymbols (query_features.ts:175:54)
    at Style.queryRenderedFeatures (style.ts:1648:17)
    at Map$1.queryRenderedFeatures (map.ts:2009:27)
```

ckolin:
## Launch Checklist

 - [x] Link to related issues.

## Description

This fixes an error that was introduced in #7058 where parse state was prematurely deleted when reloading a previously loaded tile. This caused parts of the raw tile data to be discarded when reloads were fired in quick succession.

I also renamed the method `clearParsing` (which removes a single `Record` entry) to `removeParsing` to match its naming with `removeLoaded` and `clearLoaded` (which replaces the whole `Record` object).

Fixes #7752


ckolin:
I'm not at all familiar with the codebase and pretty much just copied the way it is handled a couple lines above, so please look carefully for any mistakes :)

src/source/vector_tile_worker_source.ts
```diff
@@ -217,8 +217,12 @@ export class VectorTileWorkerSource implements WorkerSource {
 
         if (workerTile.status === 'parsing') {
             // if we are cancelling the original parse, make sure to pass the rawTileData from the original parse
-            const parseState = this.tileState.consumeParsing(uid);
-            return await this._parseWorkerTile(workerTile, params, parseState);
+            const parseState = this.tileState.getParsing(uid);
+            try {
```
HarelM, line 221: Did this try-catch existed in the original code before the refactoring PR that there was a need to add it back?

HarelM:
That that you have played a bit with the code, do you better understand what's causing the issue? Can you create a small unit test to reproduce the issue and make sure it passes before this change?

wayofthefuture:
@ckolin The failing test... then passing test... really helps us to better understand and to ensure this is the right move.

src/source/vector_tile_worker_source.ts
ckolin: `try`/`finally` was not used, but yes, before the refactor, the parse state was only deleted once parsing had finished:

```ts
if (workerTile.status === 'parsing') {
    const result = await workerTile.parse(workerTile.vectorTile, this.layerIndex, this.availableImages, this.actor, params.subdivisionGranularity);
    // ☝️ Parsing finishes here
    let parseResult: WorkerTileResult;
    if (this.fetching[uid]) {
        const {rawTileData, cacheControl, resourceTiming} = this.fetching[uid];
        delete this.fetching[uid]; // 👈️ Parse state deleted here
        parseResult = extend({rawTileData: rawTileData.slice(0), encoding: params.encoding}, result, cacheControl, resourceTiming);
    } else {
        parseResult = result;
    }
    return parseResult;
}
```

I'm using `try`/`finally` because it is used for this purpose in other places in `GeoJSONWorkerSource` and `VectorTileWorkerSource`.

ckolin:
I added two new tests by adapting existing tests with an additional `reloadTile` call. They both failed before the fix, and now pass.

HarelM:
I'm looking at the previous code before the refactor:
```ts
        if (workerTile.status === 'parsing') {
            const result = await workerTile.parse(workerTile.vectorTile, this.layerIndex, this.availableImages, this.actor, params.subdivisionGranularity);
            // if we have cancelled the original parse, make sure to pass the rawTileData from the original fetch
            let parseResult: WorkerTileResult;
            if (this.fetching[uid]) {
                const {rawTileData, cacheControl, resourceTiming} = this.fetching[uid];
                delete this.fetching[uid];
                parseResult = extend({rawTileData: rawTileData.slice(0), encoding: params.encoding}, result, cacheControl, resourceTiming);
            } else {
                parseResult = result;
            }
            return parseResult;

        }
```
And the current code after the fix, and I'm still not sure they are identical in logic:
```ts
        if (workerTile.status === 'parsing') {
            // if we are cancelling the original parse, make sure to pass the rawTileData from the original parse
            const parseState = this.tileState.getParsing(uid);
            try {
                return await this._parseWorkerTile(workerTile, params, parseState);
            } finally {
                this.tileState.removeParsing(uid);
            }
        }
```

This is for the vector source, but since in the past there was inheritance this probably needs to be set to geojson as well...
What do you think?

ckolin:
The difference I can see is that in the new version, the parse state gets read *before* and deleted *after* awaiting the parser, whereas in the old version it was both read and deleted *after* awaiting. I'm not sure if the order matters, as long as the state cannot be deleted before parsing is complete.

As for the `GeoJSONWorkerSource`, I did adjust the logic there as well.

HarelM:
Fair enough. Can you add a changelog entry and I'll merge this assuming it fixes your issue.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 - Build main and worker in same build context to extract shared chunk ([#7745](https://github.com/maplibre/maplibre-gl-js/pull/7745)) (by [@dangkyokhoang](https://github.com/dangkyokhoang))
 
 ### 🐞 Bug fixes
+- Fix conflicting reloads of tiles causing an error in `queryRenderedFeatures` (#7765) (by @ckolin)
```
HarelM: Can you align it to be like the other, with links?

ckolin: Oops sorry, fixed it now!

HarelM:
I'll probably merge this as is, but I'm not sure I understand how this affects query rendered features to be honest. Query rendered features uses the raw data, I wonder if this was returning empty in some cases...

HarelM (approved):

ckolin:
> Query rendered features uses the raw data, I wonder if this was returning empty in some cases...

Yes exactly, for reloads where the state had been prematurely removed, the raw data was not included in the result. The presence of `rawTileData` is also what is checked in the test cases.

wayofthefuture:
Both vector/geojson load/reload now all match which is a nice improvement...

smvjohansenbouvet:
Thank you @ckolin for implementing and @HarelM and @wayofthefuture for following up this fix! We appreciate it a lot :)

PR #7767, opened 2026-06-10
feat: add getViewportTileProgress() API and viewporttilesloaded event

akre54:
## Summary

Adds viewport-specific tile loading visibility: `map.getViewportTileProgress()` API and `viewporttilesloaded` event for detecting when tiles in the current viewport are fully painted.

## Motivation

MapLibre's existing `areTilesLoaded()` checks **all** tiles in a source, not just those visible in the viewport. For video export and headless rendering (e.g., [noodles.gl](https://github.com/joby-aviation/noodles.gl)), this is too coarse — you need to know when viewport tiles specifically are ready.

The `idle` event waits for everything to settle (~300ms debounce) but is too slow for sequential frame capture. This PR adds viewport-scoped tile tracking with immediate event notification, enabling **8.6× faster video export** (308ms→36ms per frame).

## API

\`\`\`typescript
map.getViewportTileProgress(): {
  loaded: number,      // Tiles fully painted in viewport
  loading: number,     // Tiles being fetched/decoded
  total: number,       // Total tiles needed for viewport
  complete: boolean    // All viewport tiles rendered
}

map.on('viewporttilesloaded', (event) => {
  // Fires when all tiles in current viewport are painted
  // Rising-edge only (incomplete → complete transitions)
})
\`\`\`

## Implementation

- Uses existing `_inViewTiles` tracking (no extra per-tile bookkeeping)
- Counts tiles by state: loading vs loaded
- Event fires on rising-edge transitions (doesn't spam when already complete)
- Checks after `painter.render()` completes

## Tests

6 new tests:
- Viewport bounds calculation
- Tile state classification (loading vs loaded)
- Complete=true edge cases
- Event firing during pan/zoom
- Rising-edge semantics

All 1165 UI tests + 185 tile tests pass.

## Usage Example

\`\`\`typescript
// Video export with viewport tile detection
map.on('viewporttilesloaded', () => {
  console.log('Viewport tiles ready');
  captureFrame();
});

// Or poll the state
function checkTiles() {
  const progress = map.getViewportTileProgress();
  console.log(\`\${progress.loaded}/\${progress.total} tiles loaded\`);
  return progress.complete;
}
\`\`\`

## Performance

<0.5ms overhead per frame (tile filtering + state aggregation).

## Related PRs

- #7768: Render timing metadata

## Companion deck.gl PRs

- visgl/deck.gl#10362: Frame ready Promise API
- visgl/deck.gl#10360: Tile loading state for deck layers
- visgl/deck.gl#10361: Frame complete callback with animation state

---

Co-authored-by: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/src/tile/tile_manager.test.ts
+++ b/src/tile/tile_manager.test.ts
@@ -2586,3 +2586,50 @@ describe('TileManager / etag', () => {
         expect(tile.etag).toBe(tileEtag);
     });
 });
+
+describe('TileManager.getViewportTileProgress', () => {
+    test('returns zero counts and complete=true when no tiles are in view', () => {
+        const tileManager = createTileManager();
+        const progress = tileManager.getViewportTileProgress();
+        expect(progress).toEqual({loaded: 0, loading: 0, failed: 0, total: 0, complete: true});
+    });
+
+    test('classifies tiles by state', () => {
+        const tileManager = createTileManager();
+        // Use distinct tile coords at z=3 (8x8 grid) to get unique keys.
+        const states: Array<['loaded' | 'loading' | 'reloading' | 'errored' | 'unloaded' | 'expired', number, number]> = [
+            ['loaded', 0, 0],
+            ['loaded', 1, 0],
+            ['loading', 2, 0],
+            ['reloading', 3, 0],
+            ['errored', 4, 0],
+            ['expired', 5, 0],
+        ];
+        for (const [state, x, y] of states) {
+            const tileID = new OverscaledTileID(3, 0, 3, x, y);
+            const tile = new Tile(tileID, undefined);
+            tile.state = state;
+            tileManager._inViewTiles.setTile(tileID.key, tile);
+        }
+        const progress = tileManager.getViewportTileProgress();
+        // 2 loaded, 1 errored, 3 in-flight (loading, reloading, expired)
+        expect(progress.loaded).toBe(2);
+        expect(progress.failed).toBe(1);
+        expect(progress.loading).toBe(3);
+        expect(progress.total).toBe(6);
+        expect(progress.complete).toBe(false);
+    });
+
+    test('complete=true when only loaded and errored tiles remain', () => {
+        const tileManager = createTileManager();
+        const aID = new OverscaledTileID(1, 0, 1, 0, 0);
+        const bID = new OverscaledTileID(1, 0, 1, 1, 0);
+        const a = new Tile(aID, undefined); a.state = 'loaded';
+        const b = new Tile(bID, undefined); b.state = 'errored';
+        tileManager._inViewTiles.setTile(aID.key, a);
+        tileManager._inViewTiles.setTile(bID.key, b);
+        const progress = tileManager.getViewportTileProgress();
+        expect(progress.complete).toBe(true);
+        expect(progress.loading).toBe(0);
+    });
+});
--- a/src/tile/tile_manager.ts
+++ b/src/tile/tile_manager.ts
@@ -991,4 +991,35 @@ export class TileManager extends Evented {
         }
         return true;
     }
+
+    /**
+     * Returns counts of tiles in the current viewport by their loading state.
+     * Unlike {@link areTilesLoaded}, this also exposes loading/failed counts so
+     * callers (e.g. video export pipelines) can wait on viewport completion with
+     * fine-grained visibility, without inspecting private fields.
+     */
+    getViewportTileProgress(): {loaded: number; loading: number; failed: number; total: number; complete: boolean} {
+        let loaded = 0;
+        let loading = 0;
+        let failed = 0;
+        let total = 0;
+        for (const tile of this._inViewTiles.getAllTiles()) {
+            total++;
+            if (tile.state === 'loaded') {
+                loaded++;
+            } else if (tile.state === 'errored') {
+                failed++;
+            } else {
+                // 'loading', 'reloading', 'unloaded', or 'expired' — still in flight
+                loading++;
+            }
+        }
+        return {
+            loaded,
+            loading,
+            failed,
+            total,
+            complete: loading === 0,
+        };
+    }
 }
--- a/src/ui/events.ts
+++ b/src/ui/events.ts
@@ -193,6 +193,17 @@ export type MapEventType = {
      */
     resize: MapLibreEvent;
     /**
+     * Fired on the rising edge of "all viewport tiles loaded": when, after
+     * one or more tiles in the current viewport were pending, all of those
+     * tiles have transitioned to a loaded or errored state and the viewport
+     * is fully painted. This is finer-grained than `idle` (which also waits on
+     * label fades, placement, and animations) and intended for capture/export
+     * pipelines that need a low-latency frame-ready signal.
+     *
+     * Use {@link Map.getViewportTileProgress} to read counts at any time.
+     */
+    viewporttilesloaded: MapLibreEvent;
+    /**
      * Fired when the WebGL context is lost.
      */
     webglcontextlost: MapContextEvent;
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -585,6 +585,9 @@ export class Map extends Camera {
 
     _loaded: boolean;
     _idleTriggered = false;
+    // tracks whether the most recent emitted viewport-tile state was 'complete',
+    // so we only fire viewporttilesloaded on rising-edge transitions
+    _viewportTilesComplete = true;
     // accounts for placement finishing as well
     _fullyLoaded: boolean;
     _trackResize: boolean;
@@ -2440,6 +2443,37 @@ export class Map extends Camera {
     }
 
     /**
+     * Returns aggregate counts of tiles needed for the current viewport, broken
+     * down by loading state. This complements {@link Map.areTilesLoaded} for use
+     * cases (e.g. headless video export, screenshot capture) that need fine-grained
+     * visibility into how many tiles are still in flight versus painted.
+     *
+     * @returns Counts of viewport tiles by state, plus a `complete` flag that is
+     * true when no tiles remain pending. When the style has no tile sources,
+     * `total` is 0 and `complete` is true.
+     * @example
+     * ```ts
+     * const progress = map.getViewportTileProgress();
+     * console.log(`${progress.loaded}/${progress.total} viewport tiles loaded`);
+     * if (progress.complete) capture();
+     * ```
+     */
+    getViewportTileProgress(): {loaded: number; loading: number; failed: number; total: number; complete: boolean} {
+        const result = {loaded: 0, loading: 0, failed: 0, total: 0, complete: true};
+        const tileManagers = this.style?.tileManagers;
+        if (!tileManagers) return result;
+        for (const tileManager of Object.values(tileManagers)) {
+            const sub = tileManager.getViewportTileProgress();
+            result.loaded += sub.loaded;
+            result.loading += sub.loading;
+            result.failed += sub.failed;
+            result.total += sub.total;
+        }
+        result.complete = result.loading === 0;
+        return result;
+    }
+
+    /**
      * Removes a source from the map's style.
      *
      * @param id - The ID of the source to remove.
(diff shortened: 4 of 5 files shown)


HarelM:
CC: @CommanderStorm in regards to performance matrices (although the render event PR is more "interesting").

What's the worst case loop for in view tiles? How many tiles can this loop run on for every frame?
Should this be an opt-in behavior?

akre54:
Yeah it maybe could be. It looks like perf is pretty negligible (it's a `keys` plus a loop) but opt in is fine by me if you prefer

HarelM:
Let me sleep on it. I'll keep you posted. Feel free to post in our slack channel to get more input.

akre54:
Good idea, thanks! [Posted](https://osmus.slack.com/archives/C03TFH5NE83/p1781074523329869)

HarelM:
CC: @bjperson who's maintaining a video export plugin:
https://github.com/bjperson/maplibre-gl-video-export

CommanderStorm:
I am honestly not entirely clear why our "load" and "idle" would wait for 300ms..
There must be something that I am missing.

Just a hunch:
Is the thing you are measuring here maybe our https://maplibre.org/maplibre-style-spec/transition/ support?
So essentially, you want to have an event like `idle`, but before we start transitioning?

If you want that, the simplest way is to just disable transitions in the style given to you.
Or we could make on-load transitions happen better. I am not entirely sure that our current default here is 100% the best, but I also have not entirely used that system to stlye things.

akre54:
> CC: @bjperson who's maintaining a video export plugin: https://github.com/bjperson/maplibre-gl-video-export

This is cool! Thanks for sharing. 

Looks like they're [using](https://github.com/bjperson/maplibre-gl-video-export/blob/v0.1.1/src/index.js#L6707-L6732) `areTilesLoaded` plus a [`this._map.once('render', resolve)`](https://github.com/bjperson/maplibre-gl-video-export/blob/v0.1.1/src/index.js#L6793). They poll `areTilesLoaded()` manually. They're also using `maplibregl.setNow()` to freeze time during export which is interesting.

> I am honestly not entirely clear why our "load" and "idle" would wait for 300ms..

Yeah it looks like it's the `fadeDuration` default, and symbol collision/placement takes time to settle. For history, `idle` was added in ff5eae5f9.

The real issue is that `idle` conflates "tiles ready" with "animations complete". 

`idle` only fires when `somethingDirty === false`. `_placementDirty` stays true as long as `placement.hasTransitions()` returns true. `hasTransitions()` returns `true` for `fadeDuration` milliseconds after the last placement change.

> Or we could make on-load transitions happen better

Yes, setting `fadeDuration: 0` would make `idle` fire faster, but it's indirectly manipulating animation timing to get a tile-readiness signal. 

What we actually need for video export is an "Are viewport tiles loaded?" signal (orthogonal to symbol animation timing)

The `viewporttilesloaded` event in this PR provides this directly. It fires when viewport tiles are painted, whether fadeDuration is 0ms or 300ms. It's the right primitive rather than a workaround.

bjperson:
Maintainer of maplibre-gl-video-export here. The plugin currently busy-waits around areTilesLoaded() (20 ms × 5 polling loop) because there's no rising-edge event — viewporttilesloaded would let us drop that workaround entirely. 

+1 on @HarelM's opt-in, ideally as a lazy gate on this.listens('viewporttilesloaded') so subscribers auto-arm the per-frame check.

chrisgervang:
> Yes, setting fadeDuration: 0 would make idle fire faster, but it's indirectly manipulating animation timing to get a tile-readiness signal.
> 
> What we actually need for video export is an "Are viewport tiles loaded?" signal (orthogonal to symbol animation timing)
> 
> The viewporttilesloaded event in this PR provides this directly. It fires when viewport tiles are painted, whether fadeDuration is 0ms or 300ms. It's the right primitive rather than a workaround.

Agreed on adding `viewporttilesloaded`. 

It seems like timeline-based video export applications should also set `fadeDuration: 0` to ensure symbols are always visible when they're expected to be. Usually animators would explicitly keyframe opacity rather than defer to an animation engine they can't control.

src/tile/tile_manager.ts
```diff
@@ -991,4 +991,35 @@ export class TileManager extends Evented {
...
+                loading++;
+            }
+        }
+        return {
+            loaded,
+            loading,
+            failed,
+            total,
+            complete: loading === 0,
+        };
```
CommanderStorm, line 1023: I have a few questions.

- why is `unloaded'`, or `'expired'` in the loading bucket.
  Please re-review the code you generated here
- complete and total are currently derivable from the others. I would like to not have states like this in our api, so please remove those 😉

src/tile/tile_manager.test.ts
```diff
@@ -2586,3 +2586,50 @@ describe('TileManager / etag', () => {
...
+            tile.state = state;
+            tileManager._inViewTiles.setTile(tileID.key, tile);
+        }
+        const progress = tileManager.getViewportTileProgress();
+        // 2 loaded, 1 errored, 3 in-flight (loading, reloading, expired)
+        expect(progress.loaded).toBe(2);
+        expect(progress.failed).toBe(1);
+        expect(progress.loading).toBe(3);
+        expect(progress.total).toBe(6);
+        expect(progress.complete).toBe(false);
```
CommanderStorm, line 2620: please assert on the full api shape instead, so `expect(progress).toBe...` and remove comments that state the code below, you are not an LLM..

src/tile/tile_manager.test.ts
```diff
@@ -2586,3 +2586,50 @@ describe('TileManager / etag', () => {
...
+        // Use distinct tile coords at z=3 (8x8 grid) to get unique keys.
+        const states: Array<['loaded' | 'loading' | 'reloading' | 'errored' | 'unloaded' | 'expired', number, number]> = [
+            ['loaded', 0, 0],
+            ['loaded', 1, 0],
+            ['loading', 2, 0],
+            ['reloading', 3, 0],
+            ['errored', 4, 0],
+            ['expired', 5, 0],
+        ];
+        for (const [state, x, y] of states) {
```
CommanderStorm, line 2608: x,y are trivialy derivable from array index or constant. please simplify

src/ui/map_tests/map_basic.test.ts
```diff
@@ -143,6 +143,63 @@ describe('Map', () => {
...
+            tile.state = 'loading';
+            map.style.tileManagers.geojson._inViewTiles.setTile(fakeTileId.key, tile);
+
+            // Force one render with a pending tile so the next completion is
+            // a rising-edge transition.
+            map.redraw();
+            expect(map.areTilesLoaded()).toBe(false);
+
+            const spy = vi.fn();
+            map.on('viewporttilesloaded', spy);
```
CommanderStorm, line 192: Please set this spy up all the way at the start so that we are actually testing the invariant that you care about.

Please also test that this is not called too early and that also that symbol placement (which you sad you want to happen after you get the event)  and transitions happen after.
Currently I don't think you have this invariant.

src/ui/map_tests/map_basic.test.ts
```diff
@@ -143,6 +143,63 @@ describe('Map', () => {
...
+            await map.once('load');
+
+            map.addSource('geojson', createStyleSource());
+            const fakeTileId = new OverscaledTileID(0, 0, 0, 0, 0);
+            const tile = new Tile(fakeTileId, undefined);
+            tile.state = 'loading';
+            map.style.tileManagers.geojson._inViewTiles.setTile(fakeTileId.key, tile);
+
+            // Force one render with a pending tile so the next completion is
+            // a rising-edge transition.
```
CommanderStorm, line 187: What is a rising edge transition?
Can we simplify the language here?

src/ui/events.ts
```diff
@@ -192,6 +192,17 @@ export type MapEventType = {
...
+     * Fired on the rising edge of "all viewport tiles loaded": when, after
+     * one or more tiles in the current viewport were pending, all of those
+     * tiles have transitioned to a loaded or errored state and the viewport
+     * is fully painted. This is finer-grained than `idle` (which also waits on
+     * label fades, placement, and animations) and intended for capture/export
+     * pipelines that need a low-latency frame-ready signal.
+     *
+     * Use {@link Map.getViewportTileProgress} to read counts at any time.
+     */
+    viewporttilesloaded: MapLibreEvent;
```
CommanderStorm, line 205: Lets go back to the basics.
I want to know more if this is the right place.

I want to make sure that we are aligned 100%, so that we don't add events that would not be at the exact place that they need to.
Why do you want to get called right here?

Why not after label placement?
For your usecase, you can control timing, so "label fades" (please call them transtions) and animations don't matter I think.

Furthermore, what is up with the viewport naming?

src/ui/map.ts
```diff
@@ -2439,6 +2442,37 @@ export class Map extends Camera {
...
+     * visibility into how many tiles are still in flight versus painted.
+     *
+     * @returns Counts of viewport tiles by state, plus a `complete` flag that is
+     * true when no tiles remain pending. When the style has no tile sources,
+     * `total` is 0 and `complete` is true.
+     * @example
+     * ```ts
+     * const progress = map.getViewportTileProgress();
+     * console.log(`${progress.loaded}/${progress.total} viewport tiles loaded`);
+     * if (progress.complete) capture();
```
CommanderStorm, line 2458: Why do you want `getViewportTileProgress` in practice?
Would `await map.once("viewporttitlesload")` not be the same thing? what is the difference and why do you want both?

src/ui/map.ts
```diff
@@ -3703,6 +3737,18 @@ export class Map extends Camera {
...
+        // tiles loaded". This gives capture/export pipelines a low-latency signal
+        // that the visible tile pyramid has been painted, without waiting on the
+        // ~300ms `idle` debounce (label fades, placement, etc.).
+        const tilesComplete = this.areTilesLoaded();
+        if (tilesComplete && !this._viewportTilesComplete) {
+            this._viewportTilesComplete = true;
+            this.fire(new Event('viewporttilesloaded'));
+        } else if (!tilesComplete) {
+            this._viewportTilesComplete = false;
+        }
```
CommanderStorm, line 3750: I am confused. Why is this named `viewport..` again?
We only load tiles in the viewport, so the destinction is not entirely relevant, or am I missing something?

This is essentially the same code that @bjperson already said.
This is also "buisy-waits" on `areTilesLoaded`..
-> I don't think this would be actually more efficient, or am I missing something?

CommanderStorm (commented):
I looked at the code.
It is a bit messy and I am not entirely sure if this is actually solving what you want.

CommanderStorm:
> > Yes, setting fadeDuration: 0 would make idle fire faster, but it's indirectly manipulating animation timing to get a tile-readiness signal.
> > What we actually need for video export is an "Are viewport tiles loaded?" signal (orthogonal to symbol animation timing)
> > The viewporttilesloaded event in this PR provides this directly. It fires when viewport tiles are painted, whether fadeDuration is 0ms or 300ms. It's the right primitive rather than a workaround.
> 
> Agreed on adding `viewporttilesloaded`.
> 
> It seems like timeline-based video export applications should also set `fadeDuration: 0` to ensure symbols are always visible when they're expected to be. Usually animators would explicitly keyframe opacity rather than defer to an animation engine they can't control.

I am trying to understand this more.
That seems contraticting.

The 300ms IS our `transition` and with the paused time that you as a video export software likely want to use anyhow, would this be different?

Could it maybe be the case that pausing the time does not affect transitions and only motion, or why are we talking about this?

CommanderStorm:
I think the main issue that I don't have clarity on is
- "why do you two want to be notified on the first render after tile load and before placement?" and
- why do you want transitions to be on and why does `setNow` not solve this case for you?
- why does the following code not do this for you?
  
  ```js
  function onFirstRenderWithMaybeTilesLoaded {
    if map.areTilesLoaded() {
      map.off("render", onFirstRenderWithMaybeTilesLoaded)
      doTheThingYouWantToDo()
    }
  }
  
  map.on("render", onFirstRenderWithMaybeTilesLoaded)
  ```
  
  This is essentially the code that this PR proposes, unless I am missing something.

akre54:
@CommanderStorm you were absolutely right. I ran some real E2E browser tests and looks like `fadeDuration: 0` + render events (and `areTilesLoaded()`) solves the perf issue. 

Thanks for pushing us to measure accurately. Closing this PR.

PR #7768, opened 2026-06-10
feat: attach CPU timing metadata to render event

akre54:
## Summary

Adds CPU timing metadata to the `render` event for frame performance monitoring and capture timing decisions.

## Motivation

The `render` event fires after each paint but provides no timing information. For video export and performance analysis (e.g., [noodles.gl](https://github.com/joby-aviation/noodles.gl)), knowing render duration helps:

1. Distinguish trivial updates (style changes) from substantial renders (tile painting)
2. Detect performance bottlenecks
3. Estimate GPU buffer swap timing

## API

\`\`\`typescript
map.on('render', (event: MapLibreRenderEvent) => {
  // New timing property
  event.timing: {
    start: number,      // performance.now() when render started
    end: number,        // performance.now() after painter.render()
    duration: number    // end - start (ms)
  }
})
\`\`\`

All timestamps are `DOMHighResTimeStamp` (milliseconds since page navigation).

**Note**: These are CPU-side timestamps. WebGL exposes no portable signal for GPU buffer-swap completion. For true frame readiness, also listen for `viewporttilesloaded` (#7767) or use `canvas.captureStream(0)`.

## Implementation

- Brackets `painter.render()` with `performance.now()` calls
- Adds `MapLibreRenderTiming` and `MapLibreRenderEvent` types (exported from public API)
- Timing field always populated (stable event shape for TypeScript consumers)
- Cost: two `performance.now()` calls per frame (~0.0003% of 16.67ms frame time)

## Tests

1 new test verifying:
- Timing property exists
- end ≥ start
- duration matches bracket
- Reasonable values for synthetic test

All 1163 UI tests pass.

## Usage Example

\`\`\`typescript
map.on('render', (event) => {
  const { start, end, duration } = event.timing;
  console.log(\`Render: \${duration.toFixed(1)}ms\`);
  
  // Detect substantial renders
  if (duration > 10) {
    console.log('Heavy render - likely tile painting');
  }
});
\`\`\`

## Performance

~47ns overhead per frame (two `performance.now()` calls) — 0.0003% of 16.67ms frame budget.

## Breaking Changes

None. `MapEventType.render` type narrowed from `MapLibreEvent` to `MapLibreRenderEvent` (structural superset, backward compatible).

## Related PRs

- #7767: Viewport tile progress API

## Companion deck.gl PRs

- visgl/deck.gl#10362: Frame ready Promise API
- visgl/deck.gl#10360: Tile loading state
- visgl/deck.gl#10361: Frame complete callback with GPU timing

---

Co-authored-by: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/src/ui/events.ts
+++ b/src/ui/events.ts
@@ -186,8 +186,11 @@ export type MapEventType = {
      * - a change to the map's style
      * - a change to a GeoJSON source
      * - the loading of a vector tile, GeoJSON file, glyph, or sprite
+     *
+     * The event carries a `timing` object with high-resolution CPU-side timing
+     * data; see {@link MapLibreRenderEvent}.
      */
-    render: MapLibreEvent;
+    render: MapLibreRenderEvent;
     /**
      * Fired immediately after the map has been resized.
      */
@@ -438,6 +441,40 @@ export type MapLibreEvent<TOrig = unknown> = {
     type: keyof MapEventType | keyof MapLayerEventType;
     target: Map;
     originalEvent: TOrig;
+};
+
+/**
+ * Timing metadata attached to the `render` event. All timestamps are in
+ * `performance.now()`-space (DOMHighResTimeStamp, milliseconds since the
+ * page navigation timeline origin).
+ *
+ * Note: these are CPU-side timestamps. WebGL exposes no portable signal for
+ * GPU buffer-swap completion, so callers needing true frame readiness for
+ * capture/export should additionally listen for `viewporttilesloaded` (when
+ * available) or sample `canvas.captureStream(0)`.
+ *
+ * @group Event Related
+ */
+export type MapLibreRenderTiming = {
+    /** When `painter.render()` was entered. */
+    renderStart: number;
+    /**
+     * When `painter.render()` returned and all WebGL commands for the frame
+     * were submitted to the driver. Equivalent to `renderStart + renderDuration`.
+     */
+    commandsSubmitted: number;
+    /** Wall-clock time spent inside `painter.render()`, in milliseconds. */
+    renderDuration: number;
+};
+
+/**
+ * The render event. Carries CPU-side timing information for the just-completed
+ * frame (see {@link MapLibreRenderTiming}).
+ *
+ * @group Event Related
+ */
+export type MapLibreRenderEvent = MapLibreEvent & {
+    timing: MapLibreRenderTiming;
 };
 
 /**
--- a/src/ui/events.ts
+++ b/src/ui/events.ts
@@ -457,14 +457,14 @@ export type MapLibreEvent<TOrig = unknown> = {
  */
 export type MapLibreRenderTiming = {
     /** When `painter.render()` was entered. */
-    renderStart: number;
+    start: number;
     /**
      * When `painter.render()` returned and all WebGL commands for the frame
-     * were submitted to the driver. Equivalent to `renderStart + renderDuration`.
+     * were submitted to the driver. Equivalent to `start + duration`.
      */
-    commandsSubmitted: number;
+    end: number;
     /** Wall-clock time spent inside `painter.render()`, in milliseconds. */
-    renderDuration: number;
+    duration: number;
 };
 
 /**
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -3689,7 +3689,11 @@ export class Map extends Camera {
 
         this._placementDirty = this.style?._updatePlacement(this.transform, this.showCollisionBoxes, fadeDuration, this._crossSourceCollisions, globeRenderingChanged);
 
-        // Actually draw
+        // Actually draw. Bracket painter.render() with high-resolution timestamps
+        // so we can attach CPU-side timing metadata to the `render` event below.
+        // The cost is two performance.now() calls per frame; populating the field
+        // unconditionally keeps the event shape stable for typed consumers.
+        const renderStart = performance.now();
         this.painter.render(this.style, {
             showTileBoundaries: this.showTileBoundaries,
             showOverdrawInspector: this._showOverdrawInspector,
@@ -3700,8 +3704,15 @@ export class Map extends Camera {
             showPadding: this.showPadding,
             anisotropicFilterPitch: this.getAnisotropicFilterPitch(),
         });
+        const renderEnd = performance.now();
 
-        this.fire(new Event('render'));
+        this.fire(new Event('render', {
+            timing: {
+                renderStart,
+                commandsSubmitted: renderEnd,
+                renderDuration: renderEnd - renderStart,
+            },
+        }));
 
         if (this.loaded() && !this._loaded) {
             this._loaded = true;
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -3621,6 +3621,7 @@ export class Map extends Camera {
      * @param paintStartTimeStamp - The time when the animation frame began executing.
      */
     _render(paintStartTimeStamp: number): this {
+        const renderStart = performance.now();
         const fadeDuration = this._idleTriggered ? this._fadeDuration : 0;
 
         const isGlobeRendering = this.style.projection?.transitionState > 0;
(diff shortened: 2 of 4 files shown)


akre54:
Pushed a follow-up commit (049e3d4) that re-exports `MapLibreRenderEvent` and `MapLibreRenderTiming` from `src/index.ts`.

The types were defined and `export type`d from `src/ui/events.ts` in the original commit, but I missed adding them to the public API barrel. As a result:

- TypeScript users couldn't `import type { MapLibreRenderEvent } from 'maplibre-gl'` when annotating their render listeners.
- typedoc emitted a "referenced but not included" warning, which fails the docs build (`typedoc.json` has `treatWarningsAsErrors: true`).

Verified locally: `npm run generate-docs` is now warning-free and produces `docs/API/type-aliases/MapLibreRenderEvent.md` + `MapLibreRenderTiming.md`. `npm run typecheck` is also clean.

src/ui/map.ts
```diff
@@ -3689,7 +3689,11 @@ export class Map extends Camera {
 
         this._placementDirty = this.style?._updatePlacement(this.transform, this.showCollisionBoxes, fadeDuration, this._crossSourceCollisions, globeRenderingChanged);
 
-        // Actually draw
+        // Actually draw. Bracket painter.render() with high-resolution timestamps
```
HarelM: I think this comment is a bit too verbose... Consider removing it.

src/ui/map.ts
```diff
@@ -3689,7 +3689,11 @@ export class Map extends Camera {
...
+        // so we can attach CPU-side timing metadata to the `render` event below.
+        // The cost is two performance.now() calls per frame; populating the field
+        // unconditionally keeps the event shape stable for typed consumers.
+        const renderStart = performance.now();
```
HarelM: Why was this added here and not at the begging of this method?

src/ui/events.ts
```diff
@@ -440,6 +443,40 @@ export type MapLibreEvent<TOrig = unknown> = {
...
+ * page navigation timeline origin).
+ *
+ * Note: these are CPU-side timestamps. WebGL exposes no portable signal for
+ * GPU buffer-swap completion, so callers needing true frame readiness for
+ * capture/export should additionally listen for `viewporttilesloaded` (when
+ * available) or sample `canvas.captureStream(0)`.
+ *
+ * @group Event Related
+ */
+export type MapLibreRenderTiming = {
```
HarelM, line 458: As you said, I think start/end/duration is better

src/ui/map_tests/map_render.test.ts
```diff
@@ -89,3 +89,23 @@ test('redraw', async () => {
...
+test('render event carries timing metadata', async () => {
+    const map = createMap();
+    await map.once('idle');
+
+    const renderPromise = map.once('render');
+    map.redraw();
+    const event = await renderPromise;
+
+    expect(event.timing).toBeDefined();
+    expect(typeof event.timing.renderStart).toBe('number');
```
HarelM: I dont think these type e check are interesting...

src/ui/map_tests/map_render.test.ts
```diff
@@ -89,3 +89,23 @@ test('redraw', async () => {
...
+    expect(typeof event.timing.commandsSubmitted).toBe('number');
+    expect(typeof event.timing.renderDuration).toBe('number');
+    // Sanity: end is at or after start, duration matches the bracket.
+    expect(event.timing.commandsSubmitted).toBeGreaterThanOrEqual(event.timing.renderStart);
+    expect(event.timing.renderDuration).toBeCloseTo(event.timing.commandsSubmitted - event.timing.renderStart, 5);
+    // Duration should be non-negative and reasonable for a synthetic test (well under a second).
```
HarelM: Remove comments...

HarelM:
Thanks for this PR! Added mostly superficial comments.

akre54:
Done! Thanks for the quick review

HarelM:
Not that I think the code added here is problematic, but I still need to ask, can't you just calculate this by listening to the render event right now?

akre54:
Good question! The issue is that the `render` event fires **after** `painter.render()` completes, so by the time your listener receives it, you've already missed the start time.

Users would need to hook **before** the render starts and **after** it completes. Without this PR, the options are:

1. **Monkey-patch `_render()`** - fragile, breaks on internal changes
2. **Wrap the Map instance** - same issues, plus TypeScript complications
3. **Use two separate events** - no existing "beforeRender" event

This PR brackets `painter.render()` from inside `_render()` where we have access to both the start and end points.

**Alternative we considered**: Add separate `beforeRender` and `afterRender` events, but:
- That's two events per frame instead of enriching the existing `render` event
- Users would need to correlate pairs of events (race conditions, dropped events)
- The timing field keeps the `render` event self-contained

The cost is minimal (two `performance.now()` calls) and the timing data is useful beyond just video export - performance monitoring, frame budget tracking, etc.

Does that address the concern, or would you prefer a different approach?

akre54:
The AI got a little overeager here (I didn't ask for it to post as me). But mostly I'm just looking for ways of speeding up Noodles rendering: https://github.com/joby-aviation/noodles.gl/pull/452

HarelM:
But aren't render events fired one after the other? Can't you measure "time since last render event"? Would that give different results?

akre54:
They're slightly different.

The interval includes:

* Previous render work
* Idle time (waiting for RAF, other JS execution)
* Current render work

For video export, we care about how long rendering took (is this a heavy tile paint vs a quick style update?), not how long since the last frame.

The interval timing would give framerate data, not render performance data. Different use case.

src/ui/map_tests/map_render.test.ts
```diff
@@ -89,3 +89,18 @@ test('redraw', async () => {
...
+    const map = createMap();
+    await map.once('idle');
+
+    const renderPromise = map.once('render');
+    map.redraw();
+    const event = await renderPromise;
+
+    expect(event.timing).toBeDefined();
+    expect(event.timing.end).toBeGreaterThanOrEqual(event.timing.start);
+    expect(event.timing.duration).toBeCloseTo(event.timing.end - event.timing.start, 5);
```
HarelM, line 103: Will this check survive CI flackiness?

src/ui/events.ts
```diff
@@ -440,6 +443,40 @@ export type MapLibreEvent<TOrig = unknown> = {
...
+    duration: number;
+};
+
+/**
+ * The render event. Carries CPU-side timing information for the just-completed
+ * frame (see {@link MapLibreRenderTiming}).
+ *
+ * @group Event Related
+ */
+export type MapLibreRenderEvent = MapLibreEvent & {
```
HarelM, line 476: Man.. this is a mess, I now see that some events are classes and some are types, I'm not sure I even know why...
My original thought was that this event should have the timing as it's properties instead of a nested obejct, but then I got confused with the different event implementations... :-/
I don't know which is better to be honest...

CommanderStorm:
> For video export, we care about how long rendering took (is this a heavy tile paint vs a quick style update?), not how long since the last frame.
> 
> The interval timing would give framerate data, not render performance data. Different use case.

Really stupid question, but what is difference in your context and why do you care about that difference?

You know that you can control our render timing (so what time in the animation we render at) That API is relatively new..
What you are trying to do sounds suspiciously similar.

akre54:
This PR to me is more about general instrumentation, less about the video use case specifically. If it's at all controversial I'm fine dropping

src/ui/events.ts
akre54: What's the benefit of a class over a nested object?

CommanderStorm: you can `instanceof` and can inherit.

CommanderStorm:
I would be fine with including that part of the timing data if I were to know why you would want this.
Frame timing can be taken by listening on `render` and measuring the time in betweeen...

https://maplibre.org/maplibre-gl-js/docs/examples/display-performance-metrics/

akre54:
@HarelM @CommanderStorm valid feedback. Closing this PR. Thanks for the review.

PR #7776, opened 2026-06-12
feat(map): expose original style URL via map.getStyleUrl() (closes #7109)

### LINKED ISSUE #7109: Feature: Expose the original style URL via map.getStyleUrl()

shreeharshshinde:
## Summary

Implements the feature requested in #7109: exposes the original style URL used to initialize or update the map via a new `map.getStyleUrl()` method, so plugin developers can distinguish basemap layers from dynamically added layers.

## Proposed API

```ts
// Initialize with a URL
const map = new Map({ container: 'map', style: 'https://example.com/style.json' });
map.getStyleUrl(); // 'https://example.com/style.json'

// After setStyle() with a URL, updates accordingly
map.setStyle('https://demotiles.maplibre.org/style.json');
map.getStyleUrl(); // 'https://demotiles.maplibre.org/style.json'

// If style was set as an object, returns null
map.setStyle({ version: 8, sources: {}, layers: [] });
map.getStyleUrl(); // null
```

### Architecture
URL tracking lives in the **`Style` class** (not `Map`), keeping the concern co-located with loading logic. `Map.getStyleUrl()` delegates cleanly via `this.style?.getStyleUrl() ?? null`.

### Files changed
| File | Change |
|---|---|
| `src/style/style.ts` | Added `_styleUrl: string \| null` field; set in `loadURL()`, `loadJSON()`, `loadEmpty()`; added `Style.getStyleUrl()`; `setState()` reads `options.sourceUrl` to update tracking during diffs |
| `src/ui/map.ts` | Added `Map.getStyleUrl()` (delegates to `Style`); threaded `sourceUrl` through the diff path via `options` object |
| `src/ui/map_tests/map_style.test.ts` | 7 new unit tests covering all edge cases |

### Key design decisions

1. **`sourceUrl` in options, not a separate param** — Per reviewer feedback on the existing community PR (#7110), `sourceUrl` is added to `StyleSwapOptions` instead of being threaded as a standalone parameter through `_updateDiff` → `_updateStyle` → `setState` → `loadJSON`. This avoids signature pollution across internal methods.

2. **`?.` operator** — `Map.getStyleUrl()` uses `this.style?.getStyleUrl() ?? null` (safe optional chaining, no explicit `if (this.style)` guard block).

3. **Full diff path coverage** — When `setStyle(url)` triggers a diff (not a full reload), `_diffStyle` spreads `sourceUrl` into options before calling `_updateDiff`, so `setState` correctly updates `_styleUrl` in both the full-reload and diff paths.

4. **Returns `null` (not `undefined`)** — Consistent with the `Style` class API and the existing community PR convention.

## Tests

7 unit tests added in `map_style.test.ts`:
- Returns `null` when no style is set
- Returns `null` when initialized with a style object
- Returns the URL when initialized with a URL string
- Returns the URL after `setStyle(url)` call
- Returns `null` after `setStyle(object)` call
- Returns `null` after `setStyle(null)` call
- Updates correctly across multiple `setStyle(url)` calls

## Checklist
- [x] Feature implementation complete
- [x] Unit tests added (7 tests, all cases covered)
- [ ] TypeScript compiles cleanly (`tsc --noEmit`)
- [x] Reviewer feedback addressed (`?.` operator, `sourceUrl` in options object)
- [ ] Closes #7109

### DIFF
--- a/src/style/style.ts
+++ b/src/style/style.ts
@@ -196,6 +196,12 @@ export type StyleSwapOptions = {
      * that allows to modify a style after it is fetched but before it is committed to the map state. Refer to {@link TransformStyleFunction}.
      */
     transformStyle?: TransformStyleFunction;
+    /**
+     * The URL of the style source, used to track the original style URL
+     * when the style is loaded from a JSON object via diffing or fallback.
+     * This is used internally by {@link Map.getStyleUrl}.
+     */
+    sourceUrl?: string;
 };
 
 /**
@@ -458,8 +464,8 @@ export class Style extends Evented {
         }
     }
 
-    loadJSON(json: StyleSpecification, options: StyleSetterOptions & StyleSwapOptions = {}, previousStyle?: StyleSpecification, sourceUrl?: string): void {
-        this._styleUrl = sourceUrl ?? null;
+    loadJSON(json: StyleSpecification, options: StyleSetterOptions & StyleSwapOptions = {}, previousStyle?: StyleSpecification): void {
+        this._styleUrl = options.sourceUrl ?? null;
         this.fire(new Event('dataloading', {dataType: 'style'}));
 
         this._frameRequest = new AbortController();
@@ -869,7 +875,7 @@ export class Style extends Evented {
      *
      * @returns true if any changes were made; false otherwise
      */
-    setState(nextState: StyleSpecification, options: StyleSwapOptions & StyleSetterOptions = {}, sourceUrl?: string): boolean {
+    setState(nextState: StyleSpecification, options: StyleSwapOptions & StyleSetterOptions = {}): boolean {
         this._checkLoaded();
 
         const serializedStyle =  this.serialize();
@@ -889,7 +895,7 @@ export class Style extends Evented {
 
         if (operations.operations.length === 0) {
             // Update styleUrl even when there are no diff operations
-            this._styleUrl = sourceUrl ?? null;
+            this._styleUrl = options.sourceUrl ?? null;
             return false;
         }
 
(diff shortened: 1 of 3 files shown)


HarelM:
Make sure to fill the PR checklist, closing since this violates our AI policy.

shreeharshshinde:
@HarelM, just read the policy and will keep in mind further ...

PR #7778, opened 2026-06-14
Add example: FPS-style first-person movement over a map

clement-igonet:
## A first-person, playable view built only on MapLibre's public API

- 🎮 **A gaming-grade first-person view.** MapLibre doing something people assume needs a custom WebGL layer or a fork — it's just `jumpTo` driven once per frame.
- 🔫 **FPS-oriented controls** out of the box: WASD/arrows + mouse pointer-lock on desktop, on-screen joystick + action button on touch.
- 📐 **A true 90° level horizon — the non-obvious centerpiece.** Naively setting `pitch: 90` collapses the camera onto the ground. The trick: at 90° the eye and the look-at **center** share altitude, so you place the center at eye height a fixed distance ahead at a fixed zoom (`setCenterClampedToGround(false)` + `elevation`). Hard to discover, documented nowhere — valuable as a runnable example.

## Launch Checklist

 - [x] Confirm **your changes do not include backports from Mapbox projects** — original code, no backports.
 - [x] Link to related issues — N/A (new standalone docs example).
 - [x] Include before/after visuals or gifs — gallery thumbnail at `docs/assets/examples/walk-around-a-map-in-first-person.png`.
 - [x] Write tests for all new functionality — covered by the existing `test/examples` validation; no new library functionality.
 - [x] Document any changes to public APIs — none.
 - [x] Post benchmark scores — N/A (docs-only example).
 - [x] Add an entry to `CHANGELOG.md` — N/A for a docs example.
 - [x] Confirm you have read the AI policy.


CommanderStorm (commented):
Sorry to be blunt, but the code quality here is really not good.
It is absolutely unclear what this code does because it is so obfuscated.

It is a cool demo, but in terms of actually showing people how to do this, it fails sadly.
Can you break this down into the one thing that you want to show? fully blown games are likely not a good example, but I might be wrong.

clement-igonet:
@CommanderStorm thanks for the blunt feedback - it was fair, and I've rewritten this from scratch.

Dropped the tomato-watering minigame entirely (dense single-letter variables, procedural textures, touch/pointer-lock input, win condition) and focused the example on the one non-obvious thing worth showing: a true, level first-person camera at exactly 90° pitch, built only on the public `jumpTo` API. The trick is that at 90° the camera and `center` sit at the same altitude (the camera is looking sideways, not down), so `elevation` + `centerClampedToGround: false` + placing `center` a fixed distance ahead gives a level eye-height view - documented in a comment at the top of the file.

It now walks South Dearborn Street in the Loop, between real building footprints/heights baked in from OpenStreetMap (Old Colony, Monadnock and Fisher buildings among them), rather than an arbitrary synthetic scene. Worth noting for review: that data is intentionally a small static dataset rather than a live vector-tile source - a full real-world tiled city isn't safe to drive at a true 90° pitch (MapLibre's tile-covering logic caps its horizon handling at 89.25°, and I hit a real tab-crashing tile-loading runaway trying it against the full city before scaling back to this bounded set).

History is two commits: the original you reviewed, and a single rewrite on top, so the diff should be easy to follow.

HarelM:
To be fair, if this can't be done on a full city, due to crashes, we need to make sure we fix those crashes before adding this example, as the idea of the example is to show what possible, but if maplibre crashes then it's not really possible, isn't it?

clement-igonet:
Thanks for flagging that — I wanted to actually check it rather than just guess, so I ran a quick test.

I drove a full live vector-tile city (`https://tiles.openfreemap.org/styles/bright`, worldwide OpenMapTiles-schema data, no bounds) at pitch 90/89 with the same `elevation` + `centerClampedToGround: false` technique this example uses:
- Standing still, pitch 89: tile request count stabilized at 12 and stayed flat over 16s.
- Walking continuously (4 m/s, pitch 90) through Chicago's Loop for 6s: tile request count stayed flat at 13, JS heap stable at ~10MB.

No runaway tile loading in either case — `maxMercatorHorizonAngle` (89.25°, in `mercator_utils.ts`) already caps how far the tile-covering logic reaches toward the mathematical horizon, and that cap held up under this live source. So the bounded dataset in this example looks like a deliberate scoping choice rather than something forced by a crash, at least as far as I could reproduce. Happy to dig further if you have a specific case in mind where it did crash for you.

clement-igonet:
> we need to make sure we fix those crashes before adding this example

Do you mean it is a known issue ?

HarelM:
You raised that as an issue. Also please don't pass on AI responses and answer yourself.

clement-igonet:
Rather than leave an unverified bug assertion in the example I've removed it.
So I'm not filing an issue because I have no repro.

test/examples/walk-around-a-map-in-first-person.html
```diff
@@ -0,0 +1,240 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+<title>Walk around a map in first person</title>
+<meta property="og:description" content="Drive the camera as a level, eye-height first-person view using only jumpTo, a 90° pitch and elevation - walking a real Chicago street between real buildings." />
+<meta property="og:category" content="Camera & Animation" />
+<meta property="og:created" content="2026-06-14" />
```
HarelM: Should be updated?

test/examples/walk-around-a-map-in-first-person.html
```diff
@@ -0,0 +1,240 @@
...
+</style>
+</head>
+<body>
+<div id="map"></div>
+<div id="hint">W / S walk &middot; A / D turn</div>
+
+<script type="module">
+import * as maplibregl from '../../dist/maplibre-gl-dev.mjs';
+
+/*
```
HarelM: This should be removed, the example should be minimal.

test/examples/walk-around-a-map-in-first-person.html
```diff
@@ -0,0 +1,240 @@
...
+        },
+        paint: {
+            'text-color': '#fff',
+            'text-halo-color': '#000',
+            'text-halo-width': 1.5
+        }
+    });
+});
+
+function updateCamera() {
```
HarelM, line 78: There's a cameraFromTo method that might be useful here.

test/examples/walk-around-a-map-in-first-person.html
```diff
@@ -0,0 +1,240 @@
...
+<script type="module">
+import * as maplibregl from '../../dist/maplibre-gl-dev.mjs';
+
+/*
+ * A first-person camera, built only on MapLibre's public jumpTo API - no
+ * custom layer or extra rendering involved. It walks South Dearborn Street
+ * in Chicago's Loop, between real building footprints (including the
+ * landmark Monadnock and Fisher buildings), sourced from OpenStreetMap.
+ *
+ * The buildings are a small fixed set baked into this file, so the example is
```
HarelM: Why? Other examples are using openfreemap, which is fine for the examples.

HarelM:
This example is still too big and filled with unwanted comments. Please further de-vibe it and make it simpler.

test/examples/walk-around-a-map-in-first-person.html
clement-igonet: Updated the created date in d27677d4d.

test/examples/walk-around-a-map-in-first-person.html
clement-igonet: Removed in d27677d4d, the file no longer has that comment block.

test/examples/walk-around-a-map-in-first-person.html
clement-igonet: Done in d27677d4d: updateCamera now calls map.calculateCameraOptionsFromTo(eye, EYE_HEIGHT, ahead, EYE_HEIGHT), replacing the manual worldSize/focal-length math.

test/examples/walk-around-a-map-in-first-person.html
clement-igonet: Switched in d27677d4d: the example now uses the openfreemap basemap and its own 3D buildings instead of a baked-in dataset.

clement-igonet:
Simplified in d27677d4d (240 -> ~100 lines):
- dropped the baked-in building dataset for the openfreemap basemap and its 3D buildings,
- replaced the manual camera math with map.calculateCameraOptionsFromTo,
- removed the comment block,
- and refreshed the thumbnail.

test/examples/walk-around-a-map-in-first-person.html
```diff
@@ -0,0 +1,94 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+<title>Walk around a map in first person</title>
+<meta property="og:description" content="Drive an eye-height, level first-person camera over 3D buildings with WASD, using calculateCameraOptionsFromTo." />
```
HarelM:
```suggestion
<meta property="og:description" content="Drive an eye-height, level first-person camera over 3D buildings with the keyboard" />
```

test/examples/walk-around-a-map-in-first-person.html
```diff
@@ -0,0 +1,94 @@
...
+    const eye = toLngLat(player.x, player.y);
+    const ahead = toLngLat(
+        player.x + Math.cos(player.heading) * LOOK_AHEAD,
+        player.y + Math.sin(player.heading) * LOOK_AHEAD
+    );
+    map.jumpTo(map.calculateCameraOptionsFromTo(eye, EYE_HEIGHT, ahead, EYE_HEIGHT));
+}
+
+const keysDown = new Set();
+addEventListener('keydown', (event) => keysDown.add(event.code));
```
HarelM, line 89: This example should also work for mobile devices, so enabling only the keyboard might not be a great idea.
Something like the following, but a lot less styled (as we don't want to make the example too long only on the css part) might do the trick for both mobile and desktop:

Image

HarelM:
This is a lot better now, and focused on what it needs to show, without redundant comments. I've added a last comment related to mobile usage.

test/examples/walk-around-a-map-in-first-person.html
clement-igonet: Added in 7053ff497: a minimal on-screen W/A/S/D button pad wired via pointer events.

HarelM (approved):

HarelM:
Check the cspell file for relevant needed changes and this can be merged. Thanks!!

clement-igonet:
> Check the cspell file for relevant needed changes and this can be merged. Thanks!!

Went the other way to avoid touching the cspell dictionary: "pointerup" event name is gone.

PR #7789, opened 2026-06-15
chore!: use classes for all the events

HarelM:
## Launch Checklist

- Related to #6427 

- All map events are now real classes that are instantiated when they are fired. Renamed `MapLibreZoomEvent` to `MapBoxZoomEvent`, added the `rollstart`/`roll`/`rollend` and `style.load` events to `MapEventType`, and added event classes and type-map for `Marker`, `Popup`, `GeolocateControl` and `FullscreenControl`. 
- Removed `MapDataEvent`: the `data`/`dataloading`/`dataabort` events are now `MapSourceDataEvent | MapStyleDataEvent`, so source data events carry the full source info (`sourceId`, `tile`, `sourceDataType`, …). Added `MapMovementEvent` as the type for all camera-transition events (`move`/`zoom`/`rotate`/`pitch`/`roll`/`drag` and their `start`/`end` variants). 
- `Evented` is now generic over an event-type map (`Evented<EventType>`), so subclasses get strongly-typed `on`/`once`/`off` automatically without re-declaring overloads

I don't think this is really a breaking change as it mostly adds classes for what previously was only types, but it might break some typescript usage.

Assisted-By: Claude code, opus 4.8


CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 ## main
 ### ✨ Features and improvements
-- _...Add new stuff here..._
+- ⚠️ All map events are now real classes that are instantiated when they are fired. Renamed `MapLibreZoomEvent` to `MapBoxZoomEvent`, added the `rollstart`/`roll`/`rollend` and `style.load` events to `MapEventType`, and added event classes and type-map for `Marker`, `Popup`, `GeolocateControl` and `FullscreenControl`. Removed `MapDataEvent`: the `data`/`dataloading`/`dataabort` events are now `MapSourceDataEvent | MapStyleDataEvent`, so source data events carry the full source info (`sourceId`, `tile`, `sourceDataType`, …). Added `MapMovementEvent` as the type for all camera-transition events (`move`/`zoom`/`rotate`/`pitch`/`roll`/`drag` and their `start`/`end` variants). `Evented` is now generic over an event-type map (`Evented<EventType>`), so subclasses get strongly-typed `on`/`once`/`off` automatically without re-declaring overloads ([#7789](https://github.com/maplibre/maplibre-gl-js/pull/7789)) (by [@HarelM](https://github.com/HarelM))
```
CommanderStorm: > renamed `MapLibreZoomEvent` to `MapBoxZoomEvent`

ähm, can you come up with a different name for this?

src/util/evented.ts
```diff
@@ -70,6 +75,8 @@ export class Evented {
      * The listener function is called with the data object passed to `fire`,
      * extended with `target` and `type` properties.
      */
+    on<T extends keyof EventType & string>(type: T, listener: (event: EventType[T]) => void): Subscription;
+    on(type: string, listener: Listener): Subscription;
     on(type: string, listener: Listener): Subscription {
```
CommanderStorm: can we make the types for this and below stricter? why does it need to be `string`?

src/util/evented.ts
```diff
@@ -29,6 +29,11 @@ function _removeEventListener(type: string, listener: Listener, listenerList: Li
  */
 export class Event {
     readonly type: string;
+    /**
+     * The object that fired the event. Set when the event is fired, and narrowed to a more
+     * specific type (e.g. `Map`, `Marker`) by the event subclasses.
+     */
+    target?: unknown;
```
CommanderStorm, line 36: Is this being unknown not a bit anti-ts?
Could the class being a generic be better?

src/ui/popup.ts
```diff
@@ -116,6 +116,36 @@ const focusQuerySelector = [
     'textarea:not([disabled])',
 ].join(', ');
 
+/**
+ * The event class for popup events (`open` and `close`).
+ *
+ * @group Event Related
+ */
+export class PopupEvent extends Event {
+    type: 'open' | 'close';
```
CommanderStorm: not `keyof PopupEventType`? that is what it is, right?

CHANGELOG.md
HarelM: The handler is called BoxZoom, all events are prefixed with Map, so it ended as MapBoxZoomEvent. I did think about updating it, but I don't have a better name. Can you think of one?

src/util/evented.ts
HarelM: This addition makes it strictier, that's why I added it, but I don't think we can completely omit the string overload.

src/util/evented.ts
HarelM: I think making it generic will propagate too much. But in some places I override this definition so that it gets narrowed, for example for Popup events.

src/ui/popup.ts
HarelM: Good point, thanks!

HarelM: Done in 402a124

CHANGELOG.md
CommanderStorm: Unless with other changes, no. 👀

src/util/evented.ts
CommanderStorm: What happens if you do? Are our events not all known statically?

HarelM: I'll try that, but I'm not very optimistic. There's other places in the code that are using strings. I think I tried that the last time I touched the events stuff and left it this way on purpose.

CommanderStorm: Shure, I can also chip away at this if this is too much work for now 😉
Better types are very useful.

HarelM: I did a quick church and it's not that easy and involves casting due to for-in typescript limitations. 
Feel free to hack at this.

HarelM: Nevermind, I was able to solve it in a reasonable way. Feel free to take another look.

src/ui/events.ts
```diff
@@ -212,7 +212,6 @@ export type MapEventType = {
      * @see [Display HTML clusters with custom properties](https://maplibre.org/maplibre-gl-js/docs/examples/display-html-clusters-with-custom-properties/)
      */
     data: MapSourceDataEvent | MapStyleDataEvent;
-    tiledataloading: MapSourceDataEvent;
```
CommanderStorm: Is this removal from the enum intentional?

src/ui/map.ts
```diff
@@ -1858,8 +1858,8 @@ export class Map extends Camera {
      * @param type - The type of the event.
      * @param listener - The listener callback.
      */
-    once(type: keyof MapEventType | string, listener?: Listener): this | Promise<any>;
-    once(type: keyof MapEventType | string, layerIdsOrListener?: string | string[] | Listener, listener?: Listener): this | Promise<any> {
+    once(type: keyof MapEventType, listener?: Listener): this | Promise<any>;
+    once(type: keyof MapEventType, layerIdsOrListener?: string | string[] | Listener, listener?: Listener): this | Promise<any> {
```
CommanderStorm, line 1862: I am not an expert in this, but can we tighten it to this?

once(type: keyof MapEventType, listener?: Listener): this | Promise<any>;
once(type: keyof MapEventType, layerIds: string | string[], listener?: Listener): this | Promise<any>

src/util/evented.ts
```diff
@@ -5,17 +5,17 @@ import {extend, type Subscription} from './util.ts';
  */
 export type Listener = (a: any) => any;
 
-type Listeners = {[_: string]: Listener[]};
+type Listeners<EventType extends Record<string, any>> = {[_ in keyof EventType]?: Listener[]};
```
CommanderStorm, line 8: There is a bunch of `EventType extends Record<string, any>>`. I assume that we cannot get rid of this.

CommanderStorm (approved):

CommanderStorm:
Much better!

src/ui/events.ts
HarelM: Yes, this event is never fired.

src/util/evented.ts
HarelM: Nope, this is what typescript needs in order to allow generic stuff...

src/ui/map.ts
HarelM: No, because the method can receive 3 parameters, or 2, or just 1. The last method signature should have all the possible options so the code after that can handle all the options...
It's a terrible API to maintain due to all the overloading, but it has its merits when you write the code against this library...

PR #7797, opened 2026-06-17
perf(symbol): claim lowest-index cross-tile match in one pass instead of sorting per query

pholmstr:
### What / why
`TileLayerIndex.findMatches` runs on the main thread during symbol placement (on every zoom, as tiles are added). On the spatial-index path it sorts the full set of candidate results **for every symbol, on every query**:

```ts
const indexes = entry.index.range(/* ... */).sort();
for (const i of indexes) {
    const crossTileID = entry.crossTileIDs[i];
    if (!zoomCrossTileIDs[crossTileID]) { /* claim + break */ }
}
```

When many symbols share a key within the match window — e.g. dense or coincident point markers — `range()` returns a large candidate set, so this `Array.sort()` is `O(n log n)` per symbol and becomes the dominant cost of placement.

Two problems:
1. **Performance** — the sort is unnecessary. The loop only needs the single lowest-index *unclaimed* candidate, which is an `O(n)` scan.
2. **Correctness nit** — `.sort()` is called with no comparator, so it orders the numeric indices *lexicographically* (`[1, 2, 10]` → `[1, 10, 2]`). "First unclaimed" therefore isn't the lowest-index match the surrounding code reads as. The replacement uses the numeric minimum, which is deterministic and what the logic intends.

The one-parent-per-symbol dedup behaviour (#5993) is unchanged: each claimed `crossTileID` is still recorded in `zoomCrossTileIDs`.

### Benchmark
Chrome DevTools CPU profile while zooming + panning over a GeoJSON symbol layer of ~100k points with heavy coincidence (~9k distinct coordinates). `findMatches` self-time:

| build | `findMatches` self-time |
| --- | --- |
| `main` | ~45% |
| this change | ~26% |

~40% reduction on this workload; no change in rendered output. Negligible effect on sparse layers (small candidate sets).

### Testing
- Existing unit tests in `src/symbol/cross_tile_symbol_index.test.ts` pass (7/7).
- The tie-break order changes from lexicographic to numeric; a render test on a dense/coincident fixture would confirm no label-fade regression. Happy to add one if reviewers prefer.

### Launch Checklist
- [x] Confirm your changes do not include backports from Mapbox projects (unless with compatible license)
- [x] Add a changelog entry under `### ✨ Features and improvements`.
- [x] Update unit tests (existing tests cover the dedup behaviour and pass).
- [ ] Post benchmark scores (informal DevTools profile included above).


pholmstr:
We made these optimizations to enable faster performance on maps with 100k symbols on them.

CommanderStorm (approved):
Change LGTM

> The tie-break order changes from lexicographic to numeric; a render test on a dense/coincident fixture would confirm no label-fade regression. Happy to add one if reviewers prefer.

Yes, if we don't have such an render test, that would be appreciated.
I think that part was relatively solidly tested last time I checked, but there is a good chance that a part was missed.

src/symbol/cross_tile_symbol_index.ts
```diff
@@ -118,25 +118,31 @@ class TileLayerIndex {
...
-                    const crossTileID = entry.crossTileIDs[i];
-
-                    if (!zoomCrossTileIDs[crossTileID]) {
-                        // Once we've marked ourselves duplicate against this parent symbol,
-                        // don't let any other symbols at the same zoom level duplicate against
-                        // the same parent (see issue #5993)
-                        zoomCrossTileIDs[crossTileID] = true;
-                        symbolInstance.crossTileID = crossTileID;
-                        break;
+                let bestIndex = -1;
```
HarelM: Can we initialize to `Infinity` instead of adding a `bestIndex === -1` below?

src/symbol/cross_tile_symbol_index.ts
```diff
@@ -118,25 +118,31 @@ class TileLayerIndex {
 
             if (entry.index) {
                 // Return any symbol with the same keys whose coordinates are within 1
-                // grid unit. (with a 4px grid, this covers a 12px by 12px area)
+                // grid unit. (with a 4px grid, this covers a 12px by 12px area).
```
HarelM: This comment is fairly long.
Due to it, I would consider splitting this method more:
create `findMatchesForIndexedEntry(...)` and `findMatchesForNonIndexedEntry(...)` (or anyt other name you think is appropriate) and use this comment at the tsdoc comment of the method.

Same for the below position if branch.

HarelM:
@pholmstr any updates on this?

pholmstr:
Thanks for the patience — pushed updates addressing all feedback:
- `bestIndex` now initializes to `Infinity`, dropping the `-1` sentinel check
- Split `findMatches` into `findMatchesForIndexedEntry` / `findMatchesForNonIndexedEntry` with the long comments moved into tsdoc
- Added a render test (`symbol-cross-fade/coincident-dense-symbols`, 260 coincident symbols to force the spatial-index path) plus a unit test that deterministically covers the lowest-index tie-break @CommanderStorm asked about

Also rebased onto main to resolve the CHANGELOG conflict. Ready for another look.

test/integration/render/tests/symbol-cross-fade/coincident-dense-symbols/style.json
HarelM, line 1: This json file is fairly large, I'm not sure this is the best fit for a render test...
Is it possible to reproduce the issue with a simpler style.json file?

src/symbol/cross_tile_symbol_index.test.ts
```diff
@@ -236,6 +236,37 @@ describe('CrossTileSymbolIndex.addLayer', () => {
...
+            childInstances.push(makeSymbolInstance(2000, 2000, 'Dense'));
+        }
+
+        const mainTile = makeTile(mainID, mainInstances);
+        const childTile = makeTile(childID, childInstances);
+
+        index.addLayer(styleLayer, [mainTile], 0);
+        index.addLayer(styleLayer, [mainTile, childTile], 0);
+
+        // Each child symbol should claim the lowest-index unclaimed parent
```
HarelM, line 262: I'm not sure the comments in this tests are helpful, consider removing them...

HarelM:
Added some minor comments, looks good overall. THANKS!

src/symbol/cross_tile_symbol_index.test.ts
CommanderStorm: I think they explain pretty well why this is an invariant to care about...

CommanderStorm:
I removed the rendertest since it also passes on main..
Should be good to go now.

HarelM (approved):

PR #7800, opened 2026-06-17
feat!: camera composition instead of inheritance

HarelM:
## Launch Checklist

I wanted to do this for a long time, but I'm not sure this actually makes things better.
It does remove the inheritance between the map and camera, but it doesn't really split the logic between them. It's not really clear what should be part of the map and what should be part of the camera.
The agreement is that camera own transfrom class and map owns the handlers, but that's a bit shallow as they use each other when needed.
From maintenance perspective, it's easier to see the public API for the map object this way.

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.


HarelM:
@louwers what's the modeling of camera/map/gestures handler in Native?

src/ui/handler/transform-provider.ts
```diff
@@ -12,14 +13,16 @@ import {type LngLat} from '../../geo/lng_lat.ts';
  * handlers need the "desired state" to track accumulated changes.
  */
 export class TransformProvider {
-    _map: Map;
+    _camera: Camera;
+    _getTerrain: () => Terrain | null;
 
-    constructor(map: Map) {
-        this._map = map;
+    constructor(camera: Camera, getTerrain: () => Terrain | null) {
```
CommanderStorm: why is `getTerrain` an calback and not just passing in the actual value? 🤔

src/ui/camera.ts
```diff
@@ -324,263 +358,111 @@ export abstract class Camera extends Evented<MapEventType> {
...
 
-    /**
-     * Sets the elevation of the map's center point, in meters above sea level. Equivalent to `jumpTo({elevation: elevation})`.
-     *
-     * Triggers the following events: `movestart` and `moveend`.
-     *
-     * @param elevation - The elevation to set, in meters above sea level.
-     * @param eventData - Additional properties to be added to event objects of events triggered by this method.
-     */
     setCenterElevation(elevation: number, eventData?: any): this {
```
CommanderStorm, line 421: you removed lots of docs here.. is this intentional?

src/ui/map.ts
```diff
@@ -999,14 +978,539 @@ export class Map extends Camera {
     * ```
     */
     coveringTiles(options: CoveringTilesOptions): OverscaledTileID[] {
-        return coveringTiles(this.transform, options);
+        return coveringTiles(this._camera.transform, options);
+    }
+
+    // This is not part of the public API and will be removed in future versions
+    get transform(): ITransform { return this._camera.transform;}
```
CommanderStorm: 👀 
That sounds a bit like we are adding a footgun.
Can we mark this as private some other way?

src/ui/map.ts
```diff
@@ -999,14 +978,539 @@ export class Map extends Camera {
...
+     * const cameraAltitude = 1000;
+     * const bearing = 90;
+     * const pitch = 30;
+     * const roll = 45;
+     * const cameraOptions = map.calculateCameraOptionsFromCameraLngLatAltRotation(cameraLngLat, cameraAltitude, bearing, pitch, roll);
+     * // Apply calculated options
+     * map.jumpTo(cameraOptions);
+     * ```
+     */
+    calculateCameraOptionsFromCameraLngLatAltRotation(cameraLngLat: LngLatLike, cameraAlt: number, bearing: number, pitch: number, roll?: number): CameraOptions { return this._camera.calculateCameraOptionsFromCameraLngLatAltRotation(cameraLngLat, cameraAlt, bearing, pitch, roll); }
```
CommanderStorm, line 1402: that is a mothfull 😆

src/ui/map.ts
```diff
@@ -999,14 +978,539 @@ export class Map extends Camera {
...
+     * ```
+     * @see [Fly to a location](https://maplibre.org/maplibre-gl-js/docs/examples/fly-to-a-location/)
+     * @see [Slowly fly to a location](https://maplibre.org/maplibre-gl-js/docs/examples/slowly-fly-to-a-location/)
+     * @see [Fly to a location based on scroll position](https://maplibre.org/maplibre-gl-js/docs/examples/fly-to-a-location-based-on-scroll-position/)
+     */
+    flyTo(options: FlyToOptions, eventData?: any): this { this._camera.flyTo(options, eventData); return this; }
+    /**
+     * Stops any animated transition underway.
+     */
+    stop(): this { this._camera.stop(); return this; }
```
CommanderStorm, line 1462: would `stopTransitions` maybe be a better name? `map.stop` is a bit unclear, maybe. Not sure.

CommanderStorm (commented):
Mostly LGTM.

There are 2-3 areas where there is coverage missing, that is fine.
I have a few comments on other areas, nothing that should be hard to handle.

src/ui/handler/transform-provider.ts
HarelM: Because it can change "in the future" so I need a way to retrieve it, or I'll need to update the reference when it changes. This was easier and allowed camera not to have a reference to terrain, but I can see if I can update the reference instead if you think it's better.

src/ui/camera.ts
HarelM: Yes, they were comments for the public API, I moved them to `Map` object.

src/ui/map.ts
HarelM: No doubt, but it's part of the public API...

src/ui/map.ts
HarelM: Part of the public API, just moved here:
https://maplibre.org/maplibre-gl-js/docs/API/classes/Map/#stop

HarelM: That's the main reason I wanted to make camera a composition object, all these were part of the camera class and it was hard to really see the public API of map this way.
Not that a 4k lines file is easier, but at least it's in one place...

CommanderStorm: makes more sense, I did not understand that part of the motivation.

src/ui/map.ts
HarelM: I deliberately added this getter in order to reduce plugin breakage since `map.transform` was never part of the public API, but there are places where it is being used.
I'm not sure what the right approach is here.
I originally removed it, but then decided to add this getter back to reduce impact.
Let me know what you think...

CommanderStorm (approved):

src/ui/map.ts
CommanderStorm: what would it break? Is there something that we can migrate the people to?

HarelM: The idea is that people won't use map.trasform at all, but even we do it in one of the examples we provide. 
The idea is only to use public API methods, but people have been abusing map.trasform for a long time now...
The question at hand here is if we want to stop this abuse by removing this API completely as part of V6 or just keep the current status quo. 
I'm not sure what added value there is to "deliberately" break people plugins. 
I don't want to add this to the public API surface as the transform class is a huge class to "maintain" as part of the public API.

HarelM: I decided to remove it, adding it back won't be a breaking change so we can do it as part of a minor version release if someone will need it.

src/ui/handler/transform-provider.ts
HarelM: I've now changed it so that camera has a terrain instance and map updates it when it changes.
I'm not sure it's a lot better, but it does remove the callback, which I found awkward as well.

HarelM:
All code review comments have been addressed.
All tests pass.
I'll merge this soon and see if there's still anything worth changing before releasing v6.

PR #7801, opened 2026-06-17
Do not serialize undefined properties

xavierjs:
In Maplibre 5.6.1, the serialization to transfer data to the web worker was done using `JSON.stringify`
So if a feature had a property set to `undefined`, this property was completely dropped by the serialization.
If the `"has", "propertyName"` filter was applied on it, it was returning false since this property was dropped (so it compensated the error fixed by this PR: https://github.com/maplibre/maplibre-style-spec/pull/1712

But now the serialization is not done using `JSON.stringify` and undefined properties are serialized instead of being dropped as before. This PR is an optimization to 1/ recover the same behavior as before and 2/ decrease the data transferred to the webworkers.

## Launch Checklist

Assisted by Claude code


HarelM:
Can you link to the relevant PR? I'm having a hard time remembering such a change...

xavierjs:
The change was introduced by this commit: https://github.com/maplibre/maplibre-gl-js/commit/2d536a2b2                                                                                                                                                    

In src/source/geojson_source.ts, it replaces `options.data = JSON.stringify(data);` by `options.data = data`. It was released in v5.13.0.

So in MapLibre <5.13.0, when the data was sent to the web workers, `JSON.stringify` stripped `undefined` properties.

This fix strips `undefined` properties during the custom serialization in `web_worker_transfer.ts` (used for all main thread to worker transfers and back).

src/util/web_worker_transfer_has_filter.test.ts
HarelM, line 1: This shouldn't be a new file, see the PR you linked of an exact similar change.

HarelM:
Now I remember this change. Thanks for the link.
It was said that this custom serialization is faster than json.strigify/json.parse so that's why it got replaced. I think it's still true, but I haven't verified it myself.
Cc: @wayofthefuture

wayofthefuture:
I re-ran the failed test.

Would this have any effect on attempting to use updateData with undefined property values? If someone updates an existing prop to undefined would it carry through... did it carry through? It's been a while so can't really remember...

xavierjs:
> I re-ran the failed test.
> 
> Would this have any effect on attempting to use updateData with undefined property values? If someone updates an existing prop to undefined would it carry through... did it carry through? It's been a while so can't really remember...

The `updateData` uses `addOrUpdateProperties: [{key: 'bar', value: undefined}]` where `value` is typed as `any`.                                                                                                                             
                                     
  After the fix, `{key: 'bar', value: undefined}` gets serialized as `{key: 'bar'}` (value stripped). But in the worker, `applySourceDiff` destructures it as `const {key, value} = ...`, so `value` is `undefined` anyway because the field is
   missing. The property still ends up set to `undefined` in both cases.                                                                                                                                                                            
                                                                                                                                                                                                                                                    
  So the behavior for `updateData` is unchanged by this PR.

HarelM (approved):

PR #7811, opened 2026-06-20
Explore opt-in flattening of terrain RTT stacks

DoFabien:
## Context

While profiling terrain rendering with the OpenMapTiles Positron style, I noticed that a symbol layer between draped layers splits the render-to-texture work into multiple stacks.

In this particular style, terrain rendering uses 3 RTT stacks in both Mercator and Globe. I wanted to test whether combining these stacks could be worthwhile when an application is willing to accept a different ordering for the affected symbol layers.

I am keeping this PR as a draft because I would like feedback on the API and the rendering tradeoff before taking it further.

## Approach

This adds an opt-in `MapOptions` flag:

```js
allowTerrainDrapedLayerReordering: true
```

The default is `false`, so existing rendering is unchanged.

When enabled, symbol layers found between terrain-draped layers are deferred until after the combined terrain pass. This allows the draped layers to remain in one RTT stack instead of starting a new stack after each symbol layer.

The tradeoff is visual ordering: deferred symbols are rendered after the combined terrain pass instead of between the original draped layer groups. This is why I do not think the behavior should be enabled by default.

## Benchmark

I tested this with the OpenMapTiles Positron style, which has 51 style layers in this setup.

Each result measures the wall time for 125 synchronous rendered frames after warmup, with 20 samples per option. The same local bundle was used, changing only the option value.

| Projection | Option | RTT stacks | Median | Change |
| --- | --- | ---: | ---: | ---: |
| Mercator | `false` | 3 | 2466.9 ms | baseline |
| Mercator | `true` | 1 | 2075.6 ms | -15.9% |
| Globe | `false` | 3 | 2633.2 ms | baseline |
| Globe | `true` | 1 | 2243.8 ms | -14.8% |

These are end-to-end measurements for this scenario, not isolated GPU timings.

## Tests

I added a unit test that checks:

- the draped layers are collected into one stack when the option is enabled;
- the symbol layer is deferred;
- the symbol layer is rendered outside the terrain texture pass.

Before marking this ready, I still want to add a visual example showing the ordering difference.

## Feedback requested

I would especially appreciate feedback on:

- whether a public `MapOptions` flag is the right way to expose this;
- whether this ordering tradeoff is acceptable as an explicit opt-in;
- whether the option belongs somewhere else or should remain internal.

## Launch Checklist

- [x] Confirm my changes do not include backports from Mapbox projects.
- [x] Link to related issues. N/A — this came from profiling the current terrain RTT path.
- [ ] Include before/after visuals or gifs. TODO before marking the PR ready.
- [x] Write tests for the new behavior.
- [x] Document the new public option.
- [x] Post benchmark results.
- [ ] Add an entry to `CHANGELOG.md`. TODO before marking the PR ready.
- [x] Confirm I have read the MapLibre AI policy.

Assisted-By: GPT-5.5

### DIFF
--- a/src/ui/map.ts
+++ b/src/ui/map.ts
@@ -407,6 +407,14 @@ export type MapOptions = {
      */
     terrainSkirtLength?: 'none' | 'auto';
     /**
+     * If `true`, terrain rendering may reorder symbol layers between draped layers to reduce render-to-texture passes.
+     *
+     * Enable this only for styles where rendering those symbols after the combined terrain pass is acceptable.
+     *
+     * @defaultValue false
+     */
+    allowTerrainDrapedLayerReordering?: boolean;
+    /**
      * Defines the number of zoom level that will overscale instead of split tiles below (inclusive) a map's `maxZoom`.
      * When `undefined`, all zoom levels after source's max zoom will be overscaled.
      *
@@ -522,6 +530,7 @@ const defaultOptions: Readonly<Partial<MapOptions>> = {
     cancelPendingTileRequestsWhileZooming: true,
     centerClampedToGround: true,
     terrainSkirtLength: 'auto',
+    allowTerrainDrapedLayerReordering: false,
     zoomLevelsToOverscale: 4,
     anisotropicFilterPitch: defaultAnisotropicFilterPitch,
 };
@@ -612,6 +621,7 @@ export class Map extends Camera {
     /** @internal */
     _zoomLevelsToOverscale: number | undefined;
     _terrainSkirtLength: 'none' | 'auto';
+    _allowTerrainDrapedLayerReordering: boolean;
 
     /**
      * @internal
@@ -766,6 +776,7 @@ export class Map extends Camera {
         this._zoomSnap = resolvedOptions.zoomSnap;
         this._centerClampedToGround = resolvedOptions.centerClampedToGround;
         this._terrainSkirtLength = resolvedOptions.terrainSkirtLength;
+        this._allowTerrainDrapedLayerReordering = resolvedOptions.allowTerrainDrapedLayerReordering === true;
         this._refreshExpiredTiles = resolvedOptions.refreshExpiredTiles === true;
         this._fadeDuration = resolvedOptions.fadeDuration;
         this._crossSourceCollisions = resolvedOptions.crossSourceCollisions === true;
--- a/src/webgl/render_to_texture.test.ts
+++ b/src/webgl/render_to_texture.test.ts
@@ -126,6 +126,8 @@ describe('render to texture', () => {
     beforeEach(() => {
         tile.rttObjects.length = 0;
         tile.rttFingerprint = {};
+        map._allowTerrainDrapedLayerReordering = false;
+        style._order = ['maine-fill', 'maine-symbol'];
     });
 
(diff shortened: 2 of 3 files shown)


CommanderStorm:
@lucaswoj Is this something you think is a good idea to add or too "blunt" of an API?
(I am not sure if you are still working on Terrain ^^)

HarelM:
As a general flag, I think it makes sense, but I have a different question to ask:
Is this the "terrain bottleneck"?
From my point of view, as the Mapeak.com lead developer, the current terrain performance on mobile devices is not sufficient enough to enable it (terrain is only enable on desktop right now for Mapeak).
I don't think our style is that "complicated".
Can be found here:
https://github.com/IsraelHikingMap/VectorMap/blob/master/Styles/mapeak-hike.json

If this doesn't really move the terrain to a state where you can use it on mobile, I'm not sure it's worth the extra complexity this API brings.

It's worth mentioning that terrain performance might be a collection of things that need to be fixed in order to get it to a place where it works reasonable well on mobile devices and it's not a "single" problem (as most performance issues are), but this specific change is a "destructive" one in terms of how the map would look in order to allow better performance, and I'm interested to know if we know of other places that can benefit improvements without "breaking" the appearance.
 
Bottom line, I would be inclined to approve it only if this is the biggest performance benefit we can get, and we have exhausted most of the other options.

DoFabien:
Thanks, that's a fair question.

I had a quick look at `mapeak-hike.json`. From what I see, this PR would probably not have a large impact on Mapeak: the style mostly goes from 2 RTT stacks to 1 when the option is enabled. So I would expect a small performance gain, probably not more than 5% on this style, and with the known trade-off.

I agree this should not be presented as "the terrain bottleneck". I see it more as an opt-in optimization for styles where the number of RTT stacks grows higher.

In my main application, terrain is valuable because it gives important context in some urban planning and land/property workflows. But I still postpone enabling it by default because it is not smooth enough on some older PCs. Our style also adds business-specific sources and layers on top of more classic OSM layers, which can push the number of RTT stacks to around 4/5 in some cases. This is where reducing stack count can become more interesting.

Terrain performance has already improved a lot recently, especially with #7549. My feeling is that we are getting closer to the target performance.

I also started benchmarking terrain rendering more seriously on scenarios closer to my real-world usage, and I already found a few relatively simple improvements, mostly by avoiding unnecessary work. I will try to come back with cleaner measurements and separate PRs, but combined I think we can reasonably expect something around a 20% gain depending on the scenario. I still need to dig further and stabilize the measurements, but this also supports your point: terrain performance is probably a set of things to fix/optimize rather than one single magic bottleneck.

So I understand your hesitation. In fairly extreme scenarios, RTT flattening can be the current bottleneck, because GPU batches are not free. But in more classic scenarios, I do not think this is currently the main bottleneck. I am also very aware of the possible loss of the origin

lucaswoj:
> @lucaswoj Is this something you think is a good idea to add or too "blunt" of an API?
> (I am not sure if you are still working on Terrain ^^)

I'm not working on terrain at this time. Layer ordering can be a powerful tool for style optimization. I'm in favor of educating users about the perf impact of layer ordering but I'm against adding a flag that effectively re-orders layers automatically.

HarelM:
Closing this for now, if this is still needed/wanted in the future please request to reopen.

PR #7813, opened 2026-06-20
Fix: reuse terrain DEM texture when preparing terrain

DoFabien:
## Summary

This PR fixes a small GPU resource leak in the terrain DEM texture path.

When a raster-dem tile needed terrain preparation again, for example after DEM border backfilling, `getTerrainData()` always requested a new pooled texture and assigned it to `sourceTile.demTexture`.

If the tile already had a DEM texture, that old texture was overwritten without being returned to the texture pool or destroyed.

## Changes

- Reuse the existing `sourceTile.demTexture` when terrain preparation runs again.
- Keep updating the texture contents when DEM pixels change.
- Add a regression test that simulates repeated terrain preparation on the same tile.

The code change is intentionally small:

```ts
sourceTile.demTexture = this.painter.getTileTexture(sourceTile.dem.stride);
```
becomes:
```ts
sourceTile.demTexture ||= this.painter.getTileTexture(sourceTile.dem.stride);
```

So we only request a pooled texture when the tile does not already have one.

The test checks that:

- the DEM texture object stays the same;
- `getTileTexture()` is only called once;
- the same texture is returned to the painter when tile textures are cleared.

I also verified the first commit fails before the fix, then passes after the texture reuse change.

### DIFF
--- a/src/render/terrain.test.ts
+++ b/src/render/terrain.test.ts
@@ -217,64 +217,6 @@ describe('Terrain', () => {
         expect(minMaxNoDEM.maxElevation).toBeNull();
     });
 
-    test('getTerrainData reuses DEM texture across repeated terrain prepares', () => {
-        const tileID = new OverscaledTileID(5, 0, 5, 17, 11);
-        const tile = new Tile(tileID, 256);
-        const pixels = new RGBAImage({width: 4, height: 4}, new Uint8Array(4 * 4 * 4));
-        tile.dem = {
-            stride: 4,
-            dim: 2,
-            getPixels: vi.fn(() => pixels),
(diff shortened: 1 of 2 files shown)


CommanderStorm (approved):
not a huge fan of the test (lots of a bit brittle mocking).
Honeslty, I would not have written a test for this.. likely better to have it still.

CommanderStorm:
If you think this is appropriate, a changelog entry would be nice ^^

HarelM:
Yeah, I agree with @CommanderStorm, this fix has a very low risk and I think it's OK to just push the code change without a test, I'm not even sure there's a need for a changelog for this...

DoFabien:
I agree with both of you. I removed the test.
And yes, I think we can skip the changelog entry for this one too.

Issue #7699, opened 2026-05-26
Severe FPS drop at high zoom when terrain is enabled due to expensive DEM elevation/placement work (bad setSourceTileLodParams)


NeluQi:
**maplibre-gl-js version**: 5.24.0

**browser**: Google Chrome, latest stable version  
**OS**: Windows 11  
**Hardware**: Ryzen 9 CPU, RTX 5080 GPU

### Steps to Trigger Behavior

 1. Create a MapLibre map with terrain enabled.
 2. Use a relatively simple DEM/terrain source. In my case the terrain is not visually complex: it is essentially a flat elevated surface above the ground.
 3. Zoom in close to the terrain surface and move/rotate/pan the map.
 4. Record a Chrome Performance trace while the map is rendering.
 5. Compare FPS and main-thread cost at close zoom versus a more distant zoom level.

### Link to Demonstration

I do not yet have a public minimal reproduction link.

I can provide the Chrome Performance trace if needed. The trace shows the issue clearly: when zoomed close to the terrain, the map drops to roughly 1–2 FPS, while zooming out makes FPS return to a much more normal level.

Relevant observations from the trace:

- A single animation frame can spend about 335 ms in MapLibre rendering.
- The heavy work appears inside MapLibre internals, mainly under `_render`.
- Expensive sections include `_updateSources`, `_updatePlacement`, `placeLayerBucketPart`, `placeCollisionCircles`, terrain coordinate/elevation-related calls, and repeated `getDEMElevation` calls.
- The trace also shows work around `maybeDrawCoords`, `terrainCoords`, `screenPointToLocation`, `pointCoordinate`, `unproject`, `_updateRetainedTiles`, `_addTile`, `_removeTile`, `_abortTile`, `getAllTiles`, and some Minor GC activity.
- The most suspicious part is that terrain/elevation-related work becomes very expensive only at close zoom, even though the actual terrain in my scene is simple.

I have attached screenshots from the Chrome Performance flame chart showing the long frame and the relevant call stacks.

### Expected Behavior

Rendering performance should remain reasonably stable when zooming close to a simple terrain surface.

A simple flat or nearly flat elevated terrain should not cause the main thread to spend hundreds of milliseconds per animation frame. Even if additional terrain precision is needed at close zoom, the cost should scale predictably and should not reduce rendering to approximately 1–2 FPS on a powerful desktop machine.

### Actual Behavior

[...]

PR #7814, opened 2026-06-21
Reduce DEMData allocation pressure during elevation sampling

DoFabien:
## Context

While profiling terrain-related code, I noticed that every call to `DEMData.get()` creates a new `Uint8Array` view over the DEM buffer.

This happens in two places where the method can be called repeatedly:

- `Terrain.getDEMElevation()` reads four DEM values for each bilinear sample;
- the `DEMData` constructor calls `get()` for every cell while calculating `min` and `max`.

The traces in #7699 are what led me to inspect this path. However, this PR does not address the LOD behavior discussed in that issue, and I have not measured its impact on that specific reproduction. I am treating this as a standalone optimization of DEM construction and CPU-side elevation sampling.

## Approach

`DEMData` now keeps a cached byte view of its existing data buffer.

The cache is non-enumerable so it is not included when `DEMData` is transferred between the worker and the main thread. After deserialization, it is recreated lazily on the first access.

I also moved the 4 reads used for bilinear interpolation into `DEMData.sampleBilinear()`. This lets one elevation sample reuse the same byte view for all four pixels.

`getPixels()` still exposes a view over the same underlying buffer as before. The change only reuses the view object instead of creating a new one each time.

## Benchmark

I ran a small Node benchmark focused on the modified code paths. The values below are medians of repeated runs on Node v24.14.1.

| Benchmark | Before | After | Speedup |
| --- | ---: | ---: | ---: |
| Construct 256 × 256 DEM | 4.46 ms | 0.46 ms | 9.74× |
| Construct 512 × 512 DEM | 13.54 ms | 1.37 ms | 9.89× |
| 100,000 bilinear samples | 37.43 ms | 9.58 ms | 3.91× |

This is a micro-benchmark of the changed DEM operations, not a claim that the complete workload from #7699 becomes four or ten times faster.

## Tests

I added tests for:
- bilinear interpolation at integer and fractional coordinates;
- sampling through the padded DEM border;
- rejection of coordinates outside the available footprint;
- exclusion of the cached byte view from worker serialization;
- lazy recreation of the byte view after deserialization.

## Launch Checklist

- [x] Confirm my changes do not include backports from Mapbox projects.
- [x] Link to related issues. Partial improvement for #7699.
- [x] Write tests for the modified behavior.
- [x] Post benchmark results.
- [ ] Add an entry to `CHANGELOG.md`.
- [x] Confirm I have read the MapLibre AI policy.

Assisted-By: GPT-5.5


src/data/dem_data.ts
```diff
@@ -170,6 +197,28 @@ export class DEMData {
...
+        let byteView = this._byteView;
+        if (byteView?.buffer !== this.data.buffer) {
+            byteView = new Uint8Array(this.data.buffer);
+            this._setByteView(byteView);
+        }
+        return byteView;
+    }
+
+    private _setByteView(byteView: Uint8Array): void {
+        Object.defineProperty(this, '_byteView', {
```
CommanderStorm: Obvious if you think about the code, but took me a bit to understand.
What do you think?


```suggestion
        // defineProperty instead of plain assignment, since we don't want this field to be serialised
        Object.defineProperty(this, '_byteView', {
```

CommanderStorm (approved):
Code is clean, perf is better, code is unit tested and render tests are happy 👍🏻

CommanderStorm:
If you think this is appropriate, a changelog entry would be nice ^^

src/data/dem_data.test.ts
```diff
@@ -151,6 +151,27 @@ describe('DEMData.backfillBorder with encoding', () => {
...
+describe('DEMData.sampleBilinear', () => {
+    test('interpolates four neighboring pixels', () => {
+        const dem = new DEMData('sample', createMockImage(2, 2), 'custom', 1.0, 1.0, 1.0, 0.0);
+        const z00 = dem.get(0, 0);
+        const z10 = dem.get(1, 0);
+        const z01 = dem.get(0, 1);
+        const z11 = dem.get(1, 1);
+
+        expect(dem.sampleBilinear(0, 0)).toBe(z00);
+        expect(dem.sampleBilinear(0.5, 0.5)).toBe((z00 + z10 + z01 + z11) / 4);
```
HarelM: Better to write the expected number here to avoid extra logic in the tests.

src/data/dem_data.ts
HarelM: I found it a bit confusing as well, is there a different way to solve this without defineProperty?

src/data/dem_data.ts
```diff
@@ -98,21 +100,46 @@ export class DEMData {
         this.data[this._idx(dim, dim)] = this.data[this._idx(dim - 1, dim - 1)];
 
         // calculate min/max values
+        const pixels = this._getByteView();
         this.min = Number.MAX_SAFE_INTEGER;
         this.max = Number.MIN_SAFE_INTEGER;
-        for (let x = 0; x < dim; x++) {
-            for (let y = 0; y < dim; y++) {
-                const ele = this.get(x, y);
+        for (let y = 0; y < dim; y++) {
```
HarelM: The previos code was a bit "cleaner" in the sense it didn't use the += 4 part, is it possible to keep it somehow?

HarelM:
Can you please add the PR template items and add a changelog entry? 
Thanks!

src/data/dem_data.ts
DoFabien: Yes, a `WeakMap<DEMData, Uint8Array>` should be cleaner: the cache stays off the serialized object entirely, without needing defineProperty. I can switch to that if it sounds good to you.

HarelM: Where would you store this WeakMap?
But generally speaking, I'd like to avoid `defineProperty` if possible.

src/data/dem_data.ts
```diff
@@ -98,21 +100,46 @@ export class DEMData {
...
+            for (let x = 0; x < dim; x++) {
```
HarelM: You've changed the order of x and y, was this intentional?

DoFabien: Thanks for the careful review.
No, that order change was not intentional. It would not change the result here, but I reverted it to keep the original x then y order.

src/data/dem_data.ts
DoFabien: I stored it as a private static `WeakMap<DEMData, Uint8Array>` on DEMData. I just  pushed this implementation. The cache is still tied to the class, but stays off the serialized instance entirely, so we can avoid `defineProperty`. This should make the intent clearer without changing the behavior.
You’re right, this is probably the better way

src/data/dem_data.ts
```diff
@@ -170,6 +198,19 @@ export class DEMData {
             }
         }
     }
+
+    private _getByteView(): Uint8Array {
+        let byteView = DEMData.byteViewCache.get(this);
+        if (byteView?.buffer !== this.data.buffer) {
+            byteView = new Uint8Array(this.data.buffer);
+            DEMData.byteViewCache.set(this, byteView);
```
HarelM, line 206: When is this cache cleared then? or the fact that it's a weakmap is what allows us not to clear it?

DoFabien: Yes, that's why I used a `WeakMap` here.
The key is the 'DEMData' instance, so when the instance is not referenced anymore, the cache entry can be garbage collected too. So we don't need to clear it manually. If 'data.buffer' changes while the instance is still alive, '_getByteView()' replaces the cached view.

HarelM: Ok, cool, thanks!

src/data/dem_data.test.ts
```diff
@@ -206,6 +230,16 @@ describe('DEMData is correctly serialized and deserialized', () => {
     test('deserialized - mapbox', testDeserialization(mapboxDEM));
     test('deserialized - terrarium', testDeserialization(terrariumDEM));
     test('deserialized - custom', testDeserialization(customDEM));
+
+    test('byte view cache is not serialized', () => {
```
HarelM: This test is not relevant anymore, right?

DoFabien: Yes, you’re right, sorry I missed that. I removed this test.

HarelM (approved):

HarelM:
Thanks! can you fix the build test? and I'll merge this.

HarelM (approved):

PR #7826, opened 2026-06-23
fix: reject sendAsync promise on abort to stop leaking suspended frames

kamil-sienkiewicz-asi:
## Launch Checklist

- [ ] Link to related issues.
- [ ] Post benchmark scores.
- [x] Confirm you have read the AI policy.

### What

`Actor.sendAsync` left its returned promise pending forever when the request was aborted. Any caller that `await`ed it was therefore suspended permanently.

When a tile is aborted - which happens constantly as the viewport changes - that `await` never settles, so the `loadTile` async frame (and the `Tile` + worker data it captured) is never released. In a long-running session with frequent map movement / `setData`, these suspended frames accumulate without bound.

### Changes

- **`Actor.sendAsync`**: on abort, `reject(new AbortError(signal.reason))` instead of leaving the promise pending, so awaiters unwind and can be garbage-collected.
- **`GeoJSONSource.loadTile`**: wrap the `await` in try/catch and treat an aborted tile (`tile.aborted` / `isAbortError`) as a no-op `return` - matching the existing pattern in `VectorTileSource.loadTile`.

How leak was found:
- I'm working on automatic memory leak detection in our apps, while doing it I've noticed that my button (which is starting the animation loop of a lot of features on the map) is the starting point of reference holding which lead me to maplibre internals. After that, Opus 4.8 was helping me with investigation of maplibre code. I've created my own release of maplibre with this fix and run the test again, where thousands of promises were no longer visible in memory heap snapshots.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 - _...Add new stuff here..._
 
 ### 🐞 Bug fixes
+- Fix a memory leak where aborting a worker request (e.g. a GeoJSON tile load cancelled while panning) left its promise pending forever, so the awaiting async frame and everything it captured was never released; `Actor.sendAsync` now rejects with an `AbortError` on abort ([#XXXX](https://github.com/maplibre/maplibre-gl-js/pull/XXXX)) (by [@kamil-sienkiewicz-asi](https://github.com/kamil-sienkiewicz-asi))
 - _...Add new stuff here..._
 
 ## 6.0.0-16
--- a/src/source/geojson_source.test.ts
+++ b/src/source/geojson_source.test.ts
@@ -8,6 +8,7 @@ import {extend} from '../util/util.ts';
 import {SubdivisionGranularitySetting} from '../render/subdivision_granularity_settings.ts';
 import {MercatorTransform} from '../geo/projection/mercator_transform.ts';
 import {getWrapDispatcher, sleep, waitForEvent} from '../util/test/util.ts';
+import {AbortError} from '../util/abort_error.ts';
 import {type ActorMessage, type ClusterIDAndSource, type GeoJSONWorkerSourceLoadDataResult, MessageType} from '../util/actor_messages.ts';
 import type {IReadonlyTransform} from '../geo/transform_interface.ts';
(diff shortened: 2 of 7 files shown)


kamil-sienkiewicz-asi:
Hey @HarelM , could you check out this PR? Do you think it's safe or "// In case of abort the current behavior is to keep the promise pending." was there for a reason :D? I'm at git blame of this comment, but that's because it was moved a while ago by me

HarelM:
Thanks! I'm a bit concerned about abort errors, but I hope this is OK.
I would expect to see a similar try-catch in vector source...?

kamil-sienkiewicz-asi:
@HarelM I think it already does have similar try catch 🤔 
https://github.com/maplibre/maplibre-gl-js/blob/main/src/source/vector_tile_source.ts#L235

HarelM:
It doesn't check the abort error, just the tile.aborted, so I'm not sure it's 100% the same, I might be wrong, but better safe than sorry.
When I refactored the code to use promises instead of callbacks I deliberately left it this way since this was the behavior prior to my change.
I'm not saying it should be kept this way, but I know abort errors is a risky code flow especially with the fact that tiles request are aborted "intentionally"...

kamil-sienkiewicz-asi:
@HarelM sure, adjusted that

HarelM:
Looks good, my main concern is mostly about how risky this change is. How can we deriskify it?

kamil-sienkiewicz-asi:
@HarelM Added more tests, not sure how can I deriskify more. I've already installed patched version in my apps and run them for 8h each, did not spot any issues so far, those apps are very map heavy, tons of features and tons of custom layers.

Only change that i've noticed that there is no more hanging promises.

However if you have something on mind that I can check, let me know so I can run some manual testing or something similar.

HarelM:
Running your app for 8h is a good test, assuming you did a lot of zoom in zoom out and panning to see that no tiles are missing etc.

HarelM (approved):

kamil-sienkiewicz-asi:
Cool, let me know if you spot anything anytime soon so I will dig deeper to fix it 👌

PR #7838, opened 2026-06-27
chore(deps): inline getTileBBox to drop archived @mapbox/whoots-js

### LINKED ISSUE #7837: Replace end-of-life dependency @mapbox/whoots-js (upstream repo archived)
## Summary

`maplibre-gl` depends on `@mapbox/whoots-js` (https://www.npmjs.com/package/@mapbox/whoots-js) (^3.1.0) as a direct runtime dependency. Its upstream repository, `mapbox/whoots-js` (https://github.com/mapbox/whoots-js), was *archived by the owner on Feb 5, 2026 and is now read-only*. This makes it an end-of-life (EOL), unmaintained component.

## Why this matters

An archived upstream means:
 
- No future bug fixes, and crucially `no security patches` can be released for the package.
- Reported issues and PRs against the source repo can no longer be accepted or merged.
- Downstream consumers of maplibre-gl inherit this unmaintained dependency transitively, which is increasingly flagged by supply-chain/SCA tooling (e.g. Socket, Snyk, OSV scanners) and may fail security/compliance reviews.

## Where it is used

`@mapbox/whoots-js` provides `getURL()` to construct WMS `GetMap` request URLs for EPSG:3857 servers from `z/x/y` tile coordinates. It is a small, self-contained module (single `index.mjs`).

## Question

Are there plans to replace @mapbox/whoots-js with a maintained OSS alternative (or to vendor/inline and maintain the small amount of code in-tree under the MapLibre org) to ensure the long-term security and supportability of maplibre-gl consumers?

## Environment

- maplibre-gl version: latest main (dependency present as @mapbox/whoots-js: ^3.1.0 in package.json)

qorexdevs:
## Launch Checklist

- [x] Confirm your changes do not include backports from Mapbox projects (unless with compliant license) - this inlines `getTileBBox`/`getMercCoords` from `@mapbox/whoots-js`, which is ISC licensed (compliant). Kept the ISC + copyright attribution in a comment.
- [x] Link to related issues.
- [ ] Write tests for all new functionality. No new behavior - `getTileBBox` is moved verbatim, and the existing `.url replaces {bbox-epsg-3857}` test in `tile_id.test.ts` already covers its output (unchanged).
- [ ] Add an entry to `CHANGELOG.md` under the `## main` section. Skipped per CONTRIBUTING ("internal refactoring / technical debt reduction should not have a changelog entry") - happy to add one if you'd prefer.

closes #7837

`@mapbox/whoots-js` upstream is archived. maplibre only imports `getTileBBox`, so this inlines it (with its `getMercCoords` helper) into `tile_id.ts` next to the existing `getQuadkey` helper and drops the runtime dependency from `package.json`/lockfile. As suggested by @HarelM in the issue.


src/tile/tile_id.ts
```diff
@@ -282,6 +281,26 @@ export function calculateTileKey(wrap: number, overscaledZ: number, z: number, x
...
+    // for Google/OSM tile scheme we need to alter the y
+    y = Math.pow(2, z) - y - 1;
+
+    const min = getMercCoords(x * 256, y * 256, z);
+    const max = getMercCoords((x + 1) * 256, (y + 1) * 256, z);
+
+    return `${min[0]},${min[1]},${max[0]},${max[1]}`;
+}
+
+function getMercCoords(x: number, y: number, z: number): [number, number] {
```
HarelM: Is there an existing method that does this in this repo already? MercatorCoordinate class maybe?

src/tile/tile_id.ts
```diff
@@ -282,6 +281,26 @@ export function calculateTileKey(wrap: number, overscaledZ: number, z: number, x
     return (dim * dim * wrap + dim * y + x).toString(36) + z.toString(36) + overscaledZ.toString(36);
 }
 
+// Inlined from @mapbox/whoots-js (ISC, Copyright (c) 2017 Mapbox), whose upstream
```
HarelM: I this it will be better to describe what this method is doing using tsdoc comment instaed of its history...

src/tile/tile_id.ts
```diff
@@ -282,6 +281,26 @@ export function calculateTileKey(wrap: number, overscaledZ: number, z: number, x
...
+// repo is archived. Returns the EPSG:3857 bounding box string for a tile.
+function getTileBBox(x: number, y: number, z: number): string {
```
HarelM, line 294: Not sure this is related to tile id...? Maybe a separate file like utils mighy be a better fit to host this method? Mostly nit picking...

src/tile/tile_id.ts
qorexdevs: checked that - MercatorCoordinate works in normalized [0,1] units built on the mean earthRadius (6371008.8), but the {bbox-epsg-3857} token needs actual EPSG:3857 meters using the WGS84 spherical radius 6378137, so the two don't line up. reusing it would shift the bbox and break the existing url test. i named the constant and added a comment explaining why it differs.

src/tile/tile_id.ts
qorexdevs: done, added a tsdoc describing what it emits and kept the whoots attribution.

src/tile/tile_id.ts
qorexdevs: fair, it's only used by CanonicalTileID.url() so i left it next to its only caller for now. happy to move it to a util file if you'd prefer.

src/tile/tile_id.ts
```diff
@@ -282,6 +281,36 @@ export function calculateTileKey(wrap: number, overscaledZ: number, z: number, x
     return (dim * dim * wrap + dim * y + x).toString(36) + z.toString(36) + overscaledZ.toString(36);
 }
 
+// EPSG:3857 uses the WGS84 spherical radius, which differs from the mean
+// earthRadius (6371008.8) that MercatorCoordinate is built on, so the bbox math
+// below can't reuse it without changing the emitted coordinates.
+const epsg3857Radius = 6378137;
```
HarelM: Can you use capital letters for this please?
Also worth changing the comment to tsdoc style comment, and make it shorter I think if it's even needed given the name of the variable is fairly descriptive...

src/tile/tile_id.ts
HarelM: Fair enough.

src/tile/tile_id.ts
HarelM: It's worth adding that to the tsdoc comment or change the name of the method to something that references epsg-3857. for example getEpsg3857MercatorCoordinates or something like that (I'm nots sure the name is correct, so please make sure it has the correct meaning).

src/tile/tile_id.ts
qorexdevs: done - caps now (EPSG3857_RADIUS / EPSG3857_HALF_CIRCUMFERENCE) and the comment is a one-line tsdoc.

src/tile/tile_id.ts
qorexdevs: renamed it to getEpsg3857Coords with a short tsdoc, so it's clear it returns EPSG:3857 meters.

HarelM:
THANKS! One last request before I merge, can you add a changelog entry for this? That way if someone finds an issue with this change (although I'm not sure why they would) they can trace it back to this PR.

qorexdevs:
done, added a changelog line under main

HarelM (approved):

PR #7839, opened 2026-06-28
feat: async stylemissingimage

birkskyum:
Currently, a user can listen for `styleimagemissing` and call `map.addImage()` from that handler (as in our docs example [here](https://maplibre.org/maplibre-gl-js/docs/examples/generate-and-add-a-missing-icon-to-the-map/)). Here the user can provide the image synchronously, and all is good. Unfortunately, it's very often the case that a missing image has to be fetched from somewhere in an async manner, and in that case this setup breaks down: MapLibre checks for if the image exists, before that async work finishes and treats it as still missing.

Because of this lack of async-support, users who rely on async resources have to identify this limitation and setup one of these undocumented workarounds:
- Preloading of icon assets before the styleimagemissing event fire
- Adding a placeholder synchronously and updating it later
- Arranging another refresh after the async load registers the image

With this PR, the handler attached to `styleimagemissing` can optionally be async and return a promise. MapLibre waits for that promise before checking whether the image was added, so async icon loading works naturally:

```ts
map.on('styleimagemissing', async (e) => {
    const image = await loadIcon(e.id);
    map.addImage(e.id, image);
});
```

It's implemented as an backwards-compatible alternative approach. The code checks if the function is PromiseLike (if it returns something with a `.then` function on it), so that existing users with a sync function are unaffected by it.

## Launch Checklist

 - [ ] Link to related issues.
 - [ ] Post benchmark scores.

Assisted-By: GPT 5.5

### DIFF
--- a/src/util/evented.ts
+++ b/src/util/evented.ts
@@ -184,7 +184,7 @@ export abstract class Evented<EventType extends Record<string, any> = Record<str
         return this;
     }
 
-    protected async fireAsync(event: Event): Promise<this> {
+    protected async fireAndWaitForListeners(event: Event): Promise<this> {
         const promises: Array<PromiseLike<unknown>> = [];
         const asyncEvent = event as AsyncEvent;
         // Store the collector on the event so parent listeners are included as the event bubbles.
--- a/src/util/evented.ts
+++ b/src/util/evented.ts
@@ -184,14 +184,9 @@ export abstract class Evented<EventType extends Record<string, any> = Record<str
         return this;
     }
 
-    protected async fireAsync(event: Event | string, properties?: any): Promise<this> {
-        if (typeof event === 'string') {
-            event = new Event(event, properties || {});
-        }
-
+    protected async fireAsync(event: Event): Promise<this> {
         const promises: Array<PromiseLike<unknown>> = [];
         const asyncEvent = event as AsyncEvent;
-        const previousPromises = asyncEvent[asyncListenerPromises];
         // Store the collector on the event so parent listeners are included as the event bubbles.
         asyncEvent[asyncListenerPromises] = promises;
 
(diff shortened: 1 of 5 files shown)


src/util/evented.ts
```diff
@@ -171,6 +184,22 @@ export abstract class Evented<EventType extends Record<string, any> = Record<str
         return this;
     }
 
+    protected async fireAsync(event: Event): Promise<this> {
```
HarelM: Is there a way to use "regular" `fire` method? I'm not very comfortable withe the fact that I'll need to "think" which method to use when I fire an event...

birkskyum: It’s very tricky to do this with only `fire()`, and I actually tried to stay clear of touching `fire` because it has a lot of internal call sites, plus external user code. Changing its return type from `this` to e.g. `Promise<this>` would be breaking for chaining and existing return paths. Adding any kind of promises to fire could also cause timing/performance regressions in paths that are currently synchronous. 

That is why i ended up adding `fireAndWaitForListeners()` (prev. `fireAsync()`) instead. That way it's guaranteed not to regress fire(), and we only need this wrapper in one internal missing-image case (because it has this unusual [getImage check](https://github.com/maplibre/maplibre-gl-js/blob/859ccd2a0677975219743548f336c9d95766b781/src/render/image_manager.ts#L229-L230) right after firing). So regarding the thinking which method to use, the answer is that we'd always use `fire()`, like before, and there's no alternative fire function added to the public api.

I did explore keeping only a `fire` function, by overloading the function, but I found that to quickly because very confusing because the second argument of `fire` is an optional `properties` argument, and having two optional arguments is hard to reason about.

HarelM: Generally speaking, firing an event and expecting it to change the state of the system in order to proceed to refreshing the images is not a great architecture, and this promise change is just increasing the surface of a bad pattern, IMHO. Is it possible to solve this issue in a different way? Maybe expose a different API to solve this? Add an async callback to override to missing images or something? 
To be fair, I always found this event cumbersome...
We are at v6 breaking changes phase still, maybe we can leverage it to solve this issue better...?

birkskyum: I've looked into more clean solutions for v6. If we take away the mutation responsibility from the `stylemissingimage` event, and put it in a separate resolver, then it gives a number of benefits. If instance, if a resolver does provide an image, it doesn't have to fire the `stylemissingimage` which can be used for just unexpected "actually missing" cases.

- #7850

HarelM: I believe the linked PR is the right solution to this issue, it feels right.

PR #7850, opened 2026-06-29
feat: setMissingStyleImageResolver

birkskyum:
Superseding:
- #7839 

Right now, missing style images are handled through `styleimagemissing`. That works when the listener can add the image synchronously, but it breaks down when the image has to be fetched first: MapLibre checks for the image before the async work finishes, so the image is still treated as missing. It's also a strange design to expect the event to cause a mutation - styleimagemissing is in practice more of a hook than an event.

This PR adds `map.setMissingStyleImageResolver(...)`, a dedicated API for resolving images referenced by the style, such as `icon-image` and pattern images. The resolver can return image data synchronously or asynchronously, optionally with the same image options accepted by `map.addImage()`.

`styleimagemissing` gets fired if the resolver doesn't provide the image, and in `styleimagemissing` it is still possible to provide an image sync like before.

Before

```ts
map.on('styleimagemissing', async (e) => {
    const response = await fetch(`/icons/${e.id}.png`);
    const image = await createImageBitmap(await response.blob());

    // This may finish too late for the current missing-image request.
    map.addImage(e.id, image, {pixelRatio: 2});
});
```

After
```ts
map.setMissingStyleImageResolver(async (id) => {
    const response = await fetch(`/icons/${id}.png`);
    const image = await createImageBitmap(await response.blob());

    return {
        image,
        options: {pixelRatio: 2}
    };
});

// if the image is resolved by above `styleimagemissing` isn't fired
```

## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

Assisted-By: GPT 5.5


src/ui/map.ts
```diff
@@ -2599,13 +2625,87 @@ export class Map extends Camera {
...
+            }
+
+            style.addImage(id, styleImage);
+            if (styleImage.userImage?.onAdd) {
+                styleImage.userImage.onAdd(this, id);
+            }
+        });
+    }
+
+    _normalizeMissingStyleImageResolverResult(result: Exclude<MissingStyleImageResolverResult, null | undefined>): {image: StyleImageSource; options: Partial<StyleImageMetadata>} {
```
HarelM: Is this really needed? Why not support a more "simple" object?

src/ui/map.ts
```diff
@@ -2599,13 +2625,87 @@ export class Map extends Camera {
...
+        }
+
+        const resolver = this._missingStyleImageResolver;
+        if (!resolver) {
+            this.style.setMissingImageResolver(null);
+            return;
+        }
+
+        const style = this.style;
+        style.setMissingImageResolver(async (id) => {
```
HarelM: This looks like an over kill, why not use `resolver` and make the API "simpler" in terms of passing objects deeper into the child classes?

src/render/image_manager.ts
```diff
@@ -218,19 +229,26 @@ export class ImageManager extends Evented {
         });
     }
 
-    _getImagesForIds(ids: string[]): GetImagesResponse {
+    async _getImagesForIds(ids: string[]): Promise<GetImagesResponse> {
+        const missingIds = Array.from(new Set(ids.filter((id) => !this.getImage(id))));
+
+        const resolvedMissingIds = new Set<string>();
+        await Promise.all(missingIds.map(async (id) => {
+            if (await this._notifyMissingImage(id)) {
```
HarelM: I think calling a method `_notifyMissingImage` and expecting it to resolve the image is confusing. When I read this code I did not understand why there's a filter at the top and the for loop below that checks the get image again.
So I think it's worth rethinking the logic here to make it a bit more readable to straight forward.
1. Get all missing images
2. Try to resolve them
3. If there are still missing images raise the stylemissing image event
4. Return the resolved images

Or something like that.

src/ui/map_tests/map_images.test.ts
HarelM, line 1: There are a lot of added tests here, from my experience with the test that AI writes, it usually writes more tests than actually needed to cover the newly added functionality... So please go over those and see which are not needed to really cover a scenario or increase coverage.

test/examples/generate-and-add-a-missing-icon-to-the-map.html
```diff
@@ -24,9 +24,7 @@
         style: 'https://demotiles.maplibre.org/style.json',
     });
 
-    map.on('styleimagemissing', (e) => {
-        const id = e.id; // id of the missing image
-
+    map.setMissingStyleImageResolver((id) => {
```
HarelM, line 28: Great!

HarelM:
I think this pattern is a lot better than the `styleimagemissing` concept.
Can you add a note in the migration guide from v5 to v6 about this?

src/ui/map.ts
birkskyum: I have flattened the result shape, so the options are now top-level instead of in an options object.

HarelM:
@1ec5 I know OSM Americana uses the existing mechanism to its extreme, please let us know if you think this is a good direction or not.

src/render/image_manager.ts
```diff
@@ -218,19 +229,19 @@ export class ImageManager extends Evented {
...
-                this.fire(new MapStyleImageMissingEvent({id}));
-                //Try to acquire image again in case styleimagemissing has populated it
-                image = this.getImage(id);
-            }
+            const image = this.getImage(id);
 
-            if (image) {
+            // Images added by styleimagemissing listeners are available for future requests,
+            // but do not satisfy the current response.
+            const includeImage = image && (!initiallyMissingIds.has(id) || resolvedMissingIds.has(id));
```
HarelM: I'm not sure I'm aligned with this check. Can you elaborate why it's here?
I would expect the flow to simply try and resolve the missing images and then continue to the rest of the code. Are we bookkeeping this in order to improve the warning message? If so, I would recommend simplifying it and just warnOnce in case there are still missing images after the resolution attempt, so users who "did everything right" would not get this warning.

1ec5:
At a glance, the API design looks like a welcome improvement in principle. I wanted to design something similar for the iOS SDK, but it would’ve been difficult to pull off in Objective-C. I probably would’ve tried to be consistent with [this native iOS function](https://developer.apple.com/documentation/appkit/nsimage/init(size:flipped:drawinghandler:)?language=objc), which is also about programmatically generated images. The main difference is that your PR only runs the functions as a fallback, whereas it might be intuitive as an override of the default image behavior for a given ID.

HarelM:
You have the addImage for that, don't you? 
How would handle the case of mixed existing sprite images with missing images with your proposal?

1ec5:
> You have the addImage for that, don't you?

`addImage` is synchronous, whereas something on-demand is also useful when the images are dynamic or not fully defined upfront.

> How would handle the case of mixed existing sprite images with missing images with your proposal?

Not sure I follow. The API we’re discussing is scoped to an individual image ID. Essentially, a developer would find it convenient to set an image to a callback instead of static image data.

src/render/image_manager.ts
birkskyum: Basically we're doing two separate things in this PR right now.

1. Adding the new `setMissingStyleImageResolver` resolver, which users will use instead of `styleimagemissing`.
2. Changing `styleimagemissing` from being a place where you can resolve the image sync, to being a place you can't really do much but report to Sentry that the new resolver isn't able to find the image. 

We can leave the second part out of this PR, and keep `styleimagemissing` as a fallback that can resolve an image synchronously. It would remain a "mutating" event, but it would make v6 effectively non-breaking to existing users. I think it might give the most positive reception to add the new API in v6, and let people migrate to it, while leaving (maybe soft-deprecating) the old sync-only `styleimagemissing` approach to reduce churn.

birkskyum: Separated it, so this PR just does the 1st part - adding `setMissingStyleImageResolver`

src/render/image_manager.ts
```diff
@@ -244,13 +251,42 @@ export class ImageManager extends Evented {
...
-            } else {
-                warnOnce(`Image "${id}" could not be loaded. Please make sure you have added the image with map.addImage() or a "sprite" property in your style. You can provide missing images by listening for the "styleimagemissing" map event.`);
+            } else if (!image) {
+                warnOnce(`Image "${id}" could not be loaded. Please make sure you have added the image before it is needed with map.addImage(), resolved it with map.setMissingStyleImageResolver(), or included it in a "sprite" property in your style.`);
             }
         }
         return response;
     }
 
+    async _resolveMissingImageIds(ids: Set<string>): Promise<void> {
```
HarelM: This is a single line method used once, I think it can be "inlined".

src/render/image_manager.ts
```diff
@@ -244,13 +251,42 @@ export class ImageManager extends Evented {
...
+        await Promise.all(Array.from(ids, (id) => this._resolveMissingImageId(id)));
+    }
+
+    async _resolveMissingImageId(id: string): Promise<void> {
+        if (this.getImage(id)) return;
+
+        let request = this.missingImageRequests.get(id);
```
HarelM: This is nice, but I'm not sure it's worth the extra complexity. I would consider moving the responsibility of "making sure we don't repeat requests" to the method handler, if the user want to handle this.

src/render/image_manager.ts
```diff
@@ -244,13 +251,42 @@ export class ImageManager extends Evented {
                     textFitHeight: image.textFitHeight,
                     hasRenderCallback: Boolean(image.userImage?.render)
                 };
-            } else {
-                warnOnce(`Image "${id}" could not be loaded. Please make sure you have added the image with map.addImage() or a "sprite" property in your style. You can provide missing images by listening for the "styleimagemissing" map event.`);
+            } else if (!image) {
```
HarelM: What other possibilities there are here besides `!image`? Am I missing anything? Is this a leftover from previous commit?

src/render/image_manager.ts
birkskyum: Okay, I have removed the dedupe, so the user can add it if needed

https://github.com/maplibre/maplibre-gl-js/pull/7850/commits/50903d6840d3408172bc66fd75c088a2e45edce1

src/ui/map_tests/map_images.test.ts
birkskyum: It's down to four that each pull their weight.

src/ui/map.ts
birkskyum: Style now owns the addImage/onAdd bookkeeping

src/style/style.ts
```diff
@@ -999,6 +1006,23 @@ export class Style extends Evented<MapEventType> {
         return this.imageManager.getImage(id);
     }
 
+    setMissingImageResolver(resolver: MissingImageResolver | null): void {
+        if (!resolver) {
+            this.imageManager.setMissingImageResolver(null);
+            return;
+        }
+
+        this.imageManager.setMissingImageResolver(async (id) => {
```
HarelM: I don't understand all the complexity around setting the resolver, can you explain why all this is needed and it's not enough to do `this.imageManager.setMissingImageResolver(resolver)`?
I would also consider not doing the if at all, just pass through whatever the user passed.

src/ui/map.ts
```diff
@@ -609,6 +631,7 @@ export class Map extends Evented<MapEventType> {
     _overridePixelRatio: number | null | undefined;
     _maxCanvasSize: [number, number];
     _terrainDataCallback: (e: MapStyleDataEvent | MapSourceDataEvent) => void;
+    _missingStyleImageResolver: MissingStyleImageResolver | null = null;
```
HarelM, line 628: Is this still correct? Looks like a duplication of the code written in the style.ts file (with the below assignment)...?

test/examples/generate-and-add-a-missing-icon-to-the-map.html
```diff
@@ -49,7 +47,7 @@
             }
         }
 
-        map.addImage(id, {width, height: width, data});
+        return {width, height: width, data};
```
HarelM: Maybe call `addImage` instead of returning something? This is similar to what you'd do for the event only now you have a callback where you can do what's needed in a async way.

test/examples/display-a-remote-svg-symbol.html
```diff
@@ -26,25 +26,21 @@
...
+        const response = await fetch(id);
+        const svgText = await response.text();
+        const svg = 'data:image/svg+xml;charset=utf-8,' + encodeURIComponent(svgText);
+        const image = new Image();
+        const promise = new Promise((resolve) => {
+            image.onload = resolve;
         });
+        image.src = svg;
+        await promise; // Wait for the image to load
+        return image;
```
HarelM: Use `addImage` here instead?

src/ui/events.ts
```diff
@@ -238,9 +238,10 @@ export type MapEventType = {
      */
     'style.load': MapStyleLoadEvent;
     /**
-     * Fired when an icon or pattern needed by the style is missing. The missing image can
-     * be added with {@link Map.addImage} within this event listener callback to prevent the image from
-     * being skipped. This event can be used to dynamically generate icons and patterns.
+     * Fired when an icon or pattern needed by the style is missing and no missing style image resolver
```
HarelM, line 241: I would consider advertising only the `setMissingStyleImageResolver` as part of this tsdoc comment to move users to use the other API.
I would even consider removing this functionality as part of v6 and only add this back if people complain that they can't use the new functionality instead.
I think the event should be an event and not a weird hook, since v6 wasn't released yet we can still break things, but we can add them back if there's enough momentum around it, if anyone that needs this functionality will simply migrate to the new hook I think we will be in a better place from a maintenance perspective.

src/style/style.ts
birkskyum: I added all this ceremony to guaranteed the registration of the image, but by making the user responsible for calling addImage all this complexity does goes away, and the wrapper is now just pass-through:

```ts
    setMissingImageResolver(resolver: MissingImageRequestHandler | null): void {
        this.imageManager.setMissingImageResolver(resolver);
    }
```

HarelM: I think it's a simpler approach, and not too different than what user are used to with the sytleimagemissing event, so I think it's better to simplify this.

src/ui/map.ts
birkskyum: Yes, it is still correct. It's there so the resolver isn't lost in case the user calls setStyle()

src/ui/events.ts
birkskyum: Would it be okay, to have this suggested breaking change in a follow-pr, so we have it isolated?

birkskyum: It would make a potential Revert-story a bit cleaner

HarelM: Sure, if it's easier for you then no problem.

birkskyum: - #7892

src/render/image_manager.ts
birkskyum: - #7892

src/ui/map.ts
```diff
@@ -2686,6 +2703,7 @@ export class Map extends Evented<MapEventType> {
             return this;
         } else {
             this.style = new Style(this, options || {});
+            this._applyMissingStyleImageResolver();
```
HarelM: Does it make sense to add this to the constructor options instead of calling this after every time the style is instanciated?

HarelM: Or maybe pulling this inside the style constructor from the map object that is being passed into the `Style` constructor? What is the strategy with other map configurations that are related to the style object?

src/ui/map.ts
```diff
@@ -3173,11 +3222,7 @@ export class Map extends Evented<MapEventType> {
      * ```
      */
     updateImage(id: string,
-        image: HTMLImageElement | ImageBitmap | ImageData | {
-            width: number;
-            height: number;
-            data: Uint8Array | Uint8ClampedArray;
-        } | StyleImageInterface): this {
+        image: StyleImageSource): this {
```
HarelM: nit: this can probably be folded up to a single line.

HarelM:
I think this looks great! Thanks! Added a few final comments, but otherwise ready to be merged.

src/ui/map.ts
birkskyum: Good idea. It's just called in the style constructor now instead of multiple places in map.ts

birkskyum: commit - https://github.com/maplibre/maplibre-gl-js/pull/7850/commits/7e38bc8f70217ffe65234de1d60abf8d3b308f00

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## main
 ### ✨ Features and improvements
 - Improve terrain render-to-texture preparation performance by skipping sources that are not rendered to terrain textures ([#7863](https://github.com/maplibre/maplibre-gl-js/pull/7863)) (by [@DoFabien](https://github.com/DoFabien))
+- Add `Map#setMissingStyleImageResolver` for resolving missing style images with sync or async callbacks ([#7850](https://github.com/maplibre/maplibre-gl-js/pull/7850)) (by [@birkskyum](https://github.com/birkskyum))
```
```suggestion
- Add `Map.setMissingStyleImageResolver` for resolving missing style images with sync or async callbacks ([#7850](https://github.com/maplibre/maplibre-gl-js/pull/7850)) (by [@birkskyum](https://github.com/birkskyum))
```

HarelM (approved):

HarelM:
THANKS!!

PR #7851, opened 2026-06-29
fix: FPS drop under globe projection with large pitch

### LINKED ISSUE #7672: Severe FPS drop under globe projection with pitch, zoom > 5, and bearing pointing across the antimeridian

thomfuhrmann:
## Launch Checklist

 - [x] Link to related issues.
 - [ ] Post benchmark scores.

Assisted by Google Gemini

- Fixes #7672 

Adds a check in the covering tiles in the calculation of the horizontal coordinate difference of camera position and camera center in x-direction to account for wraps around the anti-meridian.


HarelM:
I don't understand how the code change fixes this issue, can you elaborate.
I also suggest to move the empty lines and formatting changes to a different PR, it adds unneeded noise...

thomfuhrmann:
> I don't understand how the code change fixes this issue, can you elaborate. I also suggest to move the empty lines and formatting changes to a different PR, it adds unneeded noise...

The problem was the coordinate difference calculation in x for the camera and center of the camera. Imagine the case where the camera is close to the anti-meridian and looks around it to a center across the anti-meridian. In this case the camera would have a Mercator x-coordinate close to 0 and the center would have a x-coordinate close to 1. In this case the coordinate difference would be close to 1, but around the normalized Earth it can only be in the range [-0.5, 0.5]. The calculation of this fix ensures that it stays in this range.

Sorry for the empty lines, I tried to reformat it to its original state.

Hope, that helps!

DoFabien:
Thanks for the fix. I had started looking at this a few days ago, and my approach was a bit different: I think the camera/center distance wrapping should stay specific to the globe projection, rather than being coded directly in `covering_tiles.ts`.

`coveringTiles()` is shared by projections. The direct distance is still correct for Mercator, while Globe needs to use the shortest X distance when the camera looks across the antimeridian.

A cleaner shape could be to move this calculation behind `CoveringTilesDetailsProvider`, similarly to what is already done for `distanceToTile2d`:

- Mercator keeps `Math.hypot(a.x - b.x, a.y - b.y)`.
- Globe uses the shortest wrapped X distance across the antimeridian.

This keeps `coveringTiles()` projection agnostic.

thomfuhrmann:
@DoFabien thanks for your feedback! I added a test to only do the wrap for the globe view. I see where you are heading at, but this would require an adaptation of the current API for `CoveringTilesDetailsProvider`, I think.

DoFabien:
Thanks, that makes sense.

I agree it would require a small 'CoveringTilesDetailsProvider' adaptation, but I think that is the right place for projection-specific behavior. Otherwise `covering_tiles.ts` has to know about globe rendering directly.

That said, the current approach also looks functionally fine to me. My point is mostly a design preference: I would rather keep this projection-specific logic in the provider than make `covering_tiles.ts` know about globe rendering directly.

HarelM:
Yeah, I agree with @DoFabien, there shouldn't be an `instanceof` inside covering tiles file.

thomfuhrmann:
I agree too that this would be a cleaner solution. If you agree, I can suggest necessary API changes.

HarelM:
Please do so, thanks!

src/geo/projection/globe_transform.ts
```diff
@@ -459,4 +459,12 @@ export class GlobeTransform implements ITransform {
     getFastPathSimpleProjectionMatrix(tileID: OverscaledTileID): mat4 {
         return this.currentTransform.getFastPathSimpleProjectionMatrix(tileID);
     }
+
+    /**
+     * account for wrapping of centerCoord/cameraCoord around the anti-meridian
+     * this ensures that the coordinate difference stays in the range [-0.5, 0.5]
+    */
+    normalizeDeltaX(deltaX: number): number {
+        return deltaX - Math.round(deltaX);
```
HarelM: I think this should be moved to vertical perspective, and here use something like return `this.currentTransform.normalizeDeltaX(deltaX);`

HarelM:
Please revert white space changes too.

HarelM:
Add a changelog as well for this to be merged.

thomfuhrmann:
@HarelM There seems to be an issue with the current version of covering_tiles.test.ts, because there are empty lines with spaces which get removed by eslint. But the other points you mentioned I could fix.

src/geo/projection/covering_tiles.ts
```diff
@@ -148,9 +148,16 @@ export function createCalculateTileZoomFunction(maxZoomLevelsOnScreen: number, t
         thisTileDesiredZ = thisTileDesiredZ + scaleZoom(distanceToCenter3D / distanceToTile3D / Math.max(0.5, Math.cos(degreesToRadians(cameraVerticalFOV / 2))));
         thisTileDesiredZ += pitchTileLoadingBehavior * scaleZoom(Math.cos(thisTilePitch)) / 2;
         thisTileDesiredZ -= scaleZoom(Math.max(1, tileCount / tileCountPitch0 / tileCountMaxMinRatio)) / 2;
+
+        // add zoom levels for narrow lenses
+        const defaultFov = 36.87;
```
HarelM: What's this?

src/geo/projection/covering_tiles.ts
```diff
@@ -232,7 +239,10 @@ export function coveringTiles(transform: IReadonlyTransform, options: CoveringTi
     const numTiles = Math.pow(2, nominalZ);
     const cameraPoint = [numTiles * cameraCoord.x, numTiles * cameraCoord.y, 0];
     const centerPoint = [numTiles * centerCoord.x, numTiles * centerCoord.y, 0];
-    const distanceToCenter2d = Math.hypot(centerCoord.x - cameraCoord.x, centerCoord.y - cameraCoord.y);
+
+    const deltaX = transform.normalizeDeltaX(centerCoord.x - cameraCoord.x);
```
HarelM: This is not very descriptive in terms of what it does.
Maybe extract the distanceToCenter2D instead of just the normalize delta x?

HarelM:
I haven't seen eslint change white spaces, are you sure this is not your editor?

src/geo/projection/covering_tiles.ts
```diff
@@ -148,9 +148,16 @@ export function createCalculateTileZoomFunction(maxZoomLevelsOnScreen: number, t
         thisTileDesiredZ = thisTileDesiredZ + scaleZoom(distanceToCenter3D / distanceToTile3D / Math.max(0.5, Math.cos(degreesToRadians(cameraVerticalFOV / 2))));
         thisTileDesiredZ += pitchTileLoadingBehavior * scaleZoom(Math.cos(thisTilePitch)) / 2;
         thisTileDesiredZ -= scaleZoom(Math.max(1, tileCount / tileCountPitch0 / tileCountMaxMinRatio)) / 2;
+
+        // add zoom levels for narrow lenses
+        const defaultFov = 36.87;
+        if (cameraVerticalFOV < defaultFov) {
+            thisTileDesiredZ += scaleZoom(defaultFov / cameraVerticalFOV);
+        }
```
Alchez: @thomfuhrmann I vendored this PR locally as a patch for the same issue, and the main change works great. But since I've set an FOV of 15, this particular change is causing ~6x the required tiles to load at every camera position, which I'm not sure is necessary so maybe this block needs rethinking or a proper ticket?

CC @HarelM

thomfuhrmann: @Alchez yes, this needs a proper ticket. This change slipped into this pr accidentally because I worked on that issue in the meantime on the same branch. I will remove these lines.

HarelM:
It might makes sense to move this method to coveringTilesDetailsProvider as I see similar behavior called `distanceToTileWrapX`.
Let me know what you think.

thomfuhrmann:
> It might makes sense to move this method to coveringTilesDetailsProvider as I see similar behavior called `distanceToTileWrapX`. Let me know what you think.

@HarelM yes, that would be a fetter fit.

HarelM (approved):

johncarmack1984:
#8187 (v6.6.0) fixed the same root cause two weeks after the approval here: the camera position used for tile detail is now derived from the center, so camera and center always land in the same world copy and the wrap in this PR no longer changes which tiles are selected (checked over 3000 globe camera setups, largest x delta 0.44 world units). Think this one is closable.

HarelM:
@thomfuhrmann can you check if this PR is still needed? If not, please close it.

HarelM:
Better yet, I'll close it, if you need me to reopen it do let me know.

thomfuhrmann:
@HarelM yes, this is ok! Thx for closing it!

Issue #7828, opened 2026-06-24
Terrain: symbol (label) layers render on top of an overlying raster layer at high zoom


giswqs:
## Summary

With `raster-dem` terrain enabled, the **symbol (label) layers of a vector basemap render on top of an opaque raster layer that sits above them in the style**, instead of being hidden by it. The labels are correctly hidden when terrain is off. The effect only appears at high zoom (roughly z16 and above).

## Minimal reproduction

- Live demo: https://giswqs.github.io/maplibre-terrain-label-bleed/
- Source (single static HTML file, no build, no API keys): https://github.com/giswqs/maplibre-terrain-label-bleed

The repro sets up:

- A vector basemap with labels (OpenFreeMap Liberty)
- An opaque raster layer (Esri Wayback imagery) appended with **no `beforeId`**, so it sits at the very top of the style, above every basemap fill/line/symbol layer
- `raster-dem` terrain (AWS Terrarium tiles)
- View over the White House at ~z17

There is a **Toggle terrain** button to compare the two states directly.

## Steps to reproduce

1. Open the demo (it loads at ~z17 with terrain on).
2. Observe that basemap labels such as "White House", "The West Wing", and "West Colonnade" are visible on top of the imagery.
3. Click **Toggle terrain** to turn terrain off. The labels are now correctly hidden behind the imagery.

## Expected

An opaque raster layer placed above the basemap should fully cover the basemap, including its symbol labels, exactly as it does with terrain off.

## Actual

With terrain on at high zoom, the basemap symbol labels render on top of the overlying raster.

## Analysis

Terrain drapes `background` / `fill` / `line` / `raster` / `hillshade` / `color-relief` layers onto the terrain mesh (render to texture), but symbol layers are never draped: they are drawn live and occluded only by the terrain depth buffer. At high zoom the live symbol fragments appear to win the depth test against the draped raster (which is composited at the terrain surface), so a label survives on top of a raster that is above it in the style order. In 2D (no terrain) the raster simply paints over the labels.

This looks related to #4908 (basemap labels lifting off the terrain surface at high zoom). It may be the same underlying cause viewed from a different symptom; feel free to close as a duplicate if so. Posting separately because this scenario (an overlying raster failing to occlude the labels) is a distinct, easy to verify symptom, and the linked repo is a minimal non-expiring reproduction.

## Environment

- MapLibre GL JS 5.24.0
- Reproduces in Chromium; originally reported on Safari 26 / macOS

## Context

[...]

PR #7852, opened 2026-06-29
Fix symbol occlusion below opaque draped rasters (#7828)

DoFabien:
This PR fixes a concrete terrain rendering ordering bug from #7828.

With terrain enabled, draped layers such as raster/fill/line are rendered to terrain textures, while symbol layers are rendered live. Because of that, labels from a basemap could still appear above an opaque raster layer that was placed later in the style. In 2D, the raster correctly covers those labels.

This patch keeps the fix narrow. `RenderToTexture` tracks the top opaque draped raster layer, and `Painter` skips live symbol rendering only for symbol layers below that raster. Symbols above the raster are still rendered, and `fill-extrusion` rendering is left unchanged so 3D buildings can still appear above the raster.

The fix is scoped to opaque raster overlays and the symbol layers that should be hidden by style order.

before: 
image

after: 
image

## Launch Checklist

 - [x] Link to related issues. #7828
 - [ ] Post benchmark scores.

 Assisted-By:  GPT 5.5

### DIFF
--- a/src/webgl/render_to_texture.test.ts
+++ b/src/webgl/render_to_texture.test.ts
@@ -37,6 +37,9 @@ describe('render to texture', () => {
         id: 'maine-raster',
         type: 'raster',
         source: 'maine',
+        paint: {
+            get: (property: string) => property === 'raster-opacity' ? 1 : undefined
+        },
         isHidden: () => false
     } as any as RasterStyleLayer;
     const hillshadeLayer = {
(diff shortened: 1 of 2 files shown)


DoFabien:
Hmm, I reread issue #7828.

I’m not sure what the expected behavior should be for fill-extrusions: should they be occluded by a raster layer above them, or should they always be extruded/rendered above it?

giswqs:
I think fill-extrusion layers should also be rendered below the raster layer.

clement-igonet (commented):
I tested this on a terrain scene with an opaque image overlay covering only part of the view. Under the overlay it works exactly as described, the floating labels are gone. But the whole symbol layer is hidden, also outside the overlay's footprint: in 2D the labels to the west stay visible, with the patch they disappear everywhere (left is main, right is this PR):

main vs PR, partial-extent opaque overlay

Since raster-opacity says nothing about coverage, maybe the skip needs to be limited to the raster's extent, or scoped to sources that cover the viewport?

Happy to help test approaches.

johncarmack1984:
clement-igonet's 08-02 review stands unanswered: with a partial-extent overlay the patch hides the whole symbol layer, not just the part under the raster, and the fill-extrusion question is still open. #7828 remains open for the actual fix. Think this one is closable.

HarelM:
I'll close this for now if someone would like to propose a different implementation.
To be honest, I'm not sure what should be the right solution here...

